PR #2335, opened 2020-04-25
chore: remove safe-buffer dependency

tinovyatkin:
## Description

`safe-buffer` [does nothing](https://github.com/feross/safe-buffer/blob/7d544741d5a360613abb0d47926e107722fae723/index.js#L11) when native `Buffer.from` API is available and it's available since Nodes.JS 5.10 (https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array), so, there is no sense to bring it as production dependency (one of four) when your declared minimum supported Node.JS version is 10.x (in `master`).


mbroadst:
closing and re-opening to run CI

PR #2336, opened 2020-04-28
NODE-2487: refactors preparing for speculative authentication

mbroadst:
## Description
This is a set of changes that prepares the codebase for the introduction of speculative authentication:
- The concept of multiple authentication has been fully removed, we haven't supported this since 2.x so this is effectively dead code. As a result the API surface of `AuthProvider` has been reduced to a single `auth` method

- The legacy GSSAPI auth provider has been removed, and the modern SSPI implementation has taken its place and name. SSPI is the Win32 implementation of GSSAPI, and both providers actually did exactly the same thing. 

- The `runCommand` implementation living inside `connect.js` was moved to the legacy connection class. This allows us to treat the modern and legacy `Connection` types the same in our handshake pipeline, which greatly simplifies our implementation and looks forward to v4 when we only have one `Connection` to worry about.

- A bug was fixed with `MongoCredentials#resolveAuthMechanism`. The method actually mutated the internals of the credentials passed in, which could be problematic in mixed clusters. If a user does not specify a mechanism, but does provide a username then this would prevent future mechanism resolution once the first resolution completed. If you subsequently connected to a server which did not support the resolved mechanism, the handshake would fail.


lib/core/auth/auth_provider.js
```diff
@@ -10,136 +8,20 @@ const MongoError = require('../error').MongoError;
 class AuthProvider {
   constructor(bson) {
     this.bson = bson;
```
nbbeeken, line 10: We have a bit of an inconsistency here that might be fine to leave in since its been the existing behavior but only AWS auth uses this bson property. Meanwhile PLAIN inits its own BSON so it ignores user overrides. Since the work to take out bson will have to be done in the port to master, if done here it could make the porting work easier.

lib/core/auth/mongo_credentials.js
```diff
@@ -62,6 +62,8 @@ class MongoCredentials {
         this.mechanismProperties.AWS_SESSION_TOKEN = process.env.AWS_SESSION_TOKEN;
       }
     }
+
+    Object.freeze(this);
```
nbbeeken, line 66: Is it worth freezing `this.mechanismProperties` as well?

lib/core/auth/auth_provider.js
nbbeeken: Yea that does sound correct so no need to make changes here.

PR #2338, opened 2020-05-01
fix: hint should raise error on unacknowledged writes

emadum:
## Description

Raise error when hint option is provided on unacknowledged writes
against any server version.

[NODE-2576](https://jira.mongodb.org/browse/NODE-2576)

**What changed?**


lib/core/sdam/server.js
```diff
@@ -454,7 +454,8 @@ function executeWriteOperation(args, options, callback) {
     callback(new MongoError(`server ${server.name} does not support collation`));
     return;
   }
-  if (maxWireVersion(server) < 5) {
+  const unacknowledgedWrite = options && options.writeConcern && options.writeConcern.w === 0;
```
mbroadst, line 457: `options` always exists because of `options = options || {}` above

PR #2339, opened 2020-05-01
NODE-2487: introduce speculative authentication

mbroadst:
## Description
MongoDB 4.4+ supports speculative authentication for the SCRAM-SHA and X509 auth mechanisms, which means that these mechanisms can include the first part of their auth 
conversation in the initial ismaster, potentially reducing one round trip.

**What changed?**
The majority of the work here can be summed up in the following points:
- A `prepare` step was added to `AuthProvider` which allows a provider to manipulate the handshake document before the initial handshake

- An `AuthContext` was introduced to carry methods like `connection`, `credentials` and optionally information a mechanism might need to share between steps of an auth conversation. Currently the only piece of context tracked is the `nonce` generated for SCRAM-SHA. 

None

### DIFF
--- a/lib/core/auth/auth_provider.js
+++ b/lib/core/auth/auth_provider.js
@@ -1,5 +1,14 @@
 'use strict';
 
+/**
+ * Context used during authentication
+ *
+ * @property {Connection} connection The connection to authenticate
+ * @property {MongoCredentials} credentials The credentials to use for authentication
+ * @property {object} options The options passed to the `connect` method
+ * @property {object?} response The response of the initial handshake
+ * @property {Buffer?} nonce A random nonce generated for use in an authentication conversation
+ */
 class AuthContext {
   constructor(connection, credentials, options) {
     this.connection = connection;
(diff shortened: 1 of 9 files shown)


lib/core/auth/auth_provider.js
```diff
@@ -1,40 +1,46 @@
 'use strict';
 
-/**
- * Creates a new AuthProvider, which dictates how to authenticate for a given
- * mechanism.
- * @class
- */
+class AuthContext {
```
reggi, line 3: Is this how we will be creating ts-like interfaces moving forward? I haven't seen this style in the codebase so far. I assumed we were doing option types in JSDOC. [Something like this example here](https://jsdoc.app/tags-typedef.html):

```js
/**
 * The complete Triforce, or one or more components of the Triforce.
 * @typedef {Object} WishGranter~Triforce
 * @property {boolean} hasCourage - Indicates whether the Courage component is present.
 * @property {boolean} hasPower - Indicates whether the Power component is present.
 * @property {boolean} hasWisdom - Indicates whether the Wisdom component is present.
 */
```

lib/core/connection/connect.js
```diff
@@ -96,77 +74,115 @@ function performInitialHandshake(conn, options, _callback) {
...
+      //       relocated, or at very least restructured.
+      conn.ismaster = response;
+      conn.lastIsMasterMS = new Date().getTime() - start;
 
-    const credentials = options.credentials;
-    if (!ismaster.arbiterOnly && credentials) {
-      authenticate(conn, credentials.resolveAuthMechanism(ismaster), callback);
+      if (!response.arbiterOnly && credentials) {
+        // store the response on auth context
+        Object.assign(authContext, { response });
```
reggi, line 141: This kind of side-effect / mutation of state scares me, perhaps it's an irrational fear. I'd prefer some kind of more stream-lined obvious way of doing this that the parent class itself knows about. eg a setter `authContext.response = response` or `authContext.addResponse(response)`. I'd love to having typing somehow on `authContext`  that says it has `{ response?: any }`.

lib/core/auth/auth_provider.js
mbroadst: yes, I'll add that now, but in the future expect to get this type information directly from TypeScript. This is likely a type we will introduce early on considering its very small in scope

mbroadst: I'd added some type information for you here. In the future type information will be handled by actually using TypeScript, this will likely be an early migrated type since its so small in scope.

PR #2340, opened 2020-05-01
NODE-2579/3.6/global-promise-prereq

reggi:
Implements `promise-provider` and gets rid of native calls to `Promise`, each call needs to be  defined within scope uses `eslint-plugin-promise` to enforce. 

Because this is a large commit I tried to have a simple commits to make it easier to review:

### adds promise_provider

This commit adds the new file `promise_provider`.

### tests use promise-provider

This commit adds require calling `promise_provider` to the top of every test that uses a `Promise`, getting rid of global `Promise` call. 

### uses promise provider & no-native

This commit makes calls to `PromiseProvider.get()` in functions. Fixes some missed returns where natives shouldn't have been used. The functionality of `PromiseProvider.get` is customized to take any number of arguments, these arguments are possible stores for `promiseLibrary`. Some changes may have been made that changes current functionality to methods, either by adding optional `promiseLibrary` where it **should** have existed or by adding it where it shouldn't be, open to changing these cases.

### implements `eslint-plugin-promise`

The last commit, adds the `eslint-plugin-promise` rule, which prevents any use of a `Promise` that is not defined in scope. Tested and pass lint.

### DIFF
--- a/lib/admin.js
+++ b/lib/admin.js
@@ -1,7 +1,7 @@
 'use strict';
 
+const PromiseProvider = require('./promise_provider');
 const applyWriteConcern = require('./utils').applyWriteConcern;
-
 const AddUserOperation = require('./operations/add_user');
 const ExecuteDbAdminCommandOperation = require('./operations/execute_db_admin_command');
 const RemoveUserOperation = require('./operations/remove_user');
@@ -42,14 +42,14 @@ const executeOperation = require('./operations/execute_operation');
  * @class
  * @return {Admin} a collection instance.
  */
-function Admin(db, topology, promiseLibrary) {
+function Admin(db, topology) {
   if (!(this instanceof Admin)) return new Admin(db, topology);
 
   // Internal state
   this.s = {
     db: db,
     topology: topology,
-    promiseLibrary: promiseLibrary
+    promiseLibrary: PromiseProvider.get(db, topology)
   };
 }
 
--- a/lib/collection.js
+++ b/lib/collection.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const PromiseProvider = require('./promise_provider');
 const deprecate = require('util').deprecate;
 const deprecateOptions = require('./utils').deprecateOptions;
 const checkCollectionName = require('./utils').checkCollectionName;
@@ -128,7 +129,7 @@ function Collection(db, topology, dbName, name, pkFactory, options) {
   const namespace = new MongoDBNamespace(dbName, name);
 
   // Get the promiseLibrary
-  const promiseLibrary = options.promiseLibrary || Promise;
+  const promiseLibrary = PromiseProvider.get(options, db, topology);
 
   // Set custom primary key factory if provided
   pkFactory = pkFactory == null ? ObjectID : pkFactory;
@@ -443,7 +444,7 @@ Collection.prototype.find = deprecateOptions(
     newOptions.db = this.s.db;
 
     // Add the promise library
-    newOptions.promiseLibrary = this.s.promiseLibrary;
+    newOptions.promiseLibrary = PromiseProvider.get(options, this);
 
     // Set raw if available at collection level
     if (newOptions.raw == null && typeof this.s.raw === 'boolean') newOptions.raw = this.s.raw;
@@ -748,13 +749,14 @@ Collection.prototype.insert = deprecate(function(docs, options, callback) {
  * @return {Promise} returns Promise if no callback passed
  */
 Collection.prototype.updateOne = function(filter, update, options, callback) {
+  const Promise = PromiseProvider.get(this);
   if (typeof options === 'function') (callback = options), (options = {});
   options = options || {};
 
   const err = checkForAtomicOperators(update);
   if (err) {
     if (typeof callback === 'function') return callback(err);
-    return this.s.promiseLibrary.reject(err);
+    return Promise.reject(err);
   }
 
   options = Object.assign({}, options);
@@ -827,13 +829,14 @@ Collection.prototype.replaceOne = function(filter, doc, options, callback) {
  * @return {Promise<Collection~updateWriteOpResult>} returns Promise if no callback passed
  */
 Collection.prototype.updateMany = function(filter, update, options, callback) {
+  const Promise = PromiseProvider.get(this);
   if (typeof options === 'function') (callback = options), (options = {});
   options = options || {};
 
   const err = checkForAtomicOperators(update);
   if (err) {
     if (typeof callback === 'function') return callback(err);
-    return this.s.promiseLibrary.reject(err);
+    return Promise.reject(err);
   }
 
   options = Object.assign({}, options);
@@ -1754,6 +1757,7 @@ Collection.prototype.findOneAndReplace = function(filter, replacement, options, 
  * @return {Promise<Collection~findAndModifyWriteOpResultObject>} returns Promise if no callback passed
  */
 Collection.prototype.findOneAndUpdate = function(filter, update, options, callback) {
+  const Promise = PromiseProvider.get(this);
   if (typeof options === 'function') (callback = options), (options = {});
   options = options || {};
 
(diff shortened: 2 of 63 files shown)


lib/collection.js
```diff
@@ -128,7 +129,7 @@ function Collection(db, topology, dbName, name, pkFactory, options) {
   const namespace = new MongoDBNamespace(dbName, name);
 
   // Get the promiseLibrary
-  const promiseLibrary = options.promiseLibrary || Promise;
+  const promiseLibrary = PromiseProvider.get(options, db, topology);
```
reggi, line 132: Here's a case that I'm taking about. It's a little more comprehensive then it previously was.

lib/cursor.js
```diff
@@ -929,17 +932,15 @@ class Cursor extends CoreCursor {
       }
 
       // Return a Promise
-      return new this.s.promiseLibrary(resolve => {
-        resolve();
-      });
+      return Promise.resolve();
```
reggi, line 935: This was an opinionated change, no reason to call new here just to resolve.

lib/utils.js
```diff
@@ -705,11 +707,11 @@ function* makeCounter(seed) {
  * @returns {Promise|void} Returns nothing if a callback is supplied, else returns a Promise.
  */
 function maybePromise(parent, callback, fn) {
-  const PromiseLibrary = (parent && parent.s && parent.s.promiseLibrary) || Promise;
+  const Promise = PromiseProvider.get(parent);
```
reggi, line 710: Much cleaner here 😍

test/functional/promise_provider.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+const PromiseProvider = require('../../lib/promise_provider');
+
+const candidates = {
+  withOptions: { options: { promiseLibrary: Symbol('withOptions') } },
+  withClientOptions: { clientOptions: { promiseLibrary: Symbol('withClientOptions') } },
+  withS: { s: { promiseLibrary: Symbol('withS') } }
+};
+
+describe('PromiseProvider', () => {
```
reggi, line 12: This is a test for `PromiseProvider` just testing for the reducer, ensuring it works as expected.

lib/collection.js
```diff
@@ -443,7 +444,7 @@ Collection.prototype.find = deprecateOptions(
     newOptions.db = this.s.db;
 
     // Add the promise library
-    newOptions.promiseLibrary = this.s.promiseLibrary;
+    newOptions.promiseLibrary = PromiseProvider.get(options, this);
```
reggi, line 447: This is within `collection.find` it's not clear to me which methods you can pass in `promiseLibrary` and which you can't. Should this be `PromiseProvider.get(this);` or `PromiseProvider.get(options, this);`? Thoughts?

lib/admin.js
```diff
@@ -42,14 +42,14 @@ const executeOperation = require('./operations/execute_operation');
  * @class
  * @return {Admin} a collection instance.
  */
-function Admin(db, topology, promiseLibrary) {
```
reggi, line 45: This seems unnecessary and non-standard.

test/benchmarks/driverBench/index.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const Promise = require('../../../lib/promise_provider').get();
```
mbroadst, line 3: is it possible to use `ignorePatterns` to ignore files in `test` for this rule? I don't think it buys us anything to force these declarations in our test files, we only want it in our `lib` files to provide support to our users.

lib/collection.js
```diff
@@ -1990,7 +1994,7 @@ Collection.prototype.parallelCollectionScan = deprecate(function(options, callba
   options.readPreference = resolveReadPreference(this, options);
 
   // Add a promiseLibrary
-  options.promiseLibrary = this.s.promiseLibrary;
```
emadum, line 1993: I'm a little confused about what setting `options.promiseLibrary` before running `executeLegacyOperation` does. As you can see [here](https://github.com/mongodb/node-mongodb-native/pull/2340/files#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR377), `executeLegacyOperation` ignores the `options` and grabs the `promiseLibrary` from the `topology`. If I'm not missing something, we can just delete anything that sets `options.promiseLibrary` before a call to `executeLegacyOperation`.

lib/admin.js
```diff
@@ -42,14 +42,14 @@ const executeOperation = require('./operations/execute_operation');
...
+function Admin(db, topology) {
   if (!(this instanceof Admin)) return new Admin(db, topology);
 
   // Internal state
   this.s = {
     db: db,
     topology: topology,
-    promiseLibrary: promiseLibrary
+    promiseLibrary: PromiseProvider.get(db, topology)
```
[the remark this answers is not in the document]
reggi: It is my understanding that each time a method takes-in `promiseLibrary` it set's it for it's class-scope and it's-children. And I've tried to illustrate [this in this example](https://github.com/reggi/node-mongodb-native-playground/blob/master/NODE-2579/index.js#L27-L43). I thought we were all in-agreement that 3.6 can't change this functionality. 

```js
// here mongoClient sets `bluebird`
const client = await MongoClient("mongodb://127.0.0.1", {
    promiseLibrary: bluebird,
    useUnifiedTopology: true
});

const db = client.db("test");

// here in collection I'm setting the promiseLibrary to `q`
const collectionPromise = db.collection("test", { promiseLibrary: q }).find().toArray();

```

I would assume here the driver would provide any promises coming from `client` to be bluebird and promises coming from `collectionPromise` to be `q`.

This is the "parody" i believed we we were trying to support in 3.6

If we use a "global" setter / getter, and don't store in-parent-class and pass options, we loose this tracking of what class is using what.

My plan was to remove the stores in master and only `.set()` in `MongoClient`.

This is related to the [private-slack thread here](https://mongodb.slack.com/archives/GGWBN4ZNK/p1588188285028400?thread_ts=1588186585.026000&cid=GGWBN4ZNK).

test/benchmarks/driverBench/index.js
reggi: This is the opposite of what you [mentioned earlier](https://mongodb.slack.com/archives/GGWBN4ZNK/p1588272216015300?thread_ts=1588271504.011700&cid=GGWBN4ZNK) regarding adding these to tests because it' allows us to test the driver using multiple promise implementations.

I'm not sure if it's possible for `eslint` to `ignore` a specific rule (`eslint-plugin-promise`) for a specific folder (tests). Without possibly having two different `eslint` configs.

lib/promise_provider.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Get the current promise constructor
+ *
+ * @api private
+ */
+
+store.get = function() {
+  const values = Object.keys(arguments).map(k => arguments[k]);
+  const result = values.reduce((acq, parent) => {
```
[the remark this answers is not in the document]
reggi: Thanks, adopting change.

lib/promise_provider.js
```diff
@@ -0,0 +1,59 @@
...
+
+const assert = require('assert');
+
+/**
+ * Helper for multiplexing promise implementations
+ *
+ * @api private
+ */
+
+const store = {
```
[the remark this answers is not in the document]
reggi: I ended up copping mongooses implementation directly after previous comments regarding preferring their file-name convention, and scrapping the getter implementation I had earlier developer. I thought we were trying not to veer off what they had too much. I'd be down to use a class.

lib/collection.js
```diff
@@ -748,13 +749,14 @@ Collection.prototype.insert = deprecate(function(docs, options, callback) {
...
 Collection.prototype.updateOne = function(filter, update, options, callback) {
+  const Promise = PromiseProvider.get(this);
   if (typeof options === 'function') (callback = options), (options = {});
   options = options || {};
 
   const err = checkForAtomicOperators(update);
   if (err) {
     if (typeof callback === 'function') return callback(err);
-    return this.s.promiseLibrary.reject(err);
+    return Promise.reject(err);
```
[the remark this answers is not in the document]
reggi: Should that change be in this PR?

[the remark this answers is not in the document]
mbroadst: no, just calling it out

lib/promise_provider.js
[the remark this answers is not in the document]
mbroadst: let's use a class!

PR #2322, opened 2020-04-23
refactor(ChangeStream): use maybePromise for next/hasNext

emadum:
## Description

[NODE-2570](https://jira.mongodb.org/browse/NODE-2570)

**What changed?**

**Are there any files to ignore?**

PR #1902, opened 2019-01-18
fix(changeStream): properly handle changeStream event mid-close

daprahamian:
If a changeStream gets an event while it is in the middle of closing,
a race condition can occur where the event is still processed after
the stream has closed. This commit adds handling for this edge
case by returning an error to callbacks, rejecting promises,
and simply ignoring it in emitter mode.

Fixes NODE-1831

PR #2343, opened 2020-05-03
refactor(ChangeStream): use maybePromise for close, improve tests

emadum:
## Description

[NODE-2598](https://jira.mongodb.org/browse/NODE-2598)

**What changed?**

 - Refactor `close` to use `maybePromise` helper
 - Ensure explicitly closed change streams don't process further changes
 - Fix broken `mid-close` tests by delaying the final write
 - Add `waitForStarted` helper to tests where appropriate to reduce flakiness

### DIFF
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -143,7 +143,7 @@ class ChangeStream extends EventEmitter {
   next(callback) {
     return maybePromise(this.parent, callback, cb => {
       if (this.isClosed()) {
-        return cb(new Error('Change Stream is not open.'));
+        return cb(new MongoError('ChangeStream is closed.'));
       }
       this.cursor.next((error, change) => {
         processNewChange({ changeStream: this, error, change, callback: cb });
@@ -475,11 +475,8 @@ function processNewChange(args) {
     if (eventEmitter) {
       return;
     }
-
-    const error = new MongoError('ChangeStream is closed');
-    return typeof callback === 'function'
-      ? callback(error, null)
-      : changeStream.promiseLibrary.reject(error);
+    callback(new MongoError('ChangeStream is closed.'));
+    return;
   }
 
   const topology = changeStream.topology;
@@ -536,8 +533,7 @@ function processNewChange(args) {
     }
 
     if (eventEmitter) return changeStream.emit('error', error);
-    if (typeof callback === 'function') return callback(error, null);
-    return changeStream.promiseLibrary.reject(error);
+    return callback(error, null);
   }
 
   changeStream.attemptingResume = false;
@@ -548,8 +544,7 @@ function processNewChange(args) {
     );
 
     if (eventEmitter) return changeStream.emit('error', noResumeTokenError);
-    if (typeof callback === 'function') return callback(noResumeTokenError, null);
-    return changeStream.promiseLibrary.reject(noResumeTokenError);
+    return callback(noResumeTokenError, null);
   }
 
   // cache the resume token
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -157,7 +157,7 @@ class ChangeStream extends EventEmitter {
    * @return {boolean}
    */
   isClosed() {
-    return (this.cursor && this.cursor.isClosed()) || this.closed;
+    return this.closed || (this.cursor && this.cursor.isClosed());
   }
 
   /**
@@ -169,6 +169,8 @@ class ChangeStream extends EventEmitter {
   close(callback) {
     return maybePromise(this.parent, callback, cb => {
       if (this.closed) return cb();
+
+      // flag the change stream as explicitly closed
       this.closed = true;
 
       // Tidy up the existing cursor
@@ -463,7 +465,7 @@ function processNewChange(args) {
   const eventEmitter = args.eventEmitter || false;
   const cursor = changeStream.cursor;
 
-  // If the cursor is null, then it should not process a change.
+  // If the cursor is null or the change stream has been closed explictly, do not process a change.
   if (cursor == null || changeStream.closed) {
     // We do not error in the eventEmitter case.
     changeStream.closed = true;
@@ -492,6 +494,7 @@ function processNewChange(args) {
 
       waitForTopologyConnected(topology, { readPreference: options.readPreference }, err => {
         if (err) {
+          // if there's an error reconnecting, close the change stream
           changeStream.closed = true;
           if (eventEmitter) {
             changeStream.emit('error', err);
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -443,7 +443,7 @@ function applyKnownOptions(target, source, optionNames) {
 // ChangeStream resumability until the new SDAM layer can be used.
 const SELECTION_TIMEOUT = 30000;
 function waitForTopologyConnected(topology, options, callback) {
-  if (isUnifiedTopology(topology)) return callback();
+  // if (isUnifiedTopology(topology)) return callback();
   setTimeout(() => {
     if (options && options.start == null) options.start = process.hrtime();
     const start = options.start || process.hrtime();
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -6,7 +6,7 @@ const MongoError = require('./core').MongoError;
 const Cursor = require('./cursor');
 const relayEvents = require('./core/utils').relayEvents;
 const maxWireVersion = require('./core/utils').maxWireVersion;
-const isUnifiedTopology = require('./core/utils').isUnifiedTopology;
+// const isUnifiedTopology = require('./core/utils').isUnifiedTopology;
 const maybePromise = require('./utils').maybePromise;
 const AggregateOperation = require('./operations/aggregate');
 
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -6,7 +6,6 @@ const MongoError = require('./core').MongoError;
 const Cursor = require('./cursor');
 const relayEvents = require('./core/utils').relayEvents;
 const maxWireVersion = require('./core/utils').maxWireVersion;
-// const isUnifiedTopology = require('./core/utils').isUnifiedTopology;
 const maybePromise = require('./utils').maybePromise;
 const AggregateOperation = require('./operations/aggregate');
 
@@ -443,7 +442,6 @@ function applyKnownOptions(target, source, optionNames) {
 // ChangeStream resumability until the new SDAM layer can be used.
 const SELECTION_TIMEOUT = 30000;
 function waitForTopologyConnected(topology, options, callback) {
-  // if (isUnifiedTopology(topology)) return callback();
   setTimeout(() => {
     if (options && options.start == null) options.start = process.hrtime();
     const start = options.start || process.hrtime();
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -6,6 +6,7 @@ const MongoError = require('./core').MongoError;
 const Cursor = require('./cursor');
 const relayEvents = require('./core/utils').relayEvents;
 const maxWireVersion = require('./core/utils').maxWireVersion;
+const isUnifiedTopology = require('./core/utils').isUnifiedTopology;
 const maybePromise = require('./utils').maybePromise;
 const AggregateOperation = require('./operations/aggregate');
 
@@ -442,6 +443,7 @@ function applyKnownOptions(target, source, optionNames) {
 // ChangeStream resumability until the new SDAM layer can be used.
 const SELECTION_TIMEOUT = 30000;
 function waitForTopologyConnected(topology, options, callback) {
+  if (isUnifiedTopology(topology)) return callback();
   setTimeout(() => {
     if (options && options.start == null) options.start = process.hrtime();
     const start = options.start || process.hrtime();
@@ -453,7 +455,7 @@ function waitForTopologyConnected(topology, options, callback) {
     const elapsed = (hrElapsed[0] * 1e9 + hrElapsed[1]) / 1e6;
     if (elapsed > timeout) return callback(new MongoError('Timed out waiting for connection'));
     waitForTopologyConnected(topology, options, callback);
-  }, 3000); // this is an arbitrary wait time to allow SDAM to transition
+  }, 500); // this is an arbitrary wait time to allow SDAM to transition
 }
 
 // Handle new change events. This method brings together the routes from the callback, event emitter, and promise ways of using ChangeStream.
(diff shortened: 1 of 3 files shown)


lib/change_stream.js
```diff
@@ -158,7 +160,7 @@ class ChangeStream extends EventEmitter {
     if (this.cursor) {
       return this.cursor.isClosed();
     }
-    return true;
+    return this.closed;
```
mbroadst, line 163: I think we should _only_ return `this.close`, instead of deferring to the underlying cursor.

lib/change_stream.js
```diff
@@ -168,31 +170,21 @@ class ChangeStream extends EventEmitter {
...
   close(callback) {
-    if (!this.cursor) {
-      if (callback) return callback();
-      return this.promiseLibrary.resolve();
-    }
-
-    // Tidy up the existing cursor
-    const cursor = this.cursor;
+    return maybePromise(this.parent, callback, cb => {
+      this.closed = true;
```
mbroadst, line 174:
```
if (this.closed) return cb();
```

lib/change_stream.js
```diff
@@ -168,31 +170,21 @@ class ChangeStream extends EventEmitter {
...
 
-    if (callback) {
-      return cursor.close(err => {
-        ['data', 'close', 'end', 'error'].forEach(event => cursor.removeAllListeners(event));
-        delete this.cursor;
+      if (!this.cursor) {
```
mbroadst, line 176: I think you can remove this line after the above change is made

lib/change_stream.js
```diff
@@ -168,31 +170,21 @@ class ChangeStream extends EventEmitter {
...
-    }
+      // Tidy up the existing cursor
+      const cursor = this.cursor;
 
-    const PromiseCtor = this.promiseLibrary || Promise;
-    return new PromiseCtor((resolve, reject) => {
-      cursor.close(err => {
+      return cursor.close(err => {
         ['data', 'close', 'end', 'error'].forEach(event => cursor.removeAllListeners(event));
         delete this.cursor;
```
mbroadst, line 185: nit: `this.cursor = undefined;` (I know you didn't introduce this change)

test/functional/change_stream.test.js
```diff
@@ -2038,17 +2036,19 @@ describe('Change Streams', function() {
...
-            changeStream.next(err => {
-              let _err = null;
-              try {
-                expect(err.message).to.equal('ChangeStream is closed');
-              } catch (e) {
-                _err = e;
-              } finally {
-                done(_err);
-              }
+            changeStream.close(err => {
```
mbroadst, line 2039: this doesn't seem to jive with the name of the test, it seems like it more wants something like a race between `close` and `next`, e.g.
```
changeStream.next(() => {
  changeStream.close(() => {
  });

  changeStream.next(err => {
    expect(err).to.exist; // ...
    done();
  });
});
```

lib/change_stream.js
```diff
@@ -556,8 +525,7 @@ function processNewChange(args) {
     );
 
     if (eventEmitter) return changeStream.emit('error', noResumeTokenError);
-    if (typeof callback === 'function') return callback(noResumeTokenError, null);
-    return changeStream.promiseLibrary.reject(noResumeTokenError);
+    return callback(noResumeTokenError, null);
```
mbroadst, line 528: nit since we're here anyway: `return callback(noResumeTokenError);` (no `null`). This also applies to a number of these callbacks here.

lib/change_stream.js
```diff
@@ -506,46 +495,26 @@ function processNewChange(args) {
...
-          resolve();
-        });
-      })
-        .then(
-          () => (changeStream.cursor = createChangeStreamCursor(changeStream, cursor.resumeOptions))
-        )
-        .then(() => changeStream.next());
+        changeStream.cursor = createChangeStreamCursor(changeStream, cursor.resumeOptions);
+        if (eventEmitter) return;
+        changeStream.next(callback);
```
mbroadst, line 511: I think we still have a bug here if the thing that causes the recreate is `hasNext` (I think we discussed this the other day), I think a safer replacement would be to call `hasNext` here since that should cache the results.

lib/change_stream.js
```diff
@@ -483,11 +475,8 @@ function processNewChange(args) {
     if (eventEmitter) {
       return;
     }
-
-    const error = new MongoError('ChangeStream is closed');
-    return typeof callback === 'function'
-      ? callback(error, null)
-      : changeStream.promiseLibrary.reject(error);
+    callback(new MongoError('ChangeStream is closed.'));
```
mbroadst, line 478: nit: no period at the end of the message (acknowledging that you didn't introduce this)

lib/change_stream.js
emadum: I tried to do this, but the test `Should invalidate change stream on database drop using imperative callback form` starts failing. The underlying cursor closes due to the `invalidate` event, but `ChangeStream` doesn't have any way of knowing that except calling `this.cursor.isClosed()`. I suspect there's a way around that, but the current approach seems to work well enough.

test/functional/change_stream.test.js
emadum: Yeah, I suspected that was the case, but then I don't really understand the reasoning behind this test. Without the change I made, the test flakes out because everything works; the change event gets processed before the close. However the test requires that the close happen before the change event - why? Wouldn't it be better it if the change event happens before the close?  I suspect I'm missing something here.

lib/change_stream.js
emadum: I was planning on deferring any logical changes to `ChangeStream` unrelated to the new `closed` property to an immediate follow-up PR to this, for [NODE-2548](https://jira.mongodb.org/browse/NODE-2548). Since `processNewChange` was using `next` before, I'd like to keep that for this PR; changing that to `hasNext` isn't in this one's scope. LMK if you disagree!

test/functional/change_stream.test.js
```diff
@@ -2017,11 +2017,7 @@ describe('Change Streams', function() {
           return Promise.resolve()
             .then(() => changeStream.next())
             .then(() => changeStream.next())
-            .then(() => {
-              const nextP = changeStream.next();
-
-              return changeStream.close().then(() => nextP);
-            });
+            .then(() => Promise.all([changeStream.close(), changeStream.next()]));
```
mbroadst, line 2020: This is a behavioral change, it used to be that `next` was called (executing soon), and then the `close` call was made, after which next was awaited. Now you're essentially racing both `close` and `next`, and `close` comes first. 

Was there something you're working around here?

lib/change_stream.js
mbroadst: Hmm, okay yeah that makes sense. If the cursor receives a `getMore` response with `cursorId: 0` then it will close the cursor but we won't find out about that until we call this. We'd only be able to track the `close` event, but that would force the cursor into streaming mode.

test/functional/change_stream.test.js
emadum: I thought I wrote a comment about that but I don't see it anymore, odd. 

> it used to be that next was called (executing soon), and then the close call was made, after which next was awaited

The test fails under those conditions. The final next returns the change when it is awaited, and the test expects an error. I can only get it to return an error if I call `close` first. I'm still not exactly sure what the desired behavior is that this test is trying to confirm happens.

mbroadst: Here's the pull request where it was introduced, for more context: https://github.com/mongodb/node-mongodb-native/pull/1902

I think the idea is that we want to test the behavior when a new change comes in while the user is actively closing the change stream. It might just be an edge case that was possible before some fix that you introduced in this PR? It also seems like a test that could only really be handled through mocking or clever use of fail points, do you think we should remove the test?

emadum: Yeah this is all making more sense now. I think the difference in behavior is derived from [this change](https://github.com/mongodb/node-mongodb-native/pull/2294/files#diff-30861ceb96e5170099331504d187951eR489), at which point they started flaking out until I skipped them in #2322. 

If I'm understanding correctly, this test was checking to make sure that `processNewChange` would throw an error if the underlying cursor closed. However, that shouldn't really be the case - the underlying cursor closes during a resumable error, but the stream shouldn't error.

So yeah, if that's correct I'm all for removing the test.

mbroadst: I _think_ I agree, but now I'm wondering more about the `isClosed` dilemma also addressed in this PR. `processNewChange` has:
```
  if (cursor == null) {
    changeStream.closed = true;
    if (eventEmitter) {
      return;
    }

    callback(new MongoError('ChangeStream is closed'));
    const error = new MongoError('ChangeStream is closed');
    return;
  }
```

But what happens if `changeStream.closed === true` (i.e. the change stream was explicitly closed)? Shouldn't that trigger this path as well? Is that the behavior this test is testing?

emadum: `next` already checks if `changeStream.closed === true` and returns an error before calling `processNewChange`. I confirmed by adding this to `processNewChange` and the test suite ran with no errors.
```
  if (changeStream.closed) {
    throw new Error('Should never get here');
  }
```

emadum: That said, I'm thinking I should flip the or condition in `isClosed`.

If it was `return this.closed || (this.cursor && this.cursor.isClosed());` instead of the other way around, `closed === true` would take precedence over the cursor state.

mbroadst: `next` checks this before calling `next` on the cursor, which in turn calls `processNewChange`. What if the change stream is not closed when you call `next`, but is closed sometime before the result of `cursor.next` returns? You might be able to test this with a mock server (or a sinon mock if we had a more complex fixture in place):
```
server.setResponseHandler(res => {
   const doc = res.doc;
   if (doc.ismaster) {
     res.reply({ .. ismaster .. });
   } else if (doc.aggregate) {
     res.reply({ ok: 1, .. other aggregate stuff .. });
   } else if (doc.getMore) {
     setTimeout(() => res.reply({ ok: 1, .. other getmore stuff .. }, 200); // always respond with latency to `getMore` calls
   }
});
```

now what does this do:
```
const changeStream = coll.watch();
changeStream.next(err => { // first call will send `aggregate`
  expect(err).to.not.exist;

  changeStream.next(err => {  // second call will send `getMore`
    expect(err).to.exist; // I think we _do_ expect to get an error here, change stream was closed explicitly. 
    done();
  });

  changeStream.close(); // immediately close the change stream
});
```

It may be that the underlying cursor logic already protects us from this, but I think that this is the behavior the test is meaning to cover. I just want to make sure we're not losing that coverage here.

emadum: That makes sense, I'll check into it.

test/functional/change_stream.test.js
```diff
@@ -2609,34 +2619,24 @@ describe('Change Streams', function() {
   describe('tryNext', function() {
```
emadum, line 2619: I oddly started getting an error in the `tryNext` tests:

`MongoError: cannot open $changeStream for non-existent database: test`

I thought it would be more robust to use a specifically named database and drop it after each test, to ensure weird behavior doesn't carry over between tests. To that purpose I created the `withTempDb` helper that wraps `withClient` to create and drop the specified database. This refactor made above error go away.

test/functional/change_stream.test.js
emadum: I'm thinking it would make sense to create a separate ticket to fix those two `mid-close` tests. They were already skipped before this PR, so keeping them skipped until a followup PR doesn't hurt IMO. Thoughts, @mbroadst?

test/functional/change_stream.test.js
```diff
@@ -37,7 +38,11 @@ function triggerResumableError(changeStream, onCursorClosed) {
  * @param {function} callback
  */
 function waitForStarted(changeStream, callback) {
+  const timeout = setTimeout(() => {
+    throw new Error('Change stream never started');
+  }, 2000);
```
emadum, line 43: This _shouldn't_ do anything. I'm just concerned some tests are timing out after 60s because for some reason this never triggers. If that's the case, this will at least fail a lot sooner, and more importantly we'll know for sure whether this is the root cause of test flakiness.

test/functional/change_stream.test.js
```diff
@@ -54,9 +59,12 @@ function tryNext(changeStream, callback) {
   let complete = false;
   function done(err, result) {
     if (complete) return;
-
     // if the arity is 1 then this a callback for `more`
     if (arguments.length === 1) {
+      if (err instanceof Error) {
```
mbroadst, line 64: in what case is an error returned here? When I look at where `more` is emitted in `change_stream.js`, the errors are always handled. This would seem to imply the `getMore` callback was called with `err, err` or something.

test/functional/change_stream.test.js
```diff
@@ -2609,34 +2638,24 @@ describe('Change Streams', function() {
...
-        client.connect(err => {
-          expect(err).to.not.exist;
-
-          const changeStream = client
-            .db()
-            .collection('test')
-            .watch();
-
+      test: function() {
+        return withTempDb('testTryNext', this.configuration.newClient(), db => done => {
```
mbroadst, line 2642: thoughts about the helpers:
- can `withClient` be modified to not need the `this.configuration.newClient` parameter if `this.configuration` exists?
```
withClient((client, done) => {
...
});
```
- I'm a little wary of functions that return functions (ad infinitum) in cases like this because it makes the test a bit hard to reason about. To be clear, I think what happens behind the scenes is totally fair game, but what we expose to the user (us) should be super intuitive, something I jump to use. What about composability:
```
withClient(withTempDb('testTryNext', (client, db, done) => {
  ... test ...
}));
```

test/functional/shared.js
```diff
@@ -75,6 +75,17 @@ function makeCleanupFn(client) {
   };
 }
 
+function withTempDb(dbName, client, operation, errorHandler) {
+  return withClient(
+    client,
+    client => done => {
+      const db = client.db(dbName);
+      operation.call(this, db)(() => db.dropDatabase(done));
```
mbroadst, line 83: I imagine the reason you ran into this in the first place is because the database wasn't created with a wc: majority. We might want to ensure the same here

test/functional/change_stream.test.js
emadum: I agree it's worth thinking these helpers through a bit more. I had to make some modifications to `withMonitoredClient` in #2354 which could also be more elegant/usable.

test/functional/change_stream.test.js
emadum: Oops I think I should remove this now, I was just trying to track down an issue but it's no longer necessary.

PR #2344, opened 2020-05-03
NODE-2559: detect stale errors

mbroadst:
## Description
This is the second part of the upcoming streamable ismaster changes: detecting stale connections and errors in SDAM error handling. 

**What changed?**
The commits are generally piecemeal in describing the changes, but they can be broken down into the following sections:
- A `MongoNetworkTimeoutError` was introduced, with an ability to determine if it occurred before the handshake completed or not
- `topologyVersion` comparison rules were slightly modified to ignore nullish values
- The SDAM unit test runner had a large refactor in order to be able to simulate incoming errors from application connections. This is a risky refactor, but the reward is great in terms of the code coverage it adds for the error handling bits of our code.

There are something like ~130 new tests added for error handling scenarios, you can completely ignore these.


lib/core/error.js
```diff
@@ -83,9 +88,28 @@ class MongoError extends Error {
  * @extends MongoError
  */
 class MongoNetworkError extends MongoError {
-  constructor(message) {
+  constructor(message, options) {
     super(message);
     this.name = 'MongoNetworkError';
+
+    if (options && !!options.beforeHandshake) {
```
emadum, line 95: As I understand it, `!!` is used to coerce a value into boolean based on its truthiness, i.e. `!!value` is equivalent to `value ? true : false`. Is there a purpose to doing this as part of a logical expression inside an `if` condition? I'd think this `if` should behave the same with and without the `!!`.

lib/cmap/connection.js
```diff
@@ -75,10 +76,14 @@ class Connection extends EventEmitter {
       stream.destroy();
       this.closed = true;
       this[kQueue].forEach(op =>
-        op.cb(new MongoNetworkError(`connection ${this.id} to ${this.address} timed out`))
+        op.cb(
+          new MongoNetworkTimeoutError(`connection ${this.id} to ${this.address} timed out`, {
```
mbroadst, line 80: So I introduced a new type `MongoNetworkTimeoutError`, and I wanted to see what y'all thought about its usefulness? It helps us here by letting us switch between subtle behaviors in SDAM related to error types, but I could easily have left in the `isNetworkTimeoutError` helper to detect that (or even pass `timeout: true` into the options of the error type here). Is it useful to users that they have a `MongoNetworkError` and a `MongoNetworkTimeoutError`? I kind of lean towards not introducing a new type, but I couldn't quite make myself do it before submitting this. 

/cc @emadum @reggi

lib/core/sdam/server.js
```diff
@@ -476,26 +484,39 @@ function markServerUnknown(server, error) {
...
+function makeOperationHandler(server, connection, options, callback) {
   const session = options && options.session;
 
   return function handleOperationResult(err, result) {
-    if (err) {
+    if (err && !connectionIsStale(server.s.pool, connection)) {
       if (err instanceof MongoNetworkError) {
         if (session && !session.hasEnded) {
-          session.serverSession.isDirty = true;
+          options.session.serverSession.isDirty = true;
```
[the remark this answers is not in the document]
mbroadst: hm no I think this was mistakenly introduced because I worked on the change before the patch to `isDirty` a few weeks back. I'll switch it back.

lib/core/error.js
mbroadst: I don't think it matters all too much how this is written, but perhaps a `options && options.beforeHandshake === true` is the best approach, I'll make the change

lib/cmap/connection.js
emadum: I think we could definitely go overboard with adding new types for different errors, but I think timeout errors are distinct enough it could be useful. There are some nice shortcuts you can take advantage of when you have different classes for errors, e.g. [filtered catch in bluebird](http://bluebirdjs.com/docs/api/catch.html#filtered-catch).

mbroadst: Right, but I guess my question is: would a user ever want to catch a `MongoNetworkTimeoutError` explicitly? Try to imagine doing it yourself, whats an actual use case where I would want to catch a timeout error but not a network error?

emadum: Definitely seems like an edge case... 

Perhaps on a very spotty network, where a user wants to automatically retry certain operations in case of a timeout error (expected to occur regularly) but log it as a failure for other network errors (which do not occur regularly)?

mbroadst: right.. the original helper was:
```
function isNetworkRetryError(err) {
  return err.message.match(/timed out/);
}
```

I think we want to move towards a world where users aren't implementing their own retry logic, but rather will depend on the driver to have robust exponential retry. In my experience most people don't even use the type hierarchy but rather are informed by the text of the error. That might prove the point in both directions: if the message is all that's important then we haven't actually changed anything behaviorally here. Yet, something about introducing a new type often rubs me the wrong way; I increased the API surface we need to support for something that has potentially marginal value to the user.

reggi: I generally prefer custom error classes rather than having error string-messages or codes, because I think it's more convenient to do an `instanceof` then that sad message match. I'm all for more classes and more specificity.

reggi:
What is the spec commit used to pull in these changes?

mbroadst:
@reggi SPEC-1663

reggi:
I ran my own spec migration script and found three files from spec `371e95b18f4783d485218bb1565970be10568e8a` that weren't fully imported.

Here are the files:

```
        modified:   test/spec/server-discovery-and-monitoring/README.rst
        modified:   test/spec/server-discovery-and-monitoring/single/direct_connection_wrong_set_name.json
        modified:   test/spec/server-discovery-and-monitoring/single/direct_connection_wrong_set_name.yml
```

lib/core/utils.js
```diff
@@ -160,6 +160,34 @@ function eachAsync(arr, eachFn, callback) {
...
+  arr = arr || [];
+
+  let idx = 0;
+  let awaiting = arr.length;
+  if (awaiting === 0) {
+    callback();
+    return;
+  }
+
+  eachFn(arr[idx], eachCallback);
```
reggi, line 173: nit: It's my preferences this should be below the `eachCallback` not to violate [`no-use-before-define`](https://eslint.org/docs/rules/no-use-before-define). If we're not using that rule then it's ok.

lib/cmap/connection.js
mbroadst: ok, it sounds like consensus is that we stick with the new error type, resolving this comment

lib/core/utils.js
mbroadst: fixed

mbroadst:
@reggi the unmerged files you found will be merged as part of the work in [NODE-2564](https://jira.mongodb.org/browse/NODE-2564)

PR #2348, opened 2020-05-04
NODE-2579/master/global-promise-lib

reggi:
Implements global promise lib


.eslintrc
```diff
@@ -29,8 +29,17 @@
 
     "no-console": "off",
     "eqeqeq": ["error", "always", { "null": "ignore" }],
-    "strict": ["error", "global"]
+    "strict": ["error", "global"],
+    "promise/no-native": "error"
   },
+  "overrides": [
```
reggi, line 35: Adds `promise/no-native` lint requirement and ignores test folders.

index.js
```diff
@@ -1,12 +1,23 @@
...
 const error = require('./lib/error');
 const Instrumentation = require('./lib/apm');
 const { BSON } = require('./lib/deps');
 const { Cursor, AggregationCursor, CommandCursor } = require('./lib/cursor');
+const PromiseProvider = require('./lib/promise_provider');
 
 // Set up the connect function
 const connect = require('./lib/mongo_client').connect;
 
+Object.defineProperty(connect, 'Promise', {
```
reggi, line 12: Allows for `mongodb.Promise = Bluebird`

lib/db.js
```diff
@@ -119,7 +119,7 @@ const legalOptionNames = [
  * @param {number} [options.bufferMaxEntries=-1] Sets a cap on how many operations the driver will buffer up before giving up on getting a working connection, default is -1 which is unlimited.
  * @param {(ReadPreference|string)} [options.readPreference] The preferred read preference (ReadPreference.PRIMARY, ReadPreference.PRIMARY_PREFERRED, ReadPreference.SECONDARY, ReadPreference.SECONDARY_PREFERRED, ReadPreference.NEAREST).
  * @param {object} [options.pkFactory] A primary key factory object for generation of custom _id keys.
- * @param {object} [options.promiseLibrary] A Promise library class the application wishes to use such as Bluebird, must be ES6 compatible
+ * @deprecated {object} [options.promiseLibrary] A Promise library class the application wishes to use such as Bluebird, must be ES6 compatible
```
reggi, line 122: Unsure how to deprecate a parameter.

reggi:
evergreen is down, will run tests when back up

lib/db.js
emadum: There's a [deprecateOptions util](https://github.com/mongodb/node-mongodb-native/blob/2a3bb7cb636d1551abe0f4c3ccdde90170274ab4/lib/utils.js#L598), sample usage [here](https://github.com/mongodb/node-mongodb-native/blob/master/lib/db.js#L484)

emadum: There's a helper, sample usage [here](https://github.com/mongodb/node-mongodb-native/blob/master/lib/db.js#L484) - looks like you just add a `DEPRECATED:` prefix in the jsdoc description text.

.eslintrc
mbroadst: very nice! Can't you use glob syntax to ignore a single pattern:
```
"files": ["test/**/*.js"]
```

lib/db.js
```diff
@@ -118,7 +118,7 @@ const legalOptionNames = [
  * @param {number} [options.bufferMaxEntries=-1] Sets a cap on how many operations the driver will buffer up before giving up on getting a working connection, default is -1 which is unlimited.
  * @param {(ReadPreference|string)} [options.readPreference] The preferred read preference (ReadPreference.PRIMARY, ReadPreference.PRIMARY_PREFERRED, ReadPreference.SECONDARY, ReadPreference.SECONDARY_PREFERRED, ReadPreference.NEAREST).
  * @param {object} [options.pkFactory] A primary key factory object for generation of custom _id keys.
- * @deprecated {object} [options.promiseLibrary] A Promise library class the application wishes to use such as Bluebird, must be ES6 compatible
+ * @param {object} [options.promiseLibrary] DEPRECATED: A Promise library class the application wishes to use such as Bluebird, must be ES6 compatible
```
reggi, line 121: Switched back to @param, and added deprecated, it's a bummer this isn't more comprehensive because `promiseLibrary` will still show in type completion.

lib/db.js
reggi: Tried to implement, tests failed, my understanding is this is possible for methods, but not constructors.

lib/db.js
```diff
@@ -131,47 +131,54 @@ const legalOptionNames = [
...
-  validateDatabaseName(databaseName);
-
-  // Add a read Only property
-  getSingleProperty(this, 'serverConfig', this.s.topology);
-  getSingleProperty(this, 'bufferMaxEntries', this.s.bufferMaxEntries);
-  getSingleProperty(this, 'databaseName', this.s.namespace.db);
-
-  if (this.s.noListener) return;
-}
+const Db = deprecateOptions(
```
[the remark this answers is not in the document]
reggi: CC: Tried to implement, tests failed, my understanding is this is possible for methods, but not constructors.

lib/db.js
emadum: That's likely true, sorry about the misleading advice - I didn't look at it super closely. It looks like there's [another helper](https://github.com/mongodb/node-mongodb-native/blob/665b3524080776d836c69351c3db1cb7f211c3dc/lib/utils.js#L576) we could use to do a manual check in the constructor.

```
if (options.promiseLibrary) {
  emitDeprecationWarning('MongoClient option [promiseLibrary] is deprecated and will be removed in a later version.`);
}
```

lib/db.js
```diff
@@ -132,19 +132,18 @@ const legalOptionNames = [
...
  */
+
 function Db(databaseName, topology, options) {
   options = options || {};
   if (!(this instanceof Db)) return new Db(databaseName, topology, options);
 
-  // Get the promiseLibrary
-  const promiseLibrary = options.promiseLibrary || Promise;
+  if (options.promiseLibrary) {
+    emitDeprecationWarning('MongoClient option [promiseLibrary] is deprecated');
```
mbroadst, line 141: You're emitting this deprecation warning in `Db` not `MongoClient`, also you are only emitting this in `Db`, but we used to accept it as an option in a few other types?

reggi: Added it to:
-  lib/collection.js 
-  lib/cursor/cursor.js 
-  lib/db.js 
-  lib/sdam/topology.js

lib/utils.js
```diff
@@ -928,6 +929,14 @@ function makeClientMetadata(options) {
 
 const noop = () => {};
 
+function emitDeprecatedOptionWarning(options, name, list) {
+  list.forEach(option => {
+    if (options[option]) {
```
mbroadst, line 934: this should be `if (typeof options[option] !== 'undefined')`, otherwise you'll get some false positives

lib/utils.js
```diff
@@ -928,6 +929,14 @@ function makeClientMetadata(options) {
 
 const noop = () => {};
 
+function emitDeprecatedOptionWarning(options, name, list) {
```
mbroadst, line 932: do we need `name` here? [process.emitWarning](https://nodejs.org/api/process.html#process_process_emitwarning_warning_options) will provide a `stack` property if a user listens to `process.on('warning', ...)`, so they should have the details they need to figure out where the option is passed in. 

Also, can you please document this method and its inputs?

reggi: removed name, updated emitDeprecatedOptionWarning

lib/utils.js
reggi: updated!

PR #2349, opened 2020-05-04
feat: directConnection adds unify behavior for replica set discovery

reggi:
Adds `directConnection` option to unify behavior around configuration for replica set discovery. Migrated mongodb/specifications tests from commit "e56f5eceed7729f8b9b43a4a1f76c7e5840db49f". Skips SDAM tests for legacy topology / behavior that we do not intend to introduce to the legacy topology types. Users should switch to the unified topology.

NODE-2452


lib/sdam/server_selection.js
```diff
@@ -33,17 +33,19 @@ function maxStalenessReducer(readPreference, topologyDescription, servers) {
     return servers;
   }
 
-  const maxStaleness = readPreference.maxStalenessSeconds;
-  const maxStalenessVariance =
-    (topologyDescription.heartbeatFrequencyMS + IDLE_WRITE_PERIOD) / 1000;
-  if (maxStaleness < maxStalenessVariance) {
-    throw new MongoError(`maxStalenessSeconds must be at least ${maxStalenessVariance} seconds`);
-  }
+  if (topologyDescription.type !== TopologyType.Unknown) {
```
reggi, line 36: had to do this because of: 
https://github.com/mongodb/specifications/blob/e56f5eceed7729f8b9b43a4a1f76c7e5840db49f/source/max-staleness/tests/Unknown/SmallMaxStaleness.yml#L1

lib/sdam/server_selection.js
```diff
@@ -5,7 +5,7 @@ const { MongoError } = require('../error');
 
 // max staleness constants
 const IDLE_WRITE_PERIOD = 10000;
-const SMALLEST_MAX_STALENESS_SECONDS = 20;
+const SMALLEST_MAX_STALENESS_SECONDS = 90;
```
reggi: had to do this because of:
https://github.com/mongodb/specifications/blob/e56f5eceed7729f8b9b43a4a1f76c7e5840db49f/source/max-staleness/tests/ReplicaSetWithPrimary/MaxStalenessTooSmall.yml#L25

lib/connection_string.js
```diff
@@ -706,6 +713,22 @@ function parseConnectionString(uri, options, callback) {
...
   }
 
+  const directConnection = !!parsedOptions.directConnection;
+  if (directConnection && hosts.length !== 1) {
+    // If the option is set to true, the driver MUST validate that there is exactly one host given
+    // in the host list in the URI, and fail client creation otherwise.
+    return callback(new MongoParseError('directConnection option requires exactly one host'));
+  }
+
+  // NOTE: this behavior will go away in v4.0, we will always auto discover there
```
mbroadst, line 723: this _is_ v4.0, so it seems like something should change here?

lib/sdam/server_selection.js
mbroadst: is this a change from another PR, or new code?

reggi: this is new code to make that specific spec work in master

emadum: looks like this requirement was added in [this spec commit](https://github.com/mongodb/specifications/commit/69e4225a20c3a53428b02afec8b9cbc9a5a3a2a0#diff-badcf416f8d59df7eeb0a22ba4ca4a9fR93) quite a while back

lib/sdam/topology_description.js
```diff
@@ -277,8 +277,22 @@ class TopologyDescription {
...
+  }
+
+  if (serverType === ServerType.RSPrimary) {
+    return TopologyType.ReplicaSetWithPrimary;
+  }
+
+  if (serverType === ServerType.RSGhost || serverType === ServerType.Unknown) {
+    return TopologyType.Unknown;
+  }
+
```
emadum, line 295: I personally think switches are cleaner for this sort of thing.

```
switch (serverType) {
  case ServerType.Standalone:
    return TopologyType.single;
  case ServerType.Mongos:
    return TopologyType.Sharded;
  case ServerType.RSPrimary:
    return TopologyType.ReplicaSetWithPrimary;
  default:
    return TopologyType.Unknown;
}
```

Thoughts?

lib/connection_string.js
reggi: [This block is in the 3.6 commit](https://github.com/mongodb/node-mongodb-native/commit/c5d60fc4619227697ef2102437fe5c8b111909d2#diff-8a39aaa3a2d1338e1a7455cd10b47ae2R683-R698). I think you might have added it. I've tried to remove it but spec fails. I've tried to learn more about [auto-discovery but all i've found in the spec is line](https://github.com/mongodb/specifications/search?q=auto-discovery&unscoped_q=auto-discovery):

> Why is auto-discovery the preferred default?
> Auto-discovery is most resilient and is therefore preferred.

I'm assuming that there's a breaking change we can make in 4.0 that's not in 3.6 but I can't find what that change is or how to execute it.

lib/sdam/server_selection.js
mbroadst: I think my question is: why didn't we need this in 3.6? And I think the answer is because [this commit](https://github.com/mongodb/node-mongodb-native/commit/af57b578dd603faa7b66983232de2bc7e417dae1) was not ported to `master`. Can you please cherry-pick that one in and confirm it fixes the issue you were trying to solve here?

lib/connection_string.js
mbroadst: answered on slack

lib/sdam/topology_description.js
```diff
@@ -284,11 +284,11 @@ function topologyTypeForServerType(serverType) {
       return TopologyType.Sharded;
     case ServerType.RSPrimary:
       return TopologyType.ReplicaSetWithPrimary;
-    case ServerType.RSGhost:
-    case ServerType.Unknown:
-      return TopologyType.Unknown;
-    default:
+    case ServerType.RSOther:
+    case ServerType.RSSecondary:
       return TopologyType.ReplicaSetNoPrimary;
```
reggi, line 289: It didn't make sense to me to have the "default" be `ReplicaSetNoPrimary` so I swapped this, initial spec tests showed this worked.

test/unit/sdam/spec.test.js
```diff
@@ -51,11 +51,21 @@ describe('Server Discovery and Monitoring (spec)', function() {
     serverConnect.restore();
   });
 
+  const shouldSkip = desc => {
+    const descriptions = [
+      'Monitoring a standalone connection',
+      'Monitoring a standalone connection - suppress update events for equal server descriptions'
+    ];
```
reggi, line 58: [skipped the tests b/c this convo here](https://mongodb.slack.com/archives/C72LB5RPV/p1588864275286000)

lib/sdam/topology_description.js
```diff
@@ -163,7 +163,7 @@ class TopologyDescription {
     }
 
     if (topologyType === TopologyType.Unknown) {
-      if (serverType === ServerType.Standalone) {
+      if (serverType === ServerType.Standalone && this.servers.size !== 1) {
```
emadum, line 166: Why was it necessary to add this check against `this.servers.size`?

reggi: My understanding of this is that we got back `Standalone` from the server, but the config had multiple servers, so it's not truly `Standalone` we don't process the `serverType` and get rid of this address. Although I'm not quite sure, this was ported over from 3.6 when Matt helped me with making the changes. I've tried testing combinations around this, it all seems to fail.

reggi: Here's a test that came up failing referring to what I'm trying to explain: https://github.com/mongodb/specifications/blob/0cf00e6ec2ebc1f8959d04c674623431b58198f1/source/server-discovery-and-monitoring/tests/single/standalone_removed.json

lib/connection_string.js
```diff
@@ -37,6 +37,10 @@ function matchesParentDomain(srvAddress, parentDomain) {
 function parseSrvConnectionString(uri, options, callback) {
   const result = URL.parse(uri, true);
 
+  if (options.directConnection || options.directconnection) {
```
mbroadst, line 40: is it possible to have both forms here? I was hoping we wouldn't be introducing more cases where we had to check the upper and lowercase version of URI options.

lib/mongo_client.js
```diff
@@ -143,6 +143,7 @@ const { connect, validOptions } = require('./operations/connect');
  * @param {boolean} [options.useNewUrlParser=true] Determines whether or not to use the new url parser. Enables the new, spec-compliant, url parser shipped in the core driver. This url parser fixes a number of problems with the original parser, and aims to outright replace that parser in the near future. Defaults to true, and must be explicitly set to false to use the legacy url parser.
  * @param {DriverInfoOptions} [options.driverInfo] Allows a wrapping driver to amend the client metadata generated by the driver to include information about the wrapping driver
  * @param {AutoEncrypter~AutoEncryptionOptions} [options.autoEncryption] Optionally enable client side auto encryption
+ * @param {boolean} [options.directConnection=false] Enable directConnection
```
mbroadst, line 146: can we improve this doc string? "Indicates that a client should directly connect to a node without attempting to discover its topology type" or something like that

lib/mongo_client.js
```diff
@@ -388,6 +389,7 @@ MongoClient.prototype.isConnected = function(options) {
  * @param {number} [options.numberOfRetries=5] The number of retries for a tailable cursor
  * @param {boolean} [options.auto_reconnect=true] Enable auto reconnecting for single server instances
  * @param {number} [options.minSize] If present, the connection pool will be initialized with minSize connections, and will never dip below minSize connections
+ * @param {boolean} [options.directConnection=false] Enable directConnection
```
mbroadst, line 392: same comment

lib/sdam/topology.js
```diff
@@ -101,6 +101,10 @@ class Topology extends EventEmitter {
    * @param {number} [options.localThresholdMS=15] The size of the latency window for selecting among multiple suitable servers
    * @param {number} [options.serverSelectionTimeoutMS=30000] How long to block for server selection before throwing an error
    * @param {number} [options.heartbeatFrequencyMS=10000] The frequency with which topology updates are scheduled
+   * @param {boolean} [options.directConnection] will set topology to Single
+   * @param {string} [options.replicaSet] will set topology to ReplicaSetNoPrimary
```
mbroadst, line 105: `replicaSet` should be the only supported option, `setName` and `rs_name` can and should be removed in this major revision. Also I think the docstrings for all of these are a bit wanting, `directConnection` means "connect directly to the node without discovering its type" and "replicaSet" is the name of the replica set to connect to (used for validation during initial handshake as well). You can find decent descriptions of these in the [URI Options spec](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst)

lib/sdam/topology_description.js
mbroadst: nit: I don't find that using a `switch` here improves readability at all :man_shrugging: Maybe adding some line breaks in there would be useful, I just find it hard to visually track

test/unit/sdam/spec.test.js
mbroadst: Is there a DRIVERS ticket you can link to which fixes the change? Please include it here

lib/sdam/topology_description.js
mbroadst: I think this is a translation of the [updateUnknownWithStandalone](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updateunknownwithstandalone) guidance in the SDAM spec.

lib/sdam/topology_description.js
reggi: I generally don't prefer switch syntax, but i think @emadum's recommendation makes sense here it does help restrict what this function can do. It makes it more obvious that this function takes a `ServerType` and returns a `TopologyType` and no other logic is really allowed.

test/unit/sdam/spec.test.js
reggi: adding comment to test file

// DRIVERS-1249 should add directConnection and then update spec, remove skip

https://jira.mongodb.org/browse/DRIVERS-1249

lib/connection_string.js
reggi: During testing it came in as both, I think it performs differently when in URL and passed in, this covers all our bases.

Screen Shot 2020-05-11 at 3 52 25 PM

Is there a helper I'm missing? It's also in `CASE_TRANSLATION`. 😫

mbroadst: so case translation comes into play in [applyConnectionStringOption](https://github.com/mongodb/node-mongodb-native/blob/master/lib/connection_string.js#L299) which is in turn called by [parseQueryString](https://github.com/mongodb/node-mongodb-native/blob/master/lib/connection_string.js#L425). Your change here moves the call to `parseSrvConnectionString` to _after_ `parseQueryString` is called, which means that we should be in a world where all known values are camelCased.

lib/sdam/topology_description.js
mbroadst: default `Unknown` works for me, I'm fine to resolve this comment

lib/sdam/topology_description.js
mbroadst: OK to resolve @emadum ?

emadum: Yep, thanks!

lib/connection_string.js
mbroadst: @reggi I think we're waiting on you here

PR #2350, opened 2020-05-04
feat: support hedged reads

reggi:
Adds driver support for server Hedged Reads. Adds `hedge` property to the `ReadPreference` class and exports it using the `.toJSON` method. Introduces `withMonitoredClient` a shared testing utility for testing command monitoring. Tests command sent to the server and verifies that `hedge` property was propagated. Tests are only run on `>=3.6.0` due to how prior versions send operations to the server.

NODE-2510


test/functional/shared.js
```diff
@@ -192,12 +192,33 @@ class EventCollector {
...
+  }
+  return function(done) {
+    const configuration = this.configuration;
+    const client = configuration.newClient({ monitorCommands: true });
+    const events = [];
+    client.on('commandStarted', filterForCommands(commands, events));
+    client.connect((err, client) => {
+      expect(err).to.not.exist;
+      const d = function() {
+        const args = Object.values(arguments);
```
mbroadst, line 207: I think this is different than what made it into [3.6](https://github.com/mongodb/node-mongodb-native/blob/3.6/test/functional/shared.js#L206-L208)

reggi: Good catch

PR #2351, opened 2020-05-05
NODE-2429/master/grid-fs-index-check

reggi:
Tests if createIndex issued in shell shell on grid store shows in driver.


test/functional/gridfs_stream.test.js
```diff
@@ -1380,4 +1382,43 @@ describe('GridFS Stream', function() {
...
+
+  it('should correctly handle indexes created in shell', function(done) {
+    const configuration = this.configuration;
+    const url = format(
+      'mongodb://%s/%s',
+      format('%s:%s', configuration.host, configuration.port),
+      configuration.db
+    );
+    const cmd = `mongo ${url} --eval 'db.fs.files.createIndex({"filename" : 1,"uploadDate" : 1})'`;
+    child_process.exec(cmd, err => {
```
reggi, line 1394: How do we feel about this? Any alternatives? How can a test run command in the shell?

mbroadst: I think you might be reading the ticket a little literally. If you look at the fix for [another driver](https://github.com/rozza/mongo-java-driver/pull/362/files) you'll see that this issue is about type mismatches when creating an index. We are likely not affected by this because node is loosely typed, but the root of the ticket is about the type mismatch and so coverage should be added around that.

IMO we should _never_ spawn child processes from our tests.

test/functional/gridfs_stream.test.js
```diff
@@ -1380,4 +1382,44 @@ describe('GridFS Stream', function() {
...
+  });
+
+  it('should correctly handle indexes create with decimal', function(done) {
+    const configuration = this.configuration;
+    const client = configuration.newClient();
+    client.connect((err, client) => {
+      expect(err).to.not.exist;
+      const db = client.db(configuration.db);
+      const col = db.collection('fs.files');
+      col.createIndex({ filename: 1.0, uploadDate: 1.0 }, err => {
```
reggi, line 1413: I'm not sure of two things:

1) Are we simply trying to create indexes with the driver using floats? If so I'm a little confused what this [has to do with the shell.](https://jira.mongodb.org/browse/DRIVERS-789) It was my understanding the type mis-match would occur if the shell creates the index, and the driver loads the collection with that created index. Just need clarification on what we're looking for here.

2) What makes a collection in the Node Driver a GridFs collection? There's nothing specific to these tests that differentiates them from creating an index on a regular collection.

emadum: Not sure, but I think:
1. We don't need a test at all, because both the Mongo shell and the Node driver are using JS and these sorts of differences aren't an issue, because `1 === 1.0` in JS.
2. GridFS manages two collections, `<bucket>.files` and `<bucket>.chunks` for each bucket (in this case `fs`). You can access these just like any other collection, which is why you can create an index on them using `Collection.createIndex`, but they are also special because you can access them through GridFS.

test/functional/gridfs_stream.test.js
```diff
@@ -1422,4 +1422,25 @@ describe('GridFS Stream', function() {
...
+
+  it('should correctly handle indexes create with exponent', function(done) {
+    const configuration = this.configuration;
+    const client = configuration.newClient();
+    client.connect((err, client) => {
+      expect(err).to.not.exist;
+      const db = client.db(configuration.db);
+      const col = db.collection('fs.files');
+      // prettier-ignore
+      col.createIndex({ filename: 1e0, uploadDate: 1e0 }, err => {
```
emadum, line 1434: I don't think there's a purpose to this in Javascript. Putting `1`, `1.0`, or even `1.0000000000000000001` here will all get converted to the same internal representation, if my understanding is correct.

test/functional/gridfs_stream.test.js
reggi: @emadum thanks for clarifing

test/functional/gridfs_stream.test.js
reggi: @emadum Spoke with @mbroadst yesterday and this was the direction we were going to go in.

emadum: K, I still don't see the purpose but I'll go ahead and approve anyway.

reggi:
Currently this PR has number, decimal, exponent, and BSON Double.

Is this OK? Or do we only want Double?

PR #2353, opened 2020-05-05
feat: support speculative authentication in scram-sha and x509

mbroadst:
**this is the port to 4.0**

This introduces a new auth provider stage `prepare` which allows a provider to alter the handshake document before its used for the initial handshake. The SCRAM-SHA and X509 providers have been improved to use the prepare stage in order to speculatively authenticate, potentially reducing the roundtrips of connection authentication.

NODE-2487


lib/cmap/auth/gssapi.js
```diff
@@ -1,22 +1,15 @@
...
-   */
-  _authenticateSingleConnection(sendAuthCommand, connection, credentials, callback) {
-    const source = credentials.source;
+  auth(authContext, callback) {
+    if (Kerberos[kModuleError]) {
+      callback(Kerberos[kModuleError]);
+      return;
+    }
+
+    const { connection, credentials } = authContext;
```
emadum, line 12: Could go a bit further with destructuring here and get rid of `credentials`, which isn't used again past here.
`const { connection, credentials: { username, password, mechanismProperties } } = authContext;`

mbroadst: I thought we decided against the deep destructuring in one of Neal's PRs, because it would hose TypeScript?

emadum: Possibly, but I don't recall that discussion. Do you remember the details of how it would hose Typescript (or perhaps you do @nbbeeken)?

nbbeeken: There's no issue with doing this with TS. The discussion we had is that you can't use this nested destructuring with ES6 imports, which is potentially something we'd have to use in some far off future. So I had reverted some of the requires to just use one layer of destructuring, but that didn't make it in.

emadum: I do think it would be a bit of a downer to finally be able to use destructuring in v4, but only at the root level.

PR #2354, opened 2020-05-06
fix: honor journal=true in connection string

emadum:
## Description
Ensure the `journal` option returned by`parseQueryString` is converted
into `j` during the `connect` operation. This fixes the issue of the
write concern not being set on commands where `journal` is only specified
in the connection string.

[NODE-2422](https://jira.mongodb.org/browse/NODE-2442)

**What changed?**

- replace `journal` with `j` in options during `connect` operation
- updated test helpers
  - refactored `withMonitoredClient` so it can pass options on to `newClient`
  - added some jsdoc to helpers for better code completion when writing tests


lib/connection_string.js
```diff
@@ -427,7 +427,7 @@ function parseQueryString(query, options) {
 
   // special cases for known deprecated options
   if (result.wtimeout && result.wtimeoutms) {
-    delete result.wtimeout;
+    result.wtimeout = undefined;
```
mbroadst, line 430: I know this is counter to what I've said previously, but in these non-critical paths I think there's actual value to outright removing a known bad value. In the future we might solve this by "plucking" all the values we care about into a new object.

lib/operations/connect.js
```diff
@@ -212,6 +212,12 @@ function connect(mongoClient, url, options, callback) {
       delete _finalOptions.db_options.auth;
     }
 
+    // `journal` should be translated to `j` for the driver
+    if (_finalOptions.journal != null) {
+      _finalOptions.j = _finalOptions.journal;
```
mbroadst, line 217: to me it seems like `j` is the odd man out here, shouldn't we filter that out and depend on `journal` throughout the codebase? `j` is just the field on the command send to the server. 

**UPDATE:** I now see that this is a pretty small change, while what I'm proposing would potentially touch much more of the codebase. Maybe we can defer my suggestion here to @reggi s work on `MongoClientOptions`

emadum: Yeah I think this is the minimal change necessary to fix this issue; it's probably not worth diving deep on this now when we're already planning on overhauling this entire system; definitely worth addressing when we do the overhaul.

mbroadst: btw I think the original complaint is for 3.5, so we'll need a port there too

emadum: #2359

PR #2355, opened 2020-05-06
fix: allow connect again after close

emadum:
## Description

`connect` doesn't work again after `close` has been called once on a `MongoClient`.

[NODE-2544](https://jira.mongodb.org/browse/NODE-2544)

**What changed?**

Also introduced some CI-related cleanup:
- output `eslint` version before linting
- upgrade `nvm` and use `--no-progress` option to clean up log output

I can split this out into a separate PR if desired.

### DIFF
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -9,6 +9,7 @@ NPM_TMP_DIR="${NODE_ARTIFACTS_PATH}/tmp"
 
 # this needs to be explicitly exported for the nvm install below
 export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
+export XDG_CONFIG_HOME=${NODE_ARTIFACTS_PATH}
 
 # create node artifacts path if needed
 mkdir -p ${NODE_ARTIFACTS_PATH}
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -18,7 +18,7 @@ mkdir -p "${NPM_TMP_DIR}"
 # install Node.js
 curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | bash
 [ -s "${NVM_DIR}/nvm.sh" ] && \. "${NVM_DIR}/nvm.sh"
-nvm install --lts=${NODE_LTS_NAME}
+nvm install --no-progress --lts=${NODE_LTS_NAME}
 
 # setup npm cache in a local directory
 cat <<EOT > .npmrc
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -16,7 +16,7 @@ mkdir -p ${NPM_CACHE_DIR}
 mkdir -p "${NPM_TMP_DIR}"
 
 # install Node.js
-curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | bash
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.35.3/install.sh | bash
 [ -s "${NVM_DIR}/nvm.sh" ] && \. "${NVM_DIR}/nvm.sh"
 nvm install --no-progress --lts=${NODE_LTS_NAME}
 
--- a/lib/mongo_client.js
+++ b/lib/mongo_client.js
@@ -241,11 +241,13 @@ MongoClient.prototype.close = function(force, callback) {
     force = false;
   }
 
+  const topology = this.topology;
+
   return maybePromise(this, callback, cb => {
     const completeClose = err => {
       this.emit('close', this);
 
-      if (!(this.topology instanceof NativeTopology)) {
+      if (!(topology instanceof NativeTopology)) {
         for (const item of this.s.dbCache) {
           item[1].emit('close', this);
         }
@@ -261,13 +263,13 @@ MongoClient.prototype.close = function(force, callback) {
       cb(err);
     };
 
-    if (this.topology == null) {
+    if (topology == null) {
       completeClose();
       return;
     }
 
-    this.topology.close(force, err => {
-      const autoEncrypter = this.topology.s.options.autoEncrypter;
+    topology.close(force, err => {
+      const autoEncrypter = topology.s.options.autoEncrypter;
       if (!autoEncrypter) {
         completeClose(err);
         return;
--- a/lib/mongo_client.js
+++ b/lib/mongo_client.js
@@ -167,6 +167,8 @@ function MongoClient(url, options) {
     writeConcern: WriteConcern.fromOptions(options),
     namespace: new MongoDBNamespace('admin')
   };
+
+  this.originalOptions = Object.assign({}, this.s.options);
 }
 
 /**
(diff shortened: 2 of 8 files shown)


lib/operations/connect.js
```diff
@@ -260,6 +260,18 @@ function connect(mongoClient, url, options, callback) {
     throw new Error('no callback function provided');
   }
 
+  // Has a connection already been established?
+  if (mongoClient.topology) {
+    if (mongoClient.topology.isConnected()) {
+      throw new Error('already connected');
+    }
+    // this is a reconnect, clear out old state
```
emadum, line 268: This might be better to do somewhere inside `close`, just wanted to get some feedback.

package.json
```diff
@@ -68,7 +68,7 @@
     "test": "npm run lint && mocha --recursive test/functional test/unit test/core",
     "test-nolint": "mocha --recursive test/functional test/unit test/core",
     "coverage": "istanbul cover mongodb-test-runner -- -t 60000 test/core test/unit test/functional",
-    "lint": "eslint lib test",
+    "lint": "eslint -v && eslint lib test",
```
emadum, line 71: Unrelated to this PR. I just sometimes get weird discrepancies between running `eslint` locally and what runs on evergreen. Was thinking adding this could help make things a bit more explicit around the version of `eslint` being used, when it is being used.

test/functional/shared.js
```diff
@@ -78,9 +78,20 @@ function makeCleanupFn(client) {
 function withClient(client, operation, errorHandler) {
   const cleanup = makeCleanupFn(client);
 
+  function operationHandler(client) {
```
emadum, line 81: The alternative to this change would be wrapping those tests in promises; if that's preferred lmk.

.evergreen/install-dependencies.sh
```diff
@@ -9,16 +9,17 @@ NPM_TMP_DIR="${NODE_ARTIFACTS_PATH}/tmp"
...
 mkdir -p ${NODE_ARTIFACTS_PATH}
 mkdir -p ${NPM_CACHE_DIR}
 mkdir -p "${NPM_TMP_DIR}"
 
 # install Node.js
-curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | bash
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.35.3/install.sh | bash
 [ -s "${NVM_DIR}/nvm.sh" ] && \. "${NVM_DIR}/nvm.sh"
-nvm install --lts=${NODE_LTS_NAME}
+nvm install --no-progress --lts=${NODE_LTS_NAME}
```
emadum, line 22: Since I was already cleaning up the CI, I upgraded `nvm` so we can use the newish [`--no-progress` option.](https://github.com/nvm-sh/nvm/issues/1079)

mbroadst: oh very nice :clap:

test/functional/shared.js
```diff
@@ -89,9 +89,20 @@ function withTempDb(name, options, client, operation, errorHandler) {
...
+    // run the operation
+    const result = operation(client);
+    // if it returns a callback, wrap it in a Promise
+    if (typeof result === 'function') {
```
mbroadst, line 96: I think we have a [helper](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/utils.js#L525-L533) for this.

test/functional/connection.test.js
```diff
@@ -514,4 +515,27 @@ describe('Connection', function() {
       });
     }
   });
+
+  it('should be able to connect again after close', function() {
+    return withClient(this.configuration.newClient(), client => done => {
```
mbroadst, line 520: I don't think we want a version of `withClient` that does it this way, returning functions which return functions can be very confusing for readability. I think it would be more useful for developers if we had something like:

```
withClient(this.configuration.newClient(), (client, done) => {

});
```

`withClient` itself could still return a `Promise`, you'd just have to determine the arity of the lambda passed in to see if a `done` callback was provided.

lib/operations/connect.js
```diff
@@ -260,6 +260,18 @@ function connect(mongoClient, url, options, callback) {
     throw new Error('no callback function provided');
   }
 
+  // Has a connection already been established?
+  if (mongoClient.topology) {
+    if (mongoClient.topology.isConnected()) {
+      throw new Error('already connected');
```
mbroadst, line 266: I think this error message can be improved

lib/operations/connect.js
mbroadst: I think the eventual goal would be that the `topology` contains all of the state, and is removed on `close`. I think it's fine to do additional cleanups of these other types, but agree with you that this should most likely move to `close` or the `close` operation.

lib/mongo_client.js
```diff
@@ -241,28 +243,35 @@ MongoClient.prototype.close = function(force, callback) {
     force = false;
   }
 
-  const client = this;
```
emadum, line 244: We were using arrow functions everywhere here, so I just updated everything to use `this`

lib/mongo_client.js
```diff
@@ -241,28 +243,35 @@ MongoClient.prototype.close = function(force, callback) {
...
+          item[1].emit('close', this);
         }
       }
 
-      client.removeAllListeners('close');
+      this.removeAllListeners('close');
+
+      // clear out references to old topology
+      this.topology = undefined;
+      this.s.dbCache = new Map();
```
emadum, line 260: the old `dbCache` still references the old topology, so it is manually cleared out here in addition to `this.topology`

lib/mongo_client.js
```diff
@@ -241,28 +243,35 @@ MongoClient.prototype.close = function(force, callback) {
...
+
+      // restore options
+      this.s.options = Object.assign({}, this.originalOptions);
```
emadum, line 263: `this.s.options` can get mutated by `connect`, so we need to restore it to what it was originally constructed with. Otherwise we get the following warnings, as reported in one of the issues:

```
the options [servers] is not supported
the options [caseTranslate] is not supported
the options [dbName] is not supported
```

lib/mongo_client.js
mbroadst: iirc the reason `client` was used is because this method has significant enough depth that `this` can't be bound to the deepest nested functions. Can you confirm, for instance, that assignment to `autoEncrypter` below is actually referencing the correct `this`?

test/functional/connection.test.js
```diff
@@ -517,7 +517,7 @@ describe('Connection', function() {
   });
 
   it('should be able to connect again after close', function() {
-    return withClient(this.configuration.newClient(), client => done => {
+    return withClient.call(this, (client, done) => {
```
mbroadst, line 520: this should be equivalent to just calling `withClient` within the scope, why was this change needed?

test/functional/shared.js
```diff
@@ -57,8 +57,37 @@ function setupDatabase(configuration, dbsToClean) {
     );
 }
 
-function makeCleanupFn(client) {
-  return function(err) {
+function withTempDb(name, options, client, operation, errorHandler) {
```
mbroadst: Similar to the change streams PR, I'd like to find a way not to wrap up `withClient` calls into something that doesn't seem to create a client. At very least I think we should have composed naming for these methods, like `withClientAndTempDb`, but maybe there's an even better way. Maybe we could end up with something like:
```
withClient(withTempDb((client, db, done) => {

}));
```

test/functional/connection.test.js
emadum: Without it I get `TypeError: Cannot read property 'configuration' of undefined` referencing [this line](https://github.com/mongodb/node-mongodb-native/pull/2355/files#diff-fd8c048cd2959e76db634d7719c6c1e3R92).

lib/mongo_client.js
emadum: Confirmed that if I add `client` back in, `client === this` is true on the line before the assignment to `autoEncrypter`.

test/functional/shared.js
emadum: I was actually working on this in a separate draft PR (#2362) so I pulled in my new `withDb` helper.

emadum:
Will open a new PR for this after #2362 is merged

anzairyo0127:
#2362 seems to have been merged. 
Has a new Pull Request been issued for this change?

PR #2360, opened 2020-05-08
fix(ChangeStream): should resume from errors when iterating

emadum:
## Description
Introduced `getCursor` method to safely provide a change stream cursor
for `next`/`hasNext` across recoveries from resumable errors.

[NODE-2548](https://jira.mongodb.org/browse/NODE-2548)

**What changed?**
- `next` and `hasNext` now use `getCursor`
- `processError` was extracted from `processNewChange`
- added `withCursor` shared test helper

### DIFF
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -170,8 +170,8 @@ class ChangeStream extends EventEmitter {
    * @param  {boolean} [checkCursor=true] also check if the underlying cursor is closed
    * @return {boolean}
    */
-  isClosed(checkCursor) {
-    return !!(this.closed || (checkCursor && this.cursor && this.cursor.isClosed()));
+  isClosed() {
+    return this.closed || (this.cursor && this.cursor.isClosed());
   }
 
   /**
@@ -475,7 +475,7 @@ function waitForTopologyConnected(topology, options, callback) {
 function processNewChange(changeStream, change, callback) {
   const cursor = changeStream.cursor;
 
-  if (changeStream.isClosed()) {
+  if (changeStream.closed) {
     if (callback) callback(new MongoError('ChangeStream is closed'));
     return;
   }
@@ -506,7 +506,7 @@ function processError(changeStream, error, callback) {
   const cursor = changeStream.cursor;
 
   // If the change stream has been closed explictly, do not process error.
-  if (changeStream.isClosed()) {
+  if (changeStream.closed) {
     if (callback) callback(new MongoError('ChangeStream is closed'));
     return;
   }
@@ -568,7 +568,7 @@ function processError(changeStream, error, callback) {
  * @param {ChangeStreamCursor} [oldCursor] when resuming from an error, carry over options from previous cursor
  */
 function getCursor(changeStream, callback) {
-  if (changeStream.isClosed(true)) {
+  if (changeStream.isClosed()) {
     callback(new MongoError('ChangeStream is closed.'));
     return;
   }
@@ -593,7 +593,7 @@ function getCursor(changeStream, callback) {
 function processResumeQueue(changeStream, err) {
   while (changeStream[kResumeQueue].length) {
     const request = changeStream[kResumeQueue].pop();
-    if (changeStream.isClosed(true) && !err) {
+    if (changeStream.isClosed() && !err) {
       request(new MongoError('Change Stream is not open.'));
       return;
     }
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -469,17 +469,12 @@ function waitForTopologyConnected(topology, options, callback) {
   }, 500); // this is an arbitrary wait time to allow SDAM to transition
 }
 
-// Handle new change events. This method brings together the routes from the callback, event emitter, and promise ways of using ChangeStream.
 function processNewChange(changeStream, change, callback) {
   const cursor = changeStream.cursor;
 
   if (changeStream.closed) {
     if (callback) callback(new MongoError('ChangeStream is closed'));
     return;
-  }
-
-  if (cursor == null) {
-    throw new Error('cursor should never be null here');
   }
 
   if (change && !change._id) {
@@ -513,6 +508,13 @@ function processError(changeStream, error, callback) {
     return;
   }
 
+  // if the resume succeeds, continue with the new cursor
+  function resumeWithCursor(newCursor) {
+    changeStream.cursor = newCursor;
+    processResumeQueue(changeStream);
+  }
+
+  // otherwise, raise an error and close the change stream
   function unresumableError(err) {
     if (!callback) {
       changeStream.emit('error', err);
@@ -520,11 +522,6 @@ function processError(changeStream, error, callback) {
     }
     processResumeQueue(changeStream, err);
     changeStream.closed = true;
-  }
-
-  function resumeWithCursor(newCursor) {
-    changeStream.cursor = newCursor;
-    processResumeQueue(changeStream);
   }
 
   if (cursor && isResumableError(error, maxWireVersion(cursor.server))) {
@@ -536,13 +533,23 @@ function processError(changeStream, error, callback) {
     // close internal cursor, ignore errors
     cursor.close();
 
+    // the presence of a callback here indicates the error happened during a .next
+    // so it should be retried on resume
     if (callback) changeStream[kResumeQueue].push(() => changeStream.next(callback));
 
     waitForTopologyConnected(topology, { readPreference: cursor.options.readPreference }, err => {
+      // if the topology can't reconnect, close the stream
       if (err) return unresumableError(err);
+
+      // create a new cursor, preserving the old cursor's options
       const newCursor = createChangeStreamCursor(changeStream, cursor.resumeOptions);
+
+      // attempt to continue in emitter mode
       if (!callback) return resumeWithCursor(newCursor);
+
+      // attempt to continue in iterator mode
       newCursor.hasNext(err => {
+        // if there's an error immediately after resuming, close the stream
         if (err) return unresumableError(err);
         resumeWithCursor(newCursor);
       });
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -135,7 +135,7 @@ class ChangeStream extends EventEmitter {
   hasNext(callback) {
     return maybePromise(this.parent, callback, cb => {
       getCursor(this, (err, cursor) => {
-        if (err) return cb(err);
+        if (err) return cb(err); // failed to resume, raise an error
         cursor.hasNext(cb);
       });
     });
@@ -151,10 +151,12 @@ class ChangeStream extends EventEmitter {
   next(callback) {
     return maybePromise(this.parent, callback, cb => {
       getCursor(this, (err, cursor) => {
-        if (err) return cb(err);
+        if (err) return cb(err); // failed to resume, raise an error
         cursor.next((error, change) => {
           if (error) {
-            return processError(this, error, cb);
+            this[kResumeQueue].push(() => this.next(cb));
+            processError(this, error, cb);
+            return;
           }
           processNewChange(this, change, cb);
         });
@@ -163,12 +165,13 @@ class ChangeStream extends EventEmitter {
   }
 
   /**
-   * Is the cursor closed
+   * Is the change stream closed
    * @method ChangeStream.prototype.isClosed
+   * @param  {boolean} [checkCursor=true] also check if the underlying cursor is closed
    * @return {boolean}
    */
-  isClosed() {
-    return this.closed || (this.cursor && this.cursor.isClosed());
+  isClosed(checkCursor) {
+    return !!(this.closed || (checkCursor && this.cursor && this.cursor.isClosed()));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


test/functional/change_stream.test.js
```diff
@@ -2776,7 +2786,7 @@ describe('Change Streams', function() {
     // - MUST include a startAfter option
     // - MUST NOT include a resumeAfter option
     // when resuming a change stream.
-    it('$changeStream that has not received results must include startAfter and not resumeAfter', {
+    it('$changeStream w/o results must include startAfter and not resumeAfter', {
```
mbroadst, line 2789: nit: prefer `without`, slashes make it harder to grep for

test/functional/change_stream.test.js
```diff
@@ -2812,7 +2820,7 @@ describe('Change Streams', function() {
     // - MUST include a resumeAfter option
     // - MUST NOT include a startAfter option
     // when resuming a change stream.
-    it('$changeStream that has received results must include resumeAfter and not startAfter', {
+    it('$changeStream w/ results must include resumeAfter and not startAfter', {
```
mbroadst, line 2823: nit: same comment about greppability

test/functional/change_stream.test.js
```diff
@@ -2850,3 +2858,99 @@ describe('Change Streams', function() {
...
+      const configuration = this.configuration;
+      const client = configuration.newClient();
+      client.connect(err => {
+        expect(err).to.not.exist;
+        const db = client.db('changeStreamResumErrorTest');
+        db.createCollection(testName, (err, collection) => {
+          expect(err).to.not.exist;
+          const changeStream = collection.watch();
+          callback(collection, changeStream, () =>
+            changeStream.close(() => collection.drop(() => client.close(done)))
```
mbroadst, line 2874: this chain will ignore any errors on `close` and `drop`, as well as any error passed in from the test case itself (I think the signature should be `err => {}`).

test/functional/change_stream.test.js
```diff
@@ -2850,3 +2858,99 @@ describe('Change Streams', function() {
...
+          );
+        });
+      });
+    };
+  }
+  it('(events) should continue iterating after a resumable error', {
```
mbroadst, line 2880: `(events)` implies a `context` to me

test/functional/change_stream.test.js
```diff
@@ -2850,3 +2858,99 @@ describe('Change Streams', function() {
     });
   });
 });
+
+describe('Change Stream Resume Error Tests', function() {
+  function withChangeStream(testName, callback) {
```
mbroadst, line 2863: I think this might be useful as a more generalized helper `withCursor`, which we could alias to `withChangeStream` if that helps readability at all. Also, I think hiding a call to `withClient` in here might not be ideal since its sort of a side effect.

Finally, nitpick but `testName` seems to really be like `collectionName`?

test/functional/change_stream.test.js
```diff
@@ -18,17 +18,27 @@ chai.use(require('chai-subset'));
  * Triggers a fake resumable error on a change stream
  *
  * @param {ChangeStream} changeStream
- * @param {function} onCursorClosed callback when cursor closed due this error
+ * @param {Function} onCursorClosed callback when cursor closed due this error
+ * @param {number} [delay] optional delay before triggering error
  */
-function triggerResumableError(changeStream, onCursorClosed) {
-  const closeCursor = changeStream.cursor.close;
+function triggerResumableError(changeStream, onCursorClosed, delay) {
```
mbroadst, line 24: nit: `callback` should always be the last argument by convention

test/functional/change_stream.test.js
```diff
@@ -18,17 +18,27 @@ chai.use(require('chai-subset'));
...
+  let cbCalled = false;
+  const closeCursor = changeStream.cursor.close.bind(changeStream.cursor);
```
mbroadst, line 26: I don't know if this is any cleaner, but we have [patterns](https://github.com/mongodb/node-mongodb-native/blob/master/test/functional/scram_sha_256.test.js#L182-L193) for using `sinon` to do this type of patching

lib/change_stream.js
```diff
@@ -534,8 +494,112 @@ function processNewChange(args) {
...
+
+/**
+ * Safely provides a cursor across resume attempts
+ *
+ * @param {ChangeStream} changeStream the parent ChangeStream
+ * @param {function} callback gets the cursor or error
+ * @param {ChangeStreamCursor} [oldCursor] when resuming from an error, carry over options from previous cursor
+ */
+function getCursor(changeStream, callback) {
+  if (changeStream.isClosed()) {
```
mbroadst, line 572: just want to call this out here, but you are mixing checks for being closed between using the `isClosed` method and directly checking the `closed` boolean (in `processError`, `processNewChange`). Just want to make sure this intentional, and see if we can maybe consolidate to one approach. It occurs to me it might be because `isClosed` considers the existence of the `cursor`, so there might be some very subtle reasons to use one over the other

lib/change_stream.js
```diff
@@ -456,73 +469,20 @@ function waitForTopologyConnected(topology, options, callback) {
...
-
-  changeStream.attemptingResume = false;
-
   if (change && !change._id) {
     const noResumeTokenError = new Error(
       'A change stream document has been received that lacks a resume token (_id).'
     );
 
-    if (eventEmitter) return changeStream.emit('error', noResumeTokenError);
+    if (!callback) return changeStream.emit('error', noResumeTokenError);
```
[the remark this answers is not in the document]
emadum: I actually had that before and decided it was superfluous. We're either emitting if there's not a callback, or callback-ing if it's not an emitter; I don't think one condition is more clear than the other in the context of this behavior. This way there's no extra variable to reason about, just the function's arguments.

lib/change_stream.js
```diff
@@ -534,8 +494,112 @@ function processNewChange(args) {
...
+      if (err) return unresumableError(err);
+
+      // create a new cursor, preserving the old cursor's options
+      const newCursor = createChangeStreamCursor(changeStream, cursor.resumeOptions);
+
+      // attempt to continue in emitter mode
+      if (!callback) return resumeWithCursor(newCursor);
+
+      // attempt to continue in iterator mode
+      newCursor.hasNext(err => {
```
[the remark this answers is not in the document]
emadum: I don't think so, this is somewhat subtle and I should probably add a comment about it. 

The only place `processError` is called with a callback is in `changeStream.next`. This means that if it _can_ recover from the error, it should call that `changeStream.next` for the user.

However, this `hasNext` is just checking to see if the change stream returns _any_ error. If it does, the resume attempts fails. This is essentially the circuit-breaker that prevents an infinite loop of resumable errors from occurring.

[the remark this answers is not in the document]
mbroadst: should this line be moved to `ChangeStream#next` then:
```
cursor.next((err, doc) => {
  if (err) {
    // queue the call
    changeStream[kResumeQueue].push(() => changeStream.next(callback));
    processError(err);
    return;
  }
  ...
});
```

lib/change_stream.js
emadum: Yeah that's precisely the reason. I could consolidate them by doing something like this:
```
  /**
   * Is the change stream closed
   * @method ChangeStream.prototype.isClosed
   * @param  {boolean} checkCursor also check if the underlying cursor is closed
   * @return {boolean}
   */
  isClosed(checkCursor) {
    return this.closed || (checkCursor && this.cursor && this.cursor.isClosed());
  }
```
then using `isClosed(true)` for the current `isClosed()` checks, and `isClosed()` for the current checks against the `.closed` property.

emadum: The other direction to go would be to just inline the checks for the closed cursor in the two places they are needed, and only check the `closed` property in `isClosed`.

test/functional/change_stream.test.js
emadum: Waiting on #2362 to see how to best address this.

lib/change_stream.js
```diff
@@ -142,22 +150,28 @@ class ChangeStream extends EventEmitter {
...
   /**
-   * Is the cursor closed
+   * Is the change stream closed
    * @method ChangeStream.prototype.isClosed
+   * @param  {boolean} [checkCursor=true] also check if the underlying cursor is closed
    * @return {boolean}
    */
-  isClosed() {
-    return this.closed || (this.cursor && this.cursor.isClosed());
+  isClosed(checkCursor) {
```
mbroadst, line 173: I'm not sure we should make this change, `isClosed` is here for duck-typed parity with [Cursor#isClosed](https://github.com/mongodb/node-mongodb-native/blob/master/lib/cursor/cursor.js#L1002). Is there a way to implement the behavior that requires this boolean internally?

lib/change_stream.js
```diff
@@ -534,8 +497,108 @@ function processNewChange(args) {
...
+
+  // no cursor, queue callback until topology reconnects
+  changeStream[kResumeQueue].push(callback);
+}
+
+/**
+ * Drain the resume queue when a new has become available
+ *
+ * @param {ChangeStream} changeStream the parent ChangeStream
+ * @param {?ChangeStreamCursor} changeStream.cursor the new cursor
```
mbroadst, line 590: nit: for plain jsdoc I think the preference would be to write this as:
```
@param {ChangeStreamCursor} [changeStream.cursor]
````
but also I think both TypeScript and JavaScript prefer the question mark at the end of the symbol, rather than before.

File: test/functional/change_stream.test.js (hunk 5 above)
Comment (mbroadst) [reply test/functional/change_stream.test.js:2868]: I didn't ultimately find the `withDb` or `withCollection` helpers helped very much over there, but cursors are much more like clients, in that they have explicit `close` methods that sometimes need to be called.

File: lib/change_stream.js (hunk 11 above)
Comment (emadum) [reply lib/change_stream.js:173]: I made this change in response to [this discussion](https://github.com/mongodb/node-mongodb-native/pull/2360#discussion_r423053649). Should I revert it to the original implementation?

File: lib/change_stream.js (hunk 12 above)
Comment (emadum) [reply lib/change_stream.js:590]: This is something that's not entirely clear to me; is there a difference between an optional and a ["nullable type"](https://jsdoc.app/tags-type.html)?

I.e. I can write a function two ways. 
```
function optionalMiddleParam(a, b, c) {
  // handle options
  console.log({ a, b, c });
}
function nullableMiddleParam(a, b, c) {
  // handle options
  console.log({ a, b, c });
}
```

`optionalMiddleParam(1, 2)`  => outputs `{ a: 1, b: undefined, c: 2}`
`nullableMiddleParam(1, 2)` => outputs `{ a: 1, b: 2, c: undefined }`
and instead needs to be called as
`nullableMiddleParam(1, undefined, 2)` or `nullableMiddleParam(1, null, 2)`

My thinking was to use `[param]` vs `{?type}` to distinguish the two, but I wasn't really able to get a definitive answer from reading the jsdoc and typescript docs about nullable types vs optional params.

File: lib/change_stream.js (hunk 11 above)
Comment (mbroadst) [reply lib/change_stream.js:173]: Yeah I think I was mostly asking for rationale in that comment, and check that you knew you were doing that or if there was a bug. I think not exposing this at the api level is the right thing to do (users don't want to check isClosed conditionally, they just want to know if its closed)

File: test/functional/change_stream.test.js (hunk 5 above)
Comment (emadum) [reply test/functional/change_stream.test.js]: I added a `withCursor` helper; it's a general-purpose wrapper around `withClient`. Is this roughly what you had in mind?

```
function withCursor(getCursor, callback) {
  return withClient((client, done) => {
    getCursor(client, (err, cursor) => {
      if (err) return done(err);
      callback(cursor, () => cursor.close(done));
    });
  });
}
```

File: test/functional/change_stream.test.js (hunk 4 above)
Comment (emadum) [reply test/functional/change_stream.test.js]: I cleaned up the descriptions on those tests.

File: test/functional/change_stream.test.js (hunk 5 above)
Comment (mbroadst) [reply test/functional/change_stream.test.js]: I think of `withX` as a means of conveying the lifetime of a resource. So my thought with `withCursor` is that it would be _like_ `withClient`, but not _use_ `withClient`:
```
function withCursor(cursor, body) {
   body(cursor, (err, result) => {
     cursor.close(); // always close the cursor, this obviously would need to be async and handle errors etc
     
      callback(...);
   });
}
```

Comment (mbroadst) [reply test/functional/change_stream.test.js]: it may not be worth the effort considering we have `this.defer`? 
```
const cursor = coll.find();
this.defer(() => cursor.close());
cursor.toArray(...);
```
could work fine, but you might forget to put the defer in between creating the cursor and iterating it

File: test/functional/shared.js [hunk 13]
```diff
@@ -189,6 +189,27 @@ function withMonitoredClient(commands, options, callback) {
...
 }
 
+/**
+ * Safely perform a test with an arbitrary cursor.
+ *
+ * @param {Function} cursor any cursor that needs to be closed
+ * @param {(cursor: Object, done: Function) => void} body test body
+ * @param {Function} done called after cleanup
+ */
+function withCursor(cursor, body, done) {
```
mbroadst, line 199: Just a thought, and not one I wanted to block this review on:

This helper doesn't account for tests which return a promise, which is a similar discussion to ones we've had previously in `withClient` and friends. I wonder if what we eventually want is some helper like `withCloseable` which is internally used by `withClient`, `withMonitoredClient`, `withCursor` and maybe even something like `withSession` (though this is an existing helper in the public API). Food for thought

emadum: Agreed it would be nice for `withMonitoredClient` and `withCursor` to support promises as well as callbacks, the way `withClient` does. I think refactoring to use `withCloseable` would be a nice reduction of duplicated logic but I don't see how it would add much utility. Perhaps it's something we can work into a future PR where we're trying to write promise-based tests that need monitored clients or cursors?

PR #2362, opened 2020-05-10
test: improve shared helpers

emadum:
## Description

[NODE-2605](https://jira.mongodb.org/browse/NODE-2605)

**What changed?**

- remove `withTempDb`
- make `client` argument of `withClient` optional
- move unexported top-level function `makeCleanupFn` into `withClient`
- remove `connectToDb` helper
-  add filter function to `filterForCommands`/`filterOutCommands`/`withMonitoredClient`
- alphabetize helpers in `shared.js`

Also introduced some CI-related cleanup:
- output `eslint` version before linting
- upgrade `nvm` and use `--no-progress` option to clean up log output

### DIFF
--- a/test/functional/index.test.js
+++ b/test/functional/index.test.js
@@ -2,8 +2,7 @@
 var test = require('./shared').assert;
 var setupDatabase = require('./shared').setupDatabase;
 const expect = require('chai').expect;
-const withClient = require('./shared').withClient;
-const withMonitoredClient = require('./shared').withMonitoredClient;
+const shared = require('./shared');
 
 describe('Indexes', function() {
   before(function() {
@@ -1209,18 +1208,21 @@ describe('Indexes', function() {
       return {
         metadata: { requires: { mongodb: '<4.4' } },
         test: function() {
-          return withClient(this.configuration.newClient(), client => done => {
-            const db = client.db('test');
-            const collection = db.collection('commitQuorum');
-            testCommand(db, collection, (err, result) => {
-              expect(err).to.exist;
-              expect(err.message).to.equal(
-                '`commitQuorum` option for `createIndexes` not supported on servers < 4.4'
-              );
-              expect(result).to.not.exist;
-              done();
-            });
-          });
+          const withClient = shared.withClient.bind(this);
+          const withDb = shared.withDb.bind(this);
+          return withClient(
+            withDb('test', (db, done) => {
+              const collection = db.collection('commitQuorum');
+              testCommand(db, collection, (err, result) => {
+                expect(err).to.exist;
+                expect(err.message).to.equal(
+                  '`commitQuorum` option for `createIndexes` not supported on servers < 4.4'
+                );
+                expect(result).to.not.exist;
+                done();
+              });
+            })
+          );
         }
       };
     }
@@ -1250,7 +1252,7 @@ describe('Indexes', function() {
     function commitQuorumTest(testCommand) {
       return {
         metadata: { requires: { mongodb: '>=4.4', topology: ['replicaset', 'sharded'] } },
-        test: withMonitoredClient('createIndexes', function(client, events, done) {
+        test: shared.withMonitoredClient('createIndexes', function(client, events, done) {
           const db = client.db('test');
           const collection = db.collection('commitQuorum');
           collection.insertOne({ a: 1 }, function(err) {
--- a/test/functional/index.test.js
+++ b/test/functional/index.test.js
@@ -3,7 +3,7 @@ var test = require('./shared').assert;
 var setupDatabase = require('./shared').setupDatabase;
 const expect = require('chai').expect;
 const shared = require('./shared');
-const { withClient, withDb, withMonitoredClient } = shared;
+const { withClient, withMonitoredClient } = shared;
 
 describe('Indexes', function() {
   before(function() {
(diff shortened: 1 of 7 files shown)


emadum:
> LGTM, Makes testing easier +++++

@reggi Made some more changes after you reviewed - re-requesting. Might be easiest to just check the new commits. Sorry about that!

test/functional/shared.js
```diff
@@ -57,8 +98,26 @@ function setupDatabase(configuration, dbsToClean) {
...
 }
 
-function makeCleanupFn(client) {
-  return function(err) {
+/**
+ * Safely perform a test with provided MongoClient, ensuring client won't leak.
+ *
+ * @param {MongoClient} [client] if not provided, withClient must be bound to test function `this`
+ * @param {Function} operation (client):Promise or (client, done):void
+ * @param {Function} [errorHandler]
```
mbroadst, line 106: I don't see `errorHandler` being used anywhere, do we need it? Seems to me that just adding an `expect(client).to.exist` is enough to tell us if something went wrong, I'm not seeing the use case for returning control back to the caller.

test/functional/shared.js
```diff
@@ -57,8 +98,26 @@ function setupDatabase(configuration, dbsToClean) {
     );
 }
 
-function makeCleanupFn(client) {
-  return function(err) {
+/**
+ * Safely perform a test with provided MongoClient, ensuring client won't leak.
+ *
+ * @param {MongoClient} [client] if not provided, withClient must be bound to test function `this`
```
mbroadst, line 104: while we're here, could the first parameter also be a connection string passed to `this.configuration.newClient(connStr)`? I see cases below where you were doing `withClient(this.configuration.newClient('string'), ...`

test/functional/logger.test.js
```diff
@@ -54,29 +54,24 @@ describe('Logger', function() {
...
-        err,
-        db,
-        client
-      ) {
-        expect(err).to.not.exist;
-
-        // perform any operation that gets logged
-        db.collection('foo').findOne({}, function(err) {
-          expect(err).to.not.exist;
+      return withClient(
```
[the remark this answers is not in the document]
emadum: I don't think it'll drop the database after each test, since this isn't passing the `drop` helper option to `withDb`. 

My goal with these helpers was to move all of the startup/cleanup out of the test code, so that it is handled in a consistent manner and is more visible since it is now outside the test code.  That said I take your point, the `withDb` and `withCollection` helpers have kind of been generalized to the point of not adding value (they started out as much simpler helpers to drop the db/collection after a test).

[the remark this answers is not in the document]
emadum: I went ahead and removed them.

test/functional/index.test.js
```diff
@@ -2,8 +2,8 @@
 var test = require('./shared').assert;
 var setupDatabase = require('./shared').setupDatabase;
 const expect = require('chai').expect;
-const withClient = require('./shared').withClient;
-const withMonitoredClient = require('./shared').withMonitoredClient;
+const shared = require('./shared');
+const { withClient, withMonitoredClient } = shared;
```
mbroadst, line 6: if we're destructuring anyway it looks like just about all of these come from `shared`:
```
const {
  withClient,
  withMonitoredClient,
  assert,
  setupDatabase
} = require('./shared');
```

emadum: @mbroadst I updated [a few things](https://github.com/mongodb/node-mongodb-native/compare/d88e609..299927f) since your review, can you take a look when you get a chance? Refactoring `withMonitoredClient` exposed an issue in `withClient`, where it was swallowing errors returned by the `done(err)` callback in tests. I've updated `withClient` to now throw those errors after cleanup.

test/functional/shared.js
```diff
@@ -72,76 +135,58 @@ function makeCleanupFn(client) {
...
     });
-  };
-}
+  }
+
+  function lambda() {
+    if (!client) {
+      client = this.configuration.newClient(connectionString);
+    }
+    return client
```
mbroadst, line 144: I don't really know how much of a goal this is for us, but this does force users of this method to always treat the return type of this method as a `Promise` (as evidenced by below tests I believed). Did you try to write it in a way that would accommodate users just using callbacks? I think it means this would be more cumbersome to use if I had wanted to write this:

```
it('should do something', function(done) {
  const blah = setupSomething();

  withClient(client => {
    // do something with client
    done();
  });
});
```
now I think I would have to do:
```
it('should do something', function(done) {
  const blah = setupSomething();

  withClient(client => {
    // do something with client
    done();
  }).then(done);
});
```

emadum: I'm not sure there's any way to support the first kind of test exactly how it's written; `withClient` either needs to get a client to use as an argument, or be bound to the test function's `this` so it can make one.

The easiest option would be just doing the setup once you already have the client:

```
it('should do something', withClient((client, done) => {
  const blah = setupSomething();
  // do something with client
  done();
});
```

Or, if the setup has to happen first, you could return `withClient` inside the test, though this is a little more unwieldy. 

```
it('should do something', function() {
  const blah = setupSomething();
  return withClient.bind(this)(client, done) => {
    // do something with client
    done();
}

Technically the second option is using promises under the hood, but there aren't any promises in the individual test code.
```

mbroadst: SGTM

PR #2363, opened 2020-05-12
Node 2566/get-rid-of-readpreference.preference

reggi:


lib/core/sdam/topology.js
```diff
@@ -948,7 +948,7 @@ function translateReadPreference(options) {
   if (typeof r === 'string') {
     options.readPreference = new ReadPreference(r);
   } else if (r && !(r instanceof ReadPreference) && typeof r === 'object') {
-    const mode = r.mode || r.preference;
+    const mode = r.mode;
```
mbroadst, line 951: I think this is the only change I'm concerned about, I think its breaking behavior. If a user has passed in an object using the deprecated `preference` value, then we just dropped support for it here. I propose we still support `preference` here for compatibility.

lib/core/topologies/read_preference.js
```diff
@@ -101,7 +101,7 @@ ReadPreference.fromOptions = function(options) {
   if (typeof readPreference === 'string') {
     return new ReadPreference(readPreference, readPreferenceTags);
   } else if (!(readPreference instanceof ReadPreference) && typeof readPreference === 'object') {
-    const mode = readPreference.mode || readPreference.preference;
+    const mode = readPreference.mode;
```
mbroadst, line 104: same concern here as above

reggi:
@mbroadst updated with your edits, thanks!

lib/utils.js
```diff
@@ -21,7 +21,7 @@ var translateReadPreference = function(options) {
   if (typeof r === 'string') {
     options.readPreference = new ReadPreference(r);
   } else if (r && !(r instanceof ReadPreference) && typeof r === 'object') {
-    const mode = r.mode || r.preference;
+    const mode = r.mode;
```
mbroadst, line 24: I think this is a breaking change as well for the reasons we reverted the changes in the previous commit. Unfortunately, we do "translation" of the read preference multiple times (in most of the top-level methods on `Topology` for instance). Arguably that class is internal API, but can we prove that we didn't break something that calls one of these methods with a legacy read preference-like thing that uses `preference`?

reggi:
Closed: [Reasoning in ticket here.](https://jira.mongodb.org/browse/NODE-2566?focusedCommentId=3092955&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3092955)

PR #2364, opened 2020-05-13
add return value to bulk replaceOne

wmestrom:
## Description

Without a return value we can't use fluid bulk syntax with replaceOne statements. Example:

```
ops
  .find({ _id: 1}).upsert().replaceOne({ _id: 1, value: 2 })
  .find({ _id: 2}).upsert().replaceOne({ _id: 2, value: 1 })
```

**What changed?**

Added return statement to replaceOne()

No

### DIFF
--- a/lib/bulk/common.js
+++ b/lib/bulk/common.js
@@ -660,7 +660,7 @@ class FindOperators {
    * @returns {void} A reference to the parent BulkOperation
    */
   replaceOne(updateDocument) {
-    this.updateOne(updateDocument);
+    return this.updateOne(updateDocument);
   }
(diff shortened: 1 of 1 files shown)


wmestrom:
This doesn't seem to fix the problem, fluid syntax is actually broken in multiple ways.
`updateOne` is also broken, although it does return a value it is not of the correct type resulting in runtime errors.

mbroadst:
Thanks @wmestrom , I made this [ticket](https://jira.mongodb.org/browse/NODE-2734) to track considering this for the next breaking change.

PR #2368, opened 2020-05-14
feat(reIndex): deprecate and make standalone-only command

emadum:
## Description

The `collection.reIndex` helper has been deprecated, and will be removed
in the next major version. It is also now only supported on standalone 
servers.

[NODE-2560](https://jira.mongodb.org/browse/NODE-2560)

**What changed?**

### DIFF
--- a/lib/collection.js
+++ b/lib/collection.js
@@ -1380,7 +1380,7 @@ Collection.prototype.reIndex = deprecate(function(options, callback) {
   const reIndexOperation = new ReIndexOperation(this, options);
 
   return executeOperation(this.s.topology, reIndexOperation, callback);
-}, 'collection.reIndex is deprecated. Use db.runCommand instead.');
+}, 'collection.reIndex is deprecated. Use db.command instead.');
 
 /**
  * Get the list of all indexes information for the collection.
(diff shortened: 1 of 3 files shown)


lib/collection.js
```diff
@@ -1367,19 +1367,20 @@ Collection.prototype.dropAllIndexes = deprecate(
...
-Collection.prototype.reIndex = function(options, callback) {
+Collection.prototype.reIndex = deprecate(function(options, callback) {
   if (typeof options === 'function') (callback = options), (options = {});
   options = options || {};
 
   const reIndexOperation = new ReIndexOperation(this, options);
 
   return executeOperation(this.s.topology, reIndexOperation, callback);
-};
+}, 'collection.reIndex is deprecated. Use db.runCommand instead.');
```
mbroadst, line 1383: I think you mean `db.command`?

lib/operations/re_index.js
```diff
@@ -1,28 +1,34 @@
...
+const ServerType = require('../core/sdam/common').ServerType;
+const MongoError = require('../core').MongoError;
 
-class ReIndexOperation extends CommandOperation {
+class ReIndexOperation extends CommandOperationV2 {
   constructor(collection, options) {
-    super(collection.s.db, options, collection);
+    super(collection.s.db, options);
+    this.collection = collection;
+    this.command = { reIndex: collection.collectionName };
```
mbroadst, line 14: remember the guiding principle of operations v2 is: 
- save all things needed for building the command in the constructor. Try to make an instance of this class represent the [command](https://docs.huihoo.com/mongodb/3.4/reference/command/reIndex/index.html) as much as possible. In this case its a really limited command, it might just be that `this.collection` is saved off. 
- actually build the command in `execute`

PR #2370, opened 2020-05-14
feat: remove reIndex

emadum:
## Description

The deprecated `collection.reIndex` helper has been removed.

[NODE-2560](https://jira.mongodb.org/browse/NODE-2560)

**What changed?**


emadum:
I wasn't sure whether to delete the tests that previously used this helper, or refactor them to use `db.runCommand`, so they are just skipped for now.

test/functional/db.test.js
```diff
@@ -205,7 +205,7 @@ describe('Db', function() {
   /**
    * An example showing how to force a reindex of a collection.
    */
-  it('shouldCorrectlyForceReindexOnCollection', {
+  it.skip('shouldCorrectlyForceReindexOnCollection', {
```
mbroadst, line 208: I think its safe to remove this if we're removing `reIndex` outright. The description shows us that the test is "An example showing how to force a reindex of a collection"

mbroadst: ditto for the below tests

PR #2372, opened 2020-05-17
feat: introduce typescript migration pipeline

mbroadst:
## Description
This introduces a tool for incrementally migrating our codebase to typescript through the use of a gulp pipeline. This process allows us to ensure that no major bugs are introduced during the migration, and that our migration is conservative (no transpilation, or major refactorings in the initial move).

I've included two migrated classes (`ReadConcern`, `WriteConcern`) for illustrative purposes. The process includes a `git mv` to preserve history, as well as introduction of minimal type information to preserve the existing structure of the output files. A comparison in the PR can help us observe that minimal changes (if any) are made to the original file - a major goal of the first phase of this transition to TypeScript.


package.json
```diff
@@ -68,13 +72,13 @@
   },
   "scripts": {
     "build:evergreen": "node .evergreen/generate_evergreen_tasks.js",
-    "build:types": "tsc",
+    "build:types": "gulp -f test/tools/ts-pipeline.js",
```
nbbeeken, line 75: Publishing our own types will have to wait until we've completed conversion, so nit: lets rename this to `build:ts`

I can't comment on it but you should delete line 9 (types/index.d.ts)

package.json
```diff
@@ -68,13 +72,13 @@
...
     "check:atlas": "node test/tools/atlas_connectivity_tests.js",
     "check:bench": "node test/benchmarks/driverBench",
     "check:coverage": "nyc mocha --timeout 60000 --recursive test/functional test/unit",
     "check:lint": "eslint -v && eslint index.js lib test",
     "check:test": "mocha --recursive test/functional test/unit",
-    "check:types": "tsc -p tsconfig.check.json",
+    "check:types": "tsc -p src",
```
nbbeeken, line 81: if you go with my other comment on the scripts lets change to `check:ts` and regardless the command can be `tsc --noEmit`, right now this just crashes.

src/read_concern.ts
```diff
@@ -0,0 +1,65 @@
+'use strict';
```
nbbeeken, line 1: Can we omit this now that it is generated by ts?

src/read_concern.ts
```diff
@@ -0,0 +1,65 @@
...
+  static get LINEARIZABLE() {
+    return 'linearizable';
+  }
+
+  static get SNAPSHOT() {
+    return 'snapshot';
+  }
+}
+
+module.exports = ReadConcern;
```
nbbeeken, line 65: (Just a question, for discussion sake) We don't wanna use es6 `export` here yet cus it does emit some extra code but do we forsee a time when we're fine with that extra code? (maybe when the whole code base is ts, we would consider migrating then?)

The extra code:
```js
Object.defineProperty(exports, "__esModule", { value: true })
```

mbroadst: If the guiding principle is that we are using TypeScript for typing, not adding language features (transpiling), then I think we would want to support es6 modules when node does. That means we couldn't support es6 modules until we set a minimum supported version of Node 13 (I believe). 

I'm trying to better understand the use case, is it just more ergonomic to use `import` and `export` declarations? Or are you perceiving some other potential issues we might run into?

package.json
mbroadst: oops thanks for catching that, left it in when I was playing around

mbroadst:
@emadum yes, that's the goal on the first pass - no changes to the output javascript at all. We now have minor changes because we're depending on typescript to add `"use strict";` but other than that there should be no expected changes

src/read_concern.ts
nbbeeken: I think we'll still have type definition emitting problems unless tsc makes the fix, or we rely on es6 module syntax. So we can just wait to see if that is fixed when we get to an all ts state.

PR #2378, opened 2020-05-18
[NODE-2564] fix: clarify handle wrong set name single topology (3.6)

reggi:
[NODE-2564](https://jira.mongodb.org/browse/NODE-2564)

```
fix: clarify handle wrong set name single topology

Adds check to see if topologyDescription setName matches the serverDescription setName. In instances where they don't match, sets serverDescription type to Unknown. Incorporates spec tests from SPEC-1656 (bc864e6f9ccbdd162658cfbc08bfb1cb713143c5).

NODE-2564
```


lib/core/sdam/server_description.js
```diff
@@ -143,6 +145,29 @@ class ServerDescription {
...
     );
   }
+
+  /**
+   * Creates a new ServerDescription from this
+   * serverDescription with no ismaster response.
+   *
+   * @returns ServerDescription
+   */
+  reset() {
```
mbroadst, line 155: can we rename this to something that indicates the immutability of the action? This method name indicates action on the `ServerDescription` instance itself (mutating internal state), when its really a factory method to return a clone of this instance with its state transitioned to `Unknown`. Perhaps a better name would be `toUnknown` or something similar?

lib/core/sdam/topology_description.js
```diff
@@ -132,6 +132,8 @@ class TopologyDescription {
     let maxElectionId = this.maxElectionId;
     let commonWireVersion = this.commonWireVersion;
 
+    serverDescription = serverDescription.resetIfWrongSetName(setName);
```
mbroadst, line 135: I have a medium preference for bringing the logic inside of `resetIfWrongSetName` back into this method, and not hiding it within the implementation details of the `ServerDescription` itself. Perhaps if we were dealing with a [rule of threes](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) situation where this method would be used often in disparate areas of code it would make sense to make this refactor, but as it stands it requires the reader to jump between files to understand the logical flow of what's happening in this particularly complex piece of machinery.

PR #2385, opened 2020-05-20
Fix memory leak

dead-horse:
## Description

**What changed?**

- Clean timeout tasks when process wait task queue.
- End session when query error.

Try to fix https://jira.mongodb.org/browse/NODE-2413?focusedCommentId=3094451&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3094451

### DIFF
--- a/lib/cursor/core_cursor.js
+++ b/lib/cursor/core_cursor.js
@@ -434,6 +434,10 @@ class CoreCursor extends Readable {
     }
 
     function done(err, result) {
+      if (err && cursor._endSession) {
+        cursor._endSession();
+      }
+
       if (
         cursor.cursorState.cursorId &&
         cursor.cursorState.cursorId.isZero() &&
--- a/lib/cmap/connection_pool.js
+++ b/lib/cmap/connection_pool.js
@@ -441,11 +441,17 @@ function processWaitQueue(pool) {
     return;
   }
 
-  while (pool[kWaitQueue].length && pool.availableConnectionCount) {
+  while (pool[kWaitQueue].length) {
+    // remove timeout tasks in queue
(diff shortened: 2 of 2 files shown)


lib/cursor/core_cursor.js
```diff
@@ -434,6 +434,10 @@ class CoreCursor extends Readable {
     }
 
     function done(err, result) {
+      if (err && cursor._endSession) {
```
dead-horse, line 437: After I monkey patched mongodb with `processWaitQueue` and deployed in my environment, when this happen again, the queue is no more memory leak if I set `waitQueueTimeoutMS`. But there's another place where there's a memory leak.

Then I take a heapdump

image
image

But I'm not sure if we can fix this memory leak like this.

dead-horse:
@mbroadst any thoughts?

mbroadst:
@dead-horse thank you very much for the fix here, I think you've indeed found some bugs. I will be working on this later today, and try to include some test cases. The fix will go out in next week's patch release.

dead-horse:
@mbroadst  Any progress? Do we need to provide more information?

mbroadst:
@dead-horse the fix will be merged in as part of #2388 , and put into a patch release today. Thank you for your patience.

dead-horse:
thanks @mbroadst 

#2388 fixed the wait queue problem, but I'm not sure how the cursor problem arose and how to solve it.

mbroadst:
@dead-horse 3.5.8 has been released with a fix for the memory leak in the wait queue (#2388) and the implicit session issue (#2392). I'm going to close this PR now, let's speak on JIRA if you continue to experience issues.

PR #2388, opened 2020-05-27
fix: always clear cancelled wait queue members during processing

mbroadst:
Each time the wait queue in the connection pool is processed it is currently gated by having available connections. This can result in a memory leak where many wait queue members are cancelled, but the pool is unable to clear them out before all connections are used for new operations.

NODE-2413

### DIFF
--- a/lib/cmap/connection_pool.js
+++ b/lib/cmap/connection_pool.js
@@ -198,6 +198,10 @@ class ConnectionPool extends EventEmitter {
     return this[kConnections].length;
   }
 
+  get waitQueueSize() {
+    return this[kWaitQueue].length;
+  }
+
   /**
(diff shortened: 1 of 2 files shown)


lib/cmap/connection_pool.js
```diff
@@ -449,13 +453,17 @@ function processWaitQueue(pool) {
     return;
   }
 
-  while (pool[kWaitQueue].length && pool.availableConnectionCount) {
+  while (pool[kWaitQueue].length) {
```
reggi, line 456: Could we reuse `pool.waitQueueSize` here?

mbroadst: sure! changed

PR #2390, opened 2020-05-27
[NODE-2594] RetryableWriteError Spec-1666 (3.6)

reggi:
https://jira.mongodb.org/browse/NODE-2594

```
test: RetryableWriteError label for transactions

Adds spec tests from SPEC-1666 (49589d66d49517f10cc8e1e4b0badd61dbb1917e). Ensures Does not add the RetryableWriteError label to errors that occur during a write within a transaction excepting for commitTransaction and abortTransaction.

NODE-2594
```


mbroadst:
@reggi why were the spec tests for the `convenient-api` removed?

PR #2391, opened 2020-05-27
fix: ensure implicit session ended on query failure

mbroadst:
This is a fix and a refator, which ensures that implicit sessions are ended as soon as possible when cursor commands are executed.


mbroadst:
closed in favor of #2392

PR #2392, opened 2020-05-27
fix: ensure implicit sessions are ended consistently

mbroadst:
This is a fix and a refator, which ensures that implicit sessions are ended as soon as possible when cursor commands are executed. The current code path calls `_endSessions` in a confusing number of locations, and in some cases can accidentally call the method twice in a row. Reducing the number of calls to the method, and ensuring its called only after server command responses and the `close` method should prevent these errors

NODE-2630

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -832,9 +832,7 @@ class Cursor extends CoreCursor {
       const fetchDocs = () => {
         cursor._next((err, doc) => {
           if (err) {
-            return cursor._endSession
-              ? cursor._endSession(() => handleCallback(cb, err))
-              : handleCallback(cb, err);
+            return handleCallback(cb, err);
           }
 
           if (doc == null) {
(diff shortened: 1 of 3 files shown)


lib/cursor.js
```diff
@@ -914,38 +912,18 @@ class Cursor extends CoreCursor {
...
+      this.s.state = CursorState.CLOSED;
+      if (!options.skipKillCursors) {
+        // Kill the cursor
+        this.kill();
       }
 
-      return new this.s.promiseLibrary(resolve => {
-        this._endSession(() => completeClose().then(resolve));
+      this._endSession(() => {
+        this.emit('close');
```
reggi, line 923: I'm not 100% how session works, but just wanted to bring this to your attention. Is it possible we might need a `this.emit('close');`  outside of the `this._endSession` if there is no session?

B/C the previous code had an emit in the `completeClose` https://github.com/mongodb/node-mongodb-native/blob/3.5/lib/cursor.js#L923-L948

mbroadst: [_endSession](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/core/cursor.js#L373-L398) always calls its callback, even if its not actually sending a command to the server. I had originally maintained the two separate paths that the old code took, but I think the result is the same. The difference now is that you'll always see `_endSession` in the stacktrace, but I imagine that was the case most of the time previously.

reggi: Awesome, thanks for the explanation.

PR #2394, opened 2020-05-28
fix(GridFS): emit error on bad options

emadum:
## Description

Errors related to handling the start/end options for `createDownloadStream`
did not propagate correctly.

[NODE-2623](https://jira.mongodb.org/browse/NODE-2623)

**What changed?**

### DIFF
--- a/lib/gridfs-stream/download.js
+++ b/lib/gridfs-stream/download.js
@@ -278,56 +278,63 @@ function init(self) {
     if (error) {
       return __handleError(self, error);
     }
-    try {
-      if (!doc) {
-        var identifier = self.s.filter._id ? self.s.filter._id.toString() : self.s.filter.filename;
-        var errmsg = 'FileNotFound: file ' + identifier + ' was not found';
-        var err = new Error(errmsg);
-        err.code = 'ENOENT';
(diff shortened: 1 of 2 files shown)


lib/gridfs-stream/download.js
```diff
@@ -278,52 +278,56 @@ function init(self) {
...
       return __handleError(self, error);
     }
-    if (!doc) {
-      var identifier = self.s.filter._id ? self.s.filter._id.toString() : self.s.filter.filename;
-      var errmsg = 'FileNotFound: file ' + identifier + ' was not found';
-      var err = new Error(errmsg);
-      err.code = 'ENOENT';
-      return __handleError(self, err);
-    }
+    try {
```
mbroadst, line 281: this feels like a bit of a sledgehammer approach, is there a way to satisfy the issues in the ticket without wrapping the world in a try/catch?

emadum: Well, I could specifically wrap the calls to `handleStartOptions` and `handleEndOptions` in try/catch, but this approach seemed more robust. What makes it a "sledgehammer approach?" I'm open to other ideas.

PR #2397, opened 2020-05-29
allowDiskUse option for find should be documented as only being supported in 4.4+ (3.6)

reggi:
https://jira.mongodb.org/browse/NODE-2622


mbroadst:
@reggi I think we need a few things for this work:
- it seems like there's something wrong with the spec test format, the errors you are receiving in evergreen indicate that some collection name is probably `undefined`
- secondly, I think we will need to explicitly return an error in pre-3.2 servers when this option is specified

PR #2400, opened 2020-06-01
refactor: translate modern tls options from connect options object to ssl

bnielsen1965:
refactor: translate modern tls options from connect options object to ssl

Translation of `tls` variants into ssl options is now applied to the options
object in the connect operation in the same manner as the url query string.

### DIFF
--- a/lib/core/index.js
+++ b/lib/core/index.js
@@ -46,5 +46,6 @@ module.exports = {
   ScramSHA1: require('./auth/scram').ScramSHA1,
   ScramSHA256: require('./auth/scram').ScramSHA256,
   // Utilities
-  parseConnectionString: require('./uri_parser')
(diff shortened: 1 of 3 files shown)


bnielsen1965:
Is anyone available to do a review of this PR? The latest 3.5 release only supports tls settings in the connect URL and ignores the settings if they are in the connect options object. I'd like to avoid the deprecated ssl settings but I also don't want to use a super long connect URL.

mbroadst:
@bnielsen1965 thanks for the contribution! Could you please make a JIRA ticket about this issue, I'll review shortly.

bnielsen1965:
Jira ticket created...

[NODE-2638](https://jira.mongodb.org/browse/NODE-2638)

mbroadst:
Issue is resolved per the above mentioned jira ticket

PR #2407, opened 2020-06-07
Update aggregation.md

strigonLeader:
There appears to be a mismatch between the example code and description introduced in c43dc208ac167ea1431baa2d1bf4b3a738fd0404 when the example was fixed the text after the example was not updated.


mbroadst:
Thanks for the contribution @strigonLeader !

PR #2410, opened 2020-06-09
test: fix a number of our most notorious flakey tests

mbroadst:
- remove most `core/operations.test.js` tests using cursors in a
   sharded environment. We already have coverage for these commands
   and the returned batch sizes from our internal `_next` method
   are inconsistent

 - massive overhaul of the change streams tests which ensure that
   all async operations are awaited before cleanup, and that the
   change stream and client are closed in the correct order

 - refactor of the `defer` plugin to ensure that deferred actions
   happen LIFO, and in order rather than all at the same time


test/functional/change_stream.test.js
```diff
@@ -176,26 +180,29 @@ describe('Change Streams', function() {
...
     // The actual test we wish to run
     test: function(done) {
       let closed = false;
-      const close = _err => {
-        if (closed) {
-          return;
-        }
+      function close(err) {
+        if (closed) return;
         closed = true;
```
reggi, line 191: How is closed ever set to `true`?

PR #2411, opened 2020-06-10
Fixed data path

iyashsoni:
Without this, mongo will look for /data path i.e. starting from the root of the System. `./data` will fix it.

## Description

**What changed?** `data` directory path in the documentation.

**Are there any files to ignore?** No.

### DIFF
--- a/docs/reference/content/quick-start/quick-start.md
+++ b/docs/reference/content/quick-start/quick-start.md
@@ -43,7 +43,7 @@ For complete MongoDB installation instructions, see [the manual](https://docs.mo
 3. Install and start a ``mongod`` process.
 
 ```
-mongod --dbpath=/data
+mongod --dbpath=./data
 ```
 
 You should see the **mongod** process start up and print some status information.


reggi:
@iyashsoni Thanks for this PR, however I don't think that this is a necessary change.

I think the way it is currently is intentional. The default directory is encouraged to be `/data` at the root of a drive.

You can see examples of this in the docs:

https://docs.mongodb.com/manual/tutorial/manage-mongodb-processes/#specify-a-data-directory

iyashsoni:
@reggi, In that case, shouldn't there be a clear note saying this? Something like:
> We recommend having data directory at the root, but if you want you can modify it.

reggi:
@iyashsoni Hey! The default db path is the root of the file system. Seen in the docs [here](https://docs.mongodb.com/manual/reference/configuration-options/#storage.dbPath), having this flag in the docs isn't really necessary, but it shows users that it can be changed, as you have in your PR. I'm going to say this is ok to leave this like this.

PR #2415, opened 2020-06-16
NODE-2588: Refactor FindAndModifyOperation to use CommandOperationV2

reggi:


test/functional/shared.js
```diff
@@ -164,6 +164,7 @@ function withClient(client, callback) {
  *
  * @param {string|Array|Function} commands commands to filter for
  * @param {object} [options] options to pass on to configuration.newClient
+ * @param {Function | object} [options.connectionString] connection string options
```
reggi, line 167: This change needs explanation. I wanted to have the ability to pass `this.configuration.writeConcernMax()` to withMonitoredClient like this:

```
withMonitoredClient({ connectionString: (self) => self.configuration.writeConcernMax())
```

I didn't need to use this function, the tests I switched over passed without that `url` so I'm willing to remove this, but I still think it could be useful in the future.

reggi:
In `b7f39f3` I had added all the accrued improvements I had discovered along the way, changes past this commit aren't necessary for the work.

PR #2416, opened 2020-06-16
Handle null value of change in ChangeStream.processNewChange

dobesv:
The `change` can be `null` in `processNewChange` when the server connection is closed but the call to `cacheResumeToken` did not allow for that.


emadum:
Thanks for the PR! Closing, as this was addressed in #2421.

PR #2419, opened 2020-06-29
NODE-1452: convert codebase to TypeScript

mbroadst:
## Description
This patch constitutes the first phase of converting the whole codebase to TypeScript
- [x] Use git to move all files from lib to a new src directory, preserving history as much as possible
- [x] Convert all files in src to use TypeScript with minimal changes to the existing code
  - use the newly developed ts-convert.sh script to automate as much of the conversion as possible
  - this first pass will not be ideal, and will likely make heavy use of any
- [x] Convert our test runner to use ts-node to run all tests against TypeScript-ified code
  - Maintain all tests in vanilla JavaScript
- [x] Add a prepublish script for using the existing TypeScript "pipeline" script to generate readable transpiled javascript before publishing the module

**What changed?**
everything


mbroadst:
It looks like some of the `git mv`s were screwed up, closing to correct those

mbroadst:
reopening, looks like this might just be issues with GitHub's UI

src/admin.ts
```diff
@@ -0,0 +1,289 @@
+import { applyWriteConcern } from './utils';
```
mbroadst, line 1: this isn't actually a new file, if you check out the source locally and run `git log --follow src/admin.ts` you'll see the whole history from its javascript past.

package.json
```diff
@@ -72,12 +74,14 @@
...
     "build:evergreen": "node .evergreen/generate_evergreen_tasks.js",
     "build:ts": "gulp -f test/tools/ts-pipeline.js",
-    "check:atlas": "node test/tools/atlas_connectivity_tests.js",
     "check:bench": "node test/benchmarks/driverBench",
     "check:coverage": "nyc mocha --timeout 60000 --recursive test/functional test/unit",
-    "check:lint": "eslint -v && eslint index.js lib test",
+    "check:lint": "eslint -v && eslint src/**/*.ts test",
     "check:test": "mocha --recursive test/functional test/unit",
     "check:ts": "tsc --noEmit",
+    "check:atlas": "mocha --config \"test/manual/mocharc.json\" test/manual/atlas_connectivity.test.js",
```
mbroadst, line 82: I didn't work too hard at this, it might be possible to specify a `mocharc.json` just for this folder, overriding the top level, but it didn't immediately work so I moved on.

nbbeeken: If you add `"extends": "../../mocharc.json",` to the config in the tests/manual folder it should work by only using ts-node for the tests in this directory, Apparently the config file is handled by yargs, so whatever it supports will work for us here.

src/utils.ts
```diff
@@ -518,8 +532,8 @@ function decorateWithReadConcern(command, coll, options) {
   }
 }
 
-const emitProcessWarning = msg => process.emitWarning(msg, 'DeprecationWarning');
-const emitConsoleWarning = msg => console.error(msg);
+const emitProcessWarning = (msg: any) => process.emitWarning(msg, 'DeprecationWarning');
+const emitConsoleWarning = (msg: any) => console.error(msg);
 const emitDeprecationWarning = process.emitWarning ? emitProcessWarning : emitConsoleWarning;
```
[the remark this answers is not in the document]
nbbeeken: Figured out why I was getting this, my vscode was using typescript v3.9 and we're on v3.8, any chance we can version bump while we're at this stage?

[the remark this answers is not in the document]
mbroadst: whoops, thanks for catching that. Updated

package.json
```diff
@@ -72,12 +74,14 @@
...
+    "check:ocsp": "mocha --config \"test/manual/mocharc.json\" test/manual/ocsp_support.test.js",
+    "prepublish": "npm run build:ts",
```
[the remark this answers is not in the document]
mbroadst: hm wonder where that's coming from?

package.json
mbroadst: see I don't want it to extend the existing config, I want this one to effectively be the root config. I tried adding `"root": true` for a `.mocharc.json` in this folder, but for whatever reason it was picking up the top-level one first. I'd prefer to punt on this until after we get this PR merged, given the complexity of the change

PR #2420, opened 2020-06-29
NODE-2649/3.5/command-does-not-inherit-rp

reggi:


lib/operations/command_direct.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const CommandOperationV2 = require('./command_v2');
+const defineAspects = require('./operation').defineAspects;
+const Aspect = require('./operation').Aspect;
+
+class CommandDirect extends CommandOperationV2 {
```
mbroadst, line 7: I actually think this wrapping is great, but let's call it `RunCommand`?

lib/operations/command_v2.js
```diff
@@ -15,10 +15,18 @@ class CommandOperationV2 extends OperationBase {
   constructor(parent, options, operationOptions) {
     super(options);
 
+    const noInherit = options && options.noInherit;
```
mbroadst, line 18: Now that you're using a wrapper method, can't we use "aspects" to skip the inheritance?

lib/operations/command_v2.js
```diff
@@ -15,10 +15,18 @@ class CommandOperationV2 extends OperationBase {
...
+    const inherit = noInherit ? {} : parent;
+
     this.ns = parent.s.namespace.withCollection('$cmd');
-    this.readPreference = resolveReadPreference(parent, this.options);
-    this.readConcern = resolveReadConcern(parent, this.options);
-    this.writeConcern = resolveWriteConcern(parent, this.options);
+    try {
```
mbroadst, line 22: I think it would be preferable not to add a try/catch to solve this problem, but rather modify `resolveReadPreference` to handle a nullish input for `parent`

Also, FWIW didn't you make `ReadPreference.resolve` ?

PR #2422, opened 2020-07-01
NODE-2433/lift-restriction-auth-source

reggi:


test/spec/auth/auth.rst
```diff
@@ -0,0 +1,1218 @@
+.. role:: javascript(code)
```
mbroadst, line 1: I'd prefer not to merge in the whole spec if possible, seems like needless text duplication across all of our driver repos

test/spec/connection-string/connection-string-spec.rst
```diff
@@ -0,0 +1,392 @@
+.. role:: javascript(code)
```
mbroadst, line 1: same with this one

test/spec/uri-options/uri-options.rst
```diff
@@ -0,0 +1,393 @@
+=========================
```
mbroadst, line 1: and this one

test/spec/auth/connection-string.yml
```diff
@@ -287,6 +287,11 @@ tests:
         uri: "mongodb://localhost/foo"
         valid: true
         credential: ~
+    -
```
mbroadst, line 290: was this the only test added as part of this work?

reggi: I cherry-picked this spec, this is the only change. I removed the `.rst` files.

PR #2424, opened 2020-07-02
NODE-2649/3.6/command-does-not-inherit-rp

reggi:


lib/core/topologies/read_preference.js
```diff
@@ -123,6 +123,64 @@ ReadPreference.fromOptions = function(options) {
...
+/**
+ * Replaces options.readPreference with a ReadPreference instance
+ */
+ReadPreference.translate = function(options) {
+  var r = null;
+  if (options.readPreference) {
+    r = options.readPreference;
+  } else {
+    return options;
+  }
```
emadum, line 166: This looks a little convoluted here, I think it'd be cleaner to do something like:

```
if (options.readPreference == null) return options;
const r = options.readPreference;
```

lib/operations/command_v2.js
```diff
@@ -15,10 +15,13 @@ class CommandOperationV2 extends OperationBase {
   constructor(parent, options, operationOptions) {
     super(options);
 
+    const noInherit = this.hasAspect(Aspect.NO_INHERIT_OPTIONS);
+    const inherit = noInherit ? {} : parent;
```
emadum, line 19: Nit: this could just be one line, `const inherit = this.hasAspect(Aspect.NO_INHERIT_OPTIONS) ? {} : parent;`, I'm generally not a huge fan of variable declarations that are only used once like this.

lib/core/topologies/read_preference.js
mbroadst: +1

lib/operations/command_v2.js
```diff
@@ -15,13 +15,14 @@ class CommandOperationV2 extends OperationBase {
   constructor(parent, options, operationOptions) {
     super(options);
 
-    const noInherit = this.hasAspect(Aspect.NO_INHERIT_OPTIONS);
-    const inherit = noInherit ? {} : parent;
+    if (this.hasAspect(Aspect.NO_INHERIT_OPTIONS)) {
+      parent = undefined;
+    }
```
emadum, line 20: I think this needs to be moved down below the next line, `    this.ns = parent.s.namespace.withCollection('$cmd');` or something like that, because `parent.s` won't exist anymore now.

lib/operations/command_v2.js
```diff
@@ -15,8 +15,16 @@ class CommandOperationV2 extends OperationBase {
...
 
+    if (parent.s.logger) {
+      this.logger = parent.s.logger;
+    } else if (parent.s.db && parent.s.db.logger) {
+      this.logger = parent.s.db.logger;
+    }
     this.ns = parent.s.namespace.withCollection('$cmd');
-    this.readPreference = resolveReadPreference(parent, this.options);
+
+    if (this.hasAspect(Aspect.NO_INHERIT_OPTIONS)) parent = {};
```
mbroadst, line 25: OK, sorry we have a bit of a mess going back and forth on this PR related to `parent` and I'd like to answer this once and for all:

Right now this constructor makes a hard assumption that `parent` exists (see line 18 above, we access `parent.s` without any checks). For the purposes of "inherited properties" here we want either to pass in some class which can have inheritable properites (`MongoClient`/`Db`/`Collection`) or `undefined`, the signature to `ReadPreference.resolve` should have a first parameter `parent?: MongoClient|Db|Collection`, not `parent: MongoClient|Db|Collection|any`

I also think I confused you a bit by suggesting we reuse the `parent` variable for these purposes, let's just go ahead and create a intermediate value to preserve safety and readability:
```
constructor(parent, ..) {

   const propertyProvider = this.hasAspect(Aspect.NO_INHERIT_OPTIONS) ? undefined : parent;
   this.readPreference = ReadPreference.resolve(propertyProvider, this.options);
   ...
}
```

lib/operations/command_v2.js
reggi: Reading this after it's been fixed was super confusing. This is out of date, odd it is not labeled.

PR #2427, opened 2020-07-07
chore: sync retryable reads tests for `listIndex` changes

mbroadst:
NODE-2266


mbroadst:
@reggi it's a valid concern, but not one that can be easily addressed unless we have a consistent versioning strategy for our specifications. I attempt to solve that in the `mspec` project, by publishing specs to npmjs and using semver more carefully, but that project isn't actively being worked on.

PR #2429, opened 2020-07-07
refactor: add `mongodb+srv` tests to atlas connectivity tests

mbroadst:
This simplifies how we test our many connectivity tests while also adding a few new ones for `mongodb+srv` hosts. Specifically, it converts the connectivity tests to a new "manual" style test using mocha, and then merges many environment variables into a single JSON blob.

NODE-2021


.evergreen/config.yml
```diff
@@ -156,14 +156,23 @@ functions:
...
       type: test
       params:
-        working_dir: src
         silent: true
-        script: >
-          # DO NOT ECHO WITH XTRACE (which PREPARE_SHELL does)
+        working_dir: src
+        script: |
+          cat <<EOT > prepare_atlas_connectivity.sh
+          export ATLAS_CONNECTIVITY='${ATLAS_CONNECTIVITY}'
```
nbbeeken, line 163: Just tryna understand, is ATLAS_CONNECTIVITY set by EVG? and it seems to be a JSON string, whats the documentation around its shape?

mbroadst: Yeah `ATLAS_CONNECTIVITY` is a variable in EVG with the following shape:
```
{
  "name_of_configuration": [
    'connection-string',
    'another-connection-string'
  ]
}
```
They used to all be individual variables like `ATLASFREE`, `ATLAS_TLS_11`, etc, so this is an improvement which lets us easily add new hosts (a new one for `mongodb+srv` was added in this ticket for each variant) without having to maintain a huge number of variables.

PR #2435, opened 2020-07-09
NODE-2671: IPv6 is not supported when using dns service discovering

nbbeeken:
https://jira.mongodb.org/browse/NODE-2671

Both server class and monitoring class split the address string on `:` and assumed it would return address and port tuple. To handle IPv6 addresses we need to use the number at the end of the address and keep the ipv6 address from being split. I consolidated the logic into ServerDescription getters.

### DIFF
--- a/src/sdam/monitor.ts
+++ b/src/sdam/monitor.ts
@@ -16,6 +16,7 @@ import {
   ServerHeartbeatSucceededEvent,
   ServerHeartbeatFailedEvent
 } from './events';
+import { Server } from './server';
 
 const kServer = Symbol('server');
 const kMonitorId = Symbol('monitorId');
@@ -48,7 +49,7 @@ class Monitor extends EventEmitter {
   [kCancellationToken]: any;
   [kMonitorId]: any;
 
-  constructor(server: any, options: any) {
+  constructor(server: Server, options: any) {
     super(options);
     this[kServer] = server;
     this[kConnection] = undefined;
@@ -74,12 +75,11 @@ class Monitor extends EventEmitter {
     });
 
     // TODO: refactor this to pull it directly from the pool, requires new ConnectionPool integration
-    const addressParts = server.description.address.split(':');
     const connectOptions = Object.assign(
       {
         id: '<monitor>',
-        host: addressParts[0],
-        port: parseInt(addressParts[1], 10),
+        host: server.description.host,
+        port: server.description.port,
         connectionType: Connection
       },
       server.s.options,
--- a/src/sdam/server_description.ts
+++ b/src/sdam/server_description.ts
@@ -43,7 +43,7 @@ const ISMASTER_FIELDS = [
  * Internal type, not meant to be directly instantiated
  */
 class ServerDescription {
-  address: any;
+  address: string;
   error: any;
   roundTripTime: any;
   lastUpdateTime: any;
@@ -133,6 +133,16 @@ class ServerDescription {
    */
   get isWritable() {
     return WRITABLE_SERVER_TYPES.has(this.type);
+  }
+
+  get host() {
+    const chopLength = `:${this.port}`.length;
+    return this.address.slice(0, -chopLength);
+  }
+
+  get port() {
+    const addressParts = this.address.split(':');
+    return Number.parseInt(addressParts[addressParts.length - 1], 10);
   }
 
   /**
--- a/src/sdam/server_description.ts
+++ b/src/sdam/server_description.ts
@@ -141,8 +141,8 @@ class ServerDescription {
   }
 
   get port() {
-    const addressParts = this.address.split(':');
-    return Number.parseInt(addressParts[addressParts.length - 1], 10);
+    const port = this.address.split(':').pop();
+    return port ? Number.parseInt(port, 10) : port;
   }
 
   /**
--- a/test/unit/sdam/monitoring.test.js
+++ b/test/unit/sdam/monitoring.test.js
@@ -96,7 +96,7 @@ describe('monitoring', function() {
     });
   });
 
-  describe('Monitor class', function() {
+  describe('Monitor', function() {
     it('should connect and issue an initial server check', function(done) {
       mockServer.setMessageHandler(request => {
         const doc = request.document;
--- a/test/unit/sdam/monitoring.test.js
+++ b/test/unit/sdam/monitoring.test.js
@@ -9,10 +9,8 @@ const { ServerDescription } = require('../../../src/sdam/server_description');
 class MockServer {
   constructor(options) {
     this.s = {};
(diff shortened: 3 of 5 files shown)


test/unit/sdam/server_description.test.js
```diff
@@ -41,4 +41,10 @@ describe('ServerDescription', function() {
       });
     });
   });
+
+  it('should sensibly parse an ipv6 address', function() {
+    const description = new ServerDescription('abcd:f::abcd:abcd:abcd:abcd:27017');
```
mbroadst, line 46: Per [rfc3986](https://www.ietf.org/rfc/rfc3986.txt) section 3.2.2, this is actually an invalid representation of the address. It must be represented in text as: `[abcd:f::abcd:abcd:abcd:abcd]:27017`. I'm trying to think of better ways for us parse these values, and wondering if we can re-use the url parser in node to handle these rules for us? Maybe if you just wrap the string in a dummy `http://${input}/` and parse that? The thing is that we're _already_ parsing from the `MongoClient` constructor, so we might be introducing new complexity just for testing purposes.

src/sdam/server_description.ts
```diff
@@ -135,6 +135,16 @@ class ServerDescription {
     return WRITABLE_SERVER_TYPES.has(this.type);
   }
 
+  get host() {
```
mbroadst, line 138: This isn't documented anywhere, but one of my goals with the SDAM and CMAP rewrites was to abstract endpoints we could connect to. I prefer names like `address` or `uri` to `host` and `port` because we might eventually want to connect to something that doesn't use a classic host/port (even domain sockets break this, but what if we wanted to connect to.. a file or something). I think adding `host` and `port` on `ServerDescription` is coupling the concerns of creating a TCP socket with something that shouldn't care about transports.

Maybe we can spend a little bit of time thinking about whether such a goal is possible with limited changes: does it make sense for `connect` to check for an `address` property and do the parsing there? Does it make sense to introduce a `ServerAddress` type, and make the above intent more explicit throughout the codebase?

test/unit/sdam/server_description.test.js
nbbeeken: We pull the IPv6 address out of the square brackets here 
https://github.com/mongodb/node-mongodb-native/blob/c8495e590c9a5ed4ff7e513376b92e677c61c801/src/connection_string.ts#L688 
and then reconstruct it as ip colon port here:
https://github.com/mongodb/node-mongodb-native/blob/c8495e590c9a5ed4ff7e513376b92e677c61c801/src/sdam/topology.ts#L154-L155
So this string matches what gets passed to the ServerDescription typically, I can try to change that behavior and work out what bugs pop up there? Or I can document this reason why this test has a poorly formatted address

src/sdam/server_description.ts
nbbeeken: I think that's a good idea, it feels a bit brittle to rely on these host and port getters I'll investigate what that would look like. I also have to port the changes to 3.5 and 3.6, but my inclination in 4.0 would be to lean on typescript types to give us a union of possible addresses and use an indicator to differentiate what type of address we'd have. While that can literally translate to javascript is might not feel as assured as when you have type checking. 
But that's just an inclination, I do think this is worthy of breaking out into separate work

mbroadst: ok, can you make a ticket to do this work in the future?

test/unit/sdam/monitoring.test.js
```diff
@@ -97,7 +96,7 @@ describe('monitoring', function() {
     });
   });
 
-  describe('Monitor', function() {
+  describe('Monitor class', function() {
```
mbroadst, line 99: The convention has been to `describe('ClassName', ...)` without further modifiers. Does this context help in your mocha side bar?

nbbeeken: Oh yea whoops, because this test shares the same name as an import it was doing funky things to the discovery, I didn't mean to keep this
image

src/sdam/server_description.ts
nbbeeken: Yes! https://jira.mongodb.org/browse/NODE-2715

src/sdam/server_description.ts
```diff
@@ -135,6 +135,16 @@ class ServerDescription {
...
+    const chopLength = `:${this.port}`.length;
+    return this.address.slice(0, -chopLength);
+  }
+
+  get port() {
+    const addressParts = this.address.split(':');
+    return Number.parseInt(addressParts[addressParts.length - 1], 10);
```
emadum, line 145: I probably would have done `return Number.parseInt(this.address.split(':').pop(), 10);` here for brevity. Just wondering if you chose this slightly more verbose approach for performance reasons (I'm assuming a pop would be slightly slower than the way you've done it) or if it was just a stylistic choice.

src/sdam/monitor.ts
```diff
@@ -16,6 +16,7 @@ import {
   ServerHeartbeatSucceededEvent,
   ServerHeartbeatFailedEvent
 } from './events';
+import { Server } from './server';
```
mbroadst, line 19: actually now that I'm back here, don't we just need a `import type { Server }` here instead? I'm cautious about introducing unwanted cycles

nbbeeken: The `import type` syntax is a new insurance policy for what typescript already does, the generated code doesn't have an import cus this is only used as a type annotation. I think when we fix the prettier issue I think it was / eslint issue with this syntax I'd be interested to see if we can scan for these type only imports and make them explicit. I can change it to a type one now and do the eslint disable comment?

src/sdam/server_description.ts
nbbeeken: I'm just a bad javascript programmer who forgot pop() existed 😂 I can totally change it, that does read nicer

src/sdam/monitor.ts
nbbeeken: Found this flag: `importsNotUsedAsValues`  that can be set to `'error'`, that will block code from compiling if there's an import that is type only but isn't annotated as such, I'll give it a go and report back

nbbeeken: Yea so this flag works as described, I'm going to add to the prettier ticket that we turn the flag on there and change `import`s to `import type`s where necessary.

nbbeeken:
Not sure what happened to evg here, rerunning:
https://spruce.mongodb.com/version/5f11dfa57742ae66f7dca2be/tasks

src/sdam/monitor.ts
mbroadst: SGTM

nbbeeken:
The patch passed ✅  PTAL

PR #2437, opened 2020-07-10
test: use sinon mock timers for determinstic timer testing

mbroadst:
Using sinon's ability to mock the system clock, we can make testing the interruptable async timer more deterministc across platforms.

NODE-2684


mbroadst:
Here's a patch for the OSX build: https://evergreen.mongodb.com/version/5f085ab9d1fe07612405841f

PR #2442, opened 2020-07-13
refactor: merge `CommandOperation` and `CommandOperationV2`

mbroadst:
These last few operations built on top of `CommandOperation` could benefit from improvements made to the "execute with selection" model while also helping us migrate the codebase to a consistent operation layer.

NODE-2688


src/read_preference.ts
```diff
@@ -123,7 +123,7 @@ class ReadPreference {
    * @param {any} options The options passed into the method, potentially containing a read preference
    * @returns {(ReadPreference|null)} The resolved read preference
    */
-  static resolve(parent: any, options: any): ReadPreference | null {
+  static resolve(parent: any, options: any): ReadPreference {
```
reggi, line 126: Nice catch here, we might want to remove the JSDOC?

PR #2450, opened 2020-07-16
refactor: remove all internal usage of `executeCommand` helper

mbroadst:
This is a relatively large refactor that involves the following changes:

- remove ExecuteDbAdminCommandOperation
- remove the remaining legacy index operations, convert them into `CreateIndex`/`CreateIndexes`/`EnsureIndex` operations, and put all index-related operations into a single file
- convert find_and_modify to command operation, also merge all FindOneAnd* operations into a single file
- remove all internal usage of `executeCommand` helper and the helper itself

Note that a copy of `executeCommand` is perserved in the
FindAndModifyOperation because a larger refactor is needed to port
write commands to the operations layer

NODE-2714


src/operations/indexes.ts
```diff
@@ -1,5 +1,64 @@
-import { OperationBase } from './operation';
 import { indexInformation } from './common_functions';
+import { OperationBase, Aspect, defineAspects } from './operation';
```
mbroadst, line 2: This file looks like brand new code, but it's just copy pasta from all the other individual index related files. No functional changes were made

src/collection.ts
```diff
@@ -210,7 +218,7 @@ class Collection {
    * @memberof Collection#
    * @readonly
    */
-  get collectionName() {
+  get collectionName(): string {
```
nbbeeken, line 221: This is something we're going to encounter along our full transition to TS but if namespace.collection was correctly typed as a string then the return type annotation is not needed. But, understandably, we don't want the changes here to sprawl just to put type annotations in their ideal location. I took a quick look and it looks like as of yet there's not a linting rule to help us here, but something we can keep in mind / discuss best path

src/operations/count.ts
```diff
@@ -53,6 +53,51 @@ class CountOperation extends CommandOperation {
...
 
+/**
+ * Build the count command.
+ *
+ * @function
+ * @param {Collection|Cursor} collectionOrCursor an instance of a collection or cursor
+ * @param {any} query The query for the count.
+ * @param {any} [options] Optional settings. See Collection.prototype.count and Cursor.prototype.count for a list of options.
+ */
+function buildCountCommand(collectionOrCursor: any, query: any, options?: any) {
```
nbbeeken, line 64: Should we utilize `Collection | Cursor` as the first arg's type here? We could add a helper/util that looks something like:
```ts
function isCursor(c: Collection | Cursor): c is Cursor { /* check if cursor */ }
function isCollection(c: Collection | Cursor): c is Collection { /* check if collection */ }
```
https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/operations/find_and_modify.ts
```diff
@@ -117,4 +116,78 @@ class FindAndModifyOperation extends OperationBase {
...
+}
+
+class FindOneAndReplaceOperation extends FindAndModifyOperation {
+  constructor(collection: any, filter: any, replacement: any, options: any) {
+    // Final options
+    const finalOptions = Object.assign({}, options);
+    finalOptions.fields = options.projection;
+    finalOptions.update = true;
+    finalOptions.new = options.returnOriginal !== void 0 ? !options.returnOriginal : false;
+    finalOptions.upsert = options.upsert !== void 0 ? !!options.upsert : false;
```
nbbeeken, line 168: would it be equivalent here to use `undefined` as opposed to `void 0`? I think `undefined` is clearer

src/collection.ts
mbroadst: I think we should discuss this more at length with the whole team, but I think I'm currently of a mind that if we're adding type annotations, we should always annotate explicitly. Definitely in this case I was trying to just force the type somewhere, but in general I think we should have a rule

src/operations/count.ts
mbroadst: yes, I think we should do that as part of [NODE-2732](https://jira.mongodb.org/browse/NODE-2732)

src/operations/find_and_modify.ts
nbbeeken: Ok yea my mistake for not considering "refactor:" 😅 but thanks anyway for addressing my comments with ticket links, its good to connect it all

PR #2452, opened 2020-07-17
NODE-2680: Bump prettier version and modify config to reduce reformatting

nbbeeken:
I separated the config changes from the formatting changes for easier viewing, I prevented prettier from adding trailing commas. I can't prevent it from adding a space after `function` and that in turn made some lines that were over 100 characters reflow making the diff a bit larger in a few places. I also enabled `importsNotUsedAsValues` which revealed only two instances of importing types and I corrected them in, `src/bson.ts` and `src/operations/run_command.ts`

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -2,5 +2,6 @@
   "singleQuote": true,
   "tabWidth": 2,
   "printWidth": 100,
-  "arrowParens": "avoid"
+  "arrowParens": "avoid",
+  "trailingComma": "none"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -6871,9 +6871,9 @@
       "dev": true
     },
     "prettier": {
-      "version": "1.19.1",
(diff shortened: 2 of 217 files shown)


mbroadst:
holding off on this until the larger refactors land, as there is certain to be many conflicts and this one is automated

nbbeeken:
> I'm with Matt on holding off on this. For the record I'm not a big fan of prettier. I think we should avoid upgrading prettier and conform our eslint config to the abide by the old prettier rules if possible.

I hear you I tried looking into overriding the `function()` -> `function ()` but there doesn't seem to be a way to disable it, I found https://eslint.org/docs/rules/keyword-spacing this rule but I can't get it to override the prettier rules. Also prettier changed its reflow algo for chained statements: 
```
.a()
.b()
.c()
```
`.a().b().c()`
And that's something that's outside of eslint scope using rule controls
but in the end we can't use ts 3.8+ import syntaxes without the prettier bump, its not going to be back ported to less than 2.x

PR #2455, opened 2020-07-20
fix: IPv6 Address handling in SDAM monitoring [3.6]

nbbeeken:
NODE-2671

Port of #2435

### DIFF
--- a/lib/core/sdam/server_description.js
+++ b/lib/core/sdam/server_description.js
@@ -116,6 +116,16 @@ class ServerDescription {
     return WRITABLE_SERVER_TYPES.has(this.type);
   }
 
+  get host() {
+    const chopLength = `:${this.port}`.length;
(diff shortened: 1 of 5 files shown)


lib/core/sdam/server_description.js
```diff
@@ -116,6 +116,16 @@ class ServerDescription {
...
+  get host() {
+    const chopLength = `:${this.port}`.length;
+    return this.address.slice(0, -chopLength);
+  }
+
+  get port() {
+    const port = this.address.split(':').pop();
+    return port ? Number.parseInt(port, 10) : port;
+  }
+
```
reggi, line 128: So pretty 😹

mbroadst:
@nbbeeken hmm something is very off with the tests, I suspect the address isn't being represented correctly internally

PR #2467, opened 2020-07-21
NODE-2678/3.5/read-preference-from-options

reggi:


mbroadst:
@reggi is this ready for review? There haven't been any reviewers added

PR #2470, opened 2020-07-22
NODE-2696/options-parsing-foundation

reggi:
Added some additions to the [design doc](https://docs.google.com/document/d/1qd8J0-tP07u65Vc4XBt9v05LIgy04LD4bHzWazdancY/edit#heading=h.rgcvzgoigu2s) reflecting some changes to the design.


src/options/client_options.ts
```diff
@@ -0,0 +1,40 @@
+const url = require('url');
+const qs = require('querystring');
+import type { UriOptions, ClientOptions } from './types';
+import { CoerceCustom } from './coerce_custom';
+
+export class MongoClientOptions {
+  static parseConnectionString(connectionString: string) {
```
reggi, line 7: This is not finalized, needs errors and I will go over it again and add test specific to this function.

src/options/coerce_custom.ts
```diff
@@ -0,0 +1,239 @@
...
+  ReadConcernLevel,
+  AuthMechanism,
+  LogLevel,
+  UriOptions,
+  ClientOptions
+} from './types';
+import { Coerce as C, CoerceMatch, CoerceOptions } from './coerce';
+
+export class CoerceCustom {
+  // prettier-ignore
```
reggi, line 13: This may be a point of contention, but I'd prefer to ignore some of these blocks for our own sanity.

If not ok the alternative is this:

```ts
  static hedgeDef = (match: CoerceMatch) => ({
    ...match('enable', C.require(C.boolean))
  });

  static hedge = C.objectExact(CoerceCustom.hedgeDef);

  static readPreferenceDef = (match: CoerceMatch) => ({
    ...match('mode', C.default(C.enum({ ReadPreferenceMode }), 'primary')),
    ...match('maxStalenessSeconds', C.number),
    ...match('tags', C.default(C.tags, [])),
    ...match('hedge', C.objectExact())
  });

  static readPreference = C.warn(C.objectExact(CoerceCustom.readPreferenceDef));
```

PR #2474, opened 2020-07-23
NODE-2727: Complete migration of cmap module to TypeScript

nbbeeken:
Reduce `any` usage as much as possible, I only added a few precise interfaces while leaving more things to be `Document` which is only slightly more prescriptive than `any` giving us a basic dictionary type. 

- I start a small pattern here of adding a `types.d.ts` file to a module for shared interfaces, the .d.ts file will be omitted from the lib output which is nice for keeping our published code very javascript first.
- I created a type Callback in src/types.d.ts that gives us a very generic callback type but it can be parametrized to specify the return type `Callback<Buffer>` is `(e?: Error, r?: Buffer) => void`

Callback commentary:  
It has an issue with the second argument being optional even though this is common practice to make error and result mutually exclusive, since this is a runtime determination I don't believe we can find a type solution to this, something I considered is we could use a runtime assertion function that is disabled in prod:
```ts
function assert(value: any):  asserts value { /* assertion that only runs in dev */ }
```
Then our pattern could look like:
```ts
command('blah', (err, res) => {
  if(err) return err;
  assert(res);
  // TS will now say res exists for the rest of this scope
})
```

NODE-2727


src/cmap/events.ts
```diff
@@ -158,80 +133,110 @@ const CMAP_EVENT_NAMES = [
...
+  connectionId?: string | number;
 
   /**
    * Create a started event
    *
    * @param {Pool} pool the pool that originated the command
    * @param {object} command the command
    */
-  constructor(pool: any, command: any) {
+  constructor(pool: Connection | ConnectionPool, command: CommandTypes) {
```
mbroadst, line 151: `Connection | ConnectionPool` looks wrong

src/cmap/events.ts
```diff
@@ -249,14 +254,14 @@ const SENSITIVE_COMMANDS = new Set([
...
-const maybeRedact = (commandName: any, result: any) =>
+const extractCommandName = (commandDoc: Document) => Object.keys(commandDoc)[0];
+const namespace = (command: CommandTypes) => command.ns;
+const databaseName = (command: CommandTypes) => command.ns.split('.')[0];
+const collectionName = (command: CommandTypes) => command.ns.split('.')[1];
+const maybeRedact = (commandName: string, result: CommandResult | Error | Document) =>
   SENSITIVE_COMMANDS.has(commandName) ? {} : result;
 
-const LEGACY_FIND_QUERY_MAP: any = {
+const LEGACY_FIND_QUERY_MAP = {
```
mbroadst, line 264:
```
const LEGACY_FIND_QUERY_MAP: { [key: string]: string } = {
...
}
```

mbroadst: I'm actually confused why this one needs a type?

src/cmap/events.ts
```diff
@@ -312,41 +312,48 @@ const extractCommand = (command: any) => {
   }
 
   if (command.query && command.query.$query) {
-    let result: any;
-    if (command.ns === 'admin.$cmd') {
-      // upconvert legacy command
+    let result: Document;
+    let commandObject: any = command;
```
mbroadst, line 316: hmm any way to avoid the `any` here?

src/cmap/commands.ts
```diff
@@ -21,32 +24,33 @@ const QUERY_FAILURE = 2;
 const SHARD_CONFIG_STALE = 4;
 const AWAIT_CAPABLE = 8;
 
+export type CommandTypes = Query | Msg | GetMore | KillCursor;
```
mbroadst, line 27: prefer the singular `CommandType` since its a union

src/cmap/commands.ts
```diff
@@ -21,32 +24,33 @@ const QUERY_FAILURE = 2;
...
-  numberToSkip: any;
-  numberToReturn: any;
-  returnFieldSelector: any;
-  requestId: any;
+  numberToSkip: number;
+  numberToReturn: number;
+  returnFieldSelector?: any;
+  requestId: number;
   pre32Limit: any;
   serializeFunctions: any;
```
mbroadst, line 40: `boolean`

src/cmap/commands.ts
```diff
@@ -21,32 +24,33 @@ const QUERY_FAILURE = 2;
...
   query: any;
-  numberToSkip: any;
-  numberToReturn: any;
-  returnFieldSelector: any;
-  requestId: any;
+  numberToSkip: number;
+  numberToReturn: number;
+  returnFieldSelector?: any;
+  requestId: number;
   pre32Limit: any;
```
mbroadst, line 39: `number`

src/cmap/commands.ts
```diff
@@ -21,32 +24,33 @@ const QUERY_FAILURE = 2;
...
+export type CommandTypes = Query | Msg | GetMore | KillCursor;
+
 /**************************************************************
  * QUERY
  **************************************************************/
-class Query {
-  ns: any;
+export class Query {
+  ns: string;
   query: any;
```
mbroadst, line 34: `Document`

src/cmap/commands.ts
```diff
@@ -21,32 +24,33 @@ const QUERY_FAILURE = 2;
...
-  numberToSkip: any;
-  numberToReturn: any;
-  returnFieldSelector: any;
-  requestId: any;
+  numberToSkip: number;
+  numberToReturn: number;
+  returnFieldSelector?: any;
```
mbroadst, line 37: I think we can just get rid of this, but I think its a `string`

src/cmap/commands.ts
```diff
@@ -492,17 +502,12 @@ class Response {
...
-      typeof options.promoteLongs === 'boolean' ? options.promoteLongs : this.opts.promoteLongs;
-    var promoteValues =
-      typeof options.promoteValues === 'boolean' ? options.promoteValues : this.opts.promoteValues;
-    var promoteBuffers =
-      typeof options.promoteBuffers === 'boolean'
-        ? options.promoteBuffers
-        : this.opts.promoteBuffers;
-    var bsonSize;
+    let raw = options.raw || false;
+    let documentsReturnedIn = options.documentsReturnedIn || null;
```
mbroadst, line 506: this is a `string`

src/cmap/commands.ts
```diff
@@ -585,20 +590,20 @@ const OPTS_CHECKSUM_PRESENT = 1;
...
+  requestId: number;
+  serializeFunctions: boolean;
+  ignoreUndefined: boolean;
+  checkKeys: boolean;
+  maxBsonSize: number;
+  checksumPresent: boolean;
+  moreToCome: boolean;
+  exhaustAllowed: boolean;
+
+  constructor(ns: string, command: Document, options: any) {
```
mbroadst, line 606: `MsgOptions`?

src/cmap/commands.ts
```diff
@@ -692,27 +694,27 @@ class Msg {
...
+  checksumPresent: boolean;
+  moreToCome: boolean;
+  exhaustAllowed: boolean;
+  promoteLongs: boolean;
+  promoteValues: boolean;
+  promoteBuffers: boolean;
+  documents: Document[];
+  index?: number;
+
+  constructor(message: any, msgHeader: MessageHeader, msgBody: any, opts: any) {
```
mbroadst, line 717: `msgBody` is a `Buffer` and `opts` should be `BinMsgOptions`

src/cmap/connection_pool.ts
```diff
@@ -83,31 +84,51 @@ const VALID_POOL_OPTIONS = new Set([
...
 
-  return Object.freeze(Object.assign({}, defaults, newOptions));
+export interface ConnectionPoolOptions extends ValidPoolOptions {
+  connectionType?: typeof Connection;
+  /** The host to connect to */
+  host: string;
+  /** The port to connect to */
+  port: number;
+  /** The maximum number of connections that may be associated with a pool at a given time. This includes in use and available connections. */
+  maxPoolSize?: number;
```
mbroadst, line 115: this has a default value, should always be set. So do most of these

src/cmap/connection_pool.ts
```diff
@@ -220,7 +240,8 @@ class ConnectionPool extends EventEmitter {
     }
 
     // add this request to the wait queue
-    const waitQueueMember: any = { callback } as any;
+    const waitQueueMember = {} as WaitQueueMember;
```
mbroadst, line 243:
```
const waitQueueMember: WaitQueueMember = { callback };
```
?

src/cmap/connection_pool.ts
```diff
@@ -300,35 +323,37 @@ class ConnectionPool extends EventEmitter {
...
+        if (!waitQueueMember[kCancelled]) {
+          waitQueueMember.callback(new MongoError('connection pool closed'));
+        }
       }
     }
 
     // clear the min pool size timer
     if (this[kMinPoolSizeTimer]) {
-      clearTimeout(this[kMinPoolSizeTimer]);
+      clearTimeout(this[kMinPoolSizeTimer]!); //
```
mbroadst, line 336: some cruft left here

src/cmap/connection_pool.ts
```diff
@@ -343,10 +368,10 @@ class ConnectionPool extends EventEmitter {
    * @param {Function} callback The original callback
    * @returns {void}
    */
-  withConnection(fn: any, callback: Function): void {
-    this.checkOut((err?: any, conn?: any) => {
+  withConnection(fn: WithConnectionCallback, callback: Callback<unknown>): void {
```
mbroadst, line 371: why `unknown` for the callback?

src/cmap/connection_pool.ts
```diff
@@ -394,7 +419,7 @@ function createConnection(pool: any, callback?: Function) {
   );
 
   pool[kPermits]--;
-  connect(connectOptions, pool[kCancellationToken], (err?: any, connection?: any) => {
+  connect(connectOptions as any, pool[kCancellationToken], (err, conn) => {
```
mbroadst, line 422: shouldn't need to cast to `any` here

src/cmap/connection_pool.ts
```diff
@@ -446,14 +472,14 @@ function destroyConnection(pool: any, connection: any, reason: any) {
...
-function processWaitQueue(pool: any) {
+function processWaitQueue(pool: ConnectionPool) {
   if (pool.closed) {
     return;
   }
 
   while (pool.waitQueueSize) {
     const waitQueueMember = pool[kWaitQueue].peekFront();
-    if (waitQueueMember[kCancelled]) {
+    if (waitQueueMember && waitQueueMember[kCancelled]) {
```
mbroadst, line 482: I think this leaves a subtle bug in place, what if `waitQueueMember == null`?

src/cmap/connection_pool.ts
```diff
@@ -462,27 +488,27 @@ function processWaitQueue(pool: any) {
       break;
     }
 
-    const connection = pool[kConnections].shift();
+    const connection = pool[kConnections].shift()!;
```
mbroadst, line 491: this looks like a potential bug

src/cmap/connection_pool.ts
```diff
@@ -462,27 +488,27 @@ function processWaitQueue(pool: any) {
...
 
-  const maxPoolSize = pool.options.maxPoolSize;
+  const maxPoolSize = pool.options.maxPoolSize!;
   if (pool.waitQueueSize && (maxPoolSize <= 0 || pool.totalConnectionCount < maxPoolSize)) {
-    createConnection(pool, (err?: any, connection?: any) => {
+    createConnection(pool, (err, connection) => {
       const waitQueueMember = pool[kWaitQueue].shift();
       if (waitQueueMember == null) {
-        if (err == null) {
+        if (err == null && connection) {
```
mbroadst, line 511: looks like a bug may still exist around this

src/cmap/connection_pool.ts
```diff
@@ -494,13 +520,16 @@ function processWaitQueue(pool: any) {
       }
 
       if (err) {
-        pool.emit('connectionCheckOutFailed', new ConnectionCheckOutFailedEvent(pool, err));
-      } else {
+        pool.emit(
+          'connectionCheckOutFailed',
+          new ConnectionCheckOutFailedEvent(pool, err as MongoError)
```
mbroadst, line 525: `ConnectionCheckOutFailedEvent` should accept whatever the union for our `ErrorLike` becomes (or whatever its called), so we can avoid casting here

src/cmap/connection_pool.ts
```diff
@@ -494,13 +520,16 @@ function processWaitQueue(pool: any) {
...
+        );
+      } else if (connection) {
         pool.emit('connectionCheckedOut', new ConnectionCheckedOutEvent(pool, connection));
       }
 
-      clearTimeout(waitQueueMember.timer);
-      waitQueueMember.callback(err, connection);
+      clearTimeout(waitQueueMember.timer!);
+      waitQueueMember.callback(err as MongoError, connection);
```
mbroadst, line 532: same comment applies here as above

src/cmap/connection_pool.ts
```diff
@@ -463,26 +485,29 @@ function processWaitQueue(pool: any) {
...
-      clearTimeout(waitQueueMember.timer);
-      pool[kWaitQueue].shift();
-      waitQueueMember.callback(undefined, connection);
-      return;
+    if (connection) {
+      const isStale = connectionIsStale(pool, connection);
+      const isIdle = connectionIsIdle(pool, connection);
+      if (!isStale && !isIdle && !connection.closed) {
+        pool.emit('connectionCheckedOut', new ConnectionCheckedOutEvent(pool, connection));
+        if (waitQueueMember && waitQueueMember.timer) {
```
mbroadst, line 493: This is a bug, you should only be gating the call to `clearTimeout` on this:
```
if (waitQueueMember && waitQueueMember.timer) {
  clearTimeout(waitQueueMember.timer);
}

pool[kWaitQueue].shift();
waitQueueMember.callback(undefined, connection);
```

Additionally, I think both with `connection` and `waitQueueMember` we should take a "bail-early" strategy:
```
while (pool.waitQueueSize) {
  const waitQueueMember = pool[kWaitQueue].peekFront();
  if (waitQueueMember == null) {
    throw new Error('should never happen!');
  }
  
  ....
  const connection = pool[kConnections].shift();
  if (connection == null) {
    throw new Error('should never happen!');
  }
  ...
```
I'm throwing errors here for effect, we should probably not do that. Related question: these checks are redundant, at least in one case:
```
while (pool.waitQueueSize) {
  const waitQueueMember = pool[kWaitQueue].peekFront();
```
that guarantees that we always have something on the wait queue before peeking, and we're only peeking because we might want to leave it in place. Perhaps there are some cases where we would be comfortable using the bang (`!`).

nbbeeken: Nice find, thanks for taking a look, I agree it is redundant, if we do start using the assert `!` maybe we should enable the linter to warn on them so we have to comment an eslint-disable line to ignore it? or is that tooling overkill?

src/cmap/connection.ts
```diff
@@ -264,7 +320,7 @@ function messageHandler(conn: any) {
       }
 
       if (document.$clusterTime) {
-        conn[kClusterTime] = document.$clusterTime;
+        conn[kClusterTime] = document.$clusterTime as Document;
```
mbroadst, line 323: I wonder if you'd have to do this if you made it `const document: Document = message.documents[0]`. Also, maybe `message.documents` needs to be typed as `Document[]`

src/cmap/connection.ts
```diff
@@ -295,7 +351,7 @@ function messageHandler(conn: any) {
   };
 }
 
-function streamIdentifier(stream: any) {
+function streamIdentifier(stream: Socket) {
```
mbroadst, line 354: This relates to an earlier comment, but I'm hoping we can type this to `Stream`

src/cmap/connection.ts
```diff
@@ -304,28 +360,34 @@ function streamIdentifier(stream: any) {
...
 
   options = options || {};
-  const operationDescription = {
+  const operationDescription: OperationDescription = {
     requestId: command.requestId,
     cb: callback,
     session: options.session,
-    fullResult: typeof options.fullResult === 'boolean' ? options.fullResult : false,
-    noResponse: typeof options.noResponse === 'boolean' ? options.noResponse : false,
+    fullResult: options.fullResult ?? false,
```
mbroadst, line 379: <generic message about loss of runtime type checking>

src/cmap/connection_pool.ts
```diff
@@ -29,7 +30,7 @@ const kCancellationToken = Symbol('cancellationToken');
 const kWaitQueue = Symbol('waitQueue');
 const kCancelled = Symbol('cancelled');
 
-const VALID_POOL_OPTIONS = new Set([
+const validPoolOptionsList = [
```
mbroadst, line 33: => `VALID_POOL_OPTION_NAMES`

src/cmap/connection_pool.ts
```diff
@@ -210,7 +223,7 @@ class ConnectionPool extends EventEmitter {
    *
    * @param {ConnectionPool~checkOutCallback} callback
    */
-  checkOut(callback: Function) {
+  checkOut(callback: CheckoutCallback): void {
```
mbroadst, line 226: hm why isn't this just `Callback<Connection>`?

src/cmap/connection_pool.ts
```diff
@@ -220,7 +233,7 @@ class ConnectionPool extends EventEmitter {
     }
 
     // add this request to the wait queue
-    const waitQueueMember: any = { callback } as any;
+    const waitQueueMember: WaitQueueMember = { callback } as WaitQueueMember;
```
mbroadst, line 236: do you need the cast here?

src/cmap/connection_pool.ts
```diff
@@ -300,35 +316,38 @@ class ConnectionPool extends EventEmitter {
     // drain the wait queue
     while (this.waitQueueSize) {
       const waitQueueMember = this[kWaitQueue].pop();
-      clearTimeout(waitQueueMember.timer);
-      if (!waitQueueMember[kCancelled]) {
-        waitQueueMember.callback(new MongoError('connection pool closed'));
+      if (waitQueueMember && waitQueueMember.timer) {
```
mbroadst, line 319: you have the same bug you used to have below here, you're gating the entire logic here on the existence of the `timer`, but you only want to gate the `clearTimeout` call

src/cmap/connect.ts
```diff
@@ -1,28 +1,50 @@
...
+import type { EventEmitter } from 'events';
+
+import type { Socket, SocketConnectOpts } from 'net';
+import type { TLSSocket, ConnectionOptions as TLSConnectionOpts } from 'tls';
+
+type UniversalSocket = Socket | TLSSocket;
+
+export function connect(
+  options: MongoDBConnectionOptions,
+  callback: Callback<Connection | UniversalSocket>
```
mbroadst, line 24: `callback` should only ever return a `Connection`. Can you show me where it returns a socket directly? That looks like a bug!

src/cmap/connection_pool.ts
```diff
@@ -404,6 +423,8 @@ function createConnection(pool: any, callback?: Function) {
 
       return;
     }
+    // conn is only left as a plain socket if there was an error
+    const connection = conn as Connection;
```
mbroadst, line 427: As noted above, `conn` should only ever be a `Connection`

src/cmap/connection_pool.ts
```diff
@@ -515,6 +548,11 @@ function processWaitQueue(pool: any) {
  * @param {Connection} connection The managed connection which was checked out of the pool.
  * @param {Function} callback A function to call back after connection management is complete
  */
+type WithConnectionCallback = (
+  error?: MongoError | WaitQueueTimeoutError | null,
```
mbroadst, line 552: remove `null`, this is already an optional field

src/cmap/connection_pool.ts
```diff
@@ -523,6 +561,7 @@ function processWaitQueue(pool: any) {
  * @param {MongoError} error An error instance representing the error during checkout
  * @param {Connection} connection A connection from the pool
  */
+type CheckoutCallback = CallbackWithType<WaitQueueTimeoutError | MongoError, Connection>;
```
mbroadst, line 564: `WaitQueueTimeoutError` should be a `MongoError`, I think we can replace this with `Callback<Connection>`

src/cmap/events.ts
```diff
@@ -312,41 +311,59 @@ const extractCommand = (command: any) => {
   }
 
   if (command.query && command.query.$query) {
-    let result: any;
-    if (command.ns === 'admin.$cmd') {
-      // upconvert legacy command
+    let result: Document;
+    const commandObject: any = command;
```
mbroadst, line 315: :thinking: it looks to me like more refactoring is needed to ensure proper typing, otherwise we might as well leave `command: any` above and be done with it today

src/cursor/types.d.ts
```diff
@@ -0,0 +1,3 @@
+export interface CursorState {
```
mbroadst, line 1: Let's just put this in `cursor.ts` or `core_cursor.ts`. I'd prefer to only have a top-level `types.d.ts` if needed

src/cmap/connect.ts
nbbeeken: A few lines down, where we call makeConnection, my mistake for not reading the comment "in the error case, `socket` is the originating error event name" but as I look through I think this is outdated, none of the error cases send through a second argument.

src/cmap/auth/gssapi.ts
```diff
@@ -1,82 +1,108 @@
...
+    const { host, port } = authContext.options;
+    if (!host || !port) {
+      return callback(new MongoError('Connection must specify host and port.'));
+    }
 
-    authProcess.init(username, password, (err: any) => {
-      if (err) return callback(err, false);
-      authProcess.transition('', (err?: any, payload?: any) => {
-        if (err) return callback(err, false);
+    import('kerberos')
```
mbroadst, line 19: should we add this to the `prepare` step instead? I'm not sure this pattern is best for us, but don't feel very strongly about it:
 - its importing the module each time you authenticate
 - it uses promises inside the library, which we generally try not to do
 - nit: it adds an extra level of indentation to this very long method, which is distracting for readability

Is the old pattern in `deps` undesirable for any particular technical reason?

src/cmap/auth/scram.ts
```diff
@@ -1,37 +1,49 @@
...
+import type { HandshakeDocument } from '../connect';
 
-let saslprep: any;
+type CryptoMethod = 'sha1' | 'sha256';
+
+interface SaslPrepModule {
+  (password: string): string;
+}
+
+let saslprep: SaslPrepModule | null;
```
mbroadst, line 15: this should be moved to `deps.ts`

src/cmap/connection.ts
```diff
@@ -304,28 +361,34 @@ function streamIdentifier(stream: any) {
...
 
   options = options || {};
-  const operationDescription = {
+  const operationDescription: OperationDescription = {
     requestId: command.requestId,
     cb: callback,
     session: options.session,
-    fullResult: typeof options.fullResult === 'boolean' ? options.fullResult : false,
-    noResponse: typeof options.noResponse === 'boolean' ? options.noResponse : false,
+    fullResult: 'boolean' === typeof options.fullResult ? options.fullResult : false,
```
mbroadst, line 380: This looks like just a "whitespace" change, can we make these changes in subsequent PRs if we all feel like the pattern helps readability?

src/cmap/connect.ts
```diff
@@ -1,28 +1,50 @@
...
+): void;
+export function connect(
+  options: MongoDBConnectionOptions,
+  _cancellationToken: EventEmitter | Callback<Connection | UniversalSocket>,
+  _callback?: Callback<Connection | UniversalSocket>
+): void {
+  let cancellationToken = _cancellationToken as EventEmitter | undefined;
+  const callback = (_callback ?? _cancellationToken) as Callback<Connection | UniversalSocket>;
+  if ('function' === typeof cancellationToken) {
```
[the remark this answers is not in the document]
nbbeeken: I wanted the function signatures to have cancellationToken to be either required or omitted all together, it gives you the two distinct call signatures where you can drop the token. Here I cast to a union of EventEmitter | undefined because in the function if statement I overwrite the cancellationToken with undefined. It's all meant to resolve the complications with typing and function overloads by hiding it in the function instead of having weird signatures.

Although maybe the signatures are still not entirely accurate then here?

src/cmap/connection.ts
```diff
@@ -24,30 +32,60 @@ const kDescription = Symbol('description');
...
-  [kLastUseTime]: any;
-  [kAutoEncrypter]: any;
-  [kQueue]: any;
-  [kMessageStream]: any;
-  [kStream]: any;
-  [kIsMaster]: any;
-  [kClusterTime]: any;
-
-  constructor(stream: any, options: any) {
+export interface MongoDBConnectionOptions
```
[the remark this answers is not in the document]
nbbeeken: I thought replied to this but I think it got lost, 
'tls' exports a ConnectionOptions type, which I do rename, you can see I do `import type { ConnectionOptions as TLSConnectionOptions } from 'tls';` but I wonder if we ever export this type we'd want to avoid that clash, right?

src/cmap/events.ts
nbbeeken: Fixed this, found a typo!
`returnFieldsSelector` is mapped to `projection` should be `returnFieldSelector` right?

src/cmap/stream_description.ts
```diff
@@ -1,26 +1,34 @@
...
+] as const;
+
+export interface StreamDescriptionOptions {
+  compression: {
+    compressors: string[];
+  };
+}
+
+export class StreamDescription {
+  address: string;
```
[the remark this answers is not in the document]
nbbeeken: We can't do that here because `receiveResponse` writes to every field after class construction

src/cmap/stream_description.ts
```diff
@@ -34,20 +42,22 @@ class StreamDescription {
...
     this.type = parseServerType(response);
-    RESPONSE_FIELDS.forEach((field: any) => {
+    RESPONSE_FIELDS.forEach(field => {
       if (typeof response[field] !== 'undefined') {
-        (this as any)[field] = response[field];
+        this[field] = response[field];
+      }
+
+      // testing case
+      if ('__nodejs_mock_server__' in response) {
```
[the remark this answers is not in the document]
nbbeeken: I removed the casting but it is still in a separate statement because all the other keys in that list are of type number, but this is a boolean so you get a union type for `response[field]` that doesn't assign to `this[field]`

src/cmap/auth/gssapi.ts
nbbeeken: > should we add this to the prepare step instead?
Which step
> Is the old pattern in deps undesirable for any particular technical reason?
It's just funky to type as you saw with BSON, I suppose also I was bending over backwards to avoid the type assertion `!` that's not a good reason to start using promises here I suppose 😅 
The module and code would be cached in v8 but I understand why we wouldn't want to lean on that.

> it adds an extra level of indentation to this very long method, which is distracting for readability
The irony of a promise based approach causing this.... 😆 

I'll revert

src/deps.ts
```diff
@@ -1,36 +1,13 @@
...
-} catch (_) {} // eslint-disable-line
-
-export { Kerberos, Snappy, kModuleError };
+export function optionalRequire<T>(moduleName: string): T | OptionalModule {
+  try {
+    return require(moduleName);
+  } catch {
+    return { kModuleError: new MongoError(`Optional Module ${moduleName} not found`) };
+  }
+}
```
nbbeeken, line 13: This is my typescript take on the optional dependency, you can jump to `gssapi.ts` to see how it looks when called. I have to use a string for the prop that holds the error that might be there since since

```ts
if ('kModuleError' in Kerberos) {
  // Kerberos is `OptionalModule` type
}
// Kerberos is `typeof import('kerberos')` type
```
the `in` type guard requires a string literal on the left hand side. I think this will work for us, we were using a Symbol for kModuleError for maximum avoiding of collision, I think with TS in play if any of the modules got updated to have a property `'kModuleError'` then it can't have the type MongoError so we would get conflicts and be able to dodge bugs there.

src/cmap/auth/mongodb_aws.ts
```diff
@@ -2,44 +2,44 @@ import http = require('http');
...
+import type { Callback, BSONSerializeOptions } from '../../types';
 
-let aws4: any;
-try {
-  aws4 = require('aws4');
-} catch (e) {
-  // don't do anything;
-}
+import { optionalRequire } from '../../deps';
+const aws4 = optionalRequire<typeof import('aws4')>('aws4');
```
mbroadst, line 12: if `aws4` is not installed, what does `typeof import('aws4')` do?

src/cmap/auth/mongodb_aws.ts
```diff
@@ -132,8 +136,14 @@ class MongoDBAWS extends AuthProvider {
...
 
-function makeTempCredentials(credentials: any, callback: Function) {
-  function done(creds: any) {
+interface TemporaryCredentials {
+  AccessKeyId?: string;
+  SecretAccessKey?: string;
+  Token?: string;
+}
+
+function makeTempCredentials(credentials: MongoCredentials, callback: Callback) {
```
mbroadst, line 145: `Callback<MongoCredentials>`

src/cmap/auth/mongodb_aws.ts
```diff
@@ -132,8 +136,14 @@ class MongoDBAWS extends AuthProvider {
   }
 }
 
-function makeTempCredentials(credentials: any, callback: Function) {
-  function done(creds: any) {
+interface TemporaryCredentials {
```
mbroadst, line 139: fwiw I think these are really `AWSCredentials`

src/cmap/auth/mongodb_aws.ts
```diff
@@ -207,12 +217,29 @@ function deriveRegion(host: any) {
...
 }
 
-function request(uri: any, options: any, callback?: Function) {
-  if (typeof options === 'function') {
-    callback = options;
+interface RequestOptions {
+  json?: boolean;
+  method?: string;
+  timeout?: number;
+  headers?: {
```
mbroadst, line 224: nit: given this is a pretty generic HTTP client, headers would likely be `{ [key: string]: any }`

src/cmap/auth/scram.ts
```diff
@@ -109,13 +122,19 @@ function executeScram(cryptoMethod: any, authContext: any, callback: Function) {
...
-  callback: Function
+  cryptoMethod: CryptoMethod,
+  response: Document,
+  authContext: AuthContext,
+  callback: Callback
 ) {
   const connection = authContext.connection;
   const credentials = authContext.credentials;
+  if (!credentials) {
+    return callback(new MongoError('AuthContext must provide credentials.'));
```
mbroadst, line 133: This made me wonder: would we ever be executing an auth mechanism if credentials were not provided? We only call `auth` in `connect` if credentials are detected, so I think it should not be an optional property on `AuthContext`

src/types.d.ts
```diff
@@ -0,0 +1,85 @@
+import type { MongoError } from './error';
+import type { SerializeOptions } from 'bson';
+import type { MongoClient } from '.';
+
+export type UniversalError = MongoError | Error;
```
mbroadst, line 5:
```suggestion
export type AnyError = MongoError | Error;
```
nitpick, but I think it's more "TypeScripty"

src/cmap/commands.ts
```diff
@@ -72,11 +98,12 @@ class Query {
     // Serialization option
     this.serializeFunctions =
       typeof options.serializeFunctions === 'boolean' ? options.serializeFunctions : false;
+    'boolean' === typeof options.serializeFunctions ? options.serializeFunctions : false;
```
mbroadst, line 101: this looks like cruft

src/cmap/commands.ts
```diff
@@ -585,20 +617,31 @@ const OPTS_CHECKSUM_PRESENT = 1;
...
-  serializeFunctions: any;
-  ignoreUndefined: any;
-  checkKeys: any;
-  maxBsonSize: any;
-  checksumPresent: any;
-  moreToCome: any;
-  exhaustAllowed: any;
-
-  constructor(ns: any, command: any, options: any) {
+export interface MsgOptions {
```
mbroadst, line 620:
```suggestion
export interface OpMsgOptions {
```

src/cmap/commands.ts
```diff
@@ -421,40 +446,52 @@ class KillCursor {
...
-class Response {
+export interface MessageHeader {
+  length: number;
+  requestId: number;
+  responseTo: number;
+  opCode: number;
+  fromCompressed?: boolean;
+}
+
+export interface ResponseOptions extends BSONSerializeOptions {
```
mbroadst, line 461:
```suggestion
export interface OpResponseOptions extends BSONSerializeOptions {
```

src/cmap/commands.ts
```diff
@@ -781,11 +810,11 @@ class BinMsg {
     }
 
     if (this.documents.length === 1 && documentsReturnedIn != null && raw) {
-      const fieldsAsRaw: any = {};
+      const fieldsAsRaw: Document = {};
       fieldsAsRaw[documentsReturnedIn] = true;
       _options.fieldsAsRaw = fieldsAsRaw;
 
-      const doc = BSON.deserialize(this.documents[0], _options);
+      const doc = BSON.deserialize((this.documents[0] as unknown) as Buffer, _options);
```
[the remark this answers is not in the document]
mbroadst: @nbbeeken still looking for an answer on this one

src/cmap/connect.ts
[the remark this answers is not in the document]
mbroadst: I checked the emitted output and it looks fine, but it does look strange to me. It looks like you cast it to `EventEmitter | undefined` above, if that were the case this could never be a `function`. Is there a way to rewrite this so the validation is a little more obvious?

src/cmap/connection.ts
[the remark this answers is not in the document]
mbroadst: hm I don't know if we'd ever want to export this type. I'd prefer to cross that bridge when we come to it, but for now `MongoDBConnectionOptions` sticks out like a sore thumb since nothing else is named with the same prefix

src/cmap/message_stream.ts
```diff
@@ -1,55 +1,91 @@
...
 const kDefaultMaxBsonMessageSize = 1024 * 1024 * 16 * 4;
 const kBuffer = Symbol('buffer');
 
+interface MessageStreamOptions extends DuplexOptions {
+  maxBsonMessageSize?: number;
+}
+
+export interface OperationDescription {
+  started: number;
+  cb: Callback<CommandResult | undefined>;
```
mbroadst, line 36:
```suggestion
  cb: Callback<CommandResult>;
```

`undefined` is unnecessary since `Callback` has an optional for the result type

src/cmap/message_stream.ts
```diff
@@ -1,55 +1,91 @@
...
+  command: boolean;
+  documentsReturnedIn?: string;
+  fullResult: boolean;
+  noResponse: boolean;
+  promoteBuffers: boolean;
+  promoteLongs: boolean;
+  promoteValues: boolean;
```
mbroadst, line 43: these are `BSONSerializationOptions` right?

src/cmap/stream_description.ts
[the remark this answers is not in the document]
mbroadst: :scream_cat:

src/cmap/stream_description.ts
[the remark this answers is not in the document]
mbroadst: cool, hopefully we can get rid of that field completely in the future

src/deps.ts
mbroadst: I think the approach might work, but I'm concerned about what happens at the call site when you do something like `typeof import('kerberos')` and `kerberos` is not installed. Can we maybe leave these changes out of this PR and try to address them by themselves in a follow-on PR?

src/cmap/auth/gssapi.ts
mbroadst: yeah :) this is a tough one! As suggested below, I think we should tackle this as its own piece of work so I created [NODE-2740](https://jira.mongodb.org/browse/NODE-2740)

src/cmap/events.ts
mbroadst: nice! yes, that's correct.

src/cmap/auth/scram.ts
```diff
@@ -91,14 +99,14 @@ function makeFirstMessage(cryptoMethod: any, credentials: any, nonce: any) {
   };
 }
 
-function executeScram(cryptoMethod: any, authContext: any, callback: Function) {
+function executeScram(cryptoMethod: CryptoMethods, authContext: AuthContext, callback: Callback) {
   const connection = authContext.connection;
   const credentials = authContext.credentials;
-  const nonce = authContext.nonce;
+  const nonce = authContext.nonce!;
```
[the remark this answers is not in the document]
mbroadst: I think the if statement is fine for now!

src/cmap/wire_protocol/write_command.ts
```diff
@@ -1,16 +1,38 @@
...
+  caseLevel: boolean;
+  caseFirst: string;
+  strength: number;
+  numericOrdering: boolean;
+  alternate: string;
+  maxVariable: string;
+  backwards: boolean;
+}
+
+export interface WriteCommandOptions extends BSONSerializeOptions {
```
mbroadst, line 20: does it also `extend CommandOptions`?

src/cmap/connection_pool.ts
```diff
@@ -446,13 +467,17 @@ function destroyConnection(pool: any, connection: any, reason: any) {
...
 
-function processWaitQueue(pool: any) {
+function processWaitQueue(pool: ConnectionPool) {
   if (pool.closed) {
     return;
   }
 
   while (pool.waitQueueSize) {
     const waitQueueMember = pool[kWaitQueue].peekFront();
+    if (!waitQueueMember) {
```
mbroadst, line 477: If we ever run into this case this would lead to an infinite loop, instead I think you want to `shift` the value off and `continue`

src/cmap/connection_pool.ts
```diff
@@ -463,14 +488,20 @@ function processWaitQueue(pool: any) {
     }
 
     const connection = pool[kConnections].shift();
+    if (!connection) {
```
mbroadst, line 491: similar bug here, but more subtle. If there is no connection then we want to `break` and process the wait queue later when the driver determines there are more connections

src/cmap/connect.ts
```diff
@@ -1,37 +1,58 @@
...
-  makeConnection(family, options, cancellationToken, (err?: any, socket?: any) => {
-    if (err) {
-      callback(err, socket); // in the error case, `socket` is the originating error event name
+  const ConnectionType: typeof Connection =
+    options && options.connectionType ? options.connectionType : Connection;
+  const family = options.family !== undefined ? options.family : 0;
+
+  makeConnection(family, options, cancellationToken, (err, socket) => {
+    if (err || !socket) {
+      callback(err);
```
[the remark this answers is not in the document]
mbroadst: I missed your comment on this earlier, you're right this was likely vestigial from when `connect` was being used by the legacy and CMAP connection pool. Resolving

src/cmap/commands.ts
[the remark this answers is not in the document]
nbbeeken: Whoops missed this, I made documents have type `(Buffer | Document)[]` because it is either depending on the raw flag, it requires casting still but cleaner than this at least. I could add runtime type checks? but there's not really a clear path to how to error, since they should always be translated to Docs when we access these properties as such.

src/cmap/auth/mongodb_aws.ts
nbbeeken: Nothing, its only a type annotation so that import is actually just referencing `@types/aws4` regardless of the actual aws4 being installed.

src/cmap/auth/scram.ts
nbbeeken: This one is tough because of the way we set up our speculative auth, the prepare side of auth takes an AuthContext since it will need to tack on properties that get started in that phase, but at that time the credentials don't have to be specified so I made credentials on AuthContext optional which means threading through this if statement to each authenticator. The alternative is an if statement over in connect.ts that skips the prepare call, but code structure being the way it is, callbacks and all, I think I'd have to do a larger refactor to break out the callbacks into named functions for reusability or big copy pastas.

I could missing something here though, does that logic track?

mbroadst:
:tada: :taco: :tada:

PR #2476, opened 2020-07-24
docs: reflect possible Db options for MongoClient db method

KonanMentor:
Copied the options from `src/db.ts`, Db constructor

## Description

Reflect possible Db options for MongoClient db method

Copied the options from `src/db.ts`, `Db` constructor


mbroadst:
Hi @KonanMentor , thanks very much for your contribution. We're in the middle of adding high quality types to the driver, which will result in the type of documentation you're looking to add here. You can follow along [here](https://jira.mongodb.org/browse/NODE-1445) for progress. I'm going to close this PR now, but please take a look at the changes once that epic is complete and let us know if you think we can continue to improve the documentation.

PR #2478, opened 2020-07-27
refactor: write operations use execute operation

mbroadst:
## Description
All write operations now use `executeOperation`, ensuring that reads and writes follow the same path for command construction and operation execution. This patch also includes the removal of the `NativeTopology` type, and a number of removed redundant tests from the old `core` driver.

Most of the changes in the `operations` folder are merging existing code into similar files (`insert`/`delete`/`update`), there are no functional changes here.


mbroadst:
@reggi @emadum @nbbeeken ok EVG is all green PTAL

src/bulk/common.ts
```diff
@@ -11,6 +11,10 @@ import {
   hasAtomicOperators,
   maxWireVersion
 } from '../utils';
+import executeOperation = require('../operations/execute_operation');
```
nbbeeken, line 14: Is it worth updating the export format for this while we're here?
This applies to some other imports as well but maybe we want to save this for a TS improvement work later.

src/operations/connect.ts
```diff
@@ -561,6 +552,32 @@ function translateOptions(options: any) {
...
+    secondaryAcceptableLatencyMS: 'acceptableLatency',
+    connectWithNoPrimary: 'secondaryOnlyConnectionAllowed',
+    // Mongos options
+    acceptableLatencyMS: 'localThresholdMS'
+  };
+
+  for (const name in options) {
+    if (translations[name]) {
+      options[translations[name]] = options[name];
+    }
```
nbbeeken, line 579: Is it intentional to leave the original option behind? 
options.ca and options.sslCA will coexist for example

src/sdam/topology.ts
```diff
@@ -1112,4 +983,73 @@ function isRetryableWritesSupported(topology: any) {
...
+    setup_get_property(this, 'hasWriteCommands', writeCommands);
+    setup_get_property(this, 'hasTextSearch', textSearch);
+    setup_get_property(this, 'hasAuthCommands', authCommands);
+    setup_get_property(this, 'hasListCollectionsCommand', listCollections);
+    setup_get_property(this, 'hasListIndexesCommand', listIndexes);
+    setup_get_property(this, 'minWireVersion', ismaster.minWireVersion);
+    setup_get_property(this, 'maxWireVersion', ismaster.maxWireVersion);
+    setup_get_property(this, 'maxNumberOfDocsInBatch', maxNumberOfDocsInBatch);
+    setup_get_property(this, 'commandsTakeWriteConcern', commandsTakeWriteConcern);
+    setup_get_property(this, 'commandsTakeCollation', commandsTakeCollation);
```
nbbeeken, line 1051: I think we should change these to declarative `get xxx() {}` functions which will improve the typing information

nbbeeken:
> Prettier strikes again shouldn't these `function` keywords already have spaces?

I think prettier didn't get applied to every file because eslint isn't applied to every file, seem like an issue with the glob expansion in `check:lint` we should update it to be:
```
"check:lint": "eslint -v && eslint --ext '.js,.ts' src test",
```

src/operations/connect.ts
mbroadst: yes, this is a raw port of the existing `translateOptions` method. I agree its cleaner to pluck options and create a new set of options, but we'll be doing that as part of the mongo client options work.

src/bulk/common.ts
mbroadst: yeah I'd prefer to leave that to ts-specific work

src/sdam/topology.ts
mbroadst: Two thoughts:
- this class must be removed soon, it's a bad class that is being used to make decisions about command construction based on a random server description of a server in the topology.
- converting these to getters will require storing the `ismaster` passed in on the type

This isn't new code, even though it looks like it is. I think we should leave this alone, and work on a better replacement for it in future work.

mbroadst:
@nbbeeken okay I fixed the lint script because there were so few fixes

PR #2480, opened 2020-07-28
Node 1341/session unacknowledged writes

reggi:
https://jira.mongodb.org/browse/NODE-1341?filter=-1


lib/core/wireprotocol/write_command.js
```diff
@@ -36,6 +36,11 @@ function writeCommand(server, type, opsField, ns, ops, options, callback) {
     writeCommand.bypassDocumentValidation = options.bypassDocumentValidation;
   }
 
+  if (writeConcern.w === 0) {
+    // don't include session for unacknowledged writes
+    delete options.session;
```
mbroadst, line 41: From the ticket: 
> I understand why a session ID would be silently omitted for implicit sessions, but what is the reasoning behind omitting it for explicit sessions instead of raising a logic error to the user?

So what this change is doing is "silently omitting" the session if its an unacknowledged write. I  think we want to actually return an error in this case.

reggi: The spec mentions this portion reg SHOULD raising an error, but has a clause in for "for backwards compatibility", and I thought we'd want to maintain this and not have a breaking change for our users.

> * For unacknowledged writes with an explicit session, drivers SHOULD raise an
  error.  If a driver allows users to provide an explicit session with an
  unacknowledged write (e.g. for backwards compatibility), the driver MUST NOT
  send the session ID.

https://github.com/mongodb/specifications/commit/517ec1da43db9458aaef389c60a943c1c9ae446e

mbroadst: I'm fine with not breaking compat in 3.5/3.6, as long as we convert it to an error in 4.0. Also, the evergreen builds are very red for this change.

lib/core/sessions.js
```diff
@@ -708,6 +708,10 @@ function applySession(session, command, options) {
     return new MongoError('Cannot use a session that has ended');
   }
 
+  if (options && options.writeConcern && options.writeConcern.w === 0) {
```
emadum, line 711: I think it would make sense to put a comment here, something like
`// SPEC-1019: silently ignore explicit session with unacknowledged write for backwards compatibility`

PR #2484, opened 2020-07-29
Node 1341/3.6/session unacknowledged writes

reggi:


lib/core/wireprotocol/write_command.js
```diff
@@ -36,6 +36,13 @@ function writeCommand(server, type, opsField, ns, ops, options, callback) {
     writeCommand.bypassDocumentValidation = options.bypassDocumentValidation;
   }
 
+  if (writeConcern && writeConcern.w === 0) {
+    // don't include session for unacknowledged writes
+    if (options) {
+      delete options.session;
```
mbroadst, line 42: If we are going to be silent about it, then I don't think deleting the session is the best approach. Instead I think you'll want to bail early in the [applySession](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/core/sessions.js#L698) method if you determine that there is a unackowledged write concern requested. This will stop the driver from attaching the `lsid` to the outgoing command

reggi:
Waiting for tests to pass in CI then will merge.

PR #2485, opened 2020-07-29
NODE-1341/master/session-unacknowledged-writes

reggi:


test/functional/sessions.test.js
```diff
@@ -194,4 +195,26 @@ describe('Sessions', function () {
...
+      test: withMonitoredClient('insert', { clientOptions: { w: 0 } }, function (
+        client,
+        events,
+        done
+      ) {
+        client
+          .db('test')
+          .collection('foo')
+          .insertOne({ foo: 'bar' }, err => {
+            expect(err).to.not.exist;
```
mbroadst, line 211: Hmm, I would expect the `err` to indeed exist. Looks like something is amiss here

reggi: This is an implicit session, it is my understanding we're only throwing for explicit sessions. I added another test and fixed a throw v.s return in `applySession` bug.

test/functional/sessions.test.js
```diff
@@ -196,7 +197,7 @@ describe('Sessions', function () {
     generateTopologyTests(testSuites, testContext, testFilter);
   });
 
-  context('unacknowledged writes', () => {
+  context.only('unacknowledged writes', () => {
```
mbroadst, line 200: looks like you left an `only` in here

test/functional/sessions.test.js
```diff
@@ -216,5 +217,22 @@ describe('Sessions', function () {
...
+        client,
+        events,
+        done
+      ) {
+        const session = client.startSession({ causalConsistency: true });
+        client
+          .db('test')
+          .collection('foo')
+          .insertOne({ foo: 'bar' }, { session }, err => {
+            expect(err).to.exist;
```
mbroadst, line 232: We have a well-defined message, let's test that it matches here

reggi: sounds good

test/functional/sessions.test.js
reggi: doh

test/functional/sessions.test.js
mbroadst: oops, you're right! While we're here suggest: 
```
expect(event).nested.property('command.writeConcern.w').to.equal(0);
expect(event).nested.property('command.lsid').to.not.exist;
```

test/functional/sessions.test.js
```diff
@@ -194,4 +196,44 @@ describe('Sessions', function () {
 
     generateTopologyTests(testSuites, testContext, testFilter);
   });
+
+  context('unacknowledged writes', () => {
+    it('should not include session for unacknowledged writes', {
```
mbroadst, line 201: this test needs the same `mongodb: >= 3.6` otherwise we're just ensuring sessions aren't added in these unsupported versions

reggi: Add the version to this one test, and I updated all tests to use `>=3.6.0` and NOT `>3.6.0`.

PR #2486, opened 2020-07-29
fix: remove translateOptions export in src/utils

billouboq:
This function is not defined and so exporting it was causing issues in typescript type checker

## Description

**What changed?**
I just removed the wrong export


billouboq:
seems like this PR also fix this https://github.com/mongodb/node-mongodb-native/pull/2487#partial-pull-merging

PR #2487, opened 2020-07-29
chore: Check typescript compiles in CI

nbbeeken:
Adds our build script to our EVG's run-checks task.


.evergreen/run-checks.sh
```diff
@@ -7,3 +7,4 @@ export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 
 npm run check:lint
+npm run prepare
```
mbroadst, line 10: hm we shouldn't need to do this, `prepare` is called as part of `npm install`

nbbeeken:
✅  https://spruce.mongodb.com/version/5f22da0b2fbabe6d7b8af2f2/tasks ✅ 
It passes I swear 🤞

PR #2490, opened 2020-07-31
fix(bulk): allow upsert with empty update when using `bulkWrite()`

vkarpov15:
## Description

In v3.5 `bulkWrite()` could do an `updateOne` with an empty `update`, so long as `upsert` was set. For example, the below script succeeds in v3.5.8 but throws a "Update document requires atomic operators" error in v3.6.

```javascript
const { MongoClient, ObjectId } = require('mongodb');

void async function() {
  const client = await MongoClient.connect('mongodb://localhost:27017/test', {
    useNewUrlParser: true,
    useUnifiedTopology: true
  });

  const db = client.db();
  
  await db.collection('Test').bulkWrite([{
    updateOne: { filter: { name: 'test' }, update: {}, upsert: true }
  }]);

  console.log('Done');
  process.exit(0);
}();

```

**What changed?**

Added a check to allow the above script to pass

Nope

### DIFF
--- a/lib/bulk/common.js
+++ b/lib/bulk/common.js
@@ -973,7 +973,9 @@ class BulkOperationBase {
       if (op.replaceOne && hasAtomicOperators(op[key].replacement)) {
         throw new TypeError('Replacement document must not use atomic operators');
       } else if ((op.updateOne || op.updateMany) && !hasAtomicOperators(op[key].update)) {
-        throw new TypeError('Update document requires atomic operators');
+        if (Object.keys(op[key].update) > 0 || !op[key].upsert) {
+          throw new TypeError('Update document requires atomic operators');
+        }
       }
 
       const multi = op.updateOne || op.replaceOne ? false : true;


mbroadst:
Hey @vkarpov15 , I think this behavior change is desired. Can't you refactor the code to use an atomic operator: 
```
  await db.collection('Test').bulkWrite([{
    updateOne: { filter: { name: 'test' }, update: { $set: {} }, upsert: true }
  }]);
```
?

vkarpov15:
I could do that, the only problem is that the server throws errors when you do an empty `$set`. The below script:

```javascript
const { MongoClient, ObjectId } = require('mongodb');
  
void async function() {
  const client = await MongoClient.connect('mongodb://localhost:27017/test', {
    useNewUrlParser: true,
    useUnifiedTopology: true
  });

  const db = client.db();

  await db.collection('Test').bulkWrite([{
    updateOne: { filter: { name: 'test' }, update: { $set: {} }, upsert: true }
  }]);

  console.log('Done');
  process.exit(0);
}();
```

Throws the below error:

```
$ node ./test.js 
(node:10277) UnhandledPromiseRejectionWarning: BulkWriteError: '$set' is empty. You must specify a field like so: {$set: {<field>: ...}}
    at OrderedBulkOperation.handleWriteError (/troubleshoot-mongoose/node_modules/mongodb/lib/bulk/common.js:1213:9)
    at resultHandler (/troubleshoot-mongoose/node_modules/mongodb/lib/bulk/common.js:519:23)
```

Is this behavior defined in a spec somewhere? If not, I'd like to ask that we not go down the route of throwing errors on empty updates. It's already tricky to work around the empty `$set` issue in the server: https://jira.mongodb.org/browse/SERVER-12266, and this change breaks our workaround.

mbroadst:
@vkarpov15 oops, good catch! Let me get back to you shortly on this one

mbroadst:
@vkarpov15 I think we can close this now based on the discussion in [DRIVERS-1359](https://jira.mongodb.org/browse/DRIVERS-1359)?

PR #2491, opened 2020-08-03
refactor: complete typing for the sdam module

mbroadst:
Building upon the recently merged improved CMAP typings, this completes the SDAM typings so we can finish the rest of the project.

NODE-2726

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -1,7 +1,7 @@
 import Denque = require('denque');
 import { EventEmitter } from 'events';
 import Logger = require('../logger');
-import { Connection, StreamConnectionOptions } from './connection';
+import { Connection, ConnectionOptions } from './connection';
 import { connect } from './connect';
 import { eachAsync, relayEvents, makeCounter } from '../utils';
 import { MongoError } from '../error';
@@ -19,6 +19,7 @@ import {
   ConnectionPoolClearedEvent
 } from './events';
 import type { Callback } from '../types';
+import type { CommandResult } from './commands';
 
 const kLogger = Symbol('logger');
 const kConnections = Symbol('connections');
@@ -92,7 +93,7 @@ function resolveOptions(
   options: ConnectionPoolOptions,
   defaults: Partial<ConnectionPoolOptions>
 ): Readonly<ConnectionPoolOptions> {
-  const newOptions = {} as StreamConnectionOptions;
+  const newOptions = {} as ConnectionOptions;
 
   for (const key of VALID_POOL_OPTIONS) {
     if (key in options) {
@@ -103,7 +104,7 @@ function resolveOptions(
   return Object.freeze(Object.assign({}, defaults, newOptions)) as ConnectionPoolOptions;
 }
 
-export interface ConnectionPoolOptions extends StreamConnectionOptions {
+export interface ConnectionPoolOptions extends ConnectionOptions {
   /** The maximum number of connections that may be associated with a pool at a given time. This includes in use and available connections. */
   maxPoolSize: number;
   /** The minimum number of connections that MUST exist at any moment in a single connection pool. */
@@ -122,6 +123,16 @@ interface WaitQueueMember {
 
 export interface CloseOptions {
   force?: boolean;
+}
+
+// NOTE: to be removed as part of NODE-2745
+export interface ConnectionPool {
+  isConnected(): boolean;
+  write(
+    message: any,
+    commandOptions: any,
+    callback: (err: MongoError, ...args: CommandResult[]) => void
+  ): void;
 }
 
 /**
@@ -297,11 +308,12 @@ export class ConnectionPool extends EventEmitter {
    * @param {boolean} [options.force] Force close connections
    * @param {Function} callback
    */
-  close(callback: Callback): void;
-  close(_options: CloseOptions | Callback, _cb?: Callback): void {
+  close(callback: Callback<void>): void;
+  close(options: CloseOptions, callback: Callback<void>): void;
+  close(_options?: CloseOptions | Callback<void>, _cb?: Callback<void>): void {
     let options = _options as CloseOptions;
-    const callback = (_cb ?? _options) as Callback;
-    if ('function' === typeof options) {
+    const callback = (_cb ?? _options) as Callback<void>;
+    if (typeof options === 'function') {
       options = {};
     }
 
@@ -550,9 +562,9 @@ function processWaitQueue(pool: ConnectionPool) {
  * @param {Function} callback A function to call back after connection management is complete
  */
 type WithConnectionCallback = (
-  error?: MongoError | WaitQueueTimeoutError,
-  connection?: Connection,
-  callback?: Callback<Connection>
+  error: MongoError,
+  connection: Connection | undefined,
+  callback: Callback<Connection>
 ) => void;
 
 /**
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -90,11 +90,10 @@ const VALID_POOL_OPTION_NAMES = [
 const VALID_POOL_OPTIONS = new Set(VALID_POOL_OPTION_NAMES);
 
 function resolveOptions(
-  options: ConnectionPoolOptions,
+  options: Partial<ConnectionPoolOptions>,
   defaults: Partial<ConnectionPoolOptions>
 ): Readonly<ConnectionPoolOptions> {
-  const newOptions = {} as ConnectionOptions;
-
+  const newOptions = {};
   for (const key of VALID_POOL_OPTIONS) {
     if (key in options) {
       (newOptions as { [key: string]: any })[key] = options[key];
@@ -171,7 +170,7 @@ export class ConnectionPool extends EventEmitter {
    *
    * @param {ConnectionPoolOptions} options
    */
-  constructor(options: ConnectionPoolOptions) {
+  constructor(options: Partial<ConnectionPoolOptions>) {
     super();
 
     this.closed = false;
--- a/src/cmap/wire_protocol/command.ts
+++ b/src/cmap/wire_protocol/command.ts
@@ -168,13 +168,8 @@ function _command(
   }
 }
 
-function hasSessionSupport(topology: Server) {
-  if (topology == null) return false;
-  if (topology.description && topology.description) {
-    return topology.description.maxWireVersion >= 6;
-  }
-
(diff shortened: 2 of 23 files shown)


src/sdam/srv_polling.ts
```diff
@@ -23,7 +23,7 @@ class SrvPollingEvent {
     this.srvRecords = srvRecords;
   }
 
-  addresses() {
+  addresses(): Set<string> {
     return new Set(this.srvRecords.map((record: any) => `${record.name}:${record.port}`));
   }
 }
```
nbbeeken, line 29: I can't comment on it, but SrvPoller below looks untyped, is there a reason to leave it off?

src/sdam/server_selection.ts
```diff
@@ -84,10 +93,10 @@ function maxStalenessReducer(readPreference: any, topologyDescription: any, serv
 /**
  * Determines whether a server's tags match a given set of tags
  *
- * @param {string[]} tagSet The requested tag set to match
- * @param {string[]} serverTags The server's tags
+ * @param tagSet The requested tag set to match
```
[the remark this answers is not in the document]
mbroadst: hm the eslintrc was updated for that, maybe reload?

src/sdam/srv_polling.ts
mbroadst: oops, missed this file it looks like. Updated

src/sdam/topology.ts
```diff
@@ -66,20 +66,95 @@ const stateTransition = makeStateMachine({
...
+  srvHost?: boolean;
+  srvPoller?: SrvPoller;
+  logger?: Logger;
+  loggerLevel?: string;
+  directConnection: boolean;
+
+  metadata: ClientMetadata;
+}
+
+interface CursorOptions {
```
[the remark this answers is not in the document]
mbroadst: Hmm, I think actually no because it allows for things like `cursorFactory`. You're right though, we might have a clash, at least this isn't exported

src/sdam/topology.ts
```diff
@@ -94,8 +169,8 @@ interface Topology {
  * @fires Topology#serverHeartbeatFailed
  */
 class Topology extends EventEmitter {
-  s: any;
-  [kWaitQueue]: any;
+  s: TopologyPrivate;
```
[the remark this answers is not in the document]
mbroadst: Also leaving this to follow-on work, since its not immediately obvious how to make it actually private

src/sdam/topology.ts
```diff
@@ -986,7 +1073,7 @@ function isRetryableWritesSupported(topology: any) {
 }
 
 class ServerCapabilities {
-  constructor(ismaster: any) {
+  constructor(ismaster: Document) {
```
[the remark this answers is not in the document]
mbroadst: I know it's a hard pill to swallow, but I think the same rationale applies here as it did with the "operation refactor" a week ago: we need a better solution for this, we shouldn't do it in this PR (I think likely we would want to add this to the `StreamDescription` class, since all operations are executed against `Connection` instances, also if we made these getters we'd have to explicitly save off another copy of the ismaster document. Let's solve this in the "Operation Layer Unification" project

src/sdam/server.ts
```diff
@@ -68,52 +80,64 @@ const stateTransition = makeStateMachine({
...
+
 /**
  * @fires Server#serverHeartbeatStarted
  * @fires Server#serverHeartbeatSucceeded
  * @fires Server#serverHeartbeatFailed
  */
 class Server extends EventEmitter {
-  s: any;
-  clusterTime: any;
+  s: ServerPrivate;
```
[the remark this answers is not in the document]
nbbeeken: Ah yea I see that its accessed in cmap/wire_protocol, there's no `internal` like swift that'd be nice tho, oh well 
Probably whatever properties we use off `s` elsewhere should be made into getters but we can leave that for follow on work.

src/sdam/server_selection.ts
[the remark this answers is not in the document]
nbbeeken: I reviewed before that change came in 😅  all good now!

src/sdam/server.ts
```diff
@@ -264,26 +288,37 @@ class Server extends EventEmitter {
...
+  query(
+    ns: string,
+    cmd: Document,
+    cursorState: InternalCursorState,
+    options: QueryOptions,
+    callback: Callback
+  ) {
     if (this.s.state === STATE_CLOSING || this.s.state === STATE_CLOSED) {
       callback(new MongoError('server is closed'));
       return;
```
nbbeeken, line 324: A few lines down there's an untyped call: `withConnection((err: any, conn: any, cb: any) => {`, you can ctrl-f for this  there should be two other instances that can be typed.

src/sdam/server.ts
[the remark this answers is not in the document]
nbbeeken: I see that we have InternalCursorState, maybe we should go with InternalServer / ServerInternal types to imply that same access level, also I see MonitorPrivate that could get similar treatment.
Just bikeshedding, as is is fine.

src/sdam/topology.ts
```diff
@@ -66,20 +66,95 @@ const stateTransition = makeStateMachine({
...
+  topology?: typeof Topology;
+  maxTimeMS?: number;
+  cursorFactory?: typeof Cursor;
+}
+
+interface ConnectOptions {
+  readPreference?: ReadPreference;
+}
+
+type ServerSelector = Function;
```
nbbeeken, line 152: Is there a more descriptive type for a selector?

src/sdam/server.ts
```diff
@@ -434,14 +460,13 @@ function calculateRoundTripTime(oldRtt: any, duration: any) {
...
     return new MongoError('readPreference must be an instance of ReadPreference');
   }
 }
 
-function executeWriteOperation(args: any, options: any, callback: Function) {
-  if (typeof options === 'function') (callback = options), (options = {});
+function executeWriteOperation(args: any, options: WriteCommandOptions, callback: Callback) {
   options = options || {};
 
   // TODO: once we drop Node 4, use destructuring either here or in arguments.
```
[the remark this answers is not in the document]
mbroadst: @nbbeeken can you please resolve the comments you believe to be completed?

src/sdam/server.ts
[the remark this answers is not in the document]
mbroadst: I think ultimately we want to move away from the `XPrivate` types, and move the values directly onto the parent class, so this might not be a problem much sooner.

src/sdam/topology.ts
mbroadst: Once I turned your extra rules on I was forced to better type `server_selection.ts`, so now we have a better type for this :+1:

mbroadst:
@nbbeeken nice idea with turning those rules on, I made the changes PTAL

src/sdam/events.ts
```diff
@@ -18,9 +35,13 @@ class ServerDescriptionChangedEvent {
  * @property {object} topologyId A unique identifier for the topology
  * @property {ServerAddress} address The address (host/port pair) of the server
  */
-class ServerOpeningEvent {
-  constructor(topologyId: any, address: any) {
-    Object.assign(this, { topologyId, address });
+export class ServerOpeningEvent {
```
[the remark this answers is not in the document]
nbbeeken: @mbroadst I just wanna get your 🧅  on this style for future work, too sugary?

src/cmap/wire_protocol/command.ts
```diff
@@ -162,9 +168,9 @@ function _command(
   }
 }
 
-function hasSessionSupport(topology: Topology | Server) {
+function hasSessionSupport(topology: Server) {
   if (topology == null) return false;
-  if (topology.description) {
+  if (topology.description && topology.description) {
```
reggi, line 173: This is doubled up here.

src/cmap/connection_pool.ts
```diff
@@ -92,7 +93,7 @@ function resolveOptions(
   options: ConnectionPoolOptions,
   defaults: Partial<ConnectionPoolOptions>
 ): Readonly<ConnectionPoolOptions> {
-  const newOptions = {} as StreamConnectionOptions;
+  const newOptions = {} as ConnectionOptions;
```
reggi, line 96: I believe this should be `ConnectionPoolOptions`.

src/sdam/server.ts
```diff
@@ -68,52 +80,64 @@ const stateTransition = makeStateMachine({
...
       options,
-      // the server logger
       logger: new Logger('Server', options),
-      // the server state
       state: STATE_CLOSED,
-      credentials: options.credentials,
-      topology
+      topology,
+      pool: new ConnectionPool(
+        Object.assign({ host: description.host, port: description.port }, options)
```
reggi, line 130: I would use a spread here to help type this down the road.

```
{ ..options, host: description.host, port: description.port }
```

reggi: I just tested it `Object.assign` does type the return value, but we don't need it anymore in master.

mbroadst: This is strange, `Object.assign` works but spreading does not with the following error:
```
Argument of type '{ host: string; port: number; } | { host: string; port: number; credentials?: MongoCredentials | undefined; maxPoolSize: number; minPoolSize: number; maxIdleTimeMS: number; ... 60 more ...; appname?: string | undefined; }' is not assignable to parameter of type 'ConnectionPoolOptions'.
  Type '{ host: string; port: number; }' is missing the following properties from type 'ConnectionPoolOptions': maxPoolSize, minPoolSize, maxIdleTimeMS, waitQueueTimeoutMS, and 8 more.ts(2345)
```

src/sdam/events.ts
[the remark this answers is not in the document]
mbroadst: oh hi! whoops ,I think it's a little sugary for my taste. It definitely saves some duplication, but I think readability suffers a little bit for it. I think of it similar to shortenings like:
```
```
people often see that in our codebase and think "Is that legal JavaScript?" Similarly here I read this and was like "is that a copy/paste accident?"

src/sdam/server.ts
reggi: Yeah here changing it to a spread actually shows us some type errors. The min / max pool stuff in `ConnectionPoolOptions` should have `?` then a bunch of the other options in `ConnectionOptions` might need them as well.

reggi: `Object.assign` is not to be trusted! 👻

PR #2493, opened 2020-08-03
NODE-2747: mongo client options

reggi:
https://jira.mongodb.org/browse/NODE-2747


src/types.d.ts
```diff
@@ -58,24 +78,300 @@ export interface AutoEncryptionOptions {
...
+  appName?: string;
+  /** Enables retryable reads. */
+  retryReads?: boolean;
+  /** Enable retryable writes. */
+  retryWrites?: boolean;
+  /** Allow a driver to force a Single topology type with a connection string containing one host */
+  directConnection?: boolean;
+}
+
+export interface MongoClientOptions extends MongoURIOptions, BSONSerializeOptions {
```
mbroadst, line 284: lookin really good, but I think based on our conversation yesterday we want to put this definition into `mongo_client.ts`. The `types.d.ts` file is currently where "global" or "missing" types are going (auto encrypter is not typed for instance)

src/mongo_client.ts
```diff
@@ -26,7 +351,7 @@ import type { MongoClientOptions } from './types';
  * @property {string} [platform] Optional platform information
  */
 
-interface MongoClient {
+export default interface MongoClient {
```
mbroadst: I agree with @nbbeeken , let's not use default exports at all

PR #2495, opened 2020-08-03
chore(ci): add kerberos auth tests

emadum:
## Description

[NODE-876](https://jira.mongodb.org/browse/NODE-872)

**What changed?**


emadum:
Patch run on RHEL/Ubuntu/Mac

https://spruce.mongodb.com/version/5f288c21d1fe0706461a1693/tasks

test/manual/kerberos.test.js
```diff
@@ -0,0 +1,352 @@
+'use strict';
+
+const format = require('util').format;
+const test = require('../functional/shared').assert;
+const { MongoClient } = require('../../src');
+
+// You need to set up the kinit tab first
```
mbroadst, line 7: Let's remove all of this, we know where to look and we don't want to give any details externally

test/manual/kerberos.test.js
```diff
@@ -0,0 +1,352 @@
...
+  if (process.env.MONGODB_URI == null) {
+    console.log('skipping Kerberos tests, MONGODB_URI environment variable is not defined');
+    return;
+  }
+  if (process.platform === 'win32') {
+    console.log('Only running Win32 tests');
+    return;
+  }
+
+  it.only('Should Correctly Authenticate using kerberos with MongoClient', function (done) {
```
mbroadst, line 22: there's an `it.only` left here. Also, nit: `'should authenticate using kerberos`

test/manual/kerberos.test.js
```diff
@@ -0,0 +1,352 @@
...
+// kinit -p drivers@LDAPTEST.10GEN.CC
+// password: (not shown)
+
+describe('Kerberos', function () {
+  if (process.env.MONGODB_URI == null) {
+    console.log('skipping Kerberos tests, MONGODB_URI environment variable is not defined');
+    return;
+  }
+  if (process.platform === 'win32') {
+    console.log('Only running Win32 tests');
```
mbroadst, line 18: This message looks wrong, also we would eventually want to run this against windows

test/manual/kerberos.test.js
```diff
@@ -0,0 +1,352 @@
...
+  });
+
+  it('Validate that SERVICE_REALM and CANONICALIZE_HOST_NAME is passed in', function (done) {
+    var configuration = this.configuration;
+
+    // KDC Server
+    var server = 'ldaptest.10gen.cc';
+    var principal = 'drivers@LDAPTEST.10GEN.CC';
+    var urlEncodedPrincipal = encodeURIComponent(principal);
+    const url = format(
```
mbroadst, line 46: let's use string interpolation instead of `format`

test/manual/kerberos.test.js
```diff
@@ -0,0 +1,352 @@
+'use strict';
```
mbroadst, line 1: Can you generally spruce this file up a bit? 
- use `let`/`const` everywhere
- switch to using `chai` instead of the legacy test assert helpers
- `it.skip` the tests we don't think are useful (or better yet, delete them)
  - for instance, we no longer have a `reconnect` event for topologies
- nit: fix the test titles to be more consistent (lower case them, don't mention redundant info "should correctly", or "on MongoClient")
-

.evergreen/run-kerberos-tests.sh
```diff
@@ -0,0 +1,31 @@
...
+# set up keytab
+mkdir -p "$(pwd)/.evergreen"
+touch "$(pwd)/.evergreen/krb5.conf.empty"
+export KRB5_CONFIG="$(pwd)/.evergreen/krb5.conf.empty"
+echo "Writing keytab"
+# DON'T PRINT KEYTAB TO STDOUT
+set +o verbose
+DECODE='-d'
+OS=`uname`
+if [[ $OS == "Darwin" ]]; then
```
mbroadst, line 19: This `if` statement is set up to switch on the value `DECODE`, why not just do the work inside and save some lines:
```
if [[ "$OSTYPE" == "darwin"* ]]; then
    echo $KEYTAB | base64 -D > "$(pwd)/.evergreen/drivers.keytab"
else
    echo $KEYTAB | base64 -d > "$(pwd)/.evergreen/drivers.keytab"
fi
```

test/manual/kerberos.test.js
emadum: Just to clear up what's going on here, I added `.only` to the two tests that I fixed from the original disabled test file. I haven't touched the other tests, which either don't run because they are win32-specific and we don't currently test windows, or because they don't have `.only`.

test/manual/kerberos.test.js
emadum: This is meant to be the equivalent of `{ metadata: { os: '!win32' } }` without depending on our custom runner. There's a set of tests in the second `describe` that are win32-specific.

test/manual/kerberos.test.js
```diff
@@ -0,0 +1,83 @@
...
+        .toArray(function (err, docs) {
+          expect(err).to.not.exist;
+          expect(docs[0].kerberos).to.be.true;
+
+          client.close(done);
+        });
+    });
+  });
+
+  it.skip('should fail to authenticate due to illegal service name', function (done) {
```
emadum, line 76: Note - I had to skip this test because it connects successfully. This makes me wonder if the above tests which add additional query parameters are actually doing anything, but I left them in there since they pass.

test/manual/kerberos.test.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+  let krb5Uri = process.env.MONGODB_URI;
+  if (process.platform === 'win32') {
+    console.log('Win32 run detected');
+    if (process.env.LDAPTEST_PASSWORD == null) {
+      throw new Error('The env parameter LDAPTEST_PASSWORD must be set');
+    }
+    const parts = krb5Uri.split('@', 2);
+    krb5Uri = `${parts[0]}:${process.env.LDAPTEST_PASSWORD}@${parts[1]}`;
+  }
```
emadum, line 19: In the original disabled tests, there were separate copies of the tests for Win32, but adding the password into the URL from the environment variable was the only difference between them. I added this logic to preserve the original logic, but haven't actually tried running the tests on Windows, since we don't currently have a win32 variant.

test/manual/kerberos.test.js
emadum: Yeah here's [an example log](https://spruce.mongodb.com/task/mongo_node_driver_ubuntu_14.04_dubnium_test_auth_kerberos_patch_4fcb031bb753facbc6d9a4ebffe6279a3dd81941_5f2ab83461837d0cc464af1f_20_08_05_13_46_49/logs). The `gssapiServiceName` doesn't seem to matter; maybe it gets ignored?

PR #2498, opened 2020-08-05
NODE-2732: Driver Type Improvements

nbbeeken:
This is a collection of common type changes that will impact of each of our parallel typing work. Mostly changing `Function` to `Callback` and switching to es6 import syntax which impacts every file you import something from.

Changing to the Callback type also forced the change to setting the first/err arg to undefined instead of null, since there were many tests that explicitly checked for null you'll see those have been updated to check for not.exist instead.


src/mongo_client.ts
```diff
@@ -196,7 +197,7 @@ class MongoClient extends EventEmitter {
    * @param {MongoClient~connectCallback} [callback] The command result callback
    * @returns {Promise<MongoClient>} returns Promise if no callback passed
    */
-  connect(callback: Function): Promise<MongoClient> {
+  connect(callback: Callback): Promise<MongoClient> {
```
reggi, line 200: Can you make this optional please! Seems to have slipped though.

PR #2509, opened 2020-08-10
NODE-2759: Type annotations for cursor modules

nbbeeken:
Typings for the cursor module. There are some small changes outside the cursor module that I can merge to the base operations branch to make the diff clear, lmk.

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -254,7 +254,8 @@ export class ChangeStream extends EventEmitter {
   next(callback?: Callback): Promise<void> | void {
     return maybePromise(callback, cb => {
       getCursor(this, (err, cursor) => {
-        if (err || !cursor) return cb(err); // failed to resume, raise an error
+        if (err) return cb(err); // failed to resume, raise an error
+        if (!cursor) return cb(new MongoError('Cursor is undefined'));
         cursor.next((error, change) => {
           if (error || !change) {
             this[kResumeQueue].push(() => this.next(cb));
@@ -360,7 +361,7 @@ export interface ChangeStreamCursorOptions extends CursorOptions {
   startAfter?: boolean;
 }
 
-class ChangeStreamCursor extends Cursor<ChangeStreamCursorOptions, AggregateOperation> {
+class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamCursorOptions> {
   _resumeToken: ResumeToken;
   startAtOperationTime?: OperationTime;
   hasReceived?: boolean;
@@ -608,8 +609,6 @@ function processNewChange(
   change: ChangeStreamDocument,
   callback?: Callback
 ) {
-  const cursor = changeStream.cursor;
-
   // a null change means the cursor has been notified, implicitly closing the change stream
   if (change == null) {
     changeStream.closed = true;
@@ -630,7 +629,7 @@ function processNewChange(
   }
 
   // cache the resume token
-  cursor?.cacheResumeToken(change._id);
+  changeStream.cursor?.cacheResumeToken(change._id);
 
   // wipe the startAtOperationTime if there was one so that there won't be a conflict
   // between resumeToken and startAtOperationTime if we need to reconnect the cursor
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -2,7 +2,7 @@ import Denque = require('denque');
 import { EventEmitter } from 'events';
 import { MongoError, isResumableError } from './error';
 import { Cursor } from './cursor';
-import { AggregateOperation } from './operations/aggregate';
+import { AggregateOperation, AggregateOptions } from './operations/aggregate';
 import { loadCollection, loadDb, loadMongoClient } from './dynamic_loaders';
 import {
   relayEvents,
@@ -16,13 +16,10 @@ import type { Callback, Document, AnyError } from './types';
 import type { CursorOptions } from './cursor/cursor';
 import type { ReadPreference } from './read_preference';
 import type { Timestamp } from 'bson';
-import type { CollationOptions } from './cmap/wire_protocol/write_command';
-import type { Collection } from './collection';
-import type { Db } from './db';
-import type { MongoClient } from './mongo_client';
 import type { Topology } from './sdam/topology';
 import type { Writable } from 'stream';
 import type { StreamOptions } from './cursor/core_cursor';
+import type { Parent } from './operations/command';
 const kResumeQueue = Symbol('resumeQueue');
 
 const CHANGE_STREAM_OPTIONS = ['resumeAfter', 'startAfter', 'startAtOperationTime', 'fullDocument'];
@@ -46,8 +43,12 @@ export type ResumeToken = unknown;
  */
 export type OperationTime = Timestamp;
 
+export interface PipeOptions {
+  end?: boolean;
+}
+
 /** Options that can be passed to a ChangeStream. Note that startAfter, resumeAfter, and startAtOperationTime are all mutually exclusive, and the server will error if more than one is specified. */
-export interface ChangeStreamOptions {
+export interface ChangeStreamOptions extends AggregateOptions {
   /** Allowed values: ‘default’, ‘updateLookup’. When set to ‘updateLookup’, the change stream will include both a delta describing the changes to the document, as well as a copy of the entire document that was changed from some time after the change occurred. */
   fullDocument?: string;
   /** The maximum amount of time for the server to wait on new documents to satisfy a change stream query. */
@@ -60,10 +61,6 @@ export interface ChangeStreamOptions {
   startAtOperationTime?: OperationTime;
   /** The number of documents to return per batch. See {@link https://docs.mongodb.com/manual/reference/command/aggregate|aggregation documentation}. */
   batchSize?: number;
-  /** Specify collation settings for operation. See {@link https://docs.mongodb.com/manual/reference/command/aggregate|aggregation documentation}. */
-  collation?: CollationOptions;
-  /** The read preference. Defaults to the read preference of the database or collection. See {@link https://docs.mongodb.com/manual/reference/read-preference|read preference documentation}. */
-  readPreference?: ReadPreference;
 }
 
 interface ChangeStreamDocument {
@@ -143,9 +140,6 @@ interface UpdateDescription {
 /**
  * Creates a new Change Stream instance. Normally created using {@link Collection#watch|Collection.watch()}.
  *
- * @class ChangeStream
- * @since 3.0.0
- 
  * @fires ChangeStream#close
  * @fires ChangeStream#change
  * @fires ChangeStream#end
@@ -156,8 +150,8 @@ interface UpdateDescription {
 export class ChangeStream extends EventEmitter {
   pipeline: Document[];
   options: ChangeStreamOptions;
-  parent: Collection | Db | MongoClient;
-  namespace: string | MongoDBNamespace;
+  parent: Parent;
+  namespace: MongoDBNamespace;
   type: symbol;
   topology: Topology;
   cursor?: ChangeStreamCursor;
@@ -170,11 +164,7 @@ export class ChangeStream extends EventEmitter {
    * @param parent - The parent object that created this change stream
    * @param pipeline - An array of {@link https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/|aggregation pipeline stages} through which to pass change stream documents
    */
-  constructor(
-    parent: Collection | Db | MongoClient,
-    pipeline: Document[] = [],
-    options: ChangeStreamOptions = {}
-  ) {
+  constructor(parent: Parent, pipeline: Document[] = [], options: ChangeStreamOptions = {}) {
     super();
 
     const Collection = loadCollection();
@@ -311,12 +301,15 @@ export class ChangeStream extends EventEmitter {
    * @param {Writable} destination The destination for writing data
    * @param {object} [options] {@link https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options|Pipe options}
    */
-  pipe(destination: Writable, options?: { end?: boolean }): Writable {
+  pipe(destination: Writable, options?: PipeOptions): Writable {
     if (!this.pipeDestinations) {
       this.pipeDestinations = [];
     }
     this.pipeDestinations.push(destination);
-    return this.cursor!.pipe(destination, options);
+    if (!this.cursor) {
+      throw new MongoError('ChangeStream has no cursor, unable to pipe');
+    }
+    return this.cursor.pipe(destination, options);
   }
 
   /**
@@ -329,23 +322,35 @@ export class ChangeStream extends EventEmitter {
     if (this.pipeDestinations && destinationIndex > -1) {
       this.pipeDestinations.splice(destinationIndex, 1);
     }
-    return this.cursor!.unpipe(destination);
+    if (!this.cursor) {
+      throw new MongoError('ChangeStream has no cursor, unable to unpipe');
+    }
+    return this.cursor.unpipe(destination);
   }
 
   /** Return a modified Readable stream including a possible transform method. */
   stream(options?: StreamOptions): ChangeStreamCursor {
     this.streamOptions = options;
-    return this.cursor!.stream(options);
+    if (!this.cursor) {
+      throw new MongoError('ChangeStream has no cursor, unable to stream');
+    }
+    return this.cursor.stream(options);
   }
 
   /** This method will cause a stream in flowing mode to stop emitting data events. Any data that becomes available will remain in the internal buffer. */
   pause(): ChangeStreamCursor {
-    return this.cursor!.pause();
+    if (!this.cursor) {
+      throw new MongoError('ChangeStream has no cursor, unable to pause');
+    }
+    return this.cursor.pause();
   }
 
   /** This method will cause the readable stream to resume emitting data events. */
   resume(): ChangeStreamCursor {
-    return this.cursor!.resume();
+    if (!this.cursor) {
+      throw new MongoError('ChangeStream has no cursor, unable to resume');
+    }
+    return this.cursor.resume();
   }
 }
 
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -4,8 +4,25 @@ import { MongoError, isResumableError } from './error';
 import { Cursor } from './cursor';
 import { AggregateOperation } from './operations/aggregate';
 import { loadCollection, loadDb, loadMongoClient } from './dynamic_loaders';
-import { relayEvents, maxWireVersion, calculateDurationInMs, now, maybePromise } from './utils';
-import type { Callback } from './types';
+import {
+  relayEvents,
+  maxWireVersion,
+  calculateDurationInMs,
+  now,
+  maybePromise,
+  MongoDBNamespace
+} from './utils';
+import type { Callback, Document, AnyError } from './types';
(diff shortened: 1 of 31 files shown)


src/cursor/core_cursor.ts
```diff
@@ -799,7 +762,7 @@ function nextFunction(self: any, callback: Callback) {
     return setCursorDeadAndNotified(self, callback);
   } else if (
     self.cursorState.cursorIndex === self.cursorState.documents.length &&
-    !Long.ZERO.equals(self.cursorState.cursorId)
+    !Long.ZERO.equals(self.cursorState.cursorId!)
```
mbroadst, line 765: there should always be a cursorId at this point, otherwise we should bail much earlier

src/cursor/core_cursor.ts
```diff
@@ -813,10 +776,10 @@ function nextFunction(self: any, callback: Callback) {
 
     // Check if connection is dead and return if not possible to
     // execute a getMore on this connection
-    if (isConnectionDead(self, callback)) return;
+    if (isConnectionDead(self as Cursor, callback)) return;
```
mbroadst, line 779: should not need to cast, `isConnectionDead` takes in a `CoreCursor`

src/cursor/core_cursor.ts
```diff
@@ -905,4 +868,37 @@ function nextFunction(self: any, callback: Callback) {
   }
 }
 
-export { CursorState, CoreCursor };
+function getLimitSkipBatchSizeDefaults(
+  options: CoreCursorOptions,
+  cmd: Document | Long | undefined
```
mbroadst, line 873: `cmd` === `Long` or `undefined` seems like a bug

src/sdam/topology.ts
```diff
@@ -652,40 +652,44 @@ export class Topology extends EventEmitter {
...
-        options.session?.incrementTransactionNumber();
-        options.willRetryWrite = willRetryWrite;
-      }
-
-      server.command(ns, cmd, options, (err, result) => {
-        if (!err) return callback(undefined, result);
-        if (!shouldRetryOperation(err)) {
-          return callback(err);
+    this.selectServer(
+      readPreferenceServerSelector(readPreference as ReadPreference),
```
mbroadst, line 656: after `ReadPreference.translate` we should always be dealing with a `ReadPreference` instance, there should be no need to cast here

src/collection.ts
```diff
@@ -783,7 +783,7 @@ export class Collection {
     const cursor = new CommandCursor(
       this.s.topology,
       new ListIndexesOperation(this, options),
-      options
+      options as Document
```
mbroadst, line 786: :monocle_face:

src/cursor/cursor.ts
```diff
@@ -69,9 +74,16 @@ export interface CursorPrivate {
 }
 
 export interface CursorOptions extends CoreCursorOptions {
-  topology?: typeof Topology;
+  tailableRetryInterval?: number;
+  explicitlyIgnoreSession?: boolean;
+  cursor?: Cursor;
+  /** The internal topology of the created cursor */
+  topology?: Topology;
```
mbroadst, line 81: there seem to be duplicates between `CursorOptions` and `CoreCursorOptions`

src/cursor/cursor.ts
```diff
@@ -114,20 +125,17 @@ export interface CursorOptions extends CoreCursorOptions {
...
-   * @param {any} ns
-   * @param {any} [cmd]
-   * @param {any} [options]
-   */
-  s: any;
-  constructor(topology: any, ns: any, cmd?: any, options?: any) {
+export class Cursor<
+  T extends CursorOptions = CursorOptions,
+  O extends OperationBase = OperationBase
+> extends CoreCursor<T, O> {
```
mbroadst, line 131:
```suggestion
export class Cursor<
  O extends OperationBase = OperationBase,
  T extends CursorOptions = CursorOptions
> extends CoreCursor<T, O> {
```
Let's match the constructor: operation, then options

src/cursor/cursor.ts
```diff
@@ -114,20 +125,17 @@ export interface CursorOptions extends CoreCursorOptions {
...
+  s: CursorPrivate;
+
+  constructor(topology: Topology, ns: string | O, cmd: Document, options: T = {} as T) {
     super(topology, ns, cmd, options);
 
-    options = options || {};
     if (this.operation) {
-      options = this.operation.options;
+      options = this.operation.options as T;
```
mbroadst, line 138: I think if you remove the casting, this should actually catch errors where the options are not the same as the operation.

src/cursor/cursor.ts
```diff
@@ -158,34 +176,21 @@ export class Cursor extends CoreCursor {
...
-    } else if (typeof options.batchSize === 'number') {
-      batchSize = options.batchSize;
-    }
-
-    // Set the batchSize
-    this.setCursorBatchSize(batchSize);
   }
 
-  get readPreference() {
+  get readPreference(): ReadPreferenceLike | undefined {
```
mbroadst, line 181: this should always return a `ReadPreference`, not a `ReadPreferenceLike` and definitely not `undefined`

src/cursor/cursor.ts
```diff
@@ -206,97 +211,72 @@ export class Cursor extends CoreCursor {
...
+      throw new MongoError('Cursor is closed');
     }
 
     this.cmd.query = filter;
     return this;
   }
 
   /**
-   * Set the cursor maxScan
+   * @deprecated Set the cursor maxScan
```
mbroadst, line 273: I think `@deprecated` should be on its own line below

src/cursor/core_cursor.ts
nbbeeken: This method uses the pool property on CoreCursor which you said we don't need, is Killed or Dead equivalent?

src/cursor/core_cursor.ts
nbbeeken: Yea I fixed with a check, but if only Long.equals took undefined and just returned false wouldn't need `!` 🤷

src/cursor/core_cursor.ts
```diff
@@ -3,139 +3,189 @@ import { ReadPreference } from '../read_preference';
...
+  cmd: Document;
+  options: T;
+  topology: Topology;
   cursorState: InternalCursorState;
-  logger: any;
-  query: any;
-  s: any;
+  logger: Logger;
+  query?: Document;
+  s!: CoreCursorPrivate;
```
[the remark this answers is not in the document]
nbbeeken: Bit of a mess but only subclasses assign and make their s property but CoreCursor doesn't. I think I could move the logic of making the s prop into core cursor. The hardest part is avoiding optional `?:` on things that will always be defined because the whole s object needs to be declared at once.

src/cursor/core_cursor.ts
```diff
@@ -3,139 +3,189 @@ import { ReadPreference } from '../read_preference';
...
-  cmd: any;
-  options: any;
-  topology: any;
+export class CoreCursor<
+  T extends CoreCursorOptions = CoreCursorOptions,
+  O extends OperationBase = OperationBase
+> extends Readable {
+  operation?: O;
+  pool!: ConnectionPool;
+  server!: Server;
```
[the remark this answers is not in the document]
nbbeeken: It's assigned in _initializeCursor, which does say: '// NOTE: this goes away once cursors use `executeOperation`'

src/change_stream.ts
```diff
@@ -454,22 +542,22 @@ function createChangeStreamCursor(self: any, options: any) {
...
-  changeStreamCursor.on('error', function (error: any) {
+  changeStreamCursor.on('error', function (error) {
     processError(self, error);
   });
 
   if (self.pipeDestinations) {
     const cursorStream = changeStreamCursor.stream(self.streamOptions);
-    for (let pipeDestination in self.pipeDestinations) {
-      cursorStream.pipe(pipeDestination as any);
+    for (const pipeDestination of self.pipeDestinations) {
```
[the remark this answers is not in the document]
nbbeeken: I can keep this on my radar for after typing work
https://jira.mongodb.org/browse/NODE-2550

src/change_stream.ts
```diff
@@ -219,69 +310,63 @@ export class ChangeStream extends EventEmitter {
    * @function
    * @param {Writable} destination The destination for writing data
    * @param {object} [options] {@link https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options|Pipe options}
-   * @returns {null}
    */
-  pipe(destination: any, options?: object): null {
+  pipe(destination: Writable, options?: { end?: boolean }): Writable {
```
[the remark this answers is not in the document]
nbbeeken: I did this because its inlined in the node types, but I pulled it out, not sure if they have a set of rules over there

```ts
pipe<T extends NodeJS.WritableStream>(destination: T, options?: { end?: boolean; }): T;
```

src/cursor/cursor.ts
nbbeeken: I get this error:
```
Type 'OperationOptions' is not assignable to type 'T'.
  'OperationOptions' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'CursorOptions'
```
I don't think it can catch the type errors because none exist technically, but they could if the subtype has properties that aren't defined on the base type.

src/operations/command.ts
```diff
@@ -2,32 +2,33 @@ import { Aspect, OperationBase, OperationOptions } from './operation';
...
 const SUPPORTS_WRITE_CONCERN_AND_COLLATION = 5;
 
-export interface CommandOperationOptions extends OperationOptions, BSONSerializeOptions {
+export interface CommandOperationOptions extends OperationOptions {
   fullResponse?: boolean;
   /** Specify a read concern and level for the collection. (only MongoDB 3.2 or higher supported) */
   readConcern?: ReadConcern;
   /** The preferred read preference (ReadPreference.PRIMARY, ReadPreference.PRIMARY_PREFERRED, ReadPreference.SECONDARY, ReadPreference.SECONDARY_PREFERRED, ReadPreference.NEAREST). */
-  readPreference?: ReadPreferenceLike;
+  readPreference?: ReadPreference;
```
mbroadst, line 25: In this case I think you do want to keep `ReadPreferenceLike`. I think of `ReadPreferenceLike` being the value that is acceptable for public APIs, but internally we always want to be working with `ReadPreference`

src/operations/operation.ts
```diff
@@ -33,8 +34,9 @@ export class OperationBase<T extends OperationOptions = OperationOptions> {
   ns!: MongoDBNamespace;
   cmd!: Document;
 
-  readPreference?: ReadPreference;
+  readPreference!: ReadPreference;
```
mbroadst, line 37: why the `!`?

src/types.d.ts
```diff
@@ -4,10 +4,14 @@ import type { MongoClient } from './mongo_client';
 
 export type AnyError = MongoError | Error;
 
-export type Callback<T = any> = (error?: AnyError, result?: T) => void;
+// export type Callback<T = any> = (error?: AnyError, result?: T) => void;
 export type Callback2<T0 = any, T1 = any> = (error?: AnyError, result0?: T0, result1?: T1) => void;
 export type CallbackWithType<E = AnyError, T0 = any> = (error?: E, result?: T0) => void;
 
+export interface Callback<T = any> {
```
mbroadst, line 11: what's motivating this change?

src/cursor/core_cursor.ts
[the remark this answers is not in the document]
mbroadst: Oh I see, no this is necessary. I think it should be a `server?` which could actually second as a check for whether the cursor has been initialized. We pin to a server once the cursor begins iteration, so that's why its saved off here

src/cursor/core_cursor.ts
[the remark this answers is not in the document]
mbroadst: ok maybe we can leave this alone for now then

src/cursor/core_cursor.ts
```diff
@@ -3,139 +3,189 @@ import { ReadPreference } from '../read_preference';
...
+  topology: Topology;
   cursorState: InternalCursorState;
-  logger: any;
-  query: any;
-  s: any;
+  logger: Logger;
+  query?: Document;
+  s!: CoreCursorPrivate;
+
+  connection?: Connection;
```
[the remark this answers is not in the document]
mbroadst: this one still looks like it should be removed, the code has:
```
      // Save the returned connection to ensure all getMore's fire over the same connection
      self.connection = connection;
```
but its never actually used, fine to delete

src/cursor/core_cursor.ts
mbroadst: true, but I think in terms of defensive coding this forces us to check for a `cursorId` much earlier which is probably a good thing

src/cursor/core_cursor.ts
mbroadst: yeah hmm, maybe we don't need this method anymore. It was meant to check synchronously whether there was an active connection to be used, but we use `executeOperation` for everything now which will in turn do server selection. Everything is happening asynchronously now, so we can probably get rid of this method. Can you make a ticket to circle back around to this, and we'll leave the casting for now.

src/cursor/cursor.ts
```diff
@@ -403,52 +374,42 @@ export class Cursor extends CoreCursor {
...
-   * @param {string} flag The flag to set, must be one of following ['tailable', 'oplogReplay', 'noCursorTimeout', 'awaitData', 'partial'].
-   * @param {boolean} value The flag boolean value.
-   * @throws {MongoError}
-   * @returns {Cursor}
+   * @param flag The flag to set, must be one of following ['tailable', 'oplogReplay', 'noCursorTimeout', 'awaitData', 'partial' -.
+   *
+   * @param value - The flag boolean value.
    */
-  addCursorFlag(flag: string, value: boolean): Cursor {
+  addCursorFlag(flag: string, value: boolean): this {
```
reggi, line 402: If you define `FLAGS` as const up above, you can have a `CursorFlags` type which will be a union of all strings in the array.

```ts
const FLAGS = [
  'tailable',
  'oplogReplay',
  'noCursorTimeout',
  'awaitData',
  'exhaust',
  'partial'
] as const;
type CursorFlags = typeof FLAGS[number];
```

```suggestion
  addCursorFlag(flag: CursorFlags, value: boolean): this {
```

src/cursor/cursor.ts
```diff
@@ -836,58 +722,43 @@ export class Cursor extends CoreCursor {
...
    * results when this cursor had been previously accessed. In that case,
    * cursor.rewind() can be used to reset the cursor.
    *
-   * @function
-   * @param {Cursor~toArrayResultCallback} [callback] The result callback.
-   * @throws {MongoError}
-   * @returns {Promise<void> | void} returns Promise if no callback passed
+   * @param callback - The result callback.
    */
   toArray(callback: Callback<Document[]>): Promise<void> | void {
```
reggi, line 733: Should this return `Promise<Document[]>`?

src/change_stream.ts
```diff
@@ -143,9 +235,9 @@ export class ChangeStream extends EventEmitter {
    * @returns {Promise<void>|void} returns Promise if no callback passed
    */
   hasNext(callback?: Callback): Promise<void> | void {
-    return maybePromise(callback, (cb: any) => {
-      getCursor(this, (err?: any, cursor?: any) => {
-        if (err) return cb(err); // failed to resume, raise an error
+    return maybePromise(callback, cb => {
+      getCursor(this, (err, cursor) => {
+        if (err || !cursor) return cb(err); // failed to resume, raise an error
```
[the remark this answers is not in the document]
mbroadst: I've been guilty of this as well, but agree we need a better pattern. Mostly this convention let's us avoid writing code like this:
```
method((err, res) => {
  if (err) { return doSomething(); }
  if (res) { doSomethingElse(); }
});
```
It would be ideal if we could somehow hint using TypeScript that the optionality of `err` and `res` are mutually exclusive, i.e. when `err` is `undefined` then `res` exists. Let's come up with a convention for this outside of the scope of this PR, and then comb the codebase and apply it everywhere.

src/cursor/cursor.ts
```diff
@@ -403,52 +374,42 @@ export class Cursor extends CoreCursor {
...
-   * @param {string} field The cursor option to set ['numberOfRetries', 'tailableRetryInterval'].
-   * @param {object} value The field value.
-   * @throws {MongoError}
-   * @returns {Cursor}
+   * @param field The cursor option to set ['numberOfRetries', 'tailableRetryInterval' -.
+   *
+   * @param value - The field value.
    */
-  setCursorOption(field: string, value: object): Cursor {
+  setCursorOption(field: keyof CursorPrivate, value: CursorPrivate[keyof CursorPrivate]): this {
```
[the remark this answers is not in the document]
nbbeeken: This function only sets `'numberOfRetries' | 'tailableRetryInterval'` but I did make a type for that.

src/change_stream.ts
```diff
@@ -478,19 +571,27 @@ function applyKnownOptions(target: any, source: any, optionNames: any) {
...
   setTimeout(() => {
     if (options && options.start == null) {
       options.start = now();
     }
 
     const start = options.start || now();
     const timeout = options.timeout || SELECTION_TIMEOUT;
-    const readPreference = options.readPreference;
-    if (topology.isConnected({ readPreference })) {
+    if (topology.isConnected()) {
```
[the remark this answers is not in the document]
nbbeeken: isConnected doesn't take an argument, it must've been removed without changing the callers so it doesn't rely on it anymore

src/change_stream.ts
```diff
@@ -219,69 +300,78 @@ export class ChangeStream extends EventEmitter {
...
    */
-  unpipe(destination?: any): null {
-    if (this.pipeDestinations && this.pipeDestinations.indexOf(destination) > -1) {
-      this.pipeDestinations.splice(this.pipeDestinations.indexOf(destination), 1);
+  unpipe(destination?: Writable): ChangeStreamCursor {
+    const destinationIndex = destination ? this.pipeDestinations.indexOf(destination) : -1;
+    if (this.pipeDestinations && destinationIndex > -1) {
+      this.pipeDestinations.splice(destinationIndex, 1);
+    }
+    if (!this.cursor) {
```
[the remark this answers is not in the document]
nbbeeken: Each error message is different and I can't _pipe_ that through to the getter/setter, also I'd make the maybeCursor private but its used outside of the class so it seems like it would be confusing which one to use rather than just having the checks.

src/types.d.ts
nbbeeken: Whoops, experimenting, reverted.

src/operations/operation.ts
nbbeeken: In cursor's get readPreference it can't return undefined for a readPref, I moved the assertion there instead.

src/cursor/cursor.ts
nbbeeken: I think I got them all now, I wonder if we could find tooling to detect this.

src/sdam/topology.ts
nbbeeken: There doesn't seem to be a straight forward way to type that since translate mutates the object it takes in. You think I can do something with extending the argument passed in? I tried a few ways but it can't override an incompatible type, like string and ReadPreference are not compatible.

src/change_stream.ts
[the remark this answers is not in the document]
reggi: This makes sense, this trick may be useful elsewhere.

PR #2511, opened 2020-08-11
fix: addeds text index options

reggi:
https://jira.mongodb.org/browse/NODE-2755

### DIFF
--- a/lib/operations/create_indexes.js
+++ b/lib/operations/create_indexes.js
@@ -7,19 +7,35 @@ const MongoError = require('../core').MongoError;
 const parseIndexOptions = require('../utils').parseIndexOptions;
 const maxWireVersion = require('../core/utils').maxWireVersion;
 
-const validIndexOptions = new Set([
+const VALID_INDEX_OPTIONS = new Set([
+  'background',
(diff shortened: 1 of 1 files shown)


lib/operations/create_indexes.js
```diff
@@ -15,7 +15,11 @@ const validIndexOptions = new Set([
   'expireAfterSeconds',
   'storageEngine',
   'collation',
-  'bucketSize'
+  'bucketSize',
```
mbroadst, line 18: so it looks like for some reason we're missing a whole [bunch](https://github.com/mongodb/node-mongodb-native/blob/master/src/operations/indexes.ts#L13-L42) of options. My guess is that the old code for this would just merge in whatever options were provided, so if we're going to be strict we'll need to backport the list from 4.0

reggi: updated!

zuozp8:
@mbroadst could we we have release with this fix? it may be quite unfortunate to create index with bad options

mbroadst:
@zuozp8 I'm releasing this morning, thanks for your patience

PR #2512, opened 2020-08-11
refactor: remove legacy cursor construction

mbroadst:
We no longer need to support construction of cursors without an operation, so we can remove the logic in the constructor to check for whether one was passed in or not.

NODE-2763

### DIFF
--- a/src/cursor/core_cursor.ts
+++ b/src/cursor/core_cursor.ts
@@ -206,8 +206,8 @@ class CoreCursor extends Readable {
   }
 
   /** Clone the cursor */
-  clone(): CoreCursor {
-    return new CoreCursor(this.topology, this.operation, this.options);
+  clone(): this {
+    return new (this.constructor as any)(this.topology, this.operation, this.options);
   }
 
   /**
--- a/src/cursor/core_cursor.ts
+++ b/src/cursor/core_cursor.ts
@@ -1,12 +1,12 @@
 import Logger = require('../logger');
 import { ReadPreference } from '../read_preference';
-import { handleCallback, collationNotSupported, MongoDBNamespace } from '../utils';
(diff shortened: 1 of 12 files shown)


src/cursor/core_cursor.ts
```diff
@@ -218,14 +205,9 @@ class CoreCursor extends Readable {
...
-  /**
-   * Clone the cursor
-   *
-   * @function
-   * @returns {CoreCursor}
-   */
+  /** Clone the cursor */
   clone(): CoreCursor {
-    return this.topology.cursor(this.ns, this.cmd, this.options);
+    return new CoreCursor(this.topology, this.operation, this.options);
```
nbbeeken, line 210: topology.cursor makes use of the cursorFactory option, is there any concern about that being dropped here?

src/cursor/cursor.ts
```diff
@@ -1068,6 +1061,11 @@ class Cursor extends CoreCursor {
   getLogger(): any {
     return this.logger;
   }
+
+  /** Clone the cursor */
+  clone(): Cursor {
+    return new Cursor(this.topology, this.operation, this.options);
```
nbbeeken, line 1067: I see CommandCursor, AggregationCursor, and ChangeStreamCursor don't have clone overrides, I think this should be only implemented in CoreCursor, and take the form:
```ts
    clone() {
        return new this.constructor(/* ... */);
    }
```
I'm not sure what the return type would look like though

mbroadst: type is `clone(): this`, but there's no way to dynamically call `this.constructor` without casting to `any` or other similar tricks. I modified it to do just that for now, we can revisit the decision when we turn linting on.

PR #2514, opened 2020-08-12
NODE-2732: type annotations for GridFS

reggi:


src/gridfs-stream/upload.ts
```diff
@@ -118,7 +144,7 @@ export class GridFSBucketWriteStream extends Writable {
    * @returns {Promise<void>} if no callback specified
    */
 
-  abort(callback: Callback) {
+  abort(callback: Callback<undefined>): Promise<void> | void {
```
reggi, line 147: This should really use `maybePromise`

```suggestion
abort(): Promise<DeleteResult>;
  abort(callback: Callback<DeleteResult | undefined>): void;
  abort(callback?: Callback<DeleteResult | undefined>): Promise<DeleteResult> | void {
    return maybePromise(callback, callback => {
      if (this.state.streamEnd) {
        const error = new Error('Cannot abort a stream that has already completed');
        return callback(error);
      }
      if (this.state.aborted) {
        const error = new Error('Cannot call abort() on a stream twice');
        return callback(error);
      }
      this.state.aborted = true;
      return this.chunks.deleteMany({ files_id: this.id }, callback);
    });
  }
```

src/gridfs-stream/download.ts
```diff
@@ -262,7 +314,7 @@ function init(self: any) {
     findOneOptions.skip = self.s.options.skip;
   }
 
-  self.s.files.findOne(self.s.filter, findOneOptions, (error?: any, doc?: any) => {
+  self.s.files.findOne(self.s.filter, findOneOptions, (error?: AnyError, doc?: GridFsFile) => {
```
nbbeeken, line 317: It would be nice to drop the annotations from the callback.
Maybe we should go in and fix the Collections type to have the deprecate functions as class properties.
Or we should update the Collection interface to be better typed.

src/gridfs-stream/download.ts
```diff
@@ -372,9 +428,15 @@ function handleStartOption(stream: any, doc: any, options: any) {
 
     return options.start - stream.s.bytesRead;
   }
+  throw new Error('no option available');
```
nbbeeken, line 431: I think we should use MongoError here and in general we should try to use only MongoError, we want to eventually have that be our only error type or subclasses of it. Also [nit] maybe `'No start option defined'` slightly more clear message.

src/gridfs-stream/download.ts
```diff
@@ -138,120 +188,122 @@ export class GridFSBucketReadStream extends Readable {
...
       }
       callback && callback();
     }
   }
 }
 
-function throwIfInitialized(self: any) {
+function throwIfInitialized(self: GridFSBucketReadStream) {
   if (self.s.init) {
     throw new Error('You cannot change options after the stream has entered' + 'flowing mode!');
```
nbbeeken, line 212: Not your code I know, but looks like this doesn't need concatenation anymore maybe there used to be a var in between

src/gridfs-stream/download.ts
```diff
@@ -1,5 +1,54 @@
...
+export interface GridFSBucketReadStreamOptions {
+  sort?: number | { uploadDate: number };
+  skip?: number;
+  /** 0-based offset in bytes to start streaming from */
+  start?: number;
+  /** 0-based offset in bytes to stop streaming before */
+  end?: number;
+}
+
+export interface GridFsFile {
```
mbroadst, line 18:
```suggestion
export interface GridFSFile {
```
or change all the other references to be `GridFs`, but let's be consistent

src/gridfs-stream/upload.ts
```diff
@@ -151,48 +177,53 @@ export class GridFSBucketWriteStream extends Writable {
...
 
-  end(chunk: any, encoding?: any, callback?: Callback) {
-    var _this = this;
-    if (typeof chunk === 'function') {
-      (callback = chunk), (chunk = null), (encoding = null);
-    } else if (typeof encoding === 'function') {
-      (callback = encoding), (encoding = null);
-    }
+  end(
+    chunkOrCallback?: Buffer | Callback<GridFsFile | undefined>,
```
mbroadst, line 181: `Callback<GridFsFile>` is equivalent to `Callback<GridFsFile | undefined>`

src/gridfs-stream/upload.ts
```diff
@@ -375,19 +414,24 @@ function createFilesDoc(
   return ret;
 }
 
-function doWrite(_this: any, chunk: any, encoding: any, callback?: Callback) {
-  if (checkAborted(_this, callback)) {
+function doWrite(
+  self: GridFSBucketWriteStream,
+  chunk: Buffer,
+  encoding: BufferEncoding | undefined,
```
mbroadst, line 420:
```suggestion
  encoding?: BufferEncoding,
```

src/gridfs-stream/upload.ts
```diff
@@ -448,61 +492,65 @@ function doWrite(_this: any, chunk: any, encoding: any, callback?: Callback) {
...
 
-function getWriteOptions(_this: any) {
-  var obj = {} as any;
-  if (_this.options.writeConcern) {
-    obj.w = _this.options.writeConcern.w;
-    obj.wtimeout = _this.options.writeConcern.wtimeout;
-    obj.j = _this.options.writeConcern.j;
+function getWriteOptions(self: GridFSBucketWriteStream) {
+  var obj: WriteConcernOptions = {};
+  if (self.bucket.s.options.writeConcern) {
```
mbroadst, line 497: hm are you sure this didn't break anything? Are they indeed the same `options`?

src/collection.ts
```diff
@@ -1430,7 +1460,6 @@ Collection.prototype.update = deprecate(function (
 },
 'collection.update is deprecated. Use updateOne, updateMany, or bulkWrite instead.');
 
-Collection.prototype.removeOne = Collection.prototype.deleteOne;
 Collection.prototype.removeMany = Collection.prototype.deleteMany;
```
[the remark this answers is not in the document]
reggi: I only made changed I needed to for GridFS to work.

src/gridfs-stream/download.ts
mbroadst: I merged these methods into the main class yesterday, you should be able to use it without annotations here now

src/gridfs-stream/download.ts
```diff
@@ -1,5 +1,54 @@
...
+  _id: GridFSBucketWriteStream['id'];
+  length: GridFSBucketWriteStream['length'];
+  chunkSize: GridFSBucketWriteStream['chunkSizeBytes'];
+  md5?: boolean | string;
+  filename: GridFSBucketWriteStream['filename'];
+  contentType?: GridFSBucketWriteStream['options']['contentType'];
+  aliases?: GridFSBucketWriteStream['options']['aliases'];
+  metadata?: GridFSBucketWriteStream['options']['metadata'];
+  uploadDate: Date;
+}
```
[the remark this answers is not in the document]
reggi: There's a function that maps properties of a `GridFSBucketWriteStream` to a file.  The same way that function maps, I created the interface to do the same. Do you have a recommendation on how you would like this? I think this is the best option.

src/gridfs-stream/index.ts
```diff
@@ -1,53 +1,65 @@
...
+  bucketName?: string;
+  /** Number of bytes stored in each chunk. Defaults to 255KB */
+  chunkSizeBytes?: number;
+  /** Read preference to be passed to read operations */
+  readPreference?: ReadPreference;
+}
+
+interface GridFSBucketPrivate {
+  db: Db;
+  options: {
```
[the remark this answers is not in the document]
reggi: This was deliberate. The internal options are non-optional because we apply defaults. Ideally properties like `bucketName` would be `this.bucketName`, same for `writeConcern` is also internally narrower, again can be `this.writeConcern`.  Let me know what your preference is here.

src/gridfs-stream/index.ts
```diff
@@ -68,87 +80,34 @@ export class GridFSBucket extends EventEmitter {
    * Returns a writable stream (GridFSBucketWriteStream) for writing
    * buffers to GridFS. The stream's 'id' property contains the resulting
    * file's id.
-   *
```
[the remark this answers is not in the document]
reggi: Little confused about what you're expecting here, this?

```
   * @param filename - The value of the 'filename' key in the files doc
   * @param options - Optional settings.
```

All of the options are documented in the `GridFSBucketWriteStreamOptions` interface.

src/gridfs-stream/index.ts
```diff
@@ -68,87 +80,34 @@ export class GridFSBucket extends EventEmitter {
...
-    }
-
-    if (!options.chunkSizeBytes) {
-      options.chunkSizeBytes = this.s.options.chunkSizeBytes;
-    }
-
-    options.id = id;
-
-    return new GridFSBucketWriteStream(this, filename, options);
+  openUploadStreamWithId(
```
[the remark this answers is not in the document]
reggi: Why would `openUploadStreamWithId` have overloads? There's no difference in the return signature here, it's always returning a `GridFSBucketWriteStream` instance.

src/gridfs-stream/upload.ts
```diff
@@ -104,9 +125,14 @@ export class GridFSBucketWriteStream extends Writable {
    * @returns {boolean} False if this write required flushing a chunk to MongoDB. True otherwise.
    */
 
-  write(chunk: any, encoding: any, callback?: Callback): boolean {
-    var _this = this;
-    return waitForIndexes(this, () => doWrite(_this, chunk, encoding, callback));
+  write(
```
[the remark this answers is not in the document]
reggi: Same as the comment about overloads above. Do we need to add overloads if something always returns the same type? Please clarify what the idea is here.

src/gridfs-stream/upload.ts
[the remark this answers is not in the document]
reggi: This one is valid 👍

src/gridfs-stream/upload.ts
reggi: Updated to `void` but still needs this union for compatibility with `checkAborted`.

src/gridfs-stream/upload.ts
reggi: I don't think inheritance was happening correctly here. I added this to the class.

```
this.writeConcern = WriteConcern.fromOptions(options) || bucket.s.options.writeConcern;
```

src/gridfs-stream/upload.ts
```diff
@@ -1,11 +1,43 @@
...
+import type { Document } from './../types.d';
+import type { GridFSBucket } from './index';
+import type { GridFSFile } from './download';
+import type { WriteConcernOptions } from '../write_concern';
+import { MongoError } from '../error';
+import type { ObjectID } from 'bson';
 
 const ERROR_NAMESPACE_NOT_FOUND = 26;
 
+export type TFileId = string | number | object | ObjectId;
```
nbbeeken, line 17: Is there a convention for the leading `T` here?

reggi: @mbroadst [referenced this spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#user-content-terms:~:text=While,-GridFS) 

I'd be down to name it `GridFSFileId`, but I was assuming we'd need this in other places.

src/gridfs-stream/download.ts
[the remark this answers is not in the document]
mbroadst: SGTM, just wanted to better understand the intention here

src/gridfs-stream/index.ts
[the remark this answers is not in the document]
mbroadst: yep, that's what I was expecting

src/gridfs-stream/upload.ts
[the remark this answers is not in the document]
mbroadst: The one above was fine because the last parameter was just an optional. In this case we need overloads because we want to guide the user's completion:
```
write(chunk: Buffer, callback: Callback<void>);
write(chunk: Buffer, encoding: BufferEncoding, callback: Callback<void>);
.. so on ..
```
Overloading isn't just about the return type

src/gridfs-stream/upload.ts
mbroadst: @nbbeeken there is a convention for leading `T` on template parameters, so maybe that's where this originally came from. I don't have a strong preference either way

src/gridfs-stream/upload.ts
```diff
@@ -151,48 +189,57 @@ export class GridFSBucketWriteStream extends Writable {
...
    */
 
-  end(chunk: any, encoding?: any, callback?: Callback) {
-    var _this = this;
-    if (typeof chunk === 'function') {
-      (callback = chunk), (chunk = null), (encoding = null);
-    } else if (typeof encoding === 'function') {
-      (callback = encoding), (encoding = null);
-    }
+  end(
```
mbroadst, line 192: This has missing overloads as well

PR #2516, opened 2020-08-13
NODE-1449: Generate Type Definitions

nbbeeken:
This PR isn't ready to merge but I figured it would be easier for us to discuss next steps with something up here to comment on. So the src/bson.ts file is currently a blocker so you'll see I just changed it to not import bson-ext for now to get definition generating working. I started to delete jsdoc relating things but thought it best to keep this diff small. I committed the mongodb.d.ts file so we can all take a look but I assume we will keep that only in the published npm package?
There's the other issue of tsdoc syntax, it is (by design) a very strict language of its own to avoid ambiguity that sometimes happens with jsdoc, anyway result is we have many "syntax" warnings, none that halt doc generation at least.

If you wanna checkout how things look locally the gulp script is a good place to start, I made the following tasks:

- `gulp compile` - I temporarily switched to just using tsc to compile over our gulp script because source maps aren't correctly piped through the newline preserve and prettier, we can workshop this.
- `gulp definition` - This uses compile to output all the .d.ts per file and then runs api-extractor to do the bundling work, and then deletes all the .d.ts per file in lib. Also I run prettier on the single output mongodb.d.ts
- `gulp doc` - This will generate the typedoc powered documentation. It puts everything in `docs/gen` for now (index.html is will be in there so that's where you can start) In order to get the tutorial markdown pages I use a plugin called `typedoc-plugin-pages` and each page has to be specified manually so gulp generates the typedoc.json config runs typedoc and deletes the config because do we really need an n-th config file? 

Take it for a spin let me know what you think.


reggi:
I'm getting some yellow warnings here are a couple of them:

> tsdoc-undefined-tag: The TSDoc tag "@function" is not defined in this configurationeslinttsdoc/syntax
> tsdoc-param-tag-with-invalid-type: The @param block should not include a JSDoc-style '{type}'eslinttsdoc/syntax
> tsdoc-param-tag-missing-hyphen: The @param block should be followed by a parameter name and then a hypheneslinttsdoc/syntax

reggi:
I'm getting some yellow warnings here are a couple of them:

> tsdoc-undefined-tag: The TSDoc tag "@function" is not defined in this configurationeslinttsdoc/syntax
> tsdoc-param-tag-with-invalid-type: The @param block should not include a JSDoc-style '{type}'eslinttsdoc/syntax
> tsdoc-param-tag-missing-hyphen: The @param block should be followed by a parameter name and then a hypheneslinttsdoc/syntax

nbbeeken:
> This is so cool!
> 
> I'm getting some yellow warnings here are a couple of them:
> //... 
> May need to add in extra rules. Some of these are valid though, where there's a mismatch in typings.

Yea as I said there's many warnings luckily they don't halt documentation generation but the warnings are all valid in that that formatting come out a bit wonky in a majority of cases. For example `@function`s should be dropped everywhere, they're meaningless to tsdoc, the `@param`s need to be corrected to not have a type annotation at all and have a hyphen after the parameter name etc.

So basically by using tsdoc we aren't using jsdoc anymore they are compatible but divergent formats, the cost of good docs. Are we okay with doing another sweeping change to all the doc comments in the code base again? Do we want to do this over time? 
We could at least fix all the params pretty easily with a regex, drop types and add the hyphen.

nbbeeken:
@mbroadst @reggi @emadum Just pinging, I made the docs changes to make us tsdoc compliant everywhere but in the bulk subdirectory.
Working through the pile of warnings unveiled that API-extractor has separate warnings than the ts-doc linter, nothing major its just the type symbols we have made need to be reexported from `src/index.ts` I can do that and have that updated soon

src/cursor/cursor.ts
```diff
@@ -634,11 +624,11 @@ export class Cursor<
   }
 
   /**
-   * @deprecated
-   * Iterates over all the documents for this cursor. As with **{cursor.toArray}**,
+   * @deprecated Please use {@link Cursor.forEach} instead
```
reggi, line 627: I don't think this is formatted correctly because it starts with the tag, all of this comment is the value of the tag. I think @deprecated needs to be below the comment.

reggi: This might also apply to `@internal` i've seen start the doc.

nbbeeken: You're correct about deprecated, I'll got through and make sure it doesn't come before the summary, internal/public are different they can come anywhere since they're modifier style tags and don't collect text after them.

nbbeeken: I think however we should probably have the summary lead the doc comment, so I've changed the access modifiers position to be below the summary

src/apm.ts
```diff
@@ -1,9 +1,22 @@
 import { EventEmitter } from 'events';
 import type { Callback } from './utils';
+import { Connection } from './cmap/connection';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mbroadst, line 5: what's up with this comment?

src/apm.ts
```diff
@@ -1,9 +1,22 @@
 import { EventEmitter } from 'events';
 import type { Callback } from './utils';
+import { Connection } from './cmap/connection';
```
mbroadst, line 3: shouldn't this be an `import type`?

src/apm.ts
nbbeeken: I'm not entirely sure, we have a two other cases where this comes up
https://github.com/mongodb/node-mongodb-native/blob/ca16d554a6d40cb3cbf35164ed5c94de10537bc3/src/operations/collections.ts#L6-L8
https://github.com/mongodb/node-mongodb-native/blob/ca16d554a6d40cb3cbf35164ed5c94de10537bc3/src/operations/create_collection.ts#L10-L12
Despite the type being used its marked as unused for some reason

src/apm.ts
nbbeeken: No I'm using the static properties defined on it so I need the value.

src/apm.ts
mbroadst: oh I think this was from before we upgraded prettier, you can probably just remove the lines everywhere

nbbeeken: Oh yea you are right those were fine to remove, but this apm one still reports unused var despite it being used. Maybe because it is being used as a type annotation for a `this` on the function in instrument, which is kind of a fake param to type your function's this binding, not sure though

PR #2518, opened 2020-08-14
Can we ignore "src" on 3.6 branch?

reggi:


reggi:
I keep seeing a whole bunch of files when I switch between branches, not sure if this is why.

PR #2519, opened 2020-08-16
docs: fix JSDocs for GridFSBucketReadStream constructor

HitkoDev:
## Description

It's clear from https://github.com/mongodb/node-mongodb-native/blob/5808e12f13e39bfddb26cc15fe9b4dd751e1f363/src/gridfs-stream/index.ts#L227 and https://github.com/mongodb/node-mongodb-native/blob/5808e12f13e39bfddb26cc15fe9b4dd751e1f363/src/gridfs-stream/download.ts#L258 that `GridFSBucketReadStream` `options.sort` accepts object or array rather than a number, just like https://github.com/mongodb/node-mongodb-native/blob/4aeaedf3b12a196127852522bcc81d8b507676d5/src/collection.ts#L1828

### DIFF
--- a/src/gridfs-stream/download.ts
+++ b/src/gridfs-stream/download.ts
@@ -12,7 +12,7 @@ import { Readable } from 'stream';
  * @param {object} readPreference The read preference to use
  * @param {object} filter The query to use to find the file document
  * @param {object} [options] Optional settings.
- * @param {number} [options.sort] Optional sort for the file find query
+ * @param {(array|object)} [options.sort] Optional sort for the file find query. Array of indexes, [['a', 1]] etc.
  * @param {number} [options.skip] Optional skip for the file find query
  * @param {number} [options.start] Optional 0-based offset in bytes to start streaming from
  * @param {number} [options.end] Optional 0-based offset in bytes to stop streaming before


reggi:
Hey @HitkoDev! Nice catch!

We're in the process of converting things over to typescript, master isn't fully typed yet but we're working hard to get it in a better place. I just checked both `collection.findOne` and the `GridFSBucketReadStream` and it will be typed natively soon! 🍰🐬💕

Sort will be typed as:

```ts
/** @public */
export type SortDirection = 1 | -1 | 'asc' | 'desc' | { $meta: string };
/** @public */
export type Sort =
  | { [key: string]: SortDirection }
  | [string, SortDirection][]
  | [string, SortDirection];
```

And the `GridFSBucketReadStream` will use an interface `GridFSBucketReadStreamOptions` with the `sort?: Sort` as one of the properties.

```ts
import { GridFSBucket } from './gridfs-stream/index';
import * as mongo from './index';
const { MongoClient } = mongo;

const client = new MongoClient('uri');

(async () => {
  await client.connect();

  const db = client.db('example');

  const c = db.collection('meow');

  const x = await c.findOne({}, { sort: { updateOne: -1 } });

  const bucket = new GridFSBucket(db, { bucketName: 'gridfsdestroy', chunkSizeBytes: 10 });

  const stream = bucket.openDownloadStream('12345', { sort: { updateOne: -1 } });

})();
```

As a result we're gotten rid of some unnecessary JSDoc comments, and this one will be removed in the next release.

reggi:
@HitkoDev I'm going to close this PR for now, as we've implemented the types here for `sort` in master.

PR #2521, opened 2020-08-18
Update crud.md

rosald:
here should be createCollection

## Description

**What changed?**
createIndex -> createCollection


mbroadst:
Thanks for the contribution @rosald ! I merged your fix in, but our tutorials and reference have officially moved "on-domain" at https://docs.mongodb.com/drivers/node

PR #2522, opened 2020-08-18
Node 2732/bulk

reggi:


src/bulk/common.ts
```diff
@@ -1,43 +1,198 @@
...
+  insertOne: InsertOneModel;
+  insertMany: InsertManyModel;
+  replaceOne: ReplaceOneModel;
+  updateOne: UpdateOneModel;
+  updateMany: UpdateManyModel;
+  removeOne: RemoveOneModel;
+  removeMany: RemoveManyModel;
+  deleteOne: DeleteOneModel;
+  deleteMany: DeleteManyModel;
+}
```
nbbeeken, line 118: These could all be made `?:` style optional keys I think

src/bulk/common.ts
```diff
@@ -550,17 +580,21 @@ function mergeBatchResults(batch: any, bulkResult: any, err: any, result: any) {
...
+  bulkOperation: BulkOperationBase,
+  options: BulkOptions,
+  callback: Callback<BulkWriteResult>
+) {
   if (bulkOperation.s.batches.length === 0) {
     return callback(undefined, new BulkWriteResult(bulkOperation.s.bulkResult));
   }
 
-  const batch = bulkOperation.s.batches.shift();
+  const batch = bulkOperation.s.batches.shift() as Batch;
```
nbbeeken, line 597: One day I hope TS can use length as guard on shift and pop

src/bulk/common.ts
```diff
@@ -641,205 +663,202 @@ class BulkWriteError extends MongoError {
...
  * Is used to build a write operation that involves a query filter.
  */
 class FindOperators {
-  s: any;
+  s: {
+    currentOp?: Partial<Operation>;
+    options: {
+      addToOperationsList: (self: FindOperators, batchType: BatchTypes, document: Document) => any;
+    };
+  };
```
nbbeeken, line 683: Should we pull this out to a `FindOperatorsPrivate` interface?

src/bulk/common.ts
```diff
@@ -641,205 +663,202 @@ class BulkWriteError extends MongoError {
...
 
   /**
+   * @internal
    * Creates a new FindOperators object.
-   *
-   * **NOTE:** Internal Type, do not instantiate directly
-   *
-   * @param bulkOperation
    */
   constructor(bulkOperation: any) {
```
nbbeeken, line 689: Should this be `BulkOperationBase`? or the `T extends BulkOperationBase` trick?

src/bulk/common.ts
```diff
@@ -641,205 +663,202 @@ class BulkWriteError extends MongoError {
...
+      ...(updateDocument.hint ? { hint: updateDocument.hint } : {})
+    };
 
     if (!hasAtomicOperators(updateDocument)) {
       throw new TypeError('Update document requires atomic operators');
     }
 
-    // Clear out current Op
-    this.s.currentOp = null;
+    this.clearOutCurrentOp();
```
nbbeeken, line 733: Was this function added for readability? if so, my personal nit, maybe just `clearCurrentOp`?

src/bulk/common.ts
```diff
@@ -641,205 +663,202 @@ class BulkWriteError extends MongoError {
...
+  // Options
+  options: BulkOptions;
+  // Current operation
+  currentOp: { selector: Document } | null;
+  // Executed
+  executed: boolean;
+  // Collection
+  collection: Collection;
+  // Fundamental error
+  err: null;
```
nbbeeken, line 832: AnyError?

src/bulk/common.ts
```diff
@@ -641,205 +663,202 @@ class BulkWriteError extends MongoError {
...
+  bulkResult: BulkResult;
+  currentBatch: Batch | null;
+  currentIndex: number;
+  // ordered specific
+  currentBatchSize: number;
+  currentBatchSizeBytes: number;
+  // unordered specific
+  currentInsertBatch: Batch | null;
+  currentUpdateBatch: Batch | null;
+  currentRemoveBatch: Batch | null;
```
nbbeeken, line 810: I think these should be optional instead of `| null`

src/bulk/common.ts
```diff
@@ -938,38 +957,39 @@ class BulkOperationBase {
...
    * Builds a find operation for an update/updateOne/delete/deleteOne/replaceOne.
    * Returns a builder object used to complete the definition of the operation.
    *
-   * @function
-   * @param selector The selector for the bulk operation. See {@link https://docs.mongodb.com/manual/reference/command/update/#update-command-q q documentation}
-   * @throws {MongoError} if a selector is not specified
-   * @returns {FindOperators} A helper object with which the write operation can be defined.
-   *
    * @example
+   * ```ts
```
nbbeeken, line 992: [super nit] I'd keep this as js for our examples, this nit is made even more super by the class being internal 😅  so completely up to you

src/bulk/common.ts
```diff
@@ -1011,68 +1032,57 @@ class BulkOperationBase {
...
-   *
-   * @function
-   * @param op The raw operation to perform.
-   * @param reference for more information.
-   * @returns {BulkOperationBase} A reference to self
-   */
-  raw(op: any): BulkOperationBase {
-    const key = Object.keys(op)[0];
-
+  raw(op: Partial<BatchOperations> & { collation?: CollationOptions }): BulkOperationBase {
```
nbbeeken, line 1035: Does an optional collation not fit (as in cause type issues?) on the BatchOperations interface?

src/bulk/common.ts
```diff
@@ -1,43 +1,198 @@
...
+  ReplaceOneModel &
+  UpdateOneModel &
+  UpdateManyModel &
+  RemoveOneModel &
+  RemoveManyModel &
+  DeleteOneModel &
+  DeleteManyModel;
+
+/** @public */
+export type AnyOperation =
```
nbbeeken, line 132: [naming nit] `AnyBulkOperation`?

src/bulk/common.ts
```diff
@@ -1,43 +1,198 @@
...
+  | { replaceOne: ReplaceOneModel }
+  | { updateOne: UpdateOneModel }
+  | { updateMany: UpdateManyModel }
+  | { removeOne: RemoveOneModel }
+  | { removeMany: RemoveManyModel }
+  | { deleteOne: DeleteOneModel }
+  | { deleteMany: DeleteManyModel };
+
+/** @public */
+export type Operations = AnyOperation[];
```
nbbeeken, line 144: I think we can drop this type in favor of just using `AnyBulkOperation[]` where this is used, I think this type used to be more complex maybe?

src/bulk/common.ts
```diff
@@ -1,43 +1,198 @@
...
+  updateOne: UpdateOneModel;
+  updateMany: UpdateManyModel;
+  removeOne: RemoveOneModel;
+  removeMany: RemoveManyModel;
+  deleteOne: DeleteOneModel;
+  deleteMany: DeleteManyModel;
+}
+
+/** @public */
+export type AnyModel = InsertOneModel &
```
nbbeeken, line 121: [naming nit] `AnyWriteModel`? If this was an interface instead that extended from each of these other models is there an advantage to that? (just wondering) like maybe making it extensible for others? maybe that's not a concern

src/bulk/common.ts
```diff
@@ -1,43 +1,198 @@
...
-const REMOVE = 3;
+export const INSERT = 1;
+export const UPDATE = 2;
+export const REMOVE = 3;
+
+/** Number indicating the document type */
+export type BatchTypes = typeof INSERT | typeof UPDATE | typeof REMOVE;
+
+/** @public */
+export interface Operation {
```
nbbeeken, line 52: [naming nit] It's not named the same as OperationBase so that's not an issue but it might be confusing that its generically named but scoped to dealing with bulk api stuff, BulkOperation though could be confused with BulkOperationBase? but maybe that's a preferential confusion 😆

src/bulk/common.ts
```diff
@@ -550,17 +580,21 @@ function mergeBatchResults(batch: any, bulkResult: any, err: any, result: any) {
...
+  callback: Callback<BulkWriteResult>
+) {
   if (bulkOperation.s.batches.length === 0) {
     return callback(undefined, new BulkWriteResult(bulkOperation.s.bulkResult));
   }
 
-  const batch = bulkOperation.s.batches.shift();
+  const batch = bulkOperation.s.batches.shift() as Batch;
 
   function resultHandler(err?: any, result?: any) {
```
nbbeeken, line 599: can we get argument types here?

src/bulk/common.ts
```diff
@@ -590,17 +624,10 @@ function executeCommands(bulkOperation: any, options: any, callback: Callback<Bu
...
- * @param err
- * @param callback
- */
+/** handles write concern error */
 function handleMongoWriteConcernError(
-  batch: object,
-  bulkResult: object,
+  batch: Batch,
+  bulkResult: BulkResult,
   err: any,
```
nbbeeken, line 631: BulkWriteError?

src/bulk/common.ts
```diff
@@ -621,15 +648,10 @@ function handleMongoWriteConcernError(
  * @public
  * @category Error
  */
-class BulkWriteError extends MongoError {
+export class BulkWriteError extends MongoError {
   result: any;
```
nbbeeken, line 652: `BulkWriteResult`?

src/bulk/common.ts
```diff
@@ -1308,15 +1284,10 @@ class BulkOperationBase {
...
    * Handles the write error before executing commands
-   *
-   * An internal helper method. Do not invoke directly. Will be going away in the future
-   *
-   * @param callback
-   * @param writeResult
-   * @returns {boolean|undefined}
    */
-  handleWriteError(callback: Callback, writeResult: any): boolean | undefined {
+  handleWriteError(callback: Callback<any>, writeResult: BulkWriteResult): boolean | undefined {
```
nbbeeken, line 1290: This `<any>` is redundant, was there an actual type here at some point?

src/bulk/common.ts
```diff
@@ -1143,14 +1148,7 @@ class BulkOperationBase {
...
-   *
-   * @function
-   * @param _writeConcern
-   * @param options
-   * @param callback
-   */
+  /** An internal helper method. Do not invoke directly. Will be going away in the future */
   bulkExecute(
     _writeConcern?: WriteConcern,
     options?: object,
```
nbbeeken, line 1154: Do we have an interface for this options?

src/bulk/common.ts
reggi: saw you did some comments in `js` guess it's ok unless it's really `ts`

src/bulk/common.ts
reggi: I need consensus on this before I start changing names.

src/bulk/common.ts
reggi: I'm down for this, will implement direct array use and  wait on consensus for naming.

src/bulk/common.ts
reggi: Totally down, Need consensus

src/bulk/common.ts
reggi: Yes agree on this one, too generic, I'm not sure if this thing is already defined and can just be `AnyModel`? Not sure. Down for suggestions.

src/bulk/common.ts
reggi: This is where `AnyError` doesn't work because of that darn `.driver` and `.ok` check. I can type `result`. If you have an ideas about `err` I'm all 👂🌽👂.

src/bulk/common.ts
reggi: I fixed this but also realized we needed to make some changes to `isPromiseLike`.

reggi: Trying to mess with `isPromiseLike` it's still not correct.

src/bulk/common.ts
```diff
@@ -477,45 +499,45 @@ function mergeBatchResults(batch: any, bulkResult: any, err: any, result: any) {
...
-  if (batch.batchType === REMOVE && result.n) {
+  if (batch.batchType === BatchType.REMOVE && result?.n) {
     bulkResult.nRemoved = bulkResult.nRemoved + result.n;
   }
 
   let nUpserted = 0;
 
   // We have an array of upserted values, we need to rewrite the indexes
-  if (Array.isArray(result.upserted)) {
+  if (result && Array.isArray(result?.upserted)) {
```
mbroadst, line 531: no need for the `result?` if you check the existence of `result` here (also below). In fact, it looks like maybe we should just check for the existence of `result` above one time, rather than guarding it all over the place below?

src/bulk/common.ts
```diff
@@ -1,43 +1,195 @@
...
+  q?: Document;
+}
+
+/** @public */
+export interface DeleteManyModel extends RemoveOptions {
+  q?: Document;
+}
+
+/** @public */
+export interface BatchOperations {
```
mbroadst, line 107: I thought we discussed not using a partial approach like this, but instead using a type hierarchy to represent the models:
```
interface WriteModel {}
interface InsertModel extends WriteModel {}
interface UpdateModel extends WriteModel {}
type AnyWriteModel = InsertModel | UpdateModel | ...;
class BulkOperationBase {
  ...
  insert(model: InsertModel);
  update(model: UpdateModel);
  raw(op: AnyWriteModel);
  ...
}
```

Was there something that was preventing this approach technically?

src/bulk/common.ts
reggi: Cool yeah, I adopted this here, but below it's a bit more complex below because of the errors.

src/bulk/common.ts
reggi: I removed the .driver and cast .ok

src/bulk/common.ts
reggi: `BulkOperationBase` 👍

src/bulk/common.ts
```diff
@@ -432,9 +487,9 @@ function mergeBatchResults(batch: any, bulkResult: any, err: any, result: any) {
   } else if (result == null) {
     return;
   }
-
+  // console.log({ result, bulkResult });
```
nbbeeken, line 490: [nit] left in by mistake?

nbbeeken:
All of the eslint warnings should be fixed with this PR if you could change the `check:lint` script to allow 0 warnings that'd be grand 🚀   Also, I'm getting a compile error on my end:
```
src/cmap/wire_protocol/command.ts:109:49 - error TS2345: Argument of type 'CommandOptions' is not assignable to parameter of type 'WriteCommandOptions'.
  Types of property 'writeConcern' are incompatible.
    Type 'number | WriteConcernOptions | "majority" | WriteConcern | undefined' is not assignable to type 'WriteConcern | undefined'.
      Type 'number' is not assignable to type 'WriteConcern | undefined'.

109     const err = applySession(session, finalCmd, options);
```
Even if the WriteCommand solution isn't ideal I think we should try to merge in a temp fix so we can get master compiling.

mbroadst:
superseded by 9715c00f54a69041e4f429fc65694e602250085a

PR #2529, opened 2020-08-25
fix: SASL's signature consists of the string representation of the payload, not base64

marcj:
`payload.value()` already returns a string, so calling on it `toString('base64')`
implies the signature consists of a base64 representation, but the plain string is required.
The first argument ('base64') of String.toString is dropped, so although the code works
as expected (by accident) it implies that it was desired to have a base64 there, which is wrong.

Dropped the toString() call to not confuse driver authors when they want to get inspiration
from this JS driver.

### DIFF
--- a/src/cmap/auth/scram.ts
+++ b/src/cmap/auth/scram.ts
@@ -181,11 +181,9 @@ function continueScramConversation(
   const clientKey = HMAC(cryptoMethod, saltedPassword, 'Client Key');
   const serverKey = HMAC(cryptoMethod, saltedPassword, 'Server Key');
   const storedKey = H(cryptoMethod, clientKey);
-  const authMessage = [
-    clientFirstMessageBare(username, nonce),
-    payload.value().toString('base64'),
-    withoutProof
-  ].join(',');
+  const authMessage = [clientFirstMessageBare(username, nonce), payload.value(), withoutProof].join(
+    ','
+  );
 
   const clientSignature = HMAC(cryptoMethod, storedKey, authMessage);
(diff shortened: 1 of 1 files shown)


agolin95:
Tracked by https://jira.mongodb.org/browse/NODE-2893

nbbeeken:
Hi @marcj thanks for your contribution, it looks good to me and I'll be happy to merge but there's a some linter errors, you can fix them with:
```sh
npm run check:lint -- --fix
```

nbbeeken:
Hi @marcj if you get the chance to run the linter command I can go ahead and merge this, thanks again for contributing!

marcj:
@nbbeeken I get `✖ 258 problems (0 errors, 258 warnings)`. Which linter errors do you mean? I actually changed only one line, which I don't think generates any error.

nbbeeken:
I'm getting this error:
```
src/cmap/auth/scram.ts
184:24  error  Replace `⏎····clientFirstMessageBare(username,·nonce),⏎····payload.value(),⏎····withoutProof⏎··].join(','` with `clientFirstMessageBare(username,·nonce),·payload.value(),·withoutProof].join(⏎····','⏎··`  prettier/prettier
✖ 259 problems (1 error, 258 warnings)
```
🤔  Not sure why the prettier errors wouldn't show up on your end, did your editor auto format and the changes are still local?

marcj:
Oh, no it's because of `--fix`. It just changes the line without complaining afterwards. After the command I have the change locally. Just pushed the modified commit.

PR #2533, opened 2020-08-26
NODE-2742: adds label check to tests, fixes insertOne test where retryWrites is false

reggi:
https://jira.mongodb.org/browse/NODE-2742


test/functional/retryable_writes.test.js
```diff
@@ -94,6 +93,12 @@ function executeScenarioTest(test, ctx) {
             expect(err.message, 'expected operations to fail, but they succeeded').to.not.match(
               /expected false to be true/
             );
+            const errorLabelsContain =
```
mbroadst, line 96: I think this is one piece of the puzzle, but the ticket explicitly calls out the bug of us treating the presence of `outcome.error` and `outcome.result` as mutually exclusive. I think that entails just removing the `else` from L103 and any errors that ensue

reggi: I don't believe it makes sense to remove the check. The promise will still split this logic up. 

I added a check to check if the `result` has the label, and if not, if it's a true result, check `err.result`, this is for bulk writes. 

This change will result in test failures, there is no `err.result` it's a `MongoNetworkError`, i've been looking through the codebase to find where I would be able to change this to a Bulk write error. Any pointers?

mbroadst: You're right, we can't just get rid of the else for the reasons you listed above, didn't think about promises here but I think you've captured the gist of my suggestion.

Regarding the bugs you've uncovered, I was able to get around the first set of them with this patch:
```
diff --git a/src/bulk/common.ts b/src/bulk/common.ts
index 99b6fba00..335c426b8 100644
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -550,7 +550,7 @@ function executeCommands(
   function resultHandler(err?: AnyError, result?: Document) {
     // Error is a driver related error not a bulk op error, return early
     if (err && 'message' in err && !(err instanceof MongoWriteConcernError)) {
-      return callback(err);
+      return callback(new BulkWriteError(err, new BulkWriteResult(bulkOperation.s.bulkResult)));
     }
 
     if (err instanceof MongoWriteConcernError) {
diff --git a/test/functional/retryable_writes.test.js b/test/functional/retryable_writes.test.js
index 17c1df860..7daf3870c 100644
--- a/test/functional/retryable_writes.test.js
+++ b/test/functional/retryable_writes.test.js
@@ -98,7 +98,7 @@ function executeScenarioTest(test, ctx) {
               /expected false to be true/
             );
             if (hasResult) {
-              expect(err.result).to.equal(test.outcome.result);
+              expect(err.result).to.matchMongoSpec(test.outcome.result);
             }
             const errorLabelsContain =
               test.outcome && test.outcome.result && test.outcome.result.errorLabelsContain;
```

Two things to note here:
- I'm using the more versatile `matchMongoSpec` `chai` helper, which follows our relatively complicated rules for matching spec results
- The actual issue was that bulk was passing through errors that it probably should be wrapping. This is kind of a tough case that will need more consideration, because theoretically this would return something like a `TypeError` to the user as a `BulkWriteError` - which may be fine, but someone (you!) needs to sit and think about the implications of this change.

When I ran it with that patch everything passed against a single server, and there were three failures against a replicaset which seemed to be mostly in transactions code, and specifically about propagating error labels. That will bear some more investigation as well.

reggi:
This PR incorporates spec from here https://github.com/mongodb/specifications/pull/852/files which isn't merge in at the time of this comment.

mbroadst (changes requested):
LGTM, just some comments about code cleanup and some clarification

src/sdam/server.ts
```diff
@@ -573,7 +573,11 @@ function makeOperationHandler(
           session.serverSession.isDirty = true;
         }
 
-        if (supportsRetryableWrites(server) && !inActiveTransaction(session, cmd)) {
+        if (
+          (server.s.topology.s.options.retryWrites !== false || isTransactionCommand(cmd)) &&
```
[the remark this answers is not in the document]
reggi: * Im down to create an `isRetryableWritesEnabled` helper.
* Im a but concerned about your rearranging. Just want to bring this to your attention before I make the change:

[> commitTransaction is a retryable write command. Drivers MUST retry once after commitTransaction fails with a retryable error according to the Retryable Writes Specification, regardless of whether retryWrites is set on the MongoClient or not.](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#user-content-committransaction:~:text=Specification%2C%20regardless%20of%20whether%20retryWrites%20is%20set%20on-,the)

I see the `or` here between `retryWrites` and `isTransactionCommand`, and not `isTransactionCommand` and `inActiveTransaction`.

src/operations/execute_operation.ts
```diff
@@ -199,7 +199,7 @@ function executeWithServerSelection(topology: Topology, operation: any, callback
       operation.canRetryRead;
 
     const willRetryWrite =
-      topology.s.options.retryWrites === true &&
+      topology.s.options.retryWrites !== false &&
```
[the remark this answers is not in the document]
reggi: will revert

src/error.ts
```diff
@@ -227,6 +227,7 @@ function makeWriteConcernResultObject(input: any) {
  */
 export class MongoWriteConcernError extends MongoError {
   /** The result document (provided if ok: 1) */
+  code?: number;
```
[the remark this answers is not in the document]
reggi: will revert

nbbeeken (approved):

emadum (approved):
LGTM 👍

src/sdam/server.ts
```diff
@@ -584,6 +597,7 @@ function makeOperationHandler(
       } else {
         // if pre-4.4 server, then add error label if its a retryable write error
         if (
+          (server.s.topology.s.options.retryWrites !== false || isTransactionCommand(cmd)) &&
```
mbroadst, line 600:
```suggestion
          (isRetryableWritesEnabled(server.s.topology) || isTransactionCommand(cmd)) &&
```

src/sdam/server.ts
```diff
@@ -558,6 +558,15 @@ function inActiveTransaction(session: ClientSession | undefined, cmd: Document)
   return session && session.inTransaction() && !isTransactionCommand(cmd);
 }
 
+/** this checks the retryWrites option passed down from the client options, it
+ * does not check if the server supports retryable writes */
+function isRetryableWritesEnabled(serverOrTopology: Topology | Server) {
```
mbroadst, line 563:
```suggestion
function isRetryableWritesEnabled(topology: Topology) {
```

src/sdam/server.ts
```diff
@@ -558,6 +558,15 @@ function inActiveTransaction(session: ClientSession | undefined, cmd: Document)
...
+  const check = (topology: Topology) => topology.s.options.retryWrites !== false;
+  if (serverOrTopology instanceof Server) return check(serverOrTopology.s.topology);
+  if (serverOrTopology instanceof Topology) return check(serverOrTopology);
+  return false;
```
mbroadst, line 567:
```suggestion
  return topology.s.options.retryWrites !== false;
```

src/sdam/server.ts
```diff
@@ -573,7 +582,11 @@ function makeOperationHandler(
           session.serverSession.isDirty = true;
         }
 
-        if (supportsRetryableWrites(server) && !inActiveTransaction(session, cmd)) {
+        if (
+          (isRetryableWritesEnabled(server) || isTransactionCommand(cmd)) &&
```
mbroadst, line 586:
```suggestion
          (isRetryableWritesEnabled(server.s.topology) || isTransactionCommand(cmd)) &&
```

mbroadst (changes requested):

mbroadst (approved):
LGTM

PR #2534, opened 2020-08-27
NODE-2692: Increase linting strictness

nbbeeken:
Remove most of the disabled lint rules and correct the resulting errors.


package.json
```diff
@@ -92,7 +92,7 @@
     "build:docs": "npm run build:dts && typedoc",
     "check:bench": "node test/benchmarks/driverBench",
     "check:coverage": "nyc npm run check:test",
-    "check:lint": "eslint -v && eslint --max-warnings=0 --ext '.js,.ts' src test",
+    "check:lint": "tsc -v && tsc --noEmit && eslint -v && eslint --max-warnings=0 --ext '.js,.ts' src test",
```
nbbeeken, line 95: It seems apt that our compilation should be checked as part of linting

src/collection.ts
```diff
@@ -126,7 +126,7 @@ export interface CollectionOptions
 
 /** @internal */
 export interface CollectionPrivate {
-  pkFactory: PkFactory | typeof ObjectId;
+  pkFactory: PkFactory;
```
nbbeeken, line 129: Our ObjectId class does have a `createPk` method it's just missing from the community types so I prefer here to cast in the constructor which we can revisit when we have correct typing.

src/collection.ts
```diff
@@ -1652,16 +1649,30 @@ export class Collection implements OperationParent {
...
+  findAndRemove(query: Document): Promise<Document>;
+  findAndRemove(query: Document, sort: Sort): Promise<Document>;
+  findAndRemove(query: Document, sort: Sort, options: FindAndModifyOptions): Promise<Document>;
+  findAndRemove(
+    query: Document,
+    ...args: any[]
+  ): // TODO: Use labeled tuples when api-extractor supports TS 4.0
+  /*sort?: Sort | FindAndModifyOptions | Callback<Document>, */
+  /*options?: FindAndModifyOptions | Callback<Document>, */
+  /*callback?: Callback<Document>*/
```
nbbeeken, line 1670: At first glance this might look like comment garbage, and I can get rid of it if we don't want this to potentially sit here and get stale.
I would like to type args with
```typescript
...args: [sort?: Sort | FindAndModifyOptions | Callback<Document>, options?: FindAndModifyOptions | Callback<Document>, callback?: Callback<Document>]
```
Making use of the new labeled tuples gives a friendly auto complete and allows "optional" tuple members with the `?:` symbol, unfortunately you can't replicate that with a plain tuple, and currently api-extractor doesn't have the logic to parse this new TS feature.

src/collection.ts
```diff
@@ -1687,26 +1698,23 @@ export class Collection implements OperationParent {
...
-    const args = Array.prototype.slice.call(arguments, 3);
-    callback = typeof args[args.length - 1] === 'function' ? args.pop() : undefined;
-    reduce = args.length ? args.shift() : null;
-    finalize = args.length ? args.shift() : null;
-    command = args.length ? args.shift() : null;
-    options = args.length ? args.shift() || {} : {};
+    condition: Document,
+    initial: Document,
+    // TODO: Use labeled tuples when api-extractor supports TS 4.0
+    ...args: [/*reduce?:*/ any, /*finalize?:*/ any, /*command?:*/ any, /*callback?:*/ Callback]
```
nbbeeken, line 1706: I can leave the tuple here since most of the arguments are `any` its too permissive to error.

src/mongo_client.ts
```diff
@@ -427,13 +419,9 @@ export class MongoClient extends EventEmitter implements OperationParent {
...
   static connect(url: string, options: MongoClientOptions): Promise<MongoClient>;
   static connect(url: string, options: MongoClientOptions, callback: Callback<MongoClient>): void;
-  static connect(
-    url: string,
-    options?: MongoClientOptions | Callback<MongoClient>,
-    callback?: Callback<MongoClient>
-  ): Promise<MongoClient> | void {
-    if (typeof options === 'function') (callback = options), (options = {});
-    options = options || {};
+  static connect(url: string, ...args: any[]): Promise<MongoClient> | void {
```
nbbeeken, line 422: This is another place for the labeled tuple style that we could use when api-extractor supports TS 4

src/operations/bulk_write.ts
```diff
@@ -45,7 +45,7 @@ export class BulkWriteOperation extends OperationBase<BulkWriteOptions, BulkWrit
         : coll.initializeUnorderedBulkOp(options);
 
     // Do we have a collation
-    let collation = false;
+    const collation = false;
```
nbbeeken, line 48: collation is never changed, there's an if statement about 20 lines down that will never run, just wanted to get a second pair of eyes that this is correct and its okay to delete?

src/operations/common_functions.ts
```diff
@@ -81,19 +82,23 @@ export function indexInformation(
...
 
   // no need to modify the docs if server sets the ObjectId
   if (forceServerObjectId === true) {
     return docs;
   }
 
-  return docs.map((doc: any) => {
-    if (forceServerObjectId !== true && doc._id == null) {
+  return docs.map(doc => {
+    if (doc._id == null) {
```
nbbeeken, line 101: `forceServerObjectId` is either false or undefined after the check above asserts it can't be `true` so this was redundant

tsconfig.json
```diff
@@ -25,6 +25,7 @@
     "inlineSources": true,
     // Prevents web types from being suggested by vscode.
     "types": ["node"],
+    "forceConsistentCasingInFileNames": true
```
nbbeeken, line 28: I don't expect us to ever make a file with capital letters in it but this is some free linting to prevent potential issues on windows

src/promise_provider.ts
```diff
@@ -14,6 +14,7 @@ const store: PromiseStore = {
  */
 export class PromiseProvider {
   /** Validates the passed in promise library */
+  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
mbroadst, line 17: why is this needed? I just checked in vscode and there were no errors in this file

src/read_preference.ts
```diff
@@ -31,7 +33,9 @@ export interface ReadPreferenceOptions {
 }
 
 /** @public */
-export interface ReadPreferenceLikeOptions {
+export interface ReadPreferenceLikeOptions extends ReadPreferenceOptions {
+  readPreferenceTags?: TagSet[];
+  session?: ClientSession;
```
mbroadst, line 38: this looks wrong, session should have nothing to do with a `ReadPreferenceLikeOptions`

src/apm.ts
```diff
@@ -42,12 +43,14 @@ export class Instrumentation extends EventEmitter {
...
+    } as MongoClient['connect'];
 
     if (typeof callback === 'function') callback(undefined, this);
   }
 
-  uninstrument() {
-    this.$MongoClient.prototype.connect = this.$prototypeConnect;
+  uninstrument(): void {
+    if (this.$MongoClient) {
+      this.$MongoClient.prototype.connect = this.$prototypeConnect as any;
```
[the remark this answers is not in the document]
nbbeeken: nope 🙂  Gone Away.

src/change_stream.ts
```diff
@@ -209,6 +209,7 @@ export class ChangeStream extends EventEmitter {
       this.topology = parent.s.topology;
     } else if (parent instanceof MongoClient) {
       this.type = CHANGE_DOMAIN_TYPES.CLUSTER;
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
nbbeeken: I don't see an epic nor a ticket for this, I can make one.

src/collection.ts
nbbeeken: Is this because ObjectID shouldn't have createPk?
cus wouldn't this just be:
```
class DefaultPkFactory {
  static createPk() { return ObjectId.createPk() }
}
```
Why the pass through class?

src/collection.ts
nbbeeken: You do still make use of arguments in distinct to check the number of arguments provided. Doing a spread just makes this array explicit rather than the implicit global. I do agree though there isn't a benefit larger than that for now while we can't do the labeled tuples but they do provide nice hinting so something to consider for the future.

src/error.ts
```diff
@@ -177,8 +177,9 @@ export class MongoParseError extends MongoError {
  */
 export class MongoTimeoutError extends MongoError {
   /** An optional reason context for the timeout, generally an error saved during flow of monitoring and selecting servers */
-  reason?: string;
+  reason?: unknown;
```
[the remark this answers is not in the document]
nbbeeken: It appears to be only `TopologyDescription` good catch!

src/mongo_client.ts
nbbeeken: Reverted this, it was left over from an attempt to use the labeled tuples.

src/operations/group.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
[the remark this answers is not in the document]
nbbeeken: There's a lot of `any` left in here, and its kinda ugly to type without _bsontype existing on Code. I figured since this is deprecated API it was something we would be okay with leave untyped especially since its internal at this layer anyway.
This might actually go away actually with the annotations work

src/promise_provider.ts
nbbeeken: I was able to get rid of this but you should see an error for any function that uses `any` that's exported from a module.

src/read_preference.ts
nbbeeken: Indeed I added a new type `ReadPreferenceFromOptions` that represents this better, its public for now but I believe it will be made internal in the annotations work.

src/gridfs-stream/upload.ts
```diff
@@ -208,8 +209,9 @@ export class GridFSBucketWriteStream extends Writable {
     this.state.streamEnd = true;
 
     if (callback) {
+      const cb = callback;
       this.once(GridFSBucketWriteStream.FINISH, (result: GridFSFile) => {
-        callback!(undefined, result);
+        cb(undefined, result);
```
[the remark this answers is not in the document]
nbbeeken: There is a good reason:
```typescript
let callbackish: Function | undefined = () => console.log('yay')
if(callbackish){
  doLater(() => {
    callbackish() // error potentially nullable
  });
}
callbackish = undefined;
```
Since doLater runs async, its possible callback could be made null again by the time it runs.
Updated the code with your suggestion

src/connection_string.ts
```diff
@@ -36,7 +37,7 @@ function matchesParentDomain(srvAddress: string, parentDomain: string): boolean
  * @param options - Optional user provided connection string options
  */
 function parseSrvConnectionString(uri: string, options: any, callback: Callback) {
-  const result: any = url.parse(uri, true);
+  const result: Document = url.parse(uri, true);
```
mbroadst, line 40:
```suggestion
  const result: ArbitraryOptions = url.parse(uri, true);
```

mbroadst: also bring on the :bike: :house: but should it really be called `AnyOptions`?

src/mongo_client.ts
```diff
@@ -316,7 +308,7 @@ export class MongoClient extends EventEmitter implements OperationParent {
    * @see docs.mongodb.org/manual/reference/connection-string/
    */
   connect(): Promise<MongoClient>;
-  connect(callback: Callback<MongoClient>): void;
+  connect(callback: Callback<MongoClient> | undefined): void;
```
mbroadst, line 311:
```suggestion
  connect(callback?: Callback<MongoClient>): void;
```

src/collection.ts
```diff
@@ -213,23 +215,21 @@ export class Collection implements OperationParent {
 
   /**
    * The name of the database this collection belongs to
-   * @readonly
```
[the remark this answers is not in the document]
nbbeeken: If there's only a getter, its not just TS readonly but also runtime readonly too and TS knows to pick that up. The TS keyword can be used to create compiler errors if you want something to behave as readonly. And then there's the TSDoc `@readonly` which if you have a setter (maybe for internal use) but you want the documentation to reflect it as readonly you can use this annotation to do so.

src/connection_string.ts
nbbeeken: I'm fine with that, you could say the name is... _arbitrary_ 🥁

src/gridfs-stream/upload.ts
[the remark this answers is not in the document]
mbroadst: right, but it's unlikely that it is reassigned. It would be nice if there was some sort of way to hint to the compiler that if a value was provided for `callback` then it would be immutable for the duration of the methods scope. :shrug:

src/read_preference.ts
```diff
@@ -31,17 +33,25 @@ export interface ReadPreferenceOptions {
...
+        mode?: ReadPreferenceMode;
+        preference?: ReadPreferenceMode;
         tags: TagSet[];
         maxStalenessSeconds: number;
       };
 }
 
+/** @public */
+export interface ReadPreferenceFromOptions {
+  session?: ClientSession;
```
mbroadst, line 49: I'm still confused as to why `session` makes the cut here? It doesn't seem relevant to a read preference

nbbeeken: Maybe I should make another type, but it comes from the `resolve` method that uses the session to determine the read pref from a transaction if there is one.

src/operations/common_functions.ts
```diff
@@ -81,28 +82,32 @@ export function indexInformation(
...
   if (forceServerObjectId === true) {
     return docs;
   }
 
-  return docs.map((doc: any) => {
-    if (forceServerObjectId !== true && doc._id == null) {
-      doc._id = coll.s.pkFactory.createPk();
+  return docs.map(doc => {
+    if (doc._id == null) {
+      doc._id = coll.s.pkFactory();
```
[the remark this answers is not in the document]
nbbeeken: Reverted 👍

src/read_preference.ts
mbroadst: ah! okay thanks, I forgot about that.

PR #2537, opened 2020-09-08
fix: allow event loop to process during wait queue processing

mbroadst:
Running `processWaitQueue` on the next tick allows the event loop to process while the connection pool is processing large numbers of wait queue members.

NODE-2803

### DIFF
--- a/lib/cmap/connection_pool.js
+++ b/lib/cmap/connection_pool.js
@@ -233,7 +233,7 @@ class ConnectionPool extends EventEmitter {
     }
 
     this[kWaitQueue].push(waitQueueMember);
-    process.nextTick(() => processWaitQueue(this));
+    setImmediate(() => processWaitQueue(this));
   }
 
   /**
@@ -258,7 +258,7 @@ class ConnectionPool extends EventEmitter {
       destroyConnection(this, connection, reason);
     }
 
-    process.nextTick(() => processWaitQueue(this));
+    setImmediate(() => processWaitQueue(this));
   }
 
   /**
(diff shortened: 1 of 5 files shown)


reggi (approved):

emadum (approved):
LGTM, but it looks like this made a test start failing in CI (unless that's unrelated)

nbbeeken (approved):
LGTM

vkarpov15:
Unfortunately this is a breaking change for anyone that uses [Sinon's `useFakeTimers()` for testing](https://sinonjs.org/releases/v1.17.6/fake-timers/), see https://github.com/Automattic/mongoose/issues/9417. Mongoose strongly advises [against using timer mocks](https://mongoosejs.com/docs/jest.html#timer-mocks), but people use them anyway, so I think it's worth thinking about: do we need to use `setImmediate()` here, or can we get away with using `process.nextTick()`, which sinon doesn't stub?

mbroadst:
@vkarpov15 hmm, the original implementation used `process.nextTick` but after discussion around the guidance in [this document](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/) (specifically "We recommend developers use setImmediate() in all cases because it's easier to reason about.") we opted to use `setImmediate`. 

There's a possibility when using `process.nextTick` that we will still be too greedy and won't ever allow the event loop to process critical events. I think your guidance to those users is correct, they should not be mocking timers outside of unit tests. Perhaps your page on using mock timers could be updated to use the workarounds you provided them?

PR #2539, opened 2020-09-08
NODE-2273/poc-unified-spec-runner

reggi:


nbbeeken:
We chatted a bit about type checking js tests if you wanna give it a try I was able to get it going by adding a `jsconfig.json` to the test/unified subdirectory with the contents:
```
{
  "compilerOptions": {
    "target": "es2018",
    "checkJs": true,
    "noEmit": true
  }
}
```
and you have to add `@types/mocha` to our dev deps. Not sure how straight forward it is to resolve any issues that pop up after that.

test/unified/runner.js
```diff
@@ -0,0 +1,366 @@
...
+              done
+            );
+          });
+        });
+      });
+    });
+  }
+}
+
+module.exports = {
```
mbroadst, line 363: One of the nice properties of the existing spec test runner is that we are able to model the spec-specific helper functions in a declarative way on a subclass of the `TestRunnerContext`. You can see examples of that here:
- [SDAM integration tests](https://github.com/mongodb/node-mongodb-native/blob/master/src/operations/command.ts#L71-L73)
- [Transactions](https://github.com/mongodb/node-mongodb-native/blob/master/test/functional/transactions.test.js#L19-L81)
- [Sessions](https://github.com/mongodb/node-mongodb-native/blob/3.6/test/functional/sessions.test.js#L161-L185)

Is there any plan to support similar modular support here?

test/unified/runner.js
```diff
@@ -0,0 +1,366 @@
...
+  Version,
+  Runner
```
mbroadst, line 365: Another desirable quality (IMO) of the existing spec runner is it's ability to fit seamlessly into our existing mocha tests. You can mix the runner into a test file which also includes prose tests, and it looks pretty natural:
```
describe('Transactions', function () {
  describe('spec tests', function () {
    const testContext = new TransactionsRunnerContext();      
    const testSuites = loadSpecTests(suiteSpec.specPath);
    after(() => testContext.teardown());
    before(function () {
      return testContext.setup(this.configuration);
    });

    generateTopologyTests(testSuites, testContext, (spec, config) => {
        // NODE-2574: remove this when HELP-15010 is resolved
        if (config.topologyType === 'Sharded' && semver.satisfies(config.version, '>=4.4')) {
          return false;
        }

        return true;
    });
  });

  it('should do something else', function(done) { ... });
  it('should do something else', function(done) { ... });
  it('should do something else', function(done) { ... });
  it('should do something else', function(done) { ... });
});
```

Is there a goal in this rewrite to preserve any of that flexibility, or is it your opinion that we don't benefit from these features?

test/unified/runner.js
```diff
@@ -0,0 +1,366 @@
...
+    });
+    return satisfies;
+  }
+  static validate(supportedSchemaVersions, spec) {
+    const check = Version.check(supportedSchemaVersions, spec.schemaVersion);
+    if (!check) throw new Error(`invalid schemaVersion: ${spec.description}`);
+  }
+}
+
+class Match {
```
mbroadst, line 26: Though it does appear that we've lost the test suite for it, we have a very stable [custom chai matcher](https://github.com/mongodb/node-mongodb-native/blob/master/test/functional/spec-runner/matcher.js) already written that follows today's matching rules. I would prefer that we improve that code, and thoroughly test it, than introduce a new one.

test/unified/runner.js
```diff
@@ -0,0 +1,366 @@
...
+  }
+  static check(supportedSchemaVersions, schemaVersion) {
+    let satisfies = false;
+    supportedSchemaVersions.map(Version.clean).forEach(v => {
+      if (satisfies) return;
+      const result = semver.satisfies(v, `^${Version.clean(schemaVersion)}`);
+      if (result) satisfies = true;
+    });
+    return satisfies;
+  }
```
mbroadst, line 19: It seems to me that the `Version` class really just wraps a call to `semver.satisfies` with optional support for an array of supported schema versions, is that true? I'm wondering if we need to add this complexity

test/unified/runner.js
```diff
@@ -0,0 +1,366 @@
...
+
+// TODO: add in mappings for all methods
+const argMap = {
+  insertOne: a => [a.document]
+};
+
+class Runner {
+  constructor(spec, filePath) {
+    this.supportedSchemaVersions = ['1.0.0'];
+    Version.validate(this.supportedSchemaVersions, spec);
```
mbroadst, line 62: Do we expect to use `Version.validate` in more than one place? If not, does it make sense to proactively factor it into a helper method?

test/unified/runner.js
```diff
@@ -0,0 +1,366 @@
...
+    if (filePath) {
+      this.filePath = filePath;
+      this.fileName = path.basename(filePath);
+    }
```
mbroadst, line 66: What are these used for? I don't see them referred to elsewhere in the PR

mbroadst (commented):
I have some high-level questions here that I think we should answer before digging into the specifics of the implementation.

I think more generally I'm wondering if you think we can reuse more of the existing spec runner, rather than completely rewrite it. Have you looked into this? What was the result of your investigation?

test/unified/runner.js
reggi: Here the examples are showing custom assertion methods, I had no need for these in the POC with the example spec provided, we can bring these in where / if they match with the specs assertion / matchers.

test/unified/runner.js
reggi: All of this functionality is encapsulated in `Runner.handleSpec(example)` which can be run directly inside mocha.

If you choose to be declarative in the mocha file as you have, you could run the same code I run in that method.

```js
const r = new Runner(spec);
describe(r.spec.description, () => {
  r.spec.tests.forEach(test => {
    r.spec.runOnRequirements.forEach(req => {
      const metadata = Runner.createMetadata(req);
      const description = Runner.createTestDescription(test, metadata.requires);
      it(description, metadata, function (done) {
        Runner.callbacksAll(
          [
            cb => r.createTopClient(cb),
            cb => r.createEntities(cb),
            cb => r.removeInitialData(cb),
            cb => r.createInitialData(cb),
            cb => r.handleTest(test, cb),
            cb => r.closeClients(cb),
            cb => r.resetEntityMap(cb)
          ],
          done
        );
      });
    });
  });
});
```

test/unified/runner.js
reggi: This is great and I'm totally down to use it where it fits, but we still have to match object properties and I still believe there is some novelty here.

test/unified/runner.js
reggi: Yep seems true. I don't think it's that complex.

test/unified/runner.js
reggi: I think it makes sense.

test/unified/runner.js
reggi: This isn't used, can remove, but also can be useful for debugging when adding in multiple files.

reggi:
> I have some high-level questions here that I think we should answer before digging into the specifics of the implementation.
> 
> I think more generally I'm wondering if you think we can reuse more of the existing spec runner, rather than completely rewrite it. Have you looked into this? What was the result of your investigation?

@mbroadst The majority of this code is novel and loops over this specific properties defined in the unified test runner  schema. I totally understand your desire to reuse code that exists. There is some very useful lifecycle and matching related methods in the existing runner, chai subsets is a great example, but some others that I just didn't need yet.

PR #2541, opened 2020-09-09
fix: allow event loop to process during wait queue processing

mbroadst:
Running `processWaitQueue` on the next tick allows the event loop
to process while the connection pool is processing large numbers of
wait queue members. This also uncovered a few issues with timing
in our tests, and in some cases our top-level API:
  - `commitTransaction` / `abortTransaction` use `maybePromise` now
  - `endSession` must wait for all the machinery behind the scenes to
     check out a connection and write a message before considering its
     job finished
   - internal calls to `kill` a cursor now await the the process of fully
     sending that command, even if they ignore the response

NODE-2803

**NOTE:** this is the port from 3.6 to 4.0

### DIFF
--- a/test/functional/cursor.test.js
+++ b/test/functional/cursor.test.js
@@ -3987,21 +3987,17 @@ describe('Cursor', function () {
     }
   );
 
-  it('should return a promise when no callback supplied to forEach method', function (done) {
+  it('should return a promise when no callback supplied to forEach method', function () {
     const configuration = this.configuration;
     const client = configuration.newClient({ w: 1 }, { poolSize: 1, auto_reconnect: false });
 
-    client.connect(function (err, client) {
-      expect(err).to.not.exist;
+    return client.connect(() => {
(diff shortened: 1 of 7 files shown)


reggi (approved):

test/functional/cursor.test.js
```diff
@@ -3999,9 +3999,9 @@ describe('Cursor', function () {
       const cursor = collection.find();
       const promise = cursor.forEach();
       expect(promise).to.exist.and.to.be.an.instanceof(Promise);
-      promise.catch(() => {});
-
-      cursor.close(() => client.close(() => done()));
+      promise.then(() => {
+        cursor.close(() => client.close(() => done()));
```
reggi, line 4003: * One thought here is we may want to close the client regardless if success / error? 
* If there's no reason to use the promise response do we need to call then? Or is there some sort of lingering promise issue?

emadum (approved):

PR #2543, opened 2020-09-12
refactor: pull CursorStream out of Cursor

emadum:
## Description

[NODE-2820](https://jira.mongodb.org/browse/NODE-2820)

**What changed?**

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -1,7 +1,7 @@
 import Denque = require('denque');
 import { EventEmitter } from 'events';
 import { MongoError, AnyError, isResumableError } from './error';
-import { Cursor } from './cursor';
+import { Cursor, CursorOptions, CursorStream } from './cursor/cursor';
 import { AggregateOperation, AggregateOptions } from './operations/aggregate';
 import { loadCollection, loadDb, loadMongoClient } from './dynamic_loaders';
 import {
@@ -13,14 +13,13 @@ import {
   MongoDBNamespace,
   Callback
 } from './utils';
-import type { CursorOptions } from './cursor/cursor';
 import type { ReadPreference } from './read_preference';
 import type { Timestamp, Document } from './bson';
 import type { Topology } from './sdam/topology';
 import type { Writable } from 'stream';
-import type { StreamOptions } from './cursor/core_cursor';
 import type { OperationParent } from './operations/command';
 import type { CollationOptions } from './cmap/wire_protocol/write_command';
+import type { CursorCloseOptions, StreamOptions } from './cursor/core_cursor';
 const kResumeQueue = Symbol('resumeQueue');
 
 const CHANGE_STREAM_OPTIONS = ['resumeAfter', 'startAfter', 'startAtOperationTime', 'fullDocument'];
@@ -235,19 +234,21 @@ export class ChangeStream extends EventEmitter {
     // Create contained Change Stream cursor
     this.cursor = createChangeStreamCursor(this, options);
 
+    const cursorStream = this.cursor._stream;
+    if (!cursorStream) throw new MongoError('No stream on cursor');
     this.closed = false;
 
     // Listen for any `change` listeners being added to ChangeStream
     this.on('newListener', (eventName: string) => {
       if (eventName === 'change' && this.cursor && this.listenerCount('change') === 0) {
-        this.cursor.on('data', change => processNewChange(this, change));
+        cursorStream.on('data', (change: ChangeStreamDocument) => processNewChange(this, change));
       }
     });
 
     // Listen for all `change` listeners being removed from ChangeStream
     this.on('removeListener', (eventName: string) => {
       if (eventName === 'change' && this.listenerCount('change') === 0 && this.cursor) {
-        this.cursor.removeAllListeners('data');
+        cursorStream.removeAllListeners('data');
       }
     });
   }
@@ -302,9 +303,11 @@ export class ChangeStream extends EventEmitter {
 
       // Tidy up the existing cursor
       const cursor = this.cursor;
+      const cursorStream = cursor._stream;
+      if (!cursorStream) throw new MongoError('No stream on cursor');
 
       return cursor.close(err => {
-        ['data', 'close', 'end', 'error'].forEach(event => cursor.removeAllListeners(event));
+        ['data', 'close', 'end', 'error'].forEach(event => cursorStream.removeAllListeners(event));
         this.cursor = undefined;
 
         return cb(err);
@@ -328,7 +331,8 @@ export class ChangeStream extends EventEmitter {
     if (!this.cursor) {
       throw new MongoError('ChangeStream has no cursor, unable to pipe');
     }
-    return this.cursor.pipe(destination, options);
+    if (!this.cursor._stream) throw new MongoError('No stream on cursor');
+    return this.cursor._stream.pipe(destination, options);
   }
 
   /**
@@ -345,14 +349,16 @@ export class ChangeStream extends EventEmitter {
     if (!this.cursor) {
       throw new MongoError('ChangeStream has no cursor, unable to unpipe');
     }
-    return this.cursor.unpipe(destination);
+    if (!this.cursor._stream) throw new MongoError('No stream on cursor');
+    this.cursor._stream.unpipe(destination);
+    return this.cursor;
   }
 
   /**
    * Return a modified Readable stream including a possible transform method.
    * @throws MongoError if this.cursor is undefined
    */
-  stream(options?: StreamOptions): ChangeStreamCursor {
+  stream(options?: StreamOptions): CursorStream {
     this.streamOptions = options;
     if (!this.cursor) {
       throw new MongoError('ChangeStream has no cursor, unable to stream');
@@ -368,7 +374,9 @@ export class ChangeStream extends EventEmitter {
     if (!this.cursor) {
       throw new MongoError('ChangeStream has no cursor, unable to pause');
     }
-    return this.cursor.pause();
+    if (!this.cursor._stream) throw new MongoError('No stream on cursor');
+    this.cursor._stream.pause();
+    return this.cursor;
   }
 
   /**
@@ -379,7 +387,9 @@ export class ChangeStream extends EventEmitter {
     if (!this.cursor) {
       throw new MongoError('ChangeStream has no cursor, unable to resume');
     }
-    return this.cursor.resume();
+    if (!this.cursor._stream) throw new MongoError('No stream on cursor');
+    this.cursor._stream.resume();
+    return this.cursor;
   }
 }
 
@@ -397,6 +407,8 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
   hasReceived?: boolean;
   resumeAfter: ResumeToken;
   startAfter: ResumeToken;
+  emitter: EventEmitter;
+  _stream?: CursorStream;
 
   constructor(
     topology: Topology,
@@ -404,6 +416,8 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
     options: ChangeStreamCursorOptions
   ) {
     super(topology, operation, options);
+
+    this.emitter = new EventEmitter();
 
     options = options || {};
     this._resumeToken = null;
@@ -416,9 +430,34 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
     }
   }
 
+  close(): void;
+  close(callback: Callback): void;
+  close(options: CursorCloseOptions): Promise<void>;
+  close(options: CursorCloseOptions, callback: Callback): void;
+  close(
+    optionsOrCallback?: CursorCloseOptions | Callback,
+    callback?: Callback
+  ): Promise<void> | void {
+    this.emitter.emit(Cursor.CLOSE);
+    if (callback) return super.close(optionsOrCallback as CursorCloseOptions, callback);
+    return super.close(optionsOrCallback as Callback);
+  }
+
+  // emit(event: string | symbol, ...args: any[]): boolean {
+  //   return this.emitter.emit(event, ...args);
+  // }
+
+  // on(event: string | symbol, listener: (...args: any[]) => void): EventEmitter {
+  //   return this.emitter.on(event, listener);
+  // }
+
+  // once(event: string | symbol, listener: (...args: any[]) => void): EventEmitter {
+  //   return this.emitter.once(event, listener);
+  // }
+
   set resumeToken(token: ResumeToken) {
     this._resumeToken = token;
-    this.emit(ChangeStream.RESUME_TOKEN_CHANGED, token);
+    this.emitter.emit(ChangeStream.RESUME_TOKEN_CHANGED, token);
   }
 
   get resumeToken(): ResumeToken {
@@ -488,8 +527,8 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
 
       this._processBatch('firstBatch', response);
 
-      this.emit('init', response);
-      this.emit('response');
+      this.emitter.emit('init', response);
+      this.emitter.emit('response');
       callback(err, response);
     });
   }
@@ -503,8 +542,8 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
 
       this._processBatch('nextBatch', response);
 
-      this.emit('more', response);
-      this.emit('response');
+      this.emitter.emit('more', response);
+      this.emitter.emit('response');
       callback(err, response);
     });
   }
@@ -530,15 +569,17 @@ function createChangeStreamCursor(
     cursorOptions
   );
 
-  relayEvents(changeStreamCursor, self, ['resumeTokenChanged', 'end', 'close']);
+  relayEvents(changeStreamCursor.emitter, self, ['resumeTokenChanged', 'end', 'close']);
+
+  changeStreamCursor._stream = changeStreamCursor.stream(self.streamOptions);
+  relayEvents(changeStreamCursor._stream, self, ['end', 'close']);
 
   if (self.listenerCount(ChangeStream.CHANGE) > 0) {
-    changeStreamCursor.on(ChangeStreamCursor.DATA, function (change) {
+    changeStreamCursor._stream.on(CursorStream.DATA, function (change) {
       processNewChange(self, change);
     });
   }
-
-  changeStreamCursor.on(ChangeStream.ERROR, function (error) {
+  changeStreamCursor._stream.on(ChangeStream.ERROR, function (error) {
     processError(self, error);
   });
 
@@ -660,12 +701,9 @@ function processError(changeStream: ChangeStream, error?: AnyError, callback?: C
     changeStream.cursor = undefined;
 
     // stop listening to all events from old cursor
-    [
-      ChangeStreamCursor.DATA,
-      ChangeStreamCursor.CLOSE,
-      ChangeStreamCursor.END,
-      ChangeStreamCursor.ERROR
-    ].forEach(event => cursor.removeAllListeners(event));
+    [CursorStream.DATA, CursorStream.CLOSE, CursorStream.END, CursorStream.ERROR].forEach(event =>
+      cursor._stream?.removeAllListeners(event)
+    );
 
     // close internal cursor, ignore errors
     cursor.close();
--- a/src/cursor/cursor.ts
+++ b/src/cursor/cursor.ts
@@ -1,7 +1,7 @@
 import { emitDeprecatedOptionWarning } from '../utils';
 import { PromiseProvider } from '../promise_provider';
 import { ReadPreference, ReadPreferenceLike } from '../read_preference';
-import { Transform, PassThrough } from 'stream';
+import { Transform, PassThrough, Readable } from 'stream';
 import { deprecate } from 'util';
 import { MongoError, AnyError } from '../error';
 import {
@@ -53,6 +53,82 @@ export interface CursorOptions extends CoreCursorOptions {
   topology?: Topology;
   /** Session to use for the operation */
   numberOfRetries?: number;
+}
+
+export class CursorStream extends Readable {
+  cursor: Cursor;
+
+  /** @event */
+  static readonly CLOSE = 'close' as const;
+  /** @event */
+  static readonly DATA = 'data' as const;
+  /** @event */
+  static readonly END = 'end' as const;
+  /** @event */
+  static readonly FINISH = 'finish' as const;
+  /** @event */
+  static readonly ERROR = 'error' as const;
+  /** @event */
+  static readonly PAUSE = 'pause' as const;
+  /** @event */
+  static readonly READABLE = 'readable' as const;
+  /** @event */
+  static readonly RESUME = 'resume' as const;
+
+  constructor(cursor: Cursor) {
+    super({ objectMode: true });
+    this.cursor = cursor;
+  }
+
+  destroy(err?: AnyError): void {
+    console.log('destroy');
+    if (err) this.emit(Cursor.ERROR, err);
+    this.pause();
+    this.cursor.close(() => this.emit(Cursor.CLOSE));
+  }
+
+  /** @internal */
+  _read(): void {
+    if ((this.cursor.s && this.cursor.s.state === CursorState.CLOSED) || this.cursor.isDead()) {
+      this.push(null);
+      return;
+    }
+
+    // Get the next item
+    this.cursor._next((err, result) => {
+      if (err) {
+        if (this.listeners(CursorStream.ERROR) && this.listeners(CursorStream.ERROR).length > 0) {
+          this.emit(CursorStream.ERROR, err);
+        }
+        if (!this.cursor.isDead()) this.cursor.close();
+
+        // Emit end event
+        this.emit(CursorStream.END);
+        this.emit(CursorStream.FINISH);
+        return;
+      }
+
+      // If we provided a transformation method
+      if (
+        this.cursor.cursorState.streamOptions &&
+        typeof this.cursor.cursorState.streamOptions.transform === 'function' &&
+        result != null
+      ) {
+        this.push(this.cursor.cursorState.streamOptions.transform(result));
+        return;
+      }
+
+      // Return the result
+      this.push(result);
+
+      if (result === null && this.cursor.isDead()) {
+        this.once(CursorStream.END, () => {
+          this.cursor.close();
+          this.emit(CursorStream.FINISH);
+        });
+      }
+    });
+  }
 }
 
 /**
@@ -828,10 +904,7 @@ export class Cursor<
         this.kill();
       }
 
-      this._endSession(() => {
-        this.emit(Cursor.CLOSE);
-        cb(undefined, this);
-      });
+      this._endSession(() => cb(undefined, this));
     });
   }
 
@@ -857,17 +930,11 @@ export class Cursor<
     return this.isDead();
   }
 
-  destroy(err?: AnyError): void {
-    if (err) this.emit(Cursor.ERROR, err);
-    this.pause();
-    this.close();
-  }
-
   /** Return a modified Readable stream including a possible transform method. */
-  stream(options?: StreamOptions): this {
+  stream(options?: StreamOptions): CursorStream {
     // TODO: replace this method with transformStream in next major release
     this.cursorState.streamOptions = options || {};
-    return this;
+    return new CursorStream(this);
   }
 
   /**
@@ -887,10 +954,10 @@ export class Cursor<
         }
       });
 
-      return this.pipe(stream);
+      return this.stream().pipe(stream);
     }
 
-    return this.pipe(new PassThrough({ objectMode: true }));
+    return this.stream().pipe(new PassThrough({ objectMode: true }));
   }
 
   /**
--- a/test/functional/change_stream.test.js
+++ b/test/functional/change_stream.test.js
@@ -29,14 +29,18 @@ function withChangeStream(dbName, collectionName, callback) {
 
   return withClient((client, done) => {
     const db = client.db(dbName);
-    db.createCollection(collectionName, { w: 'majority' }, (err, collection) => {
-      if (err) return done(err);
-      withCursor(
-        collection.watch(),
-        (cursor, done) => callback(collection, cursor, done),
-        err => collection.drop(dropErr => done(err || dropErr))
-      );
-    });
+    db.dropCollection(collectionName)
+      .then(() =>
+        db.createCollection(collectionName, { w: 'majority' }, (err, collection) => {
+          if (err) return done(err);
+          withCursor(
+            collection.watch(),
+            (cursor, done) => callback(collection, cursor, done),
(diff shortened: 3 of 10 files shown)


src/change_stream.ts
```diff
@@ -235,19 +234,21 @@ export class ChangeStream extends EventEmitter {
     // Create contained Change Stream cursor
     this.cursor = createChangeStreamCursor(this, options);
 
+    const cursorStream = this.cursor._stream;
+    if (!cursorStream) throw new MongoError('No stream on cursor');
```
mbroadst, line 238: A `CursorStream` should only ever be created if a user is attempting to use the change stream as a stream. I'm actually wondering if we should continue the breaking changes here:
- let's make `Cursor` an `EventEmitter` as we discussed in team meeting yesterday
- let's remove any logic from `ChangeStream` about it being an explicit `Readable`, and use the same behavior as cursor here: you must call `stream()` on the change stream in order to use it as a `Readable` stream. I think this will cut out a lot of complexity of the implementation since you won't have to manage an internal stream anymore. What do you think?

src/change_stream.ts
```diff
@@ -405,6 +417,8 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
   ) {
     super(topology, operation, options);
 
+    this.emitter = new EventEmitter();
```
mbroadst, line 420: as discussed in team meeting, let's just keep the `EventEmitter` base on the core cursor type

src/change_stream.ts
```diff
@@ -416,9 +430,34 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
...
 
+  close(): void;
+  close(callback: Callback): void;
+  close(options: CursorCloseOptions): Promise<void>;
+  close(options: CursorCloseOptions, callback: Callback): void;
+  close(
+    optionsOrCallback?: CursorCloseOptions | Callback,
+    callback?: Callback
+  ): Promise<void> | void {
+    this.emitter.emit(Cursor.CLOSE);
```
mbroadst, line 441: We haven't completely settled on a pattern for this, but I think the majority of our code still does something like:
```
close(options?: CursorCloseOptions | Callback<void>, callback?: Callback<void>): Promise<void> | void {
  options = options || {};
  ...
}
```
It types just fine, let's you use simple names for the parameters, and doesn't require you to do things like casting `callback` below

mbroadst (commented):
sorry, jumped the gun on reviewing, but since I already did it here are some thoughts

test/functional/spec-runner/index.js
```diff
@@ -125,20 +125,19 @@ function generateTopologyTests(testSuites, testContext, filter) {
       describe(suiteName, {
         metadata: { requires },
         test: function () {
-          beforeEach(() => prepareDatabaseForSuite(testSuite, testContext));
-          afterEach(() => testContext.cleanupAfterSuite());
+          // beforeEach(() => );
```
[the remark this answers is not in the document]
emadum: I was getting leaked session errors for the tests that were skipped using `this.skip()`, which seemed to be because `beforeEach`/`afterEach` run even on tests skipped at runtime. Moving the hook behavior into the test resolved the issue, and I don't see any advantage of using the beforeEach/afterEach hooks since all of the tests are generated with the same code. 

I'm not sure if this just popped up or has been happening for a while. It wasn't a fatal error, so it wouldn't have caused any failures in evergreen.

src/change_stream.ts
emadum: Agreed, this makes sense and will definitely simplify the implementation. I think the consistency is also good; it should be easy for users to get used to calling `stream()` when they need a stream.

test/node-next/es2018/cursor_async_iterator.test.js
```diff
@@ -72,16 +72,17 @@ describe('Cursor Async Iterator Tests', function () {
     }
   });
 
-  it('should properly stop when cursor is closed', {
+  // TODO - figure out how to get this test working again
```
emadum, line 75: The `for await ... of`  statement consumes the entire stream before entering the body of the loop, so this test fails with `count` larger than 1. I'm not sure if that means there's something incorrect with the implementation, or if this is a bad test.

src/cursor/cursor.ts
```diff
@@ -24,6 +24,8 @@ import type { Sort, SortDirection } from '../operations/find';
 import type { Hint, OperationBase } from '../operations/operation';
 import type { Document } from '../bson';
 
+const kCursor = Symbol('internalCursor');
```
mbroadst, line 27: nit: I would just call this `cursor`, doesn't hugely matter but since the purpose of this is to make it easy to print out/inspect maybe brevity is preferred

src/cursor/cursor.ts
```diff
@@ -55,6 +57,83 @@ export interface CursorOptions extends CoreCursorOptions {
...
+  static readonly PAUSE = 'pause' as const;
+  /** @event */
+  static readonly READABLE = 'readable' as const;
+  /** @event */
+  static readonly RESUME = 'resume' as const;
+
+  constructor(cursor: Cursor) {
+    super({ objectMode: true });
+    this[kCursor] = cursor;
+    this.on('end', () => this[kCursor].close(() => this.emit(Cursor.CLOSE)));
```
mbroadst, line 83: should the `CursorStream` emit the `close` event still? I thought the `Cursor` would just emit this now? Seems to me that the stream should _only_ emit events that streams emit

src/cursor/cursor.ts
```diff
@@ -55,6 +57,83 @@ export interface CursorOptions extends CoreCursorOptions {
...
+    const cursor = this[kCursor];
+    if ((cursor.s && cursor.s.state === CursorState.CLOSED) || cursor.isDead()) {
+      this.push(null);
+      return;
+    }
+
+    // Get the next item
+    cursor._next((err, result) => {
+      if (err) {
+        if (this.listeners(CursorStream.ERROR) && this.listeners(CursorStream.ERROR).length > 0) {
```
mbroadst, line 103: we should always throw the error, these types of checks were around depending on whether you are using this as a stream or iterator. The iterator would return the error during next iteration, but when you use a stream you need to take all the events into consideration.

src/cursor/cursor.ts
```diff
@@ -55,6 +57,83 @@ export interface CursorOptions extends CoreCursorOptions {
...
+          this.emit(CursorStream.ERROR, err);
+        }
+        if (!cursor.isDead()) cursor.close();
```
mbroadst, line 106: should we be waiting for `cursor.close` to complete before emitting these events?

src/cursor/cursor.ts
```diff
@@ -55,6 +57,83 @@ export interface CursorOptions extends CoreCursorOptions {
...
+
+        // Emit end event
+        this.emit(CursorStream.END);
+        this.emit(CursorStream.FINISH);
```
mbroadst, line 110: are these standard events for a stream? I'm thinking we should only depend on what the `Readable` implementation emits, and it probably only emits end when you `push(null)`?

src/cursor/cursor.ts
```diff
@@ -55,6 +57,83 @@ export interface CursorOptions extends CoreCursorOptions {
...
+
+      // Return the result
+      this.push(result);
+
+      if (result === null && cursor.isDead()) {
+        this.once(CursorStream.END, () => {
+          cursor.close();
+          this.emit(CursorStream.FINISH);
+        });
+      }
```
mbroadst, line 132: I'm again unclear whether we should be doing all of this anymore. A few questions:
- when `result === null`, I believe `cursor.isClosed() === true` can you confirm? If so, we're needlessly duplicating the `cursor.close()` here.
- if `result === null` then we `push(null)` and I believe that will cause the stream to end, so we're emitting duplicate events here

src/cursor/cursor.ts
```diff
@@ -828,10 +907,7 @@ export class Cursor<
         this.kill();
       }
 
-      this._endSession(() => {
-        this.emit(Cursor.CLOSE);
-        cb(undefined, this);
-      });
+      this._endSession(() => cb(undefined, this));
```
mbroadst, line 910: I thought the idea was that the cursor would still emit `close` just for symmetry with `isClosed`? It also seems to me that a `CursorStream` created from this `Cursor` would need to listen to that event in order to proactively close itself as well.

src/cursor/cursor.ts
```diff
@@ -828,10 +907,7 @@ export class Cursor<
         this.kill();
```
mbroadst, line 907: just a thought since we're here: `this.kill()` sends a command to the server, I think we might need to await the result of that process (checking connection out of the pool, sending command, etc) even though the `killCursors` operation is sent not expecting a response. Can you take a look, if you concur can you make a ticket for this and add it to the epic?

test/examples/change_streams.js
```diff
@@ -179,7 +179,7 @@ describe('examples(change-stream):', function () {
       ];
 
       const collection = db.collection('inventory');
-      const changeStream = collection.watch(pipeline);
+      const changeStream = collection.watch(pipeline).stream();
```
mbroadst, line 182: I think maybe we crossed wires somewhere, I apologize if that's the case. I think a `ChangeStream` should continue to be an `EventEmitter` and support the `on('change', ...)` approach out of the box. We only want the `ChangeStreamStream` for cases where people want to explicitly do something like:
```
const changeStream = collection.watch(...);
changeStream.stream().pipe(process.stdout);
```

src/change_stream.ts
```diff
@@ -236,20 +235,6 @@ export class ChangeStream extends EventEmitter {
     this.cursor = createChangeStreamCursor(this, options);
 
     this.closed = false;
-
-    // Listen for any `change` listeners being added to ChangeStream
```
mbroadst, line 240: we'll want to reinstate this per my comment below

src/change_stream.ts
```diff
@@ -304,14 +295,86 @@ export class ChangeStream extends EventEmitter {
...
 
+  /**
+   * Return a modified Readable stream including a possible transform method.
+   * @throws MongoError if this.cursor is undefined
+   */
+  stream(options?: StreamOptions): ChangeStreamStream {
+    if (this[kStream] == null) {
+      this[kStream] = new ChangeStreamStream(this, options);
+    }
+    return this[kStream] as ChangeStreamStream;
```
mbroadst, line 314: is this cast necessary? What other type could `this[kStream]` be?

src/change_stream.ts
```diff
@@ -304,14 +295,86 @@ export class ChangeStream extends EventEmitter {
...
         return cb(err);
       });
     });
   }
 
+  /**
+   * Return a modified Readable stream including a possible transform method.
+   * @throws MongoError if this.cursor is undefined
+   */
+  stream(options?: StreamOptions): ChangeStreamStream {
```
mbroadst, line 310: while we're here, do you think we should rename `StreamOptions` to something less generic, maybe `CursorStreamOptions`?

src/change_stream.ts
```diff
@@ -304,14 +295,86 @@ export class ChangeStream extends EventEmitter {
...
+    if (this[kStream] == null) {
+      this[kStream] = new ChangeStreamStream(this, options);
+    }
+    return this[kStream] as ChangeStreamStream;
+  }
+
+  repipeStreams(changeStreamCursor: ChangeStreamCursor): void {
```
mbroadst, line 317: did you mean to make this public API?

src/change_stream.ts
```diff
@@ -304,14 +295,86 @@ export class ChangeStream extends EventEmitter {
...
+    }
+  }
+}
+
+class ChangeStreamStream extends EventEmitter {
+  changeStream: ChangeStream;
+  pipeDestinations: Writable[] = [];
+  cursor: ChangeStreamCursor;
+  streamOptions: StreamOptions;
+  cursorStream: CursorStream;
```
mbroadst, line 334: so this is kind of what I was talking about the other day, I think we actually want to make `ChangeStreamStream extends Readable` or at least `extends CursorStream`, so that we can be in control over the resumability of the stream. Similarly to how we want people to only write `changeStream.on('change', ...)` one time and not think about resuming, we want people doing `changeStream.stream().pipe(process.stdout)` to never have to write resume logic themselves.

src/change_stream.ts
```diff
@@ -325,10 +388,7 @@ export class ChangeStream extends EventEmitter {
       this.pipeDestinations = [];
     }
     this.pipeDestinations.push(destination);
-    if (!this.cursor) {
-      throw new MongoError('ChangeStream has no cursor, unable to pipe');
-    }
-    return this.cursor.pipe(destination, options);
+    return this.cursorStream.pipe(destination, options);
```
mbroadst, line 391: If we have `ChangeStreamStream` do we need to override `pipe` anymore? All of these methods that simulate `Readable` go away if we just inherit from `Readable`

src/change_stream.ts
```diff
@@ -416,6 +462,20 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
...
+  close(options: CursorCloseOptions): Promise<void>;
+  close(options: CursorCloseOptions, callback: Callback): void;
+  close(
+    options?: CursorCloseOptions | Callback<void>,
+    callback?: Callback<void>
+  ): Promise<void> | void {
+    if (typeof options === 'function') (callback = options), (options = {});
+    options = options || {};
+    this.emit(Cursor.CLOSE);
+    return super.close(options, callback as Callback);
```
mbroadst, line 476: I think no need to cast here?

mbroadst (changes requested):
Lookin good! I think there's some small misunderstanding about `ChangeStream` still being an `EventEmitter`, lmk if you want to chat more about that

src/cursor/cursor.ts
```diff
@@ -54,6 +54,81 @@ export interface CursorOptions extends CoreCursorOptions {
...
+  /** @event */
+  static readonly FINISH = 'finish' as const;
+  /** @event */
+  static readonly ERROR = 'error' as const;
+  /** @event */
+  static readonly PAUSE = 'pause' as const;
+  /** @event */
+  static readonly READABLE = 'readable' as const;
+  /** @event */
+  static readonly RESUME = 'resume' as const;
```
mbroadst, line 77: I think most of these can go away, right? Aren't they events that we inherit from `Readable`?

src/cursor/cursor.ts
mbroadst: to be more specific about this comment:
```
if (err) {
  this.emit(CursorStream.ERROR, err);
  cursor.close(() => this.emit(CursorStream.END));
  return;
}
```

src/cursor/cursor.ts
```diff
@@ -54,6 +54,81 @@ export interface CursorOptions extends CoreCursorOptions {
...
+          this.emit(CursorStream.ERROR, err);
+        }
+        const done = () => this.emit(CursorStream.END);
+        cursor.isDead() ? done() : cursor.close(done);
+        return;
+      }
+
+      // If we provided a transformation method
+      if (
+        cursor.streamOptions &&
```
mbroadst, line 111: shouldn't the `CursorStream` be configured with `CursorStreamOptions`? Why check for those options back on the cursor itself?

src/cursor/cursor.ts
mbroadst: @emadum did you get a chance to see this comment yet? I think it still stands.

src/cursor/cursor.ts
```diff
@@ -845,24 +920,18 @@ export class Cursor<
...
-    this.close();
-  }
-
   /** Return a modified Readable stream including a possible transform method. */
-  stream(options?: StreamOptions): this {
+  stream(options?: CursorStreamOptions): CursorStream {
     // TODO: replace this method with transformStream in next major release
     this.streamOptions = options || {};
-    return this;
+    return new CursorStream(this);
```
mbroadst, line 927: `new CursorStream(this, options)` - let's _not_ store `streamOptions` on the cursor itself, that's a side effect

src/cursor/cursor.ts
```diff
@@ -875,10 +944,10 @@ export class Cursor<
         }
       });
 
-      return this.pipe(stream);
+      return this.stream().pipe(stream);
```
mbroadst, line 947:
```suggestion
      return this.stream(options).pipe(stream);
```

src/cursor/cursor.ts
```diff
@@ -875,10 +944,10 @@ export class Cursor<
...
     }
 
-    return this.pipe(new PassThrough({ objectMode: true }));
+    return this.stream().pipe(new PassThrough({ objectMode: true }));
```
mbroadst, line 950:
```suggestion
    return this.stream(options).pipe(new PassThrough({ objectMode: true }));
```

src/change_stream.ts
```diff
@@ -235,19 +244,21 @@ export class ChangeStream extends EventEmitter {
     // Create contained Change Stream cursor
     this.cursor = createChangeStreamCursor(this, options);
 
+    const cursorStream = this.cursor[kCursorStream];
+    if (!cursorStream) throw new MongoError('No stream on cursor');
```
mbroadst, line 248: This is confusing to me, this is the constructor of the `ChangeStream` and we're effectively asserting that the `cursorStream` will _always_ exist, which shouldn't be the case. If I do something like this:
```
const changeStream = collection.watch();
for await (const doc of changeStream) {
  console.dir({ doc });
}
```
A `cursorStream` should never have been created. In fact, you might want to add a test for that? 

I think what you want is to modify the `newListener` handler below. If someone registers a listener for `change`, then we want to set up a new `cursorStream` and forward the `data` event to them. Otherwise, if they want to stream from this change stream, they should call the `stream` method.

src/change_stream.ts
```diff
@@ -235,19 +234,21 @@ export class ChangeStream extends EventEmitter {
...
 
+    const cursorStream = this.cursor._stream;
+    if (!cursorStream) throw new MongoError('No stream on cursor');
     this.closed = false;
 
     // Listen for any `change` listeners being added to ChangeStream
     this.on('newListener', (eventName: string) => {
       if (eventName === 'change' && this.cursor && this.listenerCount('change') === 0) {
-        this.cursor.on('data', change => processNewChange(this, change));
+        cursorStream.on('data', (change: ChangeStreamDocument) => processNewChange(this, change));
```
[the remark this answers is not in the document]
mbroadst: @emadum ping

src/change_stream.ts
```diff
@@ -302,62 +313,29 @@ export class ChangeStream extends EventEmitter {
 
       // Tidy up the existing cursor
       const cursor = this.cursor;
+      const cursorStream = cursor[kCursorStream];
+      if (!cursorStream) throw new MongoError('No stream on cursor');
```
mbroadst, line 317: Similar to above, I'm very surprised this isn't being triggered anywhere in the tests. This is a very aggressive check that I would expect to throw in cases where we only use the `ChangeStream` as an iterator, not as an event emitter.

src/change_stream.ts
```diff
@@ -302,62 +313,29 @@ export class ChangeStream extends EventEmitter {
...
    */
-  stream(options?: StreamOptions): ChangeStreamCursor {
+  stream(options?: CursorStreamOptions): ChangeStreamStream {
     this.streamOptions = options;
     if (!this.cursor) {
       throw new MongoError('ChangeStream has no cursor, unable to stream');
     }
-    return this.cursor.stream(options);
+    this.cursor.streamOptions = options;
+    return new ChangeStreamStream(this.cursor);
```
mbroadst, line 338:
```suggestion
    return this.cursor.stream(options);
```

For now there should be no difference, we'll tackle the `ChangeStreamStream` in another PR. Also, I'm not sure the check for `this.cursor` is valid since cursors can switch (hence the introduction of the `getCursor` method). Maybe just replace this entire method with the suggestion above?

src/change_stream.ts
```diff
@@ -368,7 +346,10 @@ export class ChangeStream extends EventEmitter {
     if (!this.cursor) {
       throw new MongoError('ChangeStream has no cursor, unable to pause');
     }
-    return this.cursor.pause();
+    const stream = this.cursor[kCursorStream];
```
mbroadst, line 349: `pause` and `resume` are methods providing API parity with Node streams, they should be removed. If users want to use the change stream as a stream, they should call the `stream` method.

src/change_stream.ts
```diff
@@ -397,6 +381,7 @@ export class ChangeStreamCursor extends Cursor<AggregateOperation, ChangeStreamC
   hasReceived?: boolean;
   resumeAfter: ResumeToken;
   startAfter: ResumeToken;
+  [kCursorStream]?: CursorStream;
```
mbroadst, line 384: I think this belongs on the `ChangeStream` type. This type is a specialization on `Cursor` which is solely focused on isolating initialization and property tracking for the cursor, I don't think we should be using this type to store private members of the parent `ChangeStream`. 

When a user uses the `ChangeStream` as an event emitter (listening to the `change` event), then we want to create a singular internal cursor stream to listen to - this is differentiated from calling `stream()` which can happen many times and create many streams for a user to consume, backed by the same cursor.

src/change_stream.ts
mbroadst: @emadum ping

src/change_stream.ts
```diff
@@ -532,22 +531,16 @@ function createChangeStreamCursor(
...
-    const cursorStream = changeStreamCursor.stream(self.streamOptions);
-    for (const pipeDestination of self.pipeDestinations) {
-      cursorStream.pipe(pipeDestination);
-    }
+  if (self.listenerCount(ChangeStream.CHANGE) > 0) {
+    stream.on(CursorStream.DATA, change => processNewChange(self, change));
   }
+  stream.on(ChangeStream.ERROR, error => processError(self, error));
+  // for internal use only, allows for triggering of mock errors on the internal stream
+  self.on('streamError', event => stream.emit('error', event));
```
mbroadst, line 543: Is it possible to make this change without introducing a new custom error event? It seems like we should be able to do something like stub a method on `CursorStream` with a fake which emits the error?

mbroadst (changes requested):

src/cursor/cursor.ts
emadum: I tried removing the `cursor.close`, but a test ("should invalidate change stream on collection rename using event listeners") started timing out (even though `cursor.isClosed() === true` before calling `cursor.close()`). I can take a closer look.

src/change_stream.ts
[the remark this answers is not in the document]
emadum: Removed, not sure why I had to add it in the first place (was getting a TS error before).

src/cursor/cursor.ts
emadum: I suppose we could take them away, `finish` is the only non-standard one. However, it's nice getting all the supported events as constants in code completion when I start typing `myCursorStream.on(CursorStream.`, which goes away if we remove these.

src/cursor/cursor.ts
emadum: On the subject of `.stream()`, what's with the TODO comment on L925?

`// TODO: replace this method with transformStream in next major release`

Maybe this is the right time to do that? Do you have any context on that TODO?

src/cursor/cursor.ts
emadum: Actually I had to revert this, passing in the options causes it to get double-transformed. Once in `read()` and once by the `Transform` stream.

src/change_stream.ts
emadum: That makes sense, I was also a bit confused why this ended up working and surprised no tests failed. I'll try the `newListener` approach out.

src/change_stream.ts
emadum: I was getting a TS error without casting, but this entire change is now unnecessary, it was based on a misunderstanding I had.

src/change_stream.ts
emadum: @mbroadst I think I got this ironed out, PTAL 🙏

src/change_stream.ts
emadum: The problem is we need some way of accessing the `processError` function (which is not exported from `change_stream.ts`). I added an `@internal` method called `triggerError` to do this for now, but open to advice about better ways of achieving this.

test/node-next/es2018/cursor_async_iterator.test.js
emadum: @mbroadst This test is still failing if un-skipped, any thoughts on what could be going wrong here?

src/change_stream.ts
```diff
@@ -304,83 +326,28 @@ export class ChangeStream extends EventEmitter {
       const cursor = this.cursor;
 
       return cursor.close(err => {
-        ['data', 'close', 'end', 'error'].forEach(event => cursor.removeAllListeners(event));
+        ['data', 'close', 'end', 'error'].forEach(event =>
```
mbroadst, line 329: nit: I would invert this, if the cursor stream is null then you don't want to needlessly loop through the event list:
```
if (this[kCursorStream]) {
   ['data', 'close', 'end', 'error'].forEach(event => this[kCursorStream]?.removeAllListeners(event));
}
```

src/change_stream.ts
```diff
@@ -648,24 +593,18 @@ function processError(changeStream: ChangeStream, error?: AnyError, callback?: C
...
     changeStream.cursor = undefined;
 
     // stop listening to all events from old cursor
-    [
-      ChangeStreamCursor.DATA,
-      ChangeStreamCursor.CLOSE,
-      ChangeStreamCursor.END,
-      ChangeStreamCursor.ERROR
-    ].forEach(event => cursor.removeAllListeners(event));
+    [CursorStream.DATA, CursorStream.CLOSE, CursorStream.END, CursorStream.ERROR].forEach(event =>
```
mbroadst, line 604: similar comment to above, no need to iterate over all of these if the cursor stream doesn't exist.

src/change_stream.ts
mbroadst: also, you use defined types for this below (`[CursorStream.DATA, CursorStream.CLOSE, CursorStream.END, CursorStream.ERROR]`)

src/cursor/cursor.ts
mbroadst: ok let's leave it alone for now

src/cursor/cursor.ts
mbroadst: ok sounds good. Yes, I think we should likely remove the `transformStream` method, it was an attempt at providing a better `stream` without breaking compatibility. Can you make a new ticket to remove that method? We'll need to deprecate it in 3.6+ and ensure we have documentation around the removal

test/functional/change_stream.test.js
```diff
@@ -61,7 +66,7 @@ function triggerResumableError(changeStream, delay, onClose) {
   });
 
   function triggerError() {
-    changeStream.cursor.emit('error', new MongoNetworkError('fake error'));
+    changeStream.triggerError(new MongoNetworkError('fake error'));
```
mbroadst, line 69: rather than creating a special method (`triggerError`) for behavior we wish to test, what if you made an internal accessor for the stream? 
```
class ChangeStream { 
  /** @internal */
  get cursorStream() {
    return this[kCursorStream];
  }
}
```

If you had that, then you would be able to renable the two tests below as well

src/change_stream.ts
```diff
@@ -240,18 +250,30 @@ export class ChangeStream extends EventEmitter {
...
     this.on('removeListener', (eventName: string) => {
       if (eventName === 'change' && this.listenerCount('change') === 0 && this.cursor) {
-        this.cursor.removeAllListeners('data');
+        this[kCursorStream]?.removeAllListeners('data');
       }
     });
   }
 
+  /** @internal */
+  streamEvents(cursor: ChangeStreamCursor): void {
```
mbroadst, line 265: it doesn't look like this method is required outside of this implementation, is there a reason you attached it to the `ChangeStream` class? Prefer to make it a file-local private method than attaching it as public API (even if its marked internal).

mbroadst (changes requested):
Looking really great, I have just a few more comments but I think we can merge this shortly

src/change_stream.ts
mbroadst: Oops missed this comment, I left my suggestion below in the test

src/cursor/cursor.ts
mbroadst: At this point I think we really want to just remove `transformStream` and standardize on `stream`. I left another comment about that, let's tackle it in another ticket.

test/node-next/es2018/cursor_async_iterator.test.js
mbroadst: There are a couple of issues here:
- We removed the [Symbol.asyncIterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) implementation for cursor because it already existed on `Readable`, so that actually needs to be reimplemented here because we want people to be able to do `for await (const doc of cursor) { ... }`

- If this is currently broken, that means that `CursorStream` is broken somehow. You might just have a silly bug here where you're creating a new stream on every iteration, in which case this fix should suffice:
```
const cursorStream = cursor.stream();
for await (const doc of cursorStream) { ... }
```
Otherwise, this would indicate that when you close the cursor it is not closing the stream which might be just fine! We assume that the stream might consume a bunch of document on that first iteration so this is probably desired behavior.

Let's fix the first problem, and revert the changes to the test here as your validation that the async iterator is implemented properly.

mbroadst (commented):
Left one

test/functional/change_stream.test.js
emadum: Unfortunately that approach only works when there *is* an internal stream we can emit events on. When the change stream is being iterated, there's no internal stream and therefore nowhere to inject the error without something that exposes `processError` more directly like this.

test/node-next/es2018/cursor_async_iterator.test.js
```diff
@@ -76,12 +77,19 @@ describe('Cursor Async Iterator Tests', function () {
     metadata: { requires: { node: '>=10.5.0' } },
     test: async function () {
       const cursor = collection.find();
+      const cursorStream = cursor.stream().pipe(
```
emadum, line 80: @mbroadst Unfortunately changing the code made it so I can't reply directly to our old thread about this. I've modified the test so that `cursor.close()` gets called before the first object is returned from the stream, which I think was the intent of the test as originally written. It works this way; do you think that's good enough, or do we want to dive further into why for stream gets consumed entirely before the first iteration of the loop's body gets executed?

test/functional/change_stream.test.js
```diff
@@ -2784,6 +2793,37 @@ describe('Change Stream Resume Error Tests', function () {
       });
     })
   });
+
+  it.skip('should continue piping changes after a resumable error', {
```
reggi, line 2797: This test is `skipped`, should it be? If so, can you comment why?

reggi (approved):
@emadum This is really cool, a lot in here! LGTM!

One request: could you add a description of the fundamental changes here to this pull? With the intention of bringing those into the body of the commit message, just because this is such a big change it would be really nice to have a plain-english description of the changes here.

nbbeeken (approved):
Every question I had was covered by the proceeding reviewers, excellent work! lgtm

mbroadst (approved):
LGTM!

PR #2544, opened 2020-09-14
NODE-2784: sets primary read preference for writes

reggi:


lib/operations/operation.js
```diff
@@ -1,11 +1,14 @@
...
+const ReadPreference = require('../core/topologies/read_preference');
+
 const Aspect = {
   READ_OPERATION: Symbol('READ_OPERATION'),
   WRITE_OPERATION: Symbol('WRITE_OPERATION'),
   RETRYABLE: Symbol('RETRYABLE'),
   EXECUTE_WITH_SELECTION: Symbol('EXECUTE_WITH_SELECTION'),
-  NO_INHERIT_OPTIONS: Symbol('NO_INHERIT_OPTIONS')
+  NO_INHERIT_OPTIONS: Symbol('NO_INHERIT_OPTIONS'),
+  USE_PRIMARY_SERVER_SELECTION: Symbol('USE_PRIMARY_SERVER_SELECTION')
```
mbroadst, line 11: Do we need a new aspect? In `4.0` we've already [solved this problem](https://github.com/mongodb/node-mongodb-native/blob/master/src/operations/command.ts#L71-L73) by fixing the `readPreference` to `primary` when constructing a `WRITE_OPERATION`.

mbroadst (changes requested):

mbroadst:
The fix itself LGTM. Maybe I'm jumping the gun before re-requesting reviews, but will there be tests added to validate that DDL operations are all using fixed read preferences?

test/functional/collection.test.js
```diff
@@ -1307,4 +1312,34 @@ describe('Collection', function() {
       });
     }
   });
+
+  it('should createIndex with serverSelection readPreference primary', {
```
mbroadst, line 1316: Lookin pretty good, I have a few thoughts:
- Can we make this a data-driven test so we're testing all [DDL](https://www.geeksforgeeks.org/sql-ddl-dql-dml-dcl-tcl-commands/) operations? 
- Since testing all of those operation types is likely to be hairy (`create` for database and collection, for instance), I think it makes sense to develop a more comprehensive mock for `Topology`. I'm thinking something along the lines of:
```
it('should do something with the mock topology', function(done) {
   const client = new MongoClient(...);
   const topology = mockTopology(client); // <-- `client` being passed in anticipates a future where `client.topology` is set during construction, for now you would have to stub the `Topology` class. 
   client.connect(err => {
      client.db('test').collection('foo').createIndex({ quote: 'text' }, err => {
         expect(err).to.not.exist;
         expect(topology).property('operations[0]').to.equal({
           /*
             `operations` would conform to some shape that we think gives us all the info we need for assertions, maybe:
             `{ readPreference: ..., command: ... }` to start
           */
         });

         client.close(done);
      });
   });
});
```

I think we can do this work as a follow-on PR, maybe another approach is better but I think we definitely want to ensure we have tests for all commands which we expect to explicitly force a primary `readPreference`

test/functional/collection.test.js
```diff
@@ -1307,4 +1312,34 @@ describe('Collection', function() {
...
+      client.connect((err, client) => {
+        expect(err).to.not.exist;
+        const db = client.db(configuration.db);
+        const collection = db.collection('db-two');
+        TopologyStub = sinon.spy(Topology.prototype, 'selectServer');
+        collection.createIndex({ quote: 'text' }, (err, res) => {
+          expect(err).to.not.exist;
+          expect(res).to.equal('quote_text');
+          expect(TopologyStub.called).to.equal(true);
+          expect(TopologyStub.args[0][0].readPreference.mode).to.equal(ReadPreference.PRIMARY);
```
mbroadst, line 1335:
```suggestion
          expect(TopologyStub).nested.property('args[0][0].readPreference.mode').to.equal(ReadPreference.PRIMARY);
```
prefer using chai's nested `property` assertion helper to better inform error messages when things go wrong:
```

File: test/functional/collection.test.js [hunk 4]
```diff
@@ -1307,4 +1312,34 @@ describe('Collection', function() {
...
+          client.close(done);
+        });
+      });
+    }
+  });
+
+  afterEach(() => {
```
mbroadst, line 1342: you can also remove this `afterEach` if you use the `mocha-sinon` plugin

mbroadst (changes requested):

test/functional/collection.test.js
```diff
@@ -1307,4 +1312,34 @@ describe('Collection', function() {
...
+      const configuration = this.configuration;
+      const client = configuration.newClient(configuration.writeConcernMax(), {
+        useUnifiedTopology: true,
+        readPreference: 'primaryPreferred'
+      });
+      client.connect((err, client) => {
+        expect(err).to.not.exist;
+        const db = client.db(configuration.db);
+        const collection = db.collection('db-two');
+        TopologyStub = sinon.spy(Topology.prototype, 'selectServer');
```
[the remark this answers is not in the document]
reggi: I think a spy makes sense here because I'm not looking to change how the `selectServer` method works, simply checking to see what argument was passed into it. Is there some reason why I'd want to create & call a fake version of the `selectServer` method?

test/functional/collection.test.js
```diff
@@ -1313,33 +1313,41 @@ describe('Collection', function() {
...
-          expect(res).to.equal('quote_text');
-          expect(TopologyStub.called).to.equal(true);
-          expect(TopologyStub.args[0][0].readPreference.mode).to.equal(ReadPreference.PRIMARY);
-          client.close(done);
-        });
-      });
-    }
-  });
+  context.only('DLL methods with serverSelection readPreference primary', () => {
+    const primaryReadPreferenceDLL = {
```
reggi, line 1314: @mbroadst is this what you mean by data-driven?

test/functional/collection.test.js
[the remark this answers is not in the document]
mbroadst: I think a spy is fine, was just calling out that you're calling the spy a stub and what a stub would otherwise look like

test/functional/collection.test.js
mbroadst: yep!

mbroadst (approved):

test/functional/collection.test.js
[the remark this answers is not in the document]
mbroadst: Yeah I think a spy is totally fine, I was just calling out that you're assigning a spy to a variable named "stub", and showing what a stub would otherwise look like

nbbeeken (approved):
Solution LGTM! 🚀 
do we wanna hold off on merging the vscode settings until we've chatted in tech chat?

test/functional/collection.test.js
```diff
@@ -1307,4 +1309,42 @@ describe('Collection', function() {
       });
     }
   });
+
+  context('DLL methods with serverSelection readPreference primary', () => {
```
emadum, line 1313: I think this should be `DDL` rather than `DLL`?

emadum (changes requested):

reggi:
@nbbeeken good catch, that was accidental :(

@emadum also good catch, updated!

nbbeeken (approved):

test/functional/collection.test.js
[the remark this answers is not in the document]
reggi: Gotcha, renamed!

emadum (approved):
LGTM

PR #2546, opened 2020-09-21
refactor: move find command building to FindOperation

mbroadst:
The primary work of this ticket was to make the `Collection#find` method behave like other read operations by moving the command construction to the `FindOperation` itself. Along the way a number of improvements were made:

 - cursors now know they are tailable and/or awaitData upon construction, rather than observing the internal `cmd` they are built around. As much as possible the cursor implementation now no longer alters behavior after initialization based on what is in the internal `cmd` document

 - a number of supported, but undocumented options were added to the `find` method itself

 - `bsonOptions` were introduced to the `OperationBase` class as a first step towards improving our story around overriding these values and passing them to wire protocol methods

NODE-2830

### DIFF
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -48,3 +48,24 @@ export interface BSONSerializeOptions extends SerializeOptions {
 
   raw?: boolean;
 }
+
+export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSerializeOptions {
+  const {
+    fieldsAsRaw,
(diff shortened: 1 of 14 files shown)


reggi (approved):
Wow, this is great I really like:

* The `Collection#find` method doesn't have any logic
* Plucking the bson options and making it a keyd option

LGTM

emadum (approved):
LGTM!

PR #2547, opened 2020-09-21
fix: user roles take single string & DDL readPreference tests

reggi:
[NODE-2832](https://jira.mongodb.org/browse/NODE-2832) 

* Annotated some main points in comments. 
* This PR adds DDL tests for several collection / db methods to ensure that a primary `readPreference` is used.
* During my research I noticed that `Collection#rename` doesn't have a `write` aspect, it's still on the old `Operation`, and still works by forcing a primary `readPreference` in the collection method handler.
* This PR also adds `.vscode` and `output` to `.gitignore` as well as deletes the output file.


test/functional/readpreference.test.js
```diff
@@ -655,3 +656,77 @@ describe('ReadPreference', function() {
...
+    const methods = {
+      'Collection#createIndex': [{ quote: 'text' }],
+      'Db#createIndex': [COLLECTION, { quote: 'text' }],
+      'Db#addUser': ['thomas', 'pass', { roles: 'readWrite' }],
+      'Db#removeUser': ['default'],
+      'Db#createCollection': ['created_collection'],
+      'Db#dropCollection': ['before_collection'],
+      'Collection#dropIndex': ['aloha_1'],
+      'Collection#rename': ['new_name'],
+      'Db#dropDatabase': []
```
reggi, line 691: Adds DDL tests here in `readpreference.test.js`

lib/operations/add_user.js
```diff
@@ -22,7 +22,11 @@ class AddUserOperation extends CommandOperation {
     const options = this.options;
 
     // Get additional values
-    let roles = Array.isArray(options.roles) ? options.roles : [];
+    let roles = Array.isArray(options.roles)
+      ? options.roles
+      : typeof options.roles === 'string'
+      ? [options.roles]
+      : [];
```
reggi, line 29: This was a bug I picked up by using the TS interface as a guide, this seems like it was / is the intention, also is a bug in master (needs port)

.gitignore
```diff
@@ -47,3 +47,5 @@ build/Release
 node_modules
 yarn.lock
 
+.vscode
+output
```
reggi, line 51: Kept noticing this random `output` file, must have sneaked in somehow this can prevent that.

lib/operations/add_user.js
mbroadst: :+1: but I think convention has been to break out these nested ternaries into explicit if statements (I think we used to enable an eslint rule for this, and it's been lost in the shuffle?).

test/functional/readpreference.test.js
```diff
@@ -655,3 +656,77 @@ describe('ReadPreference', function() {
...
   });
 });
+
+describe.only('ReadPreference - operations', () => {
+  const COLLECTION = 'ddl_collection';
+  beforeEach(function(done) {
+    const configuration = this.configuration;
+    const client = configuration.newClient(configuration.writeConcernMax(), {
+      useUnifiedTopology: true,
+      readPreference: 'primaryPreferred'
```
mbroadst, line 666: why is this `primaryPreferred`?

test/functional/readpreference.test.js
```diff
@@ -655,3 +656,77 @@ describe('ReadPreference', function() {
     });
   });
 });
+
+describe.only('ReadPreference - operations', () => {
```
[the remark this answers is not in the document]
mbroadst: also does this need to be its own `describe` block? Seems like it could be nested within `ReadPreference` (and indeed is named as if it was). What about:
```
describe('ReadPreference', function() {
  context('write operations', function() {
    beforeEach(() => ...);
    Object.keys(..).forEach(operation => {
      it('should... ', () => ...);
    });
  });
});
```

test/functional/readpreference.test.js
```diff
@@ -655,3 +656,77 @@ describe('ReadPreference', function() {
...
+  const COLLECTION = 'ddl_collection';
+  beforeEach(function(done) {
+    const configuration = this.configuration;
+    const client = configuration.newClient(configuration.writeConcernMax(), {
+      useUnifiedTopology: true,
+      readPreference: 'primaryPreferred'
+    });
+    client.connect((err, client) => {
```
mbroadst, line 668: all of the methods in this function seem to ignore the possibility for an `err`, except for `connect`. Consider using `withClient` to ensure client is cleaned up, and adding assertions for the non-existence of `err`.

test/functional/readpreference.test.js
```diff
@@ -655,3 +656,77 @@ describe('ReadPreference', function() {
...
+        db.createCollection('before_collection', () => {
+          db.createIndex(COLLECTION, { aloha: 1 }, () => {
+            client.close(done);
+          });
+        });
+      });
+    });
+  });
+
+  context('DDL methods with serverSelection readPreference primary', () => {
```
mbroadst, line 681: as noted above, I think we can remove one `context` here

test/functional/readpreference.test.js
```diff
@@ -655,3 +656,77 @@ describe('ReadPreference', function() {
...
+      'Db#addUser': ['thomas', 'pass', { roles: 'readWrite' }],
+      'Db#removeUser': ['default'],
+      'Db#createCollection': ['created_collection'],
+      'Db#dropCollection': ['before_collection'],
+      'Collection#dropIndex': ['aloha_1'],
+      'Collection#rename': ['new_name'],
+      'Db#dropDatabase': []
+    };
+    Object.keys(methods).forEach(operation => {
+      it(`should run ${operation} with serverSelection readPreference primary`, {
```
mbroadst, line 694: For readability of future maintainers, I think saying something more like `should enforce fixed primary read preference`. That might even be the title of the `context`, ad the `it` should just note the operation name?

test/functional/readpreference.test.js
```diff
@@ -655,3 +656,77 @@ describe('ReadPreference', function() {
...
+        metadata: {
+          requires: { topology: 'replicaset' }
```
mbroadst, line 696: we want to test this in sharded environments as well.

test/functional/readpreference.test.js
```diff
@@ -655,3 +656,77 @@ describe('ReadPreference', function() {
...
+            readPreference: 'primaryPreferred'
+          });
+          client.connect((err, client) => {
+            const args = methods[operation];
+            const [parentId, method] = operation.split('#');
+            expect(err).to.not.exist;
+            const db = client.db(configuration.db);
+            const collection = db.collection(COLLECTION);
+            const parent = parentId === 'Collection' ? collection : parentId === 'Db' ? db : null;
+            const TopologySpy = this.sinon.spy(Topology.prototype, 'selectServer');
```
mbroadst, line 711: nit: the code formatter is formatting this as a class name, I tend to think of these spies as being "instances":
```
const selectServerSpy = this.sinon.spy(Topology.prototype, 'selectServer');
```

mbroadst (changes requested):

test/functional/readpreference.test.js
reggi: It just needs to not be set to `primary` and `primaryPreferred` was what the user ticket had in it.

mbroadst (approved):
LGTM!

emadum (approved):

PR #2548, opened 2020-09-21
Hidden Indexes (NODE-2591)

reggi:
[NODE-2591](https://jira.mongodb.org/browse/NODE-2591?filter=-1)

<S>PR to implement the feature "hidden indexes". 

Uses a new `HideIndexOperation` to send the command specified in [SPEC-1678's scope](https://docs.google.com/document/d/1uCF7iGtW70XMZDqBQgf7cK-PI425pqtbVa7CtBR0eR4/edit#). This one operation handles both hiding indexes and unhiding index. I didn't see the point in creating two new operations when it's just a matter of adding a boolean option.

I didn't have access to the collection name within the operation so I added it in from the parent, this is because the `ns` replaces the collection with `$cmd`. Would love to know a better way of doing this if there is one.

Adds `hidden` option to `createIndex` and test for that as well.

I made an attempt at adding in the `'expireAfterSeconds'` option to `.hideIndex()` but when sent I got this error:

>Uncaught AssertionError: expected [MongoError: no expireAfterSeconds field to update] to not exist

So I figured it may not need to be added.</S>

Adds `hidden` option to `IndexOptions`.


reggi:
@mbroadst Can you give me some guidance to see if I'm on track here?

mbroadst:
@reggi where did the requirement to make `hideIndex`/`unhideIndex` helpers come from? I'm not seeing that in the [spec changes](https://github.com/mongodb/specifications/commit/fafb0076e3c2ee8c35497bd1260c1cfc672c4919) linked to the DRIVERS ticket. In adherence to the driver mantra [minimize administrative helpers](https://github.com/mongodb/specifications/#minimize-administrative-helpers) I would think we would not provide them. 

The [index management](https://github.com/mongodb/specifications/blob/master/source/index-management.rst) spec is focused on "immutable" interaction with indexes, even though the `collMod` method exists and allows indexes to be modified in place the goal of this spec change was to support a `hidden` property for index creation only.

mbroadst (approved):

emadum (approved):

PR #2549, opened 2020-09-25
refactor: remove cursorState from Cursor and wire protcol methods

mbroadst:
Cursors presently have an internal state tracking object called `cursorState` which has a high degree of coupling with multiple components in the system, and is source of great confusion and bugs. This patch removes the `cursorState` completely, mounting its relevant properties directly on the `CoreCursor` type and removing it from the signature of the `query`, `killCursors`, and `getMore` wire protocol methods.

Many changes were required as part of this refactoring, but most notable is cleanup around cursor `close`. The internal call to kill the cursor raced with the `_endsession` method, which uncovered a number of failures in our test suite when fixed.

NODE-2810

**NOTE:** this is based on #2546 , so it will appear to have the same changes until that PR is merged and this one is rebased.

### DIFF
--- a/src/cmap/wire_protocol/kill_cursors.ts
+++ b/src/cmap/wire_protocol/kill_cursors.ts
@@ -3,52 +3,28 @@ import { maxWireVersion, collectionNamespace, Callback } from '../../utils';
 import { command, CommandOptions } from './command';
 import { MongoError, MongoNetworkError } from '../../error';
 import type { Server } from '../../sdam/server';
-import type { InternalCursorState } from '../../cursor/core_cursor';
-import type { ClientSession } from '../../sessions';
-
-interface KillCursorOptions {
-  session?: ClientSession;
-  immediateRelease: boolean;
-  noResponse: boolean;
-}
+import type { Long } from '../../bson';
 
 export function killCursors(
   server: Server,
   ns: string,
-  cursorState: InternalCursorState,
+  cursorIds: Long[],
+  options: CommandOptions,
   callback: Callback
 ): void {
   callback = typeof callback === 'function' ? callback : () => undefined;
-
-  if (!cursorState.cursorId) {
-    callback(new MongoError('Invalid internal cursor state, no known cursor id'));
-    return;
+  if (!cursorIds || !Array.isArray(cursorIds)) {
+    throw new TypeError('Invalid list of cursor ids provided: ' + cursorIds);
   }
-
-  const cursorIds = [cursorState.cursorId];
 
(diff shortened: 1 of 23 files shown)


src/cmap/wire_protocol/kill_cursors.ts
```diff
@@ -3,52 +3,28 @@ import { maxWireVersion, collectionNamespace, Callback } from '../../utils';
...
   if (maxWireVersion(server) < 4) {
     const pool = server.s.pool;
     const killCursor = new KillCursor(ns, cursorIds);
-    const options: KillCursorOptions = {
-      immediateRelease: true,
-      noResponse: true
-    };
 
-    if (typeof cursorState.session === 'object') {
-      options.session = cursorState.session;
```
reggi, line 39: Are we sure we want to stop passing the session here?

src/cursor/core_cursor.ts
```diff
@@ -176,93 +173,65 @@ export class CoreCursor<
...
+    this.tailable = typeof options.tailable === 'boolean' ? options.tailable : false;
+    this.awaitData =
+      typeof options.awaitData === 'boolean'
+        ? options.awaitData
+        : typeof options.awaitdata === 'boolean'
+        ? options.awaitdata
+        : false;
+
+    // get rid of these?
+    this._limit = limit;
```
reggi, line 197: `_limit` and `_batchSize` seem to be used. What's the purpose of this comment and the deprecation  comment above? Are these meant to be `private`? If so can we use private keyword?

reggi (approved):
Wow, this has a lot of good learnings in it. LGTM

emadum (approved):
LGTM, some nice test cleanup in here 👍

src/cmap/wire_protocol/kill_cursors.ts
mbroadst: Hey, sorry it looks like we lost your comments somewhere (maybe the force push). So what's going on here is that the wire protocol `killCursors` command used to take in a `cursorState` which might have had a session, and now it takes an `options` which may have a session. The `options` are passed all the way down to the `command` method which will use the session there, so there's no need to copy it here any longer

src/cursor/core_cursor.ts
mbroadst: They are used, but won't be for much longe (referring to the cursor prototype I discussed in our meeting yesterday). For now, I wanted to keep these around to make @emadum 's work of combining `Cursor` and `CoreCursor` easier.

PR #2551, opened 2020-09-26
fix: permit waking async interval with unreliable clocks

mbroadst:
The logic for waking the AsyncInterruptibleInterval sooner than its interval is dependent on an ability to reliably mark the last call made to the wrapped function. In environments like AWS Lambda where instances can be frozen and later thawed, it's possible for the last call to be in a distant past even though the internal timer has not completed yet. This change ensures that we immediately reschedule in these situations.

NODE-2829


reggi (approved):
LGTM 🤯

emadum (approved):

PR #2552, opened 2020-09-29
fix: use options for readPreference in client

reggi:
NODE-2807

### DIFF
--- a/lib/mongo_client.js
+++ b/lib/mongo_client.js
@@ -188,7 +188,7 @@ Object.defineProperty(MongoClient.prototype, 'writeConcern', {
 Object.defineProperty(MongoClient.prototype, 'readPreference', {
   enumerable: true,
   get: function() {
-    return ReadPreference.primary;
+    return ReadPreference.fromOptions(this.s.options) || ReadPreference.primary;
   }
 });
 
--- a/lib/mongo_client.js
+++ b/lib/mongo_client.js
@@ -169,6 +169,7 @@ function MongoClient(url, options) {
     dbCache: new Map(),
     sessions: new Set(),
     writeConcern: WriteConcern.fromOptions(options),
+    readPreference: ReadPreference.fromOptions(options) || ReadPreference.primary,
     namespace: new MongoDBNamespace('admin')
   };
 }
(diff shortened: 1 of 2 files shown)


reggi:
Should I write a test for this?

lib/mongo_client.js
```diff
@@ -188,7 +188,7 @@ Object.defineProperty(MongoClient.prototype, 'writeConcern', {
 Object.defineProperty(MongoClient.prototype, 'readPreference', {
   enumerable: true,
   get: function() {
-    return ReadPreference.primary;
+    return ReadPreference.fromOptions(this.s.options) || ReadPreference.primary;
```
mbroadst, line 191: This is a property that will likely be checked very often (most operations), I think a better approach would be to resolve the `ReadPreference.fromOptions` in the constructor, save it as private/symbol property of the `MongoClient` class itself, and return the cached value here instead of computing it each time. 

Also, to answer your question on the main PR: yes, there should be tests for this change

mbroadst (changes requested):

lib/mongo_client.js
reggi: Hey Matt! Great idea caching this! Added cache + test.

test/functional/mongo_client.test.js
```diff
@@ -3,6 +3,7 @@
 var f = require('util').format;
 var test = require('./shared').assert;
 var setupDatabase = require('./shared').setupDatabase;
+const { ReadPreference } = require('../..');
```
mbroadst, line 6: The convention so far has been to require directly from the defining file (in this case `../../read_preference') . I think the concern has been mostly about the potential for circular dependency cycles

test/functional/mongo_client.test.js
```diff
@@ -831,4 +832,10 @@ describe('MongoClient', function() {
         }
       });
   });
+
+  it('should have readPreference accessor from options', function() {
```
mbroadst, line 836: I think we can help our future selves a little here by improving this test name, maybe something like:
```suggestion
  it('should cache a resolved readPreference from options', function() {
```

mbroadst (approved):
LGTM mod two nits

mbroadst (approved):
LGTM

nbbeeken (approved):
LGTM

emadum (approved):

mbroadst:
@reggi what's holding up merging this?

PR #2554, opened 2020-09-29
fix: user roles take single string & DDL readPreference tests (master)

reggi:
Bringing in changes from https://github.com/mongodb/node-mongodb-native/pull/2547


test/functional/readpreference.test.js
```diff
@@ -250,7 +263,9 @@ describe('ReadPreference', function () {
 
     test: function (done) {
       var configuration = this.configuration;
-      var client = configuration.newClient(configuration.writeConcernMax(), { poolSize: 1 });
+      var client = configuration.newClient(configuration.writeConcernMax(), {
+        poolSize: 1
+      });
```
nbbeeken, line 268: Why are these changing if the code is the same? Maybe double check prettier version?

nbbeeken (changes requested):
Looks good, consistent with the other PR, except some formatting that I think is a mistake? lmk

test/functional/readpreference.test.js
reggi: fixed!

test/functional/readpreference.test.js
```diff
@@ -587,4 +588,144 @@ describe('ReadPreference', function () {
...
+              done();
+            };
+            parent[method].apply(parent, [...args, callback]);
+          });
+        }
+      });
+    });
+  });
+
+  context('should enforce fixed primary read preference', function () {
```
mbroadst, line 662: it looks like you have two copies of the same test additions

mbroadst (changes requested):

test/functional/readpreference.test.js
reggi: doh, I was wrestling with this file / linting and manually doubled up.

mbroadst (approved):
LGTM

nbbeeken (approved):
LGTM

emadum (approved):

PR #2555, opened 2020-09-29
test: Update connection-monitoring-and-pooling specs

nbbeeken:
Update connection-monitoring-and-pooling tests
ensuring ConnectionCreated proceed ConnectionReady events

NODE-2804

 ### Note to reviewers
I have a spec sync script I suggest we bring in here, if anyone has a hesitation we can drop that.
There were a few formatting issues that made it to 3.6, the fixes are in here.

### DIFF
--- a/etc/update-spec-tests.sh
+++ b/etc/update-spec-tests.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# This script is used to fetch the latest JSON tests for the CRUD spec. It puts the tests in the
+# direcory $reporoot/data/crud. It should be run from the root of the repository.
+
+set -o errexit
+set -o nounset
+
+if [ ! -d ".git" ]; then
+    echo "$0: This script must be run from the root of the repository" >&2
+    exit 1
+fi
+
+if [ $# -ne 1 ]; then
+    echo "$0: This script must be passed exactly one argument for which tests to sync" >&2
+    exit 1
+fi
+
+spec_root="test/spec"
+
+tmpdir=$(mktemp -d -t spec_testsXXXX)
+curl -sL https://github.com/mongodb/specifications/archive/master.zip -o "$tmpdir/specs.zip"
+unzip -d "$tmpdir" "$tmpdir/specs.zip" > /dev/null
+mkdir -p "$spec_root/$1"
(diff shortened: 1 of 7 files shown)


reggi (approved):
LGTM, I like the update script 🤩

The only issue is that we may not want to pull master, I am usually cherry-picking files from a specific commit hash, or pulling all changes to a spec folder from a commit hash. I've never considered taking from master.

etc/update-spec-tests.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+# This script is used to fetch the latest JSON tests for the CRUD spec. It puts the tests in the
```
mbroadst, line 3: nit: these details aren't correct, it's not specific to the CRUD tests, and the $reproroot variable is not used.

mbroadst (approved):
LGTM with a nit for the sync script

emadum (approved):

etc/update-spec-tests.sh
nbbeeken: Updated! Thanks 😃

nbbeeken:
> LGTM, I like the update script 🤩
> 
> The only issue is that we may not want to pull master, I am usually cherry-picking files from a specific commit hash, or pulling all changes to a spec folder from a commit hash. I've never considered taking from master.

@reggi PTAL at the update script (even better would be if you could give it a whirl for me). Should be able to support a specific git hash as an optional second argument.

etc/update-spec-tests.sh
```diff
@@ -0,0 +1,26 @@
...
+    exit 1
+fi
+
+spec_root="test/spec"
+
+tmpdir=$(mktemp -d -t spec_testsXXXX)
+curl -sL "https://github.com/mongodb/specifications/archive/master.zip" -o "$tmpdir/specs.zip"
+unzip -d "$tmpdir" "$tmpdir/specs.zip" > /dev/null
+mkdir -p "$spec_root/$1"
+rsync -ah "$tmpdir/specifications-master/source/$1/tests/" "$spec_root/$1" --delete
```
reggi, line 25:
```suggestion
rsync -ah --exclude '*.rst' "$tmpdir/specifications-master/source/$1/tests/" "$spec_root/$1" --delete
```

PR #2557, opened 2020-09-30
fix: adds topology discovery for mongos

reggi:
NODE-2827

### DIFF
--- a/lib/core/uri_parser.js
+++ b/lib/core/uri_parser.js
@@ -687,15 +687,6 @@ function parseConnectionString(uri, options, callback) {
     return callback(new MongoParseError('directConnection option requires exactly one host'));
   }
 
-  // NOTE: this behavior will go away in v4.0, we will always auto discover there
-  if (
-    parsedOptions.directConnection == null &&
-    hosts.length === 1 &&
-    parsedOptions.replicaSet == null
-  ) {
-    parsedOptions.directConnection = true;
-  }
-
   const result = {
     hosts: hosts,
     auth: auth.db || auth.username ? auth : null,
--- a/test/functional/sharding_connection.test.js
+++ b/test/functional/sharding_connection.test.js
@@ -10,7 +10,7 @@ describe('Sharding (Connection)', function() {
     return setupDatabase(this.configuration);
   });
 
-  it.only('Should use sharded topology', {
+  it('Should use sharded topology', {
     metadata: { requires: { topology: 'sharded' } },
     test: function() {
       const client = this.configuration.newClient({}, { useUnifiedTopology: true });
--- a/test/functional/sharding_connection.test.js
+++ b/test/functional/sharding_connection.test.js
@@ -1,11 +1,28 @@
 'use strict';
 
+const withClient = require('./shared').withClient;
 const setupDatabase = require('./shared').setupDatabase;
 const expect = require('chai').expect;
+const TopologyType = require('../../lib/core/sdam/common').TopologyType;
 
 describe('Sharding (Connection)', function() {
   before(function() {
     return setupDatabase(this.configuration);
+  });
+
+  it.only('Should use sharded topology', {
+    metadata: { requires: { topology: 'sharded' } },
+    test: function() {
+      const client = this.configuration.newClient({}, { useUnifiedTopology: true });
+      expect(client.s.options).to.have.property('useUnifiedTopology', true);
+      return withClient(client, (client, done) => {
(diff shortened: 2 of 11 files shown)


lib/core/uri_parser.js
```diff
@@ -687,6 +687,8 @@ function parseConnectionString(uri, options, callback) {
     return callback(new MongoParseError('directConnection option requires exactly one host'));
   }
 
+  parsedOptions._directConnection = Boolean(parsedOptions.directConnection);
```
reggi, line 690: Open changing the naming here, we need a way to know if `directConnect` is defaulted to `true` or not.

lib/core/uri_parser.js
```diff
@@ -687,6 +687,8 @@ function parseConnectionString(uri, options, callback) {
...
+
```
mbroadst, line 691: isn't the fix here to just remove the code below in the `// NOTE` section?

reggi: My reasoning for not removing that code is this issue was reported in 3.6, I thought we'd want to fix it there not issue a breaking change. 

> Since changing the starting topology can reasonably be considered a backwards-breaking change, existing drivers SHOULD stage implementation according to semantic versioning guidelines. Specifically, support for the directConnection URI option can be added in a minor release. In a subsequent major release, the default starting topology can be changed to Unknown. Drivers MUST document this in a prior minor release.

[server-discovery-and-monitoring.rst](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst)

mbroadst: It's a bug reported in 3.6 using an opt-in feature where we broke compatibility with the existing behavior, I don't think anybody is depending on the bug's behavior. We should just fix it in 3.6 now, and maintain the new behavior as-is in `master`

test/functional/sharding_connection.test.js
```diff
@@ -1,13 +1,30 @@
...
 const setupDatabase = require('./shared').setupDatabase;
 const expect = require('chai').expect;
+const TopologyType = require('../../lib/core/sdam/common').TopologyType;
 
 describe('Sharding (Connection)', function() {
   before(function() {
     return setupDatabase(this.configuration);
   });
 
+  it.only('Should use sharded topology', {
```
emadum, line 13:
```suggestion
  it('Should use sharded topology', {
```

emadum (changes requested):

test/functional/sharding_connection.test.js
reggi: doh!

nbbeeken (approved):

emadum (changes requested):
Looks like there are two SDAM spec tests that are causing CI to fail.

test/functional/sharding_connection.test.js
```diff
@@ -1,13 +1,30 @@
...
 describe('Sharding (Connection)', function() {
   before(function() {
     return setupDatabase(this.configuration);
   });
 
+  it('Should use sharded topology', {
+    metadata: { requires: { topology: 'sharded' } },
+    test: function() {
+      const client = this.configuration.newClient({}, { useUnifiedTopology: true });
+      expect(client.s.options).to.have.property('useUnifiedTopology', true);
```
mbroadst, line 17: prefer not to observe private properties, I think it's satisfactory to have the check for the topology description `type` below.

test/functional/sharding_connection.test.js
```diff
@@ -1,13 +1,30 @@
...
+      return withClient(client, (client, done) => {
+        expect(client).to.exist;
+        expect(client.topology).to.exist;
+        expect(client.topology.description).to.exist;
+        expect(client.topology.description).to.have.property('type', TopologyType.Sharded);
```
mbroadst, line 22:
```suggestion
        expect(client).nested.property('topology.description.type').to.equal(TopologyType.Sharded);
```

mbroadst (approved):
LGTM, I left two notes about how I think this can be slightly improved, but no need to hold up merging the change

reggi:
> Looks like there are two SDAM spec tests that are causing CI to fail.

@emadum I don't think it's useful to put in a change request that the tests are failing, because the pull will be blocked by that, I approve PR's all the time that have failing tests I think it's implied. Just a thought here, because now that the tests are passing I'm still blocked by your change request.

emadum (approved):
LGTM 👍

PR #2558, opened 2020-09-30
fix: Fix test filters and revert mocha version

nbbeeken:
Our tests were all being filtered out on node 4 by the client_encryption_filter.
This also rolls back mocha to run on older nodejs versions.


test/tools/runner/filters/client_encryption_filter.js
```diff
@@ -43,10 +43,23 @@ class ClientSideEncryptionFilter {
...
+      }
+      if (!test.metadata.requires) {
+        test.metadata.requires = {};
+      }
+      if (!test.metadata.requires.node) {
+        test.metadata.requires.node = '>4';
+      } else if (!semver.satisfies(process.version, '>4')) {
+        test.metadata.requires.node = '>4';
+      }
+    }
```
mbroadst, line 59: A filter should not mutate the code it's observing. I suspect what's really wrong here has something to do with `ret` below, but the goal of this filter is that IFF `clientSideEncryption` is attached to a test or test suite then it should apply these filtering rules, otherwise the filter should be skipped.

mbroadst (changes requested):

test/tools/runner/filters/client_encryption_filter.js
nbbeeken: I completely agree, ran to the fast solution before the best one here. 
It will now return true for all tests that don't specify `clientSideEncryption` and check the node version if `clientSideEncryption` if defined.

reggi (approved):
LGTM 😭

mbroadst (approved):
LGTM!

PR #2562, opened 2020-10-02
npm-script-test-by-topologies

reggi:
* Moves `cluster_setup.sh` scripts into package scripts
* Adds directories (--dir) to allow for multiple topologies to exist at once
* Adds init / start / stop scripts for individual topologies and all
* Pre-loads `MONGODB_URL` env var to run test for a given topology


reggi:
we discussed maintaining this all in one script

PR #2563, opened 2020-10-04
fix: connection leak if waitque cancelled

jswangjunsheng:
## Description
cmap/connection_pool
connection may be leak if waitque cancelled

**What changed?**


mbroadst:
Thanks for the contribution @jswangjunsheng !

PR #2564, opened 2020-10-05
fix: Change socket timeout default to 0

nbbeeken:
Socket timeout by default should be infinity.
Removed 6min defaults.

NODE-2835

### DIFF
--- a/src/cmap/connect.ts
+++ b/src/cmap/connect.ts
@@ -277,7 +277,7 @@ function makeConnection(
       : typeof options.connectTimeoutMS === 'number'
       ? options.connectTimeoutMS
       : 30000;
-  const socketTimeout = typeof options.socketTimeout === 'number' ? options.socketTimeout : 360000;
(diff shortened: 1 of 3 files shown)


reggi (approved):
LGTM

Maybe worth adding a `@default` tag to the Client options interface.

https://github.com/mongodb/node-mongodb-native/blob/master/src/mongo_client.ts#L78

nbbeeken:
> LGTM
> 
> Maybe worth adding a `@default` tag to the Client options interface.
> 
> https://github.com/mongodb/node-mongodb-native/blob/master/src/mongo_client.ts#L78

I would but as of now TSDoc doesn't have _default_ 😅  support for this tag https://github.com/microsoft/tsdoc/issues/27, we could always enable it as a custom tag but we need a pipeline that does something with the information. 

I can use the `@default` tag when I port this to 3.6 however! I'll be sure to remember that.

emadum (approved):
👍

mbroadst:
@nbbeeken I would have expected some tests here, can you please add one/some?

PR #2566, opened 2020-10-07
test: adds tests for atlas data lake

reggi:
NODE-2691

### DIFF
--- a/src/sdam/topology.ts
+++ b/src/sdam/topology.ts
@@ -365,7 +365,7 @@ export class Topology extends EventEmitter {
 
       // TODO: NODE-2471
       if (server && this.s.credentials) {
-        server.command('admin.$cmd', { ping: 1 }, err => {
+        server.command('admin.$cmd', { ping: 1, thisPingShouldBeIgnored: 1 }, err => {
           if (err) {
             typeof callback === 'function' ? callback(err) : this.emit(Topology.ERROR, err);
             return;
--- a/src/sdam/topology.ts
+++ b/src/sdam/topology.ts
@@ -365,7 +365,7 @@ export class Topology extends EventEmitter {
 
       // TODO: NODE-2471
       if (server && this.s.credentials) {
-        server.command('admin.$cmd', { ping: 1, thisPingShouldBeIgnored: 1 }, err => {
+        server.command('admin.$cmd', { ping: 1 }, err => {
           if (err) {
             typeof callback === 'function' ? callback(err) : this.emit(Topology.ERROR, err);
             return;
--- a/test/functional/spec-runner/context.js
+++ b/test/functional/spec-runner/context.js
@@ -31,13 +31,13 @@ class Thread {
 
 class TestRunnerContext {
   constructor(opts) {
-    const d = {
+    const defaults = {
       password: undefined,
       user: undefined,
       useSessions: true,
       skipPrepareDatabase: false
     };
-    opts = Object.assign({}, d, opts || {});
+    opts = Object.assign({}, defaults, opts || {});
     this.skipPrepareDatabase = opts.skipPrepareDatabase;
     this.useSessions = opts.useSessions;
     this.user = opts.user;
--- a/test/functional/spec-runner/context.js
+++ b/test/functional/spec-runner/context.js
@@ -31,7 +31,12 @@ class Thread {
 
 class TestRunnerContext {
   constructor(opts) {
-    const d = { password: undefined, user: undefined, useSessions: true, skipPrepareDatabase: false };
+    const d = {
+      password: undefined,
+      user: undefined,
+      useSessions: true,
+      skipPrepareDatabase: false
+    };
     opts = Object.assign({}, d, opts || {});
     this.skipPrepareDatabase = opts.skipPrepareDatabase;
     this.useSessions = opts.useSessions;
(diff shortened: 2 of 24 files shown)


test/functional/spec-runner/index.js
```diff
@@ -289,6 +291,10 @@ function runTestSuiteTest(configuration, spec, context) {
     if (IGNORED_COMMANDS.has(event.commandName)) {
       return;
     }
+    // TODO: This should be removed if NODE-2471 accomidates it
+    if (event.command.thisPingShouldBeIgnored) {
```
mbroadst, line 295: it seems like a more general solution would be to just drop the first ping if credentials are provided?

src/sdam/topology.ts
```diff
@@ -365,7 +365,7 @@ export class Topology extends EventEmitter {
 
       // TODO: NODE-2471
       if (server && this.s.credentials) {
-        server.command('admin.$cmd', { ping: 1 }, err => {
+        server.command('admin.$cmd', { ping: 1, thisPingShouldBeIgnored: 1 }, err => {
```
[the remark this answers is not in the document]
reggi: Would love to chat more about this, open to any solutions, Shane and I came up with this one. 

I originally did ignore all ping commands. This was a tricky one to solve because there is a ping check in one of the spec tests: 

https://github.com/mongodb/specifications/blob/master/source/atlas-data-lake-testing/tests/runCommand.yml

That test runs with auth, so we can't implement your comment below:

> it seems like a more general solution would be to just drop the first ping if credentials are provided?

Would love your feedback on how to move forward on this.

reggi:
@mbroadst I've put out a fix for all the things mentioned here except for a viable solution to the contradictory `ping` exceptions.

test/manual/data_lake.test.js
```diff
@@ -0,0 +1,69 @@
...
+    before(function () {
+      return testContext.setup(this.configuration);
+    });
+
+    generateTopologyTests(testSuites, testContext);
+  });
+
+  describe('prose Tests', function () {
+    it('should properly constructs and issues a killCursors command', function () {
+      return withClient('mongodb://mhuser:pencil@localhost', (client, done) => {
```
nbbeeken, line 30: Does the withClient have to be called? like `withClient(...)()` or will it be since its returned? Just want to make sure the expectations are being run.

reggi: Darn, got me, thanks for the catch. I want to re-write this `withClient` thing so badly, ugg.

src/sdam/topology.ts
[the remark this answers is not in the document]
mbroadst: > I originally did ignore all ping commands.

I was saying if a test runs with auth then we should skip the _first_ recorded ping operation, not all of them. We know that the driver right now will _always_ send an initial `ping` if credentials are provided, so it seems safe for us to ignore them in that case

test/functional/spec-runner/index.js
```diff
@@ -307,22 +314,24 @@ function runTestSuiteTest(configuration, spec, context) {
...
-      );
 
-      savedSessionData = {
-        session0: JSON.parse(EJSON.stringify(session0.id)),
-        session1: JSON.parse(EJSON.stringify(session1.id))
-      };
-    } catch (err) {
-      // ignore
-    }
+    if (context.useSessions) {
```
mbroadst, line 318: Did you try optionally creating these if the entities exist? (i.e. if `session0` or `session1` are specified) It seems like a preferable solution to feature flagging in this particular case. I was recommending the feature flag for cases which couldn't be resolved this way (like skip init)

reggi: I did try to create them if the entities exist, it created a bug in the transactions suite. I had to dig deep into it but apparently those test require the object to exist in order to work correctly.

src/sdam/topology.ts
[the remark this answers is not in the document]
reggi: Done!

test/functional/spec-runner/context.js
```diff
@@ -30,8 +30,13 @@ class Thread {
...
 class TestRunnerContext {
-  constructor() {
-    this.url = null;
+  constructor(opts) {
+    const d = { password: undefined, user: undefined, skipInit: false, useSessions: true };
+    opts = Object.assign({}, d, opts || {});
+    this.useSessions = opts.useSessions;
+    this.skipInit = opts.skipInit;
+    this.user = opts.user;
+    this.password = opts.password;
```
nbbeeken, line 39: I suppose useSessions is pretty self explanatory, but skipInit could use a bit of description, like what "init" exactly is being skipped. might as well put a little text saying when these props are useful.

nbbeeken (approved):
LGTM-ing with a request for some comments on the new TestRunnerContext class.

test/functional/spec-runner/index.js
reggi: I just tried again, the transaction tests do not have `spec.sessionOptions` set but do require a session, that code to run.

test/functional/spec-runner/context.js
```diff
@@ -30,8 +30,18 @@ class Thread {
 }
 
 class TestRunnerContext {
-  constructor() {
-    this.url = null;
+  constructor(opts) {
+    const d = {
```
emadum, line 34: Nit: I try to avoid single letter variable names to improve readability, could we change this to `defaults`?

emadum (approved):
LGTM with one nit. 👍

test/functional/spec-runner/context.js
reggi: good rule 👍

PR #2567, opened 2020-10-07
fix: readPreferenceTags type and client readPreference getter

reggi:
NODE-2807

port of https://github.com/mongodb/node-mongodb-native/pull/2552

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -17,6 +17,7 @@ import type { AuthMechanism } from './cmap/auth/defaultAuthProviders';
 import type { Topology } from './sdam/topology';
 import type { ClientSession, ClientSessionOptions } from './sessions';
 import type { OperationParent } from './operations/command';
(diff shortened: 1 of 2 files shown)


src/mongo_client.ts
```diff
@@ -295,7 +298,7 @@ export class MongoClient extends EventEmitter implements OperationParent {
   }
 
   get readPreference(): ReadPreference {
-    return ReadPreference.primary;
+    return this.s.readPreference;
```
nbbeeken, line 301: I know this is a port, so sorry about not bringing this up on the original, but it occurred to me that this didn't effect any tests? should we throw in a quick 'respects user provided preference' test?

nbbeeken (approved):
One q, lgtm tho!

src/mongo_client.ts
reggi: did write 3.6 test, missed it in port TY!

mbroadst (approved):

PR #2568, opened 2020-10-07
cluster script enahancements

reggi:
This is my preferred cluster script, contains the subcommands `init|start|stop|prune|kill|uri|test`.

* `init` - uses `mlaunch init` to setup topology in a port folder
* `start` - uses `mlaunch start` but if that command fails because hasn't been inited, runs `mlaunch init`!
* `stop` - uses `mlaunch stop` to stop a specific topology
* `prune` - stops a given topology, then clears out it's data dir
* `kill` - Uses https://github.com/sindresorhus/fkill-cli to kill all ports related to that topology
* `uri` - `exports` and prints topology `MONGODB_URI`
* `test` - Runs start, ensuring that the topology is up, then loads in the uri, then runs just test for 3.6 branch or 4.0 with `--if-present` flag, linting takes forever, so removed that. This will not shut anything down, and is relatively quick, I'm open to having one command that does this, and another that JUST loads in uris and runs test.

Proposes `tool:` npm script prefix for this and `spec-update`.


nbbeeken (commented):
Just some suggestions and feedback, looking good! 🚀

package.json
```diff
@@ -65,6 +65,8 @@
     "url": "https://github.com/mongodb/node-mongodb-native/issues"
   },
   "scripts": {
+    "tool:spec-update": "sh ./etc/update-spec-tests.sh",
+    "tool:cluster": "sh ./test/tools/cluster.sh",
```
nbbeeken, line 69: Since we're using bash specific syntax we can just launch with bash here

test/tools/cluster.sh
```diff
@@ -0,0 +1,173 @@
...
+
+log () {
+  echo "$1 --------------------"
+}
+
+cluster () {
+    COMMAND="$1"
+    if [[ "$COMMAND" == "uri" ]]; then
+        TOPOLOGY=$2
+        if [[ "$TOPOLOGY" == "replica" ]] || [[ "$TOPOLOGY" == "replicaset" ]] || [[ "$TOPOLOGY" == "replica_set" ]]; then
```
nbbeeken, line 20: You could use a regex here:
```bash
if [[ "$TOPOLOGY" =~ ^(replica|replicaset|replica_set)$ ]]; then
```
I appreciate the flexibility of this, I would use `standalone`, `replicaset`, and `cluster` but I'm glad we can support whatever people prefer 👍

test/tools/cluster.sh
```diff
@@ -0,0 +1,173 @@
...
+            cluster uri sharded
+        elif [[ "$TOPOLOGY" == "standalone" ]] || [[ "$TOPOLOGY" == "server" ]] || [[ "$TOPOLOGY" == "single" ]]; then
+            log 'init single'
+            mlaunch init --dir "./data/$PORT_SINGLE" --single --setParameter enableTestCommands=1
+            cluster uri single
+        else
+            cluster init replica
+            cluster init sharded
+            cluster init single
+        fi
```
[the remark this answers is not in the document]
reggi: I still think it can be useful 1) as the dependency to start and 2) understanding how `mlaunch` works. I'm down to hide it from the menu, but I don't think it's necessary.

package.json
reggi: done

test/tools/cluster.sh
reggi: Yes! I tried this syntax was wonky, but will try again, prob missing a character. Was about to ask for help on this, syntax is super repetitive. I often forget what the "proper" term is that the script uses. I usually go with `single`, `replica`, and `sharded` so I like this feature.

reggi: Fixed this with your solution + functions! pretty minimal now, and easy to add / remove options.

test/tools/cluster.sh
```diff
@@ -0,0 +1,198 @@
...
+            npx -q fkill-cli :$PORT_SINGLE -fs
+        else
+            cluster kill replica
+            cluster kill sharded
+            cluster kill single
+        fi
+      elif [[ "$COMMAND" == "test" ]]; then
+        TOPOLOGY=$2
+        if is_replica $TOPOLOGY; then
+            cluster start replica && cluster uri replica && npm run test-nolint --if-present -- "${@:3}" && npm run check:test --if-present -- "${@:3}"
```
reggi, line 183: Added "${@:3}" so that you can run:

`cluster test single --bail` ect and extra args pass through to mocha

test/tools/cluster.sh
```diff
@@ -0,0 +1,198 @@
+#!/bin/bash
```
nbbeeken, line 1:
```suggestion
#!/usr/bin/env bash
```
Makes this a bit more portable

nbbeeken (approved):
Looks great, one last thing is the script could be made executable, if you could push the `chmod` on that.

test/tools/cluster.sh
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env bash
+
+if ! command -v mlaunch &> /dev/null
```
reggi, line 3: @nbbeeken I snuck this in last minute is this cross platform?

nbbeeken: Yea this is fine! Actually this script also needs mongod/s to be in the user's path as well if you want to check for that also.

mbroadst (commented):
I don't think we should merge this, it's just a layer of commands on top of `mlaunch` which we would now need to support moving forward.

PR #2569, opened 2020-10-07
refactor: Remove options generic parameter on OperationBase

nbbeeken:
Instead of maintaining the typing for bespoke options per operation,
each operation has the option properties assigned directly to itself.

NODE-2815

### Description

Each operation now `implements` its options interface to make sure we maintain the correct properties on the respective classes. There were two special cases in map_reduce and  create_collections where we used a blockList to omit properties from the command, with the new scheme it made more sense to make these into allowLists. 

I tried to minimize the impact outside of the operations subdirectory since Cursor is undergoing a number of changes, you'll notice a cast in CoreCursor and Cursor that blatantly overrides type warnings, it is my belief that we can remove this as part of the new cursor design work. 

Another exception to the options removal was the addition of the bsonOptions getter, this I think could be better but looking for feedback: the cursor looks for a `bsonOptions` property (separate from the options property) to pass through to getMore execution (tricky bug arose here where the first doc fetched by the cursor was deserialized with the user's options and the following one's were deserialized with the defaults). My solution was to make a bespoke bsonOptions object in a getter on OperationsBase despite all the options being defined on the operation directly. I could move this logic into the cursor function that fetches these options instead?

Lastly, and you'll notice a comment on it but the last line of executeCommands in `src/bulk/common.ts` we seem to swallow an error, maybe that is correct, but when fixing another bug I was having tests that had both an undefined `err` and undefined `result` which feels like odd behavior.


src/operations/add_user.ts
```diff
@@ -18,11 +18,16 @@ export interface AddUserOptions extends CommandOperationOptions {
...
 
 /** @internal */
-export class AddUserOperation extends CommandOperation<AddUserOptions, Document> {
+export class AddUserOperation extends CommandOperation implements AddUserOptions {
   db: Db;
   username: string;
   password?: string;
 
+  /** @deprecated Please use db.command('createUser', ...) instead for this option */
+  digestPassword?: null;
```
reggi, line 27: Can you add the deprecated props to the bottom of this list? This can read as `digestPassword` is the only deprecated, or all that follow are as well.

src/operations/bulk_write.ts
```diff
@@ -11,10 +17,21 @@ import type {
...
-export class BulkWriteOperation extends OperationBase<BulkWriteOptions, BulkWriteResult> {
+export class BulkWriteOperation
+  extends OperationBase<BulkWriteResult>
+  implements BulkWriteOptions, HasRetryableWrites, HasWriteConcern {
   collection: Collection;
   operations: AnyBulkWriteOperation[];
 
+  bypassDocumentValidation?: boolean;
+  ordered?: boolean;
+  /** @deprecated use `ordered` instead */
```
reggi, line 28: This one for `@deprecated` too.

src/operations/find_and_modify.ts
```diff
@@ -33,21 +34,37 @@ export interface FindAndModifyOptions extends CommandOperationOptions {
...
   collection: Collection;
   query: Document;
   sort?: Sort;
   doc?: Document;
 
+  new?: boolean;
+  remove?: boolean;
+  upsert?: boolean;
+  projection?: Document;
+  /** @deprecated Use projection instead */
```
reggi, line 59: Here too

src/operations/map_reduce.ts
```diff
@@ -96,55 +103,52 @@ export class MapReduceOperation extends CommandOperation<MapReduceOptions, Docum
...
     }
 
     // Add any other options passed in
-    for (const n in options) {
-      // Only include if not in exclusion list
-      if (exclusionList.indexOf(n) === -1) {
-        mapCommandHash[n] = (options as any)[n];
+    for (const n in this) {
+      // Only include if defined on this
+      if (optionsAllowList.has(n)) {
```
reggi, line 120: The other ops that do this check for null values / functions wondering if there's a way we can gain consistency here?

src/operations/operation.ts
```diff
@@ -34,23 +34,42 @@ export interface OperationOptions extends BSONSerializeOptions {
...
+    const bsonOptions: Document = {
+      promoteBuffers: this.promoteBuffers,
+      promoteValues: this.promoteValues,
+      promoteLongs: this.promoteLongs,
+      raw: this.raw,
+      ignoreUndefined: this.ignoreUndefined
+    } as const;
+
+    for (const key in bsonOptions) {
+      bsonOptions[key] === undefined && delete bsonOptions[key];
```
reggi, line 64: Can we do this without using `delete`? I believe we try to stay away from it when possible.

reggi: I do like this getter BTW 🤩

src/operations/replace_one.ts
```diff
@@ -19,15 +19,16 @@ export interface ReplaceOptions extends CommandOperationOptions {
   /** When true, creates a new document if no document matches the query */
   upsert?: boolean;
 
-  // FIXME:
+  // FIXME: ????!!!! neal why?
```
reggi, line 22: 😿

reggi: Maybe the intention is to have a `replace` operation that does both, or a `replace_multi`?

reggi (approved):
LGTM, some little comments here and there but nothing stands out 🙇‍♂️

Really appreciate this work, it's one step closer toward getting rid of `.s` or `.options` and having properties attached to the class is much easier to reason about. Cleaner classes!

src/operations/create_collection.ts
```diff
@@ -10,18 +10,19 @@ import type { PkFactory } from '../mongo_client';
 // eslint-disable-next-line
 import type { Collection } from '../collection';
 
-const ILLEGAL_COMMAND_FIELDS = new Set([
-  'w',
-  'wtimeout',
-  'j',
-  'fsync',
+const optionsAllowList = new Set([
```
emadum, line 13:
```suggestion
const OPTIONS_ALLOW_LIST = new Set([
```

I think our convention is to format top-level constants in uppercased snake case.

src/operations/map_reduce.ts
```diff
@@ -15,16 +15,17 @@ import type { Sort } from './find';
...
-  'readPreference',
-  'session',
-  'bypassDocumentValidation',
-  'w',
-  'wtimeout',
-  'j',
-  'writeConcern',
-  'scope' // this option is reformatted thus exclude the original
-];
+const optionsAllowList = new Set([
```
emadum, line 18:
```suggestion
const OPTIONS_ALLOW_LIST = new Set([
```

src/operations/map_reduce.ts
emadum, line 120:
```suggestion
      if (OPTIONS_ALLOW_LIST.has(n)) {
```

src/operations/create_collection.ts
```diff
@@ -74,25 +89,20 @@ export class CreateCollectionOperation extends CommandOperation<
...
     const cmd: Document = { create: name };
-    for (const n in options) {
-      if (
-        (options as any)[n] != null &&
-        typeof (options as any)[n] !== 'function' &&
-        !ILLEGAL_COMMAND_FIELDS.has(n)
-      ) {
-        cmd[n] = (options as any)[n];
+    for (const n in this) {
+      if (optionsAllowList.has(n) && this[n] != null && typeof this[n] !== 'function') {
```
emadum, line 104:
```suggestion
      if (OPTIONS_ALLOW_LIST..has(n) && this[n] != null && typeof this[n] !== 'function') {
```

src/operations/indexes.ts
```diff
@@ -189,18 +187,18 @@ export class CreateIndexesOperation extends CommandOperation<CreateIndexesOption
...
         );
         return;
       }
-      cmd.commitQuorum = options.commitQuorum;
+      cmd.commitQuorum = this.commitQuorum;
     }
 
     // collation is set on each index, it should not be defined at the root
-    this.options.collation = undefined;
+    // this.collation = undefined; FIXME!!!!!!
```
emadum, line 201: What's going on here?

src/operations/operation.ts
```diff
@@ -64,20 +83,8 @@ export abstract class OperationBase<
...
-    // NOTE: Using the bang operator here because we know there is always a
-    //       session, explicit or implicit. We should disambiguate the session
-    //       from the options and set it as an explicit field
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    return this.options.session!;
-  }
-
   clearSession(): void {
-    delete this.options.session;
+    delete this.session;
```
emadum, line 87: Echoing @reggi's comments, do want to use `delete` here or should we set it to `undefined`?

src/operations/operation.ts
```diff
@@ -34,23 +34,42 @@ export interface OperationOptions extends BSONSerializeOptions {
...
+    }
+
+    return bsonOptions;
+  }
 
-  constructor(options: T = {} as T) {
-    this.options = Object.assign({}, options);
-    this.readPreference = ReadPreference.primary;
+  constructor(options = {}) {
+    // this.readPreference = ReadPreference.primary;
```
emadum, line 71: This looks like it needs to be cleaned up, was this a temporary measure or do we want to delete this line?

src/operations/replace_one.ts
emadum: 🤔

test/functional/insert.test.js
```diff
@@ -1891,9 +1889,14 @@ describe('Insert', function () {
...
-                var doc = docs.pop();
 
-                test.ok(doc.a._bsontype === 'Long');
+                docs.forEach((d, i) => {
+                  expect(d.a, `Failed on the document at index ${i}`).to.not.be.a('number');
+                  expect(d.a, `Failed on the document at index ${i}`).to.have.property('_bsontype');
+                  expect(d.a._bsontype, `Failed on the document at index ${i}`).to.be.equal('Long');
+                });
+
+                // expect(doc.a).to.have.property('_bsontype').to.be.equal('Long');
```
emadum, line 1899: Can you explain why these changes were necessary? Is this still testing the same thing?

emadum (changes requested):
LGTM, this looks great. 🎉 Just a few nits/questions.

src/operations/indexes.ts
emadum: The `Successfully create index with collation` test is failing in CI, probably related to this?

src/bulk/common.ts
```diff
@@ -640,7 +640,7 @@ function executeCommands(
     err.ok = 0;
     // Merge top level error and return
     mergeBatchResults(batch, bulkOperation.s.bulkResult, err, undefined);
-    callback();
```
emadum, line 643: This seems like it was probably an oversight, why would we want to swallow the error here?

test/functional/insert.test.js
nbbeeken: The bsonOptions weren't being inherited properly by the cursor, the first document would respect the options because it received them from the options parameter in the cursor's constructor. But subsequent getMores would get the bsonOptions off the operation 

https://github.com/mongodb/node-mongodb-native/blob/master/src/cursor/core_cursor.ts#L588

and this was just undefined and so we would fill in defaults that were different, so the following documents would have different deserialization. 

Anyway I changed the test to check that all of the documents `a` properties remain Long class objects rather than number primitives instead of just the first.

src/operations/operation.ts
nbbeeken: Cleaned up the readPreference was previously "overwritten" if the options object specified it, but now it will be a property directly on the class.

src/operations/indexes.ts
nbbeeken: Yep fixed this, this was actually some cruft from TS conversion worth mentioning to spread the knowledge about it (@reggi), we used this syntax in few places:
```ts
class A {
  prop!: string;
}
```
because at the time the errors surrounding prop being optional were numerous and it felt right to assert the prop's presence, I just pushed changing props on `CommandOperation` that were asserted to be optional and there are no more type errors surrounding that. So we removed the dependence on the properties with further work that followed the big operations layer conversion.

src/operations/replace_one.ts
nbbeeken: Git blame says I left the FIXME there thats why I left 'neal why?' lol oops! looks like the option doesn't need to be there, it isn't a replaceOne option but it is an update option, maybe during typing this interface was being shared.

src/operations/map_reduce.ts
nbbeeken: Is there more ops than mapReduce that do this? I think MapReduce does this since the operation takes in functions it makes sense to double check the options don't have any (for example, `validator` sounds like a function but is a document). I can get more context on this if there's more

PR #2571, opened 2020-10-08
refactor!: remove deprecated find options

reggi:
* Unsure if this was done correctly.
* Do we want to remove deprecated options from the options object as well?


emadum (changes requested):
The builder methods also need to be removed, see e.g. [maxScan](https://github.com/mongodb/node-mongodb-native/blob/fe4821cb06a21c1a2a7d7350a5cd4dab7e5ebf0b/src/cursor/cursor.ts#L261-L274). Can you also run an Evergreen patch and link it here, since it's not automated on PRs not based on master?

reggi:
* [ ] Evergreen, will add check when passes: https://spruce.mongodb.com/version/5f8483980ae6061d5c01f597/tasks

emadum (approved):
LGTM, but I'm wondering why the entire `deprecate_warning.test.js` was deleted? That seems like it's intended to generally ensure our deprecation utilities are working correctly, and we should probably update it to use some other deprecated fields rather than just removing it.

reggi:
> LGTM, but I'm wondering why the entire `deprecate_warning.test.js` was deleted? That seems like it's intended to generally ensure our deprecation utilities are working correctly, and we should probably update it to use some other deprecated fields rather than just removing it.

I went crazy looking for those fields in the tests, I didn't realize it was a stub and doesn't use find, re-enabled.

nbbeeken (approved):

mbroadst:
@reggi it looks like you still deleted the `deprecate_warning.test.js` file, was that intentional?

reggi:
Total accident. Didn't push the update. https://github.com/mongodb/node-mongodb-native/pull/2585

PR #2573, opened 2020-10-09
refactor!: format sort in cursor and in sort builder

reggi:


src/operations/common_functions.ts
```diff
@@ -106,28 +99,6 @@ export function prepareDocs(
   });
 }
 
-// Get the next available document from the cursor, returns null if no more documents are available.
-export function nextObject(cursor: Cursor, callback: Callback): void {
```
reggi, line 110: dead code?

src/sort.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (SortDigest.isPair(sort)) return SortDigest.pairToObject(sort);
+    if (SortDigest.isDeep(sort)) return SortDigest.deepToObject(sort);
+    if (Array.isArray(sort)) return SortDigest.stringsToObject(sort);
+    if (typeof sort === 'object') return SortDigest.validate(sort);
+    throw new Error(`Invalid sort format: ${JSON.stringify(sort)}`);
+  }
+}
+
+/** converts a Sort type into a type that is valid for the server (SortForCmd) */
+export const formatSort = SortDigest.prepare;
```
reggi, line 94: Thought it would be good to keep this as a black box.

test/functional/cursor.test.js
```diff
@@ -405,74 +406,66 @@ describe('Cursor', function () {
...
+                cursor.close(done);
+              } else {
+                cursor.close();
               }
             };
 
             var cursor = collection.find().sort(['a', 1]);
-            test.deepEqual(['a', 1], cursor.sortValue);
-            finished();
+            test.deepEqual({ a: 1 }, cursor.sortValue);
```
reggi, line 420: these had to change, it no longer just places the value it formats & validates it.

test/functional/cursor.test.js
```diff
@@ -405,74 +406,66 @@ describe('Cursor', function () {
...
 
             cursor = collection.find().sort('a', 'asc');
-            test.deepEqual([['a', 'asc']], cursor.sortValue);
-            finished();
-
-            cursor = collection.find().sort([
-              ['a', -1],
-              ['b', 1]
-            ]);
-            var entries = cursor.sortValue.entries();
```
reggi, line 436: `.entries()` 👀 no clue what was going on here.

test/functional/cursor.test.js
```diff
@@ -405,74 +406,66 @@ describe('Cursor', function () {
...
-              .next(err => {
-                test.equal(
-                  "Illegal sort clause, must be of the form [['field1', '(ascending|descending)'], ['field2', '(ascending|descending)']]",
-                  err.message
-                );
-                finished();
-              });
+            cursor = collection.find();
+            try {
+              cursor.sort(25);
```
reggi, line 462: This now throws a sync error, which is a breaking change?

src/sort.ts
```diff
@@ -0,0 +1,94 @@
...
+/** Below stricter types were created for sort that correspond with type that the cmd takes  */
+
+/** @internal */
+type SortDirectionForCmd = 1 | -1 | { $meta: string };
+
+/** @internal */
+type SortForCmd = { [key: string]: SortDirectionForCmd };
+
+/** @internal */
+class SortDigest {
```
nbbeeken, line 28: Since all the methods are static so whats the motivation for making this a class?

src/sort.ts
```diff
@@ -0,0 +1,94 @@
...
+  static pairToObject(v: [string, SortDirection]): SortForCmd {
+    return { [v[0]]: SortDigest.prepareDirection(v[1]) };
+  }
+  static isDeep(t: Sort): t is [string, SortDirection][] {
+    return Array.isArray(t) && Array.isArray(t[0]);
+  }
+  static deepToObject(t: [string, SortDirection][]): SortForCmd {
+    return t.reduce((acq, i) => {
+      return { ...acq, ...SortDigest.pairToObject(i) };
+    }, {});
```
nbbeeken, line 68: It's unlikely that anyone will have large sort specifiers but nonetheless this nested looping feels like something to avoid despite the potentially small input.

src/sort.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+  static validate(t: { [key: string]: SortDirection }): SortForCmd {
+    return Object.keys(t).reduce((acq, key) => {
+      return { ...acq, [key]: SortDigest.prepareDirection(t[key]) };
+    }, {});
+  }
+  static prepare(sort: Sort | undefined, direction?: SortDirection): SortForCmd | undefined {
+    if (sort == null) return undefined;
+    if (Array.isArray(sort) && !sort.length) return undefined;
+    if (typeof sort === 'object' && !Object.keys(sort).length) return undefined;
```
nbbeeken, line 83: [1000% nit]: I would use `x.length !== 0` but that is a total 🧅

test/functional/cursor.test.js
nbbeeken: It returns an iterator of tuples `[index, value]`, then each next call will iterate through the array emitting this.

test/functional/cursor.test.js
```diff
@@ -4312,4 +4305,64 @@ describe('Cursor', function () {
...
+    it('should use find options string', findSort('alpha', { alpha: 1 }));
+    it('should use find options shallow array', findSort(['alpha', 1], { alpha: 1 }));
+    it('should use find options deep array', findSort([['alpha', 1]], { alpha: 1 }));
+
+    it('should use cursor.sort object', cursorSort({ alpha: 1 }, { alpha: 1 }));
+    it('should use cursor.sort string', cursorSort('alpha', { alpha: 1 }));
+    it('should use cursor.sort shallow array', cursorSort(['alpha', 1], { alpha: 1 }));
+    it('should use cursor.sort deep array', cursorSort([['alpha', 1]], { alpha: 1 }));
+
+    it('formatSort', () => {
```
nbbeeken, line 4350: Should there be some more tests that try multiple sort keys?
Also repeated keys in the list of tuples form will implicitly be reduced to one since we use an object, let's just add a case to make sure its as expected.

nbbeeken (changes requested):
Great start, a few comments

src/sort.ts
reggi: I'm confused, this seems essential, how else would you convert `[['a', 1], ['b', 'desc']]` to `{a: 1, b: -1}`?

src/sort.ts
reggi: How would this proposal work?

* `({ 'neal': true }).length` => undefined
* `Object.keys({ 'neal': true }).length` => 1

test/functional/cursor.test.js
reggi: `cursor.sortValue` is typed as Sort, just the option passed into sort from the user, there's no guarantees that `.entries()` will work.

This would have broke:

```
cursor = collection.find().sort({ a: 1 });	
var entries = cursor.sortValue.entries();
```

Can we deprecate this functionality?

test/functional/cursor.test.js
reggi: added some multi tests

src/sort.ts
nbbeeken: Wouldn't this be equivalent without the inner loop?
```ts
let sortObject = {};
for (const [name, value] of t) {
  sortObject[name] = prepareDirection(value);
}
return sortObject
```

src/sort.ts
nbbeeken: Since we haven't done it elsewhere I think its best to leave them as top level functions, the fact that they live in their own file I think already provides the boxing you're looking for.

src/sort.ts
nbbeeken: Makes sense! since we just export `prepare`, maybe we should name it `formatSort` this goes along with my comment of pulling the functions to the top level of the module, my thinking its a little more ctrl-F friendly this way.

test/functional/cursor.test.js
nbbeeken: Yea that makes sense, it won't work anymore, I don't think it was functionality per se just the tester knew it was still an array on the inside 👍

src/sort.ts
nbbeeken: I should clarify I meant in addition to what you have.
if you already check `Array.isArray(sort)` is true then sort will have a length prop that is a number.
`Object.keys(sort)` returns an array so you can check the length === 0 on that too.

src/sort.ts
reggi: done!

src/sort.ts
reggi: cool, done

src/sort.ts
reggi: Hmmm, I'm 100% down to do this, not sure I'm fully understanding, can you give me a suggestion using the github feature / code sample?

src/sort.ts
```diff
@@ -0,0 +1,110 @@
...
+  direction?: SortDirection
+): SortForCmd | undefined {
+  if (sort == null) return undefined;
+  if (Array.isArray(sort) && !sort.length) return undefined;
+  if (typeof sort === 'object' && !Object.keys(sort).length) return undefined;
+  if (typeof sort === 'string') return { [sort]: prepareDirection(direction) };
+  if (isPair(sort)) return pairToObject(sort);
+  if (isDeep(sort)) return deepToObject(sort);
+  if (Array.isArray(sort)) return stringsToObject(sort);
+  if (typeof sort === 'object') return validate(sort);
```
nbbeeken, line 108: [naming nit] `validate(..)` implies checking and returning a boolean for correctness, I would say `objectToObject` for consistency 😆  but that feels like trying to hard, idk a good name here, feel free to leave it if nothing comes to mind.

nbbeeken (changes requested):

src/sort.ts
```diff
@@ -0,0 +1,110 @@
...
+    return { ...acq, [key]: 1 };
+  }, {});
+}
+
+/** @internal */
+function validate(t: { [key: string]: SortDirection }): SortForCmd {
+  return Object.keys(t).reduce((acq, key) => {
+    return { ...acq, [key]: prepareDirection(t[key]) };
+  }, {});
+}
```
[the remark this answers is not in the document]
reggi: done, got rid of all reducers, too much lodash ingrained in my brain.

nbbeeken (approved):
🚀

src/sort.ts
reggi: done!

emadum (approved):
LGTM 👍

test/functional/cursor.test.js
emadum: I think this is fine, but we should probably add a `!` to the commit and make a note that sort direction is now being validated more strictly.

PR #2575, opened 2020-10-13
docs: Display ES2017 usage by default

outloudvi:
## Description

This PR makes ES2017 usages displayed by default.

It's 2020. `await` is [supported since Node.js v7.6.0](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await), which is released [over 3.5 years ago](https://nodejs.org/en/blog/release/v7.6.0/) (2017/02/22). Also, [93% of the browsers support `await` already](https://caniuse.com/mdn-javascript_operators_await). As a result, developers no longer need to be buried under the callback hell.


agolin95:
Tracked by https://jira.mongodb.org/browse/NODE-2892

kmahar:
Thanks for the contribution @outloudvi !

PR #2570, opened 2020-10-08
refactor: merge CoreCursor into Cursor

emadum:
NODE-2839

## Description

This PR removes `CoreCursor`, merging its functionality into the `Cursor` class that formerly extended it. This should set the stage for NODE-2809, which will refactor `Cursor` into `AbstractCursor`.

PR #2576, opened 2020-10-13
refactor: merge CoreCursor into Cursor

emadum:
NODE-2839

## Description

This PR removes `CoreCursor`, merging its functionality into the `Cursor` class that formerly extended it. This should set the stage for NODE-2809, which will refactor `Cursor` into `AbstractCursor`.

Note: replaces #2570 which looked bad, apologies @reggi for second review request, the changes are the same.


reggi (approved):
LGTM WOOT MERGE! 🌊🌊🌊🌊 🏔

src/cursor/cursor.ts
```diff
@@ -39,9 +43,58 @@ export type CursorFlag = typeof FLAGS[number];
...
+/** @public */
+export interface DocumentTransforms {
+  /** Transform each document returned */
+  doc(doc: Document): Document;
+  /** Transform the value returned from the initial query */
+  query?(doc: Document): Document | Document[];
+}
+
+/** @internal */
+export interface CoreCursorPrivate {
```
mbroadst, line 55: two questions:
- can we just rename this `CursorPrivate`, it looks like its an alias below
- do we need to export this internal type?

src/cursor/cursor.ts
```diff
@@ -39,9 +43,58 @@ export type CursorFlag = typeof FLAGS[number];
...
+
+  state: CursorState;
+  readConcern?: ReadConcern;
+}
+
 /** @internal */
 export type CursorPrivate = CoreCursorPrivate;
 
+/** @public Possible states for a cursor */
+export enum CursorState {
```
mbroadst, line 72: is `CursorState` used outside of this file? I know you didn't write this code, but I think a lot of the curious exporting we did with cursors was due to the fact that we had to split between `core_cursor` and `cursor`

src/cursor/cursor.ts
```diff
@@ -39,9 +43,58 @@ export type CursorFlag = typeof FLAGS[number];
...
+  INIT = 0,
+  OPEN = 1,
+  CLOSED = 2,
+  GET_MORE = 3
+}
+
+/** @public */
+export interface CoreCursorOptions extends CommandOperationOptions {
```
mbroadst, line 80: merge with `CursorOptions`

src/cursor/cursor.ts
```diff
@@ -54,8 +107,49 @@ export interface CursorOptions extends CoreCursorOptions {
...
+}
+
+/** @public */
+export interface CursorCloseOptions {
+  /** Bypass calling killCursors when closing the cursor. */
+  skipKillCursors?: boolean;
+}
+
+/** @internal */
+export interface InternalCursorState extends BSONSerializeOptions {
```
mbroadst, line 123: This type appears to be unused as well

src/cursor/cursor.ts
```diff
@@ -54,8 +107,49 @@ export interface CursorOptions extends CoreCursorOptions {
   numberOfRetries?: number;
 }
 
+/** @public */
+export interface StreamOptions {
```
mbroadst, line 111: this duplicates `CursorStreamOptions` below

src/cursor/cursor.ts
```diff
@@ -205,13 +405,13 @@ export class Cursor<
     const currentNumberOfRetries = numberOfRetries;
 
     // Get the batchSize
-    let batchSize = 1000;
+    let batchSizeA = 1000;
```
mbroadst, line 408: Why the rename to `batchSizeA`?  The `batchSize` pulled from `getLimitSkipBatchSizeDefaults` above seems at odds with this logic here now. Perhaps the one above should be the default, and you can omit the declaration here, only using the following logic to override the default if need be.

src/cursor/cursor.ts
```diff
@@ -264,7 +492,207 @@ export class Cursor<
       }
     }
 
-    super._initializeCursor(callback);
+    this._initializeCoreCursor(callback);
+  }
+
+  /** @internal */
+  _initializeCoreCursor(callback: Callback): void {
```
mbroadst, line 499: I'm fine leaving `_initializeCursor` and `_initializeCoreCursor` split for just this PR, but hope that the plan is to merge the two?

src/cursor/cursor.ts
```diff
@@ -973,16 +1426,281 @@ export class Cursor<
   getLogger(): Logger {
     return this.logger;
   }
+
+  // Internal methods
+
+  /** @internal Retrieve the next document from the cursor */
+  _next(callback: Callback<Document>): void {
```
mbroadst, line 1433: I believe this method can go away now, `next()` above can call `nextFunction` directly.

src/sessions.ts
```diff
@@ -18,7 +18,7 @@ import {
 import type { Topology } from './sdam/topology';
 import type { MongoClientOptions } from './mongo_client';
 import type { Cursor } from './cursor/cursor';
-import type { CoreCursor } from './cursor/core_cursor';
+// import type { CoreCursor } from './cursor/core_cursor';
```
mbroadst, line 21: delete?

test/functional/spec-runner/index.js
```diff
@@ -125,9 +125,6 @@ function generateTopologyTests(testSuites, testContext, filter) {
       describe(suiteName, {
         metadata: { requires },
         test: function () {
-          beforeEach(() => prepareDatabaseForSuite(testSuite, testContext));
```
mbroadst, line 128: I reverted this change in the last commit, I think we should revert it here. I know this leads to confusing test output for skipped tests, but I would prefer to tackle that in a separate PR specifically focused on that problem. In fact, I think the issue might really be with the leak checker, and not how this test runner is defined. 

Please make a ticket for this (if it doesn't already exist) and revert this change

mbroadst (changes requested):

src/cursor/cursor.ts
emadum: Yeah it's used in a number of places current; in the CommandCursor and in several places in the operation layer.

src/cursor/cursor.ts
emadum: Currently both `CursorPrivate` and `CoreCursorPrivate` are [exported](https://github.com/mongodb/node-mongodb-native/blob/054838f00e75a1ed6e41f320c55d62fb3b63e583/src/index.ts#L149-L157) by `index.ts`, I wasn't sure why they were being exported or if I could safely fully delete them.

src/cursor/cursor.ts
emadum: Oops, I think this slipped back in during a merge, thanks for catching

src/cursor/cursor.ts
emadum: Same issue as above, is it OK to delete the export from `index.ts` for this?

src/cursor/cursor.ts
emadum: This was an oversight on my part, they are now merged.

src/cursor/cursor.ts
mbroadst: yeah feel free to do that, I think this is because `api-extractor` complained about them not having a visibility tag, but we shouldn't be exporting internals like this

src/cursor/cursor.ts
mbroadst: yep!

src/cursor/cursor.ts
```diff
@@ -54,8 +108,43 @@ export interface CursorOptions extends CoreCursorOptions {
...
 
-const kCursor = Symbol('cursor');
+/** @public */
+export interface CursorCloseOptions {
+  /** Bypass calling killCursors when closing the cursor. */
+  skipKillCursors?: boolean;
+}
+
+/** @internal */
+interface InternalCursorState extends BSONSerializeOptions {
```
mbroadst, line 118: looks like we lost my comment on this, it looks like `InternalCursorState` is no longer used

mbroadst (commented):
Lookin really good! Just two last little things

nbbeeken (approved):
Nice work! 🚀

mbroadst (approved):

PR #2578, opened 2020-10-13
feat!: adds async iterator for custom promises

reggi:
NODE-2590

* Fixes `promiseLibrary` deprecation notice from popping up if you use it in `MongoClient`
* Fixes type overload for `Cursor#close`
* Adds `AsyncIterator` to `Cursor`
* Re-enables `AsyncIterator` tests
* Tests `AsyncIterator` for custom promise
* Removes `PromiseProvider.set` in `MongoClient#connect`


src/cursor/cursor.ts
```diff
@@ -973,6 +974,27 @@ export class Cursor<
...
+          .then(() => {
+            return this.next();
+          })
+          .then(value => {
+            if (!value) {
+              return this.close().then(() => ({ value, done: true }));
+            }
+            return { value, done: false };
+          });
+      }
```
mbroadst, line 995: Outside of the topics I brought up in the team channel the other day, I think we can at least simplify this a little bit:
```
next: () => {
  if (this.isClosed()) {
    return Promise.resolve({ value: null, done: true });
  }
  
  return this.next().then(value => ({ value, done: value === null }));
}
```

When a cursor returns the value `null` it has been exhausted, so that's where the explicit check comes from

mbroadst (changes requested):
LGTM mod two small nits

mbroadst (approved):

emadum (approved):
LGTM! 👍

nbbeeken (approved):

PR #2579, opened 2020-10-13
test: ensure connection with sharded topology

reggi:
NODE-2827

bringing in test from https://github.com/mongodb/node-mongodb-native/pull/2557

### DIFF
--- a/test/functional/sharding_connection.test.js
+++ b/test/functional/sharding_connection.test.js
@@ -1,0 +1,24 @@
+'use strict';
+
+const withClient = require('./shared').withClient;
+const setupDatabase = require('./shared').setupDatabase;
+const expect = require('chai').expect;
+const TopologyType = require('../../lib/sdam/common').TopologyType;
+
+describe('Sharding (Connection)', function () {
+  before(function () {
+    return setupDatabase(this.configuration);
+  });
+
+  it('Should use sharded topology', {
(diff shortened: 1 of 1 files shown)


mbroadst (approved):

test/functional/sharding_connection.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+describe('Sharding (Connection)', function () {
+  before(function () {
+    return setupDatabase(this.configuration);
+  });
+
+  it('Should use sharded topology', {
+    metadata: { requires: { topology: 'sharded' } },
+    test: function () {
+      const client = this.configuration.newClient({});
```
emadum, line 16: Why was this necessary? If you're not passing custom options into `this.configuration.newClient`, this should be the same as `test: withClient((client, done) => {...})`

emadum (approved):
LGTM w/ one question about `withClient` usage

test/functional/sharding_connection.test.js
reggi: Updated! This is a port, and I believe I needed to pass in useUnifiedTopology in 3.6 and that's why I did it this way, but then possibly removed the option here because of the warning?

reggi: Damn, i missed this push locally. I think it's OK to leave in, hope we revamp and check through `withClient` calls later.

PR #2581, opened 2020-10-14

HanaPearlman:
`connect` doesn't work again after `close` has been called once on a `MongoClient`.

NODE-2544

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -346,19 +346,29 @@ export class MongoClient extends EventEmitter implements OperationParent {
     const force = typeof forceOrCallback === 'boolean' ? forceOrCallback : false;
 
     return maybePromise(callback, cb => {
+      const completeClose = (err?: AnyError) => {
+        // clear out references to old topology
+        this.topology = undefined;
+        this.s.dbCache = new Map();
+        this.s.sessions = new Set();
+
+        cb(err);
+      };
+
       if (this.topology == null) {
-        return cb();
+        completeClose();
+        return;
       }
 
       const topology = this.topology;
       topology.close({ force }, err => {
         const autoEncrypter = topology.s.options.autoEncrypter;
         if (!autoEncrypter) {
-          cb(err);
+          completeClose(err);
           return;
         }
 
-        autoEncrypter.teardown(force, err2 => cb(err || err2));
+        autoEncrypter.teardown(force, err2 => completeClose(err || err2));
       });
     });
   }
--- a/src/operations/connect.ts
+++ b/src/operations/connect.ts
@@ -199,7 +199,7 @@ export function connect(
 
   // Has a connection already been established?
   if (mongoClient.topology && mongoClient.topology.isConnected()) {
-    throw new Error(`'connect' cannot be called when already connected`);
+    throw new MongoError(`'connect' cannot be called when already connected`);
   }
 
   let didRequestAuthentication = false;
--- a/src/operations/connect.ts
+++ b/src/operations/connect.ts
@@ -197,6 +197,11 @@ export function connect(
     throw new Error('no callback function provided');
   }
 
+  // Has a connection already been established?
+  if (mongoClient.topology && mongoClient.topology.isConnected()) {
+    throw new Error(`'connect' cannot be called when already connected`);
+  }
+
   let didRequestAuthentication = false;
   const logger = new Logger('MongoClient', options);
 
--- a/test/functional/connection.test.js
+++ b/test/functional/connection.test.js
@@ -276,7 +276,7 @@ describe('Connection', function () {
   });
 
   it('should be able to connect again after close', function () {
-    return withClient.call(this, (client, done) => {
+    return withClient.bind(this)((client, done) => {
       expect(client.isConnected()).to.be.true;
 
       const collection = () => client.db('testReconnect').collection('test');
--- a/test/functional/connection.test.js
+++ b/test/functional/connection.test.js
@@ -287,9 +287,11 @@ describe('Connection', function () {
 
         client.close(err => {
           expect(err).to.not.exist;
+          expect(client.isConnected()).to.be.false;
 
           client.connect(err => {
             expect(err).to.not.exist;
+            expect(client.isConnected()).to.be.true;
 
             collection().insertOne({ b: 2 }, (err, result) => {
               expect(err).to.not.exist;
--- a/test/functional/connection.test.js
+++ b/test/functional/connection.test.js
@@ -2,6 +2,7 @@
 const test = require('./shared').assert,
   setupDatabase = require('./shared').setupDatabase,
   expect = require('chai').expect;
+const withClient = require('./shared').withClient;
 
 describe('Connection', function () {
   before(function () {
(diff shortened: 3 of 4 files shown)


HanaPearlman:
Note most of this work comes from PR #2355. I added comments corresponding to some unresolved comments/important points from that PR.

test/functional/connection.test.js
```diff
@@ -273,4 +274,31 @@ describe('Connection', function () {
       done();
     }
   });
+
+  it('should be able to connect again after close', function () {
+    return withClient.call(this, (client, done) => {
```
HanaPearlman, line 279: I'm using this instead of just `withClient((client, done) ...` because I couldn't see test output (warning messages) without this change.

src/mongo_client.ts
```diff
@@ -346,19 +346,29 @@ export class MongoClient extends EventEmitter implements OperationParent {
     const force = typeof forceOrCallback === 'boolean' ? forceOrCallback : false;
 
     return maybePromise(callback, cb => {
+      const completeClose = (err?: AnyError) => {
+        // clear out references to old topology
+        this.topology = undefined;
+        this.s.dbCache = new Map();
+        this.s.sessions = new Set();
+
```
HanaPearlman, line 354: Per a discussion from #2355, we don't clean up `this.s.options` here, so we still get output warnings when re-connecting:
```
the options [servers] is not supported
the options [caseTranslate] is not supported
the options [dbName] is not supported
```

test/functional/connection.test.js
reggi: @HanaPearlman Yes! this is super confusing, there's ticket in the back log to address this I added a comment in there about how to use this as I believe it was intended. [NODE-2764](https://jira.mongodb.org/browse/NODE-2764?focusedCommentId=3444002&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3444002)

HanaPearlman: Whoops, took a closer look at `withClient` and it seems like the right approach is to bind `withClient` to `this`.

HanaPearlman: @reggi Thanks for that! Happy to make that change now to replace the test callback or keep the `bind` syntax; not 100% clear on what is best...

src/mongo_client.ts
```diff
@@ -346,19 +346,29 @@ export class MongoClient extends EventEmitter implements OperationParent {
     const force = typeof forceOrCallback === 'boolean' ? forceOrCallback : false;
 
     return maybePromise(callback, cb => {
+      const completeClose = (err?: AnyError) => {
```
reggi, line 349: nit: naming of `completeClose`, perhaps something like `resetState` is more apt?

test/functional/connection.test.js
```diff
@@ -273,4 +274,31 @@ describe('Connection', function () {
       done();
     }
   });
+
+  it('should be able to connect again after close', function () {
+    return withClient.bind(this)((client, done) => {
```
reggi, line 279: `withClient` is totally wonk right now, but because you're not passing in any extra options to it, it can be used like this:

```
it('should be able to connect again after close', withClient(function (client, done) {
```

Note you can't use arrow function to maintain mocha's `this` scoping.

reggi (changes requested):
Just Some small things 👍

src/mongo_client.ts
HanaPearlman: Had a small offline discussion. Leaning towards leaving this as-is because `completeClose` both resets the state and calls the callback. Does anyone else have thoughts on this?

test/functional/connection.test.js
```diff
@@ -273,4 +274,32 @@ describe('Connection', function () {
...
+      const collection = () => client.db('testReconnect').collection('test');
+      collection().insertOne({ a: 1 }, (err, result) => {
+        expect(err).to.not.exist;
+        expect(result).to.exist;
+
+        client.close(err => {
+          expect(err).to.not.exist;
+
+          client.connect(err => {
+            expect(err).to.not.exist;
```
emadum, line 292: And another `expect(client.isConnected()).to.be.true;` here, just as a sanity check.

emadum (approved):
LGTM 👍 just suggested a few additional expectations in the new test

src/mongo_client.ts
mbroadst: I think this is fine for today. There's another unit of work here in the future (after the `MongoClientOptions` project) where the options will be stored immutably, in which case this problem goes away.

src/mongo_client.ts
mbroadst: I'm fine with `completeClose`, it's also a pattern we use elsewhere

src/operations/connect.ts
```diff
@@ -197,6 +197,11 @@ export function connect(
     throw new Error('no callback function provided');
   }
 
+  // Has a connection already been established?
+  if (mongoClient.topology && mongoClient.topology.isConnected()) {
+    throw new Error(`'connect' cannot be called when already connected`);
```
mbroadst, line 202: Let's not use plain `Error` objects, instead prefer a `TypeError` for simple type mismatches and `MongoError` or one of its subclasses for the rest. In this case I think a `MongoError` would be best. (I know the code right above this uses an `Error`, but let's be the change we want to see in the world right?)

test/functional/connection.test.js
```diff
@@ -273,4 +274,34 @@ describe('Connection', function () {
       done();
     }
   });
+
+  it(
+    'should be able to connect again after close',
+    withClient(function (client, done) {
+      expect(client.isConnected()).to.be.true;
+
+      const collection = () => client.db('testReconnect').collection('test');
```
mbroadst, line 283:
```suggestion
      const collection = client.db('testReconnect').collection('test');
```
This isn't actually just a style nit, we should expect saving off a `Collection` like this to work through `MongoClient` reconnects, so we actually want to make a single instance and use it twice.

mbroadst (changes requested):

test/functional/connection.test.js
HanaPearlman: I agree that this should work. Unfortunately, if we keep the same collection instance, then the topology used during the second `collection.insertOne()` references the original topology (which was closed during `client.close()`), causing a `MongoError: Topology is closed` for this test case. 

I believe this means we can't keep the same instance of `Db` or `Collection` around to use again, unless we only store the topology in the `MongoClient`, and whenever we want access to it in  a `Db` or `Collection`, we have to ask to `MongoClient` for it.

test/functional/sessions.test.js
```diff
@@ -123,7 +123,7 @@ describe('Sessions', function () {
               // verify that the `endSessions` command was sent
               const lastCommand = test.commands.started[test.commands.started.length - 1];
               expect(lastCommand.commandName).to.equal('endSessions');
-              expect(client.topology.s.sessionPool.sessions).to.have.length(0);
+              expect(client.topology).to.not.exist;
```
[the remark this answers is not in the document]
HanaPearlman: As part of closing the `MongoClient`, we set `client.topology` to `undefined`, so we cannot check the size of its session pool, we can just check that the topology no longer exists. 

I agree it's not the best but it seems like we _have_ to set the topology to `undefined` (instead of just setting the state of the topology to closed) because otherwise it's ambiguous whether the topology 1) just hasn't been opened yet or 2) was closed/all cleaned up.

reggi (approved):
LGTM

test/functional/connection.test.js
mbroadst: @HanaPearlman that's a really good point, and a change I think we should make. Can you make a new NODE ticket for only storing a reference to the parent `MongoClient` in `Db`/`Collection`?

test/functional/sessions.test.js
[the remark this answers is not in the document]
mbroadst: Hmm. I think an implicit part of this work is that `Topology` can be connected, closed and reused as well. A `Topology` basically _is_ a client, the `MongoClient` being a wrapper around it to provide access to higher level API. Ideally we would want to assign a `Topology` in the constructor of `MongoClient`, rather than during `connect` (but that's not quite possible today).

Have you tried not resetting `topology = undefined` yet? This might require changes to the `connect` operation. Let's not go down a huge rabbit hole here, but I'd like to better understand the limitations to that approach before opting for this alternative.

nbbeeken (approved):
🚀 🚀 🚀

test/functional/sessions.test.js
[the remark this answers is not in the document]
HanaPearlman: Yep, you're right, it is possible to _not_ set `topology = undefined` after a small change to the `connect` logic. This means we don't need to do any additional clean-up on client close, since any references to a `Db`/`Client` can use the same topology after re-connect. I've made this change, and will write a few tests to verify that the topology still receives the correct events after client re-connect.

mbroadst (approved):

PR #2582, opened 2020-10-14
fix: correctly re-establishes pipe destinations

reggi:
NODE-2172

I tried pretty hard to recreate this locally, and I ran into issues, I was also considering stubbing out the `#pipe` here and testing the `createChangeStreamCursor` function directly but that would mean exporting it which would make it public. I can't get this code to fire and it does not exist in master anymore. 

It does seem verifiably wrong.

```js
const x = ['pipe1', 'pipe2', 'pipe3']
> for (let pipe in x) { console.log(pipe) }
0
1
2
> for (let pipe of x) { console.log(pipe) }
pipe1
pipe2
pipe3
```

### DIFF
--- a/lib/change_stream.js
+++ b/lib/change_stream.js
@@ -437,7 +437,7 @@ function createChangeStreamCursor(self, options) {
 
   if (self.pipeDestinations) {
     const cursorStream = changeStreamCursor.stream(self.streamOptions);
-    for (let pipeDestination in self.pipeDestinations) {
+    for (let pipeDestination of self.pipeDestinations) {
       cursorStream.pipe(pipeDestination);
     }
   }
--- a/test/functional/change_stream.test.js
+++ b/test/functional/change_stream.test.js
@@ -1475,7 +1475,7 @@ describe('Change Streams', function() {
     }
   });
 
-  it('should resume piping of Change Streams when a resumable error is encountered', {
+  it.only('should resume piping of Change Streams when a resumable error is encountered', {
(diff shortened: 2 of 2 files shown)


mbroadst:
@reggi I think this ticket was about reenabling [this test](https://github.com/mongodb/node-mongodb-native/blob/master/test/functional/change_stream.test.js#L1458). Looks like you found a bug here, but were you able to get that test passing? Are there other tests we should include here to validate your fix?

reggi:
> @reggi I think this ticket was about reenabling [this test](https://github.com/mongodb/node-mongodb-native/blob/master/test/functional/change_stream.test.js#L1458). Looks like you found a bug here, but were you able to get that test passing? Are there other tests we should include here to validate your fix?

@mbroadst good learning here to do a search for the jira ticket in the codebase (which I don't normally do). I fixed the test, but in 3.6 it wasn't broken because of this pipe issue, it was the file `/tmp/` dir not existing. Fixed this by setting it to a already existing dir, then cleaning up the file using `this.defer`, master may show different issues.

emadum (approved):
This fix LGTM for `3.6`. Just to clarify, we also need to get this functionality working again in `master`. That might be a little more complicated, since we've pulled the Readable stream off the Cursor object. There are currently two skipped tests in `master` - the one you fixed here, and a new somewhat simplified one I wrote. We should try to get at least one of them passing.

mbroadst (approved):

PR #2583, opened 2020-10-15
feature: add Cursor#setReadConcern builder method (3.6)

reggi:
NODE-2806


lib/cursor.js
```diff
@@ -768,6 +769,29 @@ class Cursor extends CoreCursor {
...
   }
 
+  /**
+   * Set the ReadPreference for the cursor.
+   * @method
+   * @param {(string|ReadConcern)} readConcern The new read preference for the cursor.
+   * @throws {MongoError}
+   * @return {Cursor}
+   */
+  setReadConcern(readConcern) {
```
mbroadst, line 779: nit: I think that `withReadConcern` reads better, thoughts?

lib/cursor.js
```diff
@@ -768,6 +769,29 @@ class Cursor extends CoreCursor {
...
+      throw new MongoError('cannot change cursor readConcern after cursor has been accessed');
+    }
+
+    if (readConcern instanceof ReadConcern) {
+      this.options.readConcern = readConcern;
+    } else if (typeof readConcern === 'string') {
+      this.options.readConcern = new ReadConcern(readConcern);
+    } else {
+      throw new TypeError('Invalid read concern: ' + readConcern);
+    }
```
mbroadst, line 790: I wonder if this logic belongs on the `ReadConcern` class itself, perhaps `ReadConcern.resolve(readConcernLike)`? It looks like `ReadConcern.fromOptions` does a similar check

test/functional/cursor.test.js
```diff
@@ -2839,6 +2840,28 @@ describe('Cursor', function() {
     }
   });
 
+  /**
+   * @ignore
+   */
+  it('shouldSetReadConcern', {
```
mbroadst, line 2846: nit: this compactedCamelCaseForm is from the legacy `integra` runner. Let's call this something more readable for future maintainers. 
```suggestion
  it('should allow setting the readConcern through a builder method', {
```
or something like that

test/functional/cursor.test.js
```diff
@@ -2839,6 +2840,28 @@ describe('Cursor', function() {
...
+    metadata: {
+      requires: { topology: ['single', 'replicaset', 'sharded', 'ssl', 'heap', 'wiredtiger'] }
+    },
+    test: function(done) {
```
mbroadst, line 2850: should we use `withClient` here?

test/functional/cursor.test.js
```diff
@@ -2839,6 +2840,28 @@ describe('Cursor', function() {
     }
   });
 
+  /**
+   * @ignore
+   */
+  it('shouldSetReadConcern', {
+    metadata: {
+      requires: { topology: ['single', 'replicaset', 'sharded', 'ssl', 'heap', 'wiredtiger'] }
```
mbroadst, line 2848: if you want to test on all variants then there is no need to provide this `metadata` at all.

mbroadst (changes requested):

lib/cursor.js
reggi: I think it should be consistent with `setReadPreference`.

lib/cursor.js
reggi: We've already handled this in master and have pretty good `fromOptions` methods there, I can port those over, I just thought we'd be anti the back-porting. [In the master branch I re-wrote this to use `.fromOptions` for both `setReadPreference` and `setReadConcern`.](https://github.com/mongodb/node-mongodb-native/pull/2584/files#diff-f7a848a5828483a0958ec4e436ca7f277d846a669c610b5647e192450331d09dR754-R773)

Let me know if you're interested in a back-port of those classes.

test/functional/cursor.test.js
reggi: Cool, good to know.

test/functional/cursor.test.js
reggi: Yeah I'm down.

test/functional/cursor.test.js
reggi: cool also good to know.

lib/cursor.js
reggi: done, backported updated .fromOptions from both classes and have these methods same as master.

lib/cursor.js
mbroadst: I'm thinking we should change `setReadPreference` to `withReadPreference` :smile:

lib/cursor.js
mbroadst: Ok, I think it's a good start to backport the `fromOptions` changes, but I'm still wondering if we can improve things a bit. It seems like generally in both `ReadConcern.fromOptions` and this new method that we want to be able to take something that is `ReadConcernLike`, convert it to a `ReadConcern` instance, and throw an error if the provided value doesn't satisfy the requirements of `ReadConcernLike`. It seems to me like an intermediate method could be used in both places (I suggested `resolve` above), and if we had that method we could avoid the extra allocation here when passing an object to `fromOptions` in this method.

lib/cursor.js
reggi: This is 3.6 that would be a breaking change here, are you ok with that?

test/functional/cursor.test.js
```diff
@@ -9,6 +9,8 @@ const Buffer = require('safe-buffer').Buffer;
 const Writable = require('stream').Writable;
 
 const core = require('../../lib/core');
+const ReadConcern = require('../../lib/read_concern');
+const { withClient } = require('./shared');
```
emadum, line 13:
```suggestion
const withClient = require('./shared').withClient;
```

lib/read_concern.js
```diff
@@ -29,11 +29,14 @@ class ReadConcern {
     }
 
     if (options.readConcern) {
-      if (options.readConcern instanceof ReadConcern) {
-        return options.readConcern;
+      const { readConcern } = options;
```
emadum, line 32:
```suggestion
      const readConcern = options.readConcern;
```

emadum (approved):
Some destructuring which needs to be taken out since this is `3.6`, other than that LGTM 👍

lib/cursor.js
mbroadst: I was thinking in 3.6 we would provide both, and deprecate `setReadPreference`, advising users to use `withReadPreference` instead

test/functional/shared.js
```diff
@@ -148,7 +148,7 @@ function withClient(client, callback) {
     }
     return client
       .connect()
-      .then(callback)
+      .then(callback.bind(this))
```
mbroadst, line 151: do we need to make this change in `master` as well?

lib/core/topologies/read_preference.js
```diff
@@ -100,20 +100,23 @@ const VALID_MODES = [
...
   if (typeof readPreference === 'string') {
     return new ReadPreference(readPreference, readPreferenceTags);
   } else if (!(readPreference instanceof ReadPreference) && typeof readPreference === 'object') {
     const mode = readPreference.mode || readPreference.preference;
     if (mode && typeof mode === 'string') {
       return new ReadPreference(mode, readPreference.tags, {
-        maxStalenessSeconds: readPreference.maxStalenessSeconds || maxStalenessSeconds,
-        hedge: readPreference.hedge
+        maxStalenessSeconds: readPreference.maxStalenessSeconds,
+        hedge: options.hedge
```
mbroadst, line 119: what motivated these changes? It looks like we've lost support for `maxStalenessSeconds` from `options`, and I believe `hedge` is not a valid connection string option

mbroadst (changes requested):

lib/cursor.js
reggi: I've gone down a rabbit hole to refactor and fix `ReadConcern` and `ReadPreference`. ReadPreference seems pretty buggy and there's a lot of duplication. And I also would love to replace `ReadPreference.translate` and use a `options = resolve()` in-place. This request here has a bit of scope-creep which is why I first implemented it the way I did. Complications arise when deciding default values, validation, returning undefined / throwing / default, and parental inheritance, and some type guard functions. I kind of have an opinionated way of how all this should be done. Also I think there's a bit of overlap with Neal's client options work as well.  I'm not sure what to do here, I don't want to respond with an unexpected massive refactor of these classes. Part of me just wants to rip the band-aid off and fix them.

test/functional/shared.js
reggi: yes

lib/core/topologies/read_preference.js
reggi: yes your right, this is in draft now.

reggi: I pulled changes from master thinking that had the updated code, seems like master is wrong and needs updates reflected in 3.6

reggi: Will pull in these changes in master

test/functional/cursor.test.js
reggi: 👍 new linter should help with this

nbbeeken (approved):

PR #2586, opened 2020-10-15
test: cleanup cursor tests

reggi:
NODE-2812

Merges in core cursor tests to the cursor test file.

<s>Here we seemed to have old cursor tests using the topology and server directly, I converted them to use a new `withCursorFromClient` handler, and to come from the client directly. I created a `withCursorFromServer` handler (commented out) as well, just in case we wanted these tests that way. 

I do think it may be overkill to use this `withCursorFromClient`, when I made it I thought there were going to be many more tests. I do think we need a better way to make our tests easier to run, `withClient` and `withMonitoredClient` have their issues. Ideally we can have something that helps us write less boiler-plate code for each test, and that chains / composes well with other `with` helpers.

I would love to refactor `withClient` and `withMonitoredClient` and possibly add in a stack of these `withDb` / `withCollection` / `withDocs` / `withCursor` and have some modular way of stacking them that's not a pain. And have them all behave the same way, some return lamdas, and others don't and resolve the binding issues.</s>


emadum (commented):
LGTM! But, while we're here, I think it might make sense to also pull in some other cursor-related tests into the main `cursor.test.js` from `test/functional/core`: `extend_cursor`, `tailable_cursor`, and `undefined`. @mbroadst what are your thoughts on that?

I also merged in the refactor eliminating `CoreCursor` - you'll need to replace some calls to `Cursor._next` with calls to `Cursor.next` as the former was removed in that merge.

test/functional/shared.js
```diff
@@ -276,6 +276,116 @@ class APMEventCollector {
...
+//           filter: {},
+//           batchSize: batchSize || docs.length
+//         });
+//         return callback(cursor);
+//       });
+//     });
+//   });
+// }
+
+function withCursorFromClient(optionsOrCallback, callback) {
```
mbroadst, line 344: I would really like to avoid writing so many specialized `withXXX` methods. `withClient` makes sense because there's a decent amount of configuration, but otherwise I would recommend just using `this.defer` or just explicitly closing cursors when you're done with them. `withCursorFromClient` currently connects a client, optionally monitors commands, creates a collection, inserts documents into it, and then calls a lambda with a bunch of things.. very magical

We must not be afraid of code duplication in our tests! It is meant to be code written like a user so it actually improves readability and maintainability to have clear steps, even if they are copy-pasta.

mbroadst (changes requested):

test/functional/shared.js
reggi: Rather than use the `withXXX` I create a vscode snippet. May be something we do for specific repetitive tests? Could be a good middleground?

```ts
const BATCH_SIZE = 2;
const COLLECTION = 'cursor0';
const DOCS = [{ a: 1 }, { a: 2 }, { a: 3 }];

const testCursor = cursor => {
  $1
};

const client = configuration.newClient(configuration.writeConcernMax(), { poolSize: 1 });
return client.connect((err, client) => {
  expect(err).to.not.exist;
  this.defer(() => client.close());
  const db = client.db(configuration.db);
  return db.createCollection(COLLECTION, (err, collection) => {
    expect(err).to.not.exist;
    return collection.insertMany(DOCS, this.configuration.writeConcernMax(), err => {
      expect(err).to.not.exist;
      const cursor = collection.find({});
      this.defer(() => cursor.close());
      cursor.batchSize(BATCH_SIZE);
      return testCursor(cursor);
    });
  });
});
```

PR #2587, opened 2020-10-16
fix: correctly extend the "command" object with "user"

through-a-haze:
I believe there should be `Object.assign` and not `Object.apply` otherwise `user` isn't added to the `command` object and authentication that requires this field fails.


mbroadst (approved):
Thanks for the contribution @through-a-haze !

PR #2588, opened 2020-10-16

HanaPearlman:
`connect` doesn't work again after close has been called once on a `MongoClient`.

NODE-2544

port to `3.6`

### DIFF
--- a/lib/mongo_client.js
+++ b/lib/mongo_client.js
@@ -258,6 +258,12 @@ MongoClient.prototype.close = function(force, callback) {
       }
 
       client.removeAllListeners('close');
+
+      // clear out references to old topology
+      client.topology = undefined;
+      client.s.dbCache = new Map();
+      client.s.sessions = new Set();
+
       cb(err);
     };
 
--- a/lib/mongo_client.js
+++ b/lib/mongo_client.js
@@ -272,8 +272,9 @@ MongoClient.prototype.close = function(force, callback) {
       return;
     }
 
-    client.topology.close(force, err => {
-      const autoEncrypter = client.topology.s.options.autoEncrypter;
+    const topology = this.topology;
+    topology.close(force, err => {
+      const autoEncrypter = topology.s.options.autoEncrypter;
       if (!autoEncrypter) {
         completeClose(err);
         return;
--- a/lib/operations/connect.js
+++ b/lib/operations/connect.js
@@ -266,6 +266,11 @@ function connect(mongoClient, url, options, callback) {
     throw new Error('no callback function provided');
   }
 
+  // Has a connection already been established?
+  if (mongoClient.topology && mongoClient.topology.isConnected()) {
+    throw new MongoError(`'connect' cannot be called when already connected`);
+  }
+
   let didRequestAuthentication = false;
   const logger = Logger('MongoClient', options);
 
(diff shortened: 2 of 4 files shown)


test/functional/operation_example.test.js
```diff
@@ -3599,68 +3599,6 @@ describe('Operation Examples', function() {
...
-
-          // Force close the connection
-          client.close(true, function(err) {
-            test.equal(null, err);
-
-            // Attemp to insert should fail now with correct message 'db closed by application'
-            collection.insertOne({ a: 2 }, configuration.writeConcernMax(), function(err, result) {
-              expect(err).to.exist;
-              expect(result).to.not.exist;
-
```
HanaPearlman, line 3654: This test attempts an operation against a closed client. This does fail with an error, as expected, but it also creates a session that never gets closed. We shouldn't be querying against a client that is closed, so we can delete the test.

test/functional/operation_generators_example.test.js
```diff
@@ -2791,68 +2791,6 @@ describe('Operation (Generators)', function() {
...
-        // Insert a document
-        yield collection.insertOne({ a: 1 }, configuration.writeConcernMax());
-
-        // Force close the connection
-        yield client.close(true);
-
-        try {
-          // Attemp to insert should fail now with correct message 'db closed by application'
-          yield collection.insertOne({ a: 2 }, configuration.writeConcernMax());
-        } catch (err) {
```
HanaPearlman, line 2848: Same issue with executing an op against a closed client here.

test/functional/operation_promises_example.test.js
```diff
@@ -2871,67 +2871,6 @@ describe('Operation (Promises)', function() {
...
-            test.ok(result);
-            // Force close the connection
-            return client.close(true);
-          })
-          .then(function() {
-            // Attemp to insert should fail now with correct message 'db closed by application'
-            return collection.insertOne({ a: 2 }, configuration.writeConcernMax());
-          })
-          .catch(function(err) {
-            test.ok(err);
```
HanaPearlman, line 2927: Same issue with executing an op against a closed client here.

test/functional/operation_example.test.js
mbroadst: Now that I'm looking at this, I think we might want to actually keep this test. It's correctly returning an error because the client is closed, but why isn't the second call to `close` (L3655) clearing the session cache? My guess is because the implementation bails if the `topology == null`:
```
      if (client.topology == null) {
        return cb();
      }
```

I think what we actually need to do here is modify `MongoClient#close` to clear the session pool (if it exists) if the `topology == null`

mbroadst (changes requested):

test/functional/operation_example.test.js
HanaPearlman: This is something I was actually confused about. There's a set of sessions stored on the `MongoClient` and a list of sessions stored in the session pool of the `topology`. The second `insertOne` op in this test creates a session in the (now-closed) topology's session pool. To make the test pass, it's not enough to attempt to change `MongoClient#close` to clear its own set of sessions (if it exists)-- we would need a reference to the topology to re-clear its session pool.

HanaPearlman:
Closing this PR because this approach for allowing a client to reconnect after being closed can cause session leaks, since old topologies (and sessions) may be stored on `Db`s/`Collection`s. This ticket can be revisited after NODE-2853.

PR #2589, opened 2020-10-19
refactor: unify means of applying `ignoreUndefined`

HanaPearlman:
## Description

**What changed?**
Previously, each operation in `collection.ts` needed to do the necessary logic to figure out which `BSONSerializationOptions` values to include in the options. Some operations omitted this logic completely, while others did it incorrectly. This change stores the `BSONSerializationOptions` as a new inheritable property on `MongoClient`, `Db`, `Collection`, and `OperationBase`, and makes it accessible via a new `bsonOptions` getter for each class. 

NODE-1561

### DIFF
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -1,3 +1,4 @@
+import type { OperationParent } from './operations/command';
 // import type * as _BSON from 'bson';
 // let BSON: typeof _BSON = require('bson');
 // try {
@@ -71,30 +72,23 @@ export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSe
 }
 
 /**
- * Merge the given BSONSerializeOptions, preferring options over parentOptions, and optionally
+ * Merge the given BSONSerializeOptions, preferring options over the parent's options, and
  * substituting defaults for values not set.
  *
  * @internal
- *
- * @param options - The primary options to use
- * @param parentOptions - The secondary options to use
- * @param includeDefaultOptions - Whether to include default options for any values not set
  */
-export function inheritBSONOptions(
+export function resolveBSONOptions(
   options?: BSONSerializeOptions,
-  parentOptions?: BSONSerializeOptions,
-  includeDefaultOptions?: boolean
+  parent?: OperationParent
 ): BSONSerializeOptions {
-  const defaults = {
-    raw: false,
-    promoteLongs: true,
-    promoteValues: true,
-    promoteBuffers: false,
-    ignoreUndefined: false,
-    serializeFunctions: false,
-    fieldsAsRaw: {}
+  const parentOptions = parent?.bsonOptions;
+  return {
+    raw: options?.raw ?? parentOptions?.raw ?? false,
+    promoteLongs: options?.promoteLongs ?? parentOptions?.promoteLongs ?? true,
+    promoteValues: options?.promoteValues ?? parentOptions?.promoteValues ?? true,
+    promoteBuffers: options?.promoteBuffers ?? parentOptions?.promoteBuffers ?? false,
+    ignoreUndefined: options?.ignoreUndefined ?? parentOptions?.ignoreUndefined ?? false,
+    serializeFunctions: options?.serializeFunctions ?? parentOptions?.serializeFunctions ?? false,
+    fieldsAsRaw: options?.fieldsAsRaw ?? parentOptions?.fieldsAsRaw ?? {}
   };
-
-  const base = includeDefaultOptions ? defaults : {};
-  return pluckBSONSerializeOptions({ ...base, ...parentOptions, ...options });
 }
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -70,14 +70,12 @@ export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSe
   };
 }
 
-// TODO: naming
-// TODO: testing
+// Merge the given BSONSerializeOptions, preferring options over parentOptions, and substituting a
+// default for values not set.
 export function inheritOrDefaultBSONSerializableOptions(
   options?: BSONSerializeOptions,
   parentOptions?: BSONSerializeOptions
 ): BSONSerializeOptions {
-  // Merge the BSONSerializeOptions, preferring options over parentOptions, and substituting a
-  // default for values not set.
   return {
     raw: options?.raw ?? parentOptions?.raw ?? false,
     promoteLongs: options?.promoteLongs ?? parentOptions?.promoteLongs ?? true,
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -69,3 +69,21 @@ export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSe
     raw
   };
 }
+
+// TODO: naming
+// TODO: testing
+export function inheritOrDefaultBSONSerializableOptions(
+  options?: BSONSerializeOptions,
+  parentOptions?: BSONSerializeOptions
+): BSONSerializeOptions {
+  // Merge the BSONSerializeOptions, preferring options over parentOptions, and substituting a
+  // default for values not set.
+  // Note that we exclude fieldsAsRaw and serializeFunctions because I was not sure about their usage
+  return {
+    raw: options?.raw ?? parentOptions?.raw ?? false,
+    promoteLongs: options?.promoteLongs ?? parentOptions?.promoteLongs ?? true,
+    promoteValues: options?.promoteValues ?? parentOptions?.promoteValues ?? true,
+    promoteBuffers: options?.promoteBuffers ?? parentOptions?.promoteBuffers ?? false,
+    ignoreUndefined: options?.ignoreUndefined ?? parentOptions?.ignoreUndefined ?? false
+  };
+}
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -78,12 +78,13 @@ export function inheritOrDefaultBSONSerializableOptions(
 ): BSONSerializeOptions {
   // Merge the BSONSerializeOptions, preferring options over parentOptions, and substituting a
   // default for values not set.
-  // Note that we exclude fieldsAsRaw and serializeFunctions because I was not sure about their usage //TODO: serialize functions?
   return {
     raw: options?.raw ?? parentOptions?.raw ?? false,
     promoteLongs: options?.promoteLongs ?? parentOptions?.promoteLongs ?? true,
     promoteValues: options?.promoteValues ?? parentOptions?.promoteValues ?? true,
     promoteBuffers: options?.promoteBuffers ?? parentOptions?.promoteBuffers ?? false,
-    ignoreUndefined: options?.ignoreUndefined ?? parentOptions?.ignoreUndefined ?? false
+    ignoreUndefined: options?.ignoreUndefined ?? parentOptions?.ignoreUndefined ?? false,
+    serializeFunctions: options?.serializeFunctions ?? parentOptions?.serializeFunctions ?? false,
+    fieldsAsRaw: options?.fieldsAsRaw ?? parentOptions?.fieldsAsRaw ?? {}
   };
 }
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -71,21 +71,30 @@ export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSe
 }
 
 /**
- * Merge the given BSONSerializeOptions, preferring options over parentOptions, and substituting
- * defaults for values not set.
+ * Merge the given BSONSerializeOptions, preferring options over parentOptions, and optionally
+ * substituting defaults for values not set.
+ *
  * @internal
+ *
+ * @param options - The primary options to use
+ * @param parentOptions - The secondary options to use
+ * @param includeDefaultOptions - Whether to include default options for any values not set
  */
 export function inheritBSONOptions(
   options?: BSONSerializeOptions,
-  parentOptions?: BSONSerializeOptions
+  parentOptions?: BSONSerializeOptions,
+  includeDefaultOptions?: boolean
 ): BSONSerializeOptions {
-  return {
-    raw: options?.raw ?? parentOptions?.raw ?? false,
-    promoteLongs: options?.promoteLongs ?? parentOptions?.promoteLongs ?? true,
-    promoteValues: options?.promoteValues ?? parentOptions?.promoteValues ?? true,
-    promoteBuffers: options?.promoteBuffers ?? parentOptions?.promoteBuffers ?? false,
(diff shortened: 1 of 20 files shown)


src/operations/command.ts
```diff
@@ -37,7 +37,16 @@ export interface CommandOperationOptions extends OperationOptions, WriteConcernO
...
+    namespace: MongoDBNamespace;
+    serializeFunctions?: boolean;
+    raw?: boolean;
+    promoteLongs?: boolean;
+    promoteValues?: boolean;
+    promoteBuffers?: boolean;
+    ignoreUndefined?: boolean;
+    options?: BSONSerializeOptions;
+  };
   readConcern?: ReadConcern;
```
HanaPearlman, line 50: I am not sure if there is a cleaner way to indicate that `s` 1) must have a `namespace` field, 2) may be of type `BSONSerializeOptions`, and 3) may have an `options` field of type `BSONSerializeOptions`. I think it would be nice to avoid listing all of the `BSONSerializeOptions` here.

src/operations/find.ts
```diff
@@ -113,15 +113,6 @@ export class FindOperation extends CommandOperation<FindOptions, Document> {
...
-
-    // TODO: figure out our story about inheriting BSON serialization options
-    this.bsonOptions = {
-      raw: options.raw ?? collection.s.raw ?? false,
-      promoteLongs: options.promoteLongs ?? collection.s.promoteLongs ?? true,
-      promoteValues: options.promoteValues ?? collection.s.promoteValues ?? true,
-      promoteBuffers: options.promoteBuffers ?? collection.s.promoteBuffers ?? false,
-      ignoreUndefined: options.ignoreUndefined ?? collection.s.ignoreUndefined ?? false
-    };
   }
```
HanaPearlman, line 116: This logic now occurs in `CommandOperation`. Nothing else needs to change in this file; the `execute` method here already includes `this.bsonOptions` in its options.

src/bson.ts
```diff
@@ -69,3 +69,20 @@ export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSe
     raw
   };
 }
+
+// Merge the given BSONSerializeOptions, preferring options over parentOptions, and substituting
+// defaults for values not set.
+export function inheritOrDefaultBSONSerializableOptions(
```
nbbeeken, line 75: The comments above are useful, nice work you can go ahead and make it a docstyle comment 
```
/** 
 * ...
 * @internal
 */
```
Then it will be readable by just hovering it anywhere in the codebase, also you can make it internal.

Also fine with it as is but may I suggest the name: `inheritBSONOptions` I think its fine to just state they will be defaulted in the docs.

src/operations/command.ts
```diff
@@ -90,6 +99,12 @@ export abstract class CommandOperation<
     if (parent && parent.logger) {
       this.logger = parent.logger;
     }
+
+    // Assign BSON serialize options to OperationBase, preferring options over parent options.
+    // base accounts for the fact that Collection stores bson options in s, while other parents,
+    // like Db, stores bson options in s.options
+    const base = Object.assign({}, parent?.s.options, parent?.s);
+    Object.assign(this, inheritOrDefaultBSONSerializableOptions(options, base));
```
nbbeeken, line 107: We can inline the base here like so:
```js
Object.assign(this, inheritOrDefaultBSONSerializableOptions(options, {...parent.s.options, ...parent?.s}));
```

src/operations/command.ts
```diff
@@ -90,6 +99,12 @@ export abstract class CommandOperation<
     if (parent && parent.logger) {
       this.logger = parent.logger;
     }
+
+    // Assign BSON serialize options to OperationBase, preferring options over parent options.
+    // base accounts for the fact that Collection stores bson options in s, while other parents,
+    // like Db, stores bson options in s.options
```
nbbeeken, line 105: Maybe this is cause to change that? I wonder how much it would effect to make collection keep its bsonOptions in options, or have db spread them on to `s`.

That could make this into a lot of work, so total throwaway comment if so, but curious if you have insight on what that would entail?

src/operations/delete.ts
```diff
@@ -65,7 +65,7 @@ export class DeleteOneOperation extends CommandOperation<DeleteOptions, DeleteRe
   execute(server: Server, callback: Callback<DeleteResult>): void {
     const coll = this.collection;
     const filter = this.filter;
-    const options = this.options;
+    const options = Object.assign({}, this.options, this.bsonOptions);
```
nbbeeken, line 68: I won't comment on every instance to spare you the emails, but I think we should prefer the spread operator to assign: `const options = {...this.options, ...this.bsonOptions}`. But not a strong opinion, and spread doesn't work in every case, like when assigning onto `'this`

test/functional/insert.test.js
```diff
@@ -1902,6 +1902,69 @@ describe('Insert', function () {
     }
   });
 
+  it('shouldCorrectlyInheritPromoteLongFalseNativeBSONWithGetMore', {
+    // Add a tag that our runner can trigger on
+    // in this case we are setting that node needs to be higher than 0.10.X to run
```
nbbeeken, line 1907: this comment looks left over? there's nothing about 0.10.x here

nbbeeken (changes requested):
just a few nits and 🧅 s, nice work!

src/operations/command.ts
HanaPearlman: I don't think it would have a strong effect at all to move the Collection `bsonOptions` to `s.options`-- as far as I can tell, very few places were using the Collection `bsonOptions` on `s`, and I deleted most of them in this PR. Would be happy to make this change. 

As for spreading other parent `bsonOptions` onto `s`, `Db` was just one example. I think we would also have to change `cursor` too, and possibly others.

nbbeeken: I think that would be nice for consistency
I'm thinking now the original reason that collection might pick off its own copy of these options is to make sure that it doesn't edit its parent's options, I took a quick look around and it seems like collection does get its own separate options object but that could be a bug you run into if you give this a go, the options prop that collections uses shouldn't be the same reference to Db's options object.

src/operations/command.ts
reggi: In a couple of places in the codebase (see `AdminPrivate`) we pop out the `s` property to have its own type interface. If you did that here you could then have a `OperationParentPrivate` that `extends` `BsonSerializeOptions`. Then add in the `namespace`. Ideally we dont have the bson options at the top level, as well as  `options`.

HanaPearlman: Assuming we go ahead with moving the Collection bson options into `s.options`, then we would not need any of the bson options directly on the `s` property, which would improve how this block looks. I think we can't avoid having `options` be part of `s` though, because otherwise we can't grab the parent options to do the inheriting.

src/operations/bulk_write.ts
```diff
@@ -24,18 +25,15 @@ export class BulkWriteOperation extends OperationBase<BulkWriteOptions, BulkWrit
 
     this.collection = collection;
     this.operations = operations;
+
+    // Assign BSON serialize options to OperationBase, preferring options over collection options
+    Object.assign(this, inheritBSONOptions(options, collection.s));
```
reggi, line 30: Total opinons here:

I'd love some consistency if possible. It seems that collection has it on `.s` but bulk_write is on the class itself.

Also using `Object.assign(this...` is super convenient  but usually shy away from it to ensure types not being violated. 

Here I'd prefer:

```ts
this.bsonOptions = inheritBSONOptions(options, collection.s));
```

or 

```ts
const bsonOptions = inheritBSONOptions(options, collection.s));
this.raw = bsonOptions.raw;
this.promoteLongs = bsonOptions.promoteLongs;
this.promoteValues = bsonOptions.promoteValues;
this.promoteBuffers = bsonOptions.promoteBuffers;
this.promoteBuffers = bsonOptions.promoteBuffers;
this.ignoreUndefined = bsonOptions.ignoreUndefined;
this.serializeFunctions = bsonOptions.serializeFunctions;
this.fieldsAsRaw = bsonOptions.fieldsAsRaw;
```

HanaPearlman: Your point about consistency is great, I definitely struggled with that for these changes. 

It seems to me that for the Collection/Db/Client, the bson options must be somewhere in `.s`, but classes extending `OperationBase` don't (currently) have a `.s`, so unless that changes there can't be perfect consistency...

Here is how I was thinking about it (assuming I make the change for `Collection`): `Collection`, `Db`, and `MongoClient` could all store the bson options in `s.options`, while anything extending `OperationBase` has the bson options on the class itself. The rational for the second part of this seemed to come out of @nbbeeken's work-- wondering if he has any thoughts on this.

emadum (changes requested):
Had a few minor comments about cleaning up the new tests to our modern style, but this LGTM other than that, nice work!

test/functional/ignore_undefined.test.js
```diff
@@ -148,4 +149,128 @@ describe('Ignore Undefined', function () {
...
+            client.close(done);
+          });
+        });
+      });
+    }
+  });
+
+  it(
+    'Should correctly inherit ignore undefined field from collection during insert',
+    withClient(function (client, done) {
```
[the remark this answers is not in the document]
HanaPearlman: Nope, there is no reason! I guess I didn't see exactly why that was needed, so I didn't include it. Will make this change in case there is something I am missing.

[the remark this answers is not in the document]
emadum: I think in some cases we limit very basic tests that don't depend on replica sets or sharded clusters to just run on the `single` topology, because we have extra tests that we only run on replica/sharded, so those tasks take longer to complete in Evergreen. Limiting tests to single when possible is a way we can cut down on the wait time (and cost!) of running our tests. @mbroadst might be able to provide more background about this.

emadum (approved):
LGTM 👍

src/operations/bulk_write.ts
```diff
@@ -24,18 +25,15 @@ export class BulkWriteOperation extends OperationBase<BulkWriteOptions, BulkWrit
 
     this.collection = collection;
     this.operations = operations;
+
+    // Assign BSON serialize options to OperationBase, preferring options over collection options
+    Object.assign(this, inheritBSONOptions(options, collection.s.options, true));
```
reggi, line 30: I don't think we should use `Object.assign` on `this` it does not give any information just by looking at this how `this` is being manipulated.

reggi (changes requested):

src/operations/bulk_write.ts
reggi: Alternative examples in https://github.com/mongodb/node-mongodb-native/pull/2589#discussion_r508812071

src/operations/command.ts
```diff
@@ -37,7 +37,10 @@ export interface CommandOperationOptions extends OperationOptions, WriteConcernO
 
 /** @internal */
 export interface OperationParent {
-  s: { namespace: MongoDBNamespace };
+  s: {
+    namespace: MongoDBNamespace;
+    options?: BSONSerializeOptions;
```
mbroadst, line 42: I was thinking that `bsonOptions` would be another inheritable property like `readConcern`/`writeConcern`/`readPreference` etc. outside of the `s`. Should we at least consider renaming this to `bsonOptions`?

src/bson.ts
```diff
@@ -69,3 +69,32 @@ export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSe
...
+ * @internal
+ *
+ * @param options - The primary options to use
+ * @param parentOptions - The secondary options to use
+ * @param includeDefaultOptions - Whether to include default options for any values not set
+ */
+export function inheritBSONOptions(
+  options?: BSONSerializeOptions,
+  parentOptions?: BSONSerializeOptions,
+  includeDefaultOptions?: boolean
```
mbroadst, line 86: I think we should avoid [magic booleans](https://en.wikipedia.org/wiki/Magic_number_(programming)). Are there cases where we _don't_ want to set defaults? Ultimately we do apply a [set of defaults](https://github.com/mongodb/node-mongodb-native/blob/master/src/cmap/connection.ts#L382-L387) before serializing to the wire protocol, so I'm wondering if we can always apply the defaults.

src/bson.ts
```diff
@@ -69,3 +69,32 @@ export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSe
...
+ * Merge the given BSONSerializeOptions, preferring options over parentOptions, and optionally
+ * substituting defaults for values not set.
+ *
+ * @internal
+ *
+ * @param options - The primary options to use
+ * @param parentOptions - The secondary options to use
+ * @param includeDefaultOptions - Whether to include default options for any values not set
+ */
+export function inheritBSONOptions(
```
mbroadst, line 83: small nit: we tend to call these types of methods `resolveXXX` 

Additionally, if we standardized on all "parent-like" types having a optional `bsonOptions` property (like other inheritable properties) then I think we could avoid having to pluck anything from the parent. You would just pluck the options from the user-provided options on the operation, and resolve it with a possible set of `bsonOptions` on the parent, which could be inherited up the chain back to `MongoClient`. Take a look at how `readPreference` works for a better idea of what I'm talking about here.

HanaPearlman: This makes sense, and I think it may be along the same lines as what @reggi was suggesting earlier, since it establishes a consistent place to look for `bsonOptions` across the parent-like types and `OperationBase` types. 

One small thing: I don't think the property would be optional, since we always have defaults to fall back on. Otherwise, I see how this is better!

test/functional/ignore_undefined.test.js
```diff
@@ -148,4 +149,132 @@ describe('Ignore Undefined', function () {
       });
     }
   });
+
+  it('Should correctly inherit ignore undefined field from db during insert', {
+    metadata: { requires: { topology: ['single'] } },
```
mbroadst, line 154: Is there a reason to only test against a standalone? If you want to test on any topology type/server version then you can omit the `metadata` completely:
```
if('should correctly inherit ignoreUndefined from db during insert', function(done) {
  ...
});
```

mbroadst (approved):
LGTM! I left one minor nit about test structure. :clap: great job on this work

src/bson.ts
mbroadst: agreed about defaults, once we determined there were always values then this should not be optional.

test/functional/ignore_undefined.test.js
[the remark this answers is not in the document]
mbroadst: Oops, I just left a comment about this below. I think the rule of thumb should actually be that we run tests everywhere when possible, so we maximize our coverage. Let's tackle the runtime of the tests at a later date, we can likely make a number of improvements to speed things up while maintaining as much coverage as possible across multiple cluster types

reggi (approved):
LGTM

nbbeeken (approved):

PR #2593, opened 2020-10-21
refactor: Remove generic options parameter from OperationBase

nbbeeken:
Add a `builtOptions` getter that inherits options from the parent class.

NODE-2815

### Description

In order to make tracing options sensible I went with this idea of inheriting the super class's `builtOptions` which are frozen. I'll put this small start in review to get some feedback going on how we like this methodology and I can start expanding this to other operations. 

Some of the bsonOptions work is duplicated in here to get the code to work, but I will pull that in from master when its there.

Look out for a parallel PR to this that works on removing executeLegacyOperation.

### DIFF
--- a/src/operations/command.ts
+++ b/src/operations/command.ts
@@ -1,7 +1,7 @@
 import { Aspect, OperationBase, OperationOptions } from './operation';
 import { ReadConcern } from '../read_concern';
 import { WriteConcern, WriteConcernOptions } from '../write_concern';
-import { maxWireVersion, MongoDBNamespace, Callback } from '../utils';
+import { maxWireVersion, MongoDBNamespace, Callback, deepFreeze } from '../utils';
 import { ReadPreference, ReadPreferenceLike } from '../read_preference';
 import { commandSupportsReadConcern } from '../sessions';
 import { MongoError } from '../error';
@@ -35,9 +35,14 @@ export interface CommandOperationOptions extends OperationOptions, WriteConcernO
   noResponse?: boolean;
 }
 
+export interface OperationParentPrivate extends BSONSerializeOptions {
+  options?: BSONSerializeOptions;
+  namespace: MongoDBNamespace;
+}
+
 /** @internal */
 export interface OperationParent {
-  s: { namespace: MongoDBNamespace };
+  s: OperationParentPrivate;
   readConcern?: ReadConcern;
   writeConcern?: WriteConcern;
   readPreference?: ReadPreference;
@@ -54,9 +59,29 @@ export abstract class CommandOperation<
   readPreference: ReadPreference;
   readConcern?: ReadConcern;
   writeConcern?: WriteConcern;
-  explain: boolean;
   fullResponse?: boolean;
   logger?: Logger;
+
+  protected collation;
+  protected maxTimeMS;
+  protected comment;
+  protected retryWrites;
+  protected noResponse;
+
+  get builtOptions(): Readonly<CommandOperationOptions> {
+    return deepFreeze({
+      ...super.builtOptions,
+      collation: this.collation,
+      maxTimeMS: this.maxTimeMS,
+      comment: this.comment,
+      retryWrites: this.retryWrites,
+      noResponse: this.noResponse,
+      fullResponse: this.fullResponse
+      // Override with proper type
+      // writeConcern: this.writeConcern,
+      // readConcern: this.readConcern
+    });
+  }
 
   constructor(parent?: OperationParent, options?: T) {
     super(options);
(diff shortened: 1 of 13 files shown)


src/utils.ts
```diff
@@ -1158,3 +1158,23 @@ export function hasAtomicOperators(doc: Document | Document[]): boolean {
...
+  // Retrieve the property names defined on object
+  const propNames = Object.getOwnPropertyNames(object);
+
+  // Freeze properties before freezing self
+
+  for (const name of propNames) {
+    const value = object[name];
+
+    if (name === 'session') continue;
+    if (ArrayBuffer.isView(value)) continue;
```
nbbeeken, line 1172: This is pretty lame skip logic, but I'm thinking maybe deepFreezing is overkill anyway. thoughts?

src/operations/command.ts
```diff
@@ -75,21 +100,45 @@ export abstract class CommandOperation<
...
     this.explain = false;
     this.fullResponse =
       options && typeof options.fullResponse === 'boolean' ? options.fullResponse : false;
 
+    // if (this.writeConcern && this.writeConcern.w === 0) {
+    //   if (this.session && this.session.explicit) {
+    //     throw new MongoError('Cannot have explicit session with unacknowledged writes');
+    //   }
+    //   return;
+    // }
```
nbbeeken, line 115: I ran into a test failure when trying to add a few more operations to this PR that relates to this in `applySession` which is commented saying it should occur at command construction, I'll still have to discover the bug, but shall we move this here?
Maybe as part of removing legacy operations so everything would go through this path.

src/utils.ts
```diff
@@ -1158,3 +1158,23 @@ export function hasAtomicOperators(doc: Document | Document[]): boolean {
...
   return keys.length > 0 && keys[0][0] === '$';
 }
+
+export function deepFreeze<O extends Record<string, any>>(object: O): Readonly<O> {
+  // Retrieve the property names defined on object
+  const propNames = Object.getOwnPropertyNames(object);
+
+  // Freeze properties before freezing self
+
+  for (const name of propNames) {
```
reggi, line 1168: Concern here about the combination of using deepFreeze within getters, this for loop will run on every access `this.thing`, and it's pretty hefty for loop. We may be making the object immutable but it's not the same object every-time this getter is run. Thoughts on some memoization or caching functionality?

reggi: This could easily be achieved by setting `this.cacheBuildOptions` from within the getter, and return that if it exists.

reggi (changes requested):
Curious how this overlaps with @HanaPearlman's work.

src/operations/command.ts
```diff
@@ -98,15 +147,14 @@ export abstract class CommandOperation<
...
-    const options = this.options;
     const serverWireVersion = maxWireVersion(server);
     const inTransaction = this.session && this.session.inTransaction();
 
     if (this.readConcern && commandSupportsReadConcern(cmd) && !inTransaction) {
       Object.assign(cmd, { readConcern: this.readConcern });
     }
 
-    if (options.collation && serverWireVersion < SUPPORTS_WRITE_CONCERN_AND_COLLATION) {
+    if (this.builtOptions.collation && serverWireVersion < SUPPORTS_WRITE_CONCERN_AND_COLLATION) {
```
emadum, line 157: Any reason you changed these `options` to `this.builtOptions` instead of changing above to `const options = this.builtOptions`? Seems like we're calling that deepFreeze getter repeatedly without need here.

src/utils.ts
emadum: I think it's fine in terms of performance if we minimize use of the getter, I commented about that above. In terms of immutability, I don't think we can expect something that is gradually built on an object to be the same over the object's lifespan, so I think this is OK.

src/utils.ts
emadum: I think it would be useful to explain why those need to be skipped with comments. Is there a reason we can't use a popular public domain implementation of this like the [deep-freeze](https://github.com/substack/deep-freeze/blob/master/index.js) npm module?

emadum (commented):
Couple questions but this approach looks good to me 👍

src/operations/command.ts
nbbeeken: Good catch I'm going through and changing, in fact I think I want to make this a method now, leaving this much work in a property access its easy to overlook that its doing work each time.

src/utils.ts
nbbeeken: I got the code from the same place that dep gets it from the MDN docs, I figure for the small changes we might want to make, might as well in house it.
I did document these skips but I think I can come up with better logic.

nbbeeken:
Superseded by #2659

PR #2594, opened 2020-10-21
refactor!: only store topology on MongoClient 

HanaPearlman:
## Description

**What changed?**
A `topology` is no longer stored on `Db` or `Collection`; it's only on `MongoClient`. Each class only stores its parent and accesses the topology through that parent. 

Methods that require a topology can now run into cases where the topology is undefined; an error is thrown during these cases.

This work is a step towards having a single, immutable `topology` on `MongoClient` which can never be null.

NODE-2850


test/unit/collection.test.js
```diff
@@ -1,29 +1,19 @@
...
+    test: withClient((client, done) => {
+      const db = client.db('fakeDb');
       const collection = db.collection('test');
       expect(() => {
         collection.findOneAndReplace({ a: 1 }, { $set: { a: 14 } });
       }).to.throw(/must not contain atomic operators/);
-    }
+      done();
+    })
   });
```
HanaPearlman, line 18: This test needed to change because now the `Db` constructor takes a `MongoClient` instead of a `topology`. Perhaps this test could/should be moved to `test/functional`?

src/operations/execute_operation.ts
```diff
@@ -40,24 +40,32 @@ export function executeOperation<
...
->(topology: Topology, operation: T, callback?: Callback<TResult>): Promise<TResult> | void {
+>(
+  topology: Topology | undefined,
+  operation: T,
+  callback?: Callback<TResult>
+): Promise<TResult> | void {
   if (topology == null) {
-    throw new TypeError('This method requires a valid topology instance');
+    throw new MongoClientClosedError();
   }
```
HanaPearlman, line 69: It may be a point of discussion whether this undefined topology check should happen here or where `executeOperation` is actually called. This way significantly reduces the number of checks required for this change. One downside is that we can't specifically identify the method that caused the `MongoClientClosedError` in the error message.

Another question is whether we want to be throwing the error in this way or using `maybePromise` to use the callback.

test/unit/collection.test.js
reggi: Yeah makes sense.

src/collection.ts
```diff
@@ -261,6 +259,13 @@ export class Collection implements OperationParent {
     return this.s.readPreference;
   }
 
+  /**
+   * The current topology of the collection.
+   */
+  get topology(): Topology | undefined {
```
reggi, line 265: This here exposes `topology` as a top-level API, I'm generally OK with this, but it is something to acknowledge. We're usually conscious about what we allow as apart of the stable API. We can't make it private because we need to call it internally `collection.topology`. Calling the `.s.properties` is pretty annoying. Our users have always been able to call `.s.topology` but I don't think we consider that stable. 

> Side thought may not need breaking change `!` here?
 
I really like it, and I'd also be interested to add getters in all the places we're using topology / parent / db / client. 

```ts
  get db () {
    return this.s.db;
  }

 get topology () {
    return this.db.topology;
  }
```

This would help in that we can always just call `this.topology` from within a given class.

reggi (changes requested):
Just one thought about adding getters / public API.

src/collection.ts
HanaPearlman: I think the getters idea is really good! Am happy to add that.

mbroadst: Hm, I'm not sure we should take this approach. My main question is: who is this feature helping? Do we expect users to need to refer to the `topology` or the `db` of a `Collection`, or are we adding something here that only helps us (the maintainers)?

For starters, I don't think we should make it easy to access the `Topology` type, since we want that to be private API that we can change at will. Leaving the topology in the `s` is unsavory, but at least let's us make decisions later about how we want to continue to shield users from accessing private implementation.

If it's too hard to stomach `this.s.db.client.topology`, we could consider a new `getTopology` util method. Given how boilerplate this all is, I'm not sure it's worth the effort to make it look clean though.

HanaPearlman: This is a good point. It seems like this would just be for the maintainers' benefit.

One thing to note is that a `topology` getter already exists on the `Db` class. Is this something we would consider removing?

HanaPearlman: For now I have gone with a `getTopology` util method on the `Collection` class. Since a `topology` field exists on `MongoClient` and a `topology` getter exists on `Db`, it's a bit inconsistent. To address this, I think we could remove the `topology` getter on `Db` and (potentially) replace it with a `getTopology` util.

src/collection.ts
```diff
@@ -240,6 +238,14 @@ export class Collection implements OperationParent {
     return this.s.bsonOptions;
   }
 
+  /**
+   * The current topology of the collection.
+   * @internal
+   */
+  getTopology(): Topology | undefined {
```
mbroadst, line 245: a-ha, when we were discussing this the other day I meant something a little different than this - I see how the suggestion would be interpreted this way, sorry for the confusion. What I think we should really try to do as much as possible is reduce the public api impact of this change. Rather than providing a `getTopology` method (as opposed to a `topology` property), I was suggesting that we provide some sort of completely internal `getTopology` method that _accepts_ something that provides a topology:

```
function getTopology(provider: MongoClient | Db | Collection): Topology {
  if (provider.s.topology) return provider.s.topology;
  if ....
}
```

A stricter approach might be to introduce a `TopologyProvider` interface, maybe in `topology.ts`:
```
const kTopology = Symbol('topology');
export interface TopologyProvider {
  [kTopology]: Topology
}

export getTopology(provider: TopologyProvider): Topology {
  return provider[kTopology];
}
```
If that approach works, then you have made `topology` visible and not accessible outside of the library (private)

src/utils.ts
```diff
@@ -717,6 +712,16 @@ export function maxWireVersion(topologyOrServer?: Connection | Topology | Server
...
+  const Collection = loadCollection();
+  const Db = loadDb();
+  const MongoClient = loadMongoClient();
+
+  if (provider instanceof MongoClient) return provider.topology;
+  if (provider instanceof Db) return provider.s.client.topology;
+  if (provider instanceof Collection) return provider.s.db.s.client.topology;
+}
+
 /**
```
HanaPearlman, line 725: I included the dynamic loaders because I was having circular dependency issues once I added this function. Now, I'm running into issues because the linter thinks the `MongoClient` import is unused (I am assuming because it seems to be overrided here). 

I can see at least two approaches: one is to change the argument type to `OperationParent`, another is to move this function to the `topology.ts` file, where the circular dependency issue is not happening. Is there one of these, or another approach, that is preferred?

emadum (approved):
LGTM 👍

nbbeeken (approved):
One 🧅  that's optional lgtm!

src/sdam/topology.ts
```diff
@@ -57,6 +57,7 @@ import { DestroyOptions, Connection } from '../cmap/connection';
 import { RunCommandOperation } from '../operations/run_command';
 import type { CursorOptions } from '../cursor/cursor';
 import type { MongoClientOptions } from '../mongo_client';
+import { Db, Collection, MongoClient } from '..';
```
nbbeeken, line 60:
```suggestion
import { Db } from '../db';
import { Collection } from '../collection';
import { MongoClient } from '../mongo_client';
```
Just an 🧅  I think internally we should import from the file directly

src/change_stream.ts
```diff
@@ -220,22 +220,23 @@ export class ChangeStream extends EventEmitter {
...
-      this.topology = parent.topology!;
     } else {
       throw new TypeError(
         'parent provided to ChangeStream constructor is not an instance of Collection, Db, or MongoClient'
       );
     }
 
+    const topology = getTopology(parent);
+    if (!topology) throw new MongoClientClosedError();
+    this.topology = topology;
```
mbroadst, line 238: we should be able to use `getTopology` to remove the need to attach a `topology` property on `ChangeStream` as well

src/db.ts
```diff
@@ -149,10 +149,10 @@ export class Db implements OperationParent {
    * Creates a new Db instance
    *
    * @param databaseName - The name of the database this instance represents.
-   * @param topology - The server topology for the database.
+   * @param client - The MongoClient for the database.
    * @param options - Optional settings for Db construction
    */
-  constructor(databaseName: string, topology: Topology, options?: DbOptions) {
+  constructor(databaseName: string, client: MongoClient, options?: DbOptions) {
```
mbroadst, line 155:
```suggestion
  constructor(client: MongoClient, databaseName: string, options?: DbOptions) {
```
nit, but if we're changing this anyway, let's try to go from most fundamental element to most optional element in the parameter ordering (this mirrors how we store things in our private `s` object)

src/db.ts
```diff
@@ -194,8 +194,8 @@ export class Db implements OperationParent {
   }
 
   // Topology
-  get topology(): Topology {
-    return this.s.topology;
+  get topology(): Topology | undefined {
```
mbroadst, line 197: aren't we removing all `topology` properties here?

src/collection.ts
```diff
@@ -636,8 +646,10 @@ export class Collection implements OperationParent {
       throw new TypeError('`options` parameter must not be function');
     }
 
+    const topology = getTopology(this);
+    if (!topology) throw new MongoClientClosedError('Collection.prototype.find');
```
mbroadst, line 650: Rather than create a custom error message and throwing them in place, it looks like you can actually just move the error inside `getTopology`. `getTopology` should throw if it's unable to find the topology, with a `MongoError` which says that the client has been closed. Additionally, there is no need to include the name of the method in the error message string because you'll get that context in the stack trace anyway.

src/operations/execute_operation.ts
```diff
@@ -40,25 +40,31 @@ export function executeOperation<
...
->(topology: Topology, operation: T, callback?: Callback<TResult>): Promise<TResult> | void {
-  if (topology == null) {
-    throw new TypeError('This method requires a valid topology instance');
-  }
+>(
+  topology: Topology | undefined,
+  operation: T,
+  callback?: Callback<TResult>
+): Promise<TResult> | void {
+  if (!topology) throw new MongoClientClosedError();
```
mbroadst, line 67: `getTopology` should be changed to always return a `Topology`, not `Topology | undefined` - that method will throw if one is not found. Likewise, you can remove this `topology == null` check since we now validate the presence of a `Topology` in `getTopology`

src/utils.ts
```diff
@@ -433,20 +431,15 @@ export function isPromiseLike<T = any>(
...
  *
  * @param command - the command on which to apply collation
- * @param target - target of command
+ * @param topology - topology of command
  * @param options - options containing collation settings
  */
 export function decorateWithCollation(
   command: Document,
-  target: { s: { topology: Topology } } | { topology: Topology },
+  topology: Topology | undefined,
```
mbroadst, line 439: I think we're actually looking for something like `target: MongoClient | Db | Collection`. You should be able to reuse `getTopology` here

test/functional/mongo_client.test.js
```diff
@@ -220,8 +220,8 @@ describe('MongoClient', function () {
       client.connect(function (err, client) {
         expect(err).to.not.exist;
         var db = client.db(configuration.db);
-        expect(db).nested.property('s.topology.s.options.connectTimeoutMS').to.equal(0);
-        expect(db).nested.property('s.topology.s.options.socketTimeoutMS').to.equal(0);
+        expect(db).nested.property('s.client.topology.s.options.connectTimeoutMS').to.equal(0);
```
mbroadst, line 223: reuse `getTopology` here

test/unit/collection.test.js
```diff
@@ -1,29 +0,0 @@
...
-    super();
-  }
-
-  capabilities() {
-    return {};
-  }
-}
-
-describe('Collection', function () {
-  it('should not allow atomic operators for findOneAndReplace', {
```
mbroadst, line 19: is the behavior this test is tracking covered by some other test? Otherwise, it seems like the more prudent approach would be to try to mock a `MongoClient` with a `MockTopology`

src/sdam/topology.ts
```diff
@@ -1053,3 +1054,10 @@ export class ServerCapabilities {
     return this.maxWireVersion >= 5;
   }
 }
+
+/** @internal */
+export function getTopology(provider: MongoClient | Db | Collection): Topology | undefined {
```
mbroadst, line 1059: Another way to implement this is to use ducktyping. It would remove the need for having to import these types into the file directly, which will prevent cycles in module loading:
```
export function getTopology(provider: MongoClient | Db | Collection): Topology {
  if (`topology` in provider) return provider.topology;
  if ('s' in provider && 'client' in provider.s && 'topology' in provider.s.client) return provider.s.client.topology;
  ...
```

src/sdam/topology.ts
mbroadst: +1, also you may run into issues with require "cycles" with this approach (noted below).

src/error.ts
```diff
@@ -124,6 +124,19 @@ export function isNetworkErrorBeforeHandshake(err: MongoNetworkError): boolean {
   return err[kBeforeHandshake] === true;
 }
 
+/**
+ * An error thrown when an operation requiring a connected client was called on a closed client.
+ * @public
+ * @category Error
+ */
+export class MongoClientClosedError extends MongoError {
```
mbroadst, line 132: It's uncommon for us to define a single-purpose error like this, and I think (as I said above) we can get away with moving the logic (and creation of the error) to one location inside `getTopology`. In the future I would generally say that single purpose errors may indicate an issue in code structure

mbroadst (changes requested):

test/unit/collection.test.js
HanaPearlman: I moved the test to `test/functional/find_and_modify.test.js` and just used `withClient` instead of mocking. Is that an acceptable alternative?

mbroadst: awesome, yes that's great

mbroadst (approved):
:clap: :clap: very nice work here, LGTM

PR #2595, opened 2020-10-23
test: typings for test

reggi:
NODE-2525

Adds type information to our test files.

* We can turn off `checkJs` to just gain type information and not check. 
* Discovered bug with typescript (below) (many errors can be attributed to missing `this` context within tests)
* Custom modifications to mocha types to allow for metadata
* Cleans up some overloads
* Discovers ~1198 type errors
* Many of the errors are with this within `Sinon` `callFake` (unsure how to address, attempted to fix)
* Type errors with `withClient` in that mocha can't reconcile returning promise or lambda
* Uncovers may type issues with the API / options / overloads

Code block as not to link:

```
https://github.com/microsoft/TypeScript/issues/41213
```

My learning / understanding of how to modify global types is still rocky, there may be better ways to modify `mocha` / `sinon`, decisions like `mocha.d.ts` vs `mocha.ts` I am unsure on.


reggi:
I'm not exactly sure why these tests are failing, I don't believe it was something that I introduced. I checked out master and the tests failed there as well.

PR #2599, opened 2020-10-27
feat: add explain support for non-cursor commands

HanaPearlman:
## Description

This PR implements `explain` functionality for the relevant non-cursor operations: `update`, `remove`, `distinct`, `findAndModify`, and `mapReduce`. These operations now accept an additional `explain` option, either a boolean or a string specifying the requested verbosity, and will be run as explain commands with the specified verbosity. 

Current syntax: `collection.updateOne(..., {explain: verbosity})`. 
Other syntax, e.g.: `collection.explain(verbosity).updateOne(...)` is easy to add (or switch to) if we want.

### What changed?

A new `Explainable` aspect was added so that, in `CommandOperation`, operations with that aspect can use the `explain` property already on the class. Logic for validating the explain verbosity passed through the options, adding the explain field to the command, and `canRetryWrite` have been added to `CommandOperation`. 

Not all explainable commands actually call `executeCommand`, including `updateOne`, `deleteOne`, etc. These are handled in the corresponding way: the specified explain verbosity is passed through the options until the command document is created, and then the explain is added to it.

Important note: adding an explain to a command alters its shape significantly. For this reason, it should happen at the end of the command building process. This has led to some weirdness where write operations that are being explained still seem like they are doing writes (still call `executeWriteOperation` in `server.ts`). I couldn't see a workaround for this that avoided repeating lots of command building logic. Open to any other suggestions!

NODE-2852


src/explain.ts
```diff
@@ -0,0 +1,78 @@
...
+      (op === 'remove' && wireVersion >= SUPPORTS_EXPLAIN_WITH_REMOVE) ||
+      (op === 'update' && wireVersion >= SUPPORTS_EXPLAIN_WITH_UPDATE) ||
+      (op === 'distinct' && wireVersion >= SUPPORTS_EXPLAIN_WITH_DISTINCT) ||
+      (op === 'findAndModify' && wireVersion >= SUPPORTS_EXPLAIN_WITH_FIND_AND_MODIFY) ||
+      (op === 'mapReduce' && wireVersion >= SUPPORTS_EXPLAIN_WITH_MAP_REDUCE)
+    );
+  }
+
+  /** Checks that the server supports explain on the given command.*/
+  static explainSupportedOnCmd(server: Server, cmd: Document): boolean {
```
emadum, line 68: Could we overload `explainSupported` to handle either an `op: string` or a `cmd: Document`, instead of having two separate methods that are pretty similar in implementation?

src/operations/command.ts
```diff
@@ -141,6 +139,12 @@ export abstract class CommandOperation<
...
 
+    // If a command is to be explained, we need to reformat the command after
+    // the other command properties are specified.
+    if (cmd.explain) {
+      cmd = decorateWithExplain(cmd, cmd.explain);
+    }
+
     server.command(
       this.ns.toString(),
       cmd,
```
emadum, line 150: This is a stylistic nit, but I think it'd be better to do `cmd.explain ? decorateWithExplain(cmd, cmd.explain) : cmd` here rather than reassigning the `cmd` parameter.

src/utils.ts
```diff
@@ -486,6 +487,23 @@ export function decorateWithReadConcern(
...
+ * @param command - the command on which to apply the read concern
+ * @param options - the options containing the explain verbosity
+ */
+export function decorateWithExplain(command: Document, explain: Explain): Document {
+  // A command being explained cannot have an explain field directly on it
+  if (command.explain !== undefined) {
+    delete command.explain;
+  }
+
+  command = { explain: command, verbosity: explain.verbosity };
```
emadum, line 503: I'd just return this object literal directly rather than reassigning `command`.

emadum (approved):
LGTM with some nits 👍

src/operations/explainable_command.ts
```diff
@@ -0,0 +1,44 @@
...
+export interface ExplainOptions {
+  explain?: VerbosityLike;
+}
+
+/** @internal */
+export abstract class ExplainableCommand<
+  T extends ExplainOptions & CommandOperationOptions,
+  TResult = Document
+> extends CommandOperation<T, TResult> {
+  explain?: Explain;
```
nbbeeken, line 15: It's my impression that all commands are, or will be explainable, shouldn't we put the explain logic directly into CommandOperation then? (Maybe that assumption is wrong)

src/explain.ts
```diff
@@ -0,0 +1,78 @@
...
+  executionStats = 'executionStats',
+  allPlansExecution = 'allPlansExecution'
+}
+
+/**
+ * For backwards compatibility, true is interpreted as
+ * "allPlansExecution" and false as "queryPlanner".
+ * @public
+ */
+export type VerbosityLike = Verbosity | boolean;
```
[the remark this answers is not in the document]
nbbeeken: An alternative, (I know we have enums elsewhere we should revisit that):
```typescript
/** @public */
export const Verbosity = {
  queryPlanner: 'queryPlanner',
  queryPlannerExtended: 'queryPlannerExtended',
  executionStats: 'executionStats',
  allPlansExecution: 'allPlansExecution'
} as const
export type VerbosityLike = keyof Verbosity | boolean;
```
Now those who wish an enum like behavior (I suspect only us internally) can import Verbosity, and those who want just to pass the string value, can do so with autocompletion. ✨

src/cursor/cursor.ts
```diff
@@ -1307,10 +1307,11 @@ export class Cursor<
...
     //       subclass `CommandOperationV2`. To be removed asap.
-    if (this.operation && this.operation.cmd == null) {
-      this.operation.options.explain = true;
-      return executeOperation(this.topology, this.operation as any, callback);
-    }
+    // TODO: uncomment/fix this when cursor explain is re-implemented
+    // if (this.operation && this.operation.cmd == null) {
+    //   this.operation.options.explain = true;
+    //   return executeOperation(this.topology, this.operation as any, callback);
+    // }
```
nbbeeken, line 1314: Do you know if there is a ticket for this?

src/explain.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  static fromOptions(options?: ExplainOptions): Explain | undefined {
+    if (options == null || options.explain === undefined) {
+      return;
+    }
+    return new Explain(options.explain);
+  }
+
+  static valid(options?: ExplainOptions): boolean {
+    if (options == null || options.explain === undefined) {
```
nbbeeken, line 48:
```suggestion
    if (!options?.explain) {
```
I think this should do the trick, if options is nullish then `!` makes this true, if explain is nullish then its still true. 
Was it intentional to explicitly check for options.explain being undefined vs `==` double equal nullish?

nbbeeken (changes requested):

src/explain.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @public
+ */
+export type VerbosityLike = Verbosity | boolean;
+
+// Minimum server versions which support explain with specific operations
+const SUPPORTS_EXPLAIN_WITH_REMOVE = 3;
+const SUPPORTS_EXPLAIN_WITH_UPDATE = 3;
+const SUPPORTS_EXPLAIN_WITH_DISTINCT = 3.2;
+const SUPPORTS_EXPLAIN_WITH_FIND_AND_MODIFY = 3.2;
+const SUPPORTS_EXPLAIN_WITH_MAP_REDUCE = 4.4;
```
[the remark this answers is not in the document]
HanaPearlman: Wow! Thank you!

src/operations/explainable_command.ts
HanaPearlman: Not sure about the long term plan of whether all commands will be explainable. At least for now, explain is not supported on `insert`, for example, so `insertOne` cannot be explained. This makes sense to me; since I'm not sure what explain results could even look like for an insert. 

That said, we could roll this logic into the `CommandOperation` class and just not expose explain options on the operations that are not explainable.

src/cursor/cursor.ts
HanaPearlman: Yep, it's one I am working on concurrently (and will hopefully merge immediately after this). I can still put the ticket number for convenience, though.

src/explain.ts
HanaPearlman: Not intentional. Just wanted to avoid missing cases where `options.explain = false`, since we do actually want to explain the command in that case (confusing, I know), so I was checking for `=== undefined` whenever accessing explain from options.

nbbeeken (approved):
lgtm!

src/operations/explainable_command.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  constructor(parent?: OperationParent, options?: T) {
+    super(parent, options);
+
+    if (!Explain.valid(options)) {
+      throw new MongoError(`explain must be one of ${Object.keys(Verbosity)} or a boolean`);
+    }
+
+    this.explain = Explain.fromOptions(options);
```
nbbeeken, line 25: Can we validate inside fromOptions?

src/operations/explainable_command.ts
nbbeeken: Aspects would be a good fit for explainable support we can remove the inheritance here and rely on the aspects presence to permit or remove explainability from command construction 

(chatted offline)

nbbeeken (changes requested):

src/explain.ts
```diff
@@ -0,0 +1,71 @@
+import type { Document } from '.';
+import type { ExplainOptions } from './operations/explainable_command';
+import type { Server } from './sdam/server';
+import { maxWireVersion } from './utils';
+
+export const Verbosity = {
```
emadum, line 6: Nit: can we expand `Verbosity` to `ExplainVerbosity` for clarity?

test/functional/explain.test.js
```diff
@@ -0,0 +1,429 @@
...
+const chai = require('chai');
+const expect = chai.expect;
+chai.use(require('chai-subset'));
+
+describe('Explain', function () {
+  before(function () {
+    return setupDatabase(this.configuration);
+  });
+
+  it('shouldHonorBooleanExplainWithDeleteOne', {
```
emadum, line 12: Nit: test names for new tests should be written in normal text with spaces rather than camel case.

emadum (changes requested):
A few extra nitpicks.

src/index.ts
```diff
@@ -163,6 +163,11 @@ export type {
 export type { DbPrivate, DbOptions } from './db';
 export type { AutoEncryptionOptions, AutoEncryptionLoggerLevels, AutoEncrypter } from './deps';
 export type { AnyError, ErrorDescription } from './error';
+export type {
+  ExplainOptions,
+  ExplainVerbosity as Verbosity,
```
emadum, line 168:
```suggestion
  ExplainVerbosity,
```

src/index.ts
```diff
@@ -163,6 +163,11 @@ export type {
...
+  ExplainVerbosityLike as VerbosityLike
```
emadum, line 169:
```suggestion
  ExplainVerbosityLike
```

emadum (changes requested):

src/index.ts
HanaPearlman: Oops, not sure how that happened!

nbbeeken (approved):
_lgtm_

emadum (approved):

mbroadst (approved):
LGTM!

PR #2601, opened 2020-10-27
fix: apply bson options for bulk operations

HanaPearlman:
This change extends the work from #2589 to bulk operations. The changes involved for `BulkOperationBase` parallel those for `OperationBase`: a new `bsonOptions` field was added (in `s`, so it's at the same level as `options`) with values inherited from the collection parent, it is exposed with a `bsonOptions` getter, and it is included in the options before any `executeOperation` is called.

NODE-1561

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -1285,24 +1285,12 @@ export class Collection implements OperationParent {
 
   /** Initiate an Out of order batch write operation. All operations will be buffered into insert/update/remove commands executed out of order. */
   initializeUnorderedBulkOp(options?: BulkWriteOptions): any {
-    options = options || {};
-    // Give function's options precedence over session options.
-    if (options.ignoreUndefined == null) {
-      options.ignoreUndefined = this.bsonOptions.ignoreUndefined;
-    }
-
-    return new UnorderedBulkOperation(this, options);
+    return new UnorderedBulkOperation(this, options || {});
   }
 
   /** Initiate an In order bulk write operation. Operations will be serially executed in the order they are added, creating a new operation for each switch in types. */
   initializeOrderedBulkOp(options?: BulkWriteOptions): any {
-    options = options || {};
-    // Give function's options precedence over session's options.
-    if (options.ignoreUndefined == null) {
-      options.ignoreUndefined = this.bsonOptions.ignoreUndefined;
-    }
-
-    return new OrderedBulkOperation(this, options);
+    return new OrderedBulkOperation(this, options || {});
   }
 
   /** Get the db scoped logger */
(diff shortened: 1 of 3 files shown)


mbroadst:
@HanaPearlman is this ready for review?

src/collection.ts
```diff
@@ -1285,24 +1285,12 @@ export class Collection implements OperationParent {
...
   /** Initiate an Out of order batch write operation. All operations will be buffered into insert/update/remove commands executed out of order. */
   initializeUnorderedBulkOp(options?: BulkWriteOptions): any {
-    options = options || {};
-    // Give function's options precedence over session options.
-    if (options.ignoreUndefined == null) {
-      options.ignoreUndefined = this.bsonOptions.ignoreUndefined;
-    }
-
-    return new UnorderedBulkOperation(this, options);
+    return new UnorderedBulkOperation(this, options || {});
```
nbbeeken, line 1288:
```suggestion
    return new UnorderedBulkOperation(this, options ?? {});
```

src/collection.ts
```diff
@@ -1285,24 +1285,12 @@ export class Collection implements OperationParent {
...
   /** Initiate an In order bulk write operation. Operations will be serially executed in the order they are added, creating a new operation for each switch in types. */
   initializeOrderedBulkOp(options?: BulkWriteOptions): any {
-    options = options || {};
-    // Give function's options precedence over session's options.
-    if (options.ignoreUndefined == null) {
-      options.ignoreUndefined = this.bsonOptions.ignoreUndefined;
-    }
-
-    return new OrderedBulkOperation(this, options);
+    return new OrderedBulkOperation(this, options || {});
```
nbbeeken, line 1293:
```suggestion
    return new OrderedBulkOperation(this, options ?? {});
```

test/functional/bulk.test.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
+const { withClient } = require('./shared');
 const test = require('./shared').assert,
   setupDatabase = require('./shared').setupDatabase,
   expect = require('chai').expect;
 
 const MongoError = require('../../src/error').MongoError;
 const ignoreNsNotFound = require('./shared').ignoreNsNotFound;
```
nbbeeken, line 8:
```suggestion
const { withClient, setupDatabase, ignoreNsNotFound } = require('./shared');
const test = require('./shared').assert;
const { expect } = require('chai');
const { MongoError } = require('../../src/error');
```

Another one of those, "I know its not yours but while you're here if you don't mind" changes 😅

nbbeeken (commented):
two little style nits, and a suggested improvement, looking good to me 👍

nbbeeken (approved):

emadum (approved):
LGTM 👍

mbroadst (approved):

PR #2604, opened 2020-10-30
refactor: have FindOneOperation extend CommandOperation

HanaPearlman:
This PR changes `FindOneOperation` so it extends `CommandOperation` instead of `OperationBase`. 

Recently, `FindOperation` was modified to extend `CommandOperation` as well, so this change is partially for symmetry and partially because it may help simplify the work required for implementing explain functionality on both operations.

NODE-2852/NODE-2853


mbroadst (approved):

emadum (approved):
👍

PR #2607, opened 2020-11-02
docs(mongo_client): indicate that ssl defaults to `true` for srv connection strings

vkarpov15:
Re: Automattic/mongoose#9511

## Description

If you're using `mongodb+srv://`, the `ssl` option is true by default, here's where it is set: https://github.com/mongodb/node-mongodb-native/blob/79df55383b5ba3f081b96513253e566793166199/lib/core/uri_parser.js#L78-L84 . I think it's worthwhile to document this.

**What changed?**

One line of documentation

No

### DIFF
--- a/lib/mongo_client.js
+++ b/lib/mongo_client.js
@@ -77,7 +77,7 @@ const validOptions = require('./operations/connect').validOptions;
  * @param {string} url The connection URI string
  * @param {object} [options] Optional settings
  * @param {number} [options.poolSize=5] The maximum size of the individual server pool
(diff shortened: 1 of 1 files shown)


agolin95:
Tracked by https://jira.mongodb.org/browse/NODE-2891

nbbeeken:
Hello, thanks for the suggestion, there's been some back and forth about documenting defaults in drivers vs. in the manual, and the thinking is we want to avoid additional documentation that could be inconsistent with the manual since that covers all drivers.

There's a specific sections that mentions the tls options when using an SRV record here.
https://docs.mongodb.com/manual/reference/connection-string/#dns-seed-list-connection-format

PR #2608, opened 2020-11-03
fix: move kerberos client setup from prepare to auth

emadum:
## Description

This PR fixes a bug in Kerberos authentication by moving the Kerberos
client initialization from the `AuthProvider.prepare` override into a
local `getClient` function.

NODE-2859

**What changed?**

### DIFF
--- a/lib/core/auth/gssapi.js
+++ b/lib/core/auth/gssapi.js
@@ -16,7 +16,7 @@ class GSSAPI extends AuthProvider {
     function externalCommand(command, cb) {
       return connection.command('$external.$cmd', command, cb);
     }
-    getClient(authContext, (err, client) => {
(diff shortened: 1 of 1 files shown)


nbbeeken (approved):
LGTM

lib/core/auth/gssapi.js
```diff
@@ -96,6 +52,50 @@ class GSSAPI extends AuthProvider {
 }
 module.exports = GSSAPI;
 
+function getClient(authContext, callback) {
```
mbroadst, line 55: can we make this a little less ambiguous, maybe `makeKerberosClient`? It's unclear to me which client we are referring to in this context

mbroadst (approved):
LGTM, mod a naming nitpick

PR #2609, opened 2020-11-03
refactor!: remove deprecated Collection.group helper

emadum:
## Description

The group command was deprecated in MongoDB 3.4 and removed in
MongoDB 3.6, this PR removes it from the driver.

NODE-1487

**What changed?**


mbroadst:
@emadum did you mean to request reviewers on this?

emadum:
> @emadum did you mean to request reviewers on this?

Oops meant to make this a draft, not quite ready yet.

test/functional/mapreduce.test.js
```diff
@@ -386,7 +330,7 @@ describe('MapReduce', function () {
     }
   });
 
-  it('shouldCorrectlyReturnNestedKeys', {
+  it.skip('shouldCorrectlyReturnNestedKeys', {
```
emadum, line 333: @mbroadst Do you think this test should be rewritten without `collection.group`, or is it fine to just remove it outright:?

mbroadst: I think it's fine to remove outright. It kind of looks like this test is testing the behavior of MongoDB, rather than some behavior of the driver itself

src/operations/group.ts
```diff
@@ -1,101 +0,0 @@
...
-    this.keys = keys;
-    this.condition = condition;
-    this.initial = initial;
-    this.finalize = finalize;
-    this.reduceFunction = reduce && reduce._bsontype === 'Code' ? reduce : new Code(reduce);
-  }
-
-  execute(server: Server, callback: Callback<Document>): void {
-    const cmd: Document = {
-      group: {
```
HanaPearlman, line 40: This is small but: I think this file is the only place where we ever define `cmd.group`. If so, could we delete any leftover references to `cmd.group`? I remember seeing it referenced in at least one place (`commandSupportsReadConcern`) but there may be others.

emadum: Good catch! I did some grepping and I think that was the last remaining reference to `cmd.group`.

nbbeeken (approved):
🎉

HanaPearlman (approved):

mbroadst (approved):

PR #2614, opened 2020-11-06
refactor: consolidate options inheritance logic 

HanaPearlman:
This PR introduces a new helper `resolveOptions` which includes the logic for inheriting some options from a parent. NOTE: It only merges in inherited properties (read/write concern, read preference, and bson options) from the parent, and does not add to or modify any other properties of options. 

Options are resolved before a command is constructed. This means the options passed to `CommandOperation` are already properly inherited, and inheritance is no longer needed in the `CommandOperation` constructor. So, we can get rid of the `NO_INHERIT_OPTIONS` aspect and simply not call `resolveOptions` before those commands. Similarly, the `ReadPreference.resolve()` function is no longer needed.

There is also a change to `ReadPreference.fromOptions()`, where if the options don't have a `readPreference` property and a session is present on the options, we attempt to get the read preference from the session. This might have been intended all along, since the type of the options passed to this function, `ReadPreferenceFromOptions`, already had an optional `session` property. 

NODE-2870

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -459,8 +459,7 @@ export class Db implements OperationParent {
     if (typeof options === 'function') (callback = options), (options = {});
 
     // Intentionally, we do not inherit options from parent for this operation.
-    options = options || {};
-    options.readPreference = ReadPreference.PRIMARY;
+    options = Object.assign({}, options, { readPreference: ReadPreference.PRIMARY });
 
     // Add return new collection
     options.new_collection = true;
--- a/src/db.ts
+++ b/src/db.ts
@@ -266,6 +266,9 @@ export class Db implements OperationParent {
   /**
    * Execute a command
    *
+   * @remarks
+   * This command does not inherit options from the MongoClient.
+   *
    * @param command - The command to run
    * @param options - Optional settings for the command
    * @param callback - An optional callback, a Promise will be returned if none is provided
@@ -428,6 +431,9 @@ export class Db implements OperationParent {
   /**
    * Rename a collection.
    *
+   * @remarks
+   * This operation does not inherit options from the MongoClient.
+   *
    * @param fromCollection - Name of current collection to rename
    * @param toCollection - New name of of the collection
    * @param options - Optional settings for the command
@@ -544,6 +550,9 @@ export class Db implements OperationParent {
 
   /**
    * Runs a command on the database as admin.
+   *
+   * @remarks
+   * This command does not inherit options from the MongoClient.
    *
    * @param command - The command to run
    * @param options - Optional settings for the command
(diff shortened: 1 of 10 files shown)


src/read_preference.ts
```diff
@@ -164,36 +164,6 @@ export class ReadPreference {
     return readPreference as ReadPreference;
   }
```
HanaPearlman, line 166: Note: `ReadPreference.fromOptions` should not default to `primary` if it cannot find a read pref in the options. In `resolveOptions` when `ReadPreference.fromOptions` is undefined, we default to the parent read pref. If `fromOptions` always defaults to primary, we would never use the parent read pref.

src/operations/command.ts
```diff
@@ -73,12 +73,14 @@ export abstract class CommandOperation<
         : new MongoDBNamespace('admin', '$cmd');
     }
 
-    const propertyProvider = this.hasAspect(Aspect.NO_INHERIT_OPTIONS) ? undefined : parent;
+    const readPref = ReadPreference.fromOptions(options) ?? ReadPreference.primary;
```
mbroadst, line 76: do you actually reuse `readPref` here? Should it just be inlined in the ternary?

src/utils.ts
```diff
@@ -1108,3 +1073,38 @@ export function hasAtomicOperators(doc: Document | Document[]): boolean {
...
+  }
+
+  const readPreference = ReadPreference.fromOptions(options) ?? parent?.readPreference;
+  if (readPreference) {
+    result.readPreference = readPreference;
+  }
+
+  const bsonOptions = resolveBSONOptions(result, parent);
+
+  return { ...result, ...bsonOptions };
```
mbroadst, line 1109: I think we can reduce allocations a bit by resolving the BSON objects early, otherwise you're assigning all of them using `Object.assign` above just to spread them here again. In this case `resolveBSONOptions` is more like a "santize data + provide defaults", so what if we did:
```
const result: T = Object.assign({ ...resolveBSONOptions(options, parent) }, options, );
```

Second observation: I'm sure I used `Object.assign` to get around issues with typing, but it may be worth investigating whether it's possible to gain stricter typing by spreading here for all options:
```
const result: T = { ...options, ...resolveBSONOptions(options, parent) };
```

mbroadst (approved):
LGTM!

src/db.ts
```diff
@@ -281,11 +280,11 @@ export class Db implements OperationParent {
     callback?: Callback<Document>
   ): Promise<Document> | void {
     if (typeof options === 'function') (callback = options), (options = {});
-    options = options || {};
 
+    // Intentionally, we do not inherit options from parent for this operation.
```
nbbeeken, line 284: I think we should add this to the function's docs for the ops that do this. Particularly with BSON options the return value could be unexpectedly different.

nbbeeken (approved):
LGTM-ing, my only comment is that we can do some documenting, the behavior isn't new so it's not critical but could be a good improvement. 👍

emadum (approved):

PR #2615, opened 2020-11-06

HanaPearlman:
This reinstates the work from #2581 now that NODE-2850 is complete. 

During `close` of a `MongoClient`, we remove the `topology`. Then, when a client is reconnected, a new topology is created and added to the client. First, this means references to `Db`s and `Collection`s can be re-used after the reconnect, since they will pull the new topology from the client. Also, when an operation requiring a topology is performed on a closed client, an error is thrown. 

After this, the work of figuring out how to re-connect a topology is one of the last steps towards having a single, immutable topology associated with a `MongoClient`. As in #2581, I skipped this work for now, since the refactor required to re-connect topologies may be too much work for this ticket.

NODE-2544

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -588,7 +588,6 @@ function processNewChange(
 }
 
 function processError(changeStream: ChangeStream, error: AnyError, callback?: Callback) {
-  const topology = getTopology(changeStream.parent);
   const cursor = changeStream.cursor;
 
   // If the change stream has been closed explicitly, do not process error.
@@ -618,6 +617,7 @@ function processError(changeStream: ChangeStream, error: AnyError, callback?: Ca
     // close internal cursor, ignore errors
     cursor.close();
 
+    const topology = getTopology(changeStream.parent);
     waitForTopologyConnected(topology, { readPreference: cursor.readPreference }, err => {
       // if the topology can't reconnect, close the stream
       if (err) return unresumableError(err);
--- a/src/operations/connect.ts
+++ b/src/operations/connect.ts
@@ -197,6 +197,11 @@ export function connect(
     throw new Error('no callback function provided');
   }
 
+  // Has a connection already been established?
+  if (mongoClient.topology && mongoClient.topology.isConnected()) {
+    throw new MongoError(`'connect' cannot be called when already connected`);
+  }
+
   let didRequestAuthentication = false;
   const logger = new Logger('MongoClient', options);
 
--- a/src/operations/connect.ts
+++ b/src/operations/connect.ts
@@ -199,7 +199,7 @@ export function connect(
 
   // Has a connection already been established?
   if (mongoClient.topology && mongoClient.topology.isConnected()) {
-    throw new MongoError(`'connect' cannot be called when already connected`);
+    return callback(new MongoError('MongoClient is already connected'));
   }
 
   let didRequestAuthentication = false;
--- a/src/operations/connect.ts
+++ b/src/operations/connect.ts
@@ -197,9 +197,9 @@ export function connect(
     throw new Error('no callback function provided');
   }
 
-  // Has a connection already been established?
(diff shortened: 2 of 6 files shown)


src/change_stream.ts
```diff
@@ -618,6 +617,7 @@ function processError(changeStream: ChangeStream, error: AnyError, callback?: Ca
     // close internal cursor, ignore errors
     cursor.close();
 
+    const topology = getTopology(changeStream.parent);
     waitForTopologyConnected(topology, { readPreference: cursor.readPreference }, err => {
```
HanaPearlman, line 621: Accessing the topology too early in this function causes failing tests when `processError` is called after client close.

test/functional/connection.test.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
+const { withClient, setupDatabase } = require('./shared');
 const test = require('./shared').assert,
-  setupDatabase = require('./shared').setupDatabase,
   expect = require('chai').expect;
```
nbbeeken, line 4:
```js
const { withClient, setupDatabase } = require('./shared');
const test = require('./shared').assert;
```
Couldn't make a suggestion for some reason but IYDM slowly but surely we will have consistent importing 😅

test/functional/connection.test.js
```diff
@@ -274,3 +274,55 @@ describe('Connection', function () {
...
+
+        client.connect(err => {
+          expect(err).to.not.exist;
+          expect(client.isConnected()).to.be.true;
+
+          collection.insertOne({ b: 2 }, (err, result) => {
+            expect(err).to.not.exist;
+            expect(result).to.exist;
+            expect(client.topology.isDestroyed()).to.be.false;
+            done();
```
nbbeeken, line 300: Can this instead be a find on the previous insertion instead of another insert I guess just to show the round trip of connecting back to the same mongodb with a collection instance from the previous connected context.

src/operations/connect.ts
```diff
@@ -197,6 +197,11 @@ export function connect(
     throw new Error('no callback function provided');
   }
 
+  // Has a connection already been established?
+  if (mongoClient.topology && mongoClient.topology.isConnected()) {
+    throw new MongoError(`'connect' cannot be called when already connected`);
+  }
```
nbbeeken, line 203: Error message nit: `'MongoClient is already connected'` I just think you don't need to mention 'calling' vs describing the reason for the error.

I think this error should be sent to the callback rather than being thrown because it doesn't prevent you from using the client that is connected, its just informing that this call is a no-op.

nbbeeken (changes requested):
🚀

src/operations/connect.ts
HanaPearlman: This is a good point and actually makes me wonder if it is better to just return early, since it's just a no-op and not actually an error?

nbbeeken: That's a good question I guess its something that's harmless to go undetected, there's no serious work done before we get to this point in the code.
 
@kmahar any thoughts here?

kmahar: what's the current behavior if you call `close` multiple times in a row? it would seem to me that it makes sense for it to be symmetric: either multiple calls in a row are no-ops, or are not allowed. 

more generally, what's the use case for closing a client and then reconnecting?

nbbeeken: close() is a no-op if already closed so connect can be a no-op too 👍

HanaPearlman: Closing a client multiple times is a no-op, so I'll go ahead and just return without an error here. 

Not entirely sure what the exact use case is for closing a client and then reconnecting, but there have been multiple related tickets filed/stack overflow posts about it-- since it's not too much effort it seems worth it to implement.

nbbeeken (approved):

emadum (approved):
LGTM! 👍

PR #2616, opened 2020-11-09
feat: Add MongoOptions interface

nbbeeken:
Adds MongoOptions representing the internal view of the specified options.

NODE-2698

A number of options are seemingly omitted but instead are brought in by extending NodeJS's connection options. Particularly with tls options we have a number of translations from ssl to tls options and then another set of translations to NodeJS's equivalent definition. I've left documentation of the equivalent names and they will be resolved by the following work.


src/mongo_client_options.ts
```diff
@@ -0,0 +1,143 @@
...
+import type { TcpSocketConnectOpts as ConnectionOptions } from 'net';
+import type { BSONSerializeOptions } from './bson';
+import { MongoParseError } from './error';
+import { URL } from 'url';
+
+/**
+ * Mongo Client Options
+ * @internal
+ */
+export interface MongoOptions
```
mbroadst, line 20: After our conversation yesterday, I'm wondering if `MongoOptions` shouldn't just be a "resolved" version of `MongoClientOptions`? Should `MongoOptions` basically be:
```
export type MongoOptions = Omit<MongoClientOptions, ....> & {
  moreSpecificOptions: any;
  newOptionNames: any
}
```

I wonder if that also means that you propagate the doc strings for driver dev use?

nbbeeken: Good idea! 👍  It does bring along the docs which is very nice. And in the case of tls I can redo that one prop in MongoOptions to keep the table I made as internal docs.

src/mongo_client_options.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (!match) {
+    throw new MongoParseError(`Invalid connection string ${uri}`);
+  }
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //@ts-expect-error
+  const { protocol, username, password, hosts, rest } = match.groups;
+  if (!protocol || !hosts) {
+    throw new MongoParseError('Invalid connection string, protocol and host(s) required');
+  }
+  const authString = `${username ? `${password ? `${username}:${password}` : username}` : ''}`;
```
emadum, line 117:
```suggestion
  const authString = username ? password ? `${username}:${password}` : `${username}` : '';
```

emadum (approved):
LGTM, with one small readability nit

src/mongo_client.ts
```diff
@@ -554,3 +572,92 @@ export class MongoClient extends EventEmitter implements OperationParent {
...
+  tls: boolean;
+
+  /**
+   * Turn these options into a reusable options dictionary
+   */
+  toJSON(): Record<string, any>;
+  /**
+   * Turn these options into a reusable connection URI
+   */
+  toURI(): string;
```
mbroadst, line 662:
```suggestion
  toConnectionString(): string;
```
just a thought? I'm the one that introduced calling it a URI, but I think the consensus across the drivers and the company is calling it a connection string

mbroadst (approved):
LGTM, left a single thought but I don't feel strong about it at all

src/mongo_client.ts
nbbeeken: [Connection String URI Format](https://docs.mongodb.com/manual/reference/connection-string/)
It's one of those both ways is right, I'll leave it as is, and we can consider renaming it when we get to the ticket that implements it since there might be some questions about how we want to expose it. These options are supposed to be internal, kind of, so is it something we'd want people to use: `client.options.toURI()` or would we surface something on the client itself.

PR #2617, opened 2020-11-09
refactor: consolidate options inheritance for bulk ops

HanaPearlman:
This PR extends the work from #2614 to bulk operations. Now, bulk ops use `resolveOptions` to inherit options from the parent. Also, since no bson options or read preference resolving needs to be done in subclasses of `OperationBase`, assigning the `bsonOptions` and `readPreference` properties now occurs in the `OperationBase` constructor.

The changes to `find_and_modify.ts`, `insert.ts`, and `common_functions.ts` should have been done in the last PR :(

NODE-2870

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -21,7 +21,6 @@ import type { Topology } from '../sdam/topology';
 import type { CommandOperationOptions } from '../operations/command';
 import type { CollationOptions } from '../cmap/wire_protocol/write_command';
 import type { Hint } from '../operations/operation';
-import { ReadPreference } from '../read_preference';
 
 // Error codes
 const WRITE_CONCERN_ERROR = 64;
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -21,6 +21,7 @@ import type { Topology } from '../sdam/topology';
 import type { CommandOperationOptions } from '../operations/command';
 import type { CollationOptions } from '../cmap/wire_protocol/write_command';
 import type { Hint } from '../operations/operation';
+import { ReadPreference } from '../read_preference';
 
 // Error codes
 const WRITE_CONCERN_ERROR = 64;
@@ -572,8 +573,8 @@ function executeCommands(
   }
 
   const finalOptions = resolveOptions(bulkOperation, {
-    ordered: bulkOperation.isOrdered,
-    ...options
+    ...options,
+    ordered: bulkOperation.isOrdered
   });
 
   if (finalOptions.bypassDocumentValidation !== true) {
@@ -1169,18 +1170,13 @@ export abstract class BulkOperationBase {
   }
 
   /** An internal helper method. Do not invoke directly. Will be going away in the future */
-  execute(
-    _writeConcern?: WriteConcern,
-    options?: BulkWriteOptions,
-    callback?: Callback<BulkWriteResult>
-  ): Promise<void> | void {
+  execute(options?: BulkWriteOptions, callback?: Callback<BulkWriteResult>): Promise<void> | void {
     if (typeof options === 'function') (callback = options), (options = {});
     options = options || {};
 
-    if (typeof _writeConcern === 'function') {
-      callback = _writeConcern as Callback;
-    } else if (_writeConcern && typeof _writeConcern === 'object') {
-      this.s.writeConcern = _writeConcern;
+    const writeConcern = WriteConcern.fromOptions(options);
+    if (writeConcern) {
+      this.s.writeConcern = writeConcern;
     }
 
     if (this.s.executed) {
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -2,14 +2,14 @@ import { PromiseProvider } from '../promise_provider';
 import { Long, ObjectId, Document, BSONSerializeOptions, resolveBSONOptions } from '../bson';
 import { MongoError, MongoWriteConcernError, AnyError } from '../error';
 import {
-  applyWriteConcern,
   applyRetryableWrites,
   executeLegacyOperation,
   hasAtomicOperators,
   Callback,
   MongoDBNamespace,
   maxWireVersion,
-  getTopology
+  getTopology,
+  resolveOptions
 } from '../utils';
 import { executeOperation } from '../operations/execute_operation';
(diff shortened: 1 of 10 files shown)


src/operations/common_functions.ts
```diff
@@ -218,16 +211,9 @@ export function updateDocuments(
...
+  // Final options for retryable writes
   let finalOptions = Object.assign({}, options);
   finalOptions = applyRetryableWrites(finalOptions, coll.s.db);
-  finalOptions = applyWriteConcern(finalOptions, { db: coll.s.db, collection: coll }, options);
-
-  // Do we return the actual result document
-  // Either use override on the function, or go back to default on either the collection
-  // level or db
-  finalOptions.serializeFunctions =
-    options.serializeFunctions || coll.bsonOptions.serializeFunctions;
```
HanaPearlman, line 217: Oops! These lines should have been deleted in the last PR.

src/cmap/wire_protocol/write_command.ts
```diff
@@ -64,6 +65,12 @@ export function writeCommand(
     writeCommand.bypassDocumentValidation = options.bypassDocumentValidation;
   }
 
+  // A read preference may be present in the options, inherited from a transaction or parent.
+  // For write commands, we need to ensure it is ReadPreference.primary, so we override it here.
+  if (options.readPreference) {
+    options.readPreference = ReadPreference.primary;
+  }
+
   const commandOptions = Object.assign(
```
HanaPearlman, line 74: We need to check for this case. If a read preference is in the options, it will be included in the command, even if that command is a write.

We are already doing the equivalent check for write operations which pass through `CommandOperation`, but for those that just subclass `OperationBase`, we needed a separate check.

 If this approach is too buried/hacky for us, an alternative would be manually setting the read preference to primary in the options for *each* write command, either before we construct the write command or in the constructor of the command (which is what we do in `CommandOperation`).

mbroadst: My inclination is that we would want to do the hard thing here and explicitly codify a fixed primary read preference at the point that operations are constructed/executed, if only just for readability of the codebase. The paths for "write commands" in the wire protocol layer are symbolic, and ideally would go away if we were doing command construction correctly: in a 3.6+ world we would always just use the `command` wire protocol method. 

This is another form of options resolution, but now with restrictions. We try to handle this during construction of the `CommandOperation` (as you note) by use of a `WRITABLE` aspect, codifying the restriction into the "type system". It actually looks to me that the resolution in `CommandOperation` could just be moved to `OperationBase` (I believe you're also proposing this), that seems to be where it belongs. A `readPreference` is a fundamental aspect of an operation as it is used to determine the server selected, so I think it does belong on the `OperationBase` class - and you should be able to determine if the operation is writable there as well.

Another (more functional) approach might be to encode this through a chain of functions:
```
new InsertOneOperation(topology, ..., fixedPrimaryReadPreference(resolveOptions(this, options)))`)
```
but that might get a little overwhelming as we introduce more restrictions in resolution

src/bulk/common.ts
```diff
@@ -1169,6 +1164,10 @@ export abstract class BulkOperationBase {
     return this.s.bsonOptions;
   }
 
+  get writeConcern(): WriteConcern | undefined {
```
mbroadst, line 1167: was this added so you can use `BulkOperationBase` as an `OperationParent`?

src/operations/bulk_write.ts
```diff
@@ -50,11 +46,8 @@ export class BulkWriteOperation extends OperationBase<BulkWriteOptions, BulkWrit
       return callback(err);
     }
 
-    // Final options for retryable writes and write concern
     let finalOptions = Object.assign({}, options);
     finalOptions = applyRetryableWrites(finalOptions, coll.s.db);
-    finalOptions = applyWriteConcern(finalOptions, { db: coll.s.db, collection: coll }, options);
-
     const writeCon = WriteConcern.fromOptions(finalOptions);
```
mbroadst, line 51: why was this `writeCon` left here? It seems like it should be otherwise solved through options resolution

mbroadst (changes requested):

src/bulk/common.ts
HanaPearlman: Technically we could already use it as an `OperationParent` (`writeConcern` is an optional property), but this change was so `resolveOptions` can correctly pull the `writeConcern` from `BulkOperationBase`.

Happy to change the argument type in `resolveOptions` to include `BulkOperationBase` or do something else to make this more visible.

src/operations/bulk_write.ts
HanaPearlman: The options resolution ensures that `options` has the right `writeConcern`, but in this case we still have to pull the `WriteConcern` object from the options to pass to `bulk.execute(writeConcern, options, callback)`. Happy to remove this argument from `bulk.execute(...)`, since it could just be pulled from the options within the method.

src/bulk/common.ts
mbroadst: I was actually thinking that maybe you could add a `readPreference` property which always returns `primary`, not sure if that helps at all.. Does it make sense to explicitly have things we consider operation parents `implement OperationParent`?

src/operations/bulk_write.ts
mbroadst: oh yeah, let's definitely do that!

src/bulk/common.ts
HanaPearlman: I don't think it helps to add a `readPreference` property because we actually need to modify the read pref in `options`, not the parent read preference, to avoid defaulting to a transaction read pref when doing a write. Going to do what you suggested and move the read pref resolving and corresponding options modifying to `OperationBase`, and that solves the problem.

As for the `OperationParent` issue, I am not sure if we really consider `BulkOperationBase` an `OperationParent`? Similarly, it doesn't really seem like `OperationBase` is an `OperationParent`. This is why I was wondering about just adding `BulkOperationBase` as an accepted type for the parent in `resolveOptions`

mbroadst: SGTM

mbroadst (approved):
LGTM, nice work here

nbbeeken (approved):
🚀

PR #2621, opened 2020-11-13
fix: make MongoBulkWriteError conform to CRUD spec

HanaPearlman:
This PR makes `BulkWriteError` conform to the CRUD spec by adding the necessary fields required by the spec tests to the class and renames `BulkWriteError` to `MongoBulkWriteError`.

NODE-1989, NODE-2331

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -666,38 +666,38 @@ function handleMongoWriteConcernError(
  * @category Error
  */
 export class MongoBulkWriteError extends MongoError {
-  result?: BulkWriteResult;
+  result: BulkWriteResult;
 
   /** Number of documents inserted. */
-  insertedCount?: number;
+  insertedCount: number;
   /** Number of documents matched for update. */
-  matchedCount?: number;
+  matchedCount: number;
   /** Number of documents modified. */
-  modifiedCount?: number;
+  modifiedCount: number;
   /** Number of documents deleted. */
-  deletedCount?: number;
+  deletedCount: number;
   /** Number of documents upserted. */
-  upsertedCount?: number;
+  upsertedCount: number;
   /** Inserted document generated Id's, hash key is the index of the originating operation */
-  insertedIds?: { [key: number]: ObjectId };
+  insertedIds: { [key: number]: ObjectId };
   /** Upserted document generated Id's, hash key is the index of the originating operation */
-  upsertedIds?: { [key: number]: ObjectId };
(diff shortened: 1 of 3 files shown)


src/bulk/common.ts
```diff
@@ -660,16 +665,39 @@ function handleMongoWriteConcernError(
...
+  deletedCount?: number;
+  /** Number of documents upserted. */
+  upsertedCount?: number;
+  /** Inserted document generated Id's, hash key is the index of the originating operation */
+  insertedIds?: { [key: number]: ObjectId };
+  /** Upserted document generated Id's, hash key is the index of the originating operation */
+  upsertedIds?: { [key: number]: ObjectId };
+
+  /** Creates a new MongoBulkWriteError */
   constructor(error?: AnyError, result?: BulkWriteResult) {
```
emadum, line 687: I think we should make the `error` and `result` properties/constructor options non-optional. These errors are always constructed with both arguments.

src/bulk/common.ts
```diff
@@ -660,16 +665,39 @@ function handleMongoWriteConcernError(
...
     super(error as Error);
     Object.assign(this, error);
 
-    this.name = 'BulkWriteError';
+    this.name = 'MongoBulkWriteError';
     this.result = result;
+
+    this.insertedCount = result?.insertedCount;
+    this.matchedCount = result?.matchedCount;
+    this.modifiedCount = result?.modifiedCount || 0;
```
emadum, line 696: Don't need to provide a default here, `modifiedCount` is already guaranteed to be a number in `BulkWriteResult`.

src/bulk/common.ts
```diff
@@ -660,16 +665,39 @@ function handleMongoWriteConcernError(
  * @public
  * @category Error
  */
-export class BulkWriteError extends MongoError {
+export class MongoBulkWriteError extends MongoError {
   result?: BulkWriteResult;
 
-  /** Creates a new BulkWriteError */
+  /** Number of documents inserted. */
+  insertedCount?: number;
```
emadum, line 672: If `result` isn't optional (see lower comment) none of these new properties need to be optional either.

emadum (changes requested):
LGTM! I think we can take this PR as an opportunity to tighten up some properties which are currently optional but don't need to be.

nbbeeken (approved):
Eric's suggestions seem to cover it, lgtm!

mbroadst (approved):
LGTM

emadum (approved):

PR #2622, opened 2020-11-16
feat: add explain support for cursor commands 

HanaPearlman:
This PR implements explain functionality for `findOne`, `find`, and `aggregate`. 

These operations can be explained via the `explain` option introduced in #2599 or with the `explain` method on cursors, which now takes an optional `verbosity` parameter (defaults to `true` for backwards compatibility).

### DIFF
--- a/test/functional/aggregation.test.js
+++ b/test/functional/aggregation.test.js
@@ -386,12 +386,7 @@ describe('Aggregation', function () {
    * @example-class Collection
    * @example-method aggregate
    */
-  it.skip('should correctly return a cursor and call explain', {
-    // TODO NODE-2853: This had to be skipped during NODE-2852; un-skip while re-implementing
-    // cursor explain
-
-    // Add a tag that our runner can trigger on
-    // in this case we are setting that node needs to be higher than 0.10.X to run
+  it('should correctly return a cursor and call explain', {
     metadata: {
       requires: {
         mongodb: '>2.5.3',
@@ -461,7 +456,7 @@ describe('Aggregation', function () {
           cursor.explain(function (err, result) {
             expect(err).to.not.exist;
             expect(result.stages).to.have.lengthOf.at.least(1);
-            expect(result.stages[0]).to.have.key('$cursor');
+            expect(result.stages[0]).to.have.property('$cursor');
 
             client.close(done);
           });
@@ -928,7 +923,7 @@ describe('Aggregation', function () {
     }
   });
 
-  it('should fail if you try to use explain flag with readConcern/writeConcern', {
+  it('should fail if you try to use explain flag with writeConcern', {
     metadata: {
       requires: {
         mongodb: '>3.6.0',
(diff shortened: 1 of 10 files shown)


test/functional/aggregation.test.js
```diff
@@ -938,12 +933,9 @@ describe('Aggregation', function () {
...
-      var client = this.configuration.newClient(this.configuration.writeConcernMax(), {
-        poolSize: 1
-      });
+      var client = this.configuration.newClient({ poolSize: 1 });
 
       const testCases = [
-        { readConcern: { level: 'local' } },
         { writeConcern: { j: true } },
         { readConcern: { level: 'local' }, writeConcern: { j: true } }
       ];
```
HanaPearlman, line 941: This test was passing because the write concern was inherited from the client, making it look like using read concern with explain would fail on the server, when really it was the inherited write concern causing the server to return an error.

nbbeeken (approved):
LGTM 👍

test/functional/aggregation.test.js
emadum: I'm a little confused about this, why do we want to stop throwing an error if you use explain + readConcern? To me it looks like this test would still be passing due to the an error generated by the driver if not for the change [here](https://github.com/mongodb/node-mongodb-native/pull/2622/files#diff-9cf3d2934bb06056f35b09e0e20d67d0da15324fe2090c45a4a54dc417fab74dR68).

HanaPearlman: The server doesn't actually throw an error when you just use explain & readConcern, so it is unnecessary to throw from the driver in this case. If you remove the driver error, the server would return an error for this test case because the write concern is inherited from the client, meaning explain is being used with writeConcern, not because there is any issue with explain & readConcern.  Does that help?

emadum: To clarify what I'm confused about: this test was expecting the failure to happen client-side rather than the server-side, and from what I can see it was only the change to `AggregateOperation` that caused this client-side error to stop happening, so I'm not sure I see how the server throwing an error because of the inherited write concern would effect this test. I'd expect a server-side error to cause the test to fail with the `Expected aggregation to fail on client instead of server for option` message.

I think we were using this error as a hint to the user that read concern isn't supported by the explain operation, rather than merely as a way of avoiding a server-side error, and that we're no longer providing that guidance to the user with this change. I can see an argument in favor of this - they just want to get their explain results, and we should just ignore the read concern the way the server does; just wanted to fully understand the reason for the change.

HanaPearlman: Ah, thanks for the clarification, I was misunderstanding.

I agree with what you're saying at the end. I think it makes sense to not error, since the server also ignores the read concern and since it's not really an error condition, in my opinion, in that nothing has gone wrong/is in an invalid state. I could imagine including a warning or possibly something in the documentation (for all of the explainable operations, not just aggregate), if we feel like the user needs this guidance. What do you think?

emadum: I agree it seems unnecessary to warn about a read concern on explain operations. Is there any additional context you could provide about this @mbroadst ?

mbroadst (approved):
LGTM!

emadum (approved):

test/functional/aggregation.test.js
mbroadst: Unfortunately, I can't recall the rationale for that check, I think it was added before my time and we don't have any specifications covering this. We have a similar situation [here](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/core/wireprotocol/query.js#L199-L203) where we throw a client-side error if a non-local `readConcern` is provided for servers which don't support it - the server would otherwise ignore the value, it's not clear why we throw an error client-side.

I'm pro this change if it doesn't break any of our tests

PR #2624, opened 2020-11-17
feat: Deprecate top-level write concern option keys 

HanaPearlman:
This PR is mostly documentation changes to indicate that keys `w`, `j`, `wtimeout` and `fsync` are deprecated. 

Code changes include:
- `WriteConcern.fromOptions` now warns when the options contain the top-level keys and _not_ the `writeConcern` key.
- During client creation, when creating options from a URL, the write concern options are moved under the `writeConcern` key, so that warnings are not logged later. Options passed directly to the client are also moved under the `writeConcern` key, but only after `WriteConcern.fromOptions` is called, ensuring at least one warning is logged.
- GridFS now includes `options.writeConcern`, if it exists, when getting the write concern. 

Tests now spit out a _ton_ of warns because the vast majority of tests which pass write concern options to client/operations use the top level keys.

### DIFF
--- a/lib/gridfs/grid_store.js
+++ b/lib/gridfs/grid_store.js
@@ -73,10 +73,11 @@ const deprecationFn = deprecate(() => {},
  * @param {string} [filename] optional filename for this file, no unique constrain on the field
  * @param {string} mode set the mode for this file.
  * @param {object} [options] Optional settings.
- * @param {(number|string)} [options.w] The write concern.
- * @param {number} [options.wtimeout] The write concern timeout.
- * @param {boolean} [options.j=false] Specify a journal write concern.
- * @param {boolean} [options.fsync=false] Specify a file sync write concern.
+ * @param {(number|string)} [options.w] DEPRECATED: The write concern. Use writeConcern instead.
+ * @param {number} [options.wtimeout] DEPRECATED: The write concern timeout. Use writeConcern instead.
+ * @param {boolean} [options.j=false] DEPRECATED: Specify a journal write concern. Use writeConcern instead.
+ * @param {boolean} [options.fsync=false] DEPRECATED: Specify a file sync write concern. Use writeConcern instead.
+ * @param {object|WriteConcern} [options.writeConcern] Specify write concern settings.
  * @param {string} [options.root] Root collection to use. Defaults to **{GridStore.DEFAULT_ROOT_COLLECTION}**.
  * @param {string} [options.content_type] MIME type of the file. Defaults to **{GridStore.DEFAULT_CONTENT_TYPE}**.
  * @param {number} [options.chunk_size=261120] Size for the chunk. Defaults to **{Chunk.DEFAULT_CHUNK_SIZE}**.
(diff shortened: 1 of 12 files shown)


lib/write_concern.js
```diff
@@ -37,7 +37,12 @@ class WriteConcern {
   /**
    * Construct a WriteConcern given an options object.
    *
-   * @param {object} options The options object from which to extract the write concern.
+   * @param {object} [options] The options object from which to extract the write concern.
+   * @param {(number|string)} [options.w] **Deprecated** Use `options.writeConcern` instead
+   * @param {number} [options.wtimeout] **Deprecated** Use `options.writeConcern` instead
+   * @param {boolean} [options.j] **Deprecated** Use `options.writeConcern` instead
+   * @param {boolean} [options.fsync] **Deprecated** Use `options.writeConcern` instead
```
nbbeeken, line 44: Let's be consistent with the Deprecated tag in the docs, I think the use of bold is good so can we change the `DEPRECATED:` tags to match?

lib/gridfs/grid_store.js
```diff
@@ -73,10 +73,11 @@ const deprecationFn = deprecate(() => {},
...
  * @param {string} mode set the mode for this file.
  * @param {object} [options] Optional settings.
- * @param {(number|string)} [options.w] The write concern.
- * @param {number} [options.wtimeout] The write concern timeout.
- * @param {boolean} [options.j=false] Specify a journal write concern.
- * @param {boolean} [options.fsync=false] Specify a file sync write concern.
+ * @param {(number|string)} [options.w] DEPRECATED: The write concern. Use writeConcern instead.
+ * @param {number} [options.wtimeout] DEPRECATED: The write concern timeout. Use writeConcern instead.
+ * @param {boolean} [options.j=false] DEPRECATED: Specify a journal write concern. Use writeConcern instead.
+ * @param {boolean} [options.fsync=false] DEPRECATED: Specify a file sync write concern. Use writeConcern instead.
```
nbbeeken, line 79: https://docs.mongodb.com/manual/reference/write-concern/index.html
Fsync isn't an option anymore, its equivalent to journal, maybe this is a good opportunity to write a message: "since version x use journal instead"

nbbeeken (changes requested):
looking mostly good!
As the for the test output sometime it turns into a WALL of deprecations 😆  
I'd take a look at `test/tools/runner/config.js:225` specifically the `writeConcernMax` method and see if changing that to the correct format will clean most of it up.

test/functional/bulk.test.js
```diff
@@ -760,7 +760,7 @@ describe('Bulk', function() {
           batch.insert({ b: 1 });
 
           // Execute the operations
-          batch.execute(self.configuration.writeConcernMax(), function(err, result) {
+          batch.execute(self.configuration.writeConcernMax().writeConcern, function(err, result) {
             expect(err).to.exist;
```
HanaPearlman, line 764: `writeConcernMax` was changed to return a `writeConcern` formatted the new way-- `writeConcern: {w:1, ...}`. Bulk execute takes an actual `WriteConcern` object as its first parameter (this was changed in master), so we have to un-wrap the `writeConcernMax`  result here.

test/tools/runner/config.js
```diff
@@ -224,10 +230,10 @@ class NativeConfiguration {
 
   writeConcernMax() {
     if (this.topologyType !== TopologyType.Single) {
-      return { w: 'majority', wtimeout: 30000 };
+      return { writeConcern: { w: 'majority', wtimeout: 30000 } };
     }
 
-    return { w: 1 };
+    return { writeConcern: { w: 1 } };
   }
```
HanaPearlman, line 238: Per @nbbeeken 's suggestion. The benefit of making this change should mean about half as many warnings during testing.

emadum (approved):
LGTM 👍 

One minor issue, `Db.prototype.dropDatabase` doesn't have `writeConcern` documented as an option but I think it does support it. While I don't think it needs to hold up this PR, we should probably go through and audit the `Collection`/`Db` objects to make sure all commands that support `writeConcern` have it listed in the jsdoc.

nbbeeken (approved):

wolrajhti:
Hi everyone, it seems that the new format for writeConcern options is not used everywhere in the driver (I think at least in ./lib/utils.js#mergeOptionsAndWriteConcern). So even when we pass the new format, we still get the depreciation message over and over.

https://jira.mongodb.org/browse/NODE-3114

PR #2625, opened 2020-11-18
fix: tests using lib prohibited

nbbeeken:
One of our tests was using lib which stops the tests from running if you have a clean repo, this little fix changes the import and adds a linter rule to catch this for us going forward.

### DIFF
--- a/test/.eslintrc.json
+++ b/test/.eslintrc.json
@@ -29,6 +29,9 @@
     "jsdoc/require-returns-type": "off",
     "jsdoc/valid-types": "off",
 
+    // Since we use ts-node we should always require the TS code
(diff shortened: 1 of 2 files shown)


test/.eslintrc.json
```diff
@@ -29,6 +29,9 @@
     "jsdoc/require-returns-type": "off",
     "jsdoc/valid-types": "off",
 
+    // Since we use ts-node we should always require the TS code
+    "no-restricted-modules": ["error", { "patterns": ["**/../lib/**"] }],
```
nbbeeken, line 33: The pattern is `../lib` otherwise it matched a `mocha/lib` and `bson/lib` import we use

mbroadst (approved):

emadum (approved):

PR #2626, opened 2020-11-18
feat: add explain support

HanaPearlman:
Port explain support to 3.6.

The logical changes are the same, but some things had to occur in different places: Since `OperationBase` is the only shared class between all of the explainable ops, the `explain` property is on `OperationBase`, and `decorateWithExplain` is used in a few more places.

### DIFF
--- a/lib/collection.js
+++ b/lib/collection.js
@@ -287,7 +287,6 @@ const DEPRECATED_FIND_OPTIONS = ['maxScan', 'fields', 'snapshot', 'oplogReplay']
  * @param {object} [options.fields] **Deprecated** Use `options.projection` instead
  * @param {number} [options.skip=0] Set to skip N documents ahead in your query (useful for pagination).
  * @param {Object} [options.hint] Tell the query to use specific indexes in the query. Object of indexes to use, {'_id':1}
- * @param {string|boolean} [options.explain] The verbosity mode for the explain output.
  * @param {boolean} [options.snapshot=false] DEPRECATED: Snapshot query.
  * @param {boolean} [options.timeout=false] Specify if the cursor can timeout.
  * @param {boolean} [options.tailable=false] Specify if the cursor is tailable.
@@ -310,6 +309,7 @@ const DEPRECATED_FIND_OPTIONS = ['maxScan', 'fields', 'snapshot', 'oplogReplay']
  * @param {boolean} [options.noCursorTimeout] The server normally times out idle cursors after an inactivity period (10 minutes) to prevent excess memory use. Set this option to prevent that.
  * @param {object} [options.collation] Specify collation (MongoDB 3.4 or higher) settings for update operation (see 3.4 documentation for available fields).
  * @param {boolean} [options.allowDiskUse] Enables writing to temporary files on the server.
+ * @param {string|boolean} [options.explain] The verbosity mode for the explain output.
  * @param {ClientSession} [options.session] optional session to use for this operation
  * @throws {MongoError}
  * @return {Cursor}
@@ -739,6 +739,7 @@ Collection.prototype.insert = deprecate(function(docs, options, callback) {
  * @param {boolean} [options.checkKeys=false] If true, will throw if bson documents start with `$` or include a `.` in any key value
  * @param {boolean} [options.serializeFunctions=false] Serialize functions on any object.
  * @param {boolean} [options.ignoreUndefined=false] Specify if the BSON serializer should ignore undefined fields.
+ * @param {string|boolean} [options.explain] The verbosity mode for the explain output.
  * @param {ClientSession} [options.session] optional session to use for this operation
  * @param {Collection~updateWriteOpCallback} [callback] The command result callback
  * @return {Promise} returns Promise if no callback passed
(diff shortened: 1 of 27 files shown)


lib/collection.js
```diff
@@ -310,6 +309,7 @@ const DEPRECATED_FIND_OPTIONS = ['maxScan', 'fields', 'snapshot', 'oplogReplay']
  * @param {boolean} [options.noCursorTimeout] The server normally times out idle cursors after an inactivity period (10 minutes) to prevent excess memory use. Set this option to prevent that.
  * @param {object} [options.collation] Specify collation (MongoDB 3.4 or higher) settings for update operation (see 3.4 documentation for available fields).
  * @param {boolean} [options.allowDiskUse] Enables writing to temporary files on the server.
+ * @param {string|boolean} [options.explain] The verbosity mode for the explain output.
  * @param {ClientSession} [options.session] optional session to use for this operation
```
HanaPearlman, line 313: I was not sure the best way to communicate what the valid values of `options.explain` are. We could list the valid strings and how the boolean values will be interpreted, which is what we do for `readPreference`, or we could link to some server documentation, which is what we do for `hint`.

lib/operations/delete_many.js
```diff
@@ -18,8 +19,20 @@ class DeleteManyOperation extends OperationBase {
...
     options.single = false;
-    removeDocuments(coll, filter, options, (err, r) => deleteCallback(err, r, callback));
+    removeDocuments(coll, filter, options, (err, r) => {
+      if (callback == null) return;
+      if (err && callback) return callback(err);
+      if (r == null) return callback(null, { result: { ok: 1 } });
+
+      // If an explain operation was executed, don't process the server results
+      if (this.explain) return callback(undefined, r.result);
+
```
HanaPearlman, line 29: Got rid of `deleteCallback` and `updateCallback` in `common_functions` and moved the body of the functions into the respective operation classes; this way we can reference `this.explain` (and matches what happens in master).

lib/explain.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+class Explain {
+  /**
+   * Constructs an Explain from the explain verbosity.
+   *
+   * For backwards compatibility, true is interpreted as "allPlansExecution"
+   * and false as "queryPlanner". Prior to server version 3.6, aggregate()
+   * ignores the verbosity parameter and executes in "queryPlanner".
+   *
+   * @param {string|boolean} [verbosity] The verbosity mode for the explain output ({'queryPlanner'|'queryPlannerExtended'|'executionStats'|'allPlansExecution'|boolean})
```
nbbeeken, line 15:
```suggestion
   * @param {'queryPlanner'|'queryPlannerExtended'|'executionStats'|'allPlansExecution'|boolean} [verbosity] The verbosity mode for the explain output
```

lib/explain.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+/**
+ * @class
+ * @property {string} verbosity The verbosity mode for the explain output
+ */
+class Explain {
```
nbbeeken, line 7: Just a suggestion, we can bring over that enum from 4.0 and use Object.values() for example to check the string passed in for verbosity

lib/explain.js
```diff
@@ -0,0 +1,40 @@
...
+  static fromOptions(options) {
+    if (options == null || options.explain === undefined) {
+      return;
+    }
+
+    return new Explain(options.explain);
+  }
+}
+
+module.exports = Explain;
```
nbbeeken, line 40:
```suggestion
module.exports = { Explain };
```
I know we still have this form of export elsewhere in 3.6 but it makes it harder for type inference tools to pick up that Explain is a class.

lib/collection.js
nbbeeken: I did suggest lower to put the strings in, if we put the strings in VSCode can suggest the strings to devs 🤷  but then it might error if there's a new string later.. Maybe the link is best

nbbeeken (changes requested):

lib/collection.js
HanaPearlman: If we go ahead with using the enum from 4.0 to check that the string passed in is a valid verbosity, then it will error if there's a new verbosity string added later anyway. 

Since we went ahead with doing the valid verbosity check in 4.0, my thought is to go ahead and do that here as well, and list the strings within the type. More work for us later if a new string is added, though.

lib/operations/update_one.js
```diff
@@ -26,24 +28,28 @@ class UpdateOneOperation extends OperationBase {
...
+      if (err) return callback(err);
+      if (r == null) return callback(null, { result: { ok: 1 } });
+
+      // If an explain operation was executed, don't process the server results
+      if (this.explain) return callback(undefined, r.result);
+
+      r.modifiedCount = r.result.nModified != null ? r.result.nModified : r.result.n;
+      r.upsertedId =
+        Array.isArray(r.result.upserted) && r.result.upserted.length > 0
+          ? r.result.upserted[0] // FIXME(major): should be `r.result.upserted[0]._id`
```
nbbeeken, line 42: I wonder if this fixme is still relevant..

nbbeeken (approved):
Barring the current issues we're having around testing LGTM 🙂

emadum (approved):
LGTM 👍

PR #2627, opened 2020-11-19
fix: awaitable isMaster timeout must respect connectTimeoutMS

emadum:
This PR ensures that awaitable isMaster handshakes respect an unlimited timeout (`connectionTimeoutMS=0`) as per the [SDAM specification section on socket timeouts](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#socket-timeout):

> Clients MUST use connectTimeoutMS as the timeout for the connection handshake. When connectTimeoutMS=0, the timeout is unlimited and MUST remain unlimited for awaitable isMaster replies. Otherwise, connectTimeoutMS is non-zero and clients MUST use connectTimeoutMS + heartbeatFrequencyMS as the timeout for awaitable isMaster replies.

NODE-2874

### DIFF
--- a/test/functional/mongo_client_options.test.js
+++ b/test/functional/mongo_client_options.test.js
@@ -116,6 +116,10 @@ describe('MongoClient Options', function() {
     metadata: { requires: { topology: 'replicaset', mongodb: '>= 4.4' } },
     test: function(done) {
       if (!this.configuration.usingUnifiedTopology()) return done();
+      const client = this.configuration.newClient({
+        connectTimeoutMS: 0,
+        heartbeatFrequencyMS: 500
+      });
       const stub = sinon.stub(Connection.prototype, 'command').callsFake(function() {
         const args = Array.prototype.slice.call(arguments);
         const ns = args[0];
@@ -126,15 +130,10 @@ describe('MongoClient Options', function() {
           expect(options)
             .property('socketTimeout')
             .to.equal(0);
-          done();
+          client.close(done);
         }
         stub.wrappedMethod.apply(this, args);
       });
-      const client = this.configuration.newClient({
-        connectTimeoutMS: 0,
-        heartbeatFrequencyMS: 500
-      });
-      this.defer(() => client.close());
       client.connect(err => expect(err).to.not.exist);
     }
   });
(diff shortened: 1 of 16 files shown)


test/functional/mongo_client_options.test.js
```diff
@@ -110,6 +110,28 @@ describe('MongoClient Options', function() {
     });
   });
 
+  it('NODE-2874: connectTimeoutMS=0 causes monitoring to time out', function(done) {
```
mbroadst, line 113: nit: the jira ticket number will be there in the commit associated with adding this test, I think it clutters the test output a bit to include it here

test/functional/mongo_client_options.test.js
```diff
@@ -110,6 +110,28 @@ describe('MongoClient Options', function() {
...
+    const heartbeatFrequencyMS = 500;
+    const client = this.configuration.newClient({
+      connectTimeoutMS: 0, // no connect timeout
+      heartbeatFrequencyMS // fast 500ms heartbeat
+    });
+    client.connect(() => {
+      // success after 5 heartbeats
+      const success = setTimeout(() => client.close(done), heartbeatFrequencyMS * 5);
```
mbroadst, line 121: Anything involving time without mocked timers is a potentially flaky test. I think we just want to confirm that the `socketTimeout` for the monitoring connection is never 0 given certain inputs, is there another way we can test this (ideally in unit)?

mbroadst: What I'm particularly worried about here is that you call `client.close()` from two different code paths, one of them in an arbitrary `setTimeout`. There is a possibility that `close` is called twice (and thus `done` is also called twice), and when you run this in a huge suite of tests and that race is triggered, you will spend hours trying to track down which test left a timer running in the background to produce a second call to `done`. 

Regarding the "reproduction tests", I'm not sure I see significant value in always committing exactly what the user provided to reproduce the issue. The reproduction is very useful from them, but not always very clear to the maintainers why the test is there. Imagine an exaggerated example of this, where a user reports that:
```
const client = new MongoClient('mongodb://localhost/');
await client.connect();
```
doesn't work when you connect against a secondary because `connect` is implemented as `primary` server selection instead of `primaryPreferred`. The reproduction test in this case would not be very helpful to future readers. As I said, this is sort of a contrived case, but considering that our options are passed very deeply into the driver you can imagine a whole slew of tests which ultimately would provide very similar repros

My thinking is that the reproduction code lives in the ticket, the ticket is linked to by the commit, and the thing that should be committed to our code base in most cases is a test that provides coverage very explicitly around a piece of code.

emadum: That makes sense, I can see how over time having a reproduction test for every user-submitted ticket could become cumbersome; will attempt a unit test.

mbroadst (approved):
LGTM

nbbeeken (approved):

PR #2629, opened 2020-11-20
chore(ci): skip broken windows tests, idle timeout, colors

emadum:
1. this PR skips some broken tests on Windows via the `os: '!win32'` metadata directive, to support the Continuous Matrix Testing initiative, which needs to test the `3.6` branch on Windows.
2. currently, when a test run leaks resources it sits idle until reaching our 15 minute global evergreen task timeout and being killed. This PR adds a 60 second per-task idle timeout to the driver test run on evergreen, so if a test run finishes after 5 minutes but leaks resources, the task will finish due to the idle timeout in 6 minutes rather than 15 minutes. This is a general quality-of-life improvement not related to Continuous Matrix Testing.
3. adds colors to test output in Evergreen logs 🌈

### DIFF
--- a/test/functional/change_stream.test.js
+++ b/test/functional/change_stream.test.js
@@ -1475,7 +1475,7 @@ describe('Change Streams', function() {
     }
   });
 
-  it.skip('should resume piping of Change Streams when a resumable error is encountered', {
+  it('should resume piping of Change Streams when a resumable error is encountered', {
     metadata: {
       requires: {
         generators: true,
--- a/test/functional/change_stream.test.js
+++ b/test/functional/change_stream.test.js
@@ -1475,7 +1475,7 @@ describe('Change Streams', function() {
     }
   });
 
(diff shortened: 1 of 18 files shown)


test/functional/change_stream.test.js
```diff
@@ -1475,7 +1475,7 @@ describe('Change Streams', function() {
     }
   });
 
-  it('should resume piping of Change Streams when a resumable error is encountered', {
+  it.skip('should resume piping of Change Streams when a resumable error is encountered', {
```
nbbeeken, line 1478: Is this a sometimes leaky test?

nbbeeken (commented):
Small question:

test/functional/change_stream.test.js
emadum: It was leaking the `'_nodemongodbnative_resumepipe.txt'` resource at some point in my testing, but I forget exactly when/where. I'll unskip to check if it's still an issue.

PR #2633, opened 2020-11-24
chore(ci): test on windows [master]

emadum:
## Description

Run the test suite on Windows in Evergreen for the `master` branch. Port of #2634.

[patch build](https://spruce.mongodb.com/version/5fda7044c9ec4459081f2815/tasks)

NODE-1458

**What changed?**

- `.evergreen/install-dependencies.sh` was updated to use [nvm-windows](https://github.com/coreybutler/nvm-windows) instead of [nvm](https://github.com/nvm-sh/nvm) when running in Windows environments
- only the unit and functional tests are run on Windows, manual tests may take a fair amount of work to get working and should be enabled on Windows as needed in future PRs.
- skips domain socket related tests on Windows via the `os: '!win32'` metadata directive.
- adds colors to test output in Evergreen logs 🌈 
- silences some helper tasks in Evergreen to reduce log clutter 🏎️ 
- skips CI for changes to `.md` files 🏎️ 
- adds 60 second timeout to test run. 🏎️ 
currently, when a test run leaks resources it sits idle until reaching our 15 minute global evergreen task timeout and being killed. This PR adds a 60 second per-task idle timeout to the driver test run on evergreen, so if a test run finishes after 5 minutes but leaks resources, the task will finish due to the idle timeout in 6 minutes rather than 15 minutes. 


.evergreen/install-dependencies.sh
```diff
@@ -16,22 +21,91 @@ mkdir -p ${NODE_ARTIFACTS_PATH}
...
+echo "Linux/MacOS install done"
+
+fi
+
 # NOTE: registry was overridden to not use artifactory, remove the `registry` line when
 #       BUILD-6774 is resolved.
 
 # install node dependencies
-npm install --unsafe-perm
+if [[ "$SKIP_INSTALL" == "1" ]]; then
```
emadum, line 106: Note, this will be used by the Versioned Test Executor in Continuous Matrix Testing, which will be added in a separate PR.

.evergreen/generate_evergreen_tasks.js
```diff
@@ -410,21 +396,21 @@ AWS_AUTH_VERSIONS.forEach(VERSION => {
...
-
+});
 
 const BUILD_VARIANTS = [];
 
 const getTaskList = (() => {
   const memo = {};
-  return function(mongoVersion) {
-    const key = mongoVersion;
+  return function (mongoVersion, onlyFullSuite = false) {
```
mbroadst, line 405: I thought we removed the `full-suite` tag in the 3.6 PR?

.evergreen/run-tests.sh
```diff
@@ -12,6 +12,7 @@ set -o errexit  # Exit the script with error if any of the commands fail
 AUTH=${AUTH:-noauth}
 UNIFIED=${UNIFIED:-}
 MONGODB_URI=${MONGODB_URI:-}
+TEST_NPM_SCRIPT="check:test"
```
mbroadst, line 15: nit: do you actually use this variable anywhere? It doesn't look like its ever overridden.

.evergreen/install-dependencies.sh
```diff
@@ -16,22 +20,91 @@ mkdir -p ${NODE_ARTIFACTS_PATH}
...
 registry=https://registry.npmjs.org
 EOT
+fi
 
 # NOTE: registry was overridden to not use artifactory, remove the `registry` line when
 #       BUILD-6774 is resolved.
 
 # install node dependencies
-npm install --unsafe-perm
+if [[ "$SKIP_INSTALL" == "1" ]]; then
```
mbroadst, line 105:
```suggestion
if [[ "$SKIP_NPM_INSTALL" == "1" ]]; then
```

nit: this is a file called "install dependencies", I would think a variable called `SKIP_INSTALL` would "skip installing dependencies", but it really means "skip npm install"

.evergreen/generate_evergreen_tasks.js
emadum: Oops I missed porting that over, thanks!

.evergreen/run-tests.sh
emadum: I missed porting over a few changes from `3.6`, done now. Sorry about the confusion!

This variable isn't strictly necessary; it is (now) used by the NO_EXIT flag logic, but that logic could easily be moved to the bottom where the npm script is executed.

.evergreen/install-dependencies.sh
emadum: I also missed bringing the latest update here over from `3.6`, it now accepts a `SKIP_DEPS` flag which works the way you'd think.

mbroadst (approved):
LGTM

nbbeeken (approved):

PR #2637, opened 2020-11-25
fix: transition topology state before async calls

mbroadst:
When the legacy ReplSet and Server topology types are destroyed they transition their internal state after a potentially long async close process has completed. With the recently introduced infinite default socket timeout, we are hitting this race condition more consistently. This patch ensures the types transition their state immediately, reducing the chance that such race conditions can occur.

NODE-2916


emadum (approved):
LGTM! 🎉

nbbeeken (approved):

PR #2638, opened 2020-11-25
feat: add an internal `tryNext` method

mbroadst:
The default cursor iteration behavior is to block until either a cursor is dead (has a cursor id of 0), or a non-empty batch is returned from the server for a `getMore` command. `tryNext` allows users to iterate a cursor optionally returning `null` if the `getMore` returns an empty batch (likely on a tailable + awaitData cursor).

NODE-2917

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -213,7 +213,7 @@ export abstract class AbstractCursor extends EventEmitter {
         return done(undefined, true);
       }
 
-      next(this, (err, doc) => {
+      next(this, true, (err, doc) => {
         if (err) return done(err);
 
         if (doc) {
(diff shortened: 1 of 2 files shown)


src/cursor/abstract_cursor.ts
```diff
@@ -604,7 +624,11 @@ function next(cursor: AbstractCursor, callback: Callback<Document | null>): void
       return cleanupCursor(cursor, () => callback(err, nextDocument(cursor)));
     }
 
-    next(cursor, callback);
+    if (cursor[kDocuments].length === 0 && blocking === false) {
+      return callback(undefined, null);
```
nbbeeken, line 628: Is there a specific reason to use `null` over `undefined`? I suppose it makes more sense to communicate a non-existent object, so I'm not taking issue with it just wondering if there was some thinking behind it

Although maybe one reason to use undefined is so the result doesn't respond true to: `typeof result === 'object'`.

mbroadst: I used `null` here for symmetry with our semantics around a "dead" cursor. When a cursor returns a zero cursor id, we return an explicit `null` to the user. I think this might have been left over from days when a `Cursor` inherited `Readable` since ending a stream is accomplished by `readable.push(null)`, but either way the convention exists so I stuck with it

nbbeeken (approved):

emadum (approved):

PR #2639, opened 2020-11-27
chore: change pass to password in Auth interface #2921

aherlihy:
Updated the Auth interface to include 'password' instead of 'pass'. I could not find any tests that depended on this, but let me know if anything else should be changed.

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -43,7 +43,7 @@ export interface Auth {
   /** The username for auth */
   user?: string;
   /** The password for auth */
-  pass?: string;
+  password?: string;
 }
 
 /** @public */


emadum (approved):
LGTM. FYI, this will be updated [here](https://github.com/mongodb/node-mongodb-native/pull/2623/files#diff-f005e84d9066ef889099ec2bd907abf7900f76da67603e4130e1c92fac92533dL41-R51) in #2623, but we can merge this one in first if that'll be useful.

aherlihy:
If it doesn't hurt, it'd be nice for it to be merged since I'm not aware of how long the other PR will be in review and the ticket I created was already closed, even if it wasn't fixed yet.

nbbeeken (approved):

PR #2612, opened 2020-11-05
test: restrict geoSearch tests to mongodb <= 4.4

emadum:
## Description

This PR restricts `geoHaystack`/`geoSearch` tests to mongodb `<=4.4`, which was overlooked in NODE-2546.

**What changed?**

**Are there any files to ignore?**

PR #2641, opened 2020-11-29
chore(ci): xunit integration [3.5]

emadum:
Backports of #2596 and #2612 for `3.5`

Note: also fixes a failing test on the latest server to get CI passing, related to setting collation on `listCollections`.

It seems we set collation a bit too aggressively and rely on the server ignoring it where it's not supported. However, the latest server version is returning an error:

<pre><code>FAILURE: BSON field &#x27;listCollections.collation&#x27; is an unknown field. ()
 MongoError: BSON field &#x27;listCollections.collation&#x27; is an unknown field.
     at Connection.&#x3C;anonymous&#x3E; (lib/core/connection/pool.js:451:61)
     at processMessage (lib/core/connection/connection.js:384:10)
     at Socket.&#x3C;anonymous&#x3E; (lib/core/connection/connection.js:553:15)
     at addChunk (_stream_readable.js:288:12)
     at readableAddChunk (_stream_readable.js:269:11)
     at Socket.Readable.push (_stream_readable.js:224:10)
     at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)</code></pre>

### DIFF
--- a/test/functional/readconcern.test.js
+++ b/test/functional/readconcern.test.js
@@ -59,9 +59,12 @@ describe('ReadConcern', function() {
   ];
 
   tests.forEach(test => {
+    const metadata = { requires: { topology: 'replicaset', mongodb: '>= 3.2' } };
+    if (test.commandName === 'geoSearch') {
+      metadata.requires.mongodb += ' <=4.4';
+    }
     it(test.description, {
-      metadata: { requires: { topology: 'replicaset', mongodb: '>= 3.2' } },
-
+      metadata,
       test: function(done) {
         const started = [];
         const succeeded = [];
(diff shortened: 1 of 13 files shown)


nbbeeken (approved):
nice to have this on all the branches! tyty!

test/functional/readconcern.test.js
```diff
@@ -59,9 +59,12 @@ describe('ReadConcern', function() {
   ];
 
   tests.forEach(test => {
+    const metadata = { requires: { topology: 'replicaset', mongodb: '>= 3.2' } };
+    if (test.commandName === 'geoSearch') {
+      metadata.requires.mongodb += ' <=4.4';
+    }
```
mbroadst, line 65: This is a data-driven test, shouldn't the instructions for building the tests should be in the specs above rather than special cased here? Maybe something like `serverVersion: '<=4.4'`?

mbroadst (changes requested):

mbroadst (approved):
LGTM!

PR #2642, opened 2020-11-30
feat: remove top-level write concern options

HanaPearlman:
## Description

Code changes include: the `WriteConcern` type now has only one property:  `writeConcern: WriteConcern | WriteConcernSettings`, where `WriteConcernSettings` (naming up for debate) encompasses the old top-level options. `WriteConcern.fromOptions` ignores the old top-level WC options.

The change to `legalOptionNames ` in `db.ts` means the old top-level WC options will be filtered out when creating a `Db`. The change to `validOptionNames` in `operations/connect.ts` means it will either error or warn (configurable) if the old top-level WC options are passed to a `MongoClient`.  

Test changes include everything from #2624 and now all options that are not being used to create a connection string use the `writeConcern` key. 

[NODE-1722](https://jira.mongodb.org/browse/NODE-1722)


src/operations/connect.ts
```diff
@@ -561,12 +554,24 @@ function transformUrlOptions(connStrOptions: any) {
...
+    connStrOpts.wTimeoutMS = undefined;
   }
 
   if (connStrOptions.srvHost) {
     connStrOpts.srvHost = connStrOptions.srvHost;
   }
 
+  // Any write concern options from the URL will be top-level, so we manually
+  // move them options under `object.writeConcern`
+  const wcKeys = ['w', 'wtimeout', 'j', 'journal', 'fsync'];
```
emadum, line 566: I'd prefer to see this in a const exported by `write_concern.ts`, to make future maintenance easier. Perhaps we can restore the export of `writeConcernKeys` that was deleted and add the missing `journal` key in?

emadum (approved):
LGTM, mod one nit request 👍

PR #2643, opened 2020-11-30
refactor: move find command building all into the find operation

mbroadst:
One of the changes made in merging the core and native drivers was to merge the wire protocol methods for executing find operations. A vestigial piece of this was that the find command was being built in the wire protocol layer, rather than the operation layer. This patch teases these two apart, so that the `query` wire protocol method only creates and executes an `OP_QUERY` message, and the modern and legacy find commands are built in the find operation's definition itself.

NODE-2900

### DIFF
--- a/src/operations/find.ts
+++ b/src/operations/find.ts
@@ -1,16 +1,23 @@
 import { Aspect, defineAspects, Hint } from './operation';
-import { maxWireVersion, MongoDBNamespace, Callback, normalizeHintField } from '../utils';
+import {
+  maxWireVersion,
+  MongoDBNamespace,
+  Callback,
+  normalizeHintField,
+  decorateWithExplain
+} from '../utils';
 import { MongoError } from '../error';
 import type { Document } from '../bson';
 import type { Server } from '../sdam/server';
 import type { Collection } from '../collection';
 import type { CollationOptions } from '../cmap/wire_protocol/write_command';
-import type { QueryOptions } from '../cmap/wire_protocol/query';
 import { CommandOperation, CommandOperationOptions } from './command';
 import { Sort, formatSort } from '../sort';
+import { isSharded } from '../cmap/wire_protocol/shared';
+import { ReadConcern } from '../read_concern';
 
 /** @public */
-export interface FindOptions extends QueryOptions, CommandOperationOptions {
+export interface FindOptions extends CommandOperationOptions {
   /** Sets the limit of documents returned in the query. */
   limit?: number;
   /** Set to sort the documents coming back from the query. Array of indexes, `[['a', 1]]` etc. */
@@ -59,10 +66,7 @@ const SUPPORTS_WRITE_CONCERN_AND_COLLATION = 5;
 
 /** @internal */
 export class FindOperation extends CommandOperation<FindOptions, Document> {
-  cmd: Document;
   filter: Document;
-
-  hint?: Hint;
 
   constructor(
     collection: Collection | undefined,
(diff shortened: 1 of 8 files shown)


src/operations/find.ts
```diff
@@ -89,169 +92,242 @@ export class FindOperation extends CommandOperation<FindOptions, Document> {
...
+    findCommand.$comment = options.comment;
+  }
+
+  if (typeof options.maxTimeMS === 'number') {
+    findCommand.$maxTimeMS = options.maxTimeMS;
+  }
+
+  if (options.explain) {
+    findCommand.$explain = true;
+  }
```
nbbeeken, line 328: I think this is overwriting the decorateWithExplain call at the top of the function.

nbbeeken (changes requested):

src/operations/find.ts
mbroadst: Hmm, I don't think so. This is called in a different block of the `execute` method above, where `serverWireVersion < 4`. `decorateWithExplain` it happening in a `serverWireVersion > 4` world, where we use `makeFindCommand` (this is `makeLegacyFindCommand`)

nbbeeken: Oops yea the diff made it easy to miss

emadum: Sorry to resurrect, but I think either this needs to change to `findCommand.$explain = 'queryPlanner'` or we need to update [this test](https://github.com/mongodb/node-mongodb-native/blob/master/test/functional/explain.test.js#L562) to differentiate between 3.0 and 3.2+; currently it causes CI to fail.

mbroadst: tricky little JS boolean magic :) should be fixed now, PTAL

src/operations/find.ts
```diff
@@ -89,169 +92,242 @@ export class FindOperation extends CommandOperation<FindOptions, Document> {
...
+    filter
+  };
+
+  if (options.sort) {
+    findCommand.sort = formatSort(options.sort);
+  }
+
+  if (options.projection) {
+    let projection = options.projection;
+    if (projection && !Buffer.isBuffer(projection) && Array.isArray(projection)) {
```
emadum, line 179:
```suggestion
    if (projection && Array.isArray(projection)) {
```
Nit: Shouldn't `Array.isArray` imply `!Buffer.isBuffer`?

src/operations/find.ts
```diff
@@ -89,169 +92,242 @@ export class FindOperation extends CommandOperation<FindOptions, Document> {
...
+    findCommand.hint = normalizeHintField(options.hint);
+  }
+
+  if (typeof options.skip === 'number') {
+    findCommand.skip = options.skip;
+  }
+
+  if (typeof options.limit === 'number') {
+    if (options.limit < 0) {
+      findCommand.limit = Math.abs(options.limit);
```
emadum, line 201:
```suggestion
      findCommand.limit = Math.abs(options.limit);
```

Nit: using the `-` operator is probably faster than Math.abs when we know something is negative, but also probably a fairly negligible improvement. Also applies below in the `options.batchSize < 0` condition.

emadum (approved):
LGTM! A few small nits but nothing to hold up a merge

src/operations/find.ts
mbroadst: fixed

src/operations/find.ts
mbroadst: fixed. I left two `Math.abs` in there where I felt it improved readability (`Math.abs(batchSize) < Math.abs(limit)` looks better than `-batchSize < -limit` IMO)

nbbeeken (approved):

PR #2646, opened 2020-12-01
chore: add shell flag to prepare script

nbbeeken:
This should fix the script failing to run on windows.

I had added this flag in earlier work when we wanted a more complex prepare script but we abandoned that, however we should still keep this flag to fix windows use cases.

### DIFF
--- a/etc/prepare.js
+++ b/etc/prepare.js
@@ -1,9 +1,10 @@
 #! /usr/bin/env node
 var cp = require('child_process');
 var fs = require('fs');
+var os = require('os');
 
 if (fs.existsSync('src')) {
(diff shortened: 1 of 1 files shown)


emadum (approved):
This does in fact fix it, I also made this change in #2633 but happy to merge it in separately!

nbbeeken:
> One question, is there any harm to passing `{shell: true}` on platforms other than Windows?

No harm, but someone astutely made the point that it adds expense where none is needed, I was inclined to agree. I'll merge this now just to unblock those using the git link dep style, we can simplify it back to `shell: true` in your PR if we wish. :) ty for speedy ✅

PR #2647, opened 2020-12-01
feat: reintroduce clone and rewind for cursors

mbroadst:
## Description
These two commits add back support for cloning and rewinding cursors.

**What changed?**
The changes should be fairly straightforward, but I wanted to call out a few design decisions:
~- I had originally intended `clone` to be an abstract method on `AbstractCursor`, forcing us to implement it for all possible cursor subclasses. This would have unfortunately required us to make `AbstractCursor` generic over the subclass types, since the following signature results in type ambiguity at the point of implementation in the subclass: `abstract clone(): this`~

- Therefore, `clone` is implemented by hand in `AbstractCursor` subclasses. I only implemented it for `FindCursor` and `AggregateCursor` because it's likely that's where legacy users have actually used this feature.

- There is a correction in the `rewind` implementation regarding sessions. When a session is started as part of cursor initialization I had originally marked this as an `explicit: true` start, however, this is clearly an implicit session the driver is creating in the absence of an explicitly provided one. 

none

### DIFF
--- a/src/cursor/find_cursor.ts
+++ b/src/cursor/find_cursor.ts
@@ -49,6 +49,13 @@ export class FindCursor extends AbstractCursor {
     }
   }
 
+  clone(): FindCursor {
+    return new FindCursor(this.topology, this.namespace, this[kFilter], {
+      ...this[kBuiltOptions],
+      ...this.cursorOptions
+    });
+  }
+
   /** @internal */
   _initialize(session: ClientSession | undefined, callback: Callback<ExecutionResult>): void {
     const findOperation = new FindOperation(undefined, this.namespace, this[kFilter], {
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -425,6 +425,12 @@ export class ChangeStreamCursor extends AbstractCursor {
     }
   }
 
+  clone(): ChangeStreamCursor {
+    return new ChangeStreamCursor(this.topology, this.namespace, this.pipeline, {
+      ...this.cursorOptions
+    });
+  }
+
   _initialize(session: ClientSession, callback: Callback<ExecutionResult>): void {
     const aggregateOperation = new AggregateOperation(
       { s: { namespace: this.namespace } },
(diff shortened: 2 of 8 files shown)


src/cursor/find_cursor.ts
```diff
@@ -49,6 +49,16 @@ export class FindCursor extends AbstractCursor {
     }
   }
 
+  /**
+   * Returns a new unitialized copy of this cursor, with options matching those that have been set on the current instance
```
emadum, line 53:
```suggestion
   * Returns a new uninitialized copy of this cursor, with options matching those that have been set on the current instance
```

emadum (approved):
LGTM, mod one typo

src/cursor/find_cursor.ts
mbroadst: fixed here and in aggregation_cursor too!

src/cursor/find_cursor.ts
```diff
@@ -49,6 +49,16 @@ export class FindCursor extends AbstractCursor {
...
 
+  /**
+   * Returns a new uninitialized copy of this cursor, with options matching those that have been set on the current instance
+   */
+  clone(): FindCursor {
+    return new FindCursor(this.topology, this.namespace, this[kFilter], {
+      ...this[kBuiltOptions],
+      ...this.cursorOptions
+    });
+  }
```
nbbeeken, line 60: I think you can still make this an abstract contract, the method doesn't return `this` it returns a new instance so the AbstractCursor would have:
```ts
abstract clone(): AbstractCursor;
```
and FindCursor etc. can have the more specific subclass return type.

nbbeeken (approved):
LGTM one suggestion
If we do want the abstract clone I think it's still available to us:

src/cursor/find_cursor.ts
mbroadst: duh, I should have thought of that :) implemented

PR #2649, opened 2020-12-01
fix: add all accessor tags to exported symbols

nbbeeken:
The mongodb.d.ts now only includes the public API.

NODE-2783

API-extractor enforces that things that are public can only depend upon types and symbols that are also public unless stated otherwise. So for example our ChangeStreams API hides the underlying cursor it uses from the public API so we can make both the ChangeStreamsCursor and its Options type internal which deletes them from the final type definition, implicitly (but not literally) making them inaccessible to type checking users. 
Some symbols needed to be made public since they are exposed to users. For example `BulkWriteResult` has a result property that had an internal `BulkResult` type but since the public definitions file would have that removed you would get a compilation error when it tries to type the public `result` property as the deleted `BulkResult` type.


mbroadst (approved):

tsdoc.json
```diff
@@ -1,5 +1,6 @@
 {
     "$schema": "https://developer.microsoft.com/json-schemas/tsdoc/v0/tsdoc.schema.json",
+    // Correct url: https://developer.microsoft.com/en-us/json-schemas/tsdoc/v0/tsdoc.schema.json
```
emadum, line 3: Not sure why this comment is here, but JSON doesn't support comments

emadum (changes requested):
There were a lot of changes from internal => public or from public => internal in this PR that I don't fully understand. For example, AbstractCursor became public but ChangeStreamCursorOptions became internal.

Could you update the description to explain the general reasoning behind why these had to change?

tsdoc.json
nbbeeken: I put that there because the url doesn't work and you can't change it: https://github.com/microsoft/tsdoc/issues/253
I removed, the tsdoc used jsonc so the comments didn't break anything but looks like the issue might be solved soon.

(Also updated the description)

package.json
```diff
@@ -91,11 +91,12 @@
...
     "build:evergreen": "node .evergreen/generate_evergreen_tasks.js",
     "build:ts": "rimraf lib && tsc",
-    "build:dts": "npm run build:ts && api-extractor run --typescript-compiler-folder node_modules/typescript --local && rimraf 'lib/**/*.d.ts*'",
+    "build:dts": "npm run build:ts && api-extractor run && rimraf 'lib/**/*.d.ts*'",
     "build:docs": "npm run build:dts && typedoc",
     "check:bench": "node test/benchmarks/driverBench",
     "check:coverage": "nyc npm run check:test",
-    "check:lint": "npm run check:ts && eslint -v && eslint --max-warnings=0 --ext '.js,.ts' src test",
+    "check:lint": "npm run build:dts && npm run check:eslint",
+    "check:eslint": "eslint -v && eslint --max-warnings=0 --ext '.js,.ts' src test",
```
nbbeeken, line 99: @emadum @mbroadst 
I figure we should add api-extractor to our linting now if we want to maintain access modifiers going forward, does this seem ok to y'all? If we think its overkill for now, I can roll this back

emadum (approved):

PR #2653, opened 2020-12-02
fix: only force server id generation if requested

mbroadst:
An inverted boolean expression accidentally always forced the driver to force the server to generate document ids.

NODE-2923

### DIFF
--- a/test/functional/shared.js
+++ b/test/functional/shared.js
@@ -276,6 +276,23 @@ class APMEventCollector {
   }
 }
 
+// simplified `withClient` helper that only uses callbacks
+function withClientV2(callback) {
+  return function testFunction(done) {
+    const client = this.configuration.newClient({ monitorCommands: true });
+    client.connect(err => {
+      if (err) return done(err);
+      this.defer(() => client.close());
+
+      try {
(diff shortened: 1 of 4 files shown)


emadum (approved):

test/functional/shared.js
```diff
@@ -276,6 +276,23 @@ class APMEventCollector {
   }
 }
 
+// simplified `withClient` helper that only uses callbacks
+function withClientV2(callback) {
```
nbbeeken, line 280: This will be great to use going forward, why not add a `@deprecation` warning to the old `withClient`, it'll get the ~strikethrough~ in VSCode to help remind the next dev that this exists 😃

nbbeeken (approved):

test/functional/shared.js
mbroadst: tbh, I'm not sure why the existing `withClient` didn't work, so there's likely more work that could go into merging them into a common method. This method doesn't support promises at all, for instance, so that's probably undesirable :man_shrugging:

nbbeeken: Ah fair, fair 👍

PR #2659, opened 2020-12-03
NODE-2815: alternative no generic parameter for options

mbroadst:
## Description
This is the most basic work required to remove the generic options type for the `OperationBase` type. The general rule of thumb was to add an `options` property to each subclass with a concrete type specific to that operation. Other work includes:
 - renamed `OperationBase` to `AbstractOperation` in keeping with the convention introduced in the `AbstractCursor` work
 - minor fixes uncovered when making options stricter (things like `readPreference` being stored on a copy of options in the base class)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -224,7 +224,7 @@ export type {
   MapReduceOptions,
   FinalizeFunction
 } from './operations/map_reduce';
-export type { Hint, OperationOptions, OperationBase } from './operations/operation';
+export type { Hint, OperationOptions, AbstractOperation } from './operations/operation';
 export type { ProfilingLevelOptions } from './operations/profiling_level';
 export type { RemoveUserOptions } from './operations/remove_user';
 export type { RenameOptions } from './operations/rename';
--- a/src/operations/aggregate.ts
+++ b/src/operations/aggregate.ts
@@ -7,6 +7,7 @@ import type { Callback } from '../utils';
 import type { Document } from '../bson';
 import type { Server } from '../sdam/server';
 import type { CollationOptions } from '../cmap/wire_protocol/write_command';
+import type { ClientSession } from '../sessions';
 
 /** @internal */
 export const DB_AGGREGATE_COLLECTION = 1 as const;
@@ -34,7 +35,8 @@ export interface AggregateOptions extends CommandOperationOptions {
 }
 
 /** @internal */
-export class AggregateOperation<T = Document> extends CommandOperation<AggregateOptions, T> {
+export class AggregateOperation<T = Document> extends CommandOperation<T> {
+  options: AggregateOptions;
   target: string | typeof DB_AGGREGATE_COLLECTION;
   pipeline: Document[];
   hasWriteStage: boolean;
@@ -42,6 +44,7 @@ export class AggregateOperation<T = Document> extends CommandOperation<Aggregate
   constructor(parent: OperationParent, pipeline: Document[], options?: AggregateOptions) {
     super(parent, options);
 
+    this.options = options ?? {};
     this.target =
       parent.s.namespace && parent.s.namespace.collection
         ? parent.s.namespace.collection
@@ -82,7 +85,7 @@ export class AggregateOperation<T = Document> extends CommandOperation<Aggregate
     this.pipeline.push(stage);
   }
 
-  execute(server: Server, callback: Callback<T>): void {
+  execute(server: Server, session: ClientSession, callback: Callback<T>): void {
     const options: AggregateOptions = this.options;
     const serverWireVersion = maxWireVersion(server);
     const command: Document = { aggregate: this.target, pipeline: this.pipeline };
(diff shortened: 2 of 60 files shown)


src/operations/aggregate.ts
```diff
@@ -34,14 +34,16 @@ export interface AggregateOptions extends CommandOperationOptions {
...
+export class AggregateOperation<T = Document> extends CommandOperation<T> {
+  options: AggregateOptions;
   target: string | typeof DB_AGGREGATE_COLLECTION;
   pipeline: Document[];
   hasWriteStage: boolean;
 
   constructor(parent: OperationParent, pipeline: Document[], options?: AggregateOptions) {
     super(parent, options);
 
+    this.options = options || {};
```
nbbeeken, line 46: Very nit: I'll refrain from picking them all out but is this better suited for `??`

src/operations/operation.ts
```diff
@@ -49,21 +47,22 @@ export abstract class OperationBase<
   // BSON serialization options
   bsonOptions?: BSONSerializeOptions;
 
-  constructor(options: T = {} as T) {
-    this.options = Object.assign({}, options);
+  [kSession]: ClientSession;
```
nbbeeken, line 50: I think this should be `[kSession]?: ClientSession` I guess TS has some issues around symbol properties that it's not enforcing this.

src/operations/operation.ts
```diff
@@ -49,21 +47,22 @@ export abstract class OperationBase<
...
 
+  constructor(options: OperationOptions = {}) {
     this.readPreference = this.hasAspect(Aspect.WRITE_OPERATION)
       ? ReadPreference.primary
       : ReadPreference.fromOptions(options) ?? ReadPreference.primary;
-    // TODO: A lot of our code depends on having the read preference in the options. This should
-    //       go away, but also requires massive test rewrites.
```
nbbeeken, line 59: I guess this didn't end up breaking that many tests, I'm seeing no more than 30 at the time of writing, does that seem correct?

src/index.ts
```diff
@@ -222,7 +222,7 @@ export type {
   MapReduceOptions,
   FinalizeFunction
 } from './operations/map_reduce';
-export type { Hint, OperationOptions, OperationBase } from './operations/operation';
+export type { Hint, OperationOptions, AbstractOperation as OperationBase } from './operations/operation';
```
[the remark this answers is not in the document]
mbroadst: fixed!

src/operations/aggregate.ts
mbroadst: I changed this everywhere in the whole codebase, hence the dramatic increase in file count :)

src/operations/operation.ts
mbroadst: This piece of code was particularly nasty, and lead to a lot of unexpected failures. Ultimately, removing it makes the code easier to reason about, but it required moving code from `session.ts` to `executeOperation`

src/sessions.ts
```diff
@@ -795,12 +795,12 @@ function applySession(
...
-      return new MongoError(
-        `Read preference in a transaction must be primary, not: ${readPreference.mode}`
-      );
-    }
+    // const readPreference = options.readPreference as ReadPreference;
+    // if (readPreference && !readPreference.equals(ReadPreference.primary)) {
+    //   return new MongoError(
+    //     `Read preference in a transaction must be primary, not: ${readPreference.mode}`
+    //   );
+    // }
```
nbbeeken, line 803: We don't need this anymore since the readPref is determined in the AbstractOperation constructor right?

test/functional/sessions.test.js
```diff
@@ -113,8 +113,10 @@ describe('Sessions', function () {
 
           return client
             .withSession(testCase.operation(client))
-            .catch(() => expect(client.topology.s.sessionPool.sessions).to.have.length(1))
-            .then(() => expect(client.topology.s.sessionPool.sessions).to.have.length(1))
+            .then(
+              () => expect(client.topology.s.sessionPool.sessions).to.have.length(1),
+              () => expect(client.topology.s.sessionPool.sessions).to.have.length(1)
+            )
```
nbbeeken, line 119: Is this change implying something or just seems fit b/c regardless of outcome the sessions should still be length 1?

nbbeeken (commented):
two Qs, thanks for combing through the `||` => `??`!

emadum (approved):
LGTM 🚀

test/functional/sessions.test.js
mbroadst: it's just ensuring we always check this case, no major logical difference from the code before except that this ensures we don't make this check twice in the case of an assertion failure

src/sessions.ts
mbroadst: yep, this was cruft, removed now

nbbeeken (approved):
LGTM

PR #2660, opened 2020-12-03
fix: docs should contain at least one document

DevXiaolan:
## Description

**What changed?**

### DIFF
--- a/lib/operations/insert_many.js
+++ b/lib/operations/insert_many.js
@@ -24,6 +24,12 @@ class InsertManyOperation extends OperationBase {
       );
     }
 
+    if (docs.length === 0) {
+      return callback(
+        MongoError.create({ message: 'docs should contain at least one document', driver: true })
+      );
+    }
+
     // If keep going set unordered
     options['serializeFunctions'] = options['serializeFunctions'] || coll.s.serializeFunctions;
 


nbbeeken:
Hi @DevXiaolan thanks for submitting a contribution, however this operation already generates an error here:
[lib/bulk/common.js:1107](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/bulk/common.js#L1107-L1111)

DevXiaolan:
> Hi @DevXiaolan thanks for submitting a contribution, however this operation already generates an error here:
> [lib/bulk/common.js:1107](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/bulk/common.js#L1107-L1111)

In some cases, we prefer "do nothing" instead of throw an Error

```
// 
db.collection('any_table').insertMany(arrLikeData);

```
Anyway, If it's tolerable "try-catch" always, ignore this PR

nbbeeken:
Since there's no callback specified in `insertMany` the driver returns a promise from this call:
`db.collection('any_table').insertMany(arrLikeData);`
So the change proposed would still throw an error since the error argument in the callback exists just in a different form. If it's possible you could work around the error being thrown by using a callback and ignore the error argument as needed.
```js
db.collection('any_table').insertMany(arrLikeData, (err, res) => { if(err) {/* check if ignorable */} });
// Or of course, using try-catch as needed
```
But I'd recommend insead doing a check on the input data, (if there is a length): 
```js 
if (arrLikeData.length !== 0) db.collection('any_table').insertMany(arrLikeData);
```

PR #2662, opened 2020-12-04
NODE-2825: Add tests for azure and GCP CSFLE

mbroadst:
This is the testing component for the fixes which were merged and published in the libmongocrypt bindings. The patch adds new automated YAML tests, as well as required changes to the prose tests.


emadum (commented):
This LGTM, but oddly it seems from CI that a replica set test is now leaking a client.

nbbeeken (approved):
LGTM 🚀

PR #2664, opened 2020-12-07
fix: internal access modifier on symbol properties

nbbeeken:
The process of rolling up the definitions into one file and removing the internal APIs left some typescript errors in the final definitions file, these changes fix the errors. In addition this adds a step to the `check:lint` script to ensure errors don't exist in the produced definitions file.


nbbeeken:
My original solution marked the properties on OperationParent as internal, freeing the implementers to do the same. However, I thought it better that the public classes don't implement it at all, there will still be the type checking we want, when you pass a Db for example to an operation that accepts OperationParent it will fail if Db changes to a nonconformant shape.

mbroadst (approved):
LGTM

emadum (approved):

PR #2665, opened 2020-12-07
refactor: all write commands now inherit from CommandOperation

mbroadst:
This patch moves the command construction for write operations from the wire protocol layer to the operation layer. Specifically, the insert/update/delete operations, which are also shared by the bulk implementation, are now the primary sites for write command construction. This has a few implications, primarily that you can no longer "write" with a `Server` instance, you can only execute a write operation against it.

NODE-2953


src/operations/update.ts
```diff
@@ -41,107 +46,260 @@ export interface UpdateResult {
...
+    session: ClientSession,
+    callback: Callback<UpdateResult | Document>
+  ): void {
+    super.execute(server, session, (err, res) => {
+      if (err || !res) return callback(err);
+      if (typeof this.explain !== 'undefined') return callback(undefined, res);
+
+      // TODO: copied from `updateDocuments`, not sure if these are needed anymore
+      if (res.code) return callback(new MongoError(res));
+      if (res.writeErrors) return callback(new MongoError(res.writeErrors[0]));
```
nbbeeken, line 140: TODO here or should we have a ticket?

src/operations/update.ts
```diff
@@ -41,107 +46,260 @@ export interface UpdateResult {
...
+    if (typeof options.bypassDocumentValidation === 'boolean') {
+      command.bypassDocumentValidation = options.bypassDocumentValidation;
+    }
+
+    // If a command is to be explained, we need to reformat the command after
+    // the other command properties are specified.
+    // const explain = Explain.fromOptions(options);
+    // if (explain) {
+    //   command = decorateWithExplain(writeCommand, explain);
+    // }
```
nbbeeken, line 91: Updates can be explained so was there an issue here?

src/operations/update.ts
```diff
@@ -41,107 +46,260 @@ export interface UpdateResult {
...
+  bypassDocumentValidation?: boolean;
+  /** Specifies a collation */
+  collation?: CollationOptions;
+  /** Specify that the update query should only consider plans using the hinted index */
+  hint?: string | Document;
+  /** When true, creates a new document if no document matches the query */
+  upsert?: boolean;
+
+  // FIXME:
+  multi?: boolean;
```
nbbeeken, line 205: I think the fix here would be to remove multi from the interface and instead have like `ReplaceOptions & {multi: boolean}` since it's like an internal thing, does that sound right?
We don't have to do that here but lets leave that in the fixme comment for the next dev

src/operations/update.ts
mbroadst: the command is actually decorated with explain in the `CommandOperation#executeCommand` method

src/operations/update.ts
mbroadst: oops, this was just cruft. I actually did the `& { multi: boolean }` trick already in `makeUpdateOperation` below!

src/operations/update.ts
mbroadst: I removed the `TODO`, this was actually needed. Someone may consider deduplicating the code in the future, but I think its fine for now

emadum (approved):
LGTM 👍

test/functional/aggregation.test.js
```diff
@@ -54,6 +54,8 @@ describe('Aggregation', function () {
         var collection = db.collection('shouldCorrectlyExecuteSimpleAggregationPipelineUsingArray');
         // Insert the docs
         collection.insertMany(docs, { w: 1 }, function (err, result) {
+          if (err) console.dir(err);
```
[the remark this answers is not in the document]
mbroadst: :see_no_evil: yeah it was, I'll correct that in the following PR just so we don't pay another CI round for this one

nbbeeken (approved):

PR #2666, opened 2020-12-07
refactor!: CreateIndexOp returns string, CreateIndexesOp returns array

emadum:
`CreateIndexesOperation` was returning the name of the created index
instead of the result from the server because of the logic ensuring
`CreateIndexOperation` has that behavior.

This PR makes `CreateIndexesOperation` return an *array of index names*
(breaking change), while `CreateIndexOperation` continues to return a
single index name string. This makes the behavior consistent with other
official drivers.

Users in need of the raw result from the server should use `db.command`.

NODE-2602

### DIFF
--- a/src/operations/indexes.ts
+++ b/src/operations/indexes.ts
@@ -150,7 +150,9 @@ export class IndexesOperation extends AbstractOperation<Document> {
 }
 
 /** @internal */
-export class CreateIndexesOperation<T = string[]> extends CommandOperation<T> {
+export class CreateIndexesOperation<
+  T extends string | string[] = string[]
+> extends CommandOperation<T> {
   options: CreateIndexesOptions;
   collectionName: string;
   indexes: IndexDescription[];
@@ -169,7 +171,7 @@ export class CreateIndexesOperation<T = string[]> extends CommandOperation<T> {
     this.indexes = indexes;
   }
 
-  execute(server: Server, session: ClientSession, callback: Callback<any>): void {
+  execute(server: Server, session: ClientSession, callback: Callback<T>): void {
     const options = this.options;
     const indexes = this.indexes;
 
@@ -221,8 +223,8 @@ export class CreateIndexesOperation<T = string[]> extends CommandOperation<T> {
         return;
       }
 
-      const indexNames = indexes.map(index => index.name);
-      callback(undefined, indexNames);
+      const indexNames = indexes.map(index => index.name || '');
+      callback(undefined, indexNames as T);
     });
   }
 }
@@ -245,7 +247,7 @@ export class CreateIndexOperation extends CreateIndexesOperation<string> {
   }
   execute(server: Server, session: ClientSession, callback: Callback<string>): void {
     super.execute(server, session, (err, indexNames) => {
-      if (err) return callback(err);
+      if (err || !indexNames) return callback(err);
       return callback(undefined, indexNames[0]);
     });
   }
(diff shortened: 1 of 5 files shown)


nbbeeken (approved):
LGTM

src/operations/indexes.ts
```diff
@@ -153,24 +153,23 @@ export class IndexesOperation extends AbstractOperation<Document> {
...
     collectionName: string,
     indexes: IndexDescription[],
-    options?: CreateIndexesOptions
+    options?: CreateIndexesOptions,
+    singular?: boolean
   ) {
     super(parent, options);
 
     this.options = options ?? {};
+    this.singular = singular ?? false;
```
mbroadst, line 169: Do you even need to track this anymore? It seems like the way to do this is to have this operation always return `string[]`, and then have `CreateIndexOperation` do:
```
 execute(server: Server, session: ClientSession, callback: Callback<string>): void {
   super.execute(server, session, (err, indexNames) => {
     callback(undefined, indexNames[0]);
   });
 }
```
Alternatively you could have `CreateIndexesOperation` return `string | string[]` and get rid of `CreateIndexOperation` altogether.

src/operations/indexes.ts
```diff
@@ -153,24 +153,23 @@ export class IndexesOperation extends AbstractOperation<Document> {
 export class CreateIndexesOperation extends CommandOperation<Document> {
```
mbroadst, line 153: This is `extends CommandOperation<string[]>` now right? Have the methods been changed to reflect this in `collection.ts` as well?

mbroadst (changes requested):

src/operations/indexes.ts
emadum: updated, PTAL 🙏

src/operations/indexes.ts
emadum: updated!

src/operations/indexes.ts
```diff
@@ -168,12 +167,9 @@ export class CreateIndexesOperation extends CommandOperation<Document> {
     this.collectionName = collectionName;
 
     this.indexes = indexes;
-    if (indexes.length === 1) {
-      this.onlyReturnNameOfCreatedIndex = true;
-    }
   }
 
-  execute(server: Server, session: ClientSession, callback: Callback<Document>): void {
+  execute(server: Server, session: ClientSession, callback: Callback<any>): void {
```
mbroadst, line 172: This is cheating :) But I think you can just use `Callback<T>` here since you default to `string[]`

mbroadst (changes requested):

src/operations/indexes.ts
emadum: I spent a while on this and couldn't figure out how to get around using `any`, with `T` there it gives this error:

```
Argument of type '(string | undefined)[]' is not assignable to parameter of type 'T'.
  '(string | undefined)[]' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint '{}'.ts(2345)
```

emadum: I spent a while on this and couldn't figure out how to get around using `any`, with `T` there it gives this error:

```
Argument of type 'string[]' is not assignable to parameter of type 'T'.
  'string[]' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint '{}'.ts(2345)
```

emadum: fixed!

mbroadst (approved):
LGTM

PR #2668, opened 2020-12-08
refactor: mount all wire protocol primitives on Connection

mbroadst:
We no longer use the `makeServerTrampoline` trick to pass a connection into the wire protocol primitives. Instead, the primitives are now mounted directly onto a `Connection`. An additional refactor is included which specializes `Connection` for auto encryption in the form of a `CryptoConnection`.

NODE-2133

~**NOTE:** this builds upon #2665 so the changes will look rather large until that is merged, and this is rebased upon it.~


src/cmap/connection.ts
```diff
@@ -249,52 +300,335 @@ export class Connection extends EventEmitter {
...
+            !cmd.commitTransaction &&
+            err &&
+            err instanceof MongoError &&
+            err.hasErrorLabel('TransientTransactionError')
+          ) {
+            session.transaction.unpinServer();
+          }
+
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          return callback!(err, ...args);
```
nbbeeken, line 387: Might not need this anymore the callback is nonnullish in this function looks like

src/utils.ts
```diff
@@ -560,6 +560,11 @@ export function deprecateOptions(
   return deprecated;
 }
 
+/** @internal */
+export function ns(ns: string): MongoDBNamespace {
+  return MongoDBNamespace.fromString(ns);
+}
```
nbbeeken, line 566: 👏 👏 👏  "Use namespace through and through" TODOs beware!
Did this fix all of those?

mbroadst: Hm, I've been trying to make that change as much as possible, but I don't think we can assume everything is typed as such. Someone should manually check that a string is never being used

src/cmap/connection.ts
mbroadst: fixed

mbroadst:
@emadum @nbbeeken please ignore the travis failure, I accidentally reenabled it when we were looking for the coveralls token the other day. It's been disabled again, but the result here can't be deleted

nbbeeken (approved):

emadum (approved):
LGTM 👍

PR #2669, opened 2020-12-08
feat: introduce BufferPool to replace BufferList

mbroadst:
BufferList really helped simplify a lot of code in our message stream processing, but ultimately is more powerful than we need it to be. Additionally, depending on this package makes maintenance of the driver more difficult over time. This introduces a new type called BufferList which is tailored to our particular use case.

NODE-2930

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1159,7 +1159,10 @@ export function isRecord(
 const kBuffers = Symbol('buffers');
 const kLength = Symbol('length');
 
-/** @internal */
+/**
+ * A pool of Buffers which allow you to read them as if they were one
+ * @internal
+ */
 export class BufferPool {
   [kBuffers]: Buffer[];
   [kLength]: number;
@@ -1173,22 +1176,21 @@ export class BufferPool {
     return this[kLength];
   }
 
+  /** Adds a buffer to the internal buffer pool list */
   append(buffer: Buffer): void {
     this[kBuffers].push(buffer);
     this[kLength] += buffer.length;
   }
 
+  /** Returns the requested number of bytes without consuming them */
   peek(size: number): Buffer {
     return this.read(size, false);
   }
 
+  /** Reads the requested number of bytes, optionally consuming them */
   read(size: number, consume = true): Buffer {
     if (typeof size !== 'number' || size < 0) {
       throw new TypeError('Parameter size must be a non-negative number');
-    }
-
-    if (typeof size === 'number' && size < 0) {
-      size += this.length;
     }
 
     if (size > this[kLength]) {
@@ -1243,10 +1245,6 @@ export class BufferPool {
       // compact the internal buffer array
       if (consume) {
         this[kBuffers] = this[kBuffers].slice(idx);
-      }
-
-      if (result.length > offset) {
-        return result.slice(0, offset);
       }
     }
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -961,7 +961,7 @@ export function calculateDurationInMs(started: number): number {
   return elapsed < 0 ? 0 : elapsed;
 }
 
-export interface InterruptableAsyncIntervalOptions {
+export interface InterruptibleAsyncIntervalOptions {
   /** The interval to execute a method on */
   interval: number;
   /** A minimum interval that must elapse before the method is called */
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -1155,3 +1155,101 @@ export function isRecord(
...
+      }
+
+      // compact the internal buffer array
+      if (consume) {
+        this[kBuffers] = this[kBuffers].slice(idx);
+      }
+
+      if (result.length > offset) {
+        return result.slice(0, offset);
+      }
```
nbbeeken, line 1250: How could this occur? If offset is the sum of all bytes copied into result then shouldn't it always be equal?

src/utils.ts
```diff
@@ -1155,3 +1155,101 @@ export function isRecord(
...
+      let bytesToCopy = size;
+      for (idx = 0; idx < this[kBuffers].length; ++idx) {
+        let bytesCopied;
+        if (bytesToCopy > this[kBuffers][idx].length) {
+          bytesCopied = this[kBuffers][idx].copy(result, offset, 0);
+          offset += bytesCopied;
+        } else {
+          bytesCopied = this[kBuffers][idx].copy(result, offset, 0, bytesToCopy);
+          if (consume) {
+            this[kBuffers][idx] = this[kBuffers][idx].slice(0, bytesCopied);
```
nbbeeken, line 1234: Such a minor nit, disregard if you feel like its not worth the time, in the spirit of isomorphic JS: `#subarray()` and `#set()` methods are available on Uint8Array and provide the same functionality as `#slice()` and `#copy()` respectively.

nbbeeken (changes requested):

src/utils.ts
```diff
@@ -991,10 +991,10 @@ export interface InterruptableAsyncInterval {
  *
  * @param fn - An async function to run on an interval, must accept a `callback` as its only parameter
  */
-export function makeInterruptableAsyncInterval(
+export function makeInterruptibleAsyncInterval(
```
emadum, line 994: This import needs to be updated in `test/unit/utils.test.js` to fix CI

emadum (changes requested):

src/utils.ts
mbroadst: oops thanks, fixed

src/utils.ts
```diff
@@ -1155,3 +1155,101 @@ export function isRecord(
...
+  }
+
+  read(size: number, consume = true): Buffer {
+    if (typeof size !== 'number' || size < 0) {
+      throw new TypeError('Parameter size must be a non-negative number');
+    }
+
+    if (typeof size === 'number' && size < 0) {
+      size += this.length;
+    }
```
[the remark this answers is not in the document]
mbroadst: yeah, this was left in by accident before the above checks were added. Removed

src/utils.ts
mbroadst: I think maybe if we weren't using `Buffer.alloc`/`Buffer.concat` this might make sense. There are alternatives using `UInt8Array` for those operations as well, but the reality here is that we're making a type which is optimized specifically for the Node.js `Buffer`, so I'm inclined to stick to that api

src/utils.ts
mbroadst: It was a defensive move just because `allocUnsafe` is being called above, but I agree I can't see how it might happen. Removed

src/utils.ts
```diff
@@ -1155,3 +1155,101 @@ export function isRecord(
...
+  append(buffer: Buffer): void {
+    this[kBuffers].push(buffer);
+    this[kLength] += buffer.length;
+  }
+
+  peek(size: number): Buffer {
+    return this.read(size, false);
+  }
+
+  read(size: number, consume = true): Buffer {
```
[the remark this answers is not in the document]
mbroadst: added some docs to the class in general

test/unit/utils.test.js
```diff
@@ -161,4 +161,84 @@ describe('utils', function () {
...
+        buffer.append(Buffer.from([0, 1, 2, 3]));
+        const data = buffer.peek(2);
+        expect(data).to.eql(Buffer.from([0, 1]));
+        expect(buffer).property('length').to.equal(4);
+      });
+
+      it('across multiple buffers', function () {
+        const buffer = new BufferPool();
+        buffer.append(Buffer.from([0, 1, 2, 3, 4, 5]));
+        const data = buffer.peek(6);
```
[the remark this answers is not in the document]
mbroadst: Nice catch, not sure what happened here must have been some bad copy pasta. Fixed

nbbeeken (approved):

emadum (approved):

PR #2671, opened 2020-12-09
fix: honor ignoreUndefined on findAndModify commands

emadum:
Honors the `ignoreUndefined` option for find-and-modify commands.

Note: this option doesn't seem to have any effect for `findOneAndDelete`
operations, but support was added because it is listed as a valid option.

NODE-2945


mbroadst (approved):
LGTM! Not an issue in `master` right?

emadum:
> LGTM! Not an issue in `master` right?

Nope it's fixed by Hana's work on BSON options, though adding the test on master as well couldn't hurt.

nbbeeken (approved):

PR #2675, opened 2020-12-11
fix: make tryNext and Batch public

emadum:
Exposes the `tryNext` iteration helper on cursors and change streams.

NODE-2952

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "check:test": "mocha --recursive test/functional test/unit",
     "check:ts": "tsc -v && tsc --noEmit",
     "check:atlas": "mocha --config \"test/manual/mocharc.json\" test/manual/atlas_connectivity.test.js",
-    "check:adl": "mocha --exit test/manual/data_lake.test.js",
+    "check:adl": "mocha test/manual/data_lake.test.js",
     "check:ocsp": "mocha --config \"test/manual/mocharc.json\" test/manual/ocsp_support.test.js",
     "check:kerberos": "mocha --config \"test/manual/mocharc.json\" test/manual/kerberos.test.js",
     "check:tls": "mocha --config \"test/manual/mocharc.json\" test/manual/tls_support.test.js",
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "check:test": "mocha --recursive test/functional test/unit",
     "check:ts": "tsc -v && tsc --noEmit",
     "check:atlas": "mocha --config \"test/manual/mocharc.json\" test/manual/atlas_connectivity.test.js",
-    "check:adl": "mocha test/manual/data_lake.test.js",
+    "check:adl": "mocha --exit test/manual/data_lake.test.js",
     "check:ocsp": "mocha --config \"test/manual/mocharc.json\" test/manual/ocsp_support.test.js",
     "check:kerberos": "mocha --config \"test/manual/mocharc.json\" test/manual/kerberos.test.js",
     "check:tls": "mocha --config \"test/manual/mocharc.json\" test/manual/tls_support.test.js",
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -335,6 +335,13 @@ export class ChangeStream extends EventEmitter {
     }
     return this.cursor.stream(options);
   }
+
+  tryNext(): Promise<Document | null> {
+    if (!this.cursor) {
+      throw new MongoError('ChangeStream has no cursor, unable to tryNext');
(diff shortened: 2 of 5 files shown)


src/change_stream.ts
```diff
@@ -335,6 +335,13 @@ export class ChangeStream extends EventEmitter {
     }
     return this.cursor.stream(options);
   }
+
+  tryNext(): Promise<Document | null> {
+    if (!this.cursor) {
```
mbroadst, line 340: this needs to be implemented like `next` and `hasNext`:
```
  getCursor(this, (err, cursor) => {
    if (err || !cursor) return cb(err); // failed to resume, raise an error
    cursor.tryNext(cb);
  });
});
```

mbroadst: Also, if you are going to make this public now, please document the method, Maybe there's some configuration in tslint/eslint/tsc/api-extractor (am I missing anything? :smile:) that can require documentation for public API?

mbroadst (changes requested):

nbbeeken (approved):

src/change_stream.ts
emadum: I think [completed-docs](https://palantir.github.io/tslint/rules/completed-docs/) should do it, I'll file a ticket to add that rule.

src/change_stream.ts
```diff
@@ -277,7 +278,7 @@ export class ChangeStream extends EventEmitter {
   hasNext(callback?: Callback): Promise<void> | void {
     return maybePromise(callback, cb => {
       getCursor(this, (err, cursor) => {
-        if (err || !cursor) return cb(err); // failed to resume, raise an error
+        if (err || !cursor) return cb(err || NO_CURSOR_ERROR); // failed to resume, raise an error
```
mbroadst, line 281: FWIW it looks like if you need to add `NO_CURSOR_ERROR` to each call to `getCursor` then the error should probably be returned from `getCursor`? (maybe per the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)))

mbroadst (commented):
This LGTM with one nitpick. It might also be worth filing a ticket for duplicate `maybePromise`: `ChangeStream#next|tryNext|hasNext` all are top-level API which ultimately delegate to the underlying cursor (if possible) so `maybePromise` is called twice in the happy path. It may be possible to refactor the cursors to allow you to bypass one of those calls

test/functional/change_stream.test.js
```diff
@@ -1924,7 +1924,10 @@ describe('Change Streams', function () {
     });
 
     it('when invoked using eventEmitter API', {
-      metadata: { requires: { topology: 'replicaset', mongodb: '>=3.6' } },
+      metadata: {
+        requires: { topology: 'replicaset', mongodb: '>=3.6' },
+        sessions: { skipLeakTests: true }
```
emadum, line 1929: Note, this test was leaking a session for me locally. Looks like this is because the client can close in the `afterEach` hook before the `changeStream.close` completes. This seems acceptable, so I skipped the leak checker on this test.

mbroadst: why is it:
```
if (counter === 2) {
  changeStream.close();
  setTimeout(() => close());
  ...
}
```
instead of:
```
if (counter === 2) {
  changeStream.close(close);
  ...
}
```
It seems problematic to use `setTimeout` in these cases, it can lead to leaks as you've seen here.

emadum: Good question, fixed 👍

package.json
```diff
@@ -100,7 +100,7 @@
     "check:test": "mocha --recursive test/functional test/unit",
     "check:ts": "tsc -v && tsc --noEmit",
     "check:atlas": "mocha --config \"test/manual/mocharc.json\" test/manual/atlas_connectivity.test.js",
-    "check:adl": "mocha test/manual/data_lake.test.js",
+    "check:adl": "mocha --exit test/manual/data_lake.test.js",
```
mbroadst, line 103: Maybe this is a better place to ask this question. I think adding `--exit` has great potential for hiding bugs, did you file a ticket about the issue this is meant to address so you remember to remove `--exit` in the future?

mbroadst (approved):
LGTM, mod one comment about the use of mocha's `--exit`

package.json
emadum: I was hoping the problem would go away on it's own before I merged these PRs. I'm running a patch build to check now, if it's still timing out I'll file a ticket.

PR #2677, opened 2020-12-14
test: Add a DB log filter by test name cli tool

nbbeeken:
Adds a custom xunit reporter that captures stdio and reports start
and end times of tests.

Add a CLI tool "Crawfish" can take structured mongodb logs and
the new xunit test report and merge the information to
obtain filtered mongodb logs by test name.

Deduplicate suite names to ensure grepping / filtering works as expected

Ensure all tests print to stderr to prevent
logs from interfering with custom reporters.

Update eslint config to correctly handle mjs files 
as well as ignore compiled js in the lib folder

Skunkworks


nbbeeken:
You [can see here](https://evergreen.mongodb.com/test_log/mongo_node_driver_rhel70_dubnium_test_4.2_server_patch_0f2eb49b451dcc7a0fa6b29362baf558246f4bbd_5fd7912b2fbabe7a325522a3_20_12_14_16_22_04/0/APM.should_correctly_receive_the_APM_events_for_an_insert_using_custom_operationId_and_time_generator) on a skipped test what the captured stdout/err looks like, its not spectacularly interesting but now you could conceivably run a patch with a bunch of logs and get a look quickly per test what the outputs are! hopefully that comes in handy down the road

emadum (approved):
LGTM! 🦐 One general stylistic question, why the heavy use of Reflect?

nbbeeken:
For bad TS reasons, I turn on the type checker just for me locally and I'm added properties to these objects that don't originally have them so I get errors, reflect is just me circumventing the red underlines which don't mean anything anyway. I should revert it to just be normal sets and accesses, if it were TS I'd be able to extend the `Mocha.Test` type interface and put in the props I want but alas, another time. 😄

PR #2678, opened 2020-12-15
feat(FindCursor): fluent builder for allowDiskUse option

emadum:
NODE-2955


package.json
```diff
@@ -100,7 +100,7 @@
     "check:test": "mocha --recursive test/functional test/unit",
     "check:ts": "tsc -v && tsc --noEmit",
     "check:atlas": "mocha --config \"test/manual/mocharc.json\" test/manual/atlas_connectivity.test.js",
-    "check:adl": "mocha test/manual/data_lake.test.js",
+    "check:adl": "mocha --exit test/manual/data_lake.test.js",
```
mbroadst, line 103: what's going on here?

src/operations/find.ts
```diff
@@ -52,7 +52,7 @@ export interface FindOptions extends CommandOperationOptions {
   noCursorTimeout?: boolean;
   /** Specify collation (MongoDB 3.4 or higher) settings for update operation (see 3.4 documentation for available fields). */
   collation?: CollationOptions;
-  /** Enables writing to temporary files on the server. */
+  /** Allows disk use for blocking sort operations exceeding 100MB memory. */
```
mbroadst, line 55: I wonder if this might be a good opportunity to experiment with improving these docs even more by linking to them: https://docs.mongodb.com/manual/reference/command/find/#find-cmd-allowdiskuse

Also, does typedoc support any way of marking an option supported since a certain MongoDB version? It would be cool to use jsdoc's `@since` tag here

test/functional/cursor.test.js
```diff
@@ -4089,5 +4089,40 @@ describe('Cursor', function () {
...
+    });
+
+    it('should error if allowDiskUse option used without sort', {
+      metadata: { requires: { mongodb: '>=4.4' } },
+      test: withClient(function (client, done) {
+        const db = client.db('test');
+        db.collection('test_sort_allow_disk_use', (err, collection) => {
+          expect(err).to.not.exist;
+          try {
+            collection.find({}).allowDiskUse();
```
mbroadst, line 4118:
```
expect(() => collection.find({}).allowDiskUse()).to.throw(/allowDiskUse must follow sort/)
```
also nit: maybe the error message could be a bit more informative, "allowDiskUse requires a sort specification" or something. You could write this code `collection.find({}, { allowDiskUse: true })` and it wouldn't be clear because `allowDiskUse` isn't technically following anything (this language seems specific to the builder pattern)

src/cursor/find_cursor.ts
```diff
@@ -353,6 +353,18 @@ export class FindCursor extends AbstractCursor {
     return this;
   }
 
+  /**
+   * Allows disk use for blocking sort operations exceeding 100MB memory.
+   */
+  allowDiskUse(): this {
```
mbroadst, line 359: FWIW `allowDiskUse` is also supported on the aggregation cursor in the shell, and will likely be requested of the driver in the future

package.json
emadum: The Atlas data lake seem to have started leaking a resource recently. I spent some time trying to figure out which resource, but couldn't find anything; this is a temporary measure to get CI green until we can investigate that more.

src/cursor/find_cursor.ts
emadum: Hmm, are you sure? The [documentation](https://docs.mongodb.com/manual/reference/method/cursor.allowDiskUse/) made it seem like it's only supported on find cursors:

> allowDiskUse() has the following form:
> `db.collection.find(<match>).sort(<sort>).allowDiskUse()`

Similarly the documentation for [aggregation cursors](https://docs.mongodb.com/manual/reference/method/db.collection.aggregate/#cursor-behavior) doesn't list that method. All the builder methods are [only listed](https://docs.mongodb.com/manual/reference/method/db.collection.find/#available-mongo-shell-cursor-methods) in the find cursor docs.

mbroadst: That's strange, I'm just referring to what's supported on [the command](https://docs.mongodb.com/manual/reference/command/aggregate/#syntax). I assumed there would be builders for all of those properties, maybe they accidentally omitted it in the shell as part of the server work to implement that feature?

test/functional/cursor.test.js
emadum: > this language seems specific to the builder pattern

That was actually my intent, I was trying to respect the language in the docs I quoted in my other comment where it says `.allowDiskUse()` must come after a `.find().sort()`.

Now that I think a bit more about it, that could be confusing, especially if users are mixing using options and the builder pattern, e.g. `.find({}, { sort: ... }).allowDiskUse().`. 

Maybe it would be better to not throw an error at all?

src/operations/find.ts
emadum: `@since` is not supported by tsdoc unfortunately, but we could maybe use [`@remarks`](https://tsdoc.org/pages/tags/remarks/) to separate summary/details? E.g.

```
/**
 * Allows disk use for blocking sort operations exceeding 100MB memory.
 * @remarks
 * Since MongoDB 3.2.
 * {@link https://docs.mongodb.com/manual/reference/command/find/#find-cmd-allowdiskuse | find command allowDiskUse documentation}
 * /
```

test/functional/cursor.test.js
mbroadst: I think it makes sense to tell the user that their `allowDiskUse` specification will have no effect if some sort isn't specified, I just wouldn't make it specific to needing to follow an `allowDiskUse` builder method. But that's my two cents.

src/cursor/find_cursor.ts
emadum: I think it's worth deferring that work to a separate ticket, since it doesn't seem to be reference in the shell documentation currently. The [current guidance](https://docs.mongodb.com/manual/reference/command/aggregate/#aggregate-data-using-external-sort) is to use the `options` when calling `db.collection.aggregate(pipeline, options)` which we already support.

mbroadst (approved):
LGTM

nbbeeken (approved):

PR #2667, opened 2020-12-07
chore(ci): continuous matrix integration [3.6]

emadum:
The Continuous Matrix Testing project requires a number of changes to our test infrastructure.

This PR:
- prevents testing node versions after carbon on vs2013, which they don't support 
- fixes a flaky change stream test which could sometimes produce invalid JSON and then attempt to parse it 
- adds `SKIP_DEPS` flag to `run-tests.sh` 
    - if this flag is missing, automatically run `install-dependencies.sh` (for matrix tests)
    - adds `SKIP_DEPS=1` to `run tests` Evergreen function, preserving old behavior
- adds `NO_EXIT` flag to `run-tests.sh`
    - if this flag is missing, automatically exit early from test runs that leak resources (for matrix tests)
    - adds `NO_EXIT=1` to `run tests` Evergreen function, preserving old behavior

NODE-2914

PR #2683, opened 2020-12-21
chore(ci): continuous matrix integration [3.3]

emadum:
The Continuous Matrix Testing project requires updates to some driver versions which are no longer maintained.

This PR for the `3.3` branch allows testing the Node driver released to support MongoDB 4.2.

port of #2667

[patch build](https://spruce.mongodb.com/version/5fe10b073066155bd25aa86e/tasks?page=0&taskName=4.2-era-driver)

- enables xunit reporting
- fixes testing on windows
- adds `COMPATIBILITY.md` and skips incompatible tests

NODE-2914

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -1,0 +1,5 @@
+Server Version: 5.0
+- BSON field 'listCollections.collation' is an unknown field.
+
+Server Version: 4.4
+- $where no longer supports deprecated BSON type CodeWScope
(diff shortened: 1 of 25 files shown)


COMPATIBILITY.md
```diff
@@ -0,0 +1,5 @@
+Server Version: 5.0
+- BSON field 'listCollections.collation' is an unknown field.
+
```
emadum, line 3: Note: this is a new documentation file required by the Continuous Matrix Testing design. It lists any server compatibility issues this version of the driver has, by server version. The corresponding failing tests should be skipped after they've been documented here, so that the continuous testing matrix can be green.

nbbeeken (approved):
lgtm, there's one test SASLPrep that's failing but it has to do with replicaset name mismatch, maybe an orchestration issue?

emadum:
> lgtm, there's one test SASLPrep that's failing but it has to do with replicaset name mismatch, maybe an orchestration issue?

Yeah I wasn't able to reproduce it locally, I'm going to try re-running the patch to see if it goes away, but I think it's safe to ignore this since the test passes outside of Evergreen.

emadum:
> I'm going to try re-running the patch to see if it goes away

It passed in second execution.

PR #2692, opened 2021-01-12
docs: correctly generate API docs for tsdoc comments

nbbeeken:
Some patches and updates to make our v4 API docs generate correctly

The toc setting in typedoc.json limits the symbols that appear on the side bar to only those listed, however all symbols are still available to view when clicking into the `exported` section. I selected a set of the most commonly used APIs, let me know if we should include something else.

Updated the readme a bit, removing some of the old links and updating the examples.


nbbeeken:
Here's a look at the landing page:

image

durran (approved):
LGTM

emadum (approved):
LGTM 👍

nbbeeken:
Sorry for the change after approvals, I just added MongoClientOptions to the list of quick access symbols.

durran (approved):
LGTM

mbroadst:
Yikes, sorry I was getting around to this review @nbbeeken. I was wondering if it's possible to use [categorizeByGroup](https://typedoc.org/guides/options/#categorizebygroup) to split the API between things like BSON, Errors and <the rest>. Also can we enable the search? I just want to make it incredibly easy to navigate this documentation, and my guess is that most of the time people either want to quickly drill down with a search or view `MongoClient`/`Db`/`Collection`. 

Also, FWIW the old docs include documentation for CSFLE things (like `AutoEncrypter`)

nbbeeken:
No worries, sorry for merging without your review, luckily search already is working out of the box so thats nice! 

The categorize by group requires a bit more effort, which I totally think we should do but I would suggest we move to another PR anyway for diff reasons. We have to sweep over all the tsdoc comments and add in `@category` annotations 

Made a ticket for this: NODE-3005

PR #2693, opened 2021-01-13
HELP-21340: Adding Tests For Max BSON Size Batching

durran:
## Description

This adds tests to ensure ordered and unordered bulk operations respect the max bson size and split into batches. The previous tests were only testing splitting in to batches based on the number of documents in the batch.

**What changed?**

Added 2 tests in `bulk.test.js`.

### DIFF
--- a/test/functional/bulk.test.js
+++ b/test/functional/bulk.test.js
@@ -1648,12 +1648,9 @@ describe('Bulk', function () {
       })
       .then(() => {
         const coll = db.collection('doesnt_matter');
-
-        coll.insertMany(documents, { ordered: true }, err => {
-          client.close(() => {
-            done(err);
-          });
-        });
+        coll.insertMany(documents, { ordered: true })
+      .then(() => {
+        client.close(done);
       });
   });
 
@@ -1677,12 +1674,9 @@ describe('Bulk', function () {
       })
       .then(() => {
         const coll = db.collection('doesnt_matter');
-
-        coll.insertMany(documents, { ordered: false }, err => {
-          client.close(() => {
-            done(err);
-          });
-        });
+        coll.insertMany(documents, { ordered: false })
+      .then(() => {
+        client.close(done);
       });
   });
 
--- a/test/functional/bulk.test.js
+++ b/test/functional/bulk.test.js
@@ -1628,7 +1628,7 @@ describe('Bulk', function () {
       });
   });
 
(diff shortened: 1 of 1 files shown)


emadum (approved):
LGTM with one test style nit 👍

test/functional/bulk.test.js
```diff
@@ -1625,6 +1628,64 @@ describe('Bulk', function () {
...
+      })
+      .then(() => {
+        const coll = db.collection('doesnt_matter');
+
+        coll.insertMany(documents, { ordered: false }, err => {
+          client.close(() => {
+            done(err);
+          });
+        });
+      });
```
nbbeeken, line 1686:
```suggestion
      .then(coll => {
        return coll.insertMany(documents, { ordered: false });
      })
      .finally(() => client.close());
```
Just a suggestion but you could maintain the use of promises all the way down here, seems more consistent.
This is the opposite of what @emadum you suggest since it will still implicitly pass the potential error along to mocha, thoughts?

emadum: That's a good point and SGTM, just using promises all the way down is probably more idiomatic. In addition to above suggestion the tests would also need to return the promise chain.

durran: I like this suggestion. I've updated.

test/functional/bulk.test.js
```diff
@@ -1625,6 +1628,58 @@ describe('Bulk', function () {
...
+      .then(() => {
+        return db.createCollection('doesnt_matter');
+      })
+      .then(() => {
+        const coll = db.collection('doesnt_matter');
+        coll.insertMany(documents, { ordered: true });
+      })
+      .then(() => {
+        client.close();
+      });
```
nbbeeken, line 1654:
```suggestion
      .finally(() => client.close());
```
I think the errors on CI are related to the `client.close()` promise not being part of the chain (either simple arrow or a return is fine, no style preference). 

I suggest finally here incase there's an error the tests will still exit. We've just had a number of issues in the past with tests hanging the mocha run so we wanna make best effort going forward of making sure we get a clean exit whenever we can.

nbbeeken (approved):
Barring minor CI bug: LGTM!

test/functional/bulk.test.js
durran: Updated.

PR #2694, opened 2021-01-13
HELP-21340: Adding Tests For Max BSON Size Batching (3.1)

durran:
This adds tests to ensure ordered and unordered bulk operations respect the max bson size and split into batches. The previous tests were only testing splitting in to batches based on the number of documents in the batch.

What changed?

Added 2 tests in bulk_tests.js.


durran:
Closing as to open against 3.6

PR #2695, opened 2021-01-13
fix: ingest WC as a simple object or number for w value

nbbeeken:
This reverts a change that now makes it possible to pass a
writeConcern option that maps directly to 'majority'
or a number for the w value

NODE-2836

### DIFF
--- a/test/functional/write_concern.test.js
+++ b/test/functional/write_concern.test.js
@@ -69,9 +69,7 @@ describe('Write Concern', function () {
     });
   });
 
-  after(() => {
-    mock.cleanup();
-  });
+  after(() => mock.cleanup());
 
   it('should pipe writeConcern from client down to API call', function () {
     server.setMessageHandler(request => {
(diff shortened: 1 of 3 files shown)


test/functional/write_concern.test.js
```diff
@@ -58,4 +61,40 @@ describe('Write Concern', function () {
...
+
+  let server;
+  before(() => {
+    return mock.createServer().then(s => {
+      server = s;
+    });
+  });
+
+  after(() => {
+    mock.cleanup();
```
emadum, line 73: This should also return (or just remove the curly braces)

emadum (changes requested):
Looks like there is still a leaked resource in the CI run, maybe my requested change might fix it?

test/functional/write_concern.test.js
nbbeeken: Ooo nice catch thanks!!

emadum (approved):
LGTM

durran (approved):
LGTM

PR #2696, opened 2021-01-13
fix: hasAtomicOperator check respects toBSON transformation

nbbeeken:
Certain documents cannot contain atomic operators
i.e. keys with a leading dollar sign, documents can contain
toBSON transformation functions that would modify such keys the check
now respect the transformation

NODE-2741

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -809,12 +809,11 @@ function hasAtomicOperators(doc) {
     return doc.reduce((err, u) => err || hasAtomicOperators(u), null);
   }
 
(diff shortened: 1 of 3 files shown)


emadum (approved):
Stylistic nit, I'd prefer a ternary expression inside `Object.keys` rather than the `docToTest` variable, but LGTM 👍

durran (changes requested):
Although it's a minor fix would it be possible add a test to show the behaviour works please?

durran (approved):
LGTM

PR #2698, opened 2021-01-14
feat(ci): add mongosh integration testing

addaleax:
This is part of NODE-2940 (it doesn’t address Compass integration, because I don’t know what that would even look like given the many, many packages that make up Compass).

--- 

Add a mongosh integration test task to evergreen. This currently passes CI, but could fail if more breaking changes are being added. I’m doing my best to keep up with all driver changes, but we still have to adjust our code on a regular basis to changes being made to the master branch here – if you want to avoid this task failing occasionally, you can of course wait until the actual 4.0 release (or until all planned breaking changes are done).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -137,6 +137,14 @@ functions:
         script: |
           ${PREPARE_SHELL}
           bash ${PROJECT_DIRECTORY}/.evergreen/run-checks.sh
+  run mongosh integration tests:
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: src
+        script: |
+          ${PREPARE_SHELL}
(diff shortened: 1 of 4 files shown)


durran:
I'm personally ok with going against master as we will find issues immediately rather than waiting for an official release. But I'm also willing to defer to other opinions.

durran (approved):
LGTM

nbbeeken (approved):
LGTM, I agree we can still merge if we run into issues but we'll have a good early warning system, thanks for this!

emadum (approved):

addaleax:
Fwiw, the failure on its first waterfall run is because of such a breaking change that happened here since I opened the PR :) As mentioned, that will happen from time to time, but I’ll do my best to keep up to date with changes (and try to figure out if there’s a way to get evergreen notifications for failures of specific build variants).

PR #2700, opened 2021-01-14
fix(find): correctly translate timeout option into noCursorTimeout

emadum:
The find command's `timeout` option was incorrectly being translated
into `noCursorTimeout`, this has been corrected to be `!timeout`.

NODE-2828


durran (approved):
LGTM

nbbeeken (approved):

PR #2702, opened 2021-01-15
NODE-2355: Allow GridFS write stream to destroy

durran:
## Description

The `GridFSBucketWriteStream` needs to invoke the `Writable` constructor, not just inherit from it, in order to properly initialise the internal state of the stream. This allows the stream to be destroyed properly.


addaleax (approved):

emadum (changes requested):
This LGTM, mod one nit: this test is written in the "old" style; we've been trying write new tests using `const/let` instead of `var`, and chai `expect` rather than the old `test.equal`/`test.ok` style of assertions.

durran:
@emadum Sorry about that - I was just being consistent with the tests currently in that file. Would you be ok with leaving this PR as is (so it's just the relevant changes) and opening a new PR that refactors all the tests in that file to the new style? That way this PR doesn't get cluttered with refactorings.

nbbeeken (approved):

emadum:
> @emadum Sorry about that - I was just being consistent with the tests currently in that file. Would you be ok with leaving this PR as is (so it's just the relevant changes) and opening a new PR that refactors all the tests in that file to the new style? That way this PR doesn't get cluttered with refactorings.

SGTM! 👍

emadum (approved):

PR #2704, opened 2021-01-15
fix: revert BSON Map export removal

nbbeeken:
Good news! Our mongosh early warning system works!
Looks like I dropped the Map export from BSON by mistake, here's the fix.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,7 +41,8 @@ export {
   Timestamp,
   Decimal128,
   BSONRegExp,
-  BSONSymbol
+  BSONSymbol,
+  Map
 } from './bson';
 
 export {


emadum (approved):

durran (approved):
LGTM

addaleax:
@nbbeeken Fwiw, we already worked around this in https://github.com/mongodb-js/mongosh/commit/39e6dfbafa6f18ee723aac8fa0cba0e31eab6a0b – you can keep these dropped if that’s what you prefer for 4.0.

nbbeeken:
Thanks for the heads up! I think we want to keep it until it would be removed from the BSON lib in it's major version, despite it not being the most useful.

PR #2705, opened 2021-01-17
fix(cursor): don't use other operation's session for cloned cursor operation (NODE-3008)

KonanMentor:
Fixes NODE-3008.

## Description

Please refer to NODE-3008 for more details.

I can try to write a test for this bug if someone give me a hint on an example of such tests.

Also we should check if this should be ported to v4.x

### DIFF
--- a/lib/core/cursor.js
+++ b/lib/core/cursor.js
@@ -11,6 +11,7 @@ const executeOperation = require('../operations/execute_operation');
 const Readable = require('stream').Readable;
 const SUPPORTS = require('../utils').SUPPORTS;
 const MongoDBNamespace = require('../utils').MongoDBNamespace;
+const mergeOptions = require('../utils').mergeOptions;
 const OperationBase = require('../operations/operation').OperationBase;
 
 const BSON = retrieveBSON();
@@ -207,7 +208,9 @@ class CoreCursor extends Readable {
    * @return {Cursor}
    */
   clone() {
-    return this.topology.cursor(this.ns, this.cmd, this.options);
+    const clonedOptions = mergeOptions({}, this.options);
+    delete clonedOptions.session;
+    return this.topology.cursor(this.ns, this.cmd, clonedOptions);
   }
 
   /**


durran:
Hi @KonanMentor. Thank you so much for the related JIRA ticket and the PR. What I'm trying to understand is the use case for actually wanting to clone a cursor. Would it be possible for you to provide the reasoning for this and what you are trying to achieve with this functionality? 

Once again, I thank you for your contribution and look forward to your response.

KonanMentor:
Hi @durran. My use case for using `cursor.clone` is about rendering progress of cursor iteration. By rendering I mean just a string in the following format: `` `${processed} / ${total}` ``. The progress string is rendered by [listr](https://github.com/SamVerschueren/listr/) in my case. I'll try to replicate the implementation quite close to the real thing, so there will be some RxJS.

I guess, the most crucial part is
```js
const clonedCursor = cursor.clone(); // clone cursor in order to avoid execution of both `find` and `count` operations on the same cursor
const totalPromise = clonedCursor.count(true);
// An alternative way would be passing `query` + `collection` instead of `cursor`
// const totalPromise = collection.countDocuments(query);
// But it will become even more inconvenient if the cursor has `limit` and/or `skip` operators applied
```

Here is the full story:

```js
const cursor = collection.find(QUERY);
const documents$ = asyncIterableToObservable(cursor) // this function treats cursor as asyncIterable and converts it to RxJS Observable
  .share(); // In order to avoid subscribing to the observable second time (i.e. iterating the cursor same twice) in addListrTask function
processDocuments(documents$);
addListrTask(cursor, documents$);

function processDocuments(documents$) {
  documents$
    .mergeMap(doc => doSomeWork(doc))
    .subscribe();
}

function addListrTask(cursor, documents$) {
  const clonedCursor = cursor.clone(); // clone cursor in order to avoid execution of both `find` and `count` operations on the same cursor
  const totalPromise = clonedCursor.count(true);
  // An alternative way would be passing `query` + `collection` instead of `cursor`
  // const totalPromise = collection.countDocuments(query);
  // But it will become even more inconvenient if the cursor has `limit` and/or `skip` operators applied

  // RxJS magic starts
  const progessedCount$ = documents$
    .scan((progessedCount) => progessedCount + 1, 0)
    .startWith(0);
  const progressString$ = Observable.combineLatest(progessedCount$, totalPromise)
    .map(([processed, total]) => `${processed} / ${total}`);
  // Add listr task, render progressString$
}
```

durran:
Ok I've tested this and have a test written for the suite as well. So I'll merge and then put a PR up with that test for it, and then a new PR with the port to 4.0. @emadum @nbbeeken Are you both ok with that?

emadum (approved):

nbbeeken (approved):
SGTM

PR #2709, opened 2021-01-20
feat: expose package version on MongoClient

addaleax:
## Description

Provide the current package version statically. The idea here is to use this to solve NODE-3043, by allowing mongodb-client-encryption to pass the correct options for a specific driver version.

If this is merged, I’ll make the corresponding change in the libmongocrypt repository. If you think that this is not a good approach, that’s very much okay and you can just go ahead and close this in that case.

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -11,7 +11,8 @@ import {
   resolveOptions,
   ClientMetadata,
   ns,
-  HostAddress
+  HostAddress,
+  NODE_DRIVER_VERSION
 } from './utils';
 import { deprecate } from 'util';
 import { connect } from './operations/connect';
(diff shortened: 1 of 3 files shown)


addaleax:
(Just forced-pushed this and #2710 to kick off evergreen)

durran:
@addaleax This seems a reasonable change here, but just a quick question... If I am looking at this correctly is it the intention that libmongocrypt will now with this change keep a map of sorts to map valid driver options to the driver version? If so, it seems that may be better placed in the driver itself, since those options are not specific to FLE but the core driver itself. What do you think?

addaleax:
@durran Basically, yes - for the 4.0 driver, libmongocrypt would omit the now-removed options. 

> If so, it seems that may be better placed in the driver itself, since those options are not specific to FLE but the core driver.

I'm ultimately happy with any solution here. If you are suggesting to list all valid options in a programmatically accessible way, then I think that's great to have either way, and also something that would solve this problem on the FLE side. The FLE addon would still have to do *some* version-specific code, because no published version of the driver provides a list like that (as far as I know, that is).

nbbeeken:
As discussed in sync meeting, closing this in favor of #2713 which reintroduces the options that mongo-client-encryptions uses, this will allow existing versions to keep working with version 4.0 of the driver. Thank you for your help in getting to the bottom of the breakage 🚀

PR #2712, opened 2021-01-21
test: Complete all matching operations

nbbeeken:
Adds special operations support i.e. $$unsetOrMatches for testing
nested schema equality of results and events. Adds Recursive
equality test that tracks the path into the object it is testing
for ease of error tracing. Enable Change Stream tests with the
repurposed EventsCollector class to iterate change events. Adds
find, insertMany, iterateUntilDocumentOrError, and failPoint operations.

NODE-2287

### DIFF
--- a/test/functional/unified-spec-runner/match.ts
+++ b/test/functional/unified-spec-runner/match.ts
@@ -168,24 +168,26 @@ export function expectResultCheck(
   actual: Document,
   expected: Document | number | string | boolean,
   entities: EntitiesMap,
-  path: string[] = []
+  path: string[] = [],
+  depth = 0
 ): boolean {
-  const result = resultCheck(actual, expected, entities, path);
-  if (result[0] === false) {
-    const path = result[1].join('');
+  const ok = resultCheck(actual, expected, entities, path, depth);
+  if (ok === false) {
+    const pathString = path.join('');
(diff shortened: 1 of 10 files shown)


test/functional/unified-spec-runner/match.ts
```diff
@@ -0,0 +1,301 @@
...
+    isTypeOperator(value) ||
+    isMatchesEntityOperator(value) ||
+    isMatchesHexBytesOperator(value) ||
+    isUnsetOrMatchesOperator(value) ||
+    isSessionLsidOperator(value)
+  );
+}
+
+const TYPE_MAP = new Map();
+function typeof_double(actual): actual is number | Double {
```
durran, line 81: Just out of curiosity is there a pattern of using snake case in typescript? I hadn't seen this before.

emadum: Could we perhaps have the functions inline in the Map.set calls? E.g.
```
TYPE_MAP.set('double', function (actual): actual is number | Double {
  return typeof actual === 'number' || actual._bsontype === 'Double';
});
```

emadum (approved):
LGTM, nice re-use of `EventCollector` 👍

durran (approved):
LGTM

PR #2713, opened 2021-01-22
fix: Compatibility with mongodb-client-encryption

nbbeeken:
Permits legacy options being used by mongodb-client-encryption's
use of the MongoClient constructor. Permits legacy WC option style in
insertOne.

NODE-3006

[CSFLE Tests Patch](https://spruce.mongodb.com/task/mongo_node_driver_ubuntu_18.04_erbium_test_4.2_replica_set_patch_25ef87026bca935b00d30ad943ddd28784f801e8_600b33953627e07e4a808b81_21_01_22_20_20_38/tests?execution=0&page=0&testname=Client_Side_Encryption)

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -278,7 +278,16 @@ export class Collection {
     options?: InsertOneOptions | Callback<InsertOneResult>,
     callback?: Callback<InsertOneResult>
   ): Promise<InsertOneResult> | void {
-    if (typeof options === 'function') (callback = options), (options = {});
+    if (typeof options === 'function') {
+      callback = options;
+      options = {};
+    }
+
+    // CSFLE passes in { w: 'majority' } to ensure the lib works in both 3.x and 4.x
+    // we support that option style here only
+    if (options && Reflect.get(options, 'w')) {
+      options.writeConcern = WriteConcern.fromOptions(Reflect.get(options, 'w'));
+    }
 
     return executeOperation(
       getTopology(this),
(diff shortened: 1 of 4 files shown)


src/collection.ts
```diff
@@ -278,7 +278,16 @@ export class Collection {
...
+    if (typeof options === 'function') {
+      callback = options;
+      options = {};
+    }
+
+    // CSFLE passes in { w: 'majority' } to ensure the lib works in both 3.x and 4.x
+    // we support that option style here only
+    if (options && Reflect.get(options, 'w')) {
+      options.writeConcern = WriteConcern.fromOptions(Reflect.get(options, 'w'));
+    }
```
nbbeeken, line 290: Here that is just checking for the legacy writeConcern shape

src/connection_string.ts
```diff
@@ -1062,7 +1062,10 @@ export const OPTIONS = {
   passphrase: { type: 'any' },
   pfx: { type: 'any' },
   secureProtocol: { type: 'any' },
-  index: { type: 'any' }
+  index: { type: 'any' },
+  // Legacy Options, these are unused but left here to avoid errors with CSFLE lib
+  useNewUrlParser: { type: 'boolean' } as OptionDescriptor,
+  useUnifiedTopology: { type: 'boolean' } as OptionDescriptor
```
nbbeeken, line 1068: And here that's supporting the old options. These still aren't listed in the props for MongoClientOptions (which is why we have the `as` casting). I would add that they are deprecated but then our CSFLE lib would emit the warnings with no way to disable.

nbbeeken (commented):
Instead of making a change to 'mongodb-client-encryption` and managing version compatibility I think we should just support the small patches needed to make the existing package work.

emadum (approved):
LGTM

durran (approved):
LGTM

PR #2714, opened 2021-01-23
fix(write_concern): avoid throwing error when `options` is null

vkarpov15:
## Description

**What changed?**

If you call `createCollection()` with `null` for options, you get a "Cannot read property 'writeConcern' of null" error. Not a problem in TypeScript because technically this function doesn't accept `null` as a parameter, but might be worth fixing in case JS developers run into it.


emadum (approved):
LGTM 👍

nbbeeken (approved):
LGTM, I'm betting we have a few other places where we're stricter than needed so I created [NODE-3058](https://jira.mongodb.org/browse/NODE-3058)

durran (approved):

PR #2719, opened 2021-01-27
fix: restore auto direct connection behavior

emadum:
## Description

We removed automatic direct connection for the unified topology in the `3.6.3` release of the driver. This change was preparatory for the `4.0` version of the driver, where we'll always perform automatic discovery. This PR restores automatic direct connection when connecting to a single host without a specified `replicaSet` and without `directConnection: false`. 

In #2557 this was justified because `useUnifiedTopology: true` is an opt-in feature, but we've had a number of users file tickets about this change, and it's probably more appropriate to delay it until the major release of the driver, along with a call-out/explanation in the migration guide, rather than putting it in a patch release. That said, I'm open to closing this PR if we decide the original reasoning was appropriate.

NODE-2966

**What changed?**

### DIFF
--- a/lib/core/uri_parser.js
+++ b/lib/core/uri_parser.js
@@ -687,6 +687,15 @@ function parseConnectionString(uri, options, callback) {
     return callback(new MongoParseError('directConnection option requires exactly one host'));
   }
 
+  // NOTE: this behavior will go away in v4.0, we will always auto discover there
+  if (
+    parsedOptions.directConnection == null &&
+    hosts.length === 1 &&
+    parsedOptions.replicaSet == null
+  ) {
+    parsedOptions.directConnection = true;
+  }
+
   const result = {
     hosts: hosts,
     auth: auth.db || auth.username ? auth : null,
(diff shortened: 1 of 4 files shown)


durran (changes requested):
@emadum What do you think about adding a test for the specific scenario? Even with the functionality going away I think it would be beneficial not to have a regression later in the 3.6 branch. I was expecting a test where `directConnection` and `replSet` are both not supplied when connecting directly to a replica set node.

emadum:
@durran Adding a test definitely seems reasonable! I'll do so once it's confirmed we all agree about reverting this change.

nbbeeken:
I agree with the change, should we consider a deprecation warning of sorts, as long as it can be hushed I think its a good place to notify users.

emadum:
> I agree with the change, should we consider a deprecation warning of sorts? as long as it can be hushed I think its a good place to notify users.

Just to have a record of our conversation, we aren't going to emit a deprecation warning because it would also be emitted whenever connecting to a standalone MongoDB server. However we should definitely make sure to call this out in the 4.0 migration guide.

nbbeeken (approved):

PR #2720, opened 2021-01-27
chore(ci): fix aws auth tests

emadum:
## Description

Upgrades the `boto3` python dependency used in the AWS auth tests by means of the newly added `activate_venv.sh` script.

NODE-3056

**What changed?**

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -67,8 +67,8 @@ functions:
         script: >
           ${PREPARE_SHELL}
 
-          git clone git://github.com/mongodb-labs/drivers-evergreen-tools.git $DRIVERS_TOOLS
-
+          git clone --branch DRIVERS-1462/aws-auth-upgrade-boto3 --single-branch \
+            git://github.com/emadum/drivers-evergreen-tools.git $DRIVERS_TOOLS
           echo "{ \"releases\": { \"default\": \"$MONGODB_BINARIES\" }}" >
           $MONGO_ORCHESTRATION_HOME/orchestration.config
   bootstrap mongo-orchestration:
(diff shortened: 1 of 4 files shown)


.evergreen/config.yml
```diff
@@ -243,6 +243,14 @@ functions:
...
           }
           EOF
+  upgrade boto3 aws library for python:
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: src
+        silent: true
+        script: |
+          pip install --upgrade boto3
```
ShaneHarvey, line 253: This will upgrade boto3 for the entire system. This is bad practice because it can break applications that use the system python and even break future Evergreen tasks that run on this machine. Instead you should create and use a virtualenv, like this:
```
          # The aws_e2e_assume_role script requires python3 with boto3.
          virtualenv -p ${python3_binary} mongovenv
          if [ "Windows_NT" = "$OS" ]; then
            . mongovenv/Scripts/activate
          else
            . mongovenv/bin/activate
          fi
          pip install --upgrade boto3
```

Since this is a virtualenv you'll need to activate it before running each JS script, eg:
```
          # The mongovenv was created earlier.
          if [ "Windows_NT" = "$OS" ]; then
            . mongovenv/Scripts/activate
          else
            . mongovenv/bin/activate
          fi
          cd ${DRIVERS_TOOLS}/.evergreen/auth_aws
          ${MONGODB_BINARIES}/mongo aws_e2e_regular_aws.js
```

Ideally the drivers-evergreen-tools auth testing scripts would take care of these problems so that all drivers don't need to add this finicky boilerplate.

ShaneHarvey (changes requested):

.evergreen/config.yml
ShaneHarvey: Oh, and thanks for fixing this bug! I was able to fix it in pymongo too: https://github.com/mongodb/mongo-python-driver/pull/562

emadum:
Note: depends on mongodb-labs/drivers-evergreen-tools#138

durran (approved):
LGTM

PR #2721, opened 2021-01-27
fix: Support bson property on Topology class

nbbeeken:
This addresses an incompatibility with the existing versions of
mongodb-client-encryption which falls back on the bson library
being attached to the topology class.

NODE-3041


addaleax (approved):

emadum (approved):
LGTM

durran (approved):
LGTM

PR #2725, opened 2021-02-01
NODE-3008: Dont pass session to cloned cursor

durran:
Removes the session from the cloned cursor so it is not reused.

### DIFF
--- a/test/functional/cursor.test.js
+++ b/test/functional/cursor.test.js
@@ -12,6 +12,7 @@ const { ReadPreference } = require('../../src/read_preference');
 const { ServerType } = require('../../src/sdam/common');
 const { formatSort } = require('../../src/sort');
 const { FindCursor } = require('../../src/cursor/find_cursor');
+const kSession = Symbol('session');
 
 describe('Cursor', function () {
   before(function () {
@@ -107,7 +108,6 @@ describe('Cursor', function () {
       });
     }
   });
-
 
   it('cursor should trigger getMore', {
     // Add a tag that our runner can trigger on
@@ -3785,7 +3785,7 @@ describe('Cursor', function () {
               expect(doc).to.exist;
               const clonedCursor = cursor.clone();
               expect(clonedCursor.cursorOptions.session).to.not.exist;
-              expect(clonedCursor.session).to.not.exist;
+              expect(clonedCursor[kSession]).to.not.exist;
             })
             .finally(() => {
               return cursor.close();
@@ -3805,7 +3805,7 @@ describe('Cursor', function () {
               expect(doc).to.exist;
               const clonedCursor = cursor.clone();
               expect(clonedCursor.cursorOptions.session).to.not.exist;
-              expect(clonedCursor.session).to.not.exist;
+              expect(clonedCursor[kSession]).to.not.exist;
             })
             .finally(() => {
               return cursor.close();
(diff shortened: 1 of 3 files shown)


test/functional/cursor.test.js
```diff
@@ -3755,6 +3756,65 @@ describe('Cursor', function () {
...
+        const docs = [{ name: 'test1' }, { name: 'test2' }];
+        return collection.insertMany(docs).then(() => {
+          const cursor = collection.find({}, { batchSize: 1 });
+          return cursor
+            .next()
+            .then(doc => {
+              expect(doc).to.exist;
+              const clonedCursor = cursor.clone();
+              expect(clonedCursor.cursorOptions.session).to.not.exist;
+              expect(clonedCursor[kSession]).to.not.exist;
```
nbbeeken, line 3788: This kSession isn't equal to the kSession declared in `src/cursor/abstract_cursor.ts` so this will always be true. The cursor class has a `session` getter method that would accurately check the kSession property.

nbbeeken (changes requested):

test/functional/cursor.test.js
durran: Nice catch. Updated.

nbbeeken (approved):

test/functional/cursor.test.js
```diff
@@ -3854,66 +3914,6 @@ describe('Cursor', function () {
...
-      configuration: configuration,
-      collectionName: 'stream-test-transform',
-      transformFunc: doc => ({ _id: doc._id, b: doc.a.b, c: doc.a.c }),
-      expectedSet: new Set(expectedDocs)
-    };
-
-    testTransformStream(config, done);
-  });
-
-  it('stream should return a stream of unmodified docs if no transform function applied', function (done) {
```
emadum, line 3870: Hmm, why are these tests being removed?

test/functional/cursor.test.js
```diff
@@ -4049,61 +4142,15 @@ describe('Cursor', function () {
...
-    it('should use find options object', findSort({ alpha: 1 }, { alpha: 1 }));
-    it('should use find options string', findSort('alpha', { alpha: 1 }));
-    it('should use find options shallow array', findSort(['alpha', 1], { alpha: 1 }));
-    it('should use find options deep array', findSort([['alpha', 1]], { alpha: 1 }));
-
-    it('should use cursor.sort object', cursorSort({ alpha: 1 }, { alpha: 1 }));
-    it('should use cursor.sort string', cursorSort('alpha', { alpha: 1 }));
-    it('should use cursor.sort shallow array', cursorSort(['alpha', 1], { alpha: 1 }));
-    it('should use cursor.sort deep array', cursorSort([['alpha', 1]], { alpha: 1 }));
-
```
emadum, line 4061: And these as well?

emadum (changes requested):
Some tests got dropped here that seemed unrelated/unintended. Maybe there was some kind of issue with a merge or rebase?

test/functional/cursor.test.js
durran: Oh that wasn't intentional, I'll bring those back.

durran: Fixed.

test/functional/cursor.test.js
durran: Fixed.

emadum (approved):

PR #2728, opened 2021-02-02
test: restrict destroy test to node v8 and up

emadum:
Note: [writable.destroy](https://nodejs.org/api/stream.html#stream_writable_destroy_error) was added in node v8.0.0

NODE-3065

### DIFF
--- a/test/functional/gridfs_stream.test.js
+++ b/test/functional/gridfs_stream.test.js
@@ -100,7 +100,7 @@ describe('GridFS Stream', function() {
   });
 
   it('destroy publishes provided error', {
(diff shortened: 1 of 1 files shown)


nbbeeken (approved):
LGTM, was the changes to the PR ^ just hunting down the source of the problem?

emadum:
> LGTM, was the changes to the PR ^ just hunting down the source of the problem?

There were some failures in the CSFLE tests that I wanted to include a fix for as part of this PR, but that turned out to be an error in the Evergreen project settings and didn't need any code changes.

PR #2730, opened 2021-02-03
test: more Unified runner operations

nbbeeken:
Adds a number of operations, makes use of chai expect instead of boolean
for assertions to track stack traces, adds session and bucket entity
types. Passes utilClient through to operation functions which are"
inlined into the operations map to get type checking to work.
Modifies url helper function to make multipleMongoses detection work.

### DIFF
--- a/test/functional/unified-spec-runner/match.ts
+++ b/test/functional/unified-spec-runner/match.ts
@@ -257,7 +257,7 @@ export function expectErrorCheck(
     return;
   }
 
-  if (expected.errorContains) {
+  if (expected.errorContains != null) {
     expect(error.message).to.include(expected.errorContains);
   }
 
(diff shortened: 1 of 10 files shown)


nbbeeken:
Working on the few failures left, but shouldn't be drastic so I think this is ready for a look, thanks!

test/functional/unified-spec-runner/match.ts
```diff
@@ -254,48 +258,42 @@ export function expectErrorCheck(
...
   }
 
   if (!(error instanceof MongoError)) {
-    throw new Error(`Assertions need ${error} to be a MongoError`);
+    // if statement asserts type for TS, expect will always fail
+    expect(error).to.be.instanceOf(MongoError);
+    return;
   }
 
   if (expected.errorCode) {
```
emadum, line 270: I think it'd be better to do `if (expected.errorCode != null)` here and in the other checks to allow for expectations with false-y values.

emadum (approved):
One minor request but LGTM 👍

PR #2732, opened 2021-02-05
fix: session support detection spec compliance

nbbeeken:
Discovered while working on a help ticket the reduce callback in the topology description
doesn't correctly return null if it's in the middle of the list of servers.

### DIFF
--- a/test/unit/core/common.js
+++ b/test/unit/core/common.js
@@ -41,8 +41,8 @@ class ReplSetFixture {
         arbiters: [this.arbiterServer.uri()]
       });
 
-      this.defineReplSetStates();
-      // this.configureMessageHandlers();
+      if (!options.doNotInitStates) this.defineReplSetStates();
+      if (!options.doNotInitHandlers) this.configureMessageHandlers();
     });
   }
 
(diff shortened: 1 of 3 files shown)


durran (approved):
LGTM

test/unit/core/common.js
```diff
@@ -11,30 +11,39 @@ class ReplSetFixture {
...
+        __nodejs_mock_server__: true,
+        setName: 'rs',
+        setVersion: 1,
+        electionId: this.electionIds[0],
+        hosts: this.servers.map(server => server.uri()),
+        arbiters: [this.arbiterServer.uri()]
+      });
+
+      this.defineReplSetStates();
+      // this.configureMessageHandlers();
```
emadum, line 45: Should this line be commented out? Seems there are some issues in CI.

nbbeeken: Whoops good catch, let's see if the commit I just pushed fixes it. I made the change on the 4.0 port already to add if statements to gate these extra setup calls.

emadum (approved):
LGTM 👍

PR #2734, opened 2021-02-08
fix: MongoError circular dependency

xkmgt:
Bugfix for: "Accessing non-existent property 'MongoError' of module exports inside circular dependency"

Related to https://developer.mongodb.com/community/forums/t/warning-accessing-non-existent-property-mongoerror-of-module-exports-inside-circular-dependency/15411

## Description

**What changed?**

### DIFF
--- a/lib/operations/operation.js
+++ b/lib/operations/operation.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const Explain = require('../explain').Explain;
-const MongoError = require('../core').MongoError;
+const MongoError = require('../core/error').MongoError;
 
 const Aspect = {
   READ_OPERATION: Symbol('READ_OPERATION'),
--- a/lib/operations/options_operation.js
+++ b/lib/operations/options_operation.js
@@ -2,7 +2,7 @@
 
 const OperationBase = require('./operation').OperationBase;
 const handleCallback = require('../utils').handleCallback;
-const MongoError = require('../core').MongoError;
+const MongoError = require('../core/error').MongoError;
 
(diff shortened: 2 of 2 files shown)


nbbeeken:
@xkmgt Thanks so much for helping us out with this, testing locally looks like we need to make this same change in `lib/operations/operation.js`. Are you not seeing the warning emit a stack trace leading to that file? We may need to dig deeper to make sure we catch all the cases where the require calls have this warning.

xkmgt:
@nbbeeken you are right. `lib/operations/operation.js` is the file which causes the circularity. 

These files require `MongoError` and `OperationBase` at the same time:

````
const MongoError = require('../core').MongoError;
const OperationBase = require('./operation').OperationBase;
````

- `lib\operations\bulk_write.js`
- `lib\operations\command_v2.js`
- `lib\operations\command.js`
- `lib\operations\execute_db_admin_command.js`
- `lib\operations\execute_operation.js`
- `lib\operations\find_and_modify.js`
- `lib\operations\find.js`
- `lib\operations\insert_many.js`
- `lib\operations\insert_one.js`
- `lib\operations\map_reduce.js`
- `lib\operations\options_operation.js`

Changing the path in `lib/operations/operation.js` solves it. The mod in `lib\operations\options_operation.js` is unnecessary.

It's running now for some hours in a productive project without any errors. I am happy with it ;-)

nbbeeken (approved):

durran (approved):

emadum (approved):

leon:
When can you release a new version containing this?

PR #2735, opened 2021-02-08
fix(operations): fix deprecation warning in Node.js 14

vkarpov15:
## Description

**What changed?**

#2626 introduced a deprecation warning in Node.js 14 due to a circular import, because `../core/index.js` eventually imports `operations/operation.js`. This works around the deprecation warning and makes sure `MongoError` is actually defined in this file.

Nope

### DIFF
--- a/lib/operations/operation.js
+++ b/lib/operations/operation.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const Explain = require('../explain').Explain;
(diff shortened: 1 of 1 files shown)


nbbeeken:
Thanks for helping out! since this is a duplicate of #2734 we'll just go with that one since it came in first

PR #2736, opened 2021-02-09
feat: versioned api

emadum:
Allows users to select an API version when connecting to a MongoDB instance.

NODE-2950


test/functional/unified-spec-runner/match.ts
```diff
@@ -129,10 +129,7 @@ export function resultCheck(
 
       if (depth > 1) {
         expect(actual, `Expected actual to exist at ${path.join('')}`).to.exist;
-        expect(
-          Object.keys(actual),
-          `[${Object.keys(actual)}] length !== [${Object.keys(expected)}]`
-        ).to.have.lengthOf(Object.keys(expected).length);
+        expect(Object.keys(actual)).to.include.members(Object.keys(expected));
```
nbbeeken, line 132: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#allowing-extra-fields-in-root-level-documents

Unfortunately this is part of the enforcement of the unified spec, is there a test in particular thats failing if you reverse this?
Or perhaps my depth check isn't correctly filtering out the root level document?

emadum: No, your code was correct, thanks for pointing this out. After a second look, [this line](https://github.com/mongodb/node-mongodb-native/blob/4.0/src/operations/update.ts#L277) was causing some spec tests to fail due to a presence of `multi: false`, because `multi` defaults to false and should only be passed if true.

src/cursor/abstract_cursor.ts
```diff
@@ -572,6 +572,10 @@ export abstract class AbstractCursor extends EventEmitter {
       return;
     }
 
+    if (!this[kSession]) {
+      throw new Error('Should have a session when calling getMore');
```
emadum, line 576: This ended up being the cause of a bug I had a very hard time tracking down. It only happens on `latest` running with `--setParameter requireApiVersion=1`. I had to make the changes below to avoid this error being thrown in a number of tests. @nbbeeken might this be related to your sessions work?

src/mongo_client.ts
```diff
@@ -32,6 +32,32 @@ import type { SrvPoller } from './sdam/srv_polling';
...
+/** @public */
+export const ServerApiVersion = {
+  v1: '1'
+};
+
+/** @public */
+export type ServerApiVersionId = typeof ServerApiVersion[keyof typeof ServerApiVersion];
+
+/** @public */
+export interface ServerApi {
```
emadum, line 52: The spec calls for this to be a class, but I didn't think it would be idiomatic in Node to make users import the ServerApi class and pass a `new ServerVersion(...)` in the constructor, versus accepting a plain object and providing a TypeScript interface. Does this make sense, or should it be a class like the spec advises?

src/mongo_client.ts
```diff
@@ -32,6 +32,32 @@ import type { SrvPoller } from './sdam/srv_polling';
...
+/** @public */
+export const LogLevel = {
+  error: 'error',
+  warn: 'warn',
+  info: 'info',
+  debug: 'debug'
+} as const;
+
+/** @public */
+export const ServerApiVersion = {
```
emadum, line 44: The spec calls for this to be an enum if the language supports it, but we avoid the use of TypeScript enums so I used the same pattern as `LogLevel` here, does that seem appropriate?

src/operations/command.ts
```diff
@@ -173,6 +179,18 @@ export abstract class CommandOperation<T> extends AbstractOperation<T> {
...
+    // if an API version was declared, add the apiVersion option to every command, except:
+    // a. only in the initial command of a transaction
+    // b. only in a Cursor's initiating command, not subsequent getMore commands
+    if (
+      server.serverApi &&
+      (!inTransaction || this.session.transaction.isStarting) &&
+      !cmd.commitTransaction &&
+      !cmd.getMore
+    ) {
+      applyServerApiVersion(cmd, server.serverApi);
```
emadum, line 191: Note that this catches most operations that need the version applied to them, but not all, so there are some additional calls in this PR to the `applyServerApiVersion` helper for commands that don't use this class.

src/cursor/abstract_cursor.ts
nbbeeken: Interesting, it could be, but its not obvious at first since it doesn't go through the client startSession method, its going directly to the topology one. If you can reproduce it can you take a look at:

https://github.com/mongodb/node-mongodb-native/blob/ee1a4d32ac95b7d143b08896bc486cfa8c2895a1/src/cursor/abstract_cursor.ts#L623-L625

This is where `this[kSession]` should be getting assigned so also printing the TopologyDescription with all the server descriptions as well could help

`util.inspect(cursor[kTopology].s.description, {depth: 6})` (6 usually goes plenty deep)

nbbeeken: Ok whoops, scrolled just a bit to see the changes you made to that section, that looks like the same issue to me, since you moved grabbing the session into the callback, then executeOperation has been called so server selection has run, now the function will return that sessions are supported. I am curious though what state the topology is in such that you're getting a failure, and then what state it transitions to after executeOperation runs.

src/mongo_client.ts
nbbeeken: Seems good to me. Although LogLevel is now defined in logger.ts maybe there was a rebase issue here?

src/mongo_client.ts
nbbeeken: I agree about it being idiomatic for nodejs
I guess just to consider an alternative we could make a class similar to our ReadPreference where we have static predefined instances of it? looking something like

```ts
new MongoClient('...', { serverApi: ServerApi.V1 })
```

Just putting this out there to say we considered it but I think avoiding an import is more satisfying for users.

src/mongo_client.ts
```diff
@@ -296,6 +325,7 @@ export class MongoClient extends EventEmitter {
       readConcern: this[kOptions].readConcern,
       writeConcern: this[kOptions].writeConcern,
       readPreference: this[kOptions].readPreference,
+      serverApi: this[kOptions].serverApi,
```
nbbeeken, line 328: [nit]: I don't see this being used anywhere, I think the getter on the class will serve to make it easy to access. I only mention it because I thought these `s` objects in general are something we want to get rid of, preferring symbol properties instead.

src/operations/update.ts
```diff
@@ -274,7 +274,7 @@ export function makeUpdateStatement(
     op.upsert = options.upsert;
   }
 
-  if (typeof options.multi === 'boolean') {
+  if (options.multi) {
```
emadum, line 277: Note: this should only be sent when true, since it's false by default - was causing some of the versioned api spec tests which weren't expecting `multi: false` to be there to fail.

src/mongo_client.ts
emadum: Oops yeah that's likely, fixed!

src/cmap/connection.ts
```diff
@@ -317,6 +320,23 @@ export class Connection extends EventEmitter {
 
     let clusterTime = this.clusterTime;
     let finalCmd = Object.assign({}, cmd);
+    const inTransaction = session && (session.inTransaction() || isTransactionCommand(finalCmd));
+
+    // if an API version was declared, add the apiVersion option to every command, except:
+    // a. only in the initial command of a transaction
+    // b. only in a Cursor's initiating command, not subsequent getMore commands
+    if (
+      this.serverApi &&
```
durran, line 329: What do you think about extracting this into a separate method? I tend to personally prefer doing that with complex checks as it makes this code block a bit easier to read.

src/connection_string.ts
```diff
@@ -572,6 +573,15 @@ export const OPTIONS = {
   autoEncryption: {
     type: 'record'
   },
+  serverApi: {
```
durran, line 576: Does this mean we are allowing the user to specify the api version in the connection string? The spec specifically states not to allow it as we [don't want developers changing behaviour without code changes](https://github.com/mongodb/specifications/blob/master/source/versioned-api/versioned-api.rst#no-uri-options).

src/mongo_client.ts
```diff
@@ -32,6 +32,21 @@ import type { SrvPoller } from './sdam/srv_polling';
...
+export const ServerApiVersion = {
+  v1: '1'
+};
+
+/** @public */
+export type ServerApiVersionId = typeof ServerApiVersion[keyof typeof ServerApiVersion];
+
+/** @public */
+export interface ServerApi {
+  version: string | ServerApiVersionId;
```
durran, line 45: Instead of freezing the `ServerApi` on the getter in `MongoClient`, could we instead just define the properties as `readonly` here? Or does that not give us enough protection somehow?

durran (changes requested):

nbbeeken (changes requested):

src/mongo_client.ts
emadum: I think `readonly` only applies at the level of the TypeScript compiler, so someone using plain JS would be able mutate the property. I'd be interested to hear your thoughts on this @nbbeeken

test/tools/runner/index.js
```diff
@@ -75,6 +77,11 @@ before(function (_done) {
         return;
       }
 
+      // Ensure test MongoClients set a serverApi parameter when required
+      if (MONGODB_API_VERSION) {
+        Object.assign(context, { serverApi: MONGODB_API_VERSION });
```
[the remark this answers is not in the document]
emadum: No that's a good nit, I'm not sure why I used `Object.assign` there, will update 👍

src/mongo_client.ts
nbbeeken: Indeed the `readonly` modifier only effects type checking. I think freezing is worth it here, since it is a construction time only setting, we can just jump ahead of any bugs trying to change serverApi at a later time by doing this.

src/index.ts
```diff
@@ -211,6 +211,9 @@ export type {
   Auth,
   DriverInfo,
   MongoOptions,
+  ServerApi,
```
nbbeeken, line 214: Ah just noticed, we don't have tooling to catch this unfortunately but api-extractor is satisfied when this is exported as a type, but if we want users to be able to use the enum it has to be exported normally, you can see on line 76 is where I grouped the "enums". 

Maybe a test/tooling we want to add (not here, just at some point) is importing index.js in our javascript tests and assert the actual imports we expect at runtime. 

(@durran just pinging you here for knowledge share since we all might run into this gap in tooling)

src/connection_string.ts
emadum: Ok, I added a check to throw an error if it is provided in the URI options, and a corresponding test.

src/mongo_client.ts
durran: Ok great, thanks!

nbbeeken (approved):
LGTM

durran (approved):
LGTM! Nice job!

PR #2738, opened 2021-02-12
test: add tests for azure and GCP CSFLE, fix aws boto3 error

emadum:
Ports #2662 and #2720 for 4.0

NODE-2825
NODE-3056


test/functional/client_side_encryption/prose.test.js
```diff
@@ -4,6 +4,18 @@ const chai = require('chai');
 const expect = chai.expect;
 chai.use(require('chai-subset'));
 
+const EJSON = require('mongodb-extjson');
```
nbbeeken, line 7: Can we use EJSON from bson here? and in doing so avoid bringing in the deprecated `mongodb-extjson` lib
`const { EJSON } = require('bson')`

.evergreen/run-tests.sh
```diff
@@ -53,7 +53,7 @@ if [[ -z "${CLIENT_ENCRYPTION}" ]]; then
   unset AWS_ACCESS_KEY_ID;
   unset AWS_SECRET_ACCESS_KEY;
 else
-  npm install mongodb-client-encryption
+  npm install mongodb-client-encryption@1.1.1-beta.0
```
nbbeeken, line 56: Is this to ensure we get in the (semi) recent changes to add GCP and Azure? I think we'd be better off using a range so we can get the latest as we make more releases: `npm install mongodb-client-encryption@">=1.2.1"`

nbbeeken (changes requested):
Testing LGTM, just some CI notes

.evergreen/run-tests.sh
emadum: That change came in from the 3.6 PR (#2662) so I'm not entirely sure, but agreed using a range is better.

nbbeeken: Ah my mistake, I missed that came from there we should correct it there too.

nbbeeken (approved):
LGTM

durran (approved):

PR #2739, opened 2021-02-16
fix: move session support check to operation layer

nbbeeken:
Monitor checks run on a timer and network errors cause a reset
of the topology that would be corrected in the next cycle of the
monitor, we were checking a property that gets updated by this async
work. By moving the check to the operations layer we will allow users
to obtain a session regardless of support and then emit an error if
there is not support when the session is used in an operation.

NODE-3100

### DIFF
--- a/lib/operations/execute_operation.js
+++ b/lib/operations/execute_operation.js
@@ -30,11 +30,25 @@ function executeOperation(topology, operation, callback) {
     throw new TypeError('This method requires a valid operation instance');
   }
 
-  if (isUnifiedTopology(topology) && topology.shouldCheckForSessionSupport()) {
-    return selectServerForSessionSupport(topology, operation, callback);
+  const Promise = topology.s.promiseLibrary;
+
+  let result;
+  if (typeof callback !== 'function') {
+    result = new Promise((resolve, reject) => {
+      callback = (err, res) => {
+        if (err) return reject(err);
+        resolve(res);
+      };
+    });
   }
 
-  const Promise = topology.s.promiseLibrary;
+  if (isUnifiedTopology(topology) && topology.shouldCheckForSessionSupport()) {
+    // Recursive call to executeOperation after a server selection
+    // shouldCheckForSessionSupport should return false after a server selection because
+    // there will be data bearing servers
+    selectServerForSessionSupport(topology, operation, callback);
+    return result;
+  }
 
   // The driver sessions spec mandates that we implicitly create sessions for operations
   // that are not explicitly provided with a session.
@@ -45,20 +59,14 @@ function executeOperation(topology, operation, callback) {
       session = topology.startSession({ owner });
       operation.session = session;
     } else if (operation.session.hasEnded) {
-      throw new MongoError('Use of expired sessions is not permitted');
+      callback(new MongoError('Use of expired sessions is not permitted'));
+      return result;
     }
-  } else if (operation.session && operation.session.explicit) {
-    throw new MongoError('Current topology does not support sessions');
-  }
-
-  let result;
-  if (typeof callback !== 'function') {
-    result = new Promise((resolve, reject) => {
-      callback = (err, res) => {
-        if (err) return reject(err);
-        resolve(res);
-      };
-    });
+  } else if (operation.session && !topology.hasSessionSupport()) {
+    // If the user passed an explicit session and we are still, after server selection,
+    // trying to run against a topology that doesn't support sessions we error out.
+    callback(new MongoError('Current topology does not support sessions'));
+    return result;
   }
 
   function executeCallback(err, result) {
@@ -78,7 +86,7 @@ function executeOperation(topology, operation, callback) {
     } else {
       operation.execute(executeCallback);
     }
-  } catch (e) {
+  } catch (error) {
     if (session && session.owner === owner) {
       session.endSession();
       if (operation.session === session) {
@@ -86,7 +94,7 @@ function executeOperation(topology, operation, callback) {
       }
     }
 
-    throw e;
+    callback(error);
   }
 
   return result;
@@ -141,11 +149,6 @@ function executeWithServerSelection(topology, operation, callback) {
       callback(err, null);
       return;
     }
-
-    if (serverSelectionOptions.session && topology.hasSessionSupport()) {
-      return callback(new MongoError('Current topology does not support sessions'));
-    }
-
     const shouldRetryReads =
       topology.s.options.retryReads !== false &&
       operation.session &&
@@ -165,28 +168,13 @@ function executeWithServerSelection(topology, operation, callback) {
 // TODO: This is only supported for unified topology, it should go away once
 //       we remove support for legacy topology types.
 function selectServerForSessionSupport(topology, operation, callback) {
-  const Promise = topology.s.promiseLibrary;
-
-  let result;
-  if (typeof callback !== 'function') {
-    result = new Promise((resolve, reject) => {
-      callback = (err, result) => {
-        if (err) return reject(err);
-        resolve(result);
-      };
-    });
-  }
-
   topology.selectServer(ReadPreference.primaryPreferred, err => {
     if (err) {
-      callback(err);
-      return;
+      return callback(err);
     }
 
     executeOperation(topology, operation, callback);
   });
-
-  return result;
 }
 
 module.exports = executeOperation;
--- a/lib/operations/execute_operation.js
+++ b/lib/operations/execute_operation.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const maybePromise = require('../utils').maybePromise;
 const MongoError = require('../core/error').MongoError;
 const Aspect = require('./operation').Aspect;
 const OperationBase = require('./operation').OperationBase;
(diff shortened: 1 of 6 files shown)


lib/operations/execute_operation.js
```diff
@@ -47,6 +47,8 @@ function executeOperation(topology, operation, callback) {
     } else if (operation.session.hasEnded) {
       throw new MongoError('Use of expired sessions is not permitted');
     }
+  } else if (operation.session && operation.session.explicit) {
```
mbroadst, line 50: I think a little more work is required here. The goal of the change is to defer the work of determining session support until a server is selected, that should apply to implicit and explicit sessions - so the check a few lines above this for `hasSessionsSupport` still suffers the same issue as the originally reported issue: it's a synchronous check of an asynchronously updated state. 

The refactor might be a little more involved, here are some thoughts:
- can we remove the code to distinguish between `EXECUTE_WITH_SELECTION` and not? I think this work was done in the 4.0 branch, maybe its not possible in 3.6?
- the `executeWithServerSelection` will likely need to be pulled into `executeOperation`, because the session check should happen each time a server is selected (if a retry is required, for instance).

mbroadst (changes requested):

lib/operations/execute_operation.js
nbbeeken: Actually this issue only impacts the UnifiedTopology so this was the right spot to move this for the legacy topology I believe, but a few lines above (line 33) there's a `selectServerForSessionSupport` call that will be made. So I think yes I should move the check into the callback of that selection, but also that selection will fail if it can't find a server that supports sessions. So if we want to report this error properly for the unified Topology we'd need to see if we fail selection specifically because of no session support, does that make sense?

mbroadst: Hm, if we believe that this issue is caused by synchronously checking for sessions support in `startSession` then this problem must exist for both unified and legacy topologies - perhaps it's just easier to surface in the unified because it is stricter about when and how operations are run.  I forgot we added that `selectServerForSessionSupport`, this would also seem to imply there's a bug there since that should theoretically prevent this issue from happening. I think you have the same problem here anyway, `shouldCheckForSessionSupport` is a sync check on data which may change as a result of async operations. 

What I'm proposing here is _always_ deferring the decision about sessions until the moment a server is selected for an operation. Once the server is selected, and a connection is checked out from the pool for that server, you are at a point-in-time where you can trust all the information at hand. It's here that you can say "did you provide a session? well sessions aren't supported on this connection, sorry", or "sessions are supported, but none was provided, so we should use an implicit session". The problem with the way `executeOperation` is written today is that we try to make that determination before any selection occurs at all, and thus might miss out on updated information gained _in the process of selecting a server_.

nbbeeken:
Despite what CI looks like at the time of commenting I don't believe any of those failures are related to this fix proposed here.
I've made #2747 to clean up a majority of the CI issues.

lib/operations/execute_operation.js
```diff
@@ -45,18 +59,14 @@ function executeOperation(topology, operation, callback) {
...
-
-  let result;
-  if (typeof callback !== 'function') {
-    result = new Promise((resolve, reject) => {
-      callback = (err, res) => {
-        if (err) return reject(err);
-        resolve(res);
-      };
-    });
+  } else if (operation.session && !topology.hasSessionSupport()) {
```
mbroadst, line 65: another nit: I would pull this out into a very explicit block right below the `selectServerForSessionSupport` call, just to make it very clear that we are looking for this particular case:

```
  if (isUnifiedTopology(topology) && topology.shouldCheckForSessionSupport()) {
    selectServerForSessionSupport(topology, operation, callback);
    return result;
  }
 
  if (operation.session && !topology.hasSessionSupport()) {
    callback(new MongoError('Current topology does not support sessions'));
    return result;
  }

  .. else everything goes on as expected
```

lib/operations/execute_operation.js
```diff
@@ -30,11 +30,25 @@ function executeOperation(topology, operation, callback) {
     throw new TypeError('This method requires a valid operation instance');
   }
 
-  if (isUnifiedTopology(topology) && topology.shouldCheckForSessionSupport()) {
-    return selectServerForSessionSupport(topology, operation, callback);
+  const Promise = topology.s.promiseLibrary;
```
mbroadst, line 33: If we're going to do this refactor should we just use `maybePromise`?

lib/operations/execute_operation.js
```diff
@@ -159,28 +168,13 @@ function executeWithServerSelection(topology, operation, callback) {
 // TODO: This is only supported for unified topology, it should go away once
```
mbroadst, line 168: Can you update the language of this note to better explain why this method is necessary?

test/unit/sessions/client.test.js
```diff
@@ -27,21 +27,32 @@ describe('Sessions', function() {
...
+          .connect()
+          .then(function(client) {
+            const session = client.startSession();
+            return client
+              .db()
+              .collection('t')
+              .insertOne({ a: 1 }, { session });
+          })
+          .then(() => {
+            expect.fail('Expected an error to be thrown about not supporting sessions');
```
mbroadst, line 40: I always found it cumbersome with promises that you had to add an extra assertion for the case you weren't checking. Did you convert this test to promises because of personal preference, or was there some other reason?

lib/operations/execute_operation.js
nbbeeken: Updated, let me know if there's something specific you think we should have here.

lib/operations/execute_operation.js
```diff
@@ -30,11 +30,25 @@ function executeOperation(topology, operation, callback) {
...
+      callback = (err, res) => {
+        if (err) return reject(err);
+        resolve(res);
+      };
+    });
   }
 
-  const Promise = topology.s.promiseLibrary;
+  if (isUnifiedTopology(topology) && topology.shouldCheckForSessionSupport()) {
+    // Recursive call to executeOperation after a server selection
```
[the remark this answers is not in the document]
nbbeeken: I removed a superfluous line. I left in a note that its recursive, seems worthy of calling out.

lib/operations/execute_operation.js
nbbeeken: I cleaned up the `else if` condition to make it a bit easier to read the control flow, I think that might make it more quickly understandable what is being checked here.

lib/operations/execute_operation.js
```diff
@@ -30,64 +31,57 @@ function executeOperation(topology, operation, callback) {
...
-  let session, owner;
-  if (topology.hasSessionSupport()) {
-    if (operation.session == null) {
-      owner = Symbol();
-      session = topology.startSession({ owner });
-      operation.session = session;
-    } else if (operation.session.hasEnded) {
-      throw new MongoError('Use of expired sessions is not permitted');
+  return maybePromise(topology, cb, callback => {
+    if (isUnifiedTopology(topology) && topology.shouldCheckForSessionSupport()) {
```
mbroadst, line 35: sorry to be so nitpicky, but I prefer what we do in the [4.0 branch](https://github.com/mongodb/node-mongodb-native/blob/4.0/src/operations/execute_operation.ts#L64-L75) where there's a `maybePromise` for each of these cases. The way this is written right now, you might end up nesting promises, whereas in 4.0 if you haven't provided a callback we are using the callback from the top `maybePromise` to pass into the secondary `executeOperation`.

nbbeeken: (No worries, please nitpick away!)
I don't see how you can nest promises here since `selectServerForSessionSupport` will get passed a defined callback from `maybePromise` and so the next call to executeOperation will have that callback defined so the "second" `maybePromise` won't create a new Promise.

Something to consider is that this provides one point of exit to executeOperation. Entirely an 🧅  but that pattern is preferable to what is on 4.0. It's a strange pattern to me to have more code run "synchronously" in the remainder of the function after you've returned a function that appears to do an async task. (In this instance, I know maybePromise is all sync it's just wrapping a bit of setup work, but my style point remains). I'm glossing over the callback detail by just returning but we get the idea. 
```js
function runOp() {
  if (cond)
    return doThing(() => {
      // now we've done something
      return 'one way'
    })

  // do more things
  return doThing(() => {
      // now we've done something
      return 'the other way'
  })
}
```
vs.
```js
function runOp() {
  return doThing(() => {
    if (cond) return 'one way'
    
    // do more things
    return 'the other way'
  })
}
```

mbroadst: Hm, you're right there's no Promise nesting. Maybe my ultimate concern is just about the multiple `maybePromise` calls in the `executeOperation` stacktrace. A call to `insertOne` right now will look like:

mbroadst (approved):

emadum (approved):
LGTM 🚀

lib/operations/execute_operation.js
nbbeeken: Yea we'll need to make a port of this sessions related fix to 4.0 so I can do that when I'm there (If anyone else reviewing has a differing opinion please feel free to chime in!)

durran (approved):

PR #2740, opened 2021-02-16
refactor!: remove deprecated items

nbbeeken:
Removals are noted in JIRA ticket.

NODE-2317

Some of the deprecations are not to be removed (e.g, Collection#insert) but a warning should still be emitted I wrote a small wrapper around process.emitWarning that we should also port to 3.6. This allows us to drop the deprecate calls that come after the class definition (e.g, `deprecate(Collection.prototype.insert, ...)`) and move a `emitWarningOnce` call into the method itself.


src/cmap/commands.ts
```diff
@@ -826,7 +827,7 @@ export class BinMsg {
     while (this.index < this.data.length) {
       const payloadType = this.data.readUInt8(this.index++);
       if (payloadType === 1) {
-        console.error('TYPE 1');
+        throw new MongoError('Type 1 payload'); // ???
```
emadum, line 830: Nice find, good call restricting the console functions in the linter. Can we leave a FIXME or TODO comment here instead of question marks? Or maybe it's even worth filing a JIRA ticket to investigate this?

src/operations/add_user.ts
```diff
@@ -56,25 +61,22 @@ export class AddUserOperation extends CommandOperation<Document> {
...
-    // Check the db name and add roles if needed
-    if (
-      (db.databaseName.toLowerCase() === 'admin' || options.dbName === 'admin') &&
-      !Array.isArray(options.roles)
-    ) {
-      roles = ['root'];
-    } else if (!Array.isArray(options.roles)) {
-      roles = ['dbOwner'];
+    if (roles.length === 0) {
+      // If not roles defined throw error
```
emadum, line 79:
```suggestion
      // If no roles defined throw error
```

src/utils.ts
```diff
@@ -1374,3 +1369,35 @@ export const DEFAULT_PK_FACTORY = {
...
+}
+
+const emittedWarnings = new Set();
+/**
+ * Will emit a warning once for the duration of the application.
+ * Uses the message to identify if it has already been emitted
+ * so using string interpolation can cause multiple emits
+ * @internal
+ */
+export function emitWarningOnce(message: string): void {
```
emadum, line 1398: Nice, this was exactly what I was thinking we should do 🥇

emadum (approved):
LGTM, very nice to see all that legacy code removed! 👍

src/operations/add_user.ts
nbbeeken: Actually this if statement does nothing with the added logic above it, so whoosh, it is gone! 🎉

src/cmap/commands.ts
nbbeeken: Sync-ed in the channel on this, left a comment describing this is an used part of the protocol, and made the error message a bit more clear.

durran (approved):
👍

nbbeeken:
@addaleax These are the deprecations in 3.6 that are slated for removal in 4.0, there are failures with the mongosh tests. If you (or a colleague) get the chance, could you take a look and give us some insight into what is expected/unexpected here.

addaleax:
> If you (or a colleague) get the chance, could you take a look and give us some insight into what is expected/unexpected here.

It's failing during typescript compilation in both cases (I know the test logs can be a bit messy -- it's generally helpful to grep around for the first `npm ERR!` if you're looking at them to find the source of the problem(s)):

```
 [2021/02/19 18:27:59.938] TSError: ⨯ Unable to compile TypeScript:
 [2021/02/19 18:27:59.938] src/cli-service-provider.spec.ts(295,22): error TS2339: Property 'findAndModify' does not exist on type 'StubbedInstance<Collection>'.
 [2021/02/19 18:27:59.938] src/cli-service-provider.spec.ts(308,29): error TS2339: Property 'findAndModify' does not exist on type 'StubbedInstance<Collection>'.
```

and for #2741 

```
 [2021/02/19 18:42:58.456] src/shell-internal-state.ts(223,29): error TS2749: 'TopologyType' refers to a value, but is being used as a type here. Did you mean 'typeof TopologyType'?
```

The latter is probably not hard to address, you can merge it and we'll probably really just have to add `typeof` in a places where we're referring to the type of what-was-once-an-enum.

For the `findAndModify` one, I don't think we were aware that it was deprecated. It's exposed as a shell method, so we'll need to find a way of implementing it in terms of the methods that replace it, which doesn't seem hard to me, but I haven't tried it yet so I'm not sure.

And thanks for the ping! :)

nbbeeken:
> For the `findAndModify` one, I don't think we were aware that it was deprecated. It's exposed as a shell method, so we'll need to find a way of implementing it in terms of the methods that replace it, which doesn't seem hard to me, but I haven't tried it yet so I'm not sure.

Thanks for going through the logs! Sounds like we're good to go, of course let us know if we can advise/help out with the `findAndModify` patch. There are three options that can be passed to findAndModify:

- update?: boolean;
- remove?: boolean;
- upsert?: boolean;

 and that's what makes it do the same thing as findOneAndUpdate, findOneAndDelete, findOneAndReplace.

PR #2741, opened 2021-02-16
refactor: remove enums in favor of const objects

nbbeeken:
Changes remaining enums to const objects and exports all user relevant
"enums" from index.ts. Enums are frozen to ensure they remain static.
Added a lint rule to prevent further enums.
Corrected mistake in sourcemap emit settings.

NODE-2973

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,7 +32,15 @@
     "strict": ["error", "global"],
     "promise/no-native": "error",
 
-    "@typescript-eslint/no-explicit-any": "off"
+    "@typescript-eslint/no-explicit-any": "off",
+
+    "no-restricted-syntax": [
+      "error",
+      {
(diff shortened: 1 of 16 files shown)


emadum (approved):
LGTM. I recall there being some discussion about enums, but can you remind me what motivated this change?

nbbeeken:
There were a few reasons, one was `const enums` can't be used by users since they disappear at compile time, another was that an enum forces you to use the enum even if you pass in an equivalent value.
```ts
enum Weekday { Monday = 1, ... };
function printDay(day: Weekday);
```
Our union of the possible values pattern would permit you to pass in `1` directly for example. This comes up when we have a variable that is of type number and we've done a check to ensure its one of the acceptable values but there's still a type error unless you cast etc.

durran (approved):

PR #2742, opened 2021-02-17
feat(NODE-3011): Load Balancer Support

durran:
This add support for load balancing mode in the driver.


.evergreen/config.yml
```diff
@@ -131,6 +131,40 @@ functions:
           MONGODB_API_VERSION="${MONGODB_API_VERSION}" \
           NODE_VERSION=${NODE_VERSION} SKIP_DEPS=${SKIP_DEPS|1} NO_EXIT=${NO_EXIT|1} \
             bash ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
+  start-load-balancer:
```
durran, line 134: Starts HAProxy using the drivers-evergreen-tools script.

.evergreen/config.yml
```diff
@@ -131,6 +131,40 @@ functions:
...
+    - command: shell.exec
+      params:
+        script: |
+          DRIVERS_TOOLS=${DRIVERS_TOOLS} MONGODB_URI=${MONGODB_URI} bash ${DRIVERS_TOOLS}/.evergreen/run-load-balancer.sh start
+    - command: expansions.update
+      params:
+        file: lb-expansion.yml
+  stop-load-balancer:
```
durran, line 142: Stops HAProxy using the drivers-evergreen-tools script.

.evergreen/config.yml
```diff
@@ -131,6 +131,40 @@ functions:
...
+    - command: shell.exec
+      params:
+        script: |
+          DRIVERS_TOOLS=${DRIVERS_TOOLS} MONGODB_URI=${MONGODB_URI} bash ${DRIVERS_TOOLS}/.evergreen/run-load-balancer.sh stop
+  run-lb-tests:
```
durran, line 147: Runs the load balancer tests which are a subset of the entire test suite.

.evergreen/config.yml
```diff
@@ -542,6 +576,19 @@ post:
 ignore:
   - '*.md'
 tasks:
+  - name: test-load-balancer
```
durran, line 579: Testing load balancer mode starts HAProxy, starts a sharded cluster, runs the lb tests, then stops HAProxy

.evergreen/config.yml
```diff
@@ -1634,6 +1681,14 @@ buildvariants:
       NODE_LTS_NAME: fermium
       CLIENT_ENCRYPTION: true
     tasks: *ref_3
+  - name: ubuntu-18.04-fermium-lb
```
durran, line 1684: HAProxy is only installed Ubuntu 18 & 20.

.evergreen/run-lb-tests.sh
```diff
@@ -0,0 +1,59 @@
...
+#       MONGODB_URI             Set the suggested connection MONGODB_URI (including credentials and topology info)
+#       MARCH                   Machine Architecture. Defaults to lowercase uname -m
+#       TEST_NPM_SCRIPT         Script to npm run. Defaults to "check:test"
+#       SKIP_DEPS               Skip installing dependencies
+#       NO_EXIT                 Don't exit early from tests that leak resources
+
+AUTH=${AUTH:-noauth}
+UNIFIED=${UNIFIED:-}
+MONGODB_URI=${MONGODB_URI:-}
+TEST_NPM_SCRIPT="test:load-balancer"
```
durran, line 18: This is the new script for the lb tests in the package.json

.evergreen/run-load-balancer.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
```
durran, line 1: This script comes from https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/run-load-balancer.sh

src/cmap/connect.ts
```diff
@@ -125,6 +125,15 @@ function performInitialHandshake(
         return;
       }
 
+      if (options.loadBalanced && !response.serviceId) {
```
durran, line 128: Throw an error if in LB mode and no serviceId in the handshake response per the spec.

src/cmap/connection.ts
```diff
@@ -679,8 +709,10 @@ export class CryptoConnection extends Connection {
   }
 }
 
-function hasSessionSupport(conn: Connection) {
-  return conn.description.logicalSessionTimeoutMinutes != null;
+/** @public */
+export function hasSessionSupport(conn: Connection): boolean {
```
durran, line 713: LB mode has no monitor, so no `logicalSessionTimeoutMinutes`. LB mode always has session support.

src/cmap/connection_pool.ts
```diff
@@ -99,12 +107,18 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
    * @internal
    */
   [kGeneration]: number;
+  /** A map of generations to service ids
+   * @internal
+   */
+  [kServiceGenerations]: Map<ObjectId, number>;
```
durran, line 113: The generation in LB mode is a map of `serviceId` to generation number.

src/cmap/connection_pool.ts
```diff
@@ -219,6 +235,31 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+
+  markUnpinned(connection: Connection, pinType: string): void {
+    connection.markUnpinned(pinType);
+    this[kMetrics].markUnpinned(pinType);
+  }
+
+  /**
+   * Get the metrics information for the pool when a wait queue timeout occurs.
+   */
+  waitQueueErrorMetrics(): string {
```
durran, line 259: Wait queue timeouts in LB mode have a more detailed error message on pinned connections.

src/cmap/connection_pool.ts
```diff
@@ -289,9 +335,19 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
    * Pool reset is handled by incrementing the pool's generation count. Any existing connection of a
    * previous generation will eventually be pruned during subsequent checkouts.
    */
-  clear(): void {
-    this[kGeneration] += 1;
-    this.emit('connectionPoolCleared', new ConnectionPoolClearedEvent(this));
+  clear(serviceId?: ObjectId): void {
+    if (this.loadBalanced && serviceId) {
```
durran, line 339: In LB mode we increment the generation for the specific `serviceId`

test/functional/unified-spec-runner/runner.ts
```diff
@@ -41,7 +46,27 @@ export async function runUnifiedTest(
...
+      utilClient = ctx.configuration.newClient(MULTI_MONGOS_LB_URI);
+    } else {
+      utilClient = ctx.configuration.newClient(SINGLE_MONGOS_LB_URI);
+    }
+  } else {
+    utilClient = ctx.configuration.newClient();
+  }
 
   let entities;
   try {
```
nbbeeken, line 75: I attempted running the tests locally here's a summary of the issues I ran into:

- When I try to get the tests running I run into the serviceId missing error, we need a way to mock the serviceId on ismaster responses, I think other driver's have been using some global flag as a hack.
- I hardcoded the serviceId and the load-balanced topoplogy is missing from the `topologySatisfies` check in `test/functional/unified-spec-runner/unified-utils.ts`
- The `createFindCursor` operation is missing, We are going to need to support the changes made here: https://github.com/mongodb/specifications/commit/ed4e62b48939378b814cbd7d9d4939a2c1076a6b

.evergreen/run-lb-tests.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# only run FLE tets on hosts we explicitly choose to test on
+if [[ -z "${CLIENT_ENCRYPTION}" ]]; then
+  unset AWS_ACCESS_KEY_ID;
+  unset AWS_SECRET_ACCESS_KEY;
+else
+  npm install mongodb-client-encryption@">=1.2.1"
+fi
+
+SINGLE_MONGOS_LB_URI=${SINGLE_MONGOS_LB_URI} MULTI_MONGOS_LB_URI=${MULTI_MONGOS_LB_URI} MONGODB_API_VERSION=${MONGODB_API_VERSION} MONGODB_UNIFIED_TOPOLOGY=${UNIFIED} MONGODB_URI=${MONGODB_URI} npm run ${TEST_NPM_SCRIPT}
```
nbbeeken, line 59: There isn't a precedent for this elsewhere but I think we should assert something about what tests are run, when the command above exits we do have the test results in an xunit.xml file maybe we can add some grepping to make sure we're running at least one test. I'm just thinking that our skipping logic can be agressive if all the booleans don't line up just right. 

`$readPreference is sent for load-balanced clusters` in `test/spec/load-balancers/server-selection.yml` looks really small so it could just be a good litmus to grep for in the results.

nbbeeken (changes requested):
Ok! first round in! I'm running into trouble running the tests locally, and there seems to be some missing unified runner changes that we need here.

.evergreen/run-load-balancer.sh
```diff
@@ -0,0 +1,83 @@
...
+
+set -o errexit  # Exit the script with error if any of the commands fail
+
+start() {
+  echo "Starting HAProxy..."
+
+  cat <<EOF_HAPROXY_CONFIG >> $DRIVERS_TOOLS/haproxy.conf
```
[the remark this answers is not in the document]
nbbeeken: Oh looks like we can actually resync this script, its been updated:
https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/run-load-balancer.sh

.evergreen/run-load-balancer.sh
```diff
@@ -0,0 +1,83 @@
...
+    MULTI_MONGOS_LB_URI="${MULTI_MONGOS_LB_URI}?loadBalanced=true"
+  else
+    # If there are query params then append the load balanced option to them.
+    SINGLE_MONGOS_LB_URI="${SINGLE_MONGOS_LB_URI}${SUFFIX}&loadBalanced=true"
+    MULTI_MONGOS_LB_URI="${MULTI_MONGOS_LB_URI}${SUFFIX}&loadBalanced=true"
+  fi
+
+  echo "Single Mongos LB: $SINGLE_MONGOS_LB_URI"
+  echo "Multiple Mongos LB: $MULTI_MONGOS_LB_URI"
+
```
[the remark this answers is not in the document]
nbbeeken: also can be resolved by script sync

src/cmap/connection_pool.ts
addaleax: @durran Is it intentional that this is a `Map<ObjectId, ...>`? I think this might end up being surprising when working the code here considering that Maps work based on object identity, not equality – maybe a `Map<string, number>` would be better?

src/sdam/server.ts
```diff
@@ -234,7 +257,9 @@ export class Server extends TypedEventEmitter<ServerEvents> {
    * this will be a no-op.
    */
   requestCheck(): void {
-    this[kMonitor].requestCheck();
+    if (!this.s.options.loadBalanced) {
```
addaleax, line 260: Very minor thing, but I guess for consistency with the diff immediately above, maybe this could be 
```suggestion
    if (!this.loadBalanced) {
```

throughout the file?

src/sdam/common.ts
```diff
@@ -36,7 +37,8 @@ export const ServerType = Object.freeze({
   RSArbiter: 'RSArbiter',
   RSOther: 'RSOther',
   RSGhost: 'RSGhost',
-  Unknown: 'Unknown'
+  Unknown: 'Unknown',
+  LoadBalancer: 'LoadBalancer'
```
addaleax, line 41: Are these names fixed and can we rely on them? I didn’t realize it so far, but since we make decisions based on the ServerType/TopologyType in mongosh, we’ll need to update that to account for the new values

src/cmap/connection_pool.ts
durran: Yes I think I can change this to the hex string representation of the object id in this case.

.evergreen/run-load-balancer.sh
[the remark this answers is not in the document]
durran: I've updated to the latest script now.

test/tools/mock.js
```diff
@@ -7,8 +7,14 @@ const {
   DEFAULT_ISMASTER_36
   // eslint-disable-next-line no-restricted-modules
 } = require('mongodb-mock-server');
+const { ObjectId } = require('bson');
 const { HostAddress } = require('../../src/utils');
 
+const DEFAULT_HELLO_50 = {
+  ...DEFAULT_ISMASTER_36,
+  serverId: new ObjectId()
```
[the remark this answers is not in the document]
durran: I will remove this as the mocked `serviceId` is actually handled in `src/cmap/connect.ts` during the handshake based on an env var. I've left a todo to remove as soon as the server has LB support.

test/functional/unified-spec-runner/runner.ts
```diff
@@ -176,3 +202,14 @@ export function runUnifiedSuite(specTests: uni.UnifiedSuite[], testsToSkip?: str
...
 }
+
+/**
+ * Determine if the URI is multi-host. Our test suites
+ * use simple user/password combinations so we take the
+ * simple route here and just look for commas between the
+ * scheme and db.
+ */
+function isMultiHostUri(uri: string): boolean {
+  const plainUri = uri.substring(str.lastIndexOf('://') + 1, str.lastIndexOf('/'));
```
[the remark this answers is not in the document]
durran: I've completely removed this now, and moved per the suggestion into the `TestConfiguration` class.

test/functional/unified-spec-runner/runner.ts
```diff
@@ -41,7 +46,27 @@ export async function runUnifiedTest(
...
+    // frontends.
+    expect(SINGLE_MONGOS_LB_URI).to.exist;
+    expect(MULTI_MONGOS_LB_URI).to.exist;
+
+    // We determine which URI the load balancer to point at
+    // by looking at the MONGODB_URI - if it has 1 mongos
+    // then it's the single mongos frontend and if it has
+    // multiple mongoses then we use the multi mongos
+    // frontend.
+    if (isMultiHostUri(process.env.MONGODB_URI)) {
```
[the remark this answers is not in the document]
durran: I've also removed this here now to use your suggestion and handle `useMultipleMongoses` in a special way when in LB mode.

test/functional/unified-spec-runner/runner.ts
```diff
@@ -15,6 +15,11 @@ interface MongoDBMochaTestContext extends Mocha.Context {
   configuration: TestConfiguration;
 }
 
+// Load balancer fronting 1 mongos.
+const SINGLE_MONGOS_LB_URI = process.env.SINGLE_MONGOS_LB_URI;
+// Load balancer fronting 2 mongoses.
+const MULTI_MONGOS_LB_URI = process.env.MULTI_MONGOS_LB_URI;
```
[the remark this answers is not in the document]
durran: Actually moved into `test/toos/runner/index.js` where the other env vars for uri and versioned api are handled.

test/functional/unified-spec-runner/runner.ts
durran: `createFindCursor` is now brought back as well as the `iterateUntilDocumentOrError` changes to handle both change streams and find cursors. `close` is also back as an operation as well as `assertNumberConnectionsCheckedOut`. `ignoreResultAndError` assertion is also brought back.

src/sdam/server.ts
durran: Thanks for catching that. I've updated.

src/sdam/common.ts
durran: Yes, per the spec the `TopologyType` is always `"LoadBalanced"` and the `ServerType` is always `"LoadBalancer"`

durran: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologytype

src/connection_string.ts
```diff
@@ -98,14 +109,15 @@ export function resolveSRVRecord(options: MongoOptions, callback: Callback<HostA
 
         const txtRecordOptions = new URLSearchParams(record[0].join(''));
         const txtRecordOptionKeys = [...txtRecordOptions.keys()];
-        if (txtRecordOptionKeys.some(key => key !== 'authSource' && key !== 'replicaSet')) {
+        if (txtRecordOptionKeys.some(key => !VALID_TXT_RECORDS.includes(key))) {
           return callback(
-            new MongoParseError('Text record must only set `authSource` or `replicaSet`')
+            new MongoParseError(`Text record must only set one of: ${VALID_TXT_RECORDS.join(', ')}`)
```
addaleax, line 114:
```suggestion
            new MongoParseError(`Text record may only set any of: ${VALID_TXT_RECORDS.join(', ')}`)
```

(Otherwise this sounds like it's saying that at most one option can be set -- feel free to adjust the text, of course, or feel free to ignore, as obviously this is a fairly minor thing, I just noticed it :))

.evergreen/run-load-balancer.sh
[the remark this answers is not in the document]
durran: Script synced.

src/connection_string.ts
durran: Updated to suggestion.

src/cmap/connection_pool.ts
```diff
@@ -289,9 +345,19 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+      const generation = this.serviceGenerations.get(serviceId);
```
mbroadst, line 350: what if `!has(serviceId)`?

src/cmap/connection_pool.ts
```diff
@@ -289,9 +345,19 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+      // Only need to worry if the generation exists, since it should
+      // always be there but typescript needs the check.
+      if (generation) {
+        // Increment the generation for the service id.
+        this.serviceGenerations.set(serviceId, generation + 1);
+      }
```
mbroadst, line 356: what if `generation == null`? Maybe we should throw an error, since that state should be impossible

src/cmap/connection_pool.ts
```diff
@@ -454,6 +534,12 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
   });
 }
 
+function createServiceGeneration(pool: ConnectionPool, serviceId: ObjectId) {
```
mbroadst, line 537: nit: do we use this elsewhere? If its only in one place I suggest inlining

src/cmap/connection_pool_events.ts
```diff
@@ -90,12 +91,14 @@ export class ConnectionClosedEvent extends ConnectionPoolMonitoringEvent {
   connectionId: number | '<monitor>';
   /** The reason the connection was closed */
   reason: string;
+  serviceId?: ObjectId;
 
   /** @internal */
-  constructor(pool: ConnectionPool, connection: Connection, reason: string) {
+  constructor(pool: ConnectionPool, connection: Connection, reason: string, serviceId?: ObjectId) {
```
mbroadst, line 97: Similarly to the connection events, we have a `serviceId` mounted on `Connection` so no need to add the additional argument

src/cmap/connection_pool_events.ts
```diff
@@ -166,7 +169,15 @@ export class ConnectionCheckedInEvent extends ConnectionPoolMonitoringEvent {
...
-  constructor(pool: ConnectionPool) {
+  serviceId?: ObjectId;
+
+  /** @internal */
+  constructor(pool: ConnectionPool, serviceId?: ObjectId) {
     super(pool);
+    this.serviceId = serviceId;
+  }
+
+  get hasServiceId(): boolean {
```
mbroadst, line 180: same comment as connection events

src/cmap/errors.ts
```diff
@@ -25,9 +24,9 @@ export class WaitQueueTimeoutError extends MongoDriverError {
   /** The address of the connection pool */
   address: string;
 
-  constructor(pool: Connection | ConnectionPool) {
-    super('Timed out while checking out a connection from connection pool');
+  constructor(message: string, address: string) {
+    super(`Timed out while checking out a connection from connection pool${message}`);
```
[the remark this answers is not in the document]
mbroadst: maybe add a ":" between "pool" and "${message}"?

src/connection_string.ts
```diff
@@ -33,6 +33,12 @@ import { Logger, LoggerLevel } from './logger';
 import { PromiseProvider } from './promise_provider';
 import { Encrypter } from './encrypter';
 
+const VALID_TXT_RECORDS = ['authSource', 'replicaSet', 'loadBalanced'];
+
+const LB_SINGLE_HOST = 'loadBalanced option only supported with a single host in the URI';
```
mbroadst, line 38: nit: add a suffix indicating this is an error string (maybe `LB_SINGLE_HOST_ERROR`), also consider inlining since it doesn't seem to be used in multiple places

src/connection_string.ts
```diff
@@ -118,13 +130,49 @@ export function resolveSRVRecord(options: MongoOptions, callback: Callback<HostA
...
 
       callback(undefined, hostAddresses);
     });
   });
 }
 
+function checkLoadBalancerOptions(
+  options: MongoOptions,
+  addresses: HostAddress[]
+): MongoParseError | null {
```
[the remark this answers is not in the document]
mbroadst: +1

mbroadst (changes requested):
This is a first batch of comments before digging too deeply into the SDAM changes. I'll leave a second batch later today/tomorrow morning

src/cmap/connection_pool.ts
```diff
@@ -383,6 +450,50 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       });
     });
   }
+
+  /**
+   * Runs a lambda with an implicitly checked out connection, checking that connection back in when the lambda
+   * has completed by calling back.
```
addaleax, line 456: This is looking like the `withConnection` description above, and maybe I’m missing something, but where’s the `checkIn()` call here?

durran: I updated the comments now to reflect what this method really does.

src/connection_string.ts
durran: Updated.

src/cmap/connection.ts
```diff
@@ -246,10 +247,22 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+
+  get loadBalanced(): boolean | undefined {
+    return this.description.loadBalanced;
+  }
+
   get generation(): number {
     return this[kGeneration] || 0;
   }
 
+  set generation(generation: number) {
```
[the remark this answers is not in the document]
durran: That's a tricky one. In order to maintain that, we need to know the generation when the connection is instantiated. In order to know that, we need to know the `serviceId` - which requires us to perform the initial handshake.

src/cmap/connection_pool.ts
```diff
@@ -219,6 +240,35 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+  get loadBalanced(): boolean {
+    return this.options.loadBalanced;
+  }
+
+  get serviceGenerations(): Map<ObjectId, number> {
+    return this[kServiceGenerations];
+  }
+
+  get currentCheckedOutCount(): number {
+    return this[kCheckedOut];
```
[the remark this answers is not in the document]
durran: Hmm. This didn't actually work for me. I can see the connections getting checked out and back in but this number is always 1 higher than it should be.

src/cmap/connection.ts
```diff
@@ -246,10 +247,22 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+    this[kGeneration] = generation;
```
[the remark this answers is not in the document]
durran: As per my other comment - we don't know the `serviceId` when we instantiate the `Connection` - we only know that after the handshake so the `Connection` has already been instantiated.

src/cmap/connect.ts
```diff
@@ -310,7 +333,7 @@ function makeConnection(options: ConnectionOptions, _callback: CallbackWithType<
   socket.setNoDelay(noDelay);
 
   const connectEvent = useTLS ? 'secureConnect' : 'connect';
-  let cancellationHandler: (err: Error) => void;
+  let cancellationHandler: (err: Error, serviceId?: ObjectId) => void;
```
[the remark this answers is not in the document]
durran: Removed.

src/cmap/connection_pool.ts
durran: Added throw of `MongoDriverError` in this case.

src/cmap/connection_pool.ts
durran: Based on other update an exception will now be thrown in this case if the `serviceId` is missing.

src/cmap/connection_pool_events.ts
durran: This is now pulled off the `Connection`

src/cmap/connection.ts
```diff
@@ -246,10 +247,22 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this[kIsMaster] = response;
   }
 
+  get serviceId(): ObjectId | undefined {
+    return this.ismaster?.serviceId;
+  }
+
+  get loadBalanced(): boolean | undefined {
```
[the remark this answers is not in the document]
durran: Removed `undefined` type

src/cmap/command_monitoring_events.ts
```diff
@@ -83,6 +95,11 @@ export class CommandSucceededEvent {
     this.commandName = commandName;
     this.duration = calculateDurationInMs(started);
     this.reply = maybeRedact(commandName, cmd, extractReply(command, reply));
+    this.serviceId = serviceId;
+  }
+
+  get hasServiceId(): boolean {
```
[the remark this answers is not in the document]
durran: The reason for adding these was for the unified tests. They are asserting on each event `hasServiceId` and the runner basically matches based on checking for the property generically. Maybe I just add `/* @internal */` so the API extractor removes them?

src/cmap/command_monitoring_events.ts
```diff
@@ -25,7 +26,11 @@ export class CommandStartedEvent {
    * @param pool - the pool that originated the command
    * @param command - the command
    */
-  constructor(pool: Connection | ConnectionPool, command: WriteProtocolMessageType) {
+  constructor(
+    pool: Connection | ConnectionPool,
+    command: WriteProtocolMessageType,
+    serviceId?: ObjectId
```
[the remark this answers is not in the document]
durran: I've refactored this now.

durran:
https://github.com/mongodb/node-mongodb-native/pull/2742#discussion_r665655687

@mbroadst I've now refactored all the requested build changes.

.evergreen/config.yml
```diff
@@ -131,6 +131,40 @@ functions:
           MONGODB_API_VERSION="${MONGODB_API_VERSION}" \
           NODE_VERSION=${NODE_VERSION} SKIP_DEPS=${SKIP_DEPS|1} NO_EXIT=${NO_EXIT|1} \
             bash ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
+  start-load-balancer:
+    - command: shell.exec
```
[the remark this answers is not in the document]
nbbeeken: We should also make LB a task of the existing linux variant rather than a new variant

src/cmap/command_monitoring_events.ts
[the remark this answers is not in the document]
mbroadst: Yeah I see that we have this written in the tests:
```
expect(actualEvent.hasServiceId).to.equal(expectedEvent.connectionClosedEvent.hasServiceId);
```
could they just be written like:
```
if (expectedEvent.connectionClosedEvent?.hasServiceId) {
  expect(actualEvent.serviceId).to.exist;
}
```
?

This way we wouldn't need to include the getter in the public API

src/cmap/connection.ts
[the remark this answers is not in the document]
mbroadst: I see, I think at very least we should make this a very explicitly named method then rather than a mutator (maybe `updateGenerationForServiceId`) to make the intent very clear. 

I'm still pondering whether we can do this while making `generation` immutable.. A `Connection` is created in the `connect` method, and the `generation` passed here is provided by the `ConnectionPool` when calling `connect`. Since your service generation authority is a map, perhaps you can pass a reference to that into the `connect` options, and resolve the correct generation there?
```
performHandshake(..., (err, hello) => {
  if (hello.serviceId && options.serviceGenerations.has(hello.serviceId)) {
     // modify the generation of the options passed into the `Connection` constructor
  }
});
```

It seems like a big contortion, but I think it's important for avoiding subtle bugs that we maintain immutability where it has been introduced (to solve bugs in the past!)

src/cmap/connection.ts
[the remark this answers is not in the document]
mbroadst: (resolving, we can continue this discussion in the other thread)

src/cmap/connection.ts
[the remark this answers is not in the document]
durran: Even passing in the generation we still execute the handshake after the `Connection` is instantiated, so we still don't have the `serviceId`. I'm happy to explicitly name the method to set it but I don't see any way around this unless we change the driver to execute a handshake before instantiating a `Connection`, which seems really strange to me.

src/cmap/command_monitoring_events.ts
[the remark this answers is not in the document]
durran: That works for the 2 cmap events but doesn't work for the command monitoring events which go through `resultCheck` (https://github.com/mongodb/node-mongodb-native/blob/4.0/test/functional/unified-spec-runner/match.ts#L109) in the unified runner and I'm hesitant to start adding more special cases into that method at this point. I feel like marking them as internal so they don't show up in the API docs as sufficient for now.

.evergreen/run-load-balancer.sh
durran: Done.

.evergreen/config.yml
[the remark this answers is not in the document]
durran: Updated all the Evergreen related comments to use the script generation and pass the script name now into `run-tests.sh`

src/cmap/connection_pool.ts
durran: This is in now.

src/connection_string.ts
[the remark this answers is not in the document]
durran: Refactored so all goes through `validateLoadBalancedOptions`

src/mongo_client.ts
```diff
@@ -638,6 +640,7 @@ export interface MongoOptions
         | 'heartbeatFrequencyMS'
         | 'keepAlive'
         | 'keepAliveInitialDelay'
+        | 'loadBalanced'
```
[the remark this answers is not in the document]
durran: Updated.

package.json
```diff
@@ -113,7 +113,8 @@
     "check:csfle": "mocha test/functional/client_side_encryption",
     "prepare": "node etc/prepare.js",
     "release": "standard-version -i HISTORY.md",
-    "test": "npm run check:lint && npm run check:test"
+    "test": "npm run check:lint && npm run check:test",
+    "test:load-balancer": "mocha test/manual/load-balancer.test.js"
```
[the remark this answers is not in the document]
durran: Updated.

test/functional/unified-spec-runner/runner.ts
```diff
@@ -41,7 +46,27 @@ export async function runUnifiedTest(
     ctx.skip();
   }
 
-  const utilClient = ctx.configuration.newClient();
+  let utilClient;
+  if (ctx.topologyType === uni.TopologyType.loadBalanced) {
```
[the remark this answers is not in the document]
durran: Updated.

src/cmap/connection_pool.ts
```diff
@@ -219,6 +240,35 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+  }
+
+  markPinned(connection: Connection, pinType: string): void {
+    connection.markPinned(pinType);
+    this[kMetrics].markPinned(pinType);
+  }
```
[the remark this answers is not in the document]
durran: This has now been refactored to use connection as an event emitter. 2 differences to the suggection: 1) `PinType` is not implemented and not an `enum` as we actively removed all enums from the code base and are using string constants instead and 2) `once` was not used as connections can be pinned and unpinned multiple times - so `connection.destroy()` removes all listeners for the `pinned` and `unpinned` events.

.evergreen/run-lb-tests.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
durran: Updated.

src/cmap/errors.ts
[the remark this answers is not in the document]
durran: This was so the normal message doesn't have a space or ":" in it when nothing would follow.

src/cmap/command_monitoring_events.ts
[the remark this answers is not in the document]
nbbeeken: The spec states that "driver's MAY choose how they expose the serviceId" which means there could've been any number of ways to attach this information,  I feel like a special case is warranted here because it is the case that it can be in any number of forms. [It looks like Java put this logic in their test runner](https://github.com/mongodb/mongo-java-driver/blob/master/driver-sync/src/test/functional/com/mongodb/client/unified/EventMatcher.java#L61-L69)

That being said it seems like this could've been covered by an `{ $$exists: true }` test if the spec mandated a `serviceId` property, maybe we should file a Drivers ticket to follow up on that?

src/cmap/connect.ts
```diff
@@ -125,6 +127,25 @@ function performInitialHandshake(
         return;
       }
 
+      if (options.loadBalanced) {
+        // TODO: Durran: Remove when server support exists.
```
nbbeeken, line 131: Let's make a ticket from this TODO

src/cmap/connection_pool.ts
nbbeeken: Sorry to revisit, can we use `private` instead of underscore here? We can make use of typescript to prevent usage outside this class

src/cmap/connect.ts
durran: https://jira.mongodb.org/browse/NODE-3431

src/cmap/connection.ts
```diff
@@ -266,6 +279,27 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this[kLastUseTime] = now();
   }
 
+  /**
+   * Mark the connection as pinned if it isn't already pinned. An example
+   * of this is if the connection is pinned to a transaction, and we've
+   * started a cursor inside the transaction, we want to keep the tracking
+   * for the metrics as the transaction.
```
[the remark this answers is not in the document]
durran: This has all been removed now in the metrics changes.

src/cmap/connection_pool.ts
```diff
@@ -289,9 +336,25 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+    if (this.loadBalanced && serviceId) {
+      const sid = serviceId.toHexString();
+      const generation = this.serviceGenerations.get(sid);
+      // Only need to worry if the generation exists, since it should
+      // always be there but typescript needs the check.
+      if (generation == null) {
+        throw new MongoDriverError('Service generations are required in load balancer mode.');
+      } else {
+        // Increment the generation for the service id.
+        /* eslint no-console: 0 */
```
nbbeeken, line 349: Just noting so we don't forget to remove this

src/cmap/connection_pool.ts
```diff
@@ -437,7 +525,31 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
...
+
+    if (pool.loadBalanced) {
+      const onPin = (pinType: string) => {
+        pool[kMetrics].markPinned(pinType);
+      };
+      const onUnpin = (pinType: string) => {
+        pool[kMetrics].markUnpinned(pinType);
+      };
+      connection.on('pinned', onPin);
+      connection.on('unpin', onUnpin);
```
nbbeeken, line 538: Can we make these constants on the Connection class like our other errors?

src/cmap/connection_pool.ts
```diff
@@ -437,7 +525,31 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
       connection.on(event, (e: any) => pool.emit(event, e));
     }
 
-    pool.emit(ConnectionPool.CONNECTION_POOL_CREATED, new ConnectionCreatedEvent(pool, connection));
+    pool.emit(ConnectionPool.CONNECTION_CREATED, new ConnectionCreatedEvent(pool, connection));
+
+    if (pool.loadBalanced) {
+      const onPin = (pinType: string) => {
```
nbbeeken, line 531: Seems like these could be inlined? is there a benefit to pulling them out into variables?

src/cmap/connection_pool.ts
```diff
@@ -437,7 +525,31 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
...
+      };
+      const onUnpin = (pinType: string) => {
+        pool[kMetrics].markUnpinned(pinType);
+      };
+      connection.on('pinned', onPin);
+      connection.on('unpin', onUnpin);
+
+      const serviceId = connection.serviceId;
+      if (serviceId) {
+        let generation;
```
nbbeeken, line 542: `let generation: number;` -> then you won't have to cast it on line 550.

src/cmap/stream_description.ts
```diff
@@ -42,6 +45,8 @@ export class StreamDescription {
     this.maxBsonObjectSize = 16777216;
     this.maxMessageSizeBytes = 48000000;
     this.maxWriteBatchSize = 100000;
+    this.logicalSessionTimeoutMinutes = options ? options.logicalSessionTimeoutMinutes : undefined;
+    this.loadBalanced = options ? !!options.loadBalanced : false;
```
nbbeeken, line 49:
```javascript
this.logicalSessionTimeoutMinutes = options?.logicalSessionTimeoutMinutes;
this.loadBalanced = !!options?.loadBalanced;
```
Just a suggestion

src/cursor/abstract_cursor.ts
```diff
@@ -368,6 +372,7 @@ export abstract class AbstractCursor<
     const needsToEmitClosed = !this[kClosed];
     this[kClosed] = true;
 
+    // console.log('explicitly closing cursor');
```
nbbeeken, line 375: leaving note to remember to remove

src/cursor/abstract_cursor.ts
```diff
@@ -757,15 +762,21 @@ function cursorIsDead(cursor: AbstractCursor): boolean {
...
+): void {
   if (cursor[kDocuments].length === 0) {
     cursor[kClosed] = true;
     cursor.emit(AbstractCursor.CLOSE);
   }
 
   const session = cursor[kSession];
   if (session && session.owner === cursor) {
-    session.endSession(callback);
+    /* eslint no-console: 0 */
```
nbbeeken, line 777: Note for removal

src/sdam/server.ts
```diff
@@ -151,31 +153,36 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       this.clusterTime = clusterTime;
     });
 
-    // create the monitor
-    this[kMonitor] = new Monitor(this, this.s.options);
+    if (!this.loadBalanced) {
```
nbbeeken, line 156: Could this be rewritten as `if (this.loadBalanced) return`? It looks like all the change here is indentation but just want to make sure

src/sdam/topology.ts
```diff
@@ -411,7 +417,19 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
 
     // connect all known servers, then attempt server selection to connect
-    connectServers(this, Array.from(this.s.description.servers.values()));
+    const serverDescriptions = Array.from(this.s.description.servers.values());
+    connectServers(this, serverDescriptions);
+
+    // In load balancer mode we need to fake a server description getting
+    // emitted from the monitor, since the monitor doesn't exist.
+    if (this.s.options.loadBalanced) {
+      serverDescriptions.forEach(description => {
```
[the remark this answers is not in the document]
nbbeeken: This is something we discussed in a node tech chat, when debugging its frustrating that forEach is a step into where for-of/in basic for is all step over, so its not just my preference, but not a blocker of course.

src/sessions.ts
```diff
@@ -72,6 +75,15 @@ export type ClientSessionEvents = {
 
 /** @internal */
 const kServerSession = Symbol('serverSession');
+/** @internal */
+const kPinnedConnection = Symbol('pinnedConnection');
+
+/** @public */
+export interface EndSessionOptions {
+  /** An optional error which caused the call to end this session */
```
nbbeeken, line 83: Is this only for our use? Can mark as `@internal` if so.

src/sessions.ts
```diff
@@ -157,6 +171,29 @@ class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
+    return this[kPinnedConnection];
+  }
+
+  /** @internal */
+  pinConnection(conn: Connection): void {
+    if (this[kPinnedConnection]) {
+      throw TypeError('Cant pin multiple connections to the same session');
+    }
+
+    // console.trace('pinned connection');
```
nbbeeken, line 189: removal

src/sessions.ts
```diff
@@ -157,6 +171,29 @@ class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
+    conn.emit(
+      'pinned',
```
nbbeeken, line 191: Constant on Connection class (See other comment above)

src/sessions.ts
```diff
@@ -369,6 +409,36 @@ function isUnknownTransactionCommitResult(err: MongoError) {
...
+    const servers = Array.from(session.topology.s.servers.values());
+    if (servers.length === 0) {
+      // This can happen if the client is closed when the connection is still pinned
+      // NOTE: we don't usually do this, we could instead throw an error?
+      conn.destroy();
+      session[kPinnedConnection] = undefined;
+      return;
+    }
+
+    const loadBalancer: Server = servers[0];
```
nbbeeken, line 430: Shouldn't need an annotation here.

src/sessions.ts
```diff
@@ -369,6 +409,36 @@ function isUnknownTransactionCommitResult(err: MongoError) {
...
+    loadBalancer.s.pool.checkIn(conn);
+    conn.emit(
+      'unpinned',
```
nbbeeken, line 433: Constant on Connection class.

test/functional/unified-spec-runner/entities.ts
```diff
@@ -290,7 +391,7 @@ export class EntitiesMap<E = Entity> extends Map<string, E> {
         }
 
         map.set(entity.bucket.id, new GridFSBucket(db, options));
-      } else if ('stream' in entity) {
+      } else if ('cursor' in entity) {
         throw new Error(`Unsupported Entity ${JSON.stringify(entity)}`);
```
nbbeeken, line 395: It is supported right? It just needs to be created via an operation, so maybe change the error message here

test/functional/unified-spec-runner/match.ts
```diff
@@ -235,32 +248,73 @@ export function specialCheck(
...
+  actual: (CommandEvent & CmapEvent)[],
   entities: EntitiesMap
 ): void {
   // TODO: NodeJS Driver has extra events
   // expect(actual).to.have.lengthOf(expected.length);
+  if (actual.length !== expected.length) {
+    //const actualNames = actual.map(a => a.constructor.name);
+    const expectedNames = expected.map(e => Object.keys(e)[0]);
+    console.log(
+      `Expected event count mismatch, expected ${inspect(expectedNames)} but got ${inspect(actual)}`
```
nbbeeken, line 284: Is this supposed to be an error now or is this just helping debug?

test/functional/unified-spec-runner/operations.ts
```diff
@@ -173,11 +173,33 @@ operations.set('assertSessionTransactionState', async ({ entities, operation })
...
 
+operations.set('assertNumberConnectionsCheckedOut', async ({ entities, operation }) => {
+  const client = entities.getEntity('client', operation.arguments.client);
+  const servers = Array.from(client.topology.s.servers.values());
+  const checkedOutConnections = servers.reduce((count, server) => {
+    const pool = server.s.pool;
+    return count + pool.currentCheckedOutCount;
+  }, 0);
+  process.nextTick(() => {
+    expect(checkedOutConnections).to.equal(operation.arguments.connections);
```
nbbeeken, line 184: Why is this wrapped in nextTick?

test/functional/unified-spec-runner/operations.ts
```diff
@@ -173,11 +173,33 @@ operations.set('assertSessionTransactionState', async ({ entities, operation })
...
   const collection = entities.getEntity('collection', operation.object);
   return collection.bulkWrite(operation.arguments.requests);
 });
 
+operations.set('close', async ({ entities, operation }) => {
+  try {
+    const cursor = entities.getEntity('cursor', operation.object);
+    await cursor.close();
+  } catch (e) {
+    const changeStream = entities.getEntity('stream', operation.object);
```
nbbeeken, line 198: Why does this catch block attempt to close a change stream of the same name?

test/functional/unified-spec-runner/operations.ts
```diff
@@ -219,6 +241,14 @@ operations.set('createCollection', async ({ entities, operation }) => {
   });
 });
 
+operations.set('createFindCursor', async ({ entities, operation }) => {
+  const collection = entities.getEntity('collection', operation.object);
+  const { filter, sort, batchSize, limit, let: vars } = operation.arguments;
+  const cursor = collection.find(filter, { sort, batchSize, limit, let: vars });
+  await cursor.hasNext();
```
nbbeeken, line 248: I assume this is a little hack to make it grab the first batch? Let's leave a comment

test/functional/unified-spec-runner/operations.ts
```diff
@@ -294,19 +324,36 @@ operations.set('insertMany', async ({ entities, operation }) => {
...
-  ]);
+  try {
+    const changeStream = entities.getEntity('stream', operation.object);
+    // Either change or error promise will finish
+    return Promise.race([
+      changeStream.eventCollector.waitAndShiftEvent('change'),
+      changeStream.eventCollector.waitAndShiftEvent('error')
+    ]);
+  } catch (e) {
+    const findCursor = entities.getEntity('cursor', operation.object);
```
nbbeeken, line 335: Similar to the close operation I'm not understanding why this switches to cursor logic in the catch block.

test/functional/unified-spec-runner/operations.ts
```diff
@@ -294,19 +324,36 @@ operations.set('insertMany', async ({ entities, operation }) => {
...
+    return await findCursor.next();
+  }
+});
+
+operations.set('listCollections', async ({ entities, operation }) => {
+  const db = entities.getEntity('db', operation.object);
+  const { filter, batchSize } = operation.arguments;
+  return await db.listCollections(filter, { batchSize: batchSize }).toArray();
```
nbbeeken, line 343: nit: drop await here

test/functional/unified-spec-runner/operations.ts
```diff
@@ -294,19 +324,36 @@ operations.set('insertMany', async ({ entities, operation }) => {
...
 
 operations.set('listDatabases', async ({ entities, operation }) => {
   const client = entities.getEntity('client', operation.object);
   return client.db().admin().listDatabases();
 });
 
+operations.set('listIndexes', async ({ entities, operation }) => {
+  const collection = entities.getEntity('collection', operation.object);
+  const { batchSize } = operation.arguments;
+  return await collection.listIndexes({ batchSize: batchSize }).toArray();
```
nbbeeken, line 354: nit: drop await here

test/functional/unified-spec-runner/operations.ts
```diff
@@ -441,22 +488,23 @@ export async function executeOperationAndCheck(
...
       expectErrorCheck(error, operation.expectError, entities);
       return;
-    } else {
+    } else if (!operation.ignoreResultAndError) {
       throw error;
     }
   }
 
   // We check the positive outcome here so the try-catch above doesn't catch our chai assertions
+  if (!operation.ignoreResultAndError) {
```
nbbeeken, line 497: total nit: prefer early return here

test/manual/load-balancer.test.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+});
+
+// require('../functional/retryable_reads.test');
+// require('../functional/retryable_writes.test');
+// require('../functional/uri_options_spec.test');
+// require('../functional/change_stream_spec.test');
+// require('../functional/versioned-api.test');
+// require('../unit/core/mongodb_srv.test');
+// require('../unit/sdam/server_selection/spec.test');
```
nbbeeken, line 39: Are these temporary skips for testing right now?

test/tools/mock.js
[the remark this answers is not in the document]
nbbeeken: Still needs removing

test/functional/unified-spec-runner/match.ts
durran: This is just helping debug at the moment.

test/functional/unified-spec-runner/operations.ts
durran: I can remove that - I was testing a theory.

test/functional/unified-spec-runner/operations.ts
durran: The operation executes on both a change stream and a cursor, so it tries first to get a cursor from the entities, and if there is no cursor by that name an exception is thrown, so we catch it and attempt the same for a change stream of the same name.

test/functional/unified-spec-runner/operations.ts
durran: Yes it needs to execute the command but not step the cursor forward. I'll add a comment.

test/functional/unified-spec-runner/operations.ts
durran: Same as above comment for close operation.

test/manual/load-balancer.test.js
durran: Yes just temporary to speed test runs up - these were all passing so commented out for the time being.

test/functional/unified-spec-runner/match.ts
mbroadst: oh sorry this should be a `expect.fail`

test/functional/unified-spec-runner/entities.ts
durran: That was an error in my part, I've reverted the change.

src/cmap/connection_pool.ts
durran: Removed.

src/cmap/connection_pool.ts
durran: Constants on `Connection` now.

src/cmap/connection_pool.ts
durran: No benefit, just my preferred style. I've inlined.

src/cmap/connection_pool.ts
durran: Actually that didn't work in this case and gives the same error. Typescript doesn't seem to be smart enough to know it will never be undefined even with the check if it doesn't exist in the code. Maybe it's not smart enough for Maps?

src/cmap/stream_description.ts
durran: Changed.

src/cursor/abstract_cursor.ts
durran: Removed.

src/cursor/abstract_cursor.ts
durran: Removed.

src/sdam/server.ts
durran: I've changed per suggestion.

src/sessions.ts
durran: The problem is that it's part of the public API for `endSession`, and api-extractor would not let me mark individual signatures of that method internal and some public. Or maybe I'm missing how to do that?

src/sessions.ts
durran: Removed.

src/sessions.ts
durran: Made constant.

src/sessions.ts
```diff
@@ -369,6 +409,36 @@ function isUnknownTransactionCommitResult(err: MongoError) {
...
 
+function maybeClearPinnedConnection(session: ClientSession, options: EndSessionOptions) {
+  // unpin a connection if it has been pinned
+  const conn = session[kPinnedConnection];
+
+  // NOTE: the spec talks about what to do on a network error only, but the tests seem to
+  //       to validate that we don't unpin on _all_ errors?
+  if (conn && (options.error == null || options.force)) {
+    /* eslint no-console: 0 */
+    console.log(' >  unpinConnnection');
```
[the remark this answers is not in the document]
durran: Removed.

src/sessions.ts
durran: Removed.

src/sessions.ts
durran: Made constant.

test/functional/unified-spec-runner/entities.ts
durran: I reverted. It is supported but only via a operation.

test/functional/unified-spec-runner/operations.ts
durran: Removed.

test/functional/unified-spec-runner/operations.ts
durran: Added comment.

test/functional/unified-spec-runner/operations.ts
durran: Removed it here and for `listIndexes` as well. I was hoping that was the cause of the mystery hang but alas, it was not.

test/functional/unified-spec-runner/operations.ts
durran: Changed.

src/cmap/connection_pool.ts
nbbeeken: Ah yea because `.get` can return undefined there isn't a great way to narrow it is my understanding, fine to leave as is.

test/functional/unified-spec-runner/operations.ts
```diff
@@ -173,11 +173,31 @@ operations.set('assertSessionTransactionState', async ({ entities, operation })
...
+
 operations.set('bulkWrite', async ({ entities, operation }) => {
   const collection = entities.getEntity('collection', operation.object);
   return collection.bulkWrite(operation.arguments.requests);
 });
 
+operations.set('close', async ({ entities, operation }) => {
+  try {
+    const cursor = entities.getEntity('cursor', operation.object);
+    await cursor.close();
```
nbbeeken, line 194: Not your doing, but this method claims to not return a promise, looks like the zero argument overload `close(): void` should be `close(): Promise<void>`, can we fix here?

src/cmap/connect.ts
nbbeeken: Thanks! Lets put the ticket in the comment to keep track

src/cmap/connection_pool.ts
nbbeeken: I see the constants but can we make use of them here?

src/cmap/connection_pool.ts
nbbeeken: Inlining them means you can remove the type annotation because the static `'pinned'` coerces the type of the second argument.

src/sessions.ts
nbbeeken:
```ts
/** @public */
export interface EndSessionOptions {
  /**
   * An optional error which caused the call to end this session
   * @internal
   */
  error?: AnyError;
  force?: boolean;
}
```
produces:
```ts
/** @public */
export declare interface EndSessionOptions {
    /* Excluded from this release type: error */
    force?: boolean;
}

```

So just on the error property will do

test/functional/unified-spec-runner/match.ts
nbbeeken: Can we assert the lengths match in the commented out assertion above and we can still use this string as the message for that assertion?

test/functional/unified-spec-runner/operations.ts
nbbeeken: Maybe a suggestion for making this more clear would be to just turn this into existence checks?

```javascript
  const closable =
    entities.getEntity('cursor', operation.object, false) ??
    entities.getEntity('stream', operation.object, false);
  expect(closable).to.be.an('object');
  return closable.close();
```
Maybe something similar can be done for the `iterateUntilDocumentOrError` operation, it seems like this is a good use of skipping the assertExists flag in getEntity

test/functional/unified-spec-runner/operations.ts
durran: Yes. I've updated.

test/functional/unified-spec-runner/operations.ts
durran: That's a nice suggestion but would you want the associated refactoring as part of this PR? The problem is that an entity does exist with the provided name, just the type of entity is different (`stream` vs `cursor`). So the exception thrown is not that the entity doesn't exist, rather the entity is the wrong type. So removing the existence checks still throws an exception in this case.

test/functional/unified-spec-runner/match.ts
durran: Updated for that commented out assertion.

src/sessions.ts
durran: Thanks - updated.

src/cmap/connection_pool.ts
durran: Thanks for catching that - they use it now.

src/cmap/connect.ts
durran: Added ticket to comment.

test/functional/unified-spec-runner/operations.ts
nbbeeken: 🤦 Oh right, we can also just use .get to bypass the checking logic altogether. Or just leave behind a comment mentioning why we're doing this, if this were another language the catch block's error type might sufficiently hint at what's going on but here it's not clear what's being caught (kind of proven by my insufficient suggestion haha)

test/functional/unified-spec-runner/match.ts
nbbeeken: Was it intentional to leave these commented out? I thought @mbroadst identified that we can assert the length check now?

test/functional/unified-spec-runner/schema.ts
```diff
@@ -27,25 +27,37 @@ export const TopologyType = Object.freeze({
   single: 'single',
   replicaset: 'replicaset',
   sharded: 'sharded',
-  shardedReplicaset: 'sharded-replicaset'
-} as const);
+  shardedReplicaset: 'sharded-replicaset',
+  loadBalanced: 'load-balanced'
+}) as const;
```
nbbeeken, line 32:
```suggestion
} as const);
```
the `as const` cast has to be on the object literal

test/functional/unified-spec-runner/match.ts
durran: It was intentional as the Node driver does not execute a `killCursors` after a `getMore` has errored, which causes the absence of the `killCursors` events in some cases. I could not find anywhere so far in the specs that states an errored `getMore` requires an immediate `killCursors` afterwards, and all the driver does in this case is cleanup the cursor (but not explicitly `close()` it).

src/cmap/connection_pool.ts
durran: Removed.

test/functional/unified-spec-runner/operations.ts
durran: Comment on reasoning added.

test/functional/unified-spec-runner/schema.ts
durran: Updated.

test/manual/load-balancer.test.js
durran: These are back now.

PR #2743, opened 2021-02-17
fix: backport emitWarning helper

nbbeeken:
Unify the way we print out messages relating to upcoming changes
or potential misuse of the driver. Using the emitWarning API allows
users to configure when and how messages are shown.

(code from #2740)

NODE-2317, NODE-3114

PR #2744, opened 2021-02-18
fix: unmerged writeConcern options

wolrajhti:
## Description

In 3.6.4 was added a depreciation warning about how writeConcern options should be formatted

https://github.com/mongodb/node-mongodb-native/blob/d67ffa7a2e3f86734c7e9b6944aab1d765b9e75e/lib/write_concern.js#L78-L80

#2743 solves the problem of multiple emitions of depreciation warning, but digging into the code to find why this warning is triggered so many times, I found that `sourceOptions.writeConcern` is never merged into `targetOptions.writeConcern` in `utils.mergeOptionsAndWriteConcern`
https://github.com/mongodb/node-mongodb-native/blob/d67ffa7a2e3f86734c7e9b6944aab1d765b9e75e/lib/utils.js#L289-L322

To fix it, my proposal is to add `writeConcern` to the `collectionKeys` in db.js. Also, `utils.mergeOptionsAndWriteConcern` is not part of the public API and used only 1 time in db.js, so I propose to remove the trailing, always `true`, argument : `mergeWriteConcern`

The original bug was reported here: https://jira.mongodb.org/browse/NODE-3114

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -388,6 +388,7 @@ const collectionKeys = [
   'serializeFunctions',
   'strict',
   'readConcern',
+  'writeConcern',
   'ignoreUndefined',
   'promoteValues',
   'promoteBuffers',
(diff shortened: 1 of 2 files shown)


nodesocket:
@wolrajhti thanks so much for this work. Look forward to this fix merged and a new release.

nbbeeken:
Thanks again for the contribution! 
Closing as #2743 contained the fix proposed here.

PR #2746, opened 2021-03-01
fix: remove catch for synchronous socket errors and remove validation on nodejs option

nbbeeken:
Our validation for a nodejs option suggested that an option could be a boolean when it can actually only be false or a function. In addition this removes the catching of synchronous errors thrown from socket creation. 

NODE-3061

### DIFF
--- a/src/cmap/connect.ts
+++ b/src/cmap/connect.ts
@@ -247,15 +247,6 @@ function parseSslOptions(options: ConnectionOptions): TLSConnectionOpts {
     }
   }
 
-  // Override checkServerIdentity behavior
-  if (!options.checkServerIdentity) {
-    // Skip the identity check by retuning undefined as per node documents
-    // https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
-    result.checkServerIdentity = () => undefined;
-  } else if (typeof options.checkServerIdentity === 'function') {
-    result.checkServerIdentity = options.checkServerIdentity;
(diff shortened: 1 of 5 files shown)


src/cmap/connect.ts
```diff
@@ -247,15 +247,6 @@ function parseSslOptions(options: ConnectionOptions): TLSConnectionOpts {
...
   }
 
-  // Override checkServerIdentity behavior
-  if (!options.checkServerIdentity) {
-    // Skip the identity check by retuning undefined as per node documents
-    // https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
-    result.checkServerIdentity = () => undefined;
-  } else if (typeof options.checkServerIdentity === 'function') {
-    result.checkServerIdentity = options.checkServerIdentity;
-  }
```
nbbeeken, line 257: `checkServerIdentity` is a nodejs TLS option, and nodejs asserts that it is of type function. I suggest that we don't intervene on a boolean type but allow the option to pass directly to the tls API.

mbroadst: Seems fine for 4.0, but make sure the documentation reflects this, and that your upgrade notes mention it. The change can't be made for 3.x unfortunately.

emadum (approved):
LGTM

mbroadst (approved):

durran (approved):

PR #2748, opened 2021-03-08
test: complete unified test runner

nbbeeken:
Completes the unified runner by passing all the poc spec tests.

NODE-2287

## Notes
There's some additional work in here that I'm happy to break out into its own ticket, but it came up in debugging:
- Sessions claimed to accept MongoClientOptions when they were instead receiving MongoOptions, guaranteed to have the well typed WC,RP,RC etc. 
- The issue blocking transactions from working was that my test provided an options object that had `{ writeConcern: undefined }` and that would overwrite the client options that were supposed to be inherited. I can go back and fix the test to not defined a key if the value is undefined but maybe this is unexpected enough to fix in driver, so here I used `??` and explicit fetching of the options to resolve the correct value
- APM overwrites the `client.s.options` which put `client[kOptions]` and `client.options` out of sync. There is also an opportunity for the `client.s.WC/RP/RC` values to be out of sync with `client[kOptions]`. Changing these to getters gives us one source of truth for client options.

### DIFF
--- a/src/apm.ts
+++ b/src/apm.ts
@@ -30,7 +30,7 @@ export class Instrumentation extends EventEmitter {
     const instrumentation = this;
     mongoClientClass.prototype.connect = function (this: MongoClient, callback: Callback) {
       // override monitorCommands to be switched on
-      this.s.options = { ...(this.s.options ?? {}), monitorCommands: true };
+      this.monitorCommands = true;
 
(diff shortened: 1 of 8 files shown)


emadum (approved):
LGTM 🚀

durran (approved):
LGTM

PR #2751, opened 2021-03-10
fix: ensure cursor readPreference is applied to find operations

emadum:
NODE-3129


nbbeeken (approved):
LGTM 🚀 
Should we add tests to catch this?

durran:
> LGTM 🚀
> Should we add tests to catch this?

Agreed, could be just a unit test on execute that mocks out the server.

emadum:
Added a ticket to add the test: NODE-3172

PR #2752, opened 2021-03-10
Adding catch to handled already closed socket

cpeck-bi:
## Description

**What changed?**

### DIFF
--- a/lib/core/connection/connection.js
+++ b/lib/core/connection/connection.js
@@ -271,6 +271,10 @@ class Connection extends EventEmitter {
 
       // Iterate over all buffers and write them in order to the socket
       for (let i = 0; i < buffer.length; i++) {
+        this.socket.on('error', function(){
+          console.log("CLIENT HAS ERRORED");
+          break;
+        });
         this.socket.write(buffer[i], 'binary');
       }
 


lib/core/connection/connection.js
```diff
@@ -271,6 +271,10 @@ class Connection extends EventEmitter {
 
       // Iterate over all buffers and write them in order to the socket
       for (let i = 0; i < buffer.length; i++) {
+        this.socket.on('error', function(){
+          console.log("CLIENT HAS ERRORED");
+          break;
+        });
```
nbbeeken, line 277: This will add an error listener per buffer in the BufferList passed to write and the break on line 276 won't have an effect on the synchronous for loop as the socket can emit the error event at a time later than when the loop is run.

nbbeeken (commented):
Hello! Is there more context on what error you're looking to handle here? It could potentially help to open a more descriptive JIRA ticket here: https://jira.mongodb.org/projects/NODE/issues

There's a bit of an issue with the code proposed here:

lib/core/connection/connection.js
cpeck-bi: I tried this on my fork and it's not accomplishing what I want. I'm going to close this.

PR #2754, opened 2021-03-11
NODE-2951: Don't auto destroy read stream on Node 14+

durran:
The new autoDestroy option for streams in Node 14+ would emit an extra close event when the stream has ended.

### DIFF
--- a/test/functional/cursor.test.js
+++ b/test/functional/cursor.test.js
@@ -1550,48 +1550,41 @@ describe('Cursor', function () {
     }
   });
 
-  it('does not auto destroy streams', {
-    metadata: {
-      requires: { topology: ['single', 'replicaset', 'sharded', 'ssl', 'heap', 'wiredtiger'] }
-    },
+  it('does not auto destroy streams', function (done) {
+    const docs = [];
 
-    test: function (done) {
-      const docs = [];
+    for (var i = 0; i < 10; i++) {
+      docs.push({ a: i + 1 });
+    }
 
-      for (var i = 0; i < 10; i++) {
-        docs.push({ a: i + 1 });
-      }
+    const configuration = this.configuration;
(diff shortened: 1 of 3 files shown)


test/functional/cursor.test.js
```diff
@@ -1550,6 +1550,50 @@ describe('Cursor', function () {
     }
   });
 
+  it('does not auto destroy streams', {
+    metadata: {
+      requires: { topology: ['single', 'replicaset', 'sharded', 'ssl', 'heap', 'wiredtiger'] }
```
nbbeeken, line 1555: I know this `topology` setting is what the other tests have, but for some time the valid topologies have been paired down to the first three in this list: https://github.com/mongodb/node-mongodb-native/blob/a516903bf91e34bb83a3bf565c2a39cdbfddd072/test/tools/runner/filters/mongodb_topology_filter.js#L47-L55

So you can omit this filter all together

test/functional/cursor.test.js
```diff
@@ -1550,6 +1550,50 @@ describe('Cursor', function () {
...
+
+      for (var i = 0; i < 10; i++) {
+        docs.push({ a: i + 1 });
+      }
+
+      const configuration = this.configuration;
+      const client = configuration.newClient(configuration.writeConcernMax(), { maxPoolSize: 1 });
+      client.connect((err, client) => {
+        expect(err).to.not.exist;
+        //this.defer(() => client.close());
```
nbbeeken, line 1569: cruft? I see the close in the end handler below if that was intended.

nbbeeken (changes requested):

test/functional/cursor.test.js
durran: Updated.

test/functional/cursor.test.js
durran: Oh thanks for that - I've removed now.

nbbeeken (approved):
lgtm!

PR #2755, opened 2021-03-11
NODE-3008: Remove existing session from cloned cursors.

durran:
Port up to the 4.0 branch with added tests.


nbbeeken (approved):
Just the same note about the test formatting but otherwise LGTM 🚀

durran:
> Just the same note about the test formatting but otherwise LGTM 🚀

Updated without metadata.

PR #2756, opened 2021-03-15
NODE-3152: ensure AWS environment variables are applied properly

rose-m:
## Description
The AWS environment variables were not detected properly.

The checks for `username == null` and `password == null` would always return false as the static `merge(...)` method that is used to create an instance of `MongoCredentials` defaulted to an empty string as value for both.

Furthermore, the `AWS_SESSION_TOKEN` _must be applied_ if it is present in the environment variables and takes precedence over any other value. See https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#environment-variables:
> However, if AWS_SESSION_TOKEN is set Drivers MUST use its value as the session token.

The direct assignment of `AWS_SESSION_TOKEN` would also potentially fail as it happened that `this.mechanismProperties` was initialized with an already freezed object throwing an error. That's why I changed that to a destructured assignment.

With those changes I am able to run all my E2E tests in mongosh for AWS authentication successfully.

Please note that I did not know where to add test cases for the driver and would ask you to either give me a hint so I can add them or add them on your own :)

### DIFF
--- a/src/cmap/auth/mongo_credentials.ts
+++ b/src/cmap/auth/mongo_credentials.ts
@@ -61,16 +61,19 @@ export class MongoCredentials {
     this.mechanismProperties = options.mechanismProperties || {};
 
     if (this.mechanism.match(/MONGODB-AWS/i)) {
-      if (this.username == null && process.env.AWS_ACCESS_KEY_ID) {
+      if (!this.username && process.env.AWS_ACCESS_KEY_ID) {
         this.username = process.env.AWS_ACCESS_KEY_ID;
       }
 
-      if (this.password == null && process.env.AWS_SECRET_ACCESS_KEY) {
+      if (!this.password && process.env.AWS_SECRET_ACCESS_KEY) {
         this.password = process.env.AWS_SECRET_ACCESS_KEY;
       }
 
-      if (this.mechanismProperties.AWS_SESSION_TOKEN == null && process.env.AWS_SESSION_TOKEN) {
-        this.mechanismProperties.AWS_SESSION_TOKEN = process.env.AWS_SESSION_TOKEN;
+      if (!this.mechanismProperties.AWS_SESSION_TOKEN && process.env.AWS_SESSION_TOKEN) {
+        this.mechanismProperties = {
+          ...this.mechanismProperties,
+          AWS_SESSION_TOKEN: process.env.AWS_SESSION_TOKEN
+        };
       }
     }
 


src/cmap/auth/mongo_credentials.ts
```diff
@@ -61,16 +61,19 @@ export class MongoCredentials {
...
       }
 
-      if (this.password == null && process.env.AWS_SECRET_ACCESS_KEY) {
+      if (!this.password && process.env.AWS_SECRET_ACCESS_KEY) {
         this.password = process.env.AWS_SECRET_ACCESS_KEY;
       }
 
-      if (this.mechanismProperties.AWS_SESSION_TOKEN == null && process.env.AWS_SESSION_TOKEN) {
-        this.mechanismProperties.AWS_SESSION_TOKEN = process.env.AWS_SESSION_TOKEN;
+      if (process.env.AWS_SESSION_TOKEN) {
```
addaleax, line 72: I mean … does this not seem like a spec bug? I would really wonder what their reasons for this are, especially as the description for the credentials `AWS_SESSION_TOKEN` property also states:

> ***AWS_SESSION_TOKEN***
>   Drivers MUST allow the user to specify an AWS session token for authentication with temporary credentials.

durran: Yeah the spec seems to contradict itself in that section, reading to me like URI takes precedence, but if `AWS_SESSION_TOKEN` is there it takes precedence. I think maybe a SPEC ticket to clear that up?

I think for the purposes of this PR though - this looks good to me.

durran (approved):
LGTM

src/cmap/auth/mongo_credentials.ts
rose-m: I just tried opening a SPEC ticket - maybe I lack permissions to do that? Can anyone of you two check if they could open that up? 🤔

emadum: I think [DRIVERS-1405](https://jira.mongodb.org/browse/DRIVERS-1405) relates to this.

emadum (approved):

nbbeeken (approved):
LGTM Thanks @rose-m! I'll merge this soon (edit: Actually I've enabled the AWS Auth tests, so I'll wait on those, and we still have some spec convo above)

Just a heads up to the team: Looks like we need to back port this to 3.6 so I'll put that metadata into the ticket.

src/cmap/auth/mongo_credentials.ts
addaleax: That first paragraph quoted in that ticket is also relevant:

> The order in which Drivers MUST search for credentials is:
>
> #. The URI
> #. Environment variables
> #. The ECS endpoint if ``AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`` is set. Otherwise, the EC2 endpoint.

I don’t know to what degree options as opposed to URI parameters are a Node.js-driver-specific concept, but I would also read this as saying that the environment variable should not take precedence over user-provided options.

I would undo this particular change here.

rose-m: After talking to @nbbeeken I've now changed that so the explicit authMechanismProperties take precedence similar as username and password. I'll therefore resolve that conversation.

PR #2757, opened 2021-03-15
NODE-3120: Ensure monitor has rtt pinger in when calculating rtt

durran:
I couldn't get the driver into a state where this would occur, which is why there's no unit test here. But the 1 line of code change is the same as in the 4.0 branch and wouldn't have any potential negative impact.


nbbeeken (approved):
LGTM!

PR #2758, opened 2021-03-15
NODE-3122: Always close gridfs upload stream on finish

durran:
Tests will hang on Node 14 so just updated Evergreen config with the fix.


nbbeeken (approved):
LGTM! 🚀  I suspect we need this fix in 4.0 as well right?

durran:
Yeah I'm doing a 4.0 PR now.

ex1st:
Should stream emit "close" on "error" also (after `__handleError`)? Thank you!

PR #2759, opened 2021-03-16
PORT: NODE-3122: Adds close event to GridFS stream.

durran:


emadum (commented):
The GridFS changes LGTM! 👍  

However the changes made directly to the `config.yml` to add fermium will be lost the next time the `generate_evergreen_tasks.js` script is run. Not sure if you'd prefer to fix that in this PR, or if we should just address it when we get to NODE-3106?

durran:
Oh I didn't realise it was generated. I'll remove that and pick up that ticket.

PR #2760, opened 2021-03-19
fix(NODE-2995): Add shared metadata MongoClient

nbbeeken:
## Description

Automatic client side encryption needs to perform metadata look ups
like listCollections. In situations where the connection pool size
is constrained or in full use it can be impossible for an operation
to proceed. Adding a separate client in these situations permits the
metadata look ups to proceed unblocking operations.

**What changed?**

I added a class to encapsulate the functionality of managing the internal client. We can discuss the details of how we want to manage this, for example if we feel the new class if overkill we can go to keeping things saved directly on the MongoClient. I made the prose tests somewhat programatic, but I still wanted them to be relatively easily debugged so there's some repetition in the design, of course sound off if we want this cleaner.

A spec test sync is included here, you can filter out the json and yaml files. An important note I synced with the tests written at the time of the meta client spec work since there's been changes since then relating to AWS temporary credentials. Just calling it out if you notice it doesn't match exactly.


nbbeeken:
[Tests should pass now](https://spruce.mongodb.com/version/6058cc0b2fbabe7ca009d406/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Eubuntu1804-custom-csfle-tests%24)

Something to note, I made a custom variant for this situation where we need to run tests against changes in `mongodb-client-encryption` (https://github.com/mongodb/libmongocrypt/pull/168) that aren't yet released. This means merging these changes will break our ubuntu variants where we currently test against a released CSFLE version. I've updated run-tests.sh to pull in the latest, so once we release the fix then those tests will start to pass again. The changes to `mongodb-client-encryption` aren't required to be in sync with the driver, notably, [this line](https://github.com/mongodb/libmongocrypt/blob/e41f450ad5110eb1d8b8710b32de96aa2726e05f/bindings/node/lib/autoEncrypter.js#L103) falls back on the client provided in previous/current versions of the driver.

emadum (approved):
LGTM! 👍

durran (approved):
LGTM

PR #2761, opened 2021-03-22
NODE-3050: Fix Infinite Loop with Optional Require on Windows

durran:
This changes the dependency of require_optional to optional-require, which is currently maintained.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "bl": "^2.2.1",
     "bson": "^1.1.4",
     "denque": "^1.4.1",
-    "require_optional": "^1.0.1",
+    "optional-require": "^1.0.2",
     "safe-buffer": "^5.1.2"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "bl": "^2.2.1",
     "bson": "^1.1.4",
     "denque": "^1.4.1",
-    "require_optional": "^1.0.1",
+    "optional-require": "^1.0.2",
```
nbbeeken, line 50: https://github.com/mongodb/node-mongodb-native/blob/a83541b58fbeb9fc7e323d87734c53ca4d6d2755/package.json#L19-L25

This section of package.json was used by the old lib, we can drop this along with this change.

durran: Thanks, I've updated.

emadum (approved):
LGTM 👍

nbbeeken (approved):

PR #2763, opened 2021-03-23
fix(NODE-3109): prevent servername from being IP

nbbeeken:
servername should only ever be a hostname. Add a condition
to check if the host is an IP and skip auto setting the servername.


emadum (approved):

durran (approved):
LGTM

PR #2765, opened 2021-03-24
NODE-3160: remove apm instrumentation decoration

durran:
As discussed, this PR splits out command monitoring from connection pooling as a refactoring in response to the found cyclic dependency between the connection pool and the events. The core problem is the relationship in the event names, which are defined on the connection pool but exported out of the events. This has been changed now so all the event naming is done in the connection pool to limit the depth of the refactoring for any code that would directly use the event constants on the events, as this would also be a backwards breaking change.

The main part is the removal of the instrumentation code for command monitoring, in which setting a listener directly on the client is preferred.

I've split this into 2 commits to show the clear intentions: 1 commit for the instrumentation work, and 1 commit for the refactoring to get rid of the cyclic dependency.


nbbeeken (approved):

emadum (approved):
LGTM! 👍

PR #2767, opened 2021-03-25
feat(NODE-3115): Add generic parameterization

nbbeeken:
## Add Generics to Typescript Code

Building upon the splendiferous efforts in the community typings, I've brought inhouse the functionality to create collections across a typescript type. Please see the comments I'll post in short order for implementation details.

It's worth calling out: This does not enable runtime type checking, or BSON to class mapping.

NODE-3115


package.json
```diff
@@ -109,8 +114,7 @@
     "release": "standard-version -i HISTORY.md",
     "test": "npm run check:lint && npm run check:test"
   },
-  "homepage": "https://github.com/mongodb/node-mongodb-native",
-  "optionalDependencies": {
-    "saslprep": "^1.0.0"
+  "tsd": {
+    "directory": "test/unit/types"
```
nbbeeken, line 118: This is the only place that [tsd](https://github.com/SamVerschueren/tsd#readme) accepts this configuration, otherwise I would've just put it in the added test.

src/bson.ts
```diff
@@ -86,7 +85,7 @@ export function pluckBSONSerializeOptions(options: BSONSerializeOptions): BSONSe
  */
 export function resolveBSONOptions(
   options?: BSONSerializeOptions,
-  parent?: OperationParent
+  parent?: { bsonOptions?: BSONSerializeOptions }
```
nbbeeken, line 88: This removes a circular dependency wrt TypeScript types only, since everything imports BSON, everything was importing the `OperationParent`.

src/change_stream.ts
```diff
@@ -575,22 +584,29 @@ function waitForTopologyConnected(
   }, 500); // this is an arbitrary wait time to allow SDAM to transition
 }
 
-function closeWithError(changeStream: ChangeStream, error: AnyError, callback?: Callback): void {
+function closeWithError(
+  changeStream: ChangeStream<any>,
```
nbbeeken, line 588: Since this function doesn't need to pipe through the generic (it doesn't use any APIs that depend on what the Schema is) then we can put `any` here. ChangeStream isn't defined to have a default value for the schema since it must be constructed via the `watch()` methods, which do default the generic value. However if we did make `ChangeStream<TSchema = Document>` then these functions would still need to override that with `any` or pass through the generic TSchema despite their implementation being agnostic of the provided generic type.

src/change_stream.ts
```diff
@@ -600,10 +616,10 @@ function endStream(changeStream: ChangeStream): void {
   changeStream[kCursorStream] = undefined;
 }
 
-function processNewChange(
-  changeStream: ChangeStream,
-  change: ChangeStreamDocument,
-  callback?: Callback
+function processNewChange<TSchema>(
+  changeStream: ChangeStream<TSchema>,
+  change: Nullable<ChangeStreamDocument<TSchema>>,
```
nbbeeken, line 621: This is the only place I used `Nullable` a custom type defined in `mongo_types.ts` to mirror the global typescript `NonNullable`. It feels like something that has come up in a few places, the alternative would be `change: ChangeStreamDocument<TSchema> | undefined | null`.

src/gridfs-stream/index.ts
```diff
@@ -44,8 +46,8 @@ export interface GridFSBucketPrivate {
     readPreference?: ReadPreference;
     writeConcern: WriteConcern | undefined;
   };
-  _chunksCollection: Collection;
-  _filesCollection: Collection;
+  _chunksCollection: Collection<GridFSChunk>;
```
nbbeeken, line 49: I decided to dogfood the generics changes in GridFS, it revealed a few type issues and is a good was to ensure we keep the generic usage type checked as we expect it to. I can revert if we want to do this separately.

src/index.ts
```diff
@@ -148,7 +150,7 @@ export type { AuthMechanismId } from './cmap/auth/defaultAuthProviders';
 export type { MongoCredentials, MongoCredentialsOptions } from './cmap/auth/mongo_credentials';
 export type {
   WriteProtocolMessageType,
-  Query,
+  Query as __InternalQuery,
```
nbbeeken, line 153: This wireprotocol class Query conflicts with the new `Query<>` type. Completely open to whatever naming we like.
In the community types it was referred to as `FilterQuery<>` which is just the two terms we often used pushed together but it seems like we should strive to pick one? thoughts?

src/mongo_types.ts
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable @typescript-eslint/ban-types */
+import type { Document, ObjectId } from './bson';
+
+/** @internal */
+export type TODO_NODE_2648 = any;
```
nbbeeken, line 5: When we get to revamping our operations layer, we can pipe through the generics, if this isn't our favorite way to document the TODO's we can refactor here.

src/mongo_types.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/** Given an object shaped type, return the type of the _id field or default to ObjectId @public */
+export type InferIdType<TSchema> = TSchema extends { _id: infer IdType } // user has defined a type for _id
+  ? {} extends IdType
+    ? Exclude<IdType, {}>
+    : unknown extends IdType
+    ? ObjectId
+    : IdType
+  : ObjectId; // user has not defined _id on schema
```
nbbeeken, line 14: This allows users to defined schemas that lack an `_id` property but our internal structures can still default to ObjectId when they need to reference the _id property.

src/mongo_types.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface ProjectionOperators {
+  $elemMatch?: Document;
+  $slice?: number | [number, number];
+  $meta?: MetaProjectionOperators;
+}
+
+/** @public */
+export type Projection<TSchema> = {
+  [Key in keyof TSchema]?: ProjectionOperators | 0 | 1 | boolean;
+};
```
nbbeeken, line 71: This is just a small sample of the modeling going on inside the community types. I've omitted the `UpdateQuery` for now, I can add a ticket number to the TODO soon, unless we want to see that done in this PR?

test/unit/type_check.test.ts
```diff
@@ -0,0 +1,12 @@
+import tsd from 'tsd';
+import { expect } from 'chai'
+
+describe('Exported Types', () => {
+    it('should be as expected', async () => {
+        const diagnostics = await tsd();
+        if (diagnostics.length !== 0) {
+            const messages =  diagnostics.map(d => `${d.fileName}:${d.line}:${d.column} - [${d.severity}]: ${d.message}`).join('\n')
+            expect.fail('\n' + messages)
+        }
```
nbbeeken, line 10: Unfortunately, this encapsulates all the typescript testing into one mocha test. I think this is the best we can do at the moment, but a future project might consider somehow getting more info and generating `it` blocks.

test/unit/types/basic_schema.test-d.ts
```diff
@@ -0,0 +1,34 @@
+import { expectType } from 'tsd';
```
nbbeeken, line 1: All files in this directory have to end with `test-d.ts` to be discovered by `tsd`. None of the code is actually run, its run through the TS compiler and the `expectType` etc functions allow tsd to capture the results. I tried to come up with a variety of schemas and I tested mostly with insertOne since it runs the schema through the more complex `EnhancedOmit`, `OptionalId`, and `InferIdType` types defined in mongo_types.ts

Please let me know if there's a case y'all think we should add here. 🙂

src/mongo_types.ts
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable @typescript-eslint/ban-types */
```
nbbeeken, line 1: This is to permit the `{}` type which means anything not null.

This file does compile to an empty javascript file currently. However I think that [NODE-3132](https://jira.mongodb.org/browse/NODE-3132) will add code that will actually compile to something in this file so its worth leaving in.

CONTRIBUTORS.md
```diff
@@ -1,53 +1,100 @@
...
+- andrewjstone
+- Christopher Stott
+- Corey Jewett
+- brettkiefer
+- Rob Holland
+- Senmiao Liu
+- heroic
+- gitfy
+
+## Maintainers
```
nbbeeken, line 55: Let me know if I got the info in this section right / is the way we want to keep this attribution organized.

src/index.ts
durran: In this case driver specifications have always used "filter" as the terminology post `OP_QUERY`, so my preference would be `Filter` here.

CONTRIBUTORS.md
durran: Seems fine to me, but I do use my personal email for Github and npm (durran@gmail.com). Not sure if that matters or not.

src/mongo_types.ts
durran: I think it's fine to do that in a separate PR if you prefer that.

src/index.ts
durran: > In this case driver specifications have always used "filter" as the terminology post `OP_QUERY`, so my preference would be `Filter` here.

durran (changes requested):

src/index.ts
nbbeeken: Going off the API examples in the MongoDB manual it seems we use it interchangably, I do think filter will eliminate this issue so let's go with that.

CONTRIBUTORS.md
nbbeeken: Yes! I'll fix that 😃

src/bulk/common.ts
```diff
@@ -31,37 +32,37 @@ export const BatchType = {
 export type BatchTypeId = typeof BatchType[keyof typeof BatchType];
 
 /** @public */
-export interface InsertOneModel {
+export interface InsertOneModel<TSchema> {
```
PaulEndri, line 35: Hi, I've actually been looking forwards to this feature. I was wondering if this could/should instead be `TSchema extends Document = Document` like I see you made some of the ones below work? Defaulting and expecting it to match the shape of a document seems like it might be helpful in preventing some errors and w/ type-ahead/type-hinting. Could maybe have it do `extends Partial<Document>` instead or something of the sorts, if the concern is not needing everything from the document

src/change_stream.ts
PaulEndri: As a general point for the `changeStream` typings, what do you think of adding either a second (optional) generic that users could pass in or a different type that users could customize? ChangeStream documents can vary in shape pretty wildly and I have in fact ended up having to do `changeStream as any` in the past because it conflicted with some of the expected object types, but I could have provided a concrete type instead: Typescript just didn't play nice because it read the types I was trying to convert between as conflicting.

src/cursor/abstract_cursor.ts
```diff
@@ -9,6 +9,7 @@ import { Readable, Transform } from 'stream';
 import { EventEmitter } from 'events';
 import type { ExecutionResult } from '../operations/execute_operation';
 import { ReadConcern, ReadConcernLike } from '../read_concern';
+import type { TODO_NODE_2648 } from '../mongo_types';
```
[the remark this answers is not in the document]
nbbeeken: This should be done, let me know if the new comments are what you expected.

src/cursor/abstract_cursor.ts
```diff
@@ -594,7 +601,7 @@ function nextDocument(cursor: AbstractCursor): Document | null | undefined {
   if (doc) {
     const transform = cursor[kTransform];
     if (transform) {
-      return transform(doc);
+      return transform(doc) as TSchema; // TODO: How can user's parameterize this
```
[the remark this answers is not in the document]
nbbeeken: Upon further investigation I don't think there's a problem here. But another set of eyes could help confirm that. It looks like there's no way to save a transform on to a cursor. I thought it was an option during construction that would effect the type emitted from each `.next()` but I don't see how you can specify one. So this is dead code?

test/unit/type_check.test.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const tsd = require('tsd').default;
+const { expect } = require('chai');
+
+describe('Exported Types', () => {
+  it('should be as expected', async () => {
```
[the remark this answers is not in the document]
nbbeeken: Updated the naming. tsd uses the settings I added to the package.json to look at the `test/types` directory and it compiles all files in that folder and reports any errors the compiler finds. The new name does call out the directory but let me know if you think there's a better way to capture this.

test/types/mongodb.test-d.ts
```diff
@@ -0,0 +1,13 @@
+import { expectDeprecated } from 'tsd';
```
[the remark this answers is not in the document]
nbbeeken: Everything in the test/types directory is only compiled and the potential errors are captured by `tsd`.

test/types/union_schema.test-d.ts
```diff
@@ -0,0 +1,45 @@
...
+interface Rectangle {
+  _id: ObjectId;
+  height: number;
+  width: number;
+}
+type Shape = Circle | Rectangle;
+
+type c = Collection<Shape>;
+// should not insert a portion of a type
+expectNotAssignable<Parameters<c['insertOne']>[0]>({ height: 2 });
```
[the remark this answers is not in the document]
nbbeeken: I cleaned up the tests quite a bit here. When I was first playing with tsd I was writing code that I would check the expectType against but you can cut out a lot of boilerplate if you extract say the type of the first argument to `insertOne` so I've tried to do that more here. Grabbing the return type of findOne however is a bit more difficult since it changes depending upon the arguments passed in, so it is easier to write out some fake findOne calls to check the results.

Turns out the mixed properties between unions is permitted, you can see the added test. I'm not sure if this is something to be disallowed or not 🤔

test/types/indexed_schema.test-d.ts
```diff
@@ -0,0 +1,35 @@
...
+type InsertRes<TId = ObjectId> = Promise<{ acknowledged: boolean; insertedId: TId }>;
+
+interface RandomKeysToNumberIncludingId {
+  _id: number;
+  a: number;
+  [extraKeys: string]: number;
+}
+const randomKeysIncludeIdC = new Collection<RandomKeysToNumberIncludingId>(db, '');
+expectType<InsertRes<number>>(randomKeysIncludeIdC.insertOne({ a: 2, randomKey: 23, _id: 23 }));
+expectError(randomKeysIncludeIdC.insertOne({ a: 2, randomKey: 23 }));
```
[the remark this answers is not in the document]
nbbeeken: Added a test for this!

src/collection.ts
```diff
@@ -146,7 +154,7 @@ export interface CollectionPrivate {
  * });
  * ```
  */
-export class Collection {
+export class Collection<TSchema extends Document = any> {
```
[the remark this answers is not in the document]
nbbeeken: Everywhere that we use `Collection` it is actually now `Collection<any>`. If we made the change to Document it would default to `Collection<Document>`. It causes a large number of compiler errors, but one example is:

This helper function, that is used in every crud method on collection can no longer accept `getTopology(this)`
```
export function getTopology(provider: MongoClient | Db | Collection): Topology;
```

That is because it accepts Collection<Document> _only_. But if we have some parametrized collection like `Collection<Pet>` then supposedly `getTopology` cannot operate on `Collection<Pet>`. However all that getTopology needs is access to certain properties to detect if its passed a client/db/collection and use that info to find the topology instance. 

There is more errors than just getTopology but its pretty much the same idea with each of them. The solutions I see are the following:
- Leaving Collection to default to `any`, that allows any collection to be passed to these functions that would fail if there was another default type.
- Change Collection to default to Document and make the error spots accept Collection<any> explicitly
- Remove the explicit dependence on Collection when possible and instead require objects that have the correct shape

The last bullet is something like `getTopology` only really depends upon `{s: {db:{s: {client: {toplogy: Topology}}}}}` for the Collection argument.

test/types/basic_schema.test-d.ts
```diff
@@ -0,0 +1,67 @@
...
+import { MongoClient } from '../../src/mongo_client';
+import type { InferIdType } from '../../src/mongo_types';
+
+const db = new Db(new MongoClient(''), '');
+
+type ACounter = { a: number };
+type ACounterWithId = { a: number; _id: ObjectId };
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Can defined Schema without _id
```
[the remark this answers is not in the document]
nbbeeken: done!

src/bulk/common.ts
nbbeeken: Hey! Thanks for jumping into the conversation, I appreciate feedback from the community on this effort, looks like I already made your suggestion, feel free to call out anything else you see 😃

src/mongo_types.ts
PaulEndri: Nesting ternaries for type declarations seems a bit counter intuitive and if I were trying to use this it might be a bit confusing, what do you think about using some combination of union types, and the `Extract` and/or `Pick` utility types? Something like `Extract<Pick<TSchema, '_id'> | ObjectId, NonNullable>` (not perfect/working, just a general idea/path, sorry if I'm annoying/overstepping!)

src/collection.ts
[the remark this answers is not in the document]
PaulEndri: Would using something akin to ` = Partial<Document>` or creating an interim type like `type CollectionDocument = Pick<Document, neededFields> & Partial<Document>`?

src/bulk/common.ts
PaulEndri: Yup, awesomesauce! Thank you for the work in maintaining this library :D

test/unit/type_check.test.ts
dariakp: good enough for now :)

src/index.ts
dariakp: leaving this conversation open to make sure we get rid of any outward facing instances of "query" in parameters and replace with "filter"

src/mongo_types.ts
dariakp: Leaving this comment open to see if we still even need the ban-types; option to refactor existing types to no longer need to reference {}

src/change_stream.ts
dariakp: It looks like the `ChangeStream` is currently coded to accept a generic that will apply to things like the `fullDocument` returned from the change stream update event, did you have something different in mind? We are also currently looking at typing more of the change stream methods (including this `closeWithError` one which should really be a class method but due to legacy implementation is stuck being a helper method, making strong typing somewhat awkward).

src/cursor/abstract_cursor.ts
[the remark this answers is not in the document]
dariakp: Let's create a ticket to investigate what, if anything, can we done to provide typing on this due to the complication with chaining

test/types/union_schema.test-d.ts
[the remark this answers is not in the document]
dariakp: We discussed adding a test for a true XOR setup as a sanity check to make sure our added type inference doesn't break it;

we also discussed adding some of our sample code (translating existing js to ts) as an explicit typescript compilation test to make sure users trying to use our samples don't immediately run into issues;

ideally, we also talked about potentially adding a TS compilation test for each public constructor/API to make sure they are returning expected types for each of a set of use cases - I recommend splitting this off into a separate ticket as an immediate follow up to this one

src/change_stream.ts
PaulEndri: I was thinking the `<any>` could be replaced with `<TChangeStream>` and the function can be initialized with `<TChangeStream = any>` for tighter control when needed

src/collection.ts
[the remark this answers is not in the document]
nbbeeken: This turned out to be a lot easier of a fix
```typescript
getTopology<T>(provider: MongoClient | Db | Collection<T>)
```
If we want to have our Collection default to something other than any we just have to allow any generic type to be filled in to helper functions like this. I went through and dismissed the errors generated by the operations layer using `Collection` by using the `as TODO_NODE_2648` cast.

src/cursor/abstract_cursor.ts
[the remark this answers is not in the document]
nbbeeken: Ticket: NODE-3283
And I've permitted overriding the generic in `forEach`, `next`, `tryNext`, and `toArray`. 
```
db.collection<Pet>().find().map((p) => p.age).toArray() // Still returns Pet[] in TS but number[] at runtime
db.collection<Pet>().find().map((p) => p.age).toArray<number>() // correct TS, number[]
```
Unfortunately it's not very useful for `[Symbol.asyncIterator]()` since that isn't called by users.

src/mongo_types.ts
nbbeeken: We'll revisit this in NODE-3285, I've captured the suggestion there, thanks!

test/types/union_schema.test-d.ts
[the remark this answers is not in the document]
nbbeeken: Added the test for the XOR union type, it does maintain its restrictions.

src/change_stream.ts
```diff
@@ -603,22 +610,26 @@ function waitForTopologyConnected(
...
-function streamEvents(changeStream: ChangeStream, cursor: ChangeStreamCursor): void {
+function streamEvents<T>(changeStream: ChangeStream<T>, cursor: ChangeStreamCursor<T>): void {
   const stream = changeStream[kCursorStream] || cursor.stream();
   changeStream[kCursorStream] = stream;
   stream.on('data', change => processNewChange(changeStream, change));
   stream.on('error', error => processError(changeStream, error));
 }
 
-function endStream(changeStream: ChangeStream): void {
+function endStream(changeStream: ChangeStream<any>): void {
```
dariakp, line 632: Should this one also be typed or is it an intentional omission?

src/change_stream.ts
```diff
@@ -659,7 +670,7 @@ function processNewChange(
   return callback(undefined, change);
 }
 
-function processError(changeStream: ChangeStream, error: AnyError, callback?: Callback) {
+function processError(changeStream: ChangeStream<any>, error: AnyError, callback?: Callback) {
```
dariakp, line 673: intentionally keeping `<any>` here?

dariakp: Also, should we standardize to just one of T or TSchema in this file as the generic name?

test/types/schema_helpers.test-d.ts
```diff
@@ -0,0 +1,44 @@
+import type {
+  InferIdType,
+  WithId,
+  // WithoutId,
```
dariakp, line 4: should the commented out types have been used somewhere?

test/types/schema_helpers.test-d.ts
```diff
@@ -0,0 +1,44 @@
...
+expectType<InferIdType<{ _id: number }>>(1 + 1);
+expectType<InferIdType<{ a: number } | { b: string }>>(new ObjectId());
+expectAssignable<InferIdType<{ _id: number } | { b: string }>>(new ObjectId());
+expectAssignable<InferIdType<{ _id: number } | { b: string }>>(1 + 1);
+
+// WithId
+expectAssignable<WithId<Document>>({ _id: new ObjectId() });
+expectAssignable<WithId<{ a: number }>>({ _id: new ObjectId(), a: 3 });
+expectNotType<WithId<Document>>({ _id: 3 });
+// expectType<WithId<{ _id: number }>>({ _id: 3 });
```
dariakp, line 23: lots of commented out expect lines here

dariakp (changes requested):
A few things left to clean up - almost there!

src/change_stream.ts
dariakp: I believe we have the types covered now, let us know if you see anything else, thanks!

dariakp: I believe we have the types covered now, let us know if you see anything else, thanks!

src/change_stream.ts
nbbeeken: I corrected this, and changed the use of T to TSchema, let me know if its 👍

test/types/schema_helpers.test-d.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-3287
Made a ticket about the testing we talked about earlier

dariakp: Looks good, tyvm!

dariakp (approved):
🚀

PR #2770, opened 2021-03-26
fix(NODE-3071): Ignore error message if error code is defined

nbbeeken:
Prior behavior was to check the error message if the code did not
correspond to the server state. With this change the driver will not
inspect the error message for error type. ServerDescriptions with
topology versions less than the current will be ignored.

NODE-3071, NODE-2559


src/cmap/commands.ts
```diff
@@ -30,7 +30,7 @@ export type WriteProtocolMessageType = Query | Msg | GetMore | KillCursor;
 
 /** @internal */
 export interface OpQueryOptions extends CommandOptions {
-  socketTimeout?: number;
+  socketTimeoutMS?: number;
```
nbbeeken, line 33: We had an inconsistency where `socketTimeoutMS` was being ignored because we only checked for `socketTimeout`. You'll see that update in a number of places.

src/error.ts
```diff
@@ -7,26 +7,55 @@ export type AnyError = MongoError | Error;
 
 const kErrorLabels = Symbol('errorLabels');
 
+/** @internal MongoDB Error Codes */
+export const MONGODB_ERROR_CODES = Object.freeze({
```
nbbeeken, line 11: Since we're using codes in a number of places it seemed to make sense to make an enum out of them all. Luckily the git diff shows the naming is the same as the comment that was placed next to a lot of the numbers.

src/error.ts
```diff
@@ -291,42 +320,45 @@ export function isRetryableError(error: MongoError): boolean {
...
+  MONGODB_ERROR_CODES.NotMasterOrSecondary
 ]);
 
-const SDAM_NOTMASTER_CODES = new Set([
-  10107, // NotMaster
-  13435 // NotMasterNoSlaveOk
+const SDAM_NOTMASTER_CODES = new Set<number>([
+  MONGODB_ERROR_CODES.NotMaster,
+  MONGODB_ERROR_CODES.NotMasterNoSlaveOk,
+  MONGODB_ERROR_CODES.LegacyNotPrimary
```
nbbeeken, line 334: This additional error code is part of the spec change

src/error.ts
```diff
@@ -291,42 +320,45 @@ export function isRetryableError(error: MongoError): boolean {
...
 function isRecoveringError(err: MongoError) {
-  if (err.code && SDAM_RECOVERING_CODES.has(err.code)) {
-    return true;
+  if (typeof err.code !== 'undefined') {
+    // If any error code exists, we ignore the error.message
+    return SDAM_RECOVERING_CODES.has(err.code);
   }
 
-  return err.message.match(/not master or secondary/) || err.message.match(/node is recovering/);
+  return /not master or secondary/.test(err.message) || /node is recovering/.test(err.message);
```
nbbeeken, line 348: This change was to make the return type `boolean` instead of `RegExpMatch`

src/sdam/server_description.ts
```diff
@@ -256,14 +256,10 @@ export function compareTopologyVersion(lhs?: TopologyVersion, rhs?: TopologyVers
...
-      return 0;
-    } else if (lhs.counter < rhs.counter) {
-      return -1;
-    }
-
-    return 1;
+    // tests mock counter as just number, but in a real situation counter should always be a Long
+    const lhsCounter = Long.isLong(lhs.counter) ? lhs.counter : Long.fromNumber(lhs.counter);
+    const rhsCounter = Long.isLong(rhs.counter) ? lhs.counter : Long.fromNumber(rhs.counter);
+    return lhsCounter.compare(rhsCounter);
```
nbbeeken, line 262: When testing the counter property is a number so `===` works as expected. However the counter is supposed to be a Long and `===` as well as `<` will fail when this is the case. So converting here made sense. There might be a larger issue here of internal communications with mongod losing precision since we don't breakout our own set of bson options, thoughts?

src/sdam/topology.ts
```diff
@@ -612,6 +612,11 @@ export class Topology extends EventEmitter {
       return;
     }
 
+    // ignore this server update if its from an outdated topologyVersion
+    if (isStaleServerDescription(this.s.description, serverDescription)) {
+      return;
+    }
```
nbbeeken, line 618: This is the changes from NODE-2559 that didn't get ported to 4.0 branch. It was necessary to do here to get spec tests to pass.

src/sdam/topology.ts
```diff
@@ -965,6 +970,19 @@ function processWaitQueue(topology: Topology) {
...
+  incomingServerDescription: ServerDescription
+) {
+  const currentServerDescription = topologyDescription.servers.get(
+    incomingServerDescription.address
+  );
+  const currentTopologyVersion = currentServerDescription?.topologyVersion;
+  return (
+    compareTopologyVersion(currentTopologyVersion, incomingServerDescription.topologyVersion) > 0
+  );
+}
```
nbbeeken, line 984: Also NODE-2559

test/functional/spec-runner/index.js
```diff
@@ -121,6 +121,21 @@ function generateTopologyTests(testSuites, testContext, filter) {
           afterEach(() => testContext.cleanupAfterSuite());
           testSuite.tests.forEach(spec => {
             it(spec.description, function () {
+              if (requires.authEnabled) {
+                // TODO: We do not have a way to determine if auth is enabled in our mocha metadata
+                // We need to do a admin.command({getCmdLineOpts: 1}) if it errors (code=13) auth is on
+                this.skip();
+              }
```
nbbeeken, line 128: Syncing the tests added some tests that use `authEnabled` as a runOn requirement, I believe we don't currently have a way of filtering for this, correct me if I'm wrong and I'll make the change. If not I left a comment suggesting the method the python driver uses.

test/functional/spec-runner/index.js
```diff
@@ -121,6 +121,21 @@ function generateTopologyTests(testSuites, testContext, filter) {
...
+
+              if (
+                spec.operations.some(
+                  op => op.name === 'waitForEvent' && op.arguments.event === 'PoolReadyEvent'
+                )
+              ) {
+                // TODO(NODE-2994): Connection storms work will add new events to connection pool
+                this.skip();
+              }
```
nbbeeken, line 137: We haven't implemented the paused or ready events on a connection pool yet which some tests block on a `'waitForEvent'` operation, until mocha times them out.

test/functional/spec-runner/index.js
emadum: One option could be to check the `$AUTH` environment variable. At least when running on Evergreen, it'll be `auth` if the tests are running against a server with auth enabled. A bit more hassle when running tests locally since you need to set that env var when running your tests, but the simplicity of the implementation could make that worth it.

nbbeeken:
Hey @emadum if you can take a look when you get the chance that would be much appreciated 🙂

test/functional/spec-runner/index.js
emadum: Flagging this comment because it seems I didn't leave it as a review last time.

emadum (approved):
Had one comment a few days ago but it looks like it got lost, either way I don't think that needs to hold this up; LGTM 👍

nbbeeken:
Hey @durran just a reminder to take a look at this when you get the chance, thanks!

PR #2772, opened 2021-03-30
fix(NODE-2995): Add shared metadata MongoClient [PORT 4.0]

nbbeeken:
Automatic client side encryption needs to perform metadata look ups
like listCollections. In situations where the connection pool size
is constrained or in full use it can be impossible for an operation
to proceed. Adding a separate client in these situations permits the
metadata look ups to proceed unblocking operations.


emadum (approved):

durran (approved):
LGTM

PR #2774, opened 2021-03-31
fix(NODE-3176): handle errors from MessageStream

nbbeeken:
If there's a socket that gets a large amount of data written to it our MessageStream will throw an error about exceeding the permitted BSON size, this error is now captured in SDAM.


emadum (approved):
LGTM, nice refactor 👍

durran (approved):

PR #2775, opened 2021-04-01
chore: reintroduce nodejs fermium testing

nbbeeken:
Somehow I dropped fermium testing, bringing it back here.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1430,11 +1430,11 @@ tasks:
           TOPOLOGY: server
       - func: run custom csfle tests
 buildvariants:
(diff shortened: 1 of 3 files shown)


emadum (approved):

emadum (changes requested):
Can we restrict the tests that fail on fermium to `node<14` and add a ticket to fix them?

nbbeeken:
> Can we restrict the tests that fail on fermium to `node<14` and add a ticket to fix them?

Excellent idea, went and did that, so the remaining errors are on macos only and look like the typical flakiness we usually see there.

emadum (approved):

PR #2597, opened 2020-10-27
test: add docker-based cluster setup script

emadum:
## Description

This PR introduces a Docker-based alternative to the `test/tools/cluster_setup.sh` script. An Ubuntu 18.04 based Docker image is built with `m` and `mtools` and is then used to `mlaunch` versioned instances of MongoDB. 

The new `test/tools/docker_cluster.sh` script:

- takes an additional `version` argument in addition to the `topology`
- adds support for launching all three topologies of a given version by passing `all` as the topology
- automatically builds the image if it doesn't already exist locally; no setup is required beyond having Docker running
- tags containers with the topology/version, so they can easily be stopped, monitored, or shelled into using standard Docker commands
- `test/tools/docker_cluster.sh killall` will stop all instances of mongodb launched by the script

Example usage:
```
➜ ./test/tools/docker_cluster.sh all 3.6
Image already cached, skipping build; force a rebuild with 'docker build --no-cache'
0ff754a8e898279cd39deecff5b24fec42e9b0ea0f4b57510e10a26e54369343
c74b3d1c71620767c1ffb1ee7d4c1b317ea4200ec286ddd3b013f830966fcdbb
17cdd78d17e26d5e597851cf759424adb21d1bb7cfbcbdaef33f8d19e7230c89
mongodb://localhost:27017
mongodb://localhost:31000/?replicaSet=rs
mongodb://localhost:51000,localhost:51001/
CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS                  PORTS                      NAMES
c74b3d1c7162        node-mongodb-native/docker-mongodb   "/entrypoint.sh sing…"   1 second ago        Up Less than a second   0.0.0.0:27017->27017/tcp   mongo_server_3.6

[ Tailing container logs, Ctrl+C to exit; the container is detached and will continue running until stopped with 'docker kill' ]

CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS                  PORTS                                  NAMES
0ff754a8e898        node-mongodb-native/docker-mongodb   "/entrypoint.sh repl…"   1 second ago        Up Less than a second   0.0.0.0:31000-31003->31000-31003/tcp   mongo_replica_set_3.6

[ Tailing container logs, Ctrl+C to exit; the container is detached and will continue running until stopped with 'docker kill' ]

CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS                  PORTS                                  NAMES
[...]

PR #2778, opened 2021-04-07
chore: improve docker_cluster script

emadum:
This PR improves the `test/tools/docker_cluster.sh` script added in #2597.

Downloaded mongoDB binaries are now stored in a persistent shared volume,
so they don't need to be re-downloaded each time a container is launched.


dariakp:
No longer needed

PR #2783, opened 2021-04-14
test(NODE-3187): port unified test runner

nbbeeken:
The spec tests are brought in in the first commit, so looking at the second commit forward is a good way to filter out the spec files. 
I think there's a number of things we can talk about here:

- [ ] Changing the package scripts to match 4.0 is inspired by the need to add a compile step before running the tests
- [ ] Is it okay that we need to install node 12 on boxes that test older node versions just to compile the runner? We do this for bson already
- [ ] Is adding the getters to the bulk result an acceptable change here? it is a requirement for the tests to pass, but we could skip them instead and come back to this

Among more decisions made here, I'll try and comment on things worth calling out


durran (approved):

dariakp:
TODO: pull in latest code from 3.6 to avoid potential merge conflicts

dariakp (approved):
LGTM

PR #2784, opened 2021-04-19
fix(NODE-3166): allowInvalidHostnames and allowInvalidCertificates flags are ignored

nbbeeken:
The insecure TLS flags were not being correctly translated to the equivalent nodejs options.

### DIFF
--- a/test/unit/mongo_client_options.test.js
+++ b/test/unit/mongo_client_options.test.js
@@ -298,21 +298,42 @@ describe('MongoOptions', function () {
   });
 
   it('transforms tlsAllowInvalidCertificates and tlsAllowInvalidHostnames correctly', function () {
-    const options = parseOptions('mongodb://localhost/', {
+    const optionsTrue = parseOptions('mongodb://localhost/', {
       tlsAllowInvalidCertificates: true,
       tlsAllowInvalidHostnames: true
     });
-    expect(options.rejectUnauthorized).to.equal(false);
-    expect(options.checkServerIdentity).to.be.a('function');
-    expect(options.checkServerIdentity()).to.equal(undefined);
+    expect(optionsTrue.rejectUnauthorized).to.equal(false);
+    expect(optionsTrue.checkServerIdentity).to.be.a('function');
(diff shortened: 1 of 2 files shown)


test/unit/mongo_client_options.test.js
```diff
@@ -296,4 +296,44 @@ describe('MongoOptions', function () {
...
+    expect(optionsTrue.rejectUnauthorized).to.equal(false);
+    expect(optionsTrue.checkServerIdentity).to.be.a('function');
+    expect(optionsTrue.checkServerIdentity()).to.equal(undefined);
+
+    const optionsFalse = parseOptions('mongodb://localhost/', {
+      tlsAllowInvalidCertificates: false,
+      tlsAllowInvalidHostnames: false
+    });
+    expect(optionsFalse.rejectUnauthorized).to.equal(true);
+    expect(optionsFalse.checkServerIdentity).not.exist;
```
dariakp, line 314: For this and the other lines that use the `exist` assertion I recommend using `equal(undefined)` instead for two reasons: 1) if we expect a precise outcome, it's better to use the precise assertion (e.g., the current assertion would lead me to conclude that it could be null or undefined, and then wonder what might cause it to be null or undefined), and the [chai expect docs](https://www.chaijs.com/api/bdd/#method_exist) recommend the same; 2) I generally try to avoid assertions that are properties and not function calls as a matter of best practice because they make it easier to accidentally disable a test via a typo (e.g., if you change `not.exist` to `not.exit`, that test would still pass even though it's no longer asserting anything, because the chain would resolve to `undefined`)

dariakp (changes requested):
Thanks for adding the tests, looking good minus the minor detail noted in the code comment!

test/unit/mongo_client_options.test.js
nbbeeken: You make great points, I agree the potential typo bug is particularly risky, I've made the changes.

dariakp (approved):

durran:
I'm not listed as a reviewer but LGTM

emadum (approved):

PR #2785, opened 2021-04-20
feat(NODE-3132)!: Add TypedEventEmitter

nbbeeken:
Using a mapped type of event names to function type can now provide
annotations and completion for event argument types.

Breaking Changes:
- Topology open event emits one argument which is the topology itself
- srvPollingHandler on topology now emits TopologyDescriptionChange event (erroneously was serverDescriptionChanged)


dariakp (changes requested):
Let's make sure there are tests covering at least the breaking changes

emadum (approved):
LGTM! 👍 for the refactor removing `relayEvents`

test/unit/sdam/srv_polling.test.js
```diff
@@ -425,4 +425,35 @@ describe('Mongos SRV Polling', function () {
       });
     });
   });
+
+  it('should emit a topologyDescriptionChanged event upon receiving an SRV update', function () {
```
dariakp, line 429: we talked about refactoring the test setup into a describe block to isolate the test condition better in the `it`

dariakp (changes requested):
Let's also add an actual ticket for fixing how the srv polling event handler is attached and reference it in the TODO, and either fix the duplicate event emission on updateServers now or add a ticket to do it later

test/unit/sdam/srv_polling.test.js
```diff
@@ -454,6 +439,33 @@ describe('Mongos SRV Polling', function () {
           new TopologyDescription(TopologyType.Sharded)
         )
       );
+      // This changes the topology to a sharded cluster
+      // We now have topology.s.handleSrvPolling assigned
+      expect(topology.s.handleSrvPolling).to.be.a('function');
+    });
+
+    it('two topologyDescriptionChanged events are emitted when receiving an SRV update with one server', function (done) {
```
dariakp, line 447: This is still a bug though, right? Is there a ticket for it?

test/unit/sdam/srv_polling.test.js
```diff
@@ -454,6 +439,33 @@ describe('Mongos SRV Polling', function () {
...
+      topology.on(Topology.TOPOLOGY_DESCRIPTION_CHANGED, ev => {
+        if (eventCount === 0) {
+          // topologyDescriptionChanged event emitted from the handleSrvPolling function
+          expect(ev.topologyId).to.equal(1);
+          expect(ev.previousDescription.servers.size, 'previous should have 0 servers').to.equal(0);
+          expect(ev.newDescription.servers.size, 'next should have 1 server').to.equal(1);
+        }
+
+        if (eventCount >= 1) {
+          // Testing >= 1 to allow done to be called potentially more than once here catching erroneous events
```
dariakp, line 458: Since the actual expectations are the same in both blocks, you can probably pull them outside and update the "previous" expects to just use eventCount in place of 0 or 1, and just leave the `done()` call in the `if (eventCount >= 1)` block

test/unit/sdam/srv_polling.test.js
nbbeeken: I don't think so, meant to clarify, but its good to have visibility here on the PR. 
The topology emits a change event when it receives the SRVRecord and adds the server to the topology. Another change event will be fired for each server as the SDAM machinery runs discovery on the new host names received. It's my understanding that this intentional.
Maybe the test name is still not great, its not explicitly two events, its 1 event + number of servers in the SRV resolution.

dariakp: updated test looks good! :)

test/unit/sdam/srv_polling.test.js
dariakp: resolving with the update

test/unit/sdam/srv_polling.test.js
dariakp: resolving as done

nbbeeken:
@dariakp Looks like the revert passes if you wanted to take another look here

dariakp (approved):
LGTM

PR #2788, opened 2021-04-23
fix(NODE-3174): Preserve sort key order for numeric string keys

dariakp:
## Description

Updated internal representation of sort to use a map instead of a plain object in order to preserve correct ordering for multi-key sort queries; also enabled the use of a map type to specify input sort options

### DIFF
--- a/src/sort.ts
+++ b/src/sort.ts
@@ -13,6 +13,7 @@ export type Sort =
   | string
   | string[]
   | { [key: string]: SortDirection }
+  | Map<string, SortDirection>
   | [string, SortDirection][]
   | [string, SortDirection];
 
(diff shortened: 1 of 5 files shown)


durran (approved):
LGTM. Are the todos to be addressed in another PR or ticket?

dariakp:
> LGTM. Are the todos to be addressed in another PR or ticket?

It really depends on how the investigation for the unified test framework/extended json in our internal DRIVERS-1701 ticket plays out. Unless the outcome is favorable, this won't be able to be addressed at all, but I wanted to keep the test marked regardless so that it's top of mind whenever we get around to cleaning them up.

emadum (changes requested):
Nice work! Can you create a Jira ticket for the future work described in the TODO comments, and update the comments with a reference to the ticket? These kind of comments have a tendency to never get followed up on if they aren't associated with a ticket. (Edit: perhaps the DRIVERS ticket mentioned above would suffice?)

emadum (approved):

PR #2789, opened 2021-04-25
Fix url parsing for a mongodb+srv url that has commas in the database name

pablomatiasgomez:
## Description

Fixes an invalid exception that is thrown when a mongodb+srv uri contains commas (`,`) in the database name. For example:
`mongodb+srv://valid.example.com/db,with,commas`

Unfortunatelly, `mongodb+srv` urls/uris are not tested so I couldn't add a test. I think this is because the url parsing also resolves the dns name, which is what happened when I tried to add a test. Of course we don't want to resolve dns names in tests but doing that would require a bigger refactor.

**What changed?**
Basically there was an if that checks if the numbers of hostnames that were provided were more than one for `mongodb+srv` urls/uris, but the condition was invalid as it was actually checking the `pathname`, which in the mongo uri that is the database name. I have a mongodb database with commas in its url and I am not able to connect to it.

No.

### DIFF
--- a/lib/core/uri_parser.js
+++ b/lib/core/uri_parser.js
@@ -53,11 +53,8 @@ function parseSrvConnectionString(uri, options, callback) {
 
   result.domainLength = result.hostname.split('.').length;
 
-  if (result.pathname) {
-    const maxSlashes = result.pathname.endsWith('/') ? 2 : 1;
-    if ((result.pathname.match(/\//g) || []).length > maxSlashes) {
-      return callback(new MongoParseError('Invalid URI, cannot contain multiple hostnames'));
-    }
+  if (result.pathname && (result.pathname.match(/\//g) || []).length > 1) {
+    return callback(new MongoParseError('Invalid URI, cannot contain multiple hostnames'));
   }
 
   if (result.port) {
--- a/lib/core/uri_parser.js
+++ b/lib/core/uri_parser.js
@@ -52,8 +52,12 @@ function parseSrvConnectionString(uri, options, callback) {
   }
 
   result.domainLength = result.hostname.split('.').length;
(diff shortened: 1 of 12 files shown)


durran (changes requested):
Thanks for the PR. While this does solve the issue when database names contain a comma, it does however break our tests in the following case:

- Multiple hosts are provided in the URI, ie `mongodb://test1.com,test2.com/dbname`

This is because we initially do a `URI.parse` which does not know how to handle
multiple hosts, as they are not part of the URI specification. We allow multiple hosts
with the `mongodb` scheme but not with the `mongodb+srv` scheme. So this
change actually causes our srv validation to pass when it should not be valid.

Example of a failing test in this case: https://github.com/mongodb/node-mongodb-native/blob/3.6/test/spec/connection-string/invalid-uris.yml

agolin95:
Tracked in https://jira.mongodb.org/browse/NODE-3234

pablomatiasgomez:
@durran Thanks for the review.
 
I actually didn't find that there were some tests for `+srv` uris so now I was able to run them and I also added a few for this case.

I was able to fix the issue by going back to check on the `pathname` instad of `hostname` but instead of looking for `,`, I am now looking for slashes. Given that the mongodb uri can only contain one or two slashes, if it has more than that, it means that the hostname was parsed as part of the `pathame` (as you mentioned, by the `URL.parse`).

Let me know if you agree with this approach or if you would it to be different. I thought on looking by `.` but someone could still provide a hostname without a dot...

Thanks!

durran:
@pablomatiasgomez Thanks and I think this looks good. I just need to have a conversation with the other dbx teams since we'll probably want to port those new tests across other drivers potentially.

durran (approved):

emadum (approved):

PR #2790, opened 2021-04-26
fix(NODE-3173): Preserve sort key order for numeric string keys

dariakp:
## Description

Updated internal representation of sort to use a map instead of a plain object in order to preserve correct ordering for multi-key sort queries


emadum (approved):
LGTM

durran (approved):

PR #2791, opened 2021-04-27
fix(sdam): topology no longer causes close event

durran:
The topology was emitting both a "topologyClosed" and "close"
event which are bubbled up to the MongoClient. This was causing
two "close" events to be emitted from the MongoClient when the
client is closed.

NODE-3219

There was one test on this branch looking for that event that was
looking for that event and not "topologyClosed", so I'm wondering
if this should not be backported.


emadum (approved):
Personally this feels more like a bug-fix and I think it's unlikely users are currently relying on that buggy behavior, so giving the backport my 👍

dariakp (approved):
I'm reasonably convinced that there isn't anything critical relying on this behavior

PR #2792, opened 2021-04-27
fix(sdam): topology no longer causes close event

durran:
The topology was emitting both a "topologyClosed" and "close"
event which are bubbled up to the MongoClient. This was causing
two "close" events to be emitted from the MongoClient when the
client is closed.

NODE-3219

Note that there were no tests explicitly looking for that event, but there
are tests looking for the `topologyClosed` event, so I assume this just
slipped through.


emadum (approved):

dariakp (approved):
LGTM

PR #2794, opened 2021-04-29
refactor!: remove deprecated bulk ops

emadum:
NODE-2978


dariakp (changes requested):
I think the removed operations should also be removed from the doc comments listing legal operation types in the collection.ts for bulkWrite

dariakp (approved):
lgtm

durran (approved):
LGTM

PR #2798, opened 2021-05-05
docs(NODE-3314): Changed '.equal()' to '.strictEqual()' in examples

lMartin3:
asserts.equal() is deprecated, changed it to assert.strictEqual().

## Description

**What changed?**
README.md - Changed the examples so they used asserts.strictEqual() instead of asserts.equal() (which is deprecated).

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ const dbName = 'myproject';
 const client = new MongoClient(url);
 // Use connect method to connect to the server
 client.connect(function(err) {
-  assert.equal(null, err);
+  assert.strictEqual(null, err);
   console.log('Connected successfully to server');
 
   const db = client.db(dbName);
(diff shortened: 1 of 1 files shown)


nbbeeken (approved):
Seems like a good idea to modernize the examples, thanks for your contribution!

dariakp (approved):

README.md
```diff
@@ -226,9 +226,9 @@ const insertDocuments = function(db, callback) {
   const collection = db.collection('documents');
   // Insert some documents
   collection.insertMany([{ a: 1 }, { a: 2 }, { a: 3 }], function(err, result) {
-    assert.equal(err, null);
-    assert.equal(3, result.result.n);
-    assert.equal(3, result.ops.length);
+    assert.strictEqual(err, null);
```
emadum, line 229:
```suggestion
    assert.ifError(err);
```

I think we could go a bit further and update the error checks to use `assert.ifError` here, which seems like the most idiomatic approach.

emadum (changes requested):
Thanks for this PR! Just one small suggestion.

wa-Nadoo:
IMHO the better approach is to enforce strict assertion mode for all methods. Just replace `const assert = require('assert')` with  `const assert = require('assert').strict` or  `const assert = require('assert/strict')`. In strict assertion mode all non-strict methods behave like strict methods. More info is available here: https://nodejs.org/api/assert.html#assert_strict_assertion_mode.

nbbeeken:
Closing due to inactivity and target branch needs updating, 3.7 is now the current 3.x branch

PR #2800, opened 2021-05-05
fix(NODE-3194): Ignore undefined and null options in MongoClient constructor

dariakp:
## Description

Fix NODE-3194: MongoClient constructor should not try to parse undefined or null options

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -340,7 +340,9 @@ export function parseOptions(
     }
   }
 
-  const objectOptions = new CaseInsensitiveMap(Object.entries(options));
+  const objectOptions = new CaseInsensitiveMap(
+    Object.entries(options).filter(([, v]) => (v ?? null) !== null)
+  );
 
   const allOptions = new CaseInsensitiveMap();
 
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -370,9 +370,11 @@ export function parseOptions(
     allKeys,
     Array.from(Object.keys(OPTIONS)).map(s => s.toLowerCase())
   );
+  const optionWord = unsupportedOptions.size > 1 ? 'options' : 'option';
+  const isOrAre = unsupportedOptions.size > 1 ? 'are' : 'is';
   if (unsupportedOptions.size !== 0) {
     throw new MongoParseError(
-      `options ${Array.from(unsupportedOptions).join(', ')} are not supported`
+      `${optionWord} ${Array.from(unsupportedOptions).join(', ')} ${isOrAre} not supported`
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/connection_string.ts
```diff
@@ -368,9 +370,11 @@ export function parseOptions(
     allKeys,
     Array.from(Object.keys(OPTIONS)).map(s => s.toLowerCase())
   );
+  const optionWord = unsupportedOptions.size > 1 ? 'options' : 'option';
+  const isOrAre = unsupportedOptions.size > 1 ? 'are' : 'is';
   if (unsupportedOptions.size !== 0) {
```
nbbeeken, line 375:
```suggestion
  if (unsupportedOptions.size !== 0) {
    const optionWord = unsupportedOptions.size > 1 ? 'options' : 'option';
    const isOrAre = unsupportedOptions.size > 1 ? 'are' : 'is';
```
Just a suggestion to move these checks inside the if block that throws the error.

test/unit/mongo_client_options.test.js
```diff
@@ -192,6 +191,33 @@ describe('MongoOptions', function () {
...
+    expect(options).not.to.have.property('randomopt');
+    expect(options).not.to.have.property('otherrandomopt');
+  });
+
+  it('should throw an error on unrecognized keys in the options object if they are defined', function () {
+    expect(() =>
+      parseOptions('mongodb://localhost:27017/', {
+        randomopt: 'test'
+      })
+    ).to.throw(MongoParseError, /randomopt.+not supported$/);
```
nbbeeken, line 218:
```suggestion
expect(() =>
      parseOptions('mongodb://localhost:27017/', {
        randomopt: 'test'
      })
    ).to.throw(MongoParseError, 'option randomopt is not supported');

    expect(() =>
      parseOptions('mongodb://localhost:27017/', {
        randomopt: 'test',
        randomopt2: 'test'
      })
    ).to.throw(MongoParseError, 'options randomopt, randomopt2 are not supported');
```
I figure we might as well test the pluralization fixes here, if you agree. Otherwise, I think the current tests are good!

nbbeeken (changes requested):

src/connection_string.ts
dariakp: That's actually where I was intending to put them, not sure how they ended up above it - good catch!

test/unit/mongo_client_options.test.js
dariakp: No strong objections here; the thought was that testing precise error messages can be considered too restrictive because the exact wording should not matter and changing, say, the capitalization of the first letter or providing more context in the message shouldn't break the test; I personally don't think it's that much of a pain to update the test if you update the message, so if the rest of the team agrees, it's good with me :)

nbbeeken (approved):
LGTM!

durran (approved):
LGTM

PR #2801, opened 2021-05-07
fix(NODE-3116): Move negative time check up in async interval

durran:
NODE-3116

@mbroadst Do you see any negative implications in moving this check up to the first one?

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1010,6 +1010,7 @@ export function makeInterruptibleAsyncInterval(
     // to reschedule.
     if (timeUntilNextCall < 0) {
       executeAndReschedule();
+      return;
     }
 
     // debounce multiple calls to wake within the `minInterval`
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1004,6 +1004,14 @@ export function makeInterruptibleAsyncInterval(
     // never completes, the `timeUntilNextCall` will continue to grow
     // negatively unbounded, so it will never trigger a reschedule here.
 
+    // This is possible in virtualized environments like AWS Lambda where our
+    // clock is unreliable. In these cases the timer is "running" but never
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -1004,6 +1004,14 @@ export function makeInterruptibleAsyncInterval(
     // never completes, the `timeUntilNextCall` will continue to grow
     // negatively unbounded, so it will never trigger a reschedule here.
 
+    // This is possible in virtualized environments like AWS Lambda where our
+    // clock is unreliable. In these cases the timer is "running" but never
+    // actually completes, so we want to execute immediately and then attempt
+    // to reschedule.
+    if (timeUntilNextCall < 0) {
+      executeAndReschedule();
```
mbroadst, line 1012: I think you probably want to add a `return` after `executeAndReschedule` if you move it here, since it otherwise used to be the last statement in the function. Otherwise, I think it looks okay. If you wanted to double check I think I made the unit tests for this class pass in a custom clock you can mess around with.

mbroadst (changes requested):

src/utils.ts
durran: Added: https://github.com/mongodb/node-mongodb-native/pull/2801/commits/c0abff66087722859560a300f203056efe9a0805

mbroadst: LGTM. Do you think its worth adding a test to cover this, since not adding a `return` statement caused no failure? I only suggest it because this code is pretty critical, and often hard to debug when something is wrong (i.e. the Lambda unreliable time situation)

mbroadst: @durran ping

durran: I'll add a test for that.

dariakp (changes requested):
@durran I believe this is just awaiting an additional test

PR #2802, opened 2021-05-10
fix(NODE-3197): revert setImmediate in waitQueue

nbbeeken:
A performance regression was identified in switching the connection pool to using setImmediate instead of process.nextTick for wait queue processing. This patch reverts that change.

NODE-3197

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -231,7 +231,7 @@ export class ConnectionPool extends EventEmitter {
     }
 
     this[kWaitQueue].push(waitQueueMember);
-    setImmediate(() => processWaitQueue(this));
+    process.nextTick(() => processWaitQueue(this));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


durran (approved):
LGTM

src/cmap/connection_pool.ts
```diff
@@ -231,7 +231,7 @@ export class ConnectionPool extends EventEmitter {
     }
 
     this[kWaitQueue].push(waitQueueMember);
-    setImmediate(() => processWaitQueue(this));
+    process.nextTick(() => processWaitQueue(this));
```
wa-Nadoo, line 234: `process.nextTick()` forwards the second and subsequent arguments to the callback function. It permits to avoid creating an unnecessary closure and improves performance further. It's a very popular pattern across  the Node.js core codebase (https://github.com/nodejs/node/search?p=1&q=process.nextTick+path%3Alib%2F).
```suggestion
    process.nextTick(processWaitQueue, this);
```

emadum (approved):

src/cmap/connection_pool.ts
nbbeeken: Thanks for taking the time to suggest this, I agree we can drop the closure

PR #2803, opened 2021-05-10
refactor(NODE-1812)!: replace returnOriginal with returnDocument option

dariakp:
## Description

NODE-1812

BREAKING: `returnOriginal` option no longer supported for `findOneAndReplace()` and `findOneAndUpdate()` collection methods, use the new `returnDocument` enum instead.

Replaced `returnOriginal` with `returnDocument` option in order to conform with shared drivers spec.

### DIFF
--- a/src/operations/find_and_modify.ts
+++ b/src/operations/find_and_modify.ts
@@ -12,8 +12,8 @@ import type { WriteConcern, WriteConcernSettings } from '../write_concern';
 
 /** @public */
 export const ReturnDocument = Object.freeze({
-  BEFORE: 'Before',
-  AFTER: 'After'
+  BEFORE: 'before',
+  AFTER: 'after'
 } as const);
 
 /** @public */
@@ -37,7 +37,7 @@ export interface FindOneAndReplaceOptions extends CommandOperationOptions {
   hint?: Document;
   /** Limits the fields to return for all matching documents. */
   projection?: Document;
-  /** When set to 'After', returns the updated document rather than the original. The default is 'Before'.  */
+  /** When set to 'after', returns the updated document rather than the original. The default is 'before'.  */
   returnDocument?: ReturnDocument;
   /** Determines which document the operation modifies if the query selects multiple documents. */
   sort?: Sort;
@@ -55,7 +55,7 @@ export interface FindOneAndUpdateOptions extends CommandOperationOptions {
   hint?: Document;
   /** Limits the fields to return for all matching documents. */
   projection?: Document;
-  /** When set to 'After', returns the updated document rather than the original. The default is 'Before'.  */
+  /** When set to 'after', returns the updated document rather than the original. The default is 'before'.  */
   returnDocument?: ReturnDocument;
   /** Determines which document the operation modifies if the query selects multiple documents. */
   sort?: Sort;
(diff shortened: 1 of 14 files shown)


src/operations/find_and_modify.ts
```diff
@@ -10,6 +10,15 @@ import { Sort, SortForCmd, formatSort } from '../sort';
 import type { ClientSession } from '../sessions';
 import type { WriteConcern, WriteConcernSettings } from '../write_concern';
 
+/** @public */
+export const ReturnDocument = Object.freeze({
```
durran, line 14: This is completely expected to me given what other drivers do as far as it being defined. I'd just note the options passed to the command are `true` for `AFTER` and `false` for `BEFORE`, so if those aren't the values here they'd need to be mapped somewhere.

durran: Oh wait nevermind. I saw to handle that below.

dariakp:
The corresponding deprecation for 3.6 is here: https://github.com/mongodb/node-mongodb-native/pull/2808

src/operations/find_and_modify.ts
```diff
@@ -10,6 +10,15 @@ import { Sort, SortForCmd, formatSort } from '../sort';
...
+  BEFORE: 'before',
+  AFTER: 'after'
```
nbbeeken, line 16: `ReturnDocument.before` and `ReturnDocument.after` would be my preference here, seems just 🤏  that much easier to type. Not a strong opinion though

test/functional/unified-spec-runner/operations.ts
```diff
@@ -234,17 +234,20 @@ operations.set('find', async ({ entities, operation }) => {
 
 operations.set('findOneAndReplace', async ({ entities, operation }) => {
   const collection = entities.getEntity('collection', operation.object);
-  return collection.findOneAndReplace(operation.arguments.filter, operation.arguments.replacement);
+  const { filter, replacement, ...opts } = operation.arguments;
+  if (opts.returnDocument) {
+    opts.returnDocument = opts.returnDocument.toLowerCase();
+  }
```
nbbeeken, line 240: This suggestion could be overly defensive here but can we make a `processOptions` or `translateOptions` function in `unified-utils.ts` and use it in both places here? I can foresee us needing more of these spec-to-driver translations that might be good to box up the functionality in one place. 

You'll find `patchDbOptions` and `patchCollectionOptions` in that utils file that currently do nothing since we're lucky to share camel-casing and correct value types with the spec but I can see when those hooks already exist would come in handy.

nbbeeken (commented):
Some suggestions:

src/operations/find_and_modify.ts
dariakp: I checked and we currently have about a 50-50 split on uppercase vs camelcase in existing enums; my personal preference is to use uppercase with enums to emphasize the static nature of the thing being referenced as opposed to just being a regular property on a regular object

test/functional/unified-spec-runner/operations.ts
dariakp: let me know if the added code is what you had in mind

src/operations/find_and_modify.ts
nbbeeken: SGTM

test/functional/unified-spec-runner/operations.ts
nbbeeken: Looks good!

nbbeeken (approved):
🚀

emadum (approved):
LGTM 👍

PR #2804, opened 2021-05-11
fix(NODE-3263): force closing a pool closes active connections

rose-m:
Prior to this PR using `MongoClient.close(true)` to force-close the client did also trigger a force-close of the connection pool and the individual connections. However, as the pool only tracked _available_ connections (not all connections that were created by the pool), it could not close connections that were _active_.

This fix PR adds:
* Tracking all connections that are created by the pool in a new internal `Set`.
* On force-close instead of iterating only over the _available_ connections it will iterate over all _known_ connections and force-close them.
* A test case is provided to ensure that also checked-out connections are force-closed.

The behavior and semantics of the `MongoClient.close()` (i.e. without `force = true`) remain unchanged.

The test case will fail when the iteration in line 342 is made only over the available connections.

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -20,6 +20,7 @@ import {
 } from './connection_pool_events';
 
 const kLogger = Symbol('logger');
+const kAvailableConnections = Symbol('availableConnections');
 const kConnections = Symbol('connections');
 const kPermits = Symbol('permits');
 const kMinPoolSizeTimer = Symbol('minPoolSizeTimer');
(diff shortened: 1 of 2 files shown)


dariakp:
Converting this PR to draft pending additional investigation of NODE-3263

Xunnamius:
Just got bitten by this in my CLI app. I can't do a clean exit even after all operations have completed and after calling `MongoClient.close(true)`; my program hangs unless I add a `process.exit(0)` at the end, which makes unit and integration testing a bit harder. Any plans on merging this fix?

PR #2805, opened 2021-05-12
update README to use assert.strictEqual

ikejs:
## Description

**What changed?**
- Updated README
`assert.equal` is deprecated since v9.9.0. Switched to `assert.strictEqual`


nbbeeken:
Looks like someone's already got these changes posted in #2798 so I'll defer to that one. Regardless, thank you for contributing!

PR #2806, opened 2021-05-12
fix(NODE-3192): check clusterTime is defined before access

nbbeeken:
The fix is really just making the if statement on line 53 check the correct properties before accessing them, but I also created some variables to help track which cluster time is being used where.

### DIFF
--- a/lib/core/wireprotocol/command.js
+++ b/lib/core/wireprotocol/command.js
@@ -45,14 +45,20 @@ function _command(server, ns, cmd, options, callback) {
   const shouldUseOpMsg = supportsOpMsg(server);
   const session = options.session;
 
-  let clusterTime = server.clusterTime;
+  const serverClusterTime = server.clusterTime;
+  let clusterTime = serverClusterTime;
   let finalCmd = Object.assign({}, cmd);
   if (hasSessionSupport(server) && session) {
+    const sessionClusterTime = session.clusterTime;
     if (
-      session.clusterTime &&
-      session.clusterTime.clusterTime.greaterThan(clusterTime.clusterTime)
+      serverClusterTime &&
+      serverClusterTime.clusterTime &&
+      sessionClusterTime &&
+      sessionClusterTime.clusterTime
     ) {
-      clusterTime = session.clusterTime;
+      if (sessionClusterTime.clusterTime.greaterThan(serverClusterTime.clusterTime)) {
+        clusterTime = sessionClusterTime;
+      }
     }
 
     const err = applySession(session, finalCmd, options);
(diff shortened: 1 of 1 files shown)


lib/core/wireprotocol/command.js
```diff
@@ -45,14 +45,20 @@ function _command(server, ns, cmd, options, callback) {
...
     if (
-      session.clusterTime &&
-      session.clusterTime.clusterTime.greaterThan(clusterTime.clusterTime)
+      serverClusterTime &&
+      serverClusterTime.clusterTime &&
+      sessionClusterTime &&
+      sessionClusterTime.clusterTime
     ) {
-      clusterTime = session.clusterTime;
+      if (sessionClusterTime.clusterTime.greaterThan(serverClusterTime.clusterTime)) {
```
dariakp, line 59: This may just be my own bias against nested if's, but why not just append that check to the end of the chain of the preceding 4 `&&` conditions?

lib/core/wireprotocol/command.js
```diff
@@ -45,14 +45,20 @@ function _command(server, ns, cmd, options, callback) {
...
+        clusterTime = sessionClusterTime;
```
dariakp, line 60: we want to follow up on what we want to happen if serverClusterTime.clusterTime itself is undefined - previously the greaterThan looks like it would throw an error, in the new code, it won't, but the question is what do we want to happen - should we throw, ignore, or set the server clusterTime to the session clusterTime;

also, need to double check to make sure whether this fix is also needed in 4.0

dariakp (changes requested):

lib/core/wireprotocol/command.js
nbbeeken: The error is that `clusterTime` itself is not defined and we don't have a check for that but in [4.0 we do here](https://github.com/mongodb/node-mongodb-native/blob/4.0/src/cmap/connection.ts#L353).

So our mistake was that `server.clusterTime` can be undefined so later calling `clusterTime.clusterTime` failed. This refactor is unnecessary, we can just add `clusterTime` to the sequence (`&&`) of existence checks if we prefer. I do think giving names to the sources of the clusterTime make this a bit more readable. Thoughts?

lib/core/wireprotocol/command.js
nbbeeken: I agree, I'll fix

lib/core/wireprotocol/command.js
dariakp: I'm never opposed to descriptive variable names

dariakp (approved):
LGTM

emadum (approved):

PR #2807, opened 2021-05-12
fix(NODE-3252): state transistion from DISCONNECTED

nbbeeken:
Change the state change to a legal transition


emadum (approved):

dariakp (approved):
LGTM, though it would have been nice to have tests for this that aren't disabled

nbbeeken:
Made NODE-3289 for us to look into that.

PR #2808, opened 2021-05-12
refactor(NODE-1812): Deprecate returnOriginal in favor of returnDocument

dariakp:
## Description

NODE-1812

Deprecated `returnOriginal` in favor of `returnDocument` option in order to conform with shared drivers spec for `findOneAndReplace()` and `findOneAndUpdate()` collection methods.

### DIFF
--- a/lib/operations/find_one_and_replace.js
+++ b/lib/operations/find_one_and_replace.js
@@ -1,16 +1,22 @@
 'use strict';
 
+const MongoError = require('../core').MongoError;
 const FindAndModifyOperation = require('./find_and_modify');
 const hasAtomicOperators = require('../utils').hasAtomicOperators;
 
 class FindOneAndReplaceOperation extends FindAndModifyOperation {
   constructor(collection, filter, replacement, options) {
+    if ('returnDocument' in options && 'returnOriginal' in options) {
+      throw new MongoError(
+        'findOneAndReplace option returnOriginal is deprecated in favor of returnDocument and cannot be combined'
+      );
+    }
     // Final options
     const finalOptions = Object.assign({}, options);
     finalOptions.fields = options.projection;
     finalOptions.update = true;
-    finalOptions.new = options.returnOriginal !== void 0 ? !options.returnOriginal : false;
-    finalOptions.upsert = options.upsert !== void 0 ? !!options.upsert : false;
+    finalOptions.new = options.returnDocument === 'after' || options.returnOriginal === false;
+    finalOptions.upsert = options.upsert === true;
 
     if (filter == null || typeof filter !== 'object') {
       throw new TypeError('Filter parameter must be an object');
--- a/lib/operations/find_one_and_update.js
+++ b/lib/operations/find_one_and_update.js
@@ -1,17 +1,22 @@
 'use strict';
 
+const MongoError = require('../core').MongoError;
 const FindAndModifyOperation = require('./find_and_modify');
 const hasAtomicOperators = require('../utils').hasAtomicOperators;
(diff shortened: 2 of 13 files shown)


nbbeeken (approved):
LGTM!

lib/operations/find_one_and_replace.js
```diff
@@ -1,16 +1,22 @@
 'use strict';
 
+const MongoError = require('../core').MongoError;
 const FindAndModifyOperation = require('./find_and_modify');
 const hasAtomicOperators = require('../utils').hasAtomicOperators;
 
 class FindOneAndReplaceOperation extends FindAndModifyOperation {
   constructor(collection, filter, replacement, options) {
+    if ('returnDocument' in options && 'returnOriginal' in options) {
+      throw new MongoError(
```
emadum, line 10:
```suggestion
      throw new TypeError(
```

I think this would be more appropriate than `MongoError`

lib/operations/find_one_and_update.js
```diff
@@ -1,17 +1,22 @@
 'use strict';
 
+const MongoError = require('../core').MongoError;
 const FindAndModifyOperation = require('./find_and_modify');
 const hasAtomicOperators = require('../utils').hasAtomicOperators;
 
 class FindOneAndUpdateOperation extends FindAndModifyOperation {
   constructor(collection, filter, update, options) {
+    if ('returnDocument' in options && 'returnOriginal' in options) {
+      throw new MongoError(
```
emadum, line 10: Same here

emadum (changes requested):
One small request on the type of error thrown, other than that LGTM 👍

lib/operations/find_one_and_replace.js
dariakp: I actually disagree with this use of TypeError, because a TypeError to a developer is usually an indication that the wrong data type was passed in, which in this case isn't true of the individual options. If we consider the entirety of the options object as being of a particular type (defined by the allowed combinations of keys), TypeError might technically be applicable, but I think it's a bit odd to do type validation on a dictionary as a whole instead of individual members (particularly when we don't actually validate any of the other options at the same time). What we really need is a validation error type as distinct from a server or other sort of error; unfortunately, the driver currently silently overwrites options in the other cases of deprecation and actual option validation is sparse and examining the code base, most of the other instances of option validation use MongoError, hence its use here for consistency.

emadum: The [other errors like this](https://github.com/mongodb/node-mongodb-native/blob/4.0/src/read_preference.ts#L114-L127) throw TypeErrors; I think it would be good to be consistent, though I don't feel strongly about this kind of errors being TypeErrors and would be OK with a new error type.

dariakp: Yeah those definitely should not be TypeErrors. And aside from the read preference file, other places do use MongoError:
https://github.com/mongodb/node-mongodb-native/blob/1cdc8a8738d8c2425e0ff76751331636894256b8/src/utils.ts#L55-L79
https://github.com/mongodb/node-mongodb-native/blob/4.0/src/operations/aggregate.ts#L76
https://github.com/mongodb/node-mongodb-native/blob/4.0/src/db.ts#L756
with the cleanest example using the MongoParseError:
https://github.com/mongodb/node-mongodb-native/blob/4.0/src/connection_string.ts#L136
right now though the MongoParseError is limited to the connection string parsing and I'm not sure we want to expand its use elsewhere

emadum (approved):

PR #2809, opened 2021-05-14
fix(NODE-3275): Fix enum type export naming and serverApi validation

dariakp:
## Description
NODE-3275

Make type exports for enums use the same name as the enums themselves

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -2,8 +2,8 @@ import * as dns from 'dns';
 import * as fs from 'fs';
 import { URL, URLSearchParams } from 'url';
 import { AuthMechanism } from './cmap/auth/defaultAuthProviders';
-import { ReadPreference, ReadPreferenceModeId } from './read_preference';
-import { ReadConcern, ReadConcernLevelId } from './read_concern';
+import { ReadPreference, ReadPreferenceMode } from './read_preference';
+import { ReadConcern, ReadConcernLevel } from './read_concern';
 import { W, WriteConcern } from './write_concern';
 import { MongoParseError } from './error';
 import {
@@ -818,7 +818,7 @@ export const OPTIONS = {
     transform({ values: [level], options }) {
       return ReadConcern.fromOptions({
         ...options.readConcern,
-        level: level as ReadConcernLevelId
+        level: level as ReadConcernLevel
       });
     }
   },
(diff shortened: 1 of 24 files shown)


src/mongo_client.ts
```diff
@@ -34,16 +34,16 @@ import type { Encrypter } from './encrypter';
...
+} as const);
 
 /** @public */
-export type ServerApiVersionId = typeof ServerApiVersion[keyof typeof ServerApiVersion];
+export type ServerApiVersion = typeof ServerApiVersion[keyof typeof ServerApiVersion];
 
 /** @public */
 export interface ServerApi {
-  version: string | ServerApiVersionId;
+  version: string | ServerApiVersion;
```
dariakp, line 46: @emadum Do you happen to know why this is typed as `string | ServerApiVersion`? The api version is also a string, so this typing is equivalent to just `string` - do we actually want to allow arbitrary strings here?

emadum: I don't recall a reason, we don't want to allow arbitrary strings so I think `string` can be removed.

dariakp: Actually, it looks like it's currently assigned when we parse the connection options on the condition that it is a string... do we want to restrict the option validation so that only the api versions known to the driver are supported? or do we want users to be able to specify any version regardless of whether it's known to the driver?

dariakp: Cross referencing the spec:
it looks like the `string | ServerApiVersion` given in the spec for the class is intended to accommodate languages that can't provide an enum; since we can provide an enum via typescript, it's fine for us to just have the enum, however in light of:
>  the driver MUST validate (e.g. when the application provides a version string to the ServerApi class) that the version string is valid and trigger a client-side error if an unknown API version was used

I'm going to add the check and error in the connection parser since typescript doesn't provide run time validation

src/connection_string.ts
```diff
@@ -590,10 +591,24 @@ export const OPTIONS = {
...
+      if (!versionToValidate) {
+        throw new MongoParseError(
+          `Invalid \`serverApi\` property; must specify a version as one of the following strings: "${Object.values(
+            ServerApiVersion
+          ).join('", "')}"`
+        );
+      }
+      if (!Object.values(ServerApiVersion).some(v => v === versionToValidate)) {
+        throw new MongoParseError(
+          `Invalid server API version=${versionToValidate}; must be one of the following strings: "${Object.values(
```
nbbeeken, line 606: Just a suggestion, maybe we could tighten up the error message here a bit:
```
Invalid server API version=${versionToValidate}; must be of the following: "a", "b"
```
It avoids the issue with plural and I think the quotes imply the type must be string, along with our documentation saying as much.

nbbeeken (changes requested):
A suggestion, we can discuss:

emadum (approved):
LGTM

nbbeeken (approved):
Nice error message change! 👍

PR #2810, opened 2021-05-18
fix(NODE-3245): mark symbols as internal remove from type definitions

nbbeeken:
marked as internal:
- Topology
- Connection
- Server
- Connection
- ConnectionPool

I added a linter step to our outputted definitions file that removes the unused imports.
The event types had to be changed because there was a circular reference error. We are still type checking ourselves here though with the events, since we use, for example, the `Connection.COMMAND_STARTED` variable throughout the code, there would be type errors if this variable changes from the now hardcoded `ConnectionEvents` type.
The Denque fix looks strange but the goal was to remove the Denque import from our type definitions and in order to do so we need to not use the `import Denque = require('denque')` syntax. Trying to enable es6 import syntax results in an error that requires you to enable `esModuleInterop` but that uses a helper from tslib which we don't want to add as a dependency. 

**NOTE:** Most of the interesting changes are in the mongodb.d.ts file, mainly the removal of the large types listed above. It would be ideal to take a look there to see if there's anymore symbols that should be omitted from the public API.


emadum:
> The Denque fix looks strange but the goal was to remove the Denque import from our type definitions and in order to do so we need to not use the import Denque = require('denque') syntax. Trying to enable es6 import syntax results in an error that requires you to enable esModuleInterop but that uses a helper from tslib which we don't want to add as a dependency.

Looks like this caused a lint error:

`Error: src/change_stream.ts:204:3 - (ae-forgotten-export) The symbol "DenqueLike" needs to be exported by the entry point index.d.ts`

durran (changes requested):
Looks like `EventEmitterWithState`, `RTTPinger`, `RTTPingerOptions` are only used by classes we consider internal - so maybe we should mark these as internal as well?

nbbeeken:
@durran RTT* things have been made internal, the `EventEmitterWithState` cannot be made internal because it is depended upon by our topology and server events, however the `stateChanged` event can be made internal. Good suggestions!

durran (approved):
LGTM. Nice work!

emadum (approved):
LGTM 👍

src/change_stream.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
dariakp, line 1: this comment should not be necessary anymore along with its instances in the other affected files

dariakp (changes requested):

tsconfig.json
```diff
@@ -7,10 +7,11 @@
     "target": "ES2018",
     "module": "commonJS",
     "moduleResolution": "node",
+    "esModuleInterop": true,
     "lib": ["ES2018"],
     // We don't make use of tslib helpers, all syntax used is supported by target engine
     "importHelpers": false,
-    "noEmitHelpers": true,
+    "noEmitHelpers": false,
```
dariakp, line 14: were you intending to commit the changes to this file?

src/operations/connect.ts
```diff
@@ -72,10 +72,7 @@ function createTopology(
   // save the reference to the topology on the client ASAP if the event handlers need to access it
   mongoClient.topology = topology;
 
-  // TODO(NODE-3273) - remove error
-  topology.once(Topology.OPEN, (error, topology) =>
-    mongoClient.emit(Topology.OPEN, error, topology)
-  );
+  topology.once(Topology.OPEN, () => mongoClient.emit('open', mongoClient));
```
nbbeeken, line 75: Here's the breaking change that is necessary to keep Topology `@internal` and not exposed in the type definitions. This is arguably a more useful API as the topology object that was emitted here doesn't have any of the useful, db/collections APIs most desire to use.

tsconfig.json
nbbeeken: Fixed! thanks

dariakp (changes requested):
Need to figure out what we want to do with the remaining unused line in the built ts file; if we can't automatically fix the eslint@typescript-eslint/no-unused-vars rule, we can either 1) manually remove it via regex as an extra build step or 2) leave it be for now

dariakp (approved):
:shipit:

PR #2811, opened 2021-05-19
Replace assert.equal() with assert.strictEqual()

naval7871:
## Description
README.md- Replace asserted.equal() with assert.strictEqual() as assert.equal() is deprecated in Node js latest versions. 

In assert.equal(actual, expected[,message]), the actual and expected parameters are using  "Abstract Equality Comparison ( == )" instead of "===".

**What changed?**
Replaced asserted.equal() with assert.strictEqual() in README.md.

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ const dbName = 'myproject';
 const client = new MongoClient(url);
 // Use connect method to connect to the server
 client.connect(function(err) {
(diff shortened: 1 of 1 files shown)


nbbeeken:
Looks like someone has already suggested these changes posted in #2798 so we'll defer to that one. Regardless, thank you for contributing!

PR #2812, opened 2021-05-19
fix(client): add isConnected method

j:
## Description

Added `isConnected` to client for backwards compatibility to 3.x.

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -478,6 +478,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
   }
 
   /**
+   * Check if MongoClient is connected
+   */
+  isConnected(): boolean {
+    return !!this.topology?.isConnected();
+  }
+
+  /**
    * Connect to MongoDB using a url
    *
    * @remarks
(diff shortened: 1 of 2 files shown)


emadum:
Hi @j, thanks for the PR! We removed `MongoClient.isConnected` intentionally in version 4.0, because the meaning is ambiguous. For a more detailed explanation, see our documentation for the [Unified Topology Design](http://mongodb.github.io/node-mongodb-native/3.6/reference/unified-topology/). It looks like we missed deprecating it in 3.6 though, so thanks for bringing this issue to our attention. I'm going to close this PR for now, but if you have any further questions please feel free to ask!

j:
@emadum oh, wow I just saw that!  So I can make all of my "ODM"(ish) [type-mongodb](https://github.com/j/type-mongodb) setup be synchronous.  I dig it.  Thanks.

j:
@emadum Ah, I just saw that tests still use connect, so not sure if I need to recant my assumption.

emadum:
@j It's still necessary to call the async `MongoClient.connect` method initially for now, but it'll happen automatically at some point in the future; removing `isConnected` is just taking the first step in that direction.

PR #2813, opened 2021-05-20
fix(cmap): undo flipping of `beforeHandshake` flag for timeout errors

vkarpov15:
## Description

I ran into this issue when upgrading Mongoose to 3.6.7 (https://github.com/Automattic/mongoose/pull/10265). 3.6.7 did some refactoring with https://github.com/mongodb/node-mongodb-native/commit/f1afcc4efbc41ce436812a6bfa22843e939ab5cf that seems to have unintentionally flipped the `beforeHandshake` flag on timeout errors. This causes the topology to _always_ switch to 'Unknown' state when there's an operation timeout.

**What changed?**

Undoing a change that I think may have been accidental.

### DIFF
--- a/lib/cmap/connection.js
+++ b/lib/cmap/connection.js
@@ -122,7 +122,7 @@ class Connection extends EventEmitter {
       if (issue.isTimeout) {
         op.cb(
           new MongoNetworkTimeoutError(`connection ${this.id} to ${this.address} timed out`, {
-            beforeHandshake: !!this[kIsMaster]
+            beforeHandshake: this[kIsMaster] == null
           })
         );
       } else if (issue.isClose) {
--- a/lib/cmap/connection.js
+++ b/lib/cmap/connection.js
@@ -122,7 +122,7 @@ class Connection extends EventEmitter {
       if (issue.isTimeout) {
         op.cb(
           new MongoNetworkTimeoutError(`connection ${this.id} to ${this.address} timed out`, {
-            beforeHandshake: this[kIsMaster] == null
+            beforeHandshake: this.ismaster == null
           })
         );
       } else if (issue.isClose) {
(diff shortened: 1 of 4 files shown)


nbbeeken (approved):
Hey Val, thanks so much for catching this I want to expedite this fix a bit so I've taken the liberty of adding regression tests and making a small change to the way the flag is set on the error class.

Also JIRA tracking: [NODE-3305](https://jira.mongodb.org/browse/NODE-3305)

lib/core/error.js
```diff
@@ -102,9 +102,7 @@ class MongoNetworkError extends MongoError {
     super(message);
     this.name = 'MongoNetworkError';
 
-    if (options && options.beforeHandshake === true) {
-      this[kBeforeHandshake] = true;
-    }
+    this[kBeforeHandshake] = !!(options && options.beforeHandshake === true);
```
dariakp, line 105: Just to confirm, do we actually want to explicitly set the `kBeforeHandshake` symbol on all network errors regardless of whether the options/property is passed in?

lib/cmap/connection.js
```diff
@@ -122,7 +122,7 @@ class Connection extends EventEmitter {
       if (issue.isTimeout) {
         op.cb(
           new MongoNetworkTimeoutError(`connection ${this.id} to ${this.address} timed out`, {
-            beforeHandshake: !!this[kIsMaster]
+            beforeHandshake: this[kIsMaster] == null
```
dariakp, line 125: @nbbeeken what were the cases where `!!this[kIsMaster]` was yielding an incorrect value? we should try to cover them in the tests

emadum (approved):
LGTM 👍

dariakp (approved):
LGTM, thanks!

PR #2814, opened 2021-05-20
fix(NODE-3183,NODE-3249): bring versioned API impl up to date

addaleax:
## Description

**What changed?**

Bring the versioned API code up to date by addressing NODE-3183 and
NODE-3249.

Specifically:
- Bump the versioned API tests to the latest version using
  `./etc/update-spec-tests.sh versioned-api`.
- Manually apply the SDAM updates from
  https://github.com/mongodb/specifications/commit/9946950d65a6f79fdb3746bde2701d3632f9b399
  (because a full update would have also come with load balancer
  changes that do not pass yet).
- Update the test runner slightly to account for the fact that some
  fields may have only a `$$unsetOrMatches` operator as its contents
  and thus the set of expected keys in an “expected” object is
  not a fixed-size set. (This was necessary in order for the updated
  versioned-api tests to pass)
- Remove the special check for getMore and transaction commands, and
  now pass the apiVersion for them as well when it is specified.
- Use the `hello` command instead of `isMaster` when an API version is
  specified; make sure that the slightly different response format
  does not break anything by setting the `ismaster` property manually
  in that case.
- Updating the mock server message handlers to account for both types
  of handshakes, everywhere.

I mean, the spec tests were really just updated automatically. The mock tests updates were performed semi-automatically, i.e. I did an automatic replacement and then went through all update sites + all message handlers manually to verify that they were not invalid changes. (Some of the tests are using replies like `mock.DEFAULT_ISMASTER_36`. I’m not sure if it makes sense to actually update those, considering that the name implies a MongoDB 3.6 response. I don’t think it hurts, though.)

### DIFF
--- a/src/cmap/connect.ts
+++ b/src/cmap/connect.ts
@@ -86,7 +86,7 @@ function performInitialHandshake(
   }
 
   const authContext = new AuthContext(conn, credentials, options);
-  prepareHandshakeDocument(conn.serverApi, authContext, (err, handshakeDoc) => {
+  prepareHandshakeDocument(authContext, (err, handshakeDoc) => {
     if (err || !handshakeDoc) {
       return callback(err);
     }
@@ -158,13 +158,10 @@ export interface HandshakeDocument extends Document {
   saslSupportedMechs?: string;
 }
 
-function prepareHandshakeDocument(
-  serverApi: ServerApi | undefined,
-  authContext: AuthContext,
-  callback: Callback<HandshakeDocument>
-) {
+function prepareHandshakeDocument(authContext: AuthContext, callback: Callback<HandshakeDocument>) {
   const options = authContext.options;
   const compressors = options.compressors ? options.compressors : [];
+  const { serverApi } = authContext.connection;
 
   const handshakeDoc: HandshakeDocument = {
     [serverApi?.version ? 'hello' : 'ismaster']: true,
--- a/src/cmap/connect.ts
+++ b/src/cmap/connect.ts
@@ -13,7 +13,6 @@ import {
 } from './wire_protocol/constants';
 import type { Document } from '../bson';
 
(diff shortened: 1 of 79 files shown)


src/cmap/connect.ts
```diff
@@ -85,7 +86,7 @@ function performInitialHandshake(
   }
 
   const authContext = new AuthContext(conn, credentials, options);
-  prepareHandshakeDocument(authContext, (err, handshakeDoc) => {
+  prepareHandshakeDocument(conn.serverApi, authContext, (err, handshakeDoc) => {
```
mbroadst, line 89: nit: why not store the `serverApi` in the `AuthContext`?

addaleax: Err, right, I assumed that the auth context would really only store auth info, but since the connection itself is also stored on it, that’s not even necessary :)

Thank you for pointing this out, fixed (and simplified) :)

durran (approved):
LGTM. I think the suggestions I would have had are part of NODE-3281 and don't need to happen here.

nbbeeken (commented):
There seems to be a failure still, it is the same test it just happens to be one of the duplicated tests meant to test the unified test runner. 
`A successful find event with a getmore and the server kills the cursor` in `test/spec/unified-test-format/valid-pass/poc-command-monitoring.yml`

I tried debugging a bit and it appears the server isn't killing the cursor (which I believe is returning a 0 cursor id) [so this line](https://github.com/mongodb/node-mongodb-native/blob/4.0/src/cursor/abstract_cursor.ts#L718) is kind of the source of the error. I'll check with other drivers if this test is an issue for all of us.

nbbeeken (approved):
NODE-3308
It is a shared failure so we can go ahead and merge this. It LGTM!

emadum (approved):
LGTM 🚀

PR #2817, opened 2021-05-25
refactor!: remove strict mode and callback from db.collection helper

emadum:
NODE-2752

WIP

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -311,77 +311,24 @@ export class Db {
   }
 
   /**
-   * Fetch a specific collection (containing the actual collection information). If the application does not use strict mode you
-   * can use it without a callback in the following way: `const collection = db.collection('mycollection');`
+   * Fetch a specific collection (containing the actual collection information).
+   * e.g. `const collection = db.collection('mycollection');`
    *
    * @param name - the collection name we wish to access.
-   * @returns return the new Collection instance if not in strict mode
+   * @returns return the new Collection instance
    */
   collection<TSchema extends Document = Document>(name: string): Collection<TSchema>;
   collection<TSchema extends Document = Document>(
     name: string,
-    options: CollectionOptions
-  ): Collection<TSchema>;
-  collection<TSchema extends Document = Document>(
-    name: string,
-    callback: Callback<Collection<TSchema>>
-  ): void;
-  collection<TSchema extends Document = Document>(
-    name: string,
-    options: CollectionOptions,
-    callback: Callback<Collection<TSchema>>
-  ): void;
-  collection<TSchema extends Document = Document>(
-    name: string,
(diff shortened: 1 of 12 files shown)


nbbeeken (changes requested):
All small points, I marked the skipped tests I found so to track them but maybe your reasoning for skipping and not deleting is the same for all of them?

And some docs/test suggestions

test/functional/collection.test.js
```diff
@@ -166,7 +163,7 @@ describe('Collection', function () {
       });
     });
 
-    it('should ensure strict access collection', function (done) {
+    it.skip('should ensure strict access collection', function (done) {
```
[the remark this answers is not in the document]
emadum: Oops thanks! I meant to delete these.

src/db.ts
```diff
@@ -311,77 +311,24 @@ export class Db {
...
-    if (typeof callback !== 'function') {
-      throw new MongoError(
-        `A callback is required in strict mode. While getting collection ${name}`
-      );
+    options?: CollectionOptions
+  ): Collection<TSchema> {
+    if (!options) {
+      options = {};
+    } else if (typeof options === 'function') {
+      throw new TypeError('The callback form of this helper has been removed.');
```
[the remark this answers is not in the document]
dariakp: I think this will likely go in first so I'll make a note to rebase my branch and check for any added TypeError/MongoError

test/functional/operation_example.test.js
```diff
@@ -3365,7 +3365,7 @@ describe('Operation Examples', function () {
    * @example-class Db
    * @example-method collection
    */
-  it('shouldCorrectlyAccessACollection', {
+  it.skip('shouldCorrectlyAccessACollection', {
```
dariakp, line 3368: why is this test being skipped?

test/functional/operation_promises_example.test.js
```diff
@@ -2588,7 +2588,7 @@ describe('Operation (Promises)', function () {
    * @example-class Db
    * @example-method collection
    */
-  it('shouldCorrectlyAccessACollectionWithPromises', {
+  it.skip('shouldCorrectlyAccessACollectionWithPromises', {
```
dariakp, line 2591: why is this test being skipped?

test/functional/operation_example.test.js
emadum: They are using the deprecated form of the helper; I think they should probably just be deleted rather than rewritten since they are just providing examples, not coverage, and the helper is much more self-explanatory in its behavior now. Does that sound reasonable?

dariakp: I don't mind deleting, since we need to consolidate/reorganize our test suite anyway, as long as we do have the actual coverage from elsewhere

PR #2820, opened 2021-05-25
feat(NODE-2751): add arrayFilters builder to bulk FindOperators

emadum:
NODE-2751

### DIFF
--- a/test/functional/bulk.test.js
+++ b/test/functional/bulk.test.js
@@ -1941,7 +1941,7 @@ describe('Bulk', function () {
   });
 
   it('should apply arrayFilters to bulk updates via FindOperators', {
-    metadata: { requires: { mongodb: '>= 3.4' } },
+    metadata: { requires: { mongodb: '>= 3.6' } },
     test: withMonitoredClient(['update', 'delete'], function (client, events, done) {
       client.db().dropCollection('bulkArrayFilters', () => {
         const coll = client.db().collection('bulkArrayFilters');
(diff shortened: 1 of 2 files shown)


durran (approved):

test/functional/bulk.test.js
```diff
@@ -1940,6 +1940,50 @@ describe('Bulk', function () {
...
+          .updateOne({ $set: { 'scores.$[e]': 1 } });
+        bulk
+          .find({ scores: { $gte: 10 } })
+          .arrayFilters([{ e: { $gte: 10 } }])
+          .update({ $set: { 'scores.$[e]': 10 } });
+
+        bulk.execute(err => {
+          expect(err).to.not.exist;
+          expect(events).to.be.an('array').with.length(1);
+          expect(events[0]).property('commandName').to.equal('update');
```
dariakp, line 1964: just a note: the `property` assertions can take the property name and value to compare directly, so this line is equivalent to `expect(events[0]).to.have.property('commandName', 'update')`; and if you go a step further and use the `nested` keyword, you could say `expect(events).to.have.nested.property('0.commandName', 'update')`

these are personal preferences, I just wanted to make you aware of the options in case any of them appeal to you

also be careful with the `length` assertion, since it can't be reliably chained in all cases (e.g., you can't chain it off `a`), so chai's own docs recommend to use `lengthOf` as a general rule, since that does work exactly as expected in all cases

nbbeeken (approved):

PR #2823, opened 2021-05-27
fix(NODE-2944): Reintroduce bson-ext support

nbbeeken:
- Update benchmark imports
- Add bson-ext test to CI

I've also updated the BSONSerializeOptions to use the options exported from our BSON library now, this pipes through the documentation from there so we don't have duplication. 

While I'm here, (and before the major release) how do we feel about renaming `BSONSerializeOptions` -> `BSONOptions`?
The truth is it is for both serialize and deserialize.

### DIFF
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -48,7 +48,9 @@ export interface BSONSerializeOptions
       | 'evalFunctions'
       | 'cacheFunctions'
       | 'cacheFunctionsCrc32'
+      | 'bsonRegExp'
       | 'allowObjectSmallerThanBufferSize'
+      | 'index'
     > {
   /** Return BSON filled buffers from operations */
   raw?: boolean;
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -1,10 +1,21 @@
-// import type * as _BSON from 'bson';
-// let BSON: typeof _BSON = require('bson');
-// try {
-//   BSON = require('bson-ext');
-// } catch {} // eslint-disable-line
+import type {
+  serialize as serializeFn,
+  deserialize as deserializeFn,
+  calculateObjectSize as calculateObjectSizeFn
+} from 'bson';
 
-// export = BSON;
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+let BSON = require('bson');
+try {
+  BSON = require('bson-ext');
+} catch {} // eslint-disable-line
+
+/** @internal */
+export const deserialize = BSON.deserialize as typeof deserializeFn;
+/** @internal */
+export const serialize = BSON.serialize as typeof serializeFn;
+/** @internal */
+export const calculateObjectSize = BSON.calculateObjectSize as typeof calculateObjectSizeFn;
 
 export {
   Long,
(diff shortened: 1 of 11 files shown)


emadum (approved):
LGTM 👍

test/unit/bson_import.test.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+const { expect } = require('chai');
+const BSON = require('../../src/bson');
+
+describe('BSON Library Import', function () {
```
dariakp, line 6: we discussed organizing the tests here a little differently: using require.resolve instead of require, and splitting the tests into 2 describe blocks - one for bson-ext and one for js-bson, each of which should have one it block for 'should import the correct library' and another it block for 'should correctly roundtrip supported data types'; using helpers and/or a data table structure should help minimize code duplication

src/bson.ts
```diff
@@ -21,38 +32,26 @@ export {
...
-  promoteBuffers?: boolean;
-  /** Promotes long values to number if they fit inside the 53 bits resolution */
-  promoteLongs?: boolean;
-  /** Serialize functions on any object */
-  serializeFunctions?: boolean;
-  /** Specify if the BSON serializer should ignore undefined fields */
-  ignoreUndefined?: boolean;
-
+export interface BSONSerializeOptions
+  extends Omit<SerializeOptions, 'index'>,
```
dariakp, line 45: we discovered that Serialize and Deserialize options end up being effectively the same here, so this code can be simplified accordingly

dariakp (changes requested):

src/bson.ts
nbbeeken: I looked over this too quickly in our discussion, there is a difference and these keys below are the ones we desire from the underlying BSON library:

serialize options: `checkKeys` `serializeFunctions` `ignoreUndefined`
deserialize options: `promoteLongs` `promoteBuffers` `promoteValues` `fieldsAsRaw`

I added a typescript test to cover this expectation.

test/unit/bson_import.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+    it('should correctly round trip Map', function () {
+      expect(BSON.Map).to.be.a('function');
+      const doc = { key: new BSON.Map([['2', 2]]) };
+      const outputDoc = BSON.deserialize(BSON.serialize(doc));
+      expect(outputDoc).to.have.nested.property('key.2', 2);
+    });
+  }
+
+  describe('bson-ext should be imported if it exists', function () {
```
dariakp, line 58: One tiny suggestion, can we make this describe just `'bson-ext'` and `'js-bson'` for the next one and then put the 'should be imported if it exists' into an `it` block with the corresponding expects? I just think it'll read better in the spec printout

dariakp (changes requested):

test/unit/bson_import.test.js
```diff
@@ -0,0 +1,85 @@
...
+        const typeCtor = BSON[type];
+        expect(typeCtor).to.be.a('function');
+        const doc = { key: new typeCtor(ctorArg) };
+        const outputDoc = BSON.deserialize(BSON.serialize(doc), options);
+        expect(outputDoc).to.have.property('key').that.is.instanceOf(typeCtor);
+        expect(outputDoc).to.deep.equal(doc);
+      });
+    }
+
+    it('should correctly round trip Map', function () {
```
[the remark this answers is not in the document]
nbbeeken: Deserialize doesn't return a map it returns an object so I can't reuse the deep equal check. I started by adding an if statement to change the asserts for just map but it looked more confusing than it was worth so I gave it its own test.

test/unit/bson_import.test.js
nbbeeken: Indeed, did this 👍

test/functional/insert.test.js
```diff
@@ -1577,7 +1577,7 @@ describe('Insert', function () {
         var collection = db.collection('bson_types_insert_1');
 
         var document = {
-          symbol: new BSONSymbol('abcdefghijkl'),
```
nbbeeken, line 1580: BSONExt supports deserializing but not serializing this type (which is acceptable for deprecated types). So to make the bson-ext tests pass I've updated this to a string.

dariakp (approved):
LGMT :)

PR #2825, opened 2021-05-28
fix(NODE-3272)!: emit correct event type when SRV Polling

nbbeeken:
Corrects the event type emitted from the SRV polling listener on the Topology.

Also:
Skipped the cursor tests and marked them with the unskip ticket, and corrected the visibility on session properties.

### DIFF
--- a/test/tools/utils.js
+++ b/test/tools/utils.js
@@ -246,6 +246,18 @@ class EventCollector {
   }
 }
 
+function getSymbolFrom(target, symbolName, assertExists = true) {
+  const symbol = Object.getOwnPropertySymbols(target).filter(
+    s => s.toString() === `Symbol(${symbolName})`
+  )[0];
+
+  if (assertExists && !symbol) {
+    throw new Error(`Did not find Symbol(${symbolName}) on ${target}`);
+  }
+
+  return symbol;
+}
+
 module.exports = {
   EventCollector,
   makeTestFunction,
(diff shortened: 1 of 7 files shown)


test/tools/utils.js
```diff
@@ -246,12 +246,25 @@ class EventCollector {
   }
 }
 
+function getSymbolFrom(target, symbolName, assertExists = true) {
```
emadum, line 249: Cool util! 🎉

emadum (approved):
LGTM! 👍  Looks like this also fixes NODE-3269?

test/tools/utils.js
nbbeeken: Thanks, and good catch on the other ticket I'll be sure to note it in the merge

test/unit/sdam/topology.test.js
```diff
@@ -314,5 +319,108 @@ describe('Topology (unit)', function () {
...
+              .to.be.a('object')
+              .with.property('address', 'fake:2');
+          });
+
+          topology.s.srvPoller.emit(
+            SrvPoller.SRV_RECORD_DISCOVERY,
+            new SrvPollingEvent([{ priority: 1, weight: 1, port: 2, name: 'fake' }])
+          );
+
+          // The srv event starts a monitor that we don't actually want running
```
dariakp, line 375: let's move the cleanup into an afterEach and iterate over all servers

dariakp (changes requested):

test/unit/sdam/topology.test.js
nbbeeken: Yes, good idea 🚀

dariakp (approved):
lgtm

PR #2827, opened 2021-06-02
refactor(NODE-3283): better types for mapped cursors

emadum:
NODE-3283


addaleax (approved):

test/types/mongodb.test-d.ts
```diff
@@ -11,3 +15,23 @@ expectDeprecated(Collection.prototype.remove);
 expectDeprecated(Collection.prototype.count);
 expectDeprecated(MongoError.create);
 expectDeprecated(AggregationCursor.prototype.geoNear);
+
+// test mapped cursor types
+const client = new MongoClient(process.env.MONGODB_URI || 'mongodb://localhost:27017');
```
emadum, line 20: Not sure if this is the best approach, definitely open to suggestions about better ways of testing this.

nbbeeken: You can just pass in an empty string to satisfy typescript since the file is not actually executed

test/types/mongodb.test-d.ts
```diff
@@ -11,3 +15,23 @@ expectDeprecated(Collection.prototype.remove);
...
+const coll = client.db('test').collection('test');
+const findCursor = coll.find();
+expectType<Promise<Document | null>>(findCursor.next());
```
nbbeeken, line 23: You can use `await` at the top level here, it would just cut down on the boilerplate `Promise<...>` wrapping.

src/cursor/abstract_cursor.ts
```diff
@@ -469,7 +469,7 @@ export abstract class AbstractCursor<
    *
```
nbbeeken, line 469:
```suggestion
   * @remarks
   *  
   * **NOTE:** adding a transform changes the return type of the iteration of this cursor, it **does not** return
   * a new instance of a cursor. This means when calling map, you should always assign the result to a new 
   * variable. Take note of the following example:
   * 
   * ```typescript
   * // fill in example
   * ```
```

Let's add some documentation here to call out the edge case we discussed and about best usage of this API. I think it would also be good for the manual docs to call this out as well, but that will be its own docs ticket.
I started a bit above here, but feel free to adjust as needed. 
Can you check that the documentation is threaded down to the subclasses? If you head over to your test file and hover over map for find/aggregate cursor do see these docs? We might need to replicate this on each class if not, which I think would be an acceptable cost.

nbbeeken (changes requested):
Some docs and test clean up, typings look good 👍

nbbeeken (approved):
LGTM!

PR #2828, opened 2021-06-02
feat(NODE-3325): support 'let' option for aggregate command

addaleax:
## Description

**What changed?**

Add support for `let` and cherry-pick the tests from https://github.com/mongodb/specifications/commit/3db387995bc0e64df862d621aaa31387afc1537c
(since actually updating the full test suite led to a bunch of failures).
Since the NODE ticket is scoped to only `aggregate`, nothing
else is done here.

**Are there any files to ignore?**

Test files are cherry-picked from upstream.

PR #2829, opened 2021-06-07
feat(NODE-3333): support 'let' option for CRUD commands

addaleax:
## Description

**What changed?**

Same as #2828.

The spec tests here are fake in the sense that no upstream spec tests
for this exist at this point. I’ve put them next to the other tests, more or less
just assuming that somebody would notice their existence when updating
the crud spec suite to upstream and make sure they aren’t deleted.
(They are just a temporary thing here anyway.)

### DIFF
--- a/test/spec/crud/unified/crud-let.yml
+++ b/test/spec/crud/unified/crud-let.yml
@@ -48,7 +48,7 @@ tests:
 
   - description: "Find with let option unsupported (server-side error)"
     runOnRequirements:
-      - minServerVersion: "2.6.0"
+      - minServerVersion: "4.2.0"
         maxServerVersion: "4.4.99"
     operations:
       - name: find
@@ -95,7 +95,7 @@ tests:
 
   - description: "FindOneAndUpdate with let option unsupported (server-side error)"
     runOnRequirements:
-      - minServerVersion: "2.6.0"
+      - minServerVersion: "4.2.0"
         maxServerVersion: "4.4.99"
     operations:
       - name: findOneAndUpdate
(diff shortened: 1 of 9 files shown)


test/spec/crud/unified/crud-let.yml
```diff
@@ -0,0 +1,221 @@
...
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                find: *collection0Name
+                filter: *query0
+                let: *let0
+
+  - description: "Find with let option unsupported (server-side error)"
```
nbbeeken, line 49: Looks like the error string returned changes on versions older than 4.2, and on versions older than 3.6 there's even more different error messages. 
I think it is acceptable we set the minumum version to 4.2 since these are temporary anyway.

addaleax: Sure, sounds good :+1:

nbbeeken:
Looks like the aggregate-let tests are failing, there's an update in the spec that I think fixes it. Or you could skip it for now in:
`test/functional/crud_spec.test.js:447`
change to:
```ts
await runUnifiedTest(this, crudSpecTest, test, ['Aggregate with let option']);
```

addaleax:
Err, sure, I’ll update the tests – it’s unrelated to the rest of the changes here, though – let me know if I should push them to a separate PR?

nbbeeken (approved):
I can close the related ticket (NODE-3334) and link this PR thank you so much for that, just trying to keep the waterfall 🟢  where we can :)

dariakp (approved):

PR #2832, opened 2021-06-07
fix(NODE-1502): Command monitoring commands should be copies

W-A-James:
## Description
Changed events for Command monitoring to hold copies of the commands instead of the commands themselves to prevent changing state in event monitoring by modifying `extractCommand` function in src/cmap/command_monitoring_events.ts

**What changed?**
- src/cmap/command_monitoring_events.ts
- test/functional/apm.test.js
- test/functional/multiple_db.test.js
- test/unit/cmap/command_monitoring_events.test.js

### DIFF
--- a/test/functional/apm.test.js
+++ b/test/functional/apm.test.js
@@ -660,7 +660,9 @@ describe('APM', function () {
 
   // NODE-1502
   it('should not allow mutation of internal state from commands returned by event monitoring', {
-    metadata: { requires: { mongodb: '>= 3.6.0' } },
+    metadata: {
+      /*requires: { mongodb: '>= 3.6.0' } */
+    },
     test: function () {
       const self = this;
       const started = [];
@@ -671,27 +673,28 @@ describe('APM', function () {
       );
       client.on('commandStarted', filterForCommands('insert', started));
       client.on('commandSucceeded', filterForCommands('insert', succeeded));
+      let documentToInsert = { a: { b: 1 } };
+      return client
+        .connect()
+        .then(client => {
+          const db = client.db(self.configuration.db);
+          return db.collection('apm_test').insertOne(documentToInsert);
+        })
+        .then(r => {
+          expect(r).to.have.property('insertedId').that.is.an('object');
+          expect(started.length).to.equal(1);
+          // Check if contents of returned document are equal to document inserted (by value)
+          expect(documentToInsert).to.deep.equal(started[0].command.documents[0]);
+          // Check if the returned document is a clone of the original. This confirms that the
+          // reference is not the same.
+          expect(documentToInsert !== started[0].command.documents[0]).to.equal(true);
+          expect(documentToInsert.a !== started[0].command.documents[0].a).to.equal(true);
 
-      return client.connect().then(client => {
-        const db = client.db(self.configuration.db);
-        let documentToInsert = { a: { b: 1 } };
-        return db
-          .collection('apm_test')
-          .insertOne(documentToInsert)
-          .then(r => {
-            expect(r).property('insertedId').to.exist;
-            expect(started.length).to.equal(1);
-            expect(documentToInsert).to.eql(started[0].command.documents[0]);
-            // Check if the returned document is a clone of the original
-            expect(documentToInsert).to.not.equal(started[0].command.documents[0]);
-            expect(documentToInsert.a).to.not.equal(started[0].command.documents[0].a);
-
-            expect(started[0].commandName).to.equal('insert');
-            expect(started[0].command.insert).to.equal('apm_test');
-            expect(succeeded.length).to.equal(1);
-            return client.close();
-          });
-      });
+          expect(started[0].commandName).to.equal('insert');
+          expect(started[0].command.insert).to.equal('apm_test');
+          expect(succeeded.length).to.equal(1);
+          return client.close();
+        });
     }
   });
 
--- a/test/functional/apm.test.js
+++ b/test/functional/apm.test.js
@@ -659,37 +659,40 @@ describe('APM', function () {
   });
 
   // NODE-1502
-  it('should not allow mutation of internal state from commands returned by event monitoring', function () {
-    const self = this;
-    const started = [];
-    const succeeded = [];
-    const client = self.configuration.newClient(
-      { writeConcern: { w: 1 } },
-      { maxPoolSize: 1, monitorCommands: true }
-    );
-    client.on('commandStarted', filterForCommands('insert', started));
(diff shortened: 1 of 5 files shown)


test/functional/apm.test.js
```diff
@@ -658,6 +658,46 @@ describe('APM', function () {
...
+          // reference is not the same.
+          expect(documentToInsert !== started[0].command.documents[0]).to.equal(true);
+          expect(documentToInsert.a !== started[0].command.documents[0].a).to.equal(true);
+
+          expect(started[0].commandName).to.equal('insert');
+          expect(started[0].command.insert).to.equal('apm_test');
+          expect(succeeded.length).to.equal(1);
+          return client.close();
+        });
+    }
```
nbbeeken, line 698: Leaving a note to discuss some mock tests.

nbbeeken (changes requested):

test/functional/apm.test.js
```diff
@@ -658,6 +658,46 @@ describe('APM', function () {
...
+      let documentToInsert = { a: { b: 1 } };
+      return client
+        .connect()
+        .then(client => {
+          const db = client.db(self.configuration.db);
+          return db.collection('apm_test').insertOne(documentToInsert);
+        })
+        .then(r => {
+          expect(r).to.have.property('insertedId').that.is.an('object');
+          expect(started.length).to.equal(1);
```
nbbeeken, line 685:
```suggestion
          expect(started).to.have.lengthOf(1);
```

test/functional/apm.test.js
```diff
@@ -658,6 +658,46 @@ describe('APM', function () {
...
+          // Check if contents of returned document are equal to document inserted (by value)
+          expect(documentToInsert).to.deep.equal(started[0].command.documents[0]);
+          // Check if the returned document is a clone of the original. This confirms that the
+          // reference is not the same.
+          expect(documentToInsert !== started[0].command.documents[0]).to.equal(true);
+          expect(documentToInsert.a !== started[0].command.documents[0].a).to.equal(true);
+
+          expect(started[0].commandName).to.equal('insert');
```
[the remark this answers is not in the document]
W-A-James: As far as these assertions are concerned, we want to do these to ensure that the strict equality fails which would mean we're dealing with different object references.
You're right in saying that I could try some mutation and seeing if mutating one affects the other since that would have the same effect as those two tests, but just be more explicit in how it relates to the test name, so I'll get on that.

test/unit/cmap/command_monitoring_events.test.js
```diff
@@ -0,0 +1,78 @@
...
+      if (doc.ismaster || doc.hello) {
+        request.reply(mock.DEFAULT_ISMASTER_36);
+      }
+
+      // blackhole all other requests
+    });
+
+    connect({ connectionType: Connection, hostAddress: server.hostAddress() }, (err, conn) => {
+      expect(err).to.be.undefined;
+      expect(conn).to.not.be.undefined;
```
nbbeeken, line 39: I think we can drop stubbing this connection part, and the mock server stuff if we pass:
```
{ id: 'someId', address: 'someHost' }
```
That looks like it would be accepted by the CommandStartedEvent constructor.

test/unit/cmap/command_monitoring_events.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+      commands.forEach(c => {
```
nbbeeken, line 41: Great job enumerating the cases here!
We can break up the reporting of these a bit for some logical modularity by moving the `it` to inside the loop:

```
for (const command of commands) {
  it(`testing ${command}...`, () => {
    // code from for loop
  })
}
```

test/unit/cmap/command_monitoring_events.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const commands = [
+      new Query('admin.$cmd', { a: { b: 10 }, $query: { b: 10 } }, {}),
+      new Query('hello', { a: { b: 10 }, $query: { b: 10 } }, {}),
+      new Msg('admin.$cmd', { b: { c: 20 } }, {}),
+      new Msg('hello', { b: { c: 20 } }, {}),
+      new GetMore('admin.$cmd', Long.fromNumber(10)),
+      new GetMore('hello', Long.fromNumber(10)),
+      new KillCursor('admin.$cmd', [Long.fromNumber(100), Long.fromNumber(200)]),
+      new KillCursor('hello', [Long.fromNumber(100), Long.fromNumber(200)])
+    ];
```
nbbeeken, line 26: Isn't there one more case here where it's none of these classes? I'm thinking about line 201 and forward of the extractCommand function.

nbbeeken (changes requested):

test/unit/cmap/command_monitoring_events.test.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(ev.command.cursors).to.deep.equal(c.cursorIds);
+      } else if (typeof c === 'object') {
+        if (c.ns === 'admin.$cmd') {
+          expect(ev.command !== c.query.$query).to.equal(true);
+          for (const k in c.query.$query) {
+            expect(ev.command[k]).to.deep.equal(c.query.$query[k]);
+          }
+        }
+      }
+      done();
```
nbbeeken, line 62: No need to use done anymore, there's no async callback here.

test/unit/cmap/command_monitoring_events.test.js
```diff
@@ -0,0 +1,65 @@
...
+    new Msg('hello', { b: { c: 20 } }, {}),
+    new GetMore('admin.$cmd', Long.fromNumber(10)),
+    new GetMore('hello', Long.fromNumber(10)),
+    new KillCursor('admin.$cmd', [Long.fromNumber(100), Long.fromNumber(200)]),
+    new KillCursor('hello', [Long.fromNumber(100), Long.fromNumber(200)]),
+    { ns: 'admin.$cmd', query: { $query: { a: 16 } } },
+    { ns: 'hello there', f1: { h: { a: 52, b: { c: 10, d: [1, 2, 3, 5] } } } }
+  ];
+
+  commands.forEach(c => {
```
nbbeeken, line 22: tiniest nit: can we use for-of loop here, it's completely stylistic but its friendlier when debugging.

nbbeeken (changes requested):

nbbeeken (approved):
LGTM

durran (approved):

dariakp (approved):

PR #2833, opened 2021-06-08
test(NODE-3118): run CSFLE integration tests from libmongocrypt repo

nbbeeken:
The custom CSFLE test variant will now run the mocha tests defined in the libmongocrypt/bindings/node directory.

I also skipped the count documents test because the command monitoring exposed the fact that count is run with an aggregate and not a count operation.


dariakp (approved):
LGTM

emadum (approved):
LGTM

PR #2834, opened 2021-06-09
fix(NODE-3335): do not validate explain verbosity in client

rose-m:
As referenced in NODE-3335 the client should not validate the values of explain verbosity but let the server handle the validation. This PR removes the validation check of the string values (and only verifies whether it's a boolean we can auto-transform or a string).


nbbeeken (approved):
Changes look good to me.
FYI for the team I've made a ticket to backport this: NODE-3377

durran (approved):

emadum (approved):

PR #2835, opened 2021-06-10
fix(NODE-2035): exceptions thrown from awaited cursor forEach do not propagate

W-A-James:
## Description
Catch and propagate exceptions in awaited cursor.forEach

**What changed?**
- src/cursor/abstract\_cursor.ts
- test/functional/cursor.js

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -324,12 +324,12 @@ export abstract class AbstractCursor<
           if (doc == null) return done();
           let result;
           // NOTE: no need to transform because `next` will do this automatically
-          //try {
-          result = iterator(doc); // TODO(NODE-3283): Improve transform typing
-          /*} catch (e) {
+          try {
+            result = iterator(doc); // TODO(NODE-3283): Improve transform typing
+          } catch (e) {
             return done(e);
           }
-          */
+
           if (result === false) return done();
 
           // these do need to be transformed since they are copying the rest of the batch
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -324,20 +324,25 @@ export abstract class AbstractCursor<
           if (doc == null) return done();
           let result;
           // NOTE: no need to transform because `next` will do this automatically
-          try {
-            result = iterator(doc); // TODO(NODE-3283): Improve transform typing
-          } catch (e) {
+          //try {
+          result = iterator(doc); // TODO(NODE-3283): Improve transform typing
+          /*} catch (e) {
             return done(e);
           }
+          */
           if (result === false) return done();
 
           // these do need to be transformed since they are copying the rest of the batch
           const internalDocs = this[kDocuments].splice(0, this[kDocuments].length);
           if (internalDocs) {
             for (let i = 0; i < internalDocs.length; ++i) {
-              result = iterator(
-                (transform ? transform(internalDocs[i]) : internalDocs[i]) as T // TODO(NODE-3283): Improve transform typing
-              );
+              try {
+                result = iterator(
+                  (transform ? transform(internalDocs[i]) : internalDocs[i]) as T // TODO(NODE-3283): Improve transform typing
+                );
+              } catch (e) {
+                return done(e);
+              }
               if (result === false) return done();
             }
           }
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -321,7 +321,6 @@ export abstract class AbstractCursor<
       const fetchDocs = () => {
         next<T>(this, true, (err, doc) => {
           if (err || doc == null) return done(err);
-          if (doc == null) return done();
           let result;
           // NOTE: no need to transform because `next` will do this automatically
           try {
@@ -334,17 +333,15 @@ export abstract class AbstractCursor<
 
           // these do need to be transformed since they are copying the rest of the batch
           const internalDocs = this[kDocuments].splice(0, this[kDocuments].length);
-          if (internalDocs) {
-            for (let i = 0; i < internalDocs.length; ++i) {
-              try {
-                result = iterator(
-                  (transform ? transform(internalDocs[i]) : internalDocs[i]) as T // TODO(NODE-3283): Improve transform typing
-                );
-              } catch (error) {
-                return done(error);
-              }
-              if (result === false) return done();
+          for (let i = 0; i < internalDocs.length; ++i) {
+            try {
+              result = iterator(
+                (transform ? transform(internalDocs[i]) : internalDocs[i]) as T // TODO(NODE-3283): Improve transform typing
+              );
+            } catch (error) {
+              return done(error);
             }
+            if (result === false) return done();
           }
 
           fetchDocs();
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -322,9 +322,13 @@ export abstract class AbstractCursor<
         next<T>(this, true, (err, doc) => {
           if (err || doc == null) return done(err);
           if (doc == null) return done();
-
+          let result;
           // NOTE: no need to transform because `next` will do this automatically
-          let result = iterator(doc); // TODO(NODE-3283): Improve transform typing
+          try {
+            result = iterator(doc); // TODO(NODE-3283): Improve transform typing
+          } catch (e) {
+            return done(e);
+          }
           if (result === false) return done();
 
           // these do need to be transformed since they are copying the rest of the batch
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -326,8 +326,8 @@ export abstract class AbstractCursor<
           // NOTE: no need to transform because `next` will do this automatically
           try {
             result = iterator(doc); // TODO(NODE-3283): Improve transform typing
-          } catch (e) {
-            return done(e);
+          } catch (error) {
+            return done(error);
           }
 
           if (result === false) return done();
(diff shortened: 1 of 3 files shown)


src/cursor/abstract_cursor.ts
```diff
@@ -322,18 +322,27 @@ export abstract class AbstractCursor<
         next<T>(this, true, (err, doc) => {
           if (err || doc == null) return done(err);
           if (doc == null) return done();
-
+          let result;
           // NOTE: no need to transform because `next` will do this automatically
-          let result = iterator(doc); // TODO(NODE-3283): Improve transform typing
+          try {
+            result = iterator(doc); // TODO(NODE-3283): Improve transform typing
+          } catch (e) {
```
nbbeeken, line 329: nit: prefer using `error` for the variable name here

src/cursor/abstract_cursor.ts
```diff
@@ -322,18 +322,27 @@ export abstract class AbstractCursor<
...
           if (internalDocs) {
             for (let i = 0; i < internalDocs.length; ++i) {
-              result = iterator(
-                (transform ? transform(internalDocs[i]) : internalDocs[i]) as T // TODO(NODE-3283): Improve transform typing
-              );
+              try {
+                result = iterator(
+                  (transform ? transform(internalDocs[i]) : internalDocs[i]) as T // TODO(NODE-3283): Improve transform typing
+                );
+              } catch (e) {
```
nbbeeken, line 343: nit: prefer using `error` for the variable name here

test/functional/cursor.test.js
```diff
@@ -3901,6 +3901,43 @@ describe('Cursor', function () {
...
+            promiseResult = await cursor
+              .forEach(() => {
+                throw new Error('FAILURE IN FOREACH CALL');
+              })
+              .catch(err => {
+                expect(err.message).to.deep.equal('FAILURE IN FOREACH CALL');
+              });
+          } catch (err) {
+            expect(err).to.be.undefined;
+          }
```
nbbeeken, line 3932:
```suggestion
        const cursor = collection.find({});
        expect(() => {
          cursor
              .forEach(() => {
                throw new Error('FAILURE IN FOREACH CALL');
              })
              .catch(err => {
                expect(err.message).to.deep.equal('FAILURE IN FOREACH CALL');
              });
        }).to.not.throw();
```

We can reduce some of the logic here, if we pass a function to chai with expect(...).to.not.throw(); then we can confirm that the forEach doesn't throw synchronously and still check that the error is correctly propagated to the catch callback. (ie. its part of the promise chain).

test/functional/cursor.test.js
```diff
@@ -3901,6 +3901,43 @@ describe('Cursor', function () {
...
+          expect(promiseResult).to.be.undefined;
+          cursor.close();
+          client.close();
```
nbbeeken, line 3935: If the suggested change above is made make sure you still return the client.close() promise.

nbbeeken (changes requested):
Nice job finding the fix, some clean up and its looking pretty much ready

test/functional/cursor.test.js
addaleax: Fwiw, it might also be nice here to check that the `catch` handler is actually called?

test/functional/cursor.test.js
```diff
@@ -3901,6 +3901,42 @@ describe('Cursor', function () {
...
+        const cursor = collection.find({});
+        this.defer(() => cursor.close());
+        expect(() => {
+          cursor
+            .forEach(() => {
+              throw new Error('FAILURE IN FOREACH CALL');
+            })
+            .then(
+              () => {
+                expect(false).to.equal(true, 'Failed to catch error thrown in awaited forEach');
```
nbbeeken, line 3929:
```suggestion
                expect.fail('Failed to catch error thrown in awaited forEach');
```

test/functional/cursor.test.js
```diff
@@ -3901,6 +3901,42 @@ describe('Cursor', function () {
...
+              },
+              err => {
+                expect(err.message).to.deep.equal('FAILURE IN FOREACH CALL');
+              }
+            );
+        }).to.not.throw();
+      })
+      .catch(console.error);
```
nbbeeken, line 3937: Are the console usages here for debugging? We should merge assertions if we want to catch something here.

nbbeeken (changes requested):

test/functional/cursor.test.js
W-A-James: I had forgotten to remove the console.error since I had put in there when running into so issues when writing the test so I had it there just so I could figure out where errors were coming from. As far as I know, it is no longer needed.

nbbeeken (approved):

src/cursor/abstract_cursor.ts
```diff
@@ -322,18 +322,27 @@ export abstract class AbstractCursor<
         next<T>(this, true, (err, doc) => {
           if (err || doc == null) return done(err);
           if (doc == null) return done();
```
dariakp, line 324: just a note that L324 seems redundant with L323

src/cursor/abstract_cursor.ts
```diff
@@ -322,18 +322,27 @@ export abstract class AbstractCursor<
...
           if (result === false) return done();
 
           // these do need to be transformed since they are copying the rest of the batch
           const internalDocs = this[kDocuments].splice(0, this[kDocuments].length);
           if (internalDocs) {
             for (let i = 0; i < internalDocs.length; ++i) {
-              result = iterator(
-                (transform ? transform(internalDocs[i]) : internalDocs[i]) as T // TODO(NODE-3283): Improve transform typing
-              );
+              try {
```
dariakp, line 339: since splice always returns an array, the nesting here can be simplified by removing the "if" check in L337

test/functional/cursor.test.js
```diff
@@ -3901,6 +3901,49 @@ describe('Cursor', function () {
     });
   });
 
+  // NODE-2035
+  it('should propagate error when exceptions are thrown from an awaited forEach call', function () {
```
dariakp, line 3905: is there any way we can simplify this test? I realize that the other tests in this file use similar syntax, but the better practice would be to outsource shared setup into a beforeEach (establishing the client, creating the collection) and the cleanup into an afterEach (closing the client and cursor connections) instead of the defer syntax here

dariakp (changes requested):

test/functional/cursor.test.js
W-A-James: Would it be acceptable for now to create a new describe block with a `beforeEach` setup block and `afterEach` teardown block just for this test at the moment?

emadum (changes requested):
Nice work! 

In addition to the changes in the afterEach hook, I think the test's readability would be improved by being rewritten using async/await instead of Promise syntax, now that it's available to use in driver v4. Totally up to you whether to do that though, the test LGTM otherwise.

emadum (approved):

test/functional/cursor.test.js
```diff
@@ -3901,6 +3901,60 @@ describe('Cursor', function () {
...
+    // NODE-2035
+    it('should propagate error when exceptions are thrown from an awaited forEach call', function () {
+      const docs = [{ a: 1 }, { a: 2 }, { a: 3 }];
+      let collection;
+      return client
+        .connect()
+        .then(
+          () => {
+            let db = client.db(configuration.db);
+            collection = db.collection('cursor_session_tests2');
```
dariakp, line 3928: since creating a collection is still part of the setup, I would suggest moving that into the before block, too; furthermore, I'm not sure how important it is for this collection to be unique for our use case; it seems that you are just trying to limit the number of returned documents to a known quantity, if that is accurate, you can use the same collection name as the others, and then just set a unique key and use the find filter to retrieve everything with that unique key set (e.g., `{ my_unique_test_key: 1 }, { my_unique_test_key: 2 }`); if it *is* important for the collection to be unique, then you'd want to use the before block to drop it and recreate it; otherwise you might still have left over data from previous test runs

I also agree with Eric's suggestion to use async await to simplify the flow of the code here and further down; you could use a combination of async await and catch to have something like `await cursor.forEach(() => throw err).catch(err => { // expect err.message; })` if the error is thrown synchronously, it would fail the test, and if the expect fails, it would also fail the test, but if the error is passed to the callback, things would resolve successfully

dariakp (changes requested):

test/functional/cursor.test.js
```diff
@@ -3901,6 +3901,50 @@ describe('Cursor', function () {
...
+      const docs = [{ unique_key_2035: 1 }, { unique_key_2035: 2 }, { unique_key_2035: 3 }];
+      await collection.insertMany(docs).catch(() => {
+        expect.fail('Failed to insert documents');
+      });
+      cursor = collection.find({
+        unique_key_2035: {
+          $exists: true
+        }
+      });
+      expect(async function () {
```
dariakp, line 3935: I don't think you actually need this expect, since if it does throw, the test will fail (mocha tests will fail on any uncaught error, not just "expect"; the expect library is just a handy way to generate errors or sometimes add more information)

you do, however, need to have a .then() before the .catch() where you can put an expect.fail() in case, somehow, the forEach resolves instead of rejecting

dariakp (changes requested):

W-A-James:
Rebased onto 4.0

test/functional/cursor.test.js
W-A-James: Thanks for the heads up. That was a good catch.

test/functional/cursor.test.js
```diff
@@ -3901,6 +3901,51 @@ describe('Cursor', function () {
...
+      });
+      await cursor
+        .forEach(() => {
+          throw new Error('FAILURE IN FOREACH CALL');
+        })
+        .then(() => {
+          expect.fail('Error in forEach call not caught');
+        })
+        .catch(err => {
+          console.log(err.message);
```
dariakp, line 3943: extra console log, otherwise LGTM :)

dariakp (changes requested):

dariakp (approved):

PR #2837, opened 2021-06-11
test(NODE-3288): sync command-monitoring spec tests

dariakp:
## Description

NODE-3288

see also #2838


test/functional/apm.test.js
```diff
@@ -9,6 +9,7 @@ const {
 const { loadSpecTests } = require('../spec');
```
durran, line 9: Maybe we rename this to `command_monitoring.test.js` to match the directory name change? Or would you rather do that as part of the greater test cleanup? I'm fine either way.

durran (approved):

test/functional/apm.test.js
dariakp: I thought about doing it, but (1) I wasn't sure if we had in progress work touching this file and what potential merge conflict issues that might cause and (2) there are a lot of references to APM throughout and as you pointed out, the greater test refactor later is likely going to change a lot of things around, so not sure it's worth doing it

durran: Yeah I think it's fine to wait. LGTM.

nbbeeken (approved):

PR #2838, opened 2021-06-11
test(NODE-3288): sync command-monitoring spec tests to 3.6

dariakp:
## Description

NODE-3288

See also:
#2837


test/functional/apm.test.js
```diff
@@ -9,6 +9,7 @@ const ignoreNsNotFound = shared.ignoreNsNotFound;
 const loadSpecTests = require('../spec').loadSpecTests;
```
durran, line 9: Maybe we rename this to `command_monitoring.test.js` to match the directory name change? Or would you rather do that as part of the greater test cleanup? I'm fine either way.

durran (approved):

test/functional/apm.test.js
dariakp: same thoughts here as in my reply on the other PR, but I don't mind if you think it's worth doing now

durran: Same as well. LGTM.

nbbeeken (changes requested):
LGTM, However some failures on latest are from these spec tests can we skip?

emadum:
> LGTM, However some failures on latest are from these spec tests can we skip?

Some of these are the ones we already skipped in 4.0, there is a commit referenced in NODE-3308 with the tests to skip.

nbbeeken (approved):

PR #2840, opened 2021-06-14
fix(NODE-3150): allow retrieving PCRE-style RegExp

andymina:
## Description
- Allowed retrieving PCRE-style RegExp by adding `bsonRegExp` option to commands, collections, and dbs
- Added unit tests for the `bsonRegExp` option

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -588,6 +588,9 @@ export const OPTIONS = {
   autoEncryption: {
     type: 'record'
   },
+  bsonRegExp: {
+    type: 'boolean'
+  },
   serverApi: {
     target: 'serverApi',
     transform({ values: [version] }): ServerApi {
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -55,35 +55,6 @@ import type { AutoEncrypter } from '../deps';
 import type { ServerApi } from '../mongo_client';
 import { TypedEventEmitter } from '../mongo_types';
 
-// Used for filtering out fields for logging
-const DEBUG_FIELDS = [
-  'reconnect',
-  'reconnectTries',
-  'reconnectInterval',
-  'emitError',
-  'cursorFactory',
-  'host',
-  'port',
-  'size',
-  'keepAlive',
-  'keepAliveInitialDelay',
-  'noDelay',
-  'connectionTimeout',
-  'checkServerIdentity',
-  'socketTimeoutMS',
-  'ssl',
-  'ca',
-  'crl',
-  'cert',
-  'key',
-  'rejectUnauthorized',
-  'promoteLongs',
-  'promoteValues',
-  'promoteBuffers',
-  'bsonRegExp',
-  'servername'
-];
-
 const stateTransition = makeStateMachine({
   [STATE_CLOSED]: [STATE_CLOSED, STATE_CONNECTING],
   [STATE_CONNECTING]: [STATE_CONNECTING, STATE_CLOSING, STATE_CONNECTED, STATE_CLOSED],
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -55,6 +55,35 @@ import type { AutoEncrypter } from '../deps';
 import type { ServerApi } from '../mongo_client';
 import { TypedEventEmitter } from '../mongo_types';
 
+// Used for filtering out fields for logging
+const DEBUG_FIELDS = [
+  'reconnect',
+  'reconnectTries',
+  'reconnectInterval',
+  'emitError',
+  'cursorFactory',
+  'host',
+  'port',
+  'size',
+  'keepAlive',
+  'keepAliveInitialDelay',
+  'noDelay',
+  'connectionTimeout',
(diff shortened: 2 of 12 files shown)


nbbeeken:
I don't have further background on this, we have a precedent of other values being auto converted with the `promoteValues` and like settings. However, I know it is a strict requirement that BSON libraries have an API that explicitly converts UUIDs to native types, so that is a precedence for the opposite (as well as the regex examples from other libs you mention). There is convenience in handing the user the native type, up until it fails because of the contents.

Maybe as a separate effort we should take a look at the BSON library catching a Regexp decoding error and using BSONRegex automatically instead of failing, although I imagine that would also need a flag.

durran:
Yeah it's a weird one since it is backwards breaking anyways so I guess we are stuck with a flag. Thanks for the clarification.

src/connection_string.ts
```diff
@@ -588,6 +588,9 @@ export const OPTIONS = {
   autoEncryption: {
     type: 'record'
   },
+  bsonRegExp: {
```
durran, line 591: I'm not sure we want this to be a connection string option as it's Node driver specific.

durran (changes requested):

durran:
I would also change the description not to mention `findOne` specifically, as the option is set on the command response before deserialization, which will actually take care of all commands (not just `findOne`).

src/connection_string.ts
andymina: I included the new bsonRegExp option here because this is where all possible options (uri, object param, and defaults) are defined so it might be a little overreaching to solve this issue now. I can create a new ticket to make sure you can’t use it in the connection string if that sounds better?

nbbeeken: Capturing some investigation from a slack convo:
Andy discovered we do support BSON options in the URI in 3.6 so the behavior is consistent, may be unusual, not sure what other drivers do but I think we can certainly defer changing that to another ticket.

test/unit/bson_regex.test.js
```diff
@@ -0,0 +1,68 @@
...
+  describe('bsonRegExp option', () => {
+    it('should respond with BSONRegExp class with option passed to db', async function () {
+      let client;
+      try {
+        // create and connect to client
+        client = this.configuration.newClient();
+        await client.connect();
+
+        const db = client.db('a', { bsonRegExp: true });
+        const collection = db.collection('b');
```
nbbeeken, line 16: We should use a better db/collection name here

test/unit/bson_regex.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+    it('should respond with BSONRegExp class with option passed to collection', async function () {
+      let client;
+      try {
+        // create and connect to client
+        client = this.configuration.newClient(); // bsonRegex
+        await client.connect();
+
+        const db = client.db('a');
+        const collection = db.collection('b', { bsonRegExp: true });
```
nbbeeken, line 35: The setup is nearly the same for each test, can we share some of this logic in a before / beforeEach.
Or I think we can make a loop that tests each of these, with only one implementation:

```javascript
const options = { bsonRegexp: true };
for (const passOptionsTo of ['client', 'db',  'collection', 'operation']) {
  it(`should respond with BSONRegExp class with option passed to ${passOptionsTo}`, async function () {
        client = this.configuration.newClient(passOptionsTo === 'client' ? options : undefined);
        await client.connect();

        const db = client.db('a', passOptionsTo === 'db' ? options : undefined);
        const collection = db.collection('b', passOptionsTo === 'collection' ? options : undefined);

        await collection.insertOne({ regex: new BSONRegExp('abc', 'imx') });
        const res = await collection.findOne(
          { regex: new BSONRegExp('abc', 'imx') },
          passOptionsTo === 'operation' ? options : undefined
        );
        
        // ....
  });
}
```
How does this look? needs some clean up but this is the idea.

nbbeeken (changes requested):
Some test clean up requests, driver changes LGTM

test/unit/bson_regex.test.js
durran: Looks good to me.

src/connection_string.ts
durran: I'm fine with that.

durran (approved):
I'm LGTM once @nbbeeken 's test change requests are in.

src/sdam/server.ts
```diff
@@ -55,6 +55,35 @@ import type { AutoEncrypter } from '../deps';
...
+  'crl',
+  'cert',
+  'key',
+  'rejectUnauthorized',
+  'promoteLongs',
+  'promoteValues',
+  'promoteBuffers',
+  'bsonRegExp',
+  'servername'
+];
```
nbbeeken, line 85: Looks like this got brought back in the rebase, I think this is removed now.

nbbeeken (approved):

dariakp (approved):
LGTM

emadum (approved):
LGTM

PR #2842, opened 2021-06-14
feat(NODE-3274): add type hinting for UpdateFilter

nbbeeken:
This brings in type checking / hinting for UpdateFilter. The tests in helper_types.test-d.ts and at the top of updateX.test-d.ts should cover all the new subtypes that make up the UpdateFilter type.

Additional changes:
- Updated `tsd`
  - Had to fix type errors picked up by tsd, writeConcern not defining defaults, deps.ts exports not correctly union-ed and then later narrowed
- Added running `tsd` to checking that the definitions are valid, since it is part of the linting process


test/types/community/collection/bulkWrite.test-d.ts
```diff
@@ -1,4 +1,4 @@
-import { expectError } from 'tsd';
+import { expectAssignable, expectError } from 'tsd';
```
dariakp, line 1: looks like expectAssignable is not used, and this is causing the updated build step to fail

src/write_concern.ts
```diff
@@ -88,7 +88,17 @@ export class WriteConcern {
     const parentOpts: WriteConcern | WriteConcernSettings | undefined =
       inherit instanceof WriteConcern ? inherit : inherit.writeConcern;
 
-    const { w, wtimeout, j, fsync, journal, wtimeoutMS } = { ...parentOpts, ...opts };
+    const {
+      w = undefined,
```
dariakp, line 92: why is the explicit defaulting necessary?

test/types/helper_types.test-d.ts
```diff
@@ -0,0 +1,72 @@
...
+expectAssignable<NumericType>(Long.fromInt(3));
+expectAssignable<NumericType>(new Decimal128('2.3'));
+expectAssignable<NumericType>(new Int32(23));
+expectAssignable<NumericType>(new Double(2.3));
+
+// ObjectFilterOperators
+expectAssignable<ObjectFilterOperators<{ a: number }>>({ a: { $eq: 2 } });
+
+// KeysOfAType - union of keys that are the second type argument
+expectType<KeysOfAType<{ a: number; b: string }, number>>('a');
```
dariakp, line 39: I am a little bit confused about the intent here, I think it's just the plurality of the naming - if KeysOfAType is really supposed to only apply to a single variable, wouldn't it make more sense to call it Key and not Keys? same question for the "KeysOfOtherType" helper

test/types/helper_types.test-d.ts
```diff
@@ -0,0 +1,72 @@
...
+
+// KeysOfOtherType - union of keys that aren't the second type argument
+expectType<KeysOfOtherType<{ a: number; b: string }, number>>('b');
+
+// AcceptedFields - essentially rewrites the type of keys that have the second argument type
```
dariakp, line 44: this one is somewhat confusing, too, along with its negation below

test/types/helper_types.test-d.ts
```diff
@@ -0,0 +1,72 @@
...
+expectAssignable<AcceptedFields<{ a: number; b: string }, number, boolean>>({ a: true });
+
+// NotAcceptedFields - prevents the usage of keys with an unacceptable type
+expectNotType<NotAcceptedFields<{ a: number; b: string }, number>>({ b: 'hello' }); // { b: never }
+
+// DotAndArrayNotation - map string index to type, for MongoDB's dotted and array keys
```
dariakp, line 50: should there be something about a key that doesn't have the dot and array notation?

test/types/helper_types.test-d.ts
```diff
@@ -0,0 +1,72 @@
...
+expectAssignable<DotAndArrayNotation<number>>({ 'dot.array.notation': 3 });
+expectNotType<DotAndArrayNotation<number>>({ 'dot.array.notation': true });
+
+// OnlyFieldsOfType - filters for keys of a type, and optionally replaces type (combo KeysOfAType & AcceptedFields)
+expectAssignable<OnlyFieldsOfType<{ a: number; b: string }, number>>({ a: 2 });
+expectAssignable<OnlyFieldsOfType<{ a: number; b: string }, string>>({ b: 'hello' });
+expectAssignable<OnlyFieldsOfType<{ a: number; b: string }, string, boolean>>({ b: true });
+
+// test generic schema, essentially we expect no safety here
+
```
dariakp, line 60: can we delete the whitespace here, also, not entirely clear what we're trying to test

test/types/helper_types.test-d.ts
```diff
@@ -0,0 +1,72 @@
...
+expectAssignable<OnlyFieldsOfType<Document, NumericType | undefined>>({ a: 2 });
+expectAssignable<OnlyFieldsOfType<Document, NumericType | undefined>>({ a: 'hello' });
+
+// If you use any at all, then we cannot do type hinting, is this fair?
```
dariakp, line 64: do we need to discuss this?

test/types/helper_types.test-d.ts
```diff
@@ -0,0 +1,72 @@
...
+expectAssignable<OnlyFieldsOfType<{ a: number; b: any }, NumericType | undefined>>({ a: 'hello' });
+
+// Using indexed types makes determining permitted keys tricky!
+interface IndexedSchema {
+  a: number;
+  [key: string]: string | number;
+}
+expectNotType<OnlyFieldsOfType<IndexedSchema, NumericType>>({ a: 'hello' });
```
dariakp, line 72: probably want to also have an expectType test case here, for clarity

src/mongo_types.ts
```diff
@@ -178,11 +185,136 @@ export type Projection<TSchema> = {
   Partial<Record<string, ProjectionOperators | 0 | 1 | boolean>>;
 
 /** @public */
-export type FlattenIfArray<T> = T extends ReadonlyArray<infer R> ? R : T;
+export type IsAny<Type, Yes, No> = true extends false & Type ? Yes : No;
```
dariakp, line 188: I'd like to talk through all the types defined here

dariakp (changes requested):

src/write_concern.ts
nbbeeken: Turns out it isn't. I was getting "error TS2525: Initializer provides no value for this binding element and the binding element has no default value" because the opts variable was of type `any` so it TS now warns that there isn't information about those properties from the variables being spread. The better solution is to correctly annotate `opts`.

dariakp:
looks like tsc actually doesn't like the lack of typing on write concern; other things we discussed fixing is making sure the update tests use 2 schemas: one strictly defined and one that has a key with <any> type; we want to make sure we clearly identify what we do and don't expect to be assignable in each case

nbbeeken:
@dariakp I think I've address all the comments, we discussed having more schema examples that would result in mostly duplicating the `test/types/community/collection/updateX.test-d.ts`. That isn't here because the result of looking into indexed schemas was that there isn't a way to preserve the key's narrowing. We can talk further offline, but this refers to the last example in `test/types/helper_types.test-d.ts`

dariakp (approved):

src/cmap/auth/gssapi.ts
```diff
@@ -71,6 +71,7 @@ function makeKerberosClient(authContext: AuthContext, callback: Callback<Kerbero
   if ('kModuleError' in Kerberos) {
     return callback(Kerberos['kModuleError']);
   }
+  const { initializeClient } = Kerberos;
```
emadum, line 74: What motivated pulling this out vs leaving it as `Kerberos.initializeClient`?

emadum (approved):
LGTM

src/cmap/auth/gssapi.ts
nbbeeken: Cus Typescript! tsd has its own version of TS which was a version newer and these lines were throwing errors. You'll notice the deps.ts file needed changes to make the types more accurate. Pulling out the function here after you've narrowed the type makes it exist properly inside the callback below. Where as Kerberos.initializeClient has the error that the property potentially doesn't exist. 

I can move the narrowing down into the callback too or just cast this problem away, open to diff solutions here

durran (approved):

src/cmap/auth/gssapi.ts
emadum: I think it's fine - slightly less readable, but not so bad as to necessitate a change.

PR #2843, opened 2021-06-15
fix(NODE-3150): added bsonRegExp option for v3.6

andymina:
## Description
- Allowed retrieving PCRE-style RegExp by adding bsonRegExp option to commands, collections, and dbs
- Added unit tests for the bsonRegExp option

### DIFF
--- a/test/unit/bson_regex.test.js
+++ b/test/unit/bson_regex.test.js
@@ -1,0 +1,74 @@
+'use strict';
+
+const expect = require('chai').expect;
+const BSON = require('bson');
+
+describe('BSONRegExp', () => {
+  describe('bsonRegExp option', () => {
+    it('should respond with BSONRegExp class with option passed to db', async function() {
+      let client;
+      try {
+        // create and connect to client
+        client = this.configuration.newClient();
+        await client.connect();
+
+        const db = client.db('a', { bsonRegExp: true });
+        const collection = db.collection('b');
+
+        await collection.insertOne({ regex: new BSON.BSONRegExp('abc', 'imx') });
+        const res = await collection.findOne({ regex: new BSON.BSONRegExp('abc', 'imx') });
+
+        expect(res)
+          .has.property('regex')
+          .that.is.instanceOf(BSON.BSONRegExp);
+      } finally {
+        await client.close();
+      }
+    });
+
+    it('should respond with BSONRegExp class with option passed to collection', async function() {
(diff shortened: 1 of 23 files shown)


durran (approved):

durran (approved):
Just note I did change the base branch to 3.7 from 3.6

nbbeeken (changes requested):
Just leaving a note so we can remember, there's some refactoring needed for the older node versions in the test. (i.e. cant use async)

nbbeeken (approved):

test/unit/bson_regex.test.js
```diff
@@ -0,0 +1,40 @@
...
+              { regex: new BSONRegExp('abc', 'imx') },
+              passOptionTo === 'operation' ? option : undefined
+            )
+          )
+          .then(res =>
+            expect(res)
+              .has.property('regex')
+              .that.is.instanceOf(BSONRegExp)
+          )
+          .then(() => client.close());
```
dariakp, line 36: in order to make sure the test actually does close the connection even if the test fails, please include a rejection handler, too; you can pass in a done parameter in L11, so this line would become:
`.then(() => client.close(done), (err) => client.close(() => done(err))`

dariakp (changes requested):

test/unit/bson_regex.test.js
andymina: just added!

dariakp (approved):

test/unit/bson_regex.test.js
```diff
@@ -0,0 +1,43 @@
...
+describe('BSONRegExp', () => {
+  describe('bsonRegExp option', () => {
+    // define option for tests to use
+    const option = { bsonRegExp: true };
+    for (const passOptionTo of ['client', 'db', 'collection', 'operation']) {
+      it(`should respond with BSONRegExp class with option passed to ${passOptionTo}`, function(done) {
+        const client = this.configuration.newClient(passOptionTo === 'client' ? option : undefined);
+        let collection;
+
+        return client
```
dariakp, line 15: since we are using done, this `return` should be removed, otherwise mocha gets confused

dariakp (changes requested):

dariakp (approved):

emadum (approved):
LGTM 👍

PR #2845, opened 2021-06-15
test(NODE-2856): Check that defaultTransactionOptions get used from session

W-A-James:
## Description
Test verifying that transactions get their `readPreference` from `SessionOptions` passed at session instantiation

**What changed?**
- lib/core/topologies/read_preference.js
- test/functional/readpreference.test.js

### DIFF
--- a/test/functional/readpreference.test.js
+++ b/test/functional/readpreference.test.js
@@ -761,11 +761,10 @@ describe('ReadPreference', function() {
   });
 
   it('should use session readPreference instead of client readPreference', {
-    metadata: { requires: { topology: ['single', 'replicaset'] } },
+    metadata: { requires: { unifiedTopology: true, topology: ['single', 'replicaset'] } },
     test: function(done) {
       const configuration = this.configuration;
       const client = this.configuration.newClient(configuration.writeConcernMax(), {
-        useUnifiedTopology: true,
         readPreference: 'primaryPreferred'
       });
 
@@ -773,8 +772,8 @@ describe('ReadPreference', function() {
         this.defer(() => {
           client.close();
         });
-        expect(err).to.be.null;
-        expect(client).to.not.be.null;
+        expect(err).to.not.exist;
+        expect(client).to.exist;
         const session = client.startSession({
           defaultTransactionOptions: { readPreference: 'secondary' },
           causalConsistency: true
@@ -782,7 +781,7 @@ describe('ReadPreference', function() {
 
         session.startTransaction();
         const result = ReadPreference.resolve(client, { session: session });
-        expect(result).to.not.be.undefined;
+        expect(result).to.exist;
         expect(result.mode).to.deep.equal('secondary');
         session.abortTransaction();
 
--- a/test/functional/readpreference.test.js
+++ b/test/functional/readpreference.test.js
@@ -795,8 +795,7 @@ describe('ReadPreference', function() {
         });
         session.startTransaction();
         const result = ReadPreference.resolve(client, { session: session });
-        expect(result).to.exist;
-        expect(result.mode).to.deep.equal('secondary');
+        expect(result).to.have.property('mode', 'secondary');
       }
     });
   });
--- a/test/functional/readpreference.test.js
+++ b/test/functional/readpreference.test.js
@@ -760,33 +760,44 @@ describe('ReadPreference', function() {
     })
   });
 
-  it('should use session readPreference instead of client readPreference', {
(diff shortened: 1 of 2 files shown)


test/functional/readpreference.test.js
```diff
@@ -759,4 +759,35 @@ describe('ReadPreference', function() {
         });
     })
   });
+
+  it('should use session readPreference instead of client readPreference', {
+    metadata: { requires: { topology: ['single', 'replicaset'] } },
+    test: function(done) {
+      const configuration = this.configuration;
+      const client = this.configuration.newClient(configuration.writeConcernMax(), {
+        useUnifiedTopology: true,
```
emadum, line 768: In general we try to avoid forcing the unified topology like this, and instead filter based on it. You can do this via the test metadata: `requires: { unifiedTopology: false, topology: ... }`.

test/functional/readpreference.test.js
```diff
@@ -759,4 +759,35 @@ describe('ReadPreference', function() {
...
+        readPreference: 'primaryPreferred'
+      });
+
+      client.connect((err, client) => {
+        this.defer(() => {
+          client.close();
+        });
+        expect(err).to.be.null;
+        expect(client).to.not.be.null;
```
emadum, line 777:
```suggestion
        expect(err).to.not.exist;
        expect(client).to.exist;
```
These "exists" expectations are safer than checking against null/undefined when you don't care about distinguishing them.

test/functional/readpreference.test.js
```diff
@@ -759,4 +759,35 @@ describe('ReadPreference', function() {
...
+        const session = client.startSession({
+          defaultTransactionOptions: { readPreference: 'secondary' },
+          causalConsistency: true
+        });
+
+        session.startTransaction();
+        const result = ReadPreference.resolve(client, { session: session });
+        expect(result).to.not.be.undefined;
```
emadum, line 785:
```suggestion
        expect(result).to.exist;
```

emadum (changes requested):
A few small requests, but overall LGTM! 👍

emadum (approved):
LGTM! 👍

test/functional/readpreference.test.js
```diff
@@ -759,4 +759,34 @@ describe('ReadPreference', function() {
...
+          causalConsistency: true
+        });
+
+        session.startTransaction();
+        const result = ReadPreference.resolve(client, { session: session });
+        expect(result).to.exist;
+        expect(result.mode).to.deep.equal('secondary');
+        session.abortTransaction();
+
+        session.endSession(undefined, done);
```
dariakp, line 788: I strongly recommend using beforeEach/afterEach for setup and clean up to ensure that the clean up step gets executed even if some assertions inside the test fail; for example, if your expect fails at L785, it's going to result in a thrown error and hence interrupt the flow of the code; 

to limit the impact of this change, feel free to use an extra describe block around this 'it' and declare a `let client` and `let session` in that scope; the beforeEach can take a done argument, so you should be able to write something like `beforeEach(done => { client = < config setup code >; client.connect(err => done(err); })` then inside the `it` block, have `session = client.startSession(...)` and so forth until L785; then define `afterEach(done => { if (session) { session.abortTransaction(() => session.endSession(() => client.close(() => done())))} else { client.close(() => done() } })`

dariakp (changes requested):

test/functional/readpreference.test.js
W-A-James: Thanks for this. That cleared up a lot of why we use the `beforeEach` and `afterEach`.

test/functional/readpreference.test.js
```diff
@@ -759,4 +759,45 @@ describe('ReadPreference', function() {
...
+      metadata: { requires: { unifiedTopology: true, topology: ['single', 'replicaset'] } },
+      test: function() {
+        session = client.startSession({
+          defaultTransactionOptions: { readPreference: 'secondary' },
+          causalConsistency: true
+        });
+        session.startTransaction();
+        const result = ReadPreference.resolve(client, { session: session });
+        expect(result).to.exist;
+        expect(result.mode).to.deep.equal('secondary');
```
dariakp, line 799: by the way, you can use the `property` assertion to check the property and the value, instead of the combination of the exist and the equal checks, it usually gives a nicer message on failure, like so: `expect(result).to.have.property('mode', 'secondary');`

dariakp (commented):
just a small suggested improvement on the assertion

dariakp (approved):
LGTM :)

nbbeeken (approved):
LGTM!

PR #2850, opened 2021-06-16
feat(NODE-3191): backport versioned api

emadum:
NODE-3191

Includes backports for the following `4.0` PRs:
 - #2736 - versioned api (NODE-2950)
 - #2764  - update estimated document count for v1 api (NODE-3074)
 - #2814 - versioned api updates
   - NODE-3183 - Use "hello" command when API Version is declared
   - NODE-3249 - Require passing Versioned API options to getMore and transaction-continuing commands


test/functional/cursor.test.js
```diff
@@ -4117,58 +4119,64 @@ describe('Cursor', function() {
...
   });
 
-  it('Correctly decorate the collection cursor count command with skip, limit, hint, readConcern', {
-    // Add a tag that our runner can trigger on
-    // in this case we are setting that node needs to be higher than 0.10.X to run
-    metadata: {
-      requires: { topology: ['single', 'replicaset', 'sharded', 'ssl', 'heap', 'wiredtiger'] }
-    },
+  // FIXME(NODE-3074): should this be deleted?
+  it.skip(
```
nbbeeken, line 4123: Maybe NODE-3369 should be made more general. I looked into python and it looks like they only use the aggregation method of counting if a certain wire version is met, but I guess the tests don't/can't cover that.

test/unit/core/single/sessions.test.js
```diff
@@ -484,10 +484,15 @@ describe('Sessions (Single)', function() {
...
         cursor._next(function(err) {
           expect(err).to.not.exist;
+          expect(session.id).to.exist;
 
           cursor.kill(err => {
             expect(err).to.not.exist;
-            commands.forEach(command => expect(command.lsid).to.eql(session.id));
+
+            commands.forEach(command => {
+              if (command.killCursors) return;
```
nbbeeken, line 493: Is this a new change with versioned API?

test/unit/sessions/collection.test.js
```diff
@@ -59,9 +59,9 @@ describe('Sessions', function() {
         const options = Object.freeze({});
         test.server.setMessageHandler(request => {
           const doc = request.document;
-          if (doc.ismaster) {
+          if (doc.ismaster || doc.hello) {
             request.reply(mock.DEFAULT_ISMASTER_36);
-          } else if (doc.count || doc.endSessions) {
+          } else if (doc.count || doc.aggregate || doc.endSessions) {
```
nbbeeken, line 64: We could callout here `// count changed behavior to use agg 3.6.x+`, just a nit, thinking about our future selves.

package.json
```diff
@@ -98,7 +98,7 @@
     "check:tls": "mocha --opts '{}' test/manual/tls_support.test.js",
     "format": "npm run check:lint -- --fix",
     "release": "standard-version -i HISTORY.md",
-    "test": "npm run lint && mocha --recursive test/functional test/unit"
+    "test": "npm run check:lint && mocha --recursive test/functional test/unit"
```
nbbeeken, line 101: Has this just been broken? OMG

nbbeeken (changes requested):
Despite the size, a majority of this is accepting hello and legacy hello in the tests, most of what I callout here is minor requests.

package.json
emadum: Yeah not for too long though, I think since the unified test runner was merged.

test/unit/core/single/sessions.test.js
emadum: This test (which has been moved to `disabled/` in 4.0) started failing after the NODE-3183 backport, because `command.killCursors` doesn't have an lsid. I'll do a bit more investigation into what is happening with this test.

test/functional/unified-spec-runner/entities.ts
```diff
@@ -38,10 +38,22 @@ interface UnifiedChangeStream extends ChangeStream {
...
-    useUnifiedTopology: Boolean(process.env.MONGODB_UNIFIED_TOPOLOGY)
-  });
+const options: any = { 
+    useUnifiedTopology: Boolean(process.env.MONGODB_UNIFIED_TOPOLOGY), 
+    serverApi: serverApiConfig() 
+};
+  const serverApi = serverApiConfig();
+  if (serverApi) {
+    options.serverApi = serverApi;
+  }
```
nbbeeken, line 55: Can be removed because of the refactor above right?

nbbeeken (changes requested):

test/functional/buffering_proxy.test.js
```diff
@@ -138,7 +138,7 @@ describe.skip('Buffering Proxy', function() {
           if (die) {
             request.connection.destroy();
           } else {
-            if (doc.ismaster) {
+            if (doc.ismaster || doc.hello) {
```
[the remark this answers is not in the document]
nbbeeken: NODE-3373

nbbeeken (approved):
LGTM

test/unit/core/single/sessions.test.js
emadum: Turns out this was a bug I introduced, fixed now! 👍

durran (approved):

test/functional/crud_spec.test.js
```diff
@@ -476,3 +479,19 @@ describe('CRUD v2', function() {
 
   generateTopologyTests(testSuites, testContext);
 });
+
+describe('CRUD unified', function() {
```
dariakp, line 483: were the CRUD specs and the other specs that aren't explicitly versioned API pulled in as part of the versioned API implementation? just curious about the updates that don't seem to be directly related to versioned API itself

test/functional/cursor.test.js
```diff
@@ -4117,59 +4119,6 @@ describe('Cursor', function() {
     }
   });
 
-  it('Correctly decorate the collection cursor count command with skip, limit, hint, readConcern', {
```
dariakp, line 4120: why was this test deleted?

dariakp (commented):
@emadum @nbbeeken The PR looks fine to me, minus the couple of questions I had around the test changes

test/functional/crud_spec.test.js
emadum: This comes from the backport of NODE-3074, which adds tests in the new unified format for `estimatedDocumentCount`.

test/functional/view.test.js
```diff
@@ -58,6 +58,7 @@ describe('Views', function() {
           ) {
             expect(r).to.exist;
             expect(err).to.not.exist;
+            delete commandResult.apiVersion;
```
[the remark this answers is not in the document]
emadum: I wasn't sure about the best approach to take here. Below we currently use a deep equal check on the commandResult, which may or may not contain `apiVersion` depending on the presence of the MONGODB_API_VERSION environment variable.

We could use `containsSubset` instead of doing a deep equal check, but that allows the presence of extraneous keys we don't actually want to be there. Deleting the key seemed like a reasonable compromise, but I'm open to better ideas! Or perhaps just adding a comment would suffice.

test/functional/cursor.test.js
emadum: Currently taking another look at this test actually, deleting it may have been a mistake.

test/functional/view.test.js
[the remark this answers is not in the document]
dariakp: so there's a couple of different approaches possible here: 
1) if it is possible to know what the expected apiVersion is based on the test context, then we can optionally assert on it by pulling the expected object out into its own `const expectedResult = { create, viewOn, pipeline }` and then assigning the `apiVersion` with the expected value if we expect there to be one;
2) if it is possible to know whether or not the apiVersion is set at all, but not what it's set to, we could use the keys assertion (`expect(obj).to.have.all.keys([...])`) to check that all expected keys are present and then assert on the 3 that we know one at a time;
3) if it is not possible to know whether the apiVersion is even set, then just a comment to that end would be fine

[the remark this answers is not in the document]
emadum: Went with option 1 👍

test/functional/cursor.test.js
emadum: Yeah it was a bug that was causing this test to fail, fixed now.

dariakp (approved):
Thanks for all the test updates!

PR #2851, opened 2021-06-17
refactor(NODE-3356): Remove logs that duplicate command monitoring functionality

dariakp:
## Description

NODE-3356


nbbeeken (approved):

durran (approved):
I like this refactoring. It was always expected with the command monitoring spec that a logger itself would be registering itself to all our monitoring if the users wanted easy debug, and this is how we originally implemented it in the Ruby driver at least and I believe there was discussion for the same for Node in the past that just got deprioritised. This puts us in the right direction - nice job.

emadum (changes requested):
LGTM but one of the Logger tests is failing in CI.

dariakp:
> LGTM but one of the Logger tests is failing in CI.

Oh that's weird, I could have sworn I saw them pass earlier; I'll look into it, thanks for the catch! @emadum

emadum (approved):

PR #2852, opened 2021-06-17
fix(NODE-2035): Exceptions thrown from awaited cursor forEach do not propagate

W-A-James:
## Description
Catch and propagate exceptions in awaited cursor.forEach
**What changed?**
- lib/cursor.js
- test/functional/cursor.test.js

### DIFF
--- a/test/functional/cursor.test.js
+++ b/test/functional/cursor.test.js
@@ -4372,6 +4372,60 @@ describe('Cursor', function() {
     }
   );
 
+  describe('Cursor forEach Error propagation', function() {
+    let configuration;
+    let client;
+    let cursor;
+    let collection;
+
+    beforeEach(function() {
+      configuration = this.configuration;
+      client = configuration.newClient({ w: 1 }, { maxPoolSize: 1 });
+      return client
+        .connect()
+        .then(() => {
+          collection = client.db(configuration.db).collection('cursor_session_tests2');
+        })
+        .catch(() => {
+          expect.fail('Failed to connect to client');
+        });
+    });
+
+    // NODE-2035
+    it('should propagate error when exceptions are thrown from an awaited forEach call', function(done) {
+      const docs = [{ unique_key_2035: 1 }, { unique_key_2035: 2 }, { unique_key_2035: 3 }];
+      collection.insertMany(docs).then(
+        () => {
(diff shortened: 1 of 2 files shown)


test/functional/cursor.test.js
```diff
@@ -4372,6 +4372,71 @@ describe('Cursor', function() {
...
+        .catch(error => {
+          done(error);
+        });
+    });
+
+    afterEach(function(done) {
+      if (cursor) {
+        cursor.close().then(() => {
+          client.close().then(() => done());
+        });
```
nbbeeken, line 4399:
```suggestion
        cursor.close()
          .then(() => client.close())
          .then(() => done());
```
We can rely on chaining for a bit of clarity here.

nbbeeken (changes requested):
Excellent job translating the async/await to promises, one small chainable section missed in the afterEach

nbbeeken (approved):
LGTM

emadum (approved):
LGTM

test/functional/cursor.test.js
```diff
@@ -4372,6 +4372,72 @@ describe('Cursor', function() {
...
+                try {
+                  expect(err.message).to.deep.equal('FAILURE IN FOREACH CALL');
+                  done();
+                } catch (error) {
+                  done(error);
+                }
+              }
+            );
+        },
+        error => {
```
dariakp, line 4434: Instead of having the error here, I recommend moving it down to a .catch block; that would ensure that if the .forEach throws synchronously in L4417, that error will also be caught and passed to the done callback; if you keep the error handler as you do now on the top level then, it will only handle the error from insertMany(), and there would be nothing to handle sync errors from the success handler (because the L4422 error handler is only for explicit rejections on the forEach, not for any uncaught sync errors that could arise in the execution of the .forEach handler itself). Does that make sense?

dariakp (changes requested):

test/functional/cursor.test.js
W-A-James: Just tried out adding a throw in the forEach implementation itself, so that makes perfect sense. Thanks so much for pointing that out. Wouldn't have seen it otherwise. 👍🏿

dariakp (approved):

PR #2856, opened 2021-06-22
fix(NODE-3343): allow overriding result document after projection applied

nbbeeken:
This allows for users to override the resulting documents after a projection is applied.
A furhter improvement could be to use the projection to filter the keys out of the schema but that requires the projection to be a const type. I leave that up to a future feature request.

### DIFF
--- a/test/types/community/cursor.test-d.ts
+++ b/test/types/community/cursor.test-d.ts
@@ -65,12 +65,12 @@ typedCollection
   .map(x => x.name2 && x.age2);
 typedCollection.find({ name: '123' }, { projection: { age: 1 } }).map(x => x.tag);
 
-typedCollection.find().project({ name: 1 });
-typedCollection.find().project({ notExistingField: 1 });
-typedCollection.find().project({ max: { $max: [] } });
-
-// $ExpectType Cursor<{ name: string; }>
-typedCollection.find().project<{ name: string }>({ name: 1 });
+expectType<FindCursor<{ name: string }>>(typedCollection.find().project({ name: 1 }));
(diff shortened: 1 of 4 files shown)


addaleax (approved):

test/types/community/cursor.test-d.ts
```diff
@@ -65,12 +65,12 @@ typedCollection
   .map(x => x.name2 && x.age2);
 typedCollection.find({ name: '123' }, { projection: { age: 1 } }).map(x => x.tag);
 
-typedCollection.find().project({ name: 1 });
-typedCollection.find().project({ notExistingField: 1 });
-typedCollection.find().project({ max: { $max: [] } });
-
-// $ExpectType Cursor<{ name: string; }>
-typedCollection.find().project<{ name: string }>({ name: 1 });
+expectType<FindCursor<{ name: string }>>(typedCollection.find().project({ name: 1 }));
```
dariakp, line 68: is there a way to add at least one expectNotType test here?

dariakp (changes requested):

test/types/community/cursor.test-d.ts
dariakp: Also it looks like a number of assertions are failing in the pipeline

nbbeeken: Added two! pipeline should be clear, of course clarity is key so let me know if there's more I can add to cover more use cases.

dariakp (approved):
LGTM

emadum (approved):

durran (approved):

PR #2857, opened 2021-06-23
fix(NODE-2905): support SERVICE_NAME authentication mechanism property

rose-m:
This PR adds support for the `SERVICE_NAME` authentication mechanism property that replaces the old `gssapiServiceName` query string parameter.

It also includes a test case that would set the `SERVICE_NAME` to an invalid value and verifies the connection consequently fails.

_NOTE:_ I've also started a small README file for the manual tests as I was initially unsure on how to run them locally.

Let me know if there's anything I need to adapt 😊

### DIFF
--- a/src/cmap/auth/gssapi.ts
+++ b/src/cmap/auth/gssapi.ts
@@ -6,6 +6,7 @@ import type { Document } from '../../bson';
 
 type MechanismProperties = {
   gssapiCanonicalizeHostName?: boolean;
+  SERVICE_NAME?: string;
 };
 
 import * as dns from 'dns';
@@ -72,15 +73,14 @@ function makeKerberosClient(authContext: AuthContext, callback: Callback<Kerbero
     return callback(Kerberos['kModuleError']);
   }
 
-  const { username, password, mechanismProperties } = credentials;
-  const serviceName =
-    mechanismProperties['gssapiservicename'] ||
-    mechanismProperties['gssapiServiceName'] ||
-    'mongodb';
+  const { username, password } = credentials;
+  const mechanismProperties = credentials.mechanismProperties as MechanismProperties;
+
+  const serviceName = mechanismProperties.SERVICE_NAME ?? 'mongodb';
 
   performGssapiCanonicalizeHostName(
     hostAddress.host,
-    mechanismProperties as MechanismProperties,
+    mechanismProperties,
     (err?: Error | MongoError, host?: string) => {
       if (err) return callback(err);
 
--- a/test/manual/kerberos.test.js
+++ b/test/manual/kerberos.test.js
@@ -3,12 +3,36 @@ const { MongoClient } = require('../../src');
 const chai = require('chai');
 const expect = chai.expect;
 
+function verifyKerberosAuthentication(client, done) {
+  client
+    .db('kerberos')
+    .collection('test')
+    .find()
+    .toArray(function (err, docs) {
+      let expectError;
+      try {
+        expect(err).to.not.exist;
+        expect(docs).to.have.length(1);
+        expect(docs[0].kerberos).to.be.true;
+      } catch (e) {
+        expectError = e;
+      }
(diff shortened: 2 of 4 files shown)


test/unit/core/connection_string.test.js
```diff
@@ -22,10 +22,7 @@ const skipTests = [
   'Unsupported option values are ignored',
 
   // We don't actually support `wtimeoutMS` which this test depends upon
-  'Deprecated (or unknown) options are ignored if replacement exists',
-
-  // We already handle this case in different ways
-  'may support deprecated gssapiServiceName option (GSSAPI)'
```
rose-m, line 28: I removed this as it was the last reference to `gssapiServiceName` and there is no test with that description anymore.

test/manual/kerberos.test.js
```diff
@@ -22,54 +41,46 @@ describe('Kerberos', function () {
...
 
-          client.close(done);
-        });
+    it('as part of the query string parameters', function (done) {
+      const client = new MongoClient(
+        `${krb5Uri}&authMechanismProperties=SERVICE_NAME:alternate&maxPoolSize=1`
+      );
+      client.connect(function (err) {
+        expect(err).to.exist;
+        expect(err.message).to.match(/Error from KDC: LOOKING_UP_SERVER/);
```
durran, line 81: This test is failing on Evergreen with a different error message than expected.

durran (changes requested):

test/manual/kerberos.test.js
rose-m: Yeah... dumb me did only adapt it for the other test 🙈 should be fixed now!

durran (approved):

src/cmap/auth/gssapi.ts
```diff
@@ -72,15 +73,14 @@ function makeKerberosClient(authContext: AuthContext, callback: Callback<Kerbero
...
 
-  const { username, password, mechanismProperties } = credentials;
-  const serviceName =
-    mechanismProperties['gssapiservicename'] ||
-    mechanismProperties['gssapiServiceName'] ||
-    'mongodb';
+  const { username, password } = credentials;
+  const mechanismProperties = credentials.mechanismProperties as MechanismProperties;
+
+  const serviceName = mechanismProperties.SERVICE_NAME || 'mongodb';
```
nbbeeken, line 79:
```suggestion
  const serviceName = mechanismProperties.SERVICE_NAME ?? 'mongodb';
```
super small style nit

nbbeeken (approved):
Thanks for that testing documentation! one small style request, otherwise lgtm so I'll leave a ✅

src/cmap/auth/gssapi.ts
rose-m: Done! :)

test/manual/kerberos.test.js
```diff
@@ -3,12 +3,31 @@ const { MongoClient } = require('../../src');
...
+function verifyKerberosAuthentication(client, done) {
+  client
+    .db('kerberos')
+    .collection('test')
+    .find()
+    .toArray(function (err, docs) {
+      expect(err).to.not.exist;
+      expect(docs).to.have.length(1);
+      expect(docs[0].kerberos).to.be.true;
+      client.close(done);
```
dariakp, line 15: can this be updated to ensure that client.close executes even if the assertions fail? the easiest way would be to do something like: `let expectError; try { expect ... } catch (err) { expectError = err } client.close(err => { done(expectError || err) })`

dariakp (changes requested):

test/manual/kerberos.test.js
rose-m: Just updated that so it closes it properly 👍

dariakp: Thank you very much!

dariakp (approved):

PR #2859, opened 2021-06-24
feat(NODE-3331): offer downleveled types for legacy typescript versions

nbbeeken:
Downlevel-dts is now run when ever defintions are generated and our package points typescript versions equal to 4.0.2 and older to these defintions. They target the lowest version downlevel-dts supports which is 3.4. I added a check to our CI to make sure that 4.0.2 will use and remain compatible with these types.

### DIFF
--- a/.evergreen/run-checks.sh
+++ b/.evergreen/run-checks.sh
@@ -1,9 +1,21 @@
 #!/bin/bash
-set -o errexit  # Exit the script with error if any of the commands fail
+set -o errexit # Exit the script with error if any of the commands fail
 
 export PROJECT_DIRECTORY="$(pwd)"
 NODE_ARTIFACTS_PATH="${PROJECT_DIRECTORY}/node-artifacts"
 export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 
+set -o xtrace
+
 npm run check:lint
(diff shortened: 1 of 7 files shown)


dariakp (approved):
LGTM

.evergreen/run-checks.sh
```diff
@@ -1,9 +1,21 @@
 #!/bin/bash
-set -o errexit  # Exit the script with error if any of the commands fail
+set -o errexit # Exit the script with error if any of the commands fail
 
 export PROJECT_DIRECTORY="$(pwd)"
 NODE_ARTIFACTS_PATH="${PROJECT_DIRECTORY}/node-artifacts"
 export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 
+set -o xtrace
```
emadum, line 9: Was this meant to stay here or is it left over from development? I find it generally clutters up logs having this on (and it can leak sensitive data).

emadum (changes requested):
Just confirming about xtrace being on, otherwise LGTM!

.evergreen/run-checks.sh
nbbeeken: Just testing, now that I have the useful "echo"s we dont need this

durran (approved):

emadum (approved):
LGTM

PR #2861, opened 2021-06-25
fix(NODE-3380): perform retryable write checks against server

emadum:
The driver was incorrectly checking for a `Single` topology instead of
a `Standalone` server when determining the retryability of write
operations.

NODE-3380

### DIFF
--- a/lib/core/sdam/topology.js
+++ b/lib/core/sdam/topology.js
@@ -668,12 +668,17 @@ class Topology extends EventEmitter {
         return;
       }
 
+      const notAlreadyRetrying = !options.retrying;
+      const retryWrites = !!options.retryWrites;
+      const hasSession = !!options.session;
+      const supportsRetryableWrites = server.supportsRetryableWrites;
+      const notInTransaction = !options.session.inTransaction();
       const willRetryWrite =
-        !options.retrying &&
-        !!options.retryWrites &&
-        options.session &&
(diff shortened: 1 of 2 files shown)


lib/core/sdam/topology.js
```diff
@@ -925,20 +924,20 @@ function executeWriteOperation(args, options, callback) {
...
-    !options.session.inTransaction() &&
-    options.explain === undefined;
-
   topology.selectServer(writableServerSelector(), options, (err, server) => {
     if (err) {
       callback(err, null);
       return;
     }
 
+    const willRetryWrite =
```
nbbeeken, line 933: On line 676 we have similiar logic to update and we discussed some code clarity while we're here:
```
    const currentlyNotRetrying = !args.retrying;
    const retryWrites = !!options.retryWrites;
    const hasSession = !!options.session;
    const supportsRetryableWrites = server.supportsRetryableWrites;
    const notInTransaction = !options.session.inTransaction();
    const notExplaining = options.explain === undefined;
    const willRetryWrite =
      currentlyNotRetrying &&
      retryWrites &&
      hasSession &&
      supportsRetryableWrites &&
      notInTransaction &&
      notExplaining;
```

nbbeeken (changes requested):

nbbeeken (approved):

durran (approved):

PR #2862, opened 2021-06-25
docs(NODE-3363): added docs for proposed error hierarchy

andymina:
## Description
- Added `docs/errors.md` which details the proposed new error class hierarchy. 
- Added examples of when each error class may be used.


docs/errors.md
```diff
@@ -0,0 +1,243 @@
...
+![(MongoLogicError hierarchy tree)](charts/imgs/MongoLogicError.svg)
+Children of `MongoLogicError` include:
+
+- #### `MongoInvalidArgumentError`
+
+  - Thrown when the user supplies malformed, unexpected arguments or failed to provide a required argument or field.
+
+- #### `MongoCompatibilityError`
+
+  - Thrown when a feature that is not enabled or allowed for the current configuration is used.
```
nbbeeken, line 72: Server or driver configuration?

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+- #### `MongoClientInstantiationError`
+
+  - Thrown when a user supplies a poorly-formatted connection string to the MongoClient constructor.
```
nbbeeken, line 76: If this is only for connection string then it should be in the name. We currently have MongoParseError for this which I do think effectively covers this case. But we might also be using parse errors for other cases, if that is the case then as long as we name this something more specific I'm not against it.

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+- #### `MongoMissingCredentialsError`
+
+  - Thrown when a user fails to provide authentication credentials before attempting to connect to the mongo server.
```
nbbeeken, line 80: What's the use case where this error would arise? you can run mongodb without authentication, but there isn't a way to tell that's the case until you attempt to connect.

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+- #### `MongoDependencyError`
```
nbbeeken, line 82: `MongoDependencyError` -> `MongoDependencyMissingError`

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+- #### `MongoTransactionError`
+
+  - Thrown when the user makes a mistake in the usage of transactions (e.g.: attempting to commit a transaction with a readPreference other than primary).
+
+- #### `MongoClientNotConnectedError`
+
+  - Thrown when the user attempts to operate on the data from a client that has not been connected to a MongoDB server instance.
+
+- #### `MongoKerberosClientConnectionError`
```
nbbeeken, line 113: `MongoKerberosClientConnectionError` -> `MongoKerberosConnectionError` / `MongoKerberosError`
I'm thinking that scoping this error for kerberos generally will serve us even if there's only the connection failure issue. But leaving connection in the name is also fine. I think we can drop Client since there's never a kerberos Server error on our end.

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+- #### `MongoIOError`
+
+  - Thrown when the driver fails to read or write from a file.
+
+- #### `MongoParseError`
+
+  - Thrown when the driver fails to correctly parse otherwise properly supplied input.
+
+- #### `MongoResourceClosedError`
```
nbbeeken, line 141: nit: blank line after this

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+  - Thrown when there is an attempt to access a resource which has already been or will be closed/destroyed.
+    - Children of this error class include:
+      - **`MongoServerClosedError`**: Thrown when an attempt is made to operate on a closed server.
+      - **`MongoStreamClosedError`**: Thrown when an attempt is made to operate on a closed stream.
+      - **`MongoTopologyClosedError`**: Thrown when an attempt is made to operate on a dropped, or otherwise unavailable, database.
+- #### `MongoCursorError`
+
+  - Thrown when the user incorrectly uses a cursor object.
+    - Children of this error class include:
+      - **`MongoInvalidCursorOperationError`**: Thrown when the user calls a function or method that is not supported on the current cursor.
```
nbbeeken, line 151: Would this arise because of the server version not supporting some functionality?

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+  - Thrown when the driver fails to correctly parse otherwise properly supplied input.
+
+- #### `MongoResourceClosedError`
+  - Thrown when there is an attempt to access a resource which has already been or will be closed/destroyed.
+    - Children of this error class include:
+      - **`MongoServerClosedError`**: Thrown when an attempt is made to operate on a closed server.
+      - **`MongoStreamClosedError`**: Thrown when an attempt is made to operate on a closed stream.
+      - **`MongoTopologyClosedError`**: Thrown when an attempt is made to operate on a dropped, or otherwise unavailable, database.
+- #### `MongoCursorError`
```
nbbeeken, line 147: nit: blank line above

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+  - Thrown when the user incorrectly uses a cursor object.
+    - Children of this error class include:
+      - **`MongoInvalidCursorOperationError`**: Thrown when the user calls a function or method that is not supported on the current cursor.
+      - **`MongoCursorExhaustedError`**: Thrown when an attempt is made to read from a cursor that has been exhausted.
+
+- #### `MongoStreamError`
```
nbbeeken, line 154: nit: blank line below

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+  - Thrown when a stream operation fails to execute.
+  - Children of this error class include:
+    - **`MongoChangeStreamError`**: Thrown when an error is encountered when operating on a ChangeStream.
+    - **`MongoGridFSStreamError`**: Thrown when an unexpected state is reached when operating on a GridFSStream.
+    - **`MongoGridFSChunkError`**: Thrown when a malformed or invalid chunk is encountered when reading from a GridFSStream.
+- #### `MongoBatchReExecutionError`
```
nbbeeken, line 160: nit: blank line above

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+  - Thrown when a user attempts to reexecute a batch command when one of the constituent commands has failed.
+
+- #### `MongoServerSelectionError`
+
+  - Thrown when the driver fails to select a server to complete an operation.
+
+- #### `MongoInvalidClassInstantiationError`
```
nbbeeken, line 168: nit: blank line below

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+These are errors which originate from faulty environment setup.
+
+# Test Plan
+
+The test plan consists of a series of prose tests.
+As numerous errors are being introduced, select classes will be tested.
+The classes to be tested will be selected based on two characteristics:
+
+1. The **frequency** of which users may encounter this error. Errors that users will likely catch in their code, including but not limited to `MongoIOError` and `MongoNetworkTimeoutError`, are part of the test plan. _Note:_ Error classes that should never be instantiated, such as `MongoLogicError` and `MongoRuntimeError`, will not be tested as the user should not encounter them.
```
nbbeeken, line 192: MongoNetworkTimeoutError is unlikely to be caught by users, maybe IO error depending on the example, which we have another comment thread about. ~What's the case y'all were thinking here?~ 

Edit: Ah I read the test plan now, I see how user's would encounter the timeout but I don't think it would be desirable to catch it. Anyway moot point but we can update the language a bit here.

docs/charts/mermaid/MongoError.mmd
```diff
@@ -0,0 +1,8 @@
+graph TD
```
nbbeeken, line 1: Can you make a small meta readme in this diretory of how to use these files? Nothing major, links to the right places, commands you run to generate the svgs etc.

test/functional/gridfs_stream.test.js
```diff
@@ -95,6 +95,38 @@ describe('GridFS Stream', function () {
     }
   });
 
+  it('destroy publishes provided error', {
```
nbbeeken, line 98: What's this test here doing?

nbbeeken (changes requested):

docs/errors.md
W-A-James: @nbbeeken The case we had in mind with this specifically is trying to sort a tailable cursor, so it may need some more specificity in its naming, but it have to do with the sever version.

test/functional/gridfs_stream.test.js
W-A-James: @nbbeeken I think this test was here already, but it got scooped into this PR after we rebased onto the most recent 4.0 branch. We may have some git-fu to figure out to get rid of that change in this PR.

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+![(MongoRuntimeError hierarchy tree part 2)](charts/imgs/MongoRuntimeError_1.svg)
+
+_MongoRuntimeError children (pt 3)_
+![(MongoRuntimeError hierarchy tree part 3)](charts/imgs/MongoRuntimeError_2.svg)
+
+Children of `MongoRuntimeError` include:
+
+- #### `MongoInternalDriverError`
+
+  - Thrown when a catastrophic failure occurs in the driver.
```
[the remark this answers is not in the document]
andymina: @W-A-James and I ended up removing this in an attempt to prevent another MongoError catastrophe. Our initial logic was that this error would catch anything that we couldn't think of now, but in hindsight new error classes can just be made on the fly.

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+  - Thrown when the driver fails to decompress data received from the server
+
+- #### `MongoExpiredSessionError`
+
+  - Thrown when the user attempts to operate on a session that has expired or has been closed.
+
+- #### `MongoIOError`
+
+  - Thrown when the driver fails to read or write from a file.
```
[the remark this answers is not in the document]
andymina: Removed this to prevent capturing Node.js errors.

docs/errors.md
andymina: We think MongoParseError should effectively cover this, like you said, as long as there is a detailed error message to go along with it.

docs/errors.md
andymina: Looking through some of the files, we thought this error would be used when Mongo was paired with different authentication methods like x509 or SCRAM. Almost all of the auth mechanisms under `src/cmap/auth` seem to have a generic must provide credentials error, so this class would unify that.

docs/errors.md
```diff
@@ -0,0 +1,249 @@
...
+
+  - Thrown when the user attempts to operate on the data from a client that has not been connected to a MongoDB server instance.
+
+- #### `MongoKerberosClientConnectionError`
+
+  - Thrown when the user attempts to authenticate via Kerberos, but fails to connect to the Kerberos client.
+
+- #### `MongoEncryptionError`
+
+  - Thrown when an autoencrypter is requested but unavailable.
```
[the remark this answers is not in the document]
andymina: We're deferring to you because we don't know about enough about the encrypter.

docs/errors.md
emadum: It looks like most errors in the Cursor are thrown during initialization (in methods with `assertUninitialized(this);` at the top). I'd propose we just give those a general name like `MongoCursorInitializationError`.

docs/errors.md
```diff
@@ -0,0 +1,231 @@
...
+- [`MongoServerError`](#MongoServerError)
+- [`MongoSystemError`](#MongoSystemError)
+
+## `MongoDriverError`
+
+This class represents errors which originate in the driver itself or in the user's use of the driver.
+This class should **never** be directly instantiated.
+Its children are the main classes of errors that most users will interact with: [**`MongoLogicError`**](#MongoLogicError) and [**`MongoRuntimeError`**](#MongoRuntimeError).
+
+### `MongoLogicError`
```
emadum, line 57: I find that I'm not a huge fan of this name because "logic" is quite vague, would something like `MongoAPIError` be better?

docs/errors.md
```diff
@@ -0,0 +1,231 @@
...
+
+- #### `MongoDecompressionError`
+
+  - Thrown when the driver fails to decompress data received from the server
+
+- #### `MongoExpiredSessionError`
+
+  - Thrown when the user attempts to operate on a session that has expired or has been closed.
+
+- #### `MongoParseError`
```
emadum, line 123: I think this might also be making more explicit, e.g.`MongoURIParseError`, thoughts?

docs/errors.md
```diff
@@ -0,0 +1,231 @@
...
+These are errors which originate from faulty environment setup.
+
+# Test Plan
+
+The test plan consists of a series of prose tests.
+As numerous errors are being introduced, select classes will be tested.
+The classes to be tested will be selected based on two characteristics:
+
+1. The **frequency** with which users may encounter this error. Errors that users will likely catch in their code, including but not limited to `MongoIOError` and `MongoNetworkTimeoutError`, are part of the test plan. _Note:_ Error classes that should never be instantiated, such as `MongoLogicError` and `MongoRuntimeError`, will not be tested as the user should not encounter them.
+2. The **scope** of the error. Errors that tackle a large subset of issues, including but not limited to `MongoServerError` and `MongoSystemError`, will _not_ be part of the test plan.
```
emadum, line 185: I'd propose adding "coverage" to these criteria - many errors will already be covered by existing tests; I think we should prioritize writing new tests for errors which don't already have test coverage.

emadum (changes requested):
Had a few points of discussion but this is looking great, nice job! 👍 

I noticed some of the errors in the graphs, like `MongoNoCursorInChangeStreamError`, aren't also in the markdown text. Would you be able to add any errors that are only in the graph to the markdown, so that we can reference them more easily in the review?

docs/errors.md
nbbeeken: Oh I see yes that seems like a good classification, although out of scope from this work I wonder if it ever gets used, I believe our validation is strict enough such that you would be warned at client construction time that something is not setup right. That being said lets go ahead and keep this

docs/errors.md
nbbeeken: Just curious whats the motivation for `MongoMissingDependencyError` vs `MongoDependencyMissingError`? Maybe we should actually go with `MongoModuleNotFoundError` to kinda match with nodejs.

docs/errors.md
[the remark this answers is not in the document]
nbbeeken: I don't know about "catastrophe" 😛 but yea that sounds like the right approach! We'll have a number of PRs to tackle on this.

docs/errors.md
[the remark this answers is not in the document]
nbbeeken: Was there an existing error that inspired this? I'm going based of the description, autoEncrypter being unavailable is usually cus the library that supports it is missing.

docs/errors.md
```diff
@@ -0,0 +1,231 @@
...
+
+These are errors which originate from faulty environment setup.
+
+# Test Plan
+
+The test plan consists of a series of prose tests.
+As numerous errors are being introduced, select classes will be tested.
+The classes to be tested will be selected based on two characteristics:
+
+1. The **frequency** with which users may encounter this error. Errors that users will likely catch in their code, including but not limited to `MongoIOError` and `MongoNetworkTimeoutError`, are part of the test plan. _Note:_ Error classes that should never be instantiated, such as `MongoLogicError` and `MongoRuntimeError`, will not be tested as the user should not encounter them.
```
nbbeeken, line 184: MongoIOError has been removed, you can hold off making the change right now if you want lets use this comment as an reminder to comb over the docs for dropped errors and to update the mermaid diagrams

docs/errors.md
andymina: Ooh, good point. Initially we just preferred the phrasing of the first, but I'm a little hesitant to put `MongoModuleNotFoundError` because may be interpreted as *MongoModule Not Found* instead of *Mongo: ModuleNotFound*.

nbbeeken (changes requested):

docs/errors.md
nbbeeken: Fair point! lets keep dependency in the name then, what we have up here is fine! we could go on for ages between the first two options.

docs/errors.md
andymina: Leaving a note here: this is for the server configuration. (e.g. maxWireVersion check)

docs/errors.md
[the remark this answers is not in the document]
andymina: Note: @nbbeeken and I talked about this and came to the realization that this error should never actually occur because its existence is asserted upon `CryptoConnection` initialization. No point in making extra work

docs/errors.md
andymina: @emadum: `assertUnintialized(this)` appears pretty often so I've made `MongoCursorInUseError` to cover the case where a user attempts to add options onto an already existing cursor that has failed `assertUnintialized(this)`. 

I've also renamed `MongoInvalidCursorOperationError` to `MongoTailableCursorError` because that was the only place where this error was being used. It will still be thrown when a user attempt to call a function that a TailableCursor doesn't support.

docs/errors.md
andymina: @emadum I’ve added existing coverage as a defining characteristic in the test plan, but I haven’t actually updated the test plan yet. I still have to skim through the existing tests and see what’s covered and what’s not.

docs/errors.md
andymina: @emadum I think `MongoAPIError` definitely does a better job of immediately directing the user where to look so I’ve updated `MongoLogicError`.

docs/errors.md
andymina: I've renamed `MongoParseError` to `MongoURIError` to be more specific in what it's targeting while still being able to expand this to cases where the user supplies a properly formatted URI, but it contains illegal options.

docs/errors.md
andymina: Everything has been updated 😎

nbbeeken (approved):

emadum (approved):

test/functional/gridfs_stream.test.js
dariakp: Can we leave this comment open until the mysterious test is taken care of?

docs/errors.md
```diff
@@ -0,0 +1,225 @@
...
+
+  - Thrown when a user attempts to reexecute a batch command when one of the constituent commands has failed.
+
+- #### `MongoServerSelectionError`
+
+  - Thrown when the driver fails to select a server to complete an operation.
+
+## `MongoNetworkError`
+
+These are errors encountered at runtime which occur when the driver encounters an issue in the network which leads to an inability to connect to a mongo server instance. Children of this class include:
```
dariakp, line 157: We should find a way to phrase it as less of a run on sentence

docs/errors.md
```diff
@@ -0,0 +1,225 @@
...
+
+- #### `MongoNetworkTimeoutError`
+
+  - Thrown when a timeout expires in attempting to connect to the mongo server
```
dariakp, line 161:
```suggestion
  - Thrown when a timeout expires in attempting to connect to the mongo server.
```

dariakp (approved):
Looks good, thanks for all the work here!

PR #2865, opened 2021-06-28
fix(NODE-2026): support SERVICE_REALM authentication mechanism property

rose-m:
Introduces support for the `SERVICE_REALM` authentication mechanism property. It will be work as expected in Windows environments due to the underlying nature of the GSSAPI implementation on *nix.

The logic to compute the Service Principal Name (SPN) is as follows:
1. Combine service name and hostname (with a `/` in between for Windows and an `@` otherwise)
2. If present, attach the `SERVICE_REALM` with an `@` to the result of step 1.

The logic is taken from the C driver where you can find the steps mangled as follows:
1. "Join" the array of `[serviceName, hostName, serviceRealm]` with `@` (see https://github.com/mongodb/mongo-c-driver/blob/f1093ba1686b8bf876eed67e0c7f59918649ef15/src/libmongoc/src/mongoc/mongoc-cluster-sspi.c#L61-L71)
2. If there is no `/` in the SPN (which is not the case after step 1), replace the first `@` with a `/` (see https://github.com/mongodb/mongo-c-driver/blob/5b91fb983fc5badf38e2debeb4ce6c231b73605b/src/libmongoc/src/mongoc/mongoc-sspi.c#L176-L182)

I was able to successfully test this behavior in a Windows environment with Kerberos enabled.

### DIFF
--- a/src/cmap/auth/gssapi.ts
+++ b/src/cmap/auth/gssapi.ts
@@ -7,6 +7,7 @@ import type { Document } from '../../bson';
 type MechanismProperties = {
   gssapiCanonicalizeHostName?: boolean;
   SERVICE_NAME?: string;
+  SERVICE_REALM?: string;
 };
 
 import * as dns from 'dns';
@@ -90,14 +91,15 @@ function makeKerberosClient(authContext: AuthContext, callback: Callback<Kerbero
         Object.assign(initOptions, { user: username, password: password });
       }
 
-      initializeClient(
-        `${serviceName}${process.platform === 'win32' ? '/' : '@'}${host}`,
-        initOptions,
-        (err: string, client: KerberosClient): void => {
-          if (err) return callback(new MongoDriverError(err));
-          callback(undefined, client);
-        }
-      );
+      let spn = `${serviceName}${process.platform === 'win32' ? '/' : '@'}${host}`;
+      if ('SERVICE_REALM' in mechanismProperties) {
+        spn = `${spn}@${mechanismProperties.SERVICE_REALM}`;
+      }
+
+      initializeClient(spn, initOptions, (err: string, client: KerberosClient): void => {
+        if (err) return callback(new MongoDriverError(err));
+        callback(undefined, client);
+      });
     }
   );
 }
--- a/test/manual/kerberos.test.js
+++ b/test/manual/kerberos.test.js
@@ -50,8 +50,8 @@ describe('Kerberos', function () {
     });
   });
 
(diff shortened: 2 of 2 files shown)


src/cmap/auth/gssapi.ts
```diff
@@ -90,14 +91,20 @@ function makeKerberosClient(authContext: AuthContext, callback: Callback<Kerbero
...
+      let spn: string;
+      if (process.platform === 'win32') {
+        spn = `${serviceName}/${host}`;
+        if ('SERVICE_REALM' in mechanismProperties) {
+          spn = `${spn}@${mechanismProperties.SERVICE_REALM}`;
         }
-      );
+      } else {
+        spn = `${serviceName}@${host}`;
+      }
```
nbbeeken, line 102: We want to keep the logic that uses slash instead of @ on windows, but move attaching the service realm to outside so it gets attached regardless of platform.

nbbeeken (changes requested):
Just for another example [here's the logic in python](https://github.com/mongodb/mongo-python-driver/blob/4152600ae6f7d9044d67a2601b0e7bdf3f301a30/pymongo/auth.py#L316-L318) I believe the only issue we have here is the limiting of SERVICE_REALM only on windows.

src/cmap/auth/gssapi.ts
nbbeeken: The test: `validate that SERVICE_REALM and CANONICALIZE_HOST_NAME can be passed in`
in: `test/manual/kerberos.test.js:54`
is now going to fail after making this change, you can safely skip it, we have a ticket (NODE-3060) to track unskipping / removing later.

rose-m: I discussed that with @mcasimir, too... So we want to just attach that regardless of the platform _knowing_ it will not work on any *nix systems? I've no hard feelings on that - I'll add it :)

rose-m:
Note: I was also able to successfully test the code in a Windows Kerberos environment (manually).

durran (approved):

emadum (approved):
LGTM

nbbeeken (approved):

src/cmap/auth/gssapi.ts
nbbeeken: Yea I'm just looking at python and mirroring their logic, we will revisit this when we have the right testing environment. You were able to confirm it works on windows so I'm happy with what we have here, thanks for this!

PR #2868, opened 2021-06-30
fix(NODE-1843): bulk operations ignoring provided sessions

nbbeeken:
The BulkOperationBase class wasn't making use of the options provided at construction time during execution time.

NODE-1843

### DIFF
--- a/test/functional/bulk.test.js
+++ b/test/functional/bulk.test.js
@@ -2003,4 +2003,82 @@ describe('Bulk', function () {
         );
     });
   });
+
+  describe('Bulk operation transaction rollback', () => {
+    /** @type {import('../../src/index').MongoClient} */
+    let client;
+    /** @type {import('../../src/index').Collection<{ answer: number }>} */
+    let collection;
+
+    beforeEach(async function () {
+      const config = this.configuration;
+      client = config.newClient();
+      await client.connect();
+
+      collection = client
+        .db('bulk_operation_writes_test')
+        .collection('bulk_write_transaction_test');
+
+      await collection.deleteMany({});
+    });
+
+    it('should abort ordered/unordered bulk operation writes', async () => {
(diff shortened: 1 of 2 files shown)


emadum (approved):
LGTM, though I'm sure I've seen `client.close` working in after hooks before without triggering the leak checker...

test/functional/bulk.test.js
```diff
@@ -2003,4 +2003,103 @@ describe('Bulk', function () {
...
+      }
+
+      collection = await client
+        .db('bulk_operation_writes_test')
+        .createCollection('bulk_write_transaction_test');
+
+      await collection.deleteMany({});
+    });
+
+    it('should abort ordered/unordered bulk operation writes', {
```
dariakp, line 2034: let's split up this test and the following one into 3 cases each: ordered, unordered, and mixed; and clearly identify withTransaction vs not in the name

test/functional/bulk.test.js
```diff
@@ -2003,4 +2003,103 @@ describe('Bulk', function () {
...
+        await session.endSession();
+
+        const documents = await collection.find().toArray();
+
+        expect(documents).to.be.lengthOf(
+          0,
+          'bulk operation writes were made outside of transaction'
+        );
+
+        await client.close(); // session leak checker doesn't let you put the close in an after hook
```
dariakp, line 2065: I'm pretty sure I've seen client close used in an afterEach, and it really does belong there; otherwise the cleanup is not guaranteed - in fact it would fail if that expect fails

dariakp (changes requested):

test/functional/bulk.test.js
nbbeeken: Yea I'm not sure what error I was seeing that made me think this, its gone now

dariakp (approved):

PR #2869, opened 2021-06-30
fix(NODE-3290): versioned api validation and tests

emadum:
Backport of the versioned API updates made in #2809, also pulls in updates to `test/tools/cluster_setup.sh` from 4.0.

NODE-3290


lib/core/index.js
```diff
@@ -15,6 +15,10 @@ try {
 } catch (err) {} // eslint-disable-line
 
 module.exports = {
+  // Versioned API
+  ServerApiVersion: Object.freeze({
```
nbbeeken, line 19: Do we want this exported from the top level of the driver too?
If so, can we pull it our in to its own variable and attach a small jsdoc comment to it?

lib/mongo_client.js
```diff
@@ -191,16 +192,38 @@ const validOptions = require('./operations/connect').validOptions;
  * @param {MongoClientOptions} [options] Optional settings
  */
 function MongoClient(url, options) {
+  options = options || {};
   if (!(this instanceof MongoClient)) return new MongoClient(url, options);
   // Set up event emitter
   EventEmitter.call(this);
 
   if (options && options.autoEncryption) require('./encrypter'); // Does CSFLE lib check
```
nbbeeken, line 200:
```suggestion
  if (options.autoEncryption) require('./encrypter'); // Does CSFLE lib check
```
Now that we're defaulting options, just a little cleanup.

nbbeeken (changes requested):
Some small Qs
Thanks for sync-ing the setup script!

lib/core/index.js
emadum: > Do we want this exported from the top level of the driver too?

Yes, I missed it in the initial back-port.

>  If so, can we pull it our in to its own variable and attach a small jsdoc comment to it?

Good call, will do.

nbbeeken:
@emadum This looks ready to go in the sense that it has the validation and tests but the diff looks like its bringing more in than intended, I think 3.7 needs a fastforward before merging?

nbbeeken (approved):
LGTM

dariakp (approved):
LGTM :)

PR #2870, opened 2021-06-30
NODE-3316: Update Maintainer Email Address

agolin95:
Added in the author name and email fields (omitted url).


nbbeeken (changes requested):
Update bugs link and target 3.6 (might cause some conflict so I'll leave it to you)

PR #2876, opened 2021-07-01
fix(NODE-3397): report more helpful error with unsupported authMechanism in initial handshake

vkarpov15:
## Description

I haven't been able to repro this locally, but there's a user that ran into this case at Automattic/mongoose#10377. Figured this change is sufficiently straightforward that it's worthwhile to patch without digging into it much further.

**What changed?**

Added a quick check for `null` or `undefined` before calling a method on an object

### DIFF
--- a/lib/core/connection/connect.js
+++ b/lib/core/connection/connect.js
@@ -180,7 +180,7 @@ function prepareHandshakeDocument(authContext, callback) {
 
     const authProvider = AUTH_PROVIDERS[credentials.mechanism];
     if (authProvider == null) {
-      return callback(new MongoError(`authMechanism '${credentials.mechanism}' not supported`)); 
+      return callback(new MongoError(`No AuthProvider for ${credentials.mechanism} defined.`));
     }
     authProvider.prepare(handshakeDoc, authContext, callback);
     return;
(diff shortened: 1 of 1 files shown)


lib/core/connection/connect.js
```diff
@@ -179,6 +179,9 @@ function prepareHandshakeDocument(authContext, callback) {
     }
 
     const authProvider = AUTH_PROVIDERS[credentials.mechanism];
+    if (authProvider == null) {
+      return callback(new MongoError(`authMechanism '${credentials.mechanism}' not supported`));
```
nbbeeken, line 183:
```suggestion
      return callback(new MongoError(`No AuthProvider for ${credentials.mechanism} defined.`)); 
```

nbbeeken (changes requested):
Thanks! yea, I agree catching and improving the error message would be good here

Slight message refactor suggestion so that its the same as the error message we have in 4.x

made NODE-3397 to track

lib/core/connection/connect.js
vkarpov15: Works for me :+1:

nbbeeken (approved):
LGTM

emadum (approved):

PR #2878, opened 2021-07-02
feat(NODE-3095): add timeseries options to db.createCollection

emadum:


src/operations/create_collection.ts
```diff
@@ -60,6 +60,10 @@ export interface CreateCollectionOptions extends CommandOperationOptions {
   pipeline?: Document[];
   /** A primary key factory function for generation of custom _id keys. */
   pkFactory?: PkFactory;
+  /** A document specifying configuration options for timeseries collections. */
+  timeseries?: Document;
```
nbbeeken, line 64: I think it would nice to provide some hinting here, we can have:
```typescript
interface TimeSeriesOptions extends Document {
  timeField: string;
  metaField: string;
  granularity: string;
}
```

where `extends Document` will permit future options without drivers changes

nbbeeken (changes requested):
TS suggestion, what do you think?

src/operations/create_collection.ts
emadum: Did you see those specified somewhere? I'm wondering if some of them should be optional.

nbbeeken (approved):
LGTM

test/functional/collection_management_spec.test.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const { expect } = require('chai');
+const { loadSpecTests } = require('../spec/index');
+const { runUnifiedTest } = require('./unified-spec-runner/runner');
+
+describe('Collection management unified spec tests', function () {
+  for (const collectionManagementTest of loadSpecTests('collection-management')) {
+    expect(collectionManagementTest).to.exist;
```
dariakp, line 9: just a suggestion: I generally recommend against using assertions that aren't function calls (like `exist`) because they're easy to accidentally disable with a typo (e.g., `.exists` will resolve to `undefined` so the "assertion" will continue to pass even though it's no longer testing anything); it's not a major problem since the IDE (and possibly the linter) will yell at you if you do try to access an undefined assertion, but something like `.to.be.an('object')` would technically be safer

dariakp (commented):
just a minor suggestion

test/functional/collection_management_spec.test.js
emadum: Running a test with a typo like that would result in a failure with an error like `Error: Invalid Chai property: exits. Did you mean "exists"?`.

dariakp: Oh that's awesome, they must have updated it in one of the recent-ish versions, because my favorite example used to be `expect(1).to.be.a.potato`. I fully withdraw my objection then :)

dariakp (approved):

test/functional/collection_management_spec.test.js
nbbeeken: TIL 🚀

PR #2879, opened 2021-07-02
refactor!(NODE-3368): make name prop on error classes read-only

andymina:
## Description
- Added getter function to error classes to ensure that it is a read-only property
- Added unit tests for `.name` property to ensure it is a read-only and matches the class name


src/error.ts
```diff
@@ -88,8 +88,11 @@ export class MongoError extends Error {
     } else {
       super(message);
     }
+  }
 
-    this.name = 'MongoError';
+  // Returns name of the error
```
nbbeeken, line 93: We can remove this comment, its pretty self explanatory

test/unit/errors.test.js
```diff
@@ -2,10 +2,61 @@
...
+    describe(errorName, () => {
+      it(`name should be read-only`, () => {
+        // Dynamically create error class with message
+        let error = new errorClasses[errorName]('generated by test');
+
+        // expect name property to be class name
+        expect(error.name).to.eql(errorName);
+        expect(() => {
+          error.name = 'renamed by test';
+        }).to.throw(TypeError);
```
nbbeeken, line 44: I have a suggestion that results in essentially equivalent assertion but checks precisely what we desire from all of our errors:

```javascript
const errorPrototype = Object.getPrototypeOf(error)
expect(errorPrototype).to.have.ownPropertyDescriptor('name').that.has.property('get')
```

test/unit/errors.test.js
```diff
@@ -2,10 +2,61 @@
...
+  };
+
+  for (const errorName in errorClasses) {
+    describe(errorName, () => {
+      it(`name should be read-only`, () => {
+        // Dynamically create error class with message
+        let error = new errorClasses[errorName]('generated by test');
+
+        // expect name property to be class name
+        expect(error.name).to.eql(errorName);
```
nbbeeken, line 41:
```suggestion
        expect(error).to.have.property('name', errorName);
```

nbbeeken (changes requested):
Nice work! 🚀  just some suggestions we can discuss

test/unit/errors.test.js
```diff
@@ -2,10 +2,61 @@
...
+  describe('MongoNetworkError', () => {
+    it(`name should be read-only`, () => {
+      let error = new MongoNetworkError('generated by test');
+
+      // expect name property to be class name
+      expect(error.name).to.eql('MongoNetworkError');
+      expect(() => {
+        error.name = 'renamed by test';
+      }).to.throw(TypeError);
+    });
```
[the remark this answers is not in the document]
andymina: It was originally, but Mocha threw a warning about having duplicate describe blocks: the one generated by the `errorClasses` and this one which also test handshakes.

test/unit/errors.test.js
```diff
@@ -2,10 +2,42 @@
...
+
+  for (const errorName in errorClasses) {
+    describe(errorName, () => {
+      it(`name should be read-only`, () => {
+        // Dynamically create error class with message
+        let error = new errorClasses[errorName]('generated by test');
+        const errorPrototype = Object.getPrototypeOf(error);
+
+        // expect name property to be class name
+        expect(error.name).to.have.property('name', errorName);
```
nbbeeken, line 24:
```suggestion
        expect(error).to.have.property('name', errorName);
```

andymina: Oops, typo!

test/unit/errors.test.js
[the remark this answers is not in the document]
nbbeeken: Ah yea, then we can change the name of the describe block below and then we don't have to duplicate, so maybe change the name to `when MongoNetworkError is constructed`

test/unit/errors.test.js
```diff
@@ -2,10 +2,61 @@
 
 const expect = require('chai').expect;
 const { getSymbolFrom } = require('../tools/utils');
-const MongoNetworkError = require('../../src/error').MongoNetworkError;
+const {
```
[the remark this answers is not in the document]
nbbeeken: `MongoBulkWriteError` is exported from index so we can drop the explicit import we have.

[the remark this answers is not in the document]
andymina: Removed the dup import.

test/unit/errors.test.js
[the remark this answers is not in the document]
andymina: Ahhh, that's wise. I've committed that change.

nbbeeken (approved):
LGTM

src/bulk/common.ts
```diff
@@ -684,10 +684,13 @@ export class MongoBulkWriteError extends MongoServerError {
     super(error as Error);
     Object.assign(this, error);
 
-    this.name = 'MongoBulkWriteError';
     this.result = result;
   }
 
+  get name(): string {
```
dariakp, line 690: I would advise against using getters just for the sake of making something readonly - adding a function to a class does make it heavier (less performant) than just having a plain property; typically getters are used when the corresponding attribute needs to be computed dynamically; if it is important to make this particular property readonly, `defineProperty` allows designating it as such

src/cmap/errors.ts
```diff
@@ -12,9 +12,12 @@ export class PoolClosedError extends MongoDriverError {
 
   constructor(pool: ConnectionPool) {
     super('Attempted to check out a connection from closed connection pool');
-    this.name = 'MongoPoolClosedError';
     this.address = pool.address;
   }
+
+  get name(): string {
+    return 'PoolClosedError';
```
dariakp, line 19: are the two renamings in this file intentional? these could potentially be breaking changes

test/unit/errors.test.js
```diff
@@ -2,10 +2,31 @@
...
+describe('MongoErrors', () => {
+  // import errors as object
+  let errorClasses = Object.fromEntries(
+    Object.entries(require('../../src/index')).filter(([key]) => key.endsWith('Error'))
+  );
+  errorClasses = { ...errorClasses, PoolClosedError, WaitQueueTimeoutError };
+
+  for (const errorName in errorClasses) {
+    describe(errorName, () => {
+      it(`name should be read-only`, () => {
```
dariakp, line 17: technically, the code in this test does not check whether the error is readonly, but the particular implementation of the property as a getter; I usually advise to try to test the intention of the code, and not its implementation; in this case, it would mean taking the property, attempting to set it, and then expecting the property to be the same (something like `const obj = new TestClass()`, `expect(obj).to.have.property('name', 'Test Name')`, `obj.name = 'New Name'`, `expect(obj).to.have.property('name', 'Test Name')`

dariakp (changes requested):

test/unit/errors.test.js
dariakp: Note: if you end up with a strict requirement for read only, the setting of the prop in the test should be a try/catch with an empty catch block (since that behavior is an implementation detail)

src/bulk/common.ts
addaleax: > adding a function to a class does make it heavier (less performant) than just having a plain property

You can generally expect that this kind of getter is overall faster than a plain property, because it lives on the prototype and does not require the creation of an additional property on each object instance.

src/cmap/errors.ts
andymina: This was unintentional, but this did highlight another issue. The test should assert that the `name` prop matches the actual name of the class, but this isn't true for `PoolClosedError` and `WaitQueueTimeoutError` as both of their name props add the "Mongo" prefix. For now, I've imported both of those errors and aliased them with their "Mongo" prefix, but it's only a temporary fix.

test/unit/errors.test.js
andymina: Just updated the test to include testing the intention, thank you!

src/bulk/common.ts
dariakp: That is an interesting point; my original comment was coming from the perspective of accessing the property itself via getter vs direct read, not the creation of the error object itself; having properties defined on the prototype instead of the instance would indeed make for a smaller memory footprint for the individual instances (not that we likely have anything to worry about in practice since it's unlikely anyone is compiling or iterating over large arrays of our error objects)

dariakp: we could consider setting a static property on the class that has the error name and then have `this.name = <static prop>`, which might get us the best of both worlds; @addaleax what do you think?

addaleax: Practically speaking, the cost of accessing a trivial getter is zero because engines can inline these functions with the same efficiency as an access to a plain data property.

If you want to emulate what e.g. the built-in JS Error types do, you can define a readonly data property on the prototype; but practically speaking, I think using a getter for `Error` subclasses is a common pattern and it doesn’t really hurt to do it.

test/unit/errors.test.js
```diff
@@ -2,10 +2,36 @@
...
+        // Dynamically create error class with message
+        let error = new errorClasses[errorName]('generated by test');
+        // expect name property to be class name
+        expect(error).to.have.property('name', errorName);
+
+        try {
+          error.name = 'renamed by test';
+          expect(error).to.have.property('name', 'generated by test');
```
dariakp, line 28: this expect should be outside the try/catch block since you don't want to swallow the error thrown by the expect in the case of failure, otherwise looks good :)

src/bulk/common.ts
dariakp: I have the opposing stylistic preference here but I'm happy to defer to whichever option the rest of the team feels most comfortable with. Thank you for the perspective :)

test/unit/errors.test.js
andymina: Oops, sorry! Just fixed that and replaced `'generated by test'` with `errorName` because it was also a typo.

dariakp (approved):

PR #2881, opened 2021-07-06
feat(NODE-3410): added MongoRuntimeError

andymina:
## Description
- Added `MongoRuntimeError` as per the proposed error class hierarchy

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -172,8 +172,10 @@ export class MongoDriverError extends MongoError {
 
 /**
  * An error generated when the driver encounters unexpected unexpected input
- * or reaches an unexpected/invalid internal state. Should never be
- * directly instantiated.
+ * or reaches an unexpected/invalid internal state. 
+ * 
+ * @privateRemarks
+ * Should **never** be directly instantiated.
  *
  * @public
  * @category Error
--- a/src/error.ts
+++ b/src/error.ts
@@ -171,7 +171,7 @@ export class MongoDriverError extends MongoError {
 }
 
 /**
- * An error generated when the driver encounters unexpected unexpected input
+ * An error generated when the driver encounters unexpected input
  * or reaches an unexpected/invalid internal state
  *
  * @privateRemarks
--- a/src/error.ts
+++ b/src/error.ts
@@ -170,6 +170,25 @@ export class MongoDriverError extends MongoError {
   }
 }
 
+/**
+ * An error generated when the driver encounters unexpected unexpected input
+ * or reaches an unexpected/invalid internal state. Should never be
+ * directly instantiated.
+ *
+ * @public
+ * @category Error
+ */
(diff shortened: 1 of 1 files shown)


src/error.ts
```diff
@@ -170,6 +170,25 @@ export class MongoDriverError extends MongoError {
   }
 }
 
+/**
+ * An error generated when the driver encounters unexpected unexpected input
+ * or reaches an unexpected/invalid internal state. Should never be
+ * directly instantiated.
```
nbbeeken, line 176:
```suggestion
 * An error generated when the driver encounters unexpected unexpected input
 * or reaches an unexpected/invalid internal state. 
 * 
 * @privateRemarks
 * Should **never** be directly instantiated.
```

nbbeeken (changes requested):
one small request, just no need to mention usage in public docs so you can put it under privateRemarks

nbbeeken (approved):
LGTM

src/error.ts
```diff
@@ -170,6 +170,27 @@ export class MongoDriverError extends MongoError {
...
+ * or reaches an unexpected/invalid internal state
+ *
+ * @privateRemarks
+ * Should **never** be directly instantiated.
+ *
+ * @public
+ * @category Error
+ */
+export class MongoRuntimeError extends MongoDriverError {
+  code?: string;
```
dariakp, line 184: MongoDriverError already has the code declared, so I don't think it's necessary to redeclare it on the subclass unless we're intending to remove it from the parent class; I also want to note that the code for MongoDriverError is used for exactly one error and I think there is a ticket that is meant to find a proper way to deal with that case (the ENOENT case); for now, I'd say just leave the explicit declaration off this class

dariakp (changes requested):

src/error.ts
dariakp: (You could also consider making the actual codes a string enum for the more specific, directly instantiated, classes down the line)

W-A-James (approved):
LGTM

src/error.ts
andymina: Just removed the code from this class because it was redundant like you said.

src/error.ts
```diff
@@ -170,6 +170,26 @@ export class MongoDriverError extends MongoError {
   }
 }
 
+/**
+ * An error generated when the driver encounters unexpected unexpected input
```
emadum, line 174:
```suggestion
 * An error generated when the driver encounters unexpected input
```

emadum (changes requested):
Caught a typo, LGTM other than that 👍

andymina:
Changing this PR to be merged into NODE-3399 where Warren and I will implement the new error hierarchy

dariakp (approved):

emadum (approved):

PR #2882, opened 2021-07-06
fix(NODE-3416): make change stream generic default to Document

vkarpov15:
## Description

In most places, the node driver defaults to using `Document` if no generic param is provided. `ChangeStream` looks to be an exception.

**What changed?**

Added a default generic param to ChangeStream

Nope


nbbeeken (approved):
LGTM

dariakp (approved):

PR #2883, opened 2021-07-06
refactor(NODE-3402): Implement MongoAPIError and its children

W-A-James:
## Description

Implement the errors detailed in docs/errors.md from NODE-3363 that fall under MongoAPIError

**What changed?**
Replaced some MongoDriverError instances with the following errors where appropriate

- MongoInvalidArgumentError
- MongoMissingDependencyError
- MongoMissingCredentialsError
- MongoCompatibilityError

### DIFF
--- a/src/cmap/auth/gssapi.ts
+++ b/src/cmap/auth/gssapi.ts
@@ -21,7 +21,9 @@ export class GSSAPI extends AuthProvider {
   auth(authContext: AuthContext, callback: Callback): void {
     const { connection, credentials } = authContext;
     if (credentials == null)
-      return callback(new MongoMissingCredentialsError('credentials required'));
+      return callback(
+        new MongoMissingCredentialsError('credentials required for gssapi authentication')
+      );
     const { username } = credentials;
     function externalCommand(
       command: Document,
--- a/src/cmap/auth/gssapi.ts
+++ b/src/cmap/auth/gssapi.ts
@@ -1,5 +1,10 @@
 import { AuthProvider, AuthContext } from './auth_provider';
-import { MongoDriverError, MongoError } from '../../error';
+import {
+  MongoDriverError,
+  MongoMissingCredentialsError,
+  MongoError,
+  MongoMissingDependencyError
+} from '../../error';
 import { Kerberos, KerberosClient } from '../../deps';
 import { Callback, ns } from '../../utils';
 import type { Document } from '../../bson';
@@ -15,7 +20,8 @@ import * as dns from 'dns';
 export class GSSAPI extends AuthProvider {
   auth(authContext: AuthContext, callback: Callback): void {
     const { connection, credentials } = authContext;
-    if (credentials == null) return callback(new MongoDriverError('credentials required'));
+    if (credentials == null)
+      return callback(new MongoMissingCredentialsError('credentials required'));
     const { username } = credentials;
     function externalCommand(
       command: Document,
@@ -25,7 +31,7 @@ export class GSSAPI extends AuthProvider {
     }
     makeKerberosClient(authContext, (err, client) => {
       if (err) return callback(err);
-      if (client == null) return callback(new MongoDriverError('gssapi client missing'));
+      if (client == null) return callback(new MongoMissingDependencyError('gssapi client missing'));
       client.step('', (err, payload) => {
         if (err) return callback(err);
 
--- a/src/error.ts
+++ b/src/error.ts
@@ -245,6 +245,44 @@ export class MongoParseError extends MongoDriverError {
   }
 }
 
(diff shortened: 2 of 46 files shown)


src/cmap/auth/gssapi.ts
```diff
@@ -15,7 +20,8 @@ import * as dns from 'dns';
 export class GSSAPI extends AuthProvider {
   auth(authContext: AuthContext, callback: Callback): void {
     const { connection, credentials } = authContext;
-    if (credentials == null) return callback(new MongoDriverError('credentials required'));
+    if (credentials == null)
+      return callback(new MongoMissingCredentialsError('credentials required'));
```
andymina, line 24: Could this message be more descriptive to include GSSAPI?

andymina:
I can't leave comments on things you didn't change so unfortunately this will have to do for now ;(

`src/bulk/ordered.ts:29`: I think this error should also be changed to `MongoInvalidArgumentError` since the size of the doc the user supplied is too large.

andymina:
`src/cmap/commands.ts:855`: Although realistically we shouldn't run into this error, I think it'd be good practice to replace this with `MongoCompatabilityError` since it was decided that no drivers would make use of payload type 1.

andymina:
`src/cursor/abstract_cursor.ts:549`: I think this should also be `MongoInvalidArgumentError`.

andymina:
`src/operations/execute_operations.ts:130`: I think this should be `MongoInvalidArgumentError`.

src/operations/set_profiling_level.ts
```diff
@@ -50,7 +50,9 @@ export class SetProfilingLevelOperation extends CommandOperation<ProfilingLevel>
     const level = this.level;
 
     if (!levelValues.has(level)) {
-      return callback(new MongoDriverError('Error: illegal profiling level value ' + level));
+      return callback(
+        new MongoInvalidArgumentError('Error: illegal profiling level value ' + level)
```
andymina, line 54: I think the `Error:` prefix in the message might be redundant.

andymina (changes requested):

andymina:
`src/operations/update.ts:278 and 282`: I also think these should be `MongoInvalidArgumentError`.

src/error.ts
```diff
@@ -228,6 +228,44 @@ export class MongoParseError extends MongoDriverError {
   }
 }
 
+export class MongoAPIError extends MongoDriverError {
+  constructor(message: string) {
+    super(message);
+    this.name = 'MongoAPIError';
```
andymina, line 234: NODE-3368 *just* got merged which made the name prop of error classes getter functions. Moving forward names should implemented as:
```
get name(): string {
  return 'MongoAPIError';
}
```

Can you add these changes so we can stay consistent?

src/cmap/auth/gssapi.ts
W-A-James: > `src/operations/execute_operations.ts:130 and 191`: I think this should be `MongoInvalidArgumentError`.

I left these since they seem more like they would be `MongoTransactionEror` to me, but I do see an argument for having them be `MongoInvalidArgumentError`

andymina:
`src/mongo_client.ts:559`: This should be `MongoInvalidArgumentError` I think.

andymina:
`src/sessions.ts:272`: I think `MongoCompatabilityError` might fit better here.

src/utils.ts
```diff
@@ -248,7 +254,7 @@ export function executeLegacyOperation(
       const optionsIndex = args.length - 2;
       args[optionsIndex] = Object.assign({}, args[optionsIndex], { session: session });
     } else if (opOptions.session && opOptions.session.hasEnded) {
-      throw new MongoDriverError('Use of expired sessions is not permitted');
+      throw new MongoInvalidArgumentError('Use of expired sessions is not permitted');
```
andymina, line 257: I'm a little torn about this one. The same error message appears in `src/operations/execute_operation.ts:88` as well so I feel like the verdict has to be consistent for the two cases. That being said I'm not sure whether `MongoInvalidArgumentError` or `MongoResourceClosedError` would fit better here. Definitely open to other perspectives.

andymina (changes requested):

src/error.ts
W-A-James: Will do

src/utils.ts
W-A-James: I had left the error in `src/operations/execute_operation.ts:88` because of that ambiguity. Was intending to leave that until we added the `MongoResourceClosedError` and its children, so yeah, I think you're right about changing this to `MongoResourceClosedError` as well

W-A-James:
Closing for replacement with #2891

PR #2885, opened 2021-07-07
fix(NODE-3411): remove explicit fallback batch size for getMore calls

rose-m:
Removes the explicit fallback batch size of `1000` for `getMore` calls of cursors. This will cause the driver to use the default behavior of the server which is to return up to 16MB of data in the `getMore` call.

Note that the _initial_ `find` operation for example will continue to use the server's default 101 results if the user does not immediately specify a `batchSize`.

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -512,7 +512,7 @@ export class ChangeStreamCursor<TSchema extends Document = Document> extends Abs
     });
   }
 
-  _getMore(batchSize: number, callback: Callback): void {
+  _getMore(batchSize: number | undefined, callback: Callback): void {
     super._getMore(batchSize, (err, response) => {
       if (err) {
         return callback(err);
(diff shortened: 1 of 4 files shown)


rose-m:
Something @mcasimir brought up that he kind of knew from the past that I absolutely didn't have on my radar: that 1000 limit seems to be a very well known thing and people seem to rely on that.

With that in mind my change would be hugely breaking and we would have to find an alternative that only influences the `getMore()` batch sizes. One option might be to allow specifying `undefined` as `batchSize` option which would then explicitly trigger this behavior.

rose-m:
After further research I found the following in the spec for find/getMore: https://github.com/mongodb/specifications/blob/master/source/find_getmore_killcursors_commands.rst#getmore

In the table on the linked page it says the `batchSize` parameter is required (X in the "Req") column. However, the sentence immediately below that table states:
> If batchSize is equal to 0 it must be omitted.

Omitting the `batchSize` is what the old shell does and seems to be the same behavior the Java driver has for example (see https://github.com/mongodb/mongo-java-driver/blob/1294de8a3daf0bcc0403d127f8c932511a51ea18/driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java#L308).

PR #2889, opened 2021-07-07
fix(NODE-3417): allow calling `db()` before MongoClient is connected

vkarpov15:
## Description

Removed an error case that will make Automattic/mongoose#8702 much easier.

**What changed?**

Removed a check that looks unnecessary because the `Db()` constructor no longer takes a `topology` param.

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -473,13 +473,6 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
     // Copy the options and add out internal override of the not shared flag
     const finalOptions = Object.assign({}, this[kOptions], options);
 
-    // If no topology throw an error message
-    if (!this.topology) {
-      throw new MongoDriverError(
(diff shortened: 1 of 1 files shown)


vkarpov15:
@nbbeeken any thoughts on merging this now that 4.0 is officially released?

nbbeeken (approved):
Yea LGTM!

Just for posterity: 
Tracking down the origin of this error revealed that the Db class used to take the topology as the second argument to the constructor, that's no longer the case and there's no reliance on the topology existing before grabbing a Db reference.

The UX change here is when you go to run an operation the error occurs when the topology is fetched by `getTopology`:
```
Uncaught MongoDriverError: MongoClient must be connected to perform this operation
    at Object.getTopology (lib/utils.js:383:11)
    at Collection.insertOne (lib/collection.js:151:61)
```

emadum (approved):

dariakp (approved):
LGTM, thanks!

PR #2890, opened 2021-07-07
docs(NODE-2329): write upgrade and migration guide

nbbeeken:
Upgrade guide, moved the versioned api feature mention out of this file, I'm thinking we should break out features from changes, thoughts?


docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,249 @@
...
+
+`gssapiServiceName` has been removed.
+Users should use authMechanismProperties.SERVICE_NAME like so:
+
+- In a URI query param: `?authMechanismProperties=SERVICE_NAME:alternateServiceName`
+- Or as an option: `{ authMechanismProperties: { SERVICE_NAME: 'alternateServiceName' } }`
+
+### db.collection no longer accepts a callback
+
+The only option that required the use of the callback was strict mode and that option can still be used in `db.createCollection`.
```
emadum, line 94:
```suggestion
The only option that required the use of the callback was strict mode. Users who wish to ensure operations only execute against existing collections should use `db.listCollections` directly.
```

We removed strict mode from `db.createCollection` also, where it was a no-op.

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,249 @@
...
+
+We now require node 12.9 or greater for version 4 of the driver.
+If that's outside your support matrix at this time, that's okay!
+Bug fix support for our 3.x branch will not be ending until summer 2022!
+
+### Cursor changes
+
+#### AbstractCursor, FindCursor, AggregationCursor, ChangeStreamCursor, ListCollectionsCursor
+
+Our Cursor implementation has been updated to make more clear what is possible before and after execution of an operation. Take this example:
```
emadum, line 31:
```suggestion
Our Cursor implementation has been updated to clarify what is possible before and after execution of an operation. Take this example:
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,249 @@
...
+
+### NodeJS Version
+
+We now require node 12.9 or greater for version 4 of the driver.
+If that's outside your support matrix at this time, that's okay!
+Bug fix support for our 3.x branch will not be ending until summer 2022!
+
+### Cursor changes
+
+#### AbstractCursor, FindCursor, AggregationCursor, ChangeStreamCursor, ListCollectionsCursor
```
emadum, line 29: This heading is confusing to me, what is it intended to convey?

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,249 @@
...
+const gs = new GridStore(db, filename, mode[, options])
+// new way
+const bucket = new GridFSBucket(client.db('test')[, options])
 ```
 
-### Strict mode
+#### File seeking
 
-Declaring a `strict` API version will cause the MongoDB server to reject all commands that are not part of the declared API version. This includes command options and aggregation pipeline stages. For example, the following `find` call would fail because the `tailable` option is not part of version 1:
+Since GridFSBucket uses NodeJS Stream API you can replicate file seek-ing by using the start and end options creating a downloadStream from your GridFSBucket
```
emadum, line 140:
```suggestion
Since GridFSBucket uses NodeJS Stream API you can replicate file seek-ing by using the start and end options creating a download stream from your GridFSBucket
```

emadum (changes requested):
Looking great, thank you for putting this together! 👏 Had a few small requests.

docs/CHANGES_4.0.0.md
nbbeeken: I guess just listing out the affected classes, I'll make it into a list instead

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
+
+- In a URI query param: `?authMechanismProperties=SERVICE_NAME:alternateServiceName`
+- Or as an option: `{ authMechanismProperties: { SERVICE_NAME: 'alternateServiceName' } }`
+
+### db.collection no longer accepts a callback
+
+The only option that required the use of the callback was strict mode. Users who wish to ensure operations only execute against existing collections should use `db.listCollections` directly.
+The strict option would return an error if the collection does not exist.
+
+Another way to assert a collections existence is using this snippet below:
```
dariakp, line 104:
```suggestion
Another way to assert a collection's existence is using this snippet below:
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
+#### Authentication
+
+Specifying username and password as options are supported in only these two formats:
+
+- `new MongoClient(url, { auth: { username: '', password: '' } })`
+- `new MongoClient('mongodb://username:password@myDb.host')`
+
+#### Check Server Identity Inconsistency
+
+Specifying `checkServerIdentity === false` (along with enabling tls) is different from it being leaving it `undefined`.
```
dariakp, line 85:
```suggestion
Specifying `checkServerIdentity === false` (along with enabling tls) is different from leaving it `undefined`.
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
+The [differences are described in detail here](https://mongodb.github.io/node-mongodb-native/3.6/reference/unified-topology/).
 
-Declaring a `strict` API version will cause the MongoDB server to reject all commands that are not part of the declared API version. This includes command options and aggregation pipeline stages. For example, the following `find` call would fail because the `tailable` option is not part of version 1:
+Feel free to remove the `useUnifiedTopology` and `useNewUrlParser` options at your leisure, they are no longer used by the driver.
+
+**NOTE:** With the unified topology, in order to connect to replicaSet nodes that have not been initialized you must use the new `directConnection` option.
+
+#### Authentication
+
+Specifying username and password as options are supported in only these two formats:
```
dariakp, line 78:
```suggestion
Specifying username and password as options is only supported in these two formats:
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
+
+#### Stream API
+
+The Cursor no longer extends Readable directly, it must be transformed into a stream by calling cursor.stream(), for example:
 
-cursor = client.db('database').collection('coll').find(...);
+```javascript
+const cursor = collection.find({});
+const stream = cursor.stream()
+stream.on("data", data => console.log(data));
```
dariakp, line 57: it looks like we use single quotes for strings everywhere else, would probably be good to keep it consistent here and in the line below, too

dariakp (changes requested):

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
-### Declare an API version on a client:
+We've migrated the driver to Typescript! Users can now harness the power of type hinting and intellisense in editors that support it to develop their MongoDB applications. Users don't even have to adopt Typescript to get the benefits. Along with the type hinting there's consistent docs formatting that editors should be able to display while developing.
+Recently we migrated our BSON library to TypeScript as well, this version of the driver pulls in those changes.
+
+#### Community Types users (@types/mongodb)
+
+If you are a user of the community types (@types/mongodb) there will likely be compilation errors while adopting the types from our codebase.
+Unfortunately we could not achieve a one to one match in types due to the details of writing the codebase in Typescript vs definitions for the user layer API along with the breaking changes of this major version. Please let us know if there's anything that is a blocker to upgrading [on JIRA](https://jira.mongodb.org/browse/NODE).
+
+### NodeJS Version
```
addaleax, line 21:
```suggestion
### Node.js Version
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
+
+- `new MongoClient(url, { auth: { username: '', password: '' } })`
+- `new MongoClient('mongodb://username:password@myDb.host')`
+
+#### Check Server Identity Inconsistency
+
+Specifying `checkServerIdentity === false` (along with enabling tls) is different from it being leaving it `undefined`.
+The 3.x version intercepted `checkServerIdentity: false` and turned it into a no-op function which is the required way to skip checking the server identity by nodejs.
+Setting this option to `false` is only for testing anyway as it disables essential verification to TLS.
+So it made sense for our library to directly expose the option validation from nodejs.
```
addaleax, line 88:
```suggestion
So it made sense for our library to directly expose the option validation from Node.js.
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
+```javascript
+// old way
+const gs = new GridStore(db, filename, mode[, options])
+// new way
+const bucket = new GridFSBucket(client.db('test')[, options])
+```
+
+#### File seeking
+
+Since GridFSBucket uses NodeJS Stream API you can replicate file seek-ing by using the start and end options creating a download stream from your GridFSBucket
```
addaleax, line 147:
```suggestion
Since GridFSBucket uses the Node.js Stream API you can replicate file seek-ing by using the start and end options creating a download stream from your GridFSBucket
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
+const db = client.db()
+const bucket = new GridFSBucket(db)
+
+fs.createReadStream(filename)
+    .pipe(bucket.openUploadStream(filename))
+    .on('error', console.error)
+    .on('finish', () => {
+    console.log('done writing to db!')
+
+    bucket.find().toArray().then((files) => {
```
addaleax, line 167: The indentation here is a bit odd?

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,256 @@
...
+- `MongoClient.prototype.logout`
+- `MongoClient.prototype.addUser: Creating a user without roles`
+- `MongoClient.prototype.connect`
+- `Remove MongoClient.isConnected` - calling connect is a no-op if already connected
+- `Remove MongoClient.logOut`
+- `require('mongodb').instrument`
+  - Use command monitoring: `client.on('commandStarted', (ev) => {})`
+- Top-Level export no longer a function: `typeof require('mongodb') !== 'function'`
+  - Must construct a MongoClient and call `.connect()` on it.
+- No more `Symbol` export, now `BSONSymbol` which is a deprecated BSON type
```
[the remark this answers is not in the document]
nbbeeken: Wrote some text, looks good?

docs/CHANGES_4.0.0.md
nbbeeken: Ran prettier, so all code blocks should be pretty now 😄

docs/CHANGES_4.0.0.md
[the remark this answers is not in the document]
dariakp: maybe change the order and make one sentence?
"Existing bson symbols in your database will be deserialized to a BSONSymbol instance; however, users should use plain strings instead of BSONSymbol"

dariakp (changes requested):
Resolved all comments, just the BSONSymbol wording still an open question

emadum:
> moved the versioned api feature mention out of this file, I'm thinking we should break out features from changes, thoughts?

SGTM, there is also now a page in the official driver docs describing the feature, we don't need to duplicate that info here.

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,273 @@
...
+
+- In a URI query param: `?authMechanismProperties=SERVICE_NAME:alternateServiceName`
+- Or as an option: `{ authMechanismProperties: { SERVICE_NAME: 'alternateServiceName' } }`
+
+### db.collection no longer accepts a callback
+
+The only option that required the use of the callback was strict mode. Users who wish to ensure operations only execute against existing collections should use `db.listCollections` directly.
+The strict option would return an error if the collection does not exist.
+
+Another way to assert a collection's existence is using this snippet below:
```
emadum, line 104:
```suggestion
The only option that required the use of the callback was strict mode. 
The strict option would return an error if the collection does not exist.
Users who wish to ensure operations only execute against existing collections should use `db.listCollections` directly.

For example:
```
Some re-ordering to make things clearer.

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,273 @@
...
 
 ```javascript
-// Declare API version "1" for the client
-client = new MongoClient(uri, { serverApi: { version: '1' } });
+const collections = (await db.listCollections({}, { nameOnly: true }).toArray()).map(
+  ({ name }) => name
+); // map to get string[]
+if (!collections.includes(myNewCollectionName)) {
+  await db.createCollection(myNewCollectionName);
```
emadum, line 111:
```suggestion
  throw new Error(`${myCollectionName} doesn't exist`);
```
This change makes the snippet closer to the old behavior.

emadum (changes requested):

docs/CHANGES_4.0.0.md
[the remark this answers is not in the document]
nbbeeken: SGTM, done!

emadum (changes requested):
Noticed a few more things and had a few styling nits, I think we're getting close 👍

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,274 @@
...
+```javascript
+const cursor = collection.find({ a: 2.3 }).skip(1);
+for await (const doc of cursor) {
+  console.log(doc);
+  fc.limit(1); // bad.
+}
+```
+
+Prior to the this release there was inconsistency surrounding how the cursor would error if a setting like limit was applied after cursor execution had begun. Now, an error along the lines of: `Cursor is already initialized` is thrown.
+
```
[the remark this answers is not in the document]
nbbeeken: Yes! added, thoughts?

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,297 @@
...
+Prior to the this release there was inconsistency surrounding how the cursor would error if a setting like limit was applied after cursor execution had begun.
+Now, an error along the lines of: `Cursor is already initialized` is thrown.
 
-cursor = client.db('database').collection('coll').find(...);
+#### ChangeStream must be used as an iterator or an event emitter
+
+You cannot use ChangeStream as iterator after using as an EventEmitter nor visa versa.
+Previously the driver would permit this kind of usage but with undefined results of which type of usage would actually receive the change documents.
+It's unlikely this kind of usage was useful but to be sure we can now warn users about the risk.
```
emadum, line 58:
```suggestion
You cannot use ChangeStream as an iterator after using as an EventEmitter nor visa versa.
Previously the driver would permit this kind of usage but it could lead to unpredictable behavior and obscure errors.
It's unlikely this kind of usage was useful but to be sure we now prevent it by throwing a clear error.
```

docs/CHANGES_4.0.0.md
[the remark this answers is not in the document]
emadum: LGTM, had a few suggestions

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,297 @@
...
+The Db instance is no longer an EventEmitter, all events your application is concerned with can be listened to directly from the `MongoClient` instance.
+
+### Collection.group() removed
+
+The collection `group()` helper has been deprecated in MongoDB since 3.4 and is now removed from the Driver.
+The same functionality can be achieved using the aggregation pipeline's `$group` operator.
+
+### GridStore removed
+
+The deprecated GirdStore API has been removed from the driver.
```
emadum, line 156:
```suggestion
The deprecated GridStore API has been removed from the driver.
```

emadum (changes requested):

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,297 @@
-# Changes in 4.x
+# Changes in 4.x (and how to migrate!)
 
-WIP
+_Hello dear reader, **thank you** for adopting version 4.x of the MongoDB Node.js driver, from the bottom of our developer hearts we thank you so much for taking the time to upgrade to our latest and greatest offering of a stunning database experience.
+We hope you enjoy your upgrade experience and this guide gives you all the answers you are searching for.
+If anything, and we mean anything, hinders you upgrade experience please let us know via [JIRA](https://jira.mongodb.org/browse/NODE).
```
emadum, line 5:
```suggestion
If anything, and we mean anything, hinders your upgrade experience please let us know via [JIRA](https://jira.mongodb.org/browse/NODE).
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,297 @@
...
 
 ```javascript
-// Declare API version "1" for the client, with strict on
-client = new MongoClient(uri, { serverApi: { version: '1', strict: true } });
+const cursor = collection.find({});
+const stream = cursor.stream();
+stream.on('data', data => console.log(data));
+stream.on('end', () => client.close());
+```
+
```
emadum, line 84:
```suggestion

`Cursor.transformStream()` has been removed. `Cursor.stream()` accepts a transform function, so that API was redundant.
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,297 @@
...
+Specifying `checkServerIdentity === false` (along with enabling tls) is different from leaving it `undefined`.
+The 3.x version intercepted `checkServerIdentity: false` and turned it into a no-op function which is the required way to skip checking the server identity by nodejs.
+Setting this option to `false` is only for testing anyway as it disables essential verification to TLS.
+So it made sense for our library to directly expose the option validation from Node.js.
+If you need to test TLS connections without verifying server identity pass in `{ checkServerIdentity: () => {} }`.
+
+#### Kerberos / GSSAPI
+
+`gssapiServiceName` has been removed.
+Users should use authMechanismProperties.SERVICE_NAME like so:
```
emadum, line 118:
```suggestion
Users should use `authMechanismProperties.SERVICE_NAME` like so:
```

docs/CHANGES_4.0.0.md
```diff
@@ -1,39 +1,297 @@
...
+Users testing for `BulkWriteError`s should be sure to import the new class name `MongoBulkWriteError`.
 
-The `deprecationErrors` option can be used to enable command failures when using functionality that is deprecated from version 1. Note that at the time of this writing, no deprecations in version 1 exist.
+### Db no longer emits events
+
+The Db instance is no longer an EventEmitter, all events your application is concerned with can be listened to directly from the `MongoClient` instance.
+
+### Collection.group() removed
+
+The collection `group()` helper has been deprecated in MongoDB since 3.4 and is now removed from the Driver.
```
emadum, line 151:
```suggestion
The collection `group()` helper has been deprecated in MongoDB since 3.4 and is now removed from the driver.
```

emadum (changes requested):
Probably the final round of feedback 🚀

emadum (approved):
LGTM 🎉

dariakp (approved):
LGTM

docs/CHANGES_4.0.0.md
```diff
@@ -227,6 +228,40 @@ File metadata that used to be accessible on the GridStore instance can be found
 const fileMetaDataList: GridFSFile[] = bucket.find({}).toArray();
 ```
 
+#### Hashing an upload
+
+The automatic MD5 hashing has been removed from the upload family of functions.
+This makes the default Grid FS behavior compliant with systems that do not permit usage of MD5 hashing.
+The `disableMD5` option is no longer used and has no effect.
+
+If you still want to add an MD5 hash to your file upload here's a simple example that can be used with [any hashing algorithm](https://nodejs.org/dist/latest-v14.x/docs/api/crypto.html#crypto_crypto_createhash_algorithm_options) provided by Node.js:
```
dariakp, line 237: should have a `,` after "file upload"

emadum (approved):

PR #2891, opened 2021-07-07
refactor(NODE-3402): implement MongoAPIError and its children

W-A-James:
## Description

Implement the errors detailed in [docs/errors.md](https://github.com/mongodb/node-mongodb-native/blob/4.0/docs/errors.md) from NODE-3363 that fall under MongoAPIError

**What changed?**
Replacing MongoDriverError instances with the following errors where appropriate

- MongoInvalidArgumentError
- MongoMissingDependencyError
- MongoMissingCredentialsError
- MongoCompatibilityError


W-A-James:
Currently going back through the driver to locate anywhere I may have missed that would benefit from a child instance of `MongoAPIError` in place of `MongoDriverError`

src/db.ts
```diff
@@ -324,6 +324,7 @@ export class Db {
     if (!options) {
       options = {};
     } else if (typeof options === 'function') {
+      // TODO: Replace this error with a more appropriate error class
```
W-A-James, line 327: Considering making this a MongoCompatibilityError

emadum: I think it's good to only use MongoCompatiblityError when the server doesn't support a given feature. I'd leave this as is or maybe introduce a MongoDeprecationError that is thrown when a user tries to use a previously deprecated functionality that is no longer supported. This error is only here as a convenience to users who are upgrading, so my vote would be to just leave as is, maybe with a note to remove it entirely in driver 5.0.

andymina:
`src/bulk/common.ts:1150`: I think it might be worth to make this `MongoInvalidArgumentError`.

andymina:
`src/cmap/auth/scram.ts:31`: Should this be `MongoMissingCredentials`? My reasoning is that it would fit with the error message thrown here more, but I'm not too sure.

andymina:
`src/cmap/commands.ts:85`: I'm not quite sure how I feel about this being `MongoInvalidArgumentError. I feel like `MongoURIError` might be a little more fitting here since the string to be parsed has an invalid char. Happy to defer to the rest of them.

andymina:
`src/cmap/commands.ts:855`: Although realistically we shouldn't run into this error, I think it'd be good practice to replace this with `MongoCompatabilityError` since it was decided that no drivers would make use of payload type 1.

andymina:
`src/cmap/connection.ts:661`: This should be `MongoCompatabilityError`.

W-A-James:
> src/cmap/commands.ts:85: I'm not quite sure how I feel about this being MongoInvalidArgumentError. I feel like MongoURIError might be a little more fitting here since the string to be parsed has an invalid char. Happy to defer to the rest of them.

@andymina 
I think it might be worth keeping MongoParseError since in instances like this, we're not doing anything with the URI string itself, but rather just parsing a namespace string (`<database>.<collection>`)

andymina:
> > src/cmap/commands.ts:85: I'm not quite sure how I feel about this being MongoInvalidArgumentError. I feel like MongoURIError might be a little more fitting here since the string to be parsed has an invalid char. Happy to defer to the rest of team.
> 
> @andymina
> I think it might be worth keeping MongoParseError since in instances like this, we're not doing anything with the URI string itself, but rather just parsing a namespace string (`<database>.<collection>`)

From the description of MongoURIError that we've got in `docs/errors.md`, I think this use case should be covered by it. Although if you think it's worth to split the two and reintroduce MongoParseError, I'm down.

andymina:
`src/cmap/connection.ts:651`: This should `MongoMissingDependencyError` per [NODE-3633](https://github.com/mongodb/node-mongodb-native/pull/2862#discussion_r660016883).

src/change_stream.ts
```diff
@@ -357,7 +363,7 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
    */
   stream(options?: CursorStreamOptions): Readable {
     this.streamOptions = options;
-    if (!this.cursor) throw new MongoDriverError(NO_CURSOR_ERROR);
+    if (!this.cursor) throw new MongoInvalidArgumentError(NO_CURSOR_ERROR);
```
andymina, line 366: Happy to merge this for now, but I think this should ultimately end up as `MongoChangeStreamError` which is covered by [NODE-3404](https://jira.mongodb.org/browse/NODE-3404).

src/db.ts
```diff
@@ -724,15 +725,15 @@ export class Db {
...
   if (databaseName.length === 0)
-    throw new MongoDriverError('database name cannot be the empty string');
+    throw new MongoInvalidArgumentError('database name cannot be the empty string');
   if (databaseName === '$external') return;
 
   const invalidChars = [' ', '.', '$', '/', '\\'];
   for (let i = 0; i < invalidChars.length; i++) {
     if (databaseName.indexOf(invalidChars[i]) !== -1)
-      throw new MongoDriverError(
+      throw new MongoInvalidArgumentError(
```
andymina, line 736: This falls under the [earlier convo](https://github.com/mongodb/node-mongodb-native/pull/2891#issuecomment-876668400) we had. I wanted raise this so we can be consistent for whatever decision we make.

andymina:
`src/encrypter.ts:29`: Should this also be `MongoInvalidArgumentError`?

src/utils.ts
```diff
@@ -248,7 +254,7 @@ export function executeLegacyOperation(
       const optionsIndex = args.length - 2;
       args[optionsIndex] = Object.assign({}, args[optionsIndex], { session: session });
     } else if (opOptions.session && opOptions.session.hasEnded) {
-      throw new MongoDriverError('Use of expired sessions is not permitted');
+      throw new MongoInvalidArgumentError('Use of expired sessions is not permitted');
```
andymina, line 257: Again, happy to merge this for now but I think this would fall under `MongoExpiredSessionError` which would be under [NODE-3405](https://jira.mongodb.org/browse/NODE-3405).

src/utils.ts
```diff
@@ -574,7 +582,7 @@ export class MongoDBNamespace {
 
   static fromString(namespace?: string): MongoDBNamespace {
     if (!namespace) {
-      throw new MongoDriverError(`Cannot parse namespace from "${namespace}"`);
+      throw new MongoInvalidArgumentError(`Cannot parse namespace from "${namespace}"`);
```
andymina, line 585: It seems like an error is only throw when namespace is the empty string. I'm in favor of changing the error message to reflect that, but I'll defer to the rest of the team since I don't know all of the context behind this test.

andymina (changes requested):
I think once the changes requested in the comments are resolved, LGTM.

W-A-James:
> > > src/cmap/commands.ts:85: I'm not quite sure how I feel about this being MongoInvalidArgumentError. I feel like MongoURIError might be a little more fitting here since the string to be parsed has an invalid char. Happy to defer to the rest of team.
> > 
> > @andymina
> > I think it might be worth keeping MongoParseError since in instances like this, we're not doing anything with the URI string itself, but rather just parsing a namespace string (`<database>.<collection>`)

> @W-A-James 
> From the description of MongoURIError that we've got in `docs/errors.md`, I think this use case should be covered by it. Although if you think it's worth to split the two and reintroduce MongoParseError, I'm down.

@andymina
From the description we have for MongoURIError, namespaces, outside the context of connecting to the server itself, seem like they come up enough to justify having MongoParseError continue to be a separate error. There are other cases (src/cmap/message_stream.ts: 138, 144 & src/sdam/topology.ts: 279) where MongoParseError fits better than MongoURIError in my opinion, so my vote is to keep MongoParseError and add it to our hierarchy alongside MongoURIError

andymina:
> > > > src/cmap/commands.ts:85: I'm not quite sure how I feel about this being MongoInvalidArgumentError. I feel like MongoURIError might be a little more fitting here since the string to be parsed has an invalid char. Happy to defer to the rest of team.
> > > 
> > > 
> > > @andymina
> > > I think it might be worth keeping MongoParseError since in instances like this, we're not doing anything with the URI string itself, but rather just parsing a namespace string (`<database>.<collection>`)
> 
> > @W-A-James
> > From the description of MongoURIError that we've got in `docs/errors.md`, I think this use case should be covered by it. Although if you think it's worth to split the two and reintroduce MongoParseError, I'm down.
> 
> @andymina
> From the description we have for MongoURIError, namespaces, outside the context of connecting to the server itself, seem like they come up enough to justify having MongoParseError continue to be a separate error. There are other cases (src/cmap/message_stream.ts: 138, 144 & src/sdam/topology.ts: 279) where MongoParseError fits better than MongoURIError in my opinion, so my vote is to keep MongoParseError and add it to our hierarchy alongside MongoURIError

@W-A-James sounds good to me, lets do it

src/utils.ts
W-A-James: It also throws when no argument is provided and I think the error message lets you know what went wrong pretty well

The message would look something like 

`MongoInvalidArgumentError: Cannot parse namespace from ""`

or 

`MongoInvalidArgumentError: Cannot parse namespace from "undefined"`

If you're still in favour to change it, I can go ahead, but that seems fairly clear to me

andymina: Ahh good point, don't think needs a change then.

andymina (approved):
LGTM, nice job

src/cmap/commands.ts
```diff
@@ -77,12 +77,12 @@ export class Query {
 
   constructor(ns: string, query: Document, options: OpQueryOptions) {
     // Basic options needed to be passed in
-    if (ns == null) throw new MongoDriverError('ns must be specified for query');
-    if (query == null) throw new MongoDriverError('query must be specified for query');
+    if (ns == null) throw new MongoInvalidArgumentError('ns must be specified for query');
```
nbbeeken, line 80:
```suggestion
    if (ns == null) throw new MongoInvalidArgumentError('Namespace must be specified for query');
```

src/cmap/commands.ts
```diff
@@ -852,7 +852,7 @@ export class BinMsg {
         this.index += bsonSize;
       } else if (payloadType === 1) {
         // It was decided that no driver makes use of payload type 1
-        throw new MongoDriverError('OP_MSG Payload Type 1 detected unsupported protocol');
+        throw new MongoInvalidArgumentError('OP_MSG Payload Type 1 detected unsupported protocol');
```
nbbeeken, line 855: This is actually something a poorly behaving server would send us, its an unused payload type. So this wouldn't be an argument error, its still a driver error in the sense that the driver doesn't have the logic to handle that payload type, is there something better that fits this?

src/cursor/abstract_cursor.ts
```diff
@@ -319,7 +319,7 @@ export abstract class AbstractCursor<
     callback?: Callback<void>
   ): Promise<void> | void {
     if (typeof iterator !== 'function') {
-      throw new MongoDriverError('Missing required parameter `iterator`');
+      throw new MongoInvalidArgumentError('Missing required parameter `iterator`');
```
nbbeeken, line 322:
```suggestion
      throw new MongoInvalidArgumentError('Iterator must be a function');
```
This is the same comment along the lines of trying to make errors state what "must be"

src/operations/execute_operation.ts
```diff
@@ -85,12 +87,13 @@ export function executeOperation<
         owner = Symbol();
         session = topology.startSession({ owner, explicit: false });
       } else if (session.hasEnded) {
+        // TODO: Change this out for MongoExpiredSessionError
```
nbbeeken, line 90: Maybe we should put ticket numbers on these todos?

src/operations/profiling_level.ts
```diff
@@ -17,14 +17,17 @@ export class ProfilingLevelOperation extends CommandOperation<string> {
...
   execute(server: Server, session: ClientSession, callback: Callback<string>): void {
     super.executeCommand(server, session, { profile: -1 }, (err, doc) => {
       if (err == null && doc.ok === 1) {
         const was = doc.was;
         if (was === 0) return callback(undefined, 'off');
         if (was === 1) return callback(undefined, 'slow_only');
         if (was === 2) return callback(undefined, 'all');
-        return callback(new MongoDriverError('Error: illegal profiling level value ' + was));
+        return callback(
+          new MongoInvalidArgumentError('Error: illegal profiling level value ' + was)
```
nbbeeken, line 29:
```suggestion
          new MongoInvalidArgumentError('Illegal profiling level value ' + was)
```

src/operations/set_profiling_level.ts
```diff
@@ -50,9 +50,10 @@ export class SetProfilingLevelOperation extends CommandOperation<ProfilingLevel>
     const level = this.level;
 
     if (!levelValues.has(level)) {
-      return callback(new MongoDriverError('Error: illegal profiling level value ' + level));
+      return callback(new MongoInvalidArgumentError('Illegal profiling level value ' + level));
     }
 
+    // TODO: Determine error to put here
```
nbbeeken, line 56: What errors?

src/operations/validate_collection.ts
```diff
@@ -40,6 +40,7 @@ export class ValidateCollectionOperation extends CommandOperation<Document> {
     super.executeCommand(server, session, this.command, (err, doc) => {
       if (err != null) return callback(err);
 
+      // TODO: Decide on error to put here
```
nbbeeken, line 43: What error?

src/sdam/topology.ts
```diff
@@ -274,6 +274,8 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
       } else if (seed instanceof HostAddress) {
         seedlist.push(seed);
       } else {
+        // NODE-3402
+        // FIXME: May need to me a MongoParseError
```
nbbeeken, line 278:
```suggestion
        // FIXME(NODE-3402): May need to be a MongoParseError
```

src/sdam/topology_description.ts
```diff
@@ -431,7 +431,7 @@ function updateRsWithPrimaryFromMember(
   setName?: string
 ): TopologyType {
   if (setName == null) {
-    throw new MongoDriverError('setName is required');
+    throw new MongoInvalidArgumentError('setName is required');
```
nbbeeken, line 434: Weirdly this seems to me to be a MongoCompatibilityError, we've been using that to mean version incompatibility but here it would mean topology incompatibility. I'm fine with leaving this as is, but its an interesting case of if the setName is missing then we're not connected to a replica set so trying to `updateRsWithPrimaryFromMember` is illogical. 

But anyway this is internal logic that ideally would be caught in testing, it wouldn't surface

nbbeeken (changes requested):

nbbeeken:
@W-A-James Our linter is supposed to make all `@public` symbols forced to be exported at the top-level `src/index.ts` I'm not sure why it isn't forcing it here, can you make sure all the new errors are added to the export here: https://github.com/mongodb/node-mongodb-native/blob/4.0/src/index.ts#L32-L42

nbbeeken:
> > > > > src/cmap/commands.ts:85: I'm not quite sure how I feel about this being MongoInvalidArgumentError. I feel like MongoURIError might be a little more fitting here since the string to be parsed has an invalid char. Happy to defer to the rest of team.
> > > > 
> > > > 
> > > > @andymina
> > > > I think it might be worth keeping MongoParseError since in instances like this, we're not doing anything with the URI string itself, but rather just parsing a namespace string (`<database>.<collection>`)
> > 
> > 
> > > @W-A-James
> > > From the description of MongoURIError that we've got in `docs/errors.md`, I think this use case should be covered by it. Although if you think it's worth to split the two and reintroduce MongoParseError, I'm down.
> > 
> > 
> > @andymina
> > From the description we have for MongoURIError, namespaces, outside the context of connecting to the server itself, seem like they come up enough to justify having MongoParseError continue to be a separate error. There are other cases (src/cmap/message_stream.ts: 138, 144 & src/sdam/topology.ts: 279) where MongoParseError fits better than MongoURIError in my opinion, so my vote is to keep MongoParseError and add it to our hierarchy alongside MongoURIError
> 
> @W-A-James SGTM, lets see what the rest of the team thinks because this would likely cause another PR to 4.0 to update the doc

I think maybe we consider a `MongoInvalidNamespaceError` which could be useful in the Db and collection constructor as well as MongoDBNameSpace class. ParseError still feels too generic to communicate effectively. Thoughts? We could add that here and make a follow up ticket to update the doc

src/cmap/auth/gssapi.ts
```diff
@@ -15,7 +20,10 @@ import * as dns from 'dns';
 export class GSSAPI extends AuthProvider {
   auth(authContext: AuthContext, callback: Callback): void {
     const { connection, credentials } = authContext;
-    if (credentials == null) return callback(new MongoDriverError('credentials required'));
+    if (credentials == null)
+      return callback(
+        new MongoMissingCredentialsError('credentials required for gssapi authentication')
```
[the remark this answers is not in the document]
W-A-James: Does this also apply to errors that start by referencing a variable name? If so, does that mean that we should restructure the message so that this is not the case?

For example, in `src/operations/update.ts: 131` we have the line:

```javascript
callback(new MongoCompatibilityError('arrayFilters are only supported in MongoDB 3.6+'));
```

Should this be changed to something like: 

```javascript
callback(new MongoCompatibilityError('The "arrayFilters" option is only supported in MongoDB 3.6+'));
```

src/cmap/commands.ts
W-A-James: At the moment, we don't have an error that covers this cleanly, but we could conceivably have something like a `MongoInvalidServerResponseError` or a `MongoUnexpectedServerResponseError` as a subclass of `MongoRuntimeError`

W-A-James:
@andymina and I have discussed and come to the conclusion that we need a few more error classes to cover the few cases that I have marked with `TODO` or `FIXME` comments. The new errors are enumerated here:

## MongoDeprecationError

**extends MongoAPIError**

An error encountered when a user calls a function or uses a feature that is deprecated in the current driver version. This should serve both as a a tool for the driver team to softly deprecate a feature if its removal would break other components as well as a warning for users that a feature will be removed in a future version.

## MongoUnexpectedServerResponseError

**extends MongoRuntimeError**

An error encountered when the driver receives a response from the server that is not interpreted as a ServerError, but also cannot be properly interpreted or used by the driver. The case for this error was made with `src/cmap/commands.ts:855`

## MongoGridFSFileNotFoundError

**extends MongoRuntimeError**

An error encountered when an attempt is made to operate on a non-existent GridFS file. This is placed here since it seems as though it doesn't wrap an existent ENOENT error.

## MongoCommandOperationError

**extends MongoRuntimeError**

An error encountered when an operation on the documents in a collection is encountered.

### Unknown

* How much does the _source_ of the error that is encountered matter?
* Can we cover the range of sources in the error message?

src/cmap/auth/gssapi.ts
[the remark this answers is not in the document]
nbbeeken: I guess that would be an exception to the rule, fine to leave that lowercase

src/cmap/commands.ts
nbbeeken: So for the sake of this PR do we want to leave this as a driver error and we'll leave a todo with a ticket that'll bring in the new proposed error?
`MongoInvalidResponseError` seems like it could work

src/operations/insert.ts
```diff
@@ -100,7 +100,7 @@ export class InsertManyOperation extends AbstractOperation<InsertManyResult> {
     super(options);
 
     if (!Array.isArray(docs)) {
-      throw new MongoDriverError('docs parameter must be an array of documents');
+      throw new MongoInvalidArgumentError('Parameter "docs" must be an array of documents');
```
nbbeeken, line 103: Okay this is better than the `The xx option must be` in the other comment 🚀

src/operations/update.ts
```diff
@@ -108,25 +108,29 @@ export class UpdateOperation extends CommandOperation<Document> {
...
+      callback(
+        new MongoCompatibilityError(`Server ${server.name} does not support explain on update`)
+      );
       return;
     }
 
     if (this.statements.some(statement => !!statement.arrayFilters) && maxWireVersion(server) < 6) {
-      callback(new MongoDriverError('arrayFilters are only supported on MongoDB 3.6+'));
+      callback(
+        new MongoCompatibilityError('The "arrayFilters" option is only supported on MongoDB 3.6+')
```
nbbeeken, line 132: `Option "arrayFilters" is only..`

nbbeeken (changes requested):

src/cmap/wire_protocol/shared.ts
```diff
@@ -28,7 +28,7 @@ export function getReadPreference(cmd: Document, options?: ReadPreferenceOption)
   }
 
   if (!(readPreference instanceof ReadPreference)) {
-    throw new MongoDriverError('read preference must be a ReadPreference instance');
+    throw new MongoInvalidArgumentError('Read preference must be a ReadPreference instance');
```
nbbeeken, line 31:
```suggestion
    throw new MongoInvalidArgumentError('Option readPreference must be a ReadPreference instance');
```

src/change_stream.ts
```diff
@@ -255,8 +261,8 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
     } else if (parent instanceof MongoClient) {
       this.type = CHANGE_DOMAIN_TYPES.CLUSTER;
     } else {
-      throw new MongoDriverError(
-        'parent provided to ChangeStream constructor is not an instance of Collection, Db, or MongoClient'
+      throw new MongoInvalidArgumentError(
+        'parent provided to ChangeStream constructor should be an instance of Collection, Db, or MongoClient'
```
nbbeeken, line 265:
```suggestion
        'Parent provided to ChangeStream constructor must be an instance of Collection, Db, or MongoClient'
```

src/collection.ts
```diff
@@ -1448,8 +1456,8 @@ export class Collection<TSchema extends Document = Document> {
     // Out must always be defined (make sure we don't break weirdly on pre 1.8+ servers)
     // TODO NODE-3339: Figure out if this is still necessary given we no longer officially support pre-1.8
     if (options?.out == null) {
-      throw new MongoDriverError(
-        'the out option parameter must be defined, see mongodb docs for possible values'
+      throw new MongoInvalidArgumentError(
+        'The "out" option parameter must be defined, see mongodb docs for possible values'
```
nbbeeken, line 1460: `Option "out"...`

src/sort.ts
```diff
@@ -45,7 +45,11 @@ function prepareDirection(direction: any = 1): SortDirectionForCmd {
     case '-1':
       return -1;
     default:
-      throw new MongoDriverError(`Invalid sort direction: ${JSON.stringify(direction)}`);
+      throw new MongoInvalidArgumentError(
+        `Invalid sort direction: ${JSON.stringify(
+          direction
+        )}\nSort direction must be one of 'ascending', 'asc', '1', 'descending', 'desc', '-1'`
```
nbbeeken, line 51: Maybe we can omit the possible correct values here. We stringify the input at the top of the function to make the switch statement simple but actually `1` and `-1` shouldn't be in quotes.

src/utils.ts
```diff
@@ -289,7 +295,9 @@ export function executeLegacyOperation(
 
   // Return a Promise
   if (args[args.length - 1] != null) {
-    throw new MongoDriverError('final argument to `executeLegacyOperation` must be a callback');
+    throw new MongoInvalidArgumentError(
+      'final argument to `executeLegacyOperation` must be a callback'
```
nbbeeken, line 299:
```suggestion
      'Final argument to `executeLegacyOperation` must be a callback'
```

nbbeeken (changes requested):
Sorry for the broken up review, great work! just some details left

src/operations/insert.ts
W-A-James: I'm realizing I made some of them be in the form `Parameter "xx" must ..."` and some of them in the form `Argument "xx" must ...`

I think I'm going to change all of them to be in the second form to match better with the naming of `MongoInvalidArgumentError`

src/cmap/commands.ts
W-A-James: I'll go ahead and change that back to a MongoDriverError and add that todo 👍🏿

andymina (changes requested):

src/encrypter.ts
```diff
@@ -26,7 +26,7 @@ export class Encrypter {
 
   constructor(client: MongoClient, uri: string, options: MongoClientOptions) {
     if (typeof options.autoEncryption !== 'object') {
-      throw new MongoDriverError('Options autoEncryption must be specified');
+      throw new MongoInvalidArgumentError('Options autoEncryption must be specified');
```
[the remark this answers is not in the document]
W-A-James: I think I can keep it as `Option "autoEncryption" must be specified in argument "options"`

src/operations/find_and_modify.ts
```diff
@@ -203,7 +205,7 @@ export class FindOneAndDeleteOperation extends FindAndModifyOperation {
   constructor(collection: Collection, filter: Document, options: FindOneAndDeleteOptions) {
     // Basic validation
     if (filter == null || typeof filter !== 'object') {
-      throw new MongoDriverError('Filter parameter must be an object');
+      throw new MongoInvalidArgumentError('Filter parameter must be an object');
```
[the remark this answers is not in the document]
W-A-James: Talked to Neal and we've agreed on doing `Argument "\<argument name\>" must be ...`
So yeah, I'll go ahead and fix it

src/change_stream.ts
emadum: I think this should be changed back to `MongoDriverError`, at least until `MongoChangeStreamError` is introduced. A change stream not having an internal cursor would not be a result of invalid user input.

src/db.ts
```diff
@@ -324,6 +324,7 @@ export class Db {
     if (!options) {
       options = {};
     } else if (typeof options === 'function') {
+      // TODO: Remove this error and refactor this function in driver 5.0
```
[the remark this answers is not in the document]
W-A-James: I added this comment here by Eric's suggestion, but I'll add the ticket number in the comment

nbbeeken (approved):
Just a nit but LGTM

src/cursor/abstract_cursor.ts
```diff
@@ -517,7 +517,7 @@ export abstract class AbstractCursor<
     } else if (typeof readPreference === 'string') {
       this[kOptions].readPreference = ReadPreference.fromString(readPreference);
     } else {
-      throw new MongoDriverError('Invalid read preference: ' + readPreference);
+      throw new MongoInvalidArgumentError('Invalid read preference: ' + readPreference);
```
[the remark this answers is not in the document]
W-A-James: Gonna go ahead and replace as many of those concatenations as I can with template strings now 👍🏿

andymina (approved):
LGTM

emadum (changes requested):
Other than that one request this LGTM! 👍

nbbeeken (approved):
LGTM (needs the link fix ^ and a rebase too though)

emadum (approved):
After a second pass over the errors and thinking about it more, I retract the concern I brought up in our Zoom @W-A-James 

LGTM 👍

src/gridfs/download.ts
```diff
@@ -391,26 +391,16 @@ function handleStartOption(
...
-          'more than the length of the file (' +
-          doc.length +
-          ')'
+      throw new MongoInvalidArgumentError(
+        `Stream start (${options.start}) must not be more than the length of the file (${doc.length})`
       );
     }
     if (options.start < 0) {
-      throw new MongoDriverError('Stream start (' + options.start + ') must not be ' + 'negative');
+      throw new MongoDriverError(`Stream start (${options.start}) must not be negative`);
```
andymina, line 399: This should probably be `MongoInvalidArgumentError`.

andymina (changes requested):

andymina:
`src/sessions.ts:142`: This seems new and should likely be a MongoInvalidArgumentError.

andymina:
`src/cmap/connect.ts:150, 196, 204`: These are all of the same type of error which I believe can be replaced with `MongoInvalidArgumentError`.

andymina:
`src/cmap/auth/auth_provider.ts:57`: This is a funky one, but I feel like it may fall under `MongoInvalidArugmentError` because the subclass the user provided fails to override this method.

andymina:
`src/cmap/mongo_credentials.ts:134, 141, 149`: These are the same general type of error which may benefit from `MongoInvalidArgumentError`. I'm not quite sure where else they'd fit unless we revive `MongoAuthMechanismError`.

andymina:
`src/cmap/auth/gssapi.ts:77`: This should likely be `MongoInvalidArgumentError`.

andymina (approved):

src/operations/set_profiling_level.ts
```diff
@@ -50,9 +51,14 @@ export class SetProfilingLevelOperation extends CommandOperation<ProfilingLevel>
...
     if (!levelValues.has(level)) {
-      return callback(new MongoDriverError('Error: illegal profiling level value ' + level));
+      return callback(
+        new MongoInvalidArgumentError(
+          `Profiling level must be one of "${enumToString(ProfilingLevel)}"`
+        )
+      );
     }
 
+    // TODO(NODE-3402): Determine error to put here
```
dariakp, line 61: follow up ticket?

src/operations/validate_collection.ts
```diff
@@ -40,6 +40,8 @@ export class ValidateCollectionOperation extends CommandOperation<Document> {
     super.executeCommand(server, session, this.command, (err, doc) => {
       if (err != null) return callback(err);
 
+      // TODO(NODE-3402): Decide on errors to put here
+      // Would this justify a MongoValidationError?
```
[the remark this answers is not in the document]
dariakp: Unresolved TODO?

src/sdam/server.ts
```diff
@@ -260,15 +262,15 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
     }
 
     if (ns.db == null || typeof ns === 'string') {
-      throw new MongoDriverError('ns must not be a string');
+      throw new MongoInvalidArgumentError('Namespace must not be a string');
     }
 
     if (this.s.state === STATE_CLOSING || this.s.state === STATE_CLOSED) {
-      callback(new MongoDriverError('server is closed'));
+      callback(new MongoInvalidArgumentError('Server is closed'));
```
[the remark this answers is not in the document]
dariakp: We really don't want to switch these to a specific class that the next class won't extend from; leave it as MongoDriverError if it's not ready to have the correct class applied

src/sdam/srv_polling.ts
```diff
@@ -67,7 +67,7 @@ export class SrvPoller extends TypedEventEmitter<SrvPollerEvents> {
     super();
 
     if (!options || !options.srvHost) {
-      throw new MongoDriverError('options for SrvPoller must exist and include srvHost');
+      throw new MongoInvalidArgumentError('Options for SrvPoller must exist and include srvHost');
```
dariakp, line 70: I think there is a different place where we validate the srvHost option from the user; I don't think they directly call SrvPoller so this would be a runtime error

src/sdam/topology.ts
```diff
@@ -274,6 +274,7 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
       } else if (seed instanceof HostAddress) {
         seedlist.push(seed);
       } else {
+        // FIXME(NODE-3402): May need to be a MongoParseError
```
dariakp, line 277: unresolved TODO

src/sdam/topology_description.ts
```diff
@@ -431,7 +431,12 @@ function updateRsWithPrimaryFromMember(
   setName?: string
 ): TopologyType {
   if (setName == null) {
-    throw new MongoDriverError('setName is required');
+    // FIXME(NODE-3402): Consider replacing with MongoCompatibilityError or making
```
dariakp, line 434: unresolved TODO

src/sdam/topology_description.ts
```diff
@@ -431,7 +431,12 @@ function updateRsWithPrimaryFromMember(
...
+    // error message more verbose to indicate that the incompatibility is occurring
+    // with the topology rather than the current server version
+    throw new MongoCompatibilityError(
```
dariakp, line 437: I think this is a runtime error, this means we failed to acquire or pass in the setName, it's not user configuration specific

src/sessions.ts
```diff
@@ -125,19 +127,19 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
     super();
 
     if (topology == null) {
-      throw new MongoDriverError('ClientSession requires a topology');
+      throw new MongoInvalidArgumentError('ClientSession requires a topology');
```
dariakp, line 130: this line and the one below reference internal classes; it's not meant to be instantiated directly and I don't think the topology and server session pool are things that can be directly controlled by the user, so it's probably better typed as a runtime error

src/utils.ts
```diff
@@ -248,6 +254,7 @@ export function executeLegacyOperation(
       const optionsIndex = args.length - 2;
       args[optionsIndex] = Object.assign({}, args[optionsIndex], { session: session });
     } else if (opOptions.session && opOptions.session.hasEnded) {
+      // TODO(3405): Replace this with MongoExpiredSessionError
```
dariakp, line 257:
```suggestion
      // TODO(NODE-3405): Replace this with MongoExpiredSessionError
```

src/utils.ts
dariakp: is this function meant to be user facing? if not, the error should probably be runtime

src/utils.ts
```diff
@@ -289,7 +296,9 @@ export function executeLegacyOperation(
 
   // Return a Promise
   if (args[args.length - 1] != null) {
-    throw new MongoDriverError('final argument to `executeLegacyOperation` must be a callback');
+    throw new MongoInvalidArgumentError(
```
dariakp, line 299: this is an internal use function, so we should use a different error

src/utils.ts
```diff
@@ -228,7 +234,7 @@ export function executeLegacyOperation(
   const Promise = PromiseProvider.get();
 
   if (!Array.isArray(args)) {
-    throw new MongoDriverError('This method requires an array of arguments to apply');
+    throw new MongoInvalidArgumentError('This method requires an array of arguments to apply');
```
dariakp, line 237: should use a different error since this is not user facing

dariakp (changes requested):
I think it would be wortwhile to do a search on the repo for NODE-3402 (or just 3402) and make sure any pending TODOs are resolved or updated to a new ticket before we merge this. The majority of the rest of the comments I have are in relation to only using the MongoAPIError for cases where the input is user controlled

src/bulk/ordered.ts
```diff
@@ -26,7 +26,7 @@ export class OrderedBulkOperation extends BulkOperationBase {
 
     // Throw error if the doc is bigger than the max BSON size
     if (bsonSize >= this.s.maxBsonObjectSize)
-      throw new MongoDriverError(
+      throw new MongoInvalidArgumentError(
```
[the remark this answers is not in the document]
nbbeeken: maxBsonObjectSize is a property from the server, I believe it can be bumped or reduced in config. But regardless the check at this point determines if the JS object passed in would serialize to something that exceeds the maximum so I think InvalidArgument does apply. 

We could also consider making a MongoBSONError right now and thinking about subclasses for that later. Could cover `MongoBSONSerializationError, MongoBSONDeserializationError, MongoBSONSizeError`

src/operations/profiling_level.ts
```diff
@@ -17,15 +17,17 @@ export class ProfilingLevelOperation extends CommandOperation<string> {
...
     super.executeCommand(server, session, { profile: -1 }, (err, doc) => {
       if (err == null && doc.ok === 1) {
         const was = doc.was;
         if (was === 0) return callback(undefined, 'off');
         if (was === 1) return callback(undefined, 'slow_only');
         if (was === 2) return callback(undefined, 'all');
-        return callback(new MongoDriverError('Error: illegal profiling level value ' + was));
+        return callback(new MongoInvalidArgumentError('Illegal profiling level value ' + was));
       } else {
+        // TODO(NODE-3402): Consider adding some kind of MongoRuntimeError here
```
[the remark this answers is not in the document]
W-A-James: With regards to these, I'm going through and replacing the ticket number to refer to the new tickets we have made to cover the missing errors

src/cmap/wire_protocol/compression.ts
```diff
@@ -54,9 +54,7 @@ export function compress(
       break;
     default:
       throw new MongoDriverError(
-        'Attempt to compress message using unknown compressor "' +
-          self.options.agreedCompressor +
-          '".'
+        `Attempt to compress message using unknown compressor "${self.options.agreedCompressor}".`
```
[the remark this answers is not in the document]
W-A-James: Good catch here. I think Andy has already addressed this one in NODE-3404 ([PR link here](https://github.com/mongodb/node-mongodb-native/pull/2912/files?file-filters%5B%5D=.ts))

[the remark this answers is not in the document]
dariakp: In that case, can you revert this change to be the same as in main, so that we avoid merge conflicts?

W-A-James:
@dariakp What I've done is to replace the instances you've mentioned of `MongoInvalidArgumentError` that were in internal functions, reverted them to `MongoDriverError` instances and flagged them to be addressed in NODE-3483, which is the ticket @andymina and I are using to fill out the children of `MongoRuntimeError` that we are building out to fit the cases that we missed in the original design stage

src/operations/execute_operation.ts
```diff
@@ -4,6 +4,8 @@ import {
   isRetryableError,
   MONGODB_ERROR_CODES,
   MongoDriverError,
+  MongoInvalidArgumentError,
```
dariakp, line 7: Is this still used in this file?

dariakp (approved):
LGTM, keep up the great work!

PR #2897, opened 2021-07-08
feat(NODE-3392): enable snapshot reads on secondaries

dariakp:
## Description

A base implementation of snapshot sessions per latest spec (NODE-3392, NODE-3393 and NODE-3394).

NOTE: Some spec tests are skipped due to test runner issues with session setup.


src/sessions.ts
```diff
@@ -88,6 +91,7 @@ class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
   clientOptions?: MongoOptions;
   supports: { causalConsistency: boolean };
   clusterTime?: ClusterTime;
+  snapshotTime?: Timestamp; // TODO: should this be more "private"?
```
nbbeeken, line 94: Use symbol property

src/sessions.ts
```diff
@@ -96,6 +100,8 @@ class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
   transaction: Transaction;
   /** @internal */
   [kServerSession]?: ServerSession;
+  /** @internal */
+  snapshot = false; // TODO: should this be more "private"?
```
nbbeeken, line 104: Use symbol property

src/sessions.ts
```diff
@@ -123,15 +129,23 @@ class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
 
     options = options ?? {};
 
+    if (options.snapshot === true) {
+      this.snapshot = true;
+      if (options.causalConsistency === true) {
+        // TODO: throw validation error; MongoDriverError or more specific
+      }
+
+      // TODO(NODE-3394): also validate server version >= 5.0
```
nbbeeken, line 138: Use topology.capabilities (with a new method) to check this

src/sessions.ts
```diff
@@ -801,28 +820,35 @@ function applySession(
...
-    if (session.supports.causalConsistency && session.operationTime && shouldApplyReadConcern) {
+    if (
+      session.supports.causalConsistency &&
+      session.operationTime &&
+      commandSupportsReadConcern(command, options)
+    ) {
       command.readConcern = command.readConcern || {};
       Object.assign(command.readConcern, { afterClusterTime: session.operationTime });
+    } else if (session.snapshot) {
+      command.readConcern = command.readConcern || { level: ReadConcernLevel.snapshot }; // TODO: is there a better place to set this?
```
nbbeeken, line 841: Should we be using the ReadConcern proper class here? If it doesn't cause failures I think this style of usage is just legacy

src/sessions.ts
```diff
@@ -855,6 +881,10 @@ function updateSessionFromResponse(session: ClientSession, document: Document):
   if (document.recoveryToken && session && session.inTransaction()) {
     session.transaction._recoveryToken = document.recoveryToken;
   }
+
+  if (document.cursor?.atClusterTime && session?.snapshot && session.snapshotTime === undefined) {
+    session.snapshotTime = document.cursor.atClusterTime;
+  }
 }
 
 export {
```
nbbeeken, line 890: If you want to do some legacy clean up, putting the export on each symbol is what we try to do everywhere else in the driver, completely optional.

nbbeeken (changes requested):
Leaving some notes from out meeting:

src/sessions.ts
dariakp: should be implemented

src/sessions.ts
dariakp: implemented

src/sessions.ts
dariakp: I think we should revisit this later because based on what's defined in the ReadConcern class proper, only the level is present, and based on the usage, there is ReadConcern defined for entire entities, like collections, and then there are specific read concerns passed with individual commands (containing things like the afterClusterTime and atClusterTime); from a typescript hinting perspective, it would not be helpful to mix the two since these additional properties can't be defined at a higher level... so we probably want to think more on how we want to design this internally going forward

src/sessions.ts
dariakp: Was easy enough, pushed

src/operations/execute_operation.ts
```diff
@@ -86,6 +86,8 @@ export function executeOperation<
         session = topology.startSession({ owner, explicit: false });
       } else if (session.hasEnded) {
         return cb(new MongoDriverError('Use of expired sessions is not permitted'));
+      } else if (session.snapshotEnabled && !topology.capabilities.supportsSnapshotReads) {
+        return cb(new MongoDriverError('Snapshot reads require MongoDB 5.0 or later'));
```
dariakp, line 90: Do we want to use a more specific error here now that they are available?

src/sessions.ts
```diff
@@ -123,15 +138,23 @@ class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
 
     options = options ?? {};
 
+    if (options.snapshot === true) {
+      this[kSnapshotEnabled] = true;
+      if (options.causalConsistency === true) {
+        throw new MongoDriverError(
```
dariakp, line 144: Do we want to use a more specific error here now that they are available?

Also, this requires a prose test per spec, making a note so I don't forget

src/sessions.ts
```diff
@@ -257,6 +280,10 @@ class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    * @param options - Options for the transaction
    */
   startTransaction(options?: TransactionOptions): void {
+    if (this[kSnapshotEnabled]) {
+      throw new MongoDriverError('Transactions are not allowed with snapshot sessions');
```
dariakp, line 284: Do we want to use a more specific error here now that they are available?

src/sessions.ts
```diff
@@ -96,6 +103,14 @@ class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
   transaction: Transaction;
   /** @internal */
   [kServerSession]?: ServerSession;
+  /** @internal */
+  [kSnapshotTime]?: Timestamp;
+  /** @internal */
+  [kSnapshotEnabled] = false;
+
+  get snapshotEnabled(): boolean {
```
nbbeeken, line 111: suggestion: `isSnapshotEnabled`?
Can you move it be next to the other getters?
I'm in favor of the helper though regardless of these two asks

src/operations/execute_operation.ts
nbbeeken: Compat and InvalidArgument are still not available so I think leaving the DriverError in there would make sense for now.

src/sessions.ts
dariakp: Moved the getter (as discussed, leaving name unchanged so that the verb doesn't mislead anyone into thinking this is a function call)

nbbeeken (approved):

emadum (approved):

PR #2901, opened 2021-07-13
test: new test runner changes from lb support

durran:


test/functional/unified-spec-runner/entities.ts
```diff
@@ -38,52 +62,105 @@ function getClient(address) {
...
+    this.cmapEvents = [];
     this.failPoints = [];
     this.ignoredEvents = [
       ...(description.ignoreCommandMonitoringEvents ?? []),
       'configureFailPoint'
     ];
-    // apm
-    this.observedEvents = (description.observeEvents ?? []).map(
-      e => UnifiedMongoClient.EVENT_NAME_LOOKUP[e]
+    this.observedCommandEvents = (description.observeEvents ?? []).map(
```
dariakp, line 118: should these two observed event arrays be filtered so that only the defined ones get listeners since they both derive from `description.observeEvents`?

test/functional/unified-spec-runner/entities.ts
```diff
@@ -222,7 +317,7 @@ export class EntitiesMap<E = Entity> extends Map<string, E> {
     for (const entity of entities ?? []) {
       if ('client' in entity) {
         const useMultipleMongoses =
-          config.topologyType === 'Sharded' && entity.client.useMultipleMongoses;
+          config.topologyType === 'LoadBalanced' && entity.client.useMultipleMongoses;
```
dariakp, line 320: why is this no longer desired for the sharded topologyType?

test/functional/unified-spec-runner/schema.ts
```diff
@@ -118,7 +130,41 @@ export interface Test {
 }
 export interface ExpectedEventsForClient {
   client: string;
-  events: ExpectedEvent[];
+  eventType?: string;
+  events: (ExpectedCommandEvent | ExpectedCmapEvent)[];
+}
+export interface ExpectedCommandEvent {
```
dariakp, line 136: since this is a duplicate of the original ExpectedEvent interface, should `ExpectedEvent` be deleted below? Or is it still used somewhere?

test/functional/unified-spec-runner/match.ts
```diff
@@ -235,32 +248,70 @@ export function specialCheck(
...
+  connectionCheckOutFailedEvent: ConnectionCheckOutFailedEvent,
+  connectionCheckedOutEvent: ConnectionCheckedOutEvent,
+  connectionCheckedInEvent: ConnectionCheckedInEvent
+};
+
+function validEmptyCmapEvent(
+  expected: ExpectedCommandEvent | ExpectedCmapEvent,
+  actual: CommandEvent | CmapEvent
+) {
+  return Object.keys(EMPTY_CMAP_EVENTS).some(key => {
```
dariakp, line 267: why not just iterate over `Object.values`?

test/functional/unified-spec-runner/match.ts
```diff
@@ -235,32 +248,70 @@ export function specialCheck(
...
-  expected: ExpectedEvent[],
-  actual: CommandEvent[],
+  expected: (ExpectedCommandEvent & ExpectedCmapEvent)[],
+  actual: (CommandEvent & CmapEvent)[],
   entities: EntitiesMap
 ): void {
-  // TODO: NodeJS Driver has extra events
-  // expect(actual).to.have.lengthOf(expected.length);
+  if (actual.length !== expected.length) {
+    // const actualNames = actual.map(a => a.constructor.name);
```
dariakp, line 279: this seems to be an empty code block

test/functional/unified-spec-runner/match.ts
```diff
@@ -235,32 +248,70 @@ export function specialCheck(
...
+    } else if (expectedEvent.commandSucceededEvent) {
+      expect(actualEvent).to.be.instanceOf(CommandSucceededEvent);
       resultCheck(actualEvent, expectedEvent.commandSucceededEvent, entities, [
         `events[${index}].commandSucceededEvent`
       ]);
-    } else if (expectedEvent.commandFailedEvent && actualEvent instanceof CommandFailedEvent) {
+    } else if (expectedEvent.commandFailedEvent) {
+      expect(actualEvent).to.be.instanceOf(CommandFailedEvent);
       expect(actualEvent.commandName).to.equal(expectedEvent.commandFailedEvent.commandName);
+    } else if (validEmptyCmapEvent(expectedEvent, actualEvent)) {
```
dariakp, line 301: I would suggest moving this into L313 instead, as the final clause `else if (!validEmptyCmapEvent...) { expect.fail ... }`, that would avoid having an odd empty block in the middle of otherwise uniform checks

test/functional/unified-spec-runner/operations.ts
```diff
@@ -294,19 +328,36 @@ operations.set('insertMany', async ({ entities, operation }) => {
...
+    ]);
+  } catch (e) {
+    const findCursor = entities.getEntity('cursor', operation.object);
+    return await findCursor.next();
+  }
+});
+
+operations.set('listCollections', async ({ entities, operation }) => {
+  const db = entities.getEntity('db', operation.object);
+  const { filter, batchSize } = operation.arguments;
```
dariakp, line 346: are we intentionally ignoring potential other options? I am concerned that this approach to implementing commands (here and at `listIndexes` in L355) might cause incorrect tests sneak in when we sync specs that introduce new arguments, I feel like a better approach would be to do something like `{ filter, ...opts } = operation.arguments` and then pass them in as `(filter, opts)`, with any future unsupported parameters explicitly omitted (e.g., `{ filter, unsupportedOpt, ...opts } = operation.arguments`)

dariakp (changes requested):
Just a few outstanding questions/concerns here

test/functional/unified-spec-runner/match.ts
nbbeeken: Relevant convo over on the bigger PR: https://github.com/mongodb/node-mongodb-native/pull/2742#discussion_r668296050

dariakp: @nbbeeken Thanks! still think that we should maybe remove the if block and the code that attempts to do things and add a comment in the code summarizing this: 

> It was intentional as the Node driver does not execute a killCursors after a getMore has errored, which causes the absence of the killCursors events in some cases. I could not find anywhere so far in the specs that states an errored getMore requires an immediate killCursors afterwards, and all the driver does in this case is cleanup the cursor (but not explicitly close() it).

(maybe even file a ticket to explore this further, since if we know exactly what's out of the ordinary, we may be able to filter explicitly those cases out)

test/functional/unified-spec-runner/operations.ts
durran: Updated to pass the options through.

test/functional/unified-spec-runner/match.ts
durran: This check is now back and asserting the event lengths, so no comment needed.

test/functional/unified-spec-runner/match.ts
durran: Done.

test/functional/unified-spec-runner/schema.ts
durran: `ExpectedEvent` now removed.

test/functional/unified-spec-runner/entities.ts
durran: Was a mistake - both sharded and load balanced are now checked here.

test/functional/unified-spec-runner/entities.ts
durran: These are now filtered so no `undefined` elements are in the arrays.

test/functional/unified-spec-runner/match.ts
dariakp: It looks like some of the new snapshot tests are failing now with that check, I don't think you need to take care of it here because I need to sort out those tests anyway, but if you could add them to the skip list, that would allow the pipeline to stay green on main

test/functional/unified-spec-runner/match.ts
dariakp: what do you think is the benefit of having it as a separate block as:
```
    } else if (validEmptyCmapEvent(expectedEvent, actualEvent)) {
      // This should just always pass since the event must exist and match the type.
    } else {
      expect.fail(`Events must be one of the known types, got ${inspect(actualEvent)}`);
    }
```
vs just
```
    } else if (!validEmptyCmapEvent(expectedEvent, actualEvent)) {
      expect.fail(`Events must be one of the known types, got ${inspect(actualEvent)}`);
    }
    // remove final else
```

test/functional/unified-spec-runner/match.ts
dariakp: snapshot-sessions-unsupported-ops
- Server returns an error on findOneAndUpdate with snapshot
- Server returns an error on deleteOne with snapshot
- Server returns an error on updateOne with snapshot

test/functional/unified-spec-runner/match.ts
durran: Updated - nice catch.

test/functional/unified-spec-runner/match.ts
durran: I've skipped the one failing test now.

dariakp (approved):
Thank you very much!

nbbeeken (approved):
LGTM

emadum (approved):
LGTM 👍

PR #2903, opened 2021-07-14
fix(NODE-3199): unable to bundle driver due to uncaught require

nbbeeken:
Tested locally webpack only returns warnings with this change


emadum (approved):
LGTM! I think we discussed adding comments to any requires in try/catch blocks to note the necessity of it for bundling, should we do that in this PR?

nbbeeken:
Oh yes, thanks for the reminder I'll do that now, plus gotta do this rebase

dariakp (approved):
LGTM

PR #2905, opened 2021-07-14
test(NODE-3072): serverless testing 

emadum:
Adds testing against "proxy as mongos"


.evergreen/run-serverless-tests.sh
```diff
@@ -0,0 +1,13 @@
...
+SERVERLESS_ATLAS_USER="${SERVERLESS_ATLAS_USER}" \
+SERVERLESS_ATLAS_PASSWORD="${SERVERLESS_ATLAS_PASSWORD}" \
+SERVERLESS=1 AUTH=auth SSL=ssl \
+MONGODB_URI=${MONGODB_URI} npx mocha \
+  test/functional/crud_spec.test.js \
+  test/functional/retryable_reads.test.js \
+  test/functional/retryable_writes.test.js \
+  test/functional/sessions.test.js \
+  test/functional/transactions.test.js \
+  test/functional/versioned-api.test.js
```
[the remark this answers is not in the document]
emadum: Yes this list comes from the spec:

```
Tests defined in the following specifications MUST be included in a driver's
Atlas Serverless testing suite, including prose tests:

- CRUD, including the v1 and unified tests
- Retryable Writes
- Retryable Reads
- Versioned API
- Driver Sessions
- Transactions (excluding convenient API)
```

test/functional/unified-spec-runner/entities.ts
```diff
@@ -106,7 +101,8 @@ export class UnifiedMongoClient extends MongoClient {
     super(url, {
       monitorCommands: true,
       ...description.uriOptions,
-      serverApi: description.serverApi ? description.serverApi : serverApiConfig()
+      ...getEnvironmentalOptions(),
+      ...(description.serverApi ? { serverApi: description.serverApi } : {})
```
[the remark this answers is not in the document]
emadum: That just provides default options based on the testing configuration; if there are more specific options in the spec test's descriptions for `serverApi`, they need to take precedence over the environmental options.

test/functional/unified-spec-runner/entities.ts
```diff
@@ -51,15 +53,8 @@ export type CmapEvent =
...
-function serverApiConfig() {
-  if (process.env.MONGODB_API_VERSION) {
-    return { version: process.env.MONGODB_API_VERSION };
-  }
-}
-
 function getClient(address) {
-  const serverApi = serverApiConfig();
-  return new MongoClient(`mongodb://${address}`, serverApi ? { serverApi } : {});
+  return new MongoClient(`mongodb://${address}`, getEnvironmentalOptions() as MongoClientOptions);
```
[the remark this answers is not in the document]
emadum: The helper is defined in a js file so the return type can't go there unfortunately. I remember needing to put that type assertion because I was getting a ts error, but I just tried removing it and it's now compiling with no complaints. :shrug:

test/functional/sessions.test.js
```diff
@@ -66,7 +66,12 @@ describe('Sessions', function () {
     });
 
     describe('withSession', {
-      metadata: { requires: { mongodb: '>=3.6.0' } },
+      metadata: {
+        requires: {
+          mongodb: '>=3.6.0',
+          serverless: 'forbid' // UnhandledPromiseRejectionWarning: MongoServerError: CMD_NOT_ALLOWED: dropAllUsersFromDatabase
```
dariakp, line 72: could you explain what this comment is referencing?

test/functional/unified-spec-runner/entities.ts
```diff
@@ -116,6 +111,10 @@ export class UnifiedMongoClient extends MongoClient {
       ...(description.ignoreCommandMonitoringEvents ?? []),
       'configureFailPoint'
     ];
+    // FIXME: hack to get tests passing, extra unexpected events otherwise
```
dariakp, line 114: can we add a jira ticket for this?

test/tools/runner/config.js
```diff
@@ -104,13 +109,13 @@ class TestConfiguration {
...
-      Object.assign(defaultOptions, { serverApi: this.serverApi });
-    }
+    const defaultOptions = Object.assign(
+      { minHeartbeatFrequencyMS: 100 },
+      getEnvironmentalOptions()
+    );
     // support MongoClient constructor form (url, options) for `newClient`
     if (typeof dbOptions === 'string') {
-      return new MongoClient(dbOptions, Object.assign(defaultOptions, serverOptions));
+      return new MongoClient(dbOptions, Object.assign({}, defaultOptions, serverOptions));
```
dariakp, line 118: since the defaultOptions are generated in this test, why generate a new object instead of just passing in the updated defaultOptions as before?

test/tools/runner/config.js
```diff
@@ -251,6 +256,11 @@ class TestConfiguration {
       }
     }
 
+    if (this.isServerless) {
```
dariakp, line 259: what happens if the authSource is already set by the test? (e.g., in L236) Do we still want to override? Should this have a check whether ssl/authSource are already set and only update them in that case?

dariakp (changes requested):

test/functional/sessions.test.js
emadum: The command `dropAllUsers`used by that test isn't allowed on serverless, so the test needs to be skipped. Do you think the comment needs to be more descriptive? E.g. "Skipped due to error on serverless: ..."

test/tools/runner/config.js
emadum: None of the tests required for serverless deal with auth in any way, so this should be fine for this iteration of the testing - that said, it couldn't hurt to future-proof the code now rather than dealing it with if/when such tests are added to the serverless suite. We can't really validate it works though, unless such tests are added; do you think it's worth doing now?

test/tools/runner/config.js
emadum: Good point, I'll clean this up 👍

test/functional/unified-spec-runner/entities.ts
emadum: Yep I'm just first going to confirm this is still an issue with the LB work merged in, I know there were a number of changes made to the unified runner there and I put this hack in before LB was done.

test/functional/sessions.test.js
dariakp: is it the test.setup that does that? `dropAllUsersFromDatabase` doesn't seem essential to sessions (though I may be missing something here), in which case, instead of skipping these tests, it might be better to just refactor the setup to make it compatible with serverless? if that sounds accurate but isn't a quick fix, I think just logging a follow jira ticket to re-enable them later would be fine

test/tools/runner/config.js
dariakp: would it make sense to put this block as an `else if` to the one in L224? that way, if there are tests added that have the auth props set and they end up not being compatible with serverless, the test will fail and we'll be able to make an informed decision how to proceed at that point

test/tools/runner/config.js
emadum: Refactored, PTAL 🙏

test/functional/unified-spec-runner/entities.ts
emadum: Confirmed it's still an issue and filed NODE-3549

test/functional/sessions.test.js
emadum: Took a second look at this, the test actually passes and doesn't need to be skipped.

dariakp (approved):

PR #2907, opened 2021-07-15
test(NODE-3082): details property from write errors is accessible

nbbeeken:
The new details property in 5.0 is already accesible we should just test to ensure it stays that way.

### DIFF
--- a/test/functional/write_concern.test.js
+++ b/test/functional/write_concern.test.js
@@ -8,7 +8,7 @@ const loadSpecTests = require('../spec').loadSpecTests;
 const { withMonitoredClient } = require('./shared');
 
 const mock = require('../tools/mock');
-const { MongoClient, MongoServerError } = require('../../src');
+const { MongoClient } = require('../../src');
 
 describe('Write Concern', function () {
   describe('spec tests', function () {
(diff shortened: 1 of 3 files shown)


test/functional/write_concern.test.js
```diff
@@ -117,4 +117,56 @@ describe('Write Concern', function () {
...
+        const evCapture = require('events').once(client, 'commandSucceeded');
+
+        let insertError;
+        try {
+          await collection.insertOne({ x: /not a string/ });
+          expect.fail('The insert should fail the validation that x must be a string');
+        } catch (error) {
+          expect(error).to.be.instanceOf(MongoServerError);
+          expect(error).to.have.property('code', 121);
+          expect(error).to.have.nested.property('errInfo.details').that.is.an('object');
```
dariakp, line 155: the way I am reading the spec, it says that the `errInfo` property is what should be exposed as `error.details` in the driver; what we currently do is expose `errInfo` as `error.errInfo`, it also says we shouldn't try to parse the inside of `errInfo` so accessing `errInfo.details` probably should not be in the test

test/functional/write_concern.test.js
```diff
@@ -117,4 +117,56 @@ describe('Write Concern', function () {
...
+          await client.db().collection('wc_details').drop();
+        } catch {
+          // don't care
+        }
+
+        const collection = await client
+          .db()
+          .createCollection('wc_details', { validator: { x: { $type: 'string' } } });
+
+        const evCapture = require('events').once(client, 'commandSucceeded');
```
dariakp, line 146: since node will cache any requires, I don't think hiding a require inside a test instead of declaring it up top is a good idea; we should also be careful to reset event listeners in an after block in case it isn't triggered in the test

dariakp (changes requested):

test/functional/write_concern.test.js
nbbeeken: Did a bit of hunting on this one: 

- [Looks like we need to leave errInfo as is](https://jira.mongodb.org/browse/DRIVERS-820?focusedCommentId=3604057&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3604057)
- [The spec permits us to have errInfo on generic Server write failures](https://github.com/mongodb/specifications/blob/master/source/server_write_commands.rst#error-document)
- [Following PHPs lead here](https://github.com/mongodb/mongo-php-driver/pull/1229/files#diff-f5695d70f8e6c3e6adedcd927165379ee9dbf1be9284dd219473db5b077a2eb8R43) I agree with removing the parsing of errInfo I'll push that up in the next commit

package.json
```diff
@@ -54,6 +54,7 @@
     "@types/node": "^15.3.1",
     "@types/saslprep": "^1.0.0",
     "@types/semver": "^7.3.4",
+    "@types/whatwg-url": "^8.2.1",
```
dariakp, line 57: what's this one for?

test/unit/core/write_concern_error.test.js
```diff
@@ -159,4 +161,51 @@ describe('WriteConcernError', function () {
...
+        } catch (error) {
+          expect(error).to.be.instanceOf(MongoServerError);
+          expect(error).to.have.property('code', 121);
+          expect(error).to.have.property('errInfo').that.is.an('object');
+        }
+
+        const commandSucceededEvents = await evCapture;
+        expect(commandSucceededEvents).to.have.lengthOf(1);
+        const ev = commandSucceededEvents[0];
+        expect(ev).to.have.nested.property('reply.writeErrors[0].errInfo').that.is.an('object');
```
dariakp, line 207: do we want to check to make sure that this errInfo is deep equal to the caught error's errInfo above?

test/functional/write_concern.test.js
dariakp: Thanks for the thorough check!

test/unit/core/write_concern_error.test.js
nbbeeken: Oo yes that would be good

package.json
nbbeeken: It was to fix this: https://github.com/mongodb-js/mongodb-connection-string-url/issues/2
But its been merged in one of my other PRs so I'll rebase :)

dariakp (approved):
LGTM!

emadum (approved):

PR #2908, opened 2021-07-15
fix(NODE-3393): snapshot time not applied if distinct executed first

dariakp:
## Description

NODE-3393, NODE-3394

This PR started out as a fix to our spec runner which was selectively passing in sessions instead of applying them to all operations. Fixing the last test revealed a bug in the original snapshot session reads implementation: the `atClusterTime` from the server response was not applied to the session if the first operation executed against it was `distinct`. This PR includes the fix for that issue.


nbbeeken (approved):
LGTM

emadum (approved):
LGTM 👍

PR #2909, opened 2021-07-15
NODE-3011: Load Balancer Support

durran:


src/cmap/connect.ts
```diff
@@ -129,6 +131,25 @@ function performInitialHandshake(
         return;
       }
 
+      if (options.loadBalanced) {
+        // TODO: Durran: Remove when server support exists. (NODE-3431)
+        if (FAKE_MONGODB_SERVICE_ID) {
+          if (response.topologyVersion) {
```
dariakp, line 137: this might look prettier as a ternary:
`response.serviceId = response.topologyVersion ? response.topologyVersion.processId : new ObjectId();`

for my own understanding, could you explain why we fall back to `new ObjectId()` specifically if `topologyVersion` is not returned?

src/cmap/connection.ts
```diff
@@ -416,35 +421,10 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
-            session.transaction.unpinServer();
-          }
-
-          return callback(err, ...args);
-        }
-      : callback;
-
     try {
-      write(this, message, commandOptions, commandResponseHandler);
+      write(this, message, commandOptions, callback);
```
dariakp, line 425: where did we land with the TransientTransactionError label? Is it just no longer required?

src/cmap/connection.ts
```diff
@@ -680,8 +660,10 @@ export class CryptoConnection extends Connection {
   }
 }
 
-function hasSessionSupport(conn: Connection) {
-  return conn.description.logicalSessionTimeoutMinutes != null;
+/** @public */
```
dariakp, line 663: A general question on this file... it looks like this code is non-breaking apart from the unpin server logic and possibly the transient error thing - would it be possible to pull out just the changes here and related files into a separate PR? That way we could ensure that existing functionality, is, in fact, not broken with the new pinning logic? (It's fine if the logic is too intertwined in the other files and the answer is no, I am just trying to think of how to make applying these changes a little safer and easier to review)

src/cmap/errors.ts
```diff
@@ -28,9 +27,9 @@ export class WaitQueueTimeoutError extends MongoDriverError {
   /** The address of the connection pool */
   address: string;
 
-  constructor(pool: Connection | ConnectionPool) {
-    super('Timed out while checking out a connection from connection pool');
-    this.address = pool.address;
+  constructor(message: string, address: string) {
+    super(`Timed out while checking out a connection from connection pool${message}`);
```
dariakp, line 31: I think from a human reading perspective, it would make more sense for the metrics function to return the metrics formatted as a standalone sentence, and for this function to assume that the message will be a standalone sentence... just because it feels a little odd to have a constructor that takes a seemingly arbitrary message assume that it can always be appended directly to `pool`.

src/cmap/metrics.ts
```diff
@@ -0,0 +1,56 @@
...
+      this.otherConnections -= 1;
+    }
+  }
+
+  /**
+   * Return information about the cmap metrics as a string.
+   */
+  info(): string {
+    return (
+      `connections in use by cursors: ${this.cursorConnections},` +
```
dariakp, line 51: for readability, probably want to add a space after the comma here and in the line below

test/tools/runner/index.js
```diff
@@ -66,6 +66,7 @@ before(function (_done) {
   // );
 
   const options = MONGODB_API_VERSION ? { serverApi: MONGODB_API_VERSION } : {};
+
```
dariakp, line 69: this is just a whitespace change, if we get rid of it, we have one less file in the PR

dariakp (commented):
I got through almost half of the PR, going to take a fresh look at the rest of it Monday; meanwhile, some thoughts/questions

src/cmap/connection.ts
mbroadst: Hm, thanks for calling this out. I made a perhaps debatable decision to always make local network/timeout errors include this error label [here](https://github.com/mongodb/node-mongodb-native/blob/NODE-3011/load-balancer-support/src/error.ts#L562-L563). The transactions spec [states](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#transienttransactionerror): 

> In the case of command errors, the server adds the label; in the case of network errors or server selection errors where the client receives no server reply, the client adds the label.

I believe this used to be specific to the context of being in a transaction, but may be more important now for load balancers (and thus applied universally). A quick check in the dev chat did not elucidate much concern about including it always, except that libmongoc does indeed only apply it during transactions. It also didn't break any tests :shrug: Perhaps we should make a NODE ticket for it to track in the future, or even better a DRIVERS ticket to clarify if that language is missing an "when encountered in an active transaction" clause

dariakp: @mbroadst Looking over the transactions spec, it does appear to [define the transient transaction error](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#id16) as 

> Any command error that includes the "TransientTransactionError" error label in the "errorLabels" field. Any network error encountered running any command other than commitTransaction **in a transaction**. 

and it has the same "in a transaction" language in the sentence immediately preceding the one you quoted above:

> Any network error or server selection error encountered running any command besides commitTransaction **in a transaction**.

so to me at least it doesn't read like we should be adding that logic to all network errors in general outside of transactions, though I'm not sure I fully understand the implications for load balancers here

test/unit/mongo_client_options.test.js
```diff
@@ -450,4 +451,32 @@ describe('MongoOptions', function () {
       );
     });
   });
+
+  context('when loadBalanced=true is in the URI', function () {
```
dariakp, line 455: should there be analogous tests for when it's passed through the options object?

test/unit/cmap/stream_description.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+        it('sets the property', function () {
+          expect(description.logicalSessionTimeoutMinutes).to.eq(5);
+        });
+      });
+
+      context('when logicalSessionTimeoutMinutes is not in the options', function () {
+        const description = new StreamDescription('a:27017', {});
+
+        it('sets logicalSessionTimeoutMinutes to undefined', function () {
```
dariakp, line 21: do we want the behavior to be that it sets a property to undefined or that it does not define the property? as written, it sounds like it should be the former, in which case the corresponding assertion is `expect(description).to.have.property('logical...', undefined)`

test/unit/cmap/stream_description.test.js
```diff
@@ -0,0 +1,56 @@
...
+          expect(description.logicalSessionTimeoutMinutes).to.not.exist;
+        });
+      });
+
+      context('when loadBalanced is in the options', function () {
```
dariakp, line 26: should this be "and set to true"? and then should there be another test for "and set to false"?

test/unit/cmap/stream_description.test.js
```diff
@@ -0,0 +1,56 @@
...
+        it('sets loadBalanced to false', function () {
+          expect(description.loadBalanced).to.be.false;
+        });
+      });
+    });
+
+    context('when options are not provided', function () {
+      const description = new StreamDescription('a:27017');
+
+      it('defaults logicalSessionTimeoutMinutes to undefined', function () {
```
dariakp, line 47: same comment here regarding using the `property` assertion when asserting that the key is present but set to undefined (vs the current assertion which does not differentiate between that scenario and the key not being present at all)

dariakp (commented):
Some more questions/comments, still not quite finished with the PR because I want to take a very close look at connection pool and sessions changes

src/transactions.ts
```diff
@@ -133,9 +132,19 @@ export class Transaction {
       TxnState.STARTING_TRANSACTION,
       TxnState.TRANSACTION_IN_PROGRESS
     ];
+
     return activeStates.includes(this.state);
   }
 
+  get isCommitted(): boolean {
+    const committedStates: TxnState[] = [
```
emadum, line 140: Can we move this definition to the top level of the file?

test/functional/spec-runner/index.js
```diff
@@ -193,10 +193,15 @@ function prepareDatabaseForSuite(suite, context) {
...
-      if (err.message.match(/no such (cmd|command)/) || err.code === 11601) {
+      if (
+        err.message.match(/no such (cmd|command)/) ||
+        err.message.match(/Failed to kill on some hosts/) ||
+        err.code === 11601
+      ) {
         return;
       }
 
+      console.log('err', err);
```
emadum, line 204: Is this left over from debugging?

emadum (changes requested):
Did a full read through and left a few nit comments - I'll follow up with a final pass once some of the active comments on the PR have been resolved. Very nice work here! 👍

package.json
```diff
@@ -119,6 +119,7 @@
     "check:ts": "tsc -v && tsc --noEmit",
     "check:atlas": "mocha --config \"test/manual/mocharc.json\" test/manual/atlas_connectivity.test.js",
     "check:adl": "mocha test/manual/data_lake.test.js",
+    "check:load-balancer": "mocha test/manual/load-balancer.test.js",
```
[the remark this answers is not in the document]
durran: I've created https://jira.mongodb.org/browse/NODE-3461

src/operations/execute_operation.ts
```diff
@@ -241,8 +250,9 @@ function shouldRetryWrite(err: any) {
 
 function supportsRetryableWrites(server: Server) {
   return (
-    server.description.maxWireVersion >= 6 &&
-    server.description.logicalSessionTimeoutMinutes &&
-    server.description.type !== ServerType.Standalone
+    server.loadBalanced ||
```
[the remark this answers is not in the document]
durran: `maxWireVersion` and `logicalSessionTimeoutMinutes` would both not exist in this case in the server description since the monitor is not running.

src/utils.ts
```diff
@@ -663,6 +664,9 @@ export function uuidV4(): Buffer {
  */
 export function maxWireVersion(topologyOrServer?: Connection | Topology | Server): number {
   if (topologyOrServer) {
+    if (topologyOrServer.loadBalanced) {
+      return MAX_SUPPORTED_WIRE_VERSION;
```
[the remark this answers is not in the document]
durran: We are assuming that load balancer is running against the latest server for the time being. If the server starts backporting support, which it might, then this will need to change.

test/manual/load-balancer.test.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
```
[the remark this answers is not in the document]
durran: To remove in https://jira.mongodb.org/browse/NODE-3461

src/cmap/connect.ts
```diff
@@ -129,6 +131,25 @@ function performInitialHandshake(
         return;
       }
 
+      if (options.loadBalanced) {
+        // TODO: Durran: Remove when server support exists. (NODE-3431)
```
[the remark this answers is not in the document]
durran: I updated the ticket with a link to the PM ticket, since a SERVER ticket does not exist yet.

test/unit/cmap/stream_description.test.js
durran: Updated to suggestion.

test/unit/cmap/stream_description.test.js
durran: Improved tests to test for both true and false values.

test/unit/cmap/stream_description.test.js
durran: Updated to suggestion.

test/unit/mongo_client_options.test.js
durran: Actually it's only allowed in the URI, not the options object, so I've added tests for those cases and fixed the behaviour to throw when provided in the options object.

test/unit/cmap/metrics.test.js
```diff
@@ -0,0 +1,102 @@
...
+      );
+    });
+  });
+
+  describe('#markPinned', function () {
+    const metrics = new ConnectionPoolMetrics();
+
+    context('when the type is TXN', function () {
+      before(function () {
+        metrics.markPinned(ConnectionPoolMetrics.TXN);
```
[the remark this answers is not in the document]
durran: Updated the tests to look at the entire object properties as suggested, refactored with a `reset()` method on the object that can be used both in the constructor to default the values and in the test suite to reset it back to the default.

test/unit/cmap/connection_pool.test.js
```diff
@@ -243,13 +252,77 @@ describe('Connection Pool', function () {
...
 
-      pool.withConnection((err, conn, cb) => {
+      pool.withConnection(undefined, (err, conn, cb) => {
         expect(err).to.not.exist;
         cb();
       });
     });
   });
 
+  describe.skip('#closeConnections', function () {
```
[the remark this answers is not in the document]
durran: I've removed these as the spec tests cover these cases of closing per the service id.

test/unit/cmap/metrics.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+  describe('#markUnpinned', function () {
+    const metrics = new ConnectionPoolMetrics();
+
+    context('when the type is TXN', function () {
+      before(function () {
+        metrics.markUnpinned(ConnectionPoolMetrics.TXN);
+      });
+
+      it('increments the txnConnections count', function () {
```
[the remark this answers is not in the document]
durran: Updated the typo.

test/unit/cmap/metrics.test.js
```diff
@@ -0,0 +1,102 @@
...
+        expect(metrics.txnConnections).to.equal(-1);
+      });
+    });
+
+    context('when the type is CURSOR', function () {
+      before(function () {
+        metrics.markUnpinned(ConnectionPoolMetrics.CURSOR);
+      });
+
+      it('increments the cursorConnections count', function () {
```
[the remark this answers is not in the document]
durran: Updated the typo.

test/unit/cmap/metrics.test.js
```diff
@@ -0,0 +1,102 @@
...
+        expect(metrics.cursorConnections).to.equal(-1);
+      });
+    });
+
+    context('when the type is OTHER', function () {
+      before(function () {
+        metrics.markUnpinned(ConnectionPoolMetrics.OTHER);
+      });
+
+      it('increments the otherConnections count', function () {
```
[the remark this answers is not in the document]
durran: Updated the typo.

test/unit/cmap/connection.test.js
```diff
@@ -106,4 +107,49 @@ describe('Connection - unit/cmap', function () {
       done();
     });
   });
+
+  describe('.hasSessionSupport', function () {
+    let connection;
+    const stream = new Socket();
+
+    context('when logicalSessionTimeoutMinutes is preset', function () {
```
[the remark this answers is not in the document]
durran: Updated the typo.

src/cursor/abstract_cursor.ts
```diff
@@ -757,18 +719,71 @@ function cursorIsDead(cursor: AbstractCursor): boolean {
...
+  cursor: AbstractCursor,
+  options: { error?: AnyError | undefined; needsToEmitClosed?: boolean } | undefined,
+  callback: Callback
+): void {
+  const cursorId = cursor[kId];
+  const cursorNs = cursor[kNamespace];
+  const server = cursor[kServer];
+  const session = cursor[kSession];
+  const error = options?.error;
+  const needsToEmitClosed =
```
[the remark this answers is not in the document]
durran: Updated to suggestion.

test/tools/runner/index.js
durran: Whitespace removed.

src/cmap/metrics.ts
durran: Formatting updated with the spaces.

src/cmap/errors.ts
durran: Moved the full error message into the metrics object itself, and the complete message is always passed in and used as is.

src/transactions.ts
durran: Are you actually asking that all the getters defined are in alphabetical order? It's sitting after the `isActive` getter so I assumed that would be a natural place for it.

test/functional/spec-runner/index.js
durran: Yes that was left over. Removed now.

src/transactions.ts
emadum: Sorry for being unclear, I meant moving the `committedStates` constant out of the getter, it seems like something that could go at the top of the file as an UPPER_CASE const. Total nit, feel free to disregard as well if you think this makes it more readable.

src/cmap/connection.ts
durran: I think personally I'm happy when the existing tests are running and passing (most notably the transactions spec tests) that the changes here have not broken the existing functionality. Since right now at the time of writing there are only 2 failing tests related to change streams in the main suite I'm pretty confident this is fine.

src/cmap/connection.ts
dariakp: My concern is that the current implementation adds the label to all network/timeout errors, regardless of load balancer mode, so even if we say the load balancer should always behave as being in a transaction of sorts, the same isn't true of non-load balancer mode. As you said, it is at best misleading and at worst breaking in some edge case that we don't test, though if we think the risk of the latter is low enough, it may be something we can live with, depending on how difficult it would be to implement a more correct solution.

test/unit/mongo_client_options.test.js
dariakp: Hmm, how is the test in L146 still passing given that we're passing it in in L83?

src/utils.ts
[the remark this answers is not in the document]
dariakp: Would there be an issue when the server is updated to a later version? I'm not 100% clear on the implications of setting maxWireVersion higher than the "actual" max wire version of the underlying server

src/sessions.ts
```diff
@@ -181,6 +198,41 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
+
+  /** @internal */
+  get pinnedConnection(): Connection | undefined {
+    return this[kPinnedConnection];
+  }
+
+  /** @internal */
+  pin(conn: Connection): void {
+    if (this[kPinnedConnection]) {
+      throw TypeError('Cant pin multiple connections to the same session');
```
dariakp, line 213:
```suggestion
      throw TypeError('Cannot pin multiple connections to the same session');
```

src/sessions.ts
```diff
@@ -189,21 +241,25 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
     callback?: Callback<void>
   ): void | Promise<void> {
     if (typeof options === 'function') (callback = options), (options = {});
-    options = options ?? {};
+    const _options = options ?? ({} as EndSessionOptions);
+    if (_options.force == null) _options.force = true;
 
     return maybePromise(callback, done => {
       if (this.hasEnded) {
+        maybeClearPinnedConnection(this, _options);
```
dariakp, line 256: do the new tests fully exercise this new pinning/unpinning logic? the code *seems* fine, but without it being clearly outlined in the tests (or elsewhere?) exactly when things are supposed to get pinned/unpinned, it's hard to tell by just looking at this file whether the resulting behavior will be 100% correct... case in point, that comment regarding unpinning on errors in L477; I'd be more comfortable with these changes if we had a single set of tests that outlined the expected behavior

dariakp (changes requested):
I went through and resolved the comments that were fully addressed; there are still a few that don't have replies (they might have gotten hidden by github on the main page) and some others that are open for discussion

src/utils.ts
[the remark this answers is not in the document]
dariakp: Based on our conversation and the fact that this is something that it sounds like was an acceptable risk per the spec, can we add a comment here briefly mentioning the potential issues and that this is intentional?

test/unit/mongo_client_options.test.js
durran: Nice catch. It wasn't passing so I've now updated the tests for the correct expectations.

src/transactions.ts
durran: Ah. Nice suggestion. I've refactored that now and also with `isActive` as well.

test/unit/cmap/stream_description.test.js
durran: Whoops, missed that. It's there now.

src/cmap/connection_pool.ts
```diff
@@ -219,6 +245,25 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+
+  get currentCheckedOutCount(): number {
+    return this[kCheckedOut];
+  }
+
+  /**
+   * Get the metrics information for the pool when a wait queue timeout occurs.
+   */
+  private waitQueueErrorMetrics(): string {
+    return `. maxPoolSize: ${this.options.maxPoolSize}, ${this[kMetrics].info()}`;
```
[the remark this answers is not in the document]
durran: I've changed to now pass the `maxPoolSize` to the metrics' `info` method to get the entire string there.

src/sessions.ts
durran: The load balancer spec tests add a new assertion to the unified test runner, `assertNumberConnectionsCheckedOut`, which is checking to make sure connections remained pinned or are returned to the pool based on the pool's count. These are tested for both cursors (https://github.com/mongodb/node-mongodb-native/blob/4.0/test/spec/load-balancers/cursors.json) and transactions (https://github.com/mongodb/node-mongodb-native/blob/4.0/test/spec/load-balancers/transactions.json)

src/cmap/connection.ts
mbroadst: What I like about this solution is we remove one callback from every operation in a transaction, and make the code in the CMAP module a little easier to understand. I _think_ it might be enough to add the logic to [the server error handling code](https://github.com/mongodb/node-mongodb-native/blob/4.0/src/sdam/server.ts#L450) since we still execute all operations against a `Server` (rather than a `Connection`).

src/cmap/connection_pool.ts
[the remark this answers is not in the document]
dariakp: works for me, thanks!

src/sessions.ts
dariakp: are they also tested for non load balanced clusters? and is there something that needs to happen to address the comment in L477?

test/unit/cmap/metrics.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+const { expect } = require('chai');
+const { ConnectionPoolMetrics } = require('../../../src/cmap/metrics');
+
+describe('ConnectionPoolMetrics', function () {
+  describe('#constructor', function () {
+    const metrics = new ConnectionPoolMetrics();
+
+    it('defaults txnConnections to zero', function () {
+      expect(metrics.txnConnections).to.equal(0);
```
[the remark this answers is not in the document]
mbroadst: glad somebody wrote this, I saw another case and came here to leave a comment. I personally prefer the somewhat shorter:
```
expect(metrics).property('txnConnections').to.equal(0);
```

src/sessions.ts
```diff
@@ -189,21 +241,25 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
+  endSession(options: EndSessionOptions, callback: Callback<void>): void;
   endSession(
-    options?: Record<string, unknown> | Callback<void>,
+    options?: EndSessionOptions | Callback<void>,
     callback?: Callback<void>
   ): void | Promise<void> {
     if (typeof options === 'function') (callback = options), (options = {});
-    options = options ?? {};
+    const _options = options ?? ({} as EndSessionOptions);
+    if (_options.force == null) _options.force = true;
```
nbbeeken, line 252:
```suggestion
    if (typeof options === 'function') {
      callback = options;
    }
    const finalOptions = { force: true, ...options };
```

We have some duplication of defaulting the options to an empty object here, and we can combine the defaulting of force. I suggest finalOptions here, we've just used that pattern elsewhere. 

Why is force default true now?

src/sessions.ts
```diff
@@ -397,6 +457,49 @@ function isUnknownTransactionCommitResult(err: MongoError) {
...
+    return;
+  }
+
+  // NOTE: the spec talks about what to do on a network error only, but the tests seem to
+  //       to validate that we don't unpin on _all_ errors?
+  if (conn && (options?.error == null || options?.force)) {
+    const servers = Array.from(session.topology.s.servers.values());
+    if (servers.length === 0) {
+      // This can happen if the client is closed when the connection is still pinned
+      // NOTE: we don't usually do this, we could instead throw an error?
```
nbbeeken, line 483: Before we decide on whether to throw or just destory is there a snippet that can make this code reachable? Where would I close the client such that a connection is still pinned?

src/sessions.ts
```diff
@@ -748,6 +863,11 @@ export class ServerSessionPool {
    */
   release(session: ServerSession): void {
     const sessionTimeoutMinutes = this.topology.logicalSessionTimeoutMinutes;
+
+    if (this.topology.loadBalanced) {
+      this.sessions.unshift(session);
```
nbbeeken, line 868: Is it possible to have a double unshift here? we're falling through to the logic beneath which could call unshift again

src/transactions.ts
```diff
@@ -45,6 +45,14 @@ const stateMachine: { [state in TxnState]: TxnState[] } = {
   ]
 };
 
+const ACTIVE_STATES: TxnState[] = [TxnState.STARTING_TRANSACTION, TxnState.TRANSACTION_IN_PROGRESS];
```
nbbeeken, line 48: Should these be Sets and we can use has? the teensiest speed improvement.

test/functional/unified-spec-runner/runner.ts
```diff
@@ -15,6 +16,25 @@ interface MongoDBMochaTestContext extends Mocha.Context {
   configuration: TestConfiguration;
 }
 
+export function trace(message: string): void {
+  if (process.env.UTR_TRACE) {
+    console.log(` > ${message}`);
```
nbbeeken, line 21: Can we change this to `console.error`? even though its for debugging its important to not interfere with the mocha reporter's output.

test/functional/unified-spec-runner/runner.ts
```diff
@@ -125,7 +157,15 @@ export async function runUnifiedTest(
     }
 
     for (const operation of test.operations) {
-      await executeOperationAndCheck(operation, entities, utilClient);
+      trace(operation.name);
+      try {
+        await executeOperationAndCheck(operation, entities, utilClient);
+      } catch (err) {
```
nbbeeken, line 163: Using `finally` here will let you run the `terminateOpenTransactions` helper and bubble up the error without needing to rethrow it explicitly.

test/functional/change_stream_spec.test.js
```diff
@@ -280,3 +281,19 @@ describe('Change Stream Spec', function () {
     return () => target[command].apply(target, args);
   }
 });
+
+describe('Change Streams Spec Unified Tests', function () {
```
nbbeeken, line 285: Change stream tests has been merged so this can be rebased now.

test/manual/load-balancer.test.js
```diff
@@ -0,0 +1,49 @@
...
+require('../functional/retryable_writes.test');
+require('../functional/transactions.test');
+require('../functional/uri_options_spec.test');
+require('../functional/change_stream_spec.test');
+require('../functional/transactions.test');
+require('../functional/versioned-api.test');
+require('../unit/core/mongodb_srv.test');
+require('../unit/sdam/server_selection/spec.test');
+
+describe('Load Balancer Spec Unified Tests', function () {
```
nbbeeken, line 31: I do see some custom logic in here, can we use `runUnifiedSuite` or does it not support what's needed?

test/tools/mock.js
```diff
@@ -7,8 +7,14 @@ const {
   DEFAULT_ISMASTER_36
   // eslint-disable-next-line no-restricted-modules
 } = require('mongodb-mock-server');
+const { ObjectId } = require('bson');
 const { HostAddress } = require('../../src/utils');
 
+const DEFAULT_HELLO_50 = {
```
nbbeeken, line 13: This looks unused, maybe we drop this change?

nbbeeken (changes requested):
I just merged in the changes stream test changes, so you can sync those in making the PR a bit smaller. 
Got some nits and some more concept questions here:

test/functional/unified-spec-runner/runner.ts
durran: Changed to use finally.

test/functional/unified-spec-runner/runner.ts
durran: Updated to use `console.error`

src/cmap/connection_pool.ts
```diff
@@ -365,7 +432,26 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
    * @param callback - The original callback
    */
-  withConnection(fn: WithConnectionCallback, callback?: Callback<Connection>): void {
+  withConnection(
+    conn: Connection | undefined,
+    fn: WithConnectionCallback,
+    callback?: Callback<Connection>
+  ): void {
+    if (conn) {
+      // use the provided connection, and do _not_ check it in after execution
```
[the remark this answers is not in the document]
durran: TSDoc updated per suggestion.

src/cmap/connection.ts
```diff
@@ -680,8 +660,10 @@ export class CryptoConnection extends Connection {
...
+export function hasSessionSupport(conn: Connection): boolean {
```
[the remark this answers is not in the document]
durran: Marked `@internal`

src/sdam/server.ts
```diff
@@ -397,9 +462,10 @@ Object.defineProperty(Server.prototype, 'clusterTime', {
 
 function supportsRetryableWrites(server: Server) {
   return (
-    server.description.maxWireVersion >= 6 &&
-    server.description.logicalSessionTimeoutMinutes &&
-    server.description.type !== ServerType.Standalone
+    server.loadBalanced ||
```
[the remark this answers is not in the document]
durran: Moved to utils, but there is still a weird typescript error that forced me to show the function's return type as `boolean`, `number` or `undefined`, which I still need to sort out.

src/utils.ts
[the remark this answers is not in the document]
durran: Added comment to the check.

test/functional/retryable_writes.test.js
```diff
@@ -99,8 +99,8 @@ function executeScenarioTest(test, ctx) {
               /expected false to be true/
             );
             if (hasResult) expect(err.result).to.matchMongoSpec(test.outcome.result);
-            if (errorLabelsContain) expect(err.errorLabels).to.have.members(errorLabelsContain);
-            if (errorLabelsOmit) expect(err.errorLabels).to.not.have.members(errorLabelsOmit);
+            if (errorLabelsContain) expect(err.errorLabels).to.containSubset(errorLabelsContain);
```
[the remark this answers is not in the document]
durran: Updated to suggestion.

test/functional/retryable_writes.test.js
```diff
@@ -99,8 +99,8 @@ function executeScenarioTest(test, ctx) {
...
+            if (errorLabelsOmit) expect(err.errorLabels).to.not.containSubset(errorLabelsOmit);
```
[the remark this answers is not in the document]
durran: I've changed to iterate and test individually.

test/unit/cmap/metrics.test.js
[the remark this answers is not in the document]
durran: Changed all the assertions to check the property now like Matt's first suggestion.

nbbeeken (changes requested):
My review request got cleared early I think, there are still comments outstanding (maybe collapsed by GH)

src/utils.ts
```diff
@@ -1385,3 +1394,17 @@ export function emitWarningOnce(message: string): void {
     return emitWarning(message);
   }
 }
+
+/**
+ * Determine if a server supports retryable writes.
+ *
+ * @internal
+ */
+export function supportsRetryableWrites(server: Server): boolean | undefined | number {
```
dariakp, line 1403: I think this can be tightened to boolean if we `!!` the `server.loadBalanced` and/or `server.description.logicalSessionTimeoutMinutes`, which is where I imagine the number and undefined could be coming from

src/sdam/server.ts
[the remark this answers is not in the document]
dariakp: I think it's because we are referencing the actual properties and not doing `!!` on the checks, I'm leaving a comment to that end on the utils file and resolving this conversation

dariakp (approved):
I went through and resolved all of my comments that didn't require follow up. There's a few things left on outstanding TODOs, but I think we're pretty close

dariakp (changes requested):
Accidentally cleared my review, resetting to "request changes" for the unresolved comments

src/cmap/connection.ts
durran: I've moved this out now over the the server class error handling to check if we are in a transaction and apply the label and everything is passing.

src/cmap/connection_pool.ts
```diff
@@ -437,7 +530,29 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
...
+      if (serviceId) {
+        let generation;
+        const sid = serviceId.toHexString();
+        if (!pool.serviceGenerations.has(sid)) {
+          generation = 0;
+          pool.serviceGenerations.set(sid, generation);
+        } else {
+          generation = pool.serviceGenerations.get(sid);
+        }
+        connection.generation = generation as number;
```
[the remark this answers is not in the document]
durran: Updated per suggestion with some eslint errors fixed (needed extra parens to avoid generation not expression error)

src/cmap/metrics.ts
```diff
@@ -0,0 +1,65 @@
...
+  txnConnections!: number;
+  cursorConnections!: number;
+  otherConnections!: number;
+
+  /**
+   * Create the metrics object.
+   */
+  constructor() {
+    this.reset();
+  }
```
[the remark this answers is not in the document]
durran: Updated to suggestion.

src/operations/command.ts
```diff
@@ -109,6 +112,10 @@ export abstract class CommandOperation<T> extends AbstractOperation<T> {
     return true;
   }
 
+  get bypassPinningCheck(): boolean {
```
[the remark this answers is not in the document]
durran: Moved up to `AbstractOperation`, added as a property there, and removed the two getters.

src/sdam/server.ts
```diff
@@ -43,14 +44,15 @@ import {
...
+import type { Long } from '../bson';
 import type { Topology } from './topology';
 import type {
   ServerHeartbeatFailedEvent,
   ServerHeartbeatStartedEvent,
   ServerHeartbeatSucceededEvent
 } from './events';
 import type { ClientSession } from '../sessions';
-import type { Document, Long } from '../bson';
+import type { Document } from '../bson';
```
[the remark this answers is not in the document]
durran: Updated.

src/sdam/server.ts
```diff
@@ -281,19 +303,43 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
-        markServerUnknown(this, err);
-        return cb(err);
-      }
+    const session = finalOptions.session;
+    const conn = session?.pinnedConnection;
+
+    // NOTE: This is a hack! We can't retrieve the connections used for executing an operation
+    //       (and prevent them from being checked back in) at the point of operation execution.
+    //       This should be considered as part of the work for NODE-2882
+    if (session && conn == null && this.loadBalanced && isPinnableCommand(cmd, session)) {
```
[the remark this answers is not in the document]
durran: Moved lb check to first conditional.

src/sdam/server.ts
```diff
@@ -331,19 +386,24 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
     }
 
-    this.s.pool.withConnection((err, conn, cb) => {
-      if (err || !conn) {
-        markServerUnknown(this, err);
-        return cb(err);
-      }
+    const conn = options.session?.pinnedConnection;
+    this.s.pool.withConnection(
+      conn,
```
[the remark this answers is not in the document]
durran: Updated all the similar operations to this now.

src/sdam/topology.ts
```diff
@@ -411,7 +417,19 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
 
     // connect all known servers, then attempt server selection to connect
-    connectServers(this, Array.from(this.s.description.servers.values()));
+    const serverDescriptions = Array.from(this.s.description.servers.values());
+    connectServers(this, serverDescriptions);
+
+    // In load balancer mode we need to fake a server description getting
+    // emitted from the monitor, since the monitor doesn't exist.
+    if (this.s.options.loadBalanced) {
+      serverDescriptions.forEach(description => {
```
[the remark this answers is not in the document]
durran: Changed to for of loop.

src/sessions.ts
durran: Changed to use `finalOptions` pattern here now. As for forcing `true` to default, ending the session should always unpin the connection in the default case, even out of lb mode. (Since it won't ever be pinned anyway and would be a no-op). The only thing we ever want to specify is when *not* to unpin.

src/sessions.ts
durran: This would cause this to get called in lb mode.

```
cursor.next((err, doc) => {
  client.close();
});
```

src/transactions.ts
durran: Changed to sets.

test/manual/load-balancer.test.js
durran: Updated to use `runUnifiedSuite`.

src/utils.ts
durran: Ah bingo. This is updated now.

src/cmap/connect.ts
durran: This is now updated to always use `processId`.

src/error.ts
```diff
@@ -546,25 +552,22 @@ export class MongoNetworkError extends MongoError {
...
 
-  constructor(message: string | Error, options?: { beforeHandshake?: boolean }) {
+  constructor(message: string | Error, options?: MongoNetworkErrorOptions) {
     super(message);
 
     if (options && typeof options.beforeHandshake === 'boolean') {
       this[kBeforeHandshake] = options.beforeHandshake;
     }
+
+    // TODO: must not apply when running `commitTransaction`
```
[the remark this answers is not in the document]
durran: When the label was moved to `Server`, we now check for commit and abort before applying it.

src/cursor/abstract_cursor.ts
```diff
@@ -757,18 +719,71 @@ function cursorIsDead(cursor: AbstractCursor): boolean {
...
-  } else {
-    callback();
+  if (cursorId == null || server == null || cursorId.isZero() || cursorNs == null) {
+    if (needsToEmitClosed) {
+      cursor[kClosed] = true;
+      cursor[kId] = Long.ZERO;
+      cursor.emit(AbstractCursor.CLOSE);
+    }
+
+    if (session) {
```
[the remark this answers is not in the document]
durran: I've wrapped this all into `completeCleanup` now, which also fixed the cursor's close event from getting emitted multiple times. Ever I left a todo to rename an option name to the cursor since we've already released 4.0.

src/cursor/abstract_cursor.ts
```diff
@@ -757,18 +719,71 @@ function cursorIsDead(cursor: AbstractCursor): boolean {
...
   }
 
-  const session = cursor[kSession];
-  if (session && session.owner === cursor) {
-    session.endSession(callback);
-  } else {
-    callback();
+  if (cursorId == null || server == null || cursorId.isZero() || cursorNs == null) {
+    if (needsToEmitClosed) {
+      cursor[kClosed] = true;
```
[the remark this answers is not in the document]
durran: I've removed all the need for this in multiple places now. There is one thing I did not refactor as it would have changed the public `close` api is that the `skipKillCursors` option is never used that I could see, and passing `options` to `close` could be removed. However since 4.0 has already been released I didn't make this change. I created https://jira.mongodb.org/browse/NODE-3489 for this.

src/operations/command.ts
```diff
@@ -47,6 +47,9 @@ export interface CommandOperationOptions
   dbName?: string;
   authdb?: string;
   noResponse?: boolean;
+
+  /** @internal Hints to `executeOperation` that this operation should not unpin on an ended transaction */
+  bypassPinningCheck?: boolean;
```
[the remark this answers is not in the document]
durran: https://jira.mongodb.org/browse/NODE-3490 has been created for this.

src/sessions.ts
durran: No there are no tests for that as connection pinning does not happen outside of load balancer mode and is only part of the LB test suite.

src/cursor/abstract_cursor.ts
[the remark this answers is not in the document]
durran: Actually I had to revert this commit as it caused changes that broke cursors in regular mode in the case where a cursor was closed and there were still documents in memory to be iterated. I'll create another ticket to address that.

test/functional/unified-spec-runner/runner.ts
```diff
@@ -15,6 +16,25 @@ interface MongoDBMochaTestContext extends Mocha.Context {
...
 }
 
+export function trace(message: string): void {
+  if (process.env.UTR_TRACE) {
+    console.log(` > ${message}`);
+  }
+}
+
+async function terminateOpenTransactions(client: MongoClient) {
+  // TODO: on sharded clusters this has to be run on each mongos
```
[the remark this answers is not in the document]
durran: https://jira.mongodb.org/browse/NODE-3491

src/cursor/abstract_cursor.ts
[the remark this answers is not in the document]
mbroadst: FWIW, I think the real reason that `close` sets the flag immediately is because there is a slim chance that cursor state changes between the sync call to `close` and the `maybePromise` code called in its body, iirc a "tick" might go by between those two calls.

src/sessions.ts
nbbeeken: This code already throws with `MongoDriverError: Attempted to use a session that has ended` which makes sense because close is an async operation that can't stop .next() from moving on to the next document. I tried without awaiting close:

```js
for await (const doc of cursor) {
  client.close()
}
```
which throws 
```
./src/sdam/server.ts:330
      callback(new MongoDriverError('server is closed'));
```

And awaiting close
```js
for await (const doc of cursor) {
  await client.close()
}
```
which throws:
```
./src/cmap/connection_pool.ts:323
          waitQueueMember.callback(new MongoDriverError('connection pool closed'));
```

So I guess we have a precedent of throwing, I also don't know how safe it is to call destroy here without providing a callback. Don't we need to confirm when destroy is complete? (I'm actually not sure, it might be fine). If we do throw here we need maybClearPinnedConnection to take a callback or it could return the error? Or if we destroy then we still might need to involve a callback here, thoughts?

src/sessions.ts
nbbeeken: Can we be in loadbalanced mode with a non-zero logicalSessionTimeoutMinutes (this should always be non-zero in LB) and the current session has not timed out and isn't dirty? If so then we're duplicating the insert if I'm reading the booleans correctly.

dariakp (commented):
Just following up on a couple of questions, resolved most outstanding comments

src/sessions.ts
nbbeeken: Ah ok yea I can see the the logicalSessionTimeoutMinutes isn't defined. I agree we should assert something here I think then

nbbeeken (changes requested):
Just looking for some discussion on these:
- [kProcessingWaitQueue question](https://github.com/mongodb/node-mongodb-native/pull/2909#discussion_r676855931)
- [Throw to prevent double unshift?](https://github.com/mongodb/node-mongodb-native/pull/2909#discussion_r676860773)
- And I replied to [the throwing when maybeClearingPinned](https://github.com/mongodb/node-mongodb-native/pull/2909#discussion_r674264936)

src/sessions.ts
dariakp: I guess my confusion is more around the two types of pinning that we do, some of which is now handled by the same logic - there was already session server pinning and the session.transaction.unpinServer code that was in place... I'm trying to understand how the purpose of the original logic relates to the new connection pinning

src/sessions.ts
durran: It seems maybe more correct that we probably ensure destroy finishes, as it also removes the listeners from the pinned connection, and then we return a better worded error in the callback. Would you be ok with that?

src/sessions.ts
durran: I've added the assertion that `sessionTimeoutMinutes` also does not exist.

src/cmap/connection_pool.ts
```diff
@@ -465,10 +580,11 @@ function destroyConnection(pool: ConnectionPool, connection: Connection, reason:
 }
 
 function processWaitQueue(pool: ConnectionPool) {
-  if (pool.closed) {
+  if (pool.closed || pool[kProcessingWaitQueue]) {
```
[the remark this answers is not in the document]
durran: I think it had to do with testing where the queue was getting extremely fast checkouts in succession in the new unified tests, but I'll open another ticket for further investigation.

src/cursor/abstract_cursor.ts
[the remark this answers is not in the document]
durran: I've created https://jira.mongodb.org/browse/NODE-3495 to address refactoring this.

test/functional/unified-spec-runner/runner.ts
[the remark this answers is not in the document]
durran: It wasn't causing any failures before (as it wasn't implemented as the spec stated) so I think it's ok to address as a unified test format specific ticket.

src/sessions.ts
durran: Yes the pinning logic was combined, just the resource getting pinned is different. But according to the transaction and load balancer specs, the pinning behaviour is the same, we are just pinning to a `Connection` in LB mode (to ensure executing against the same mongos behind the load balancer) and a `Server` in non-lb mode (since it's a direct connection to the mongos and any connection in the server's connection pool with execute on it)

src/cmap/connection_pool.ts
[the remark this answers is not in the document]
durran: https://jira.mongodb.org/browse/NODE-3495

src/sessions.ts
nbbeeken: I took a look at the usages its going to be quite the refactor to make this function accept a callback, its doable but ClientSession#unpin for example doesn't currently take a callback so that will need it too, which branch out to its usages.

Throwing synchronously I think would also cause its own set of problems where `promise.catch()`s wouldn't work correctly.

The downside of not destroying is that we're leaking this connection right? If someone calls close while looping their cursor then .endSessions is called: https://github.com/mongodb/node-mongodb-native/blob/4.0/src/sdam/topology.ts#L486-L488 but it isn't currently awaited (or callback used) so even if we piped through the callback there would never be one defined 😕 this is tricky. 

I'm inclined we throw synchronously here and come back to fix this up? (@dariakp just want a second 👀 on this weird situation)

dariakp: I don't fully understand the implications of either side here... can connections jump between multiple pools? If not, it seems that there is no reason for a pool or the original server not to keep track of pinned connections so that when the server or the pool is destroyed, that connection is taken care of along with all the rest. I am assuming since the server length is 0, the actual associated server and pool was, in fact, destroyed by some existing mechanism.

dariakp: I guess I also want to understand 1) If we remove this code entirely, what's the undesirable result that may happen? and 2) If we leave this code as is and don't throw or call back with an error, what's the undesirable result that may happen? so, ultimately, the question is 3) In the ideal scenario, if we do reach this point, what's end state that we want to propagate to the user?

durran: I think the end state that we want is that 1) the connection is destroyed (and finishes that process so the event listeners are removed and the socket is closed, and 2) we give the user a meaningful error about what they did to get in that state.

So if we remove entirely, we are leaking connections as per point 1. If we leave as is, the user gets an error that might not truly explain why they got it. So I think the suggestion by @nbbeeken to just throw for now is the simplest solution in order to refactor into some callback craziness.

nbbeeken: > Just dropping a note from standup, I think the fact that we don't await endSessions in the topology.close method is a bug and causes this servers.length === 0 case to happen.

durran: My latest changes to address this (make sure all `endSession` calls finished before moving forward in the topology) didn't prove to be working, so I'll revert and make another ticket.

durran: This is now resolved as per the suggestion of waiting for all the end session calls to complete during closing of the topology.

dariakp:
LGTM pending merge conflict resolution and green tests

src/cursor/abstract_cursor.ts
```diff
@@ -836,7 +852,7 @@ function makeCursorStream<TSchema extends Document>(cursor: AbstractCursor<TSche
         // NOTE: This is also perhaps questionable. The rationale here is that these errors tend
         //       to be "operation interrupted", where a cursor has been closed but there is an
         //       active getMore in-flight.
-        if (cursor.killed) {
+        if (err.message.match(/interrupted/)) {
```
nbbeeken, line 855: Is this a safe check? Maybe we want to capture rational in a comment here

src/operations/aggregate.ts
```diff
@@ -79,6 +79,10 @@ export class AggregateOperation<T = Document> extends CommandOperation<T> {
     }
   }
 
+  get isCursorCreating(): boolean {
```
nbbeeken, line 82: This seems it would be better suited to be an [aspect](https://github.com/mongodb/node-mongodb-native/blob/4.0/src/operations/aggregate.ts#L130). If we could refactor to that it would be great but also something we can defer as well.

mbroadst: Do we actually need this on operations classes? The cursor design is such that subclasses of `AbstractCursor` implement their own initialization which execute an operation, so we _should_ have all the context we need about which operations create cursors without explicitly adding that context to operations classes. Is there a reason you can't move the check in `executeOperation` into the `AbstractCursor` call to [_initialize](https://github.com/mongodb/node-mongodb-native/blob/4.0/src/cursor/abstract_cursor.ts#L686) ?

src/cursor/abstract_cursor.ts
durran: The cursor was previously only killed on cursor.close() but it changed to be killed during cleanup. Which meant that legitimate errors were not causing the stream to destroy because at this point the cursor was _always_ killed.

src/operations/aggregate.ts
durran: Well the addition of this was specifically for retry behaviour, and I wanted to keep all the retry related code in the same place and not have it scattered in multiple places.

durran: Also because of the `findOne` case - it doesn't directly instantiate a cursor, it calls `collection.find` in it's operation which means that the session in that case it not owned by the cursor as `executeOperation` creates the implicit session, not `cursor._initialize`.

durran: Refactored in a026ce4ca2d7dd337f4b13b0cac09d7a813a6924 to use aspects.

src/cursor/abstract_cursor.ts
durran: Also updated the comment in 363d41b22523c6324693f92ea6f6db88c5b85458

src/operations/aggregate.ts
mbroadst: Hm, but I think this actually complicates the retry behavior. The logic in `executeOperation` now is very clear: if the operation is retryable, and the error is retryable, the operation is retried. Now we have a special logical block that has specific knowledge about cursors, and I'm worried that we will continue to lump extraneous checks there. 

I'm not sure I follow the `findOne` case, can you explain that a bit more? It should be fine that it calls another operation internally. In fact, I think marking `FindOneOperation` as being  "cursor creating" is incorrect: it issues a command which returns a single-batch, exhausted cursor. The session should not have a pinned connection after that completes, and if it does that would mean we have a bug somewhere else. Unfortunately, we'd never see that now because we put this aspect on the operation, and that means in `executeOperation` we are just force unpinning, and force resetting the pool. Seems dangerous.

nbbeeken: What tests were failing without this condition? I tried running the tests locally without the creates_cursor aspect and all passed. There are some skips so maybe my environment is incomplete. (I have both multi and single env variables)

durran: No tests fail without this condition, but it leaks the connection each time `findOne` retries. This is because `findOne` calls `executeOperation`, which itself creates an implicit session with the `owner` being a symbol. Then inside it's execute method it calls `collection.find`, which creates a cursor with the already created implicit session. The connection is pinned at the connection pool level because `find` is a pinnable operation, and then when the cursor exhausts it does not unpin because the cursor is not the owner of the session. So the connection remains pinned to the session.

durran: Screenshot 2021-08-03 at 16 18 31

durran: You can see at the end of the test run here without this code that there are 2 sessions still open to the load balancer, which correlated directly with 2 findOne retries that had an extra connection checked out of the pool when the client was closed.

durran: Screenshot 2021-08-03 at 16 18 31

Here are the extra 2 checked out connections from the findOne retries at the end of the test run.

durran: I do agree in this case that the pin should never have occurred, but until we refactor the operation layer it's always going to be pinned here because we pin on connection checkout from the pool and the find command is pinnable.

nbbeeken (approved):

test/functional/unified-spec-runner/runner.ts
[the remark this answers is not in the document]
emadum:
```suggestion
  // TODO(NODE-3491): on sharded clusters this has to be run on each mongos
```

src/sdam/server.ts
```diff
@@ -153,6 +154,8 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       this.clusterTime = clusterTime;
     });
 
+    if (this.loadBalanced) return;
```
emadum, line 157:
```suggestion
    // monitoring is disabled in load balancing mode
    if (this.loadBalanced) return;
```
I think it might be useful to call this out here for future maintainers.

emadum (approved):
Couple of small suggestions but this is looking good to me. 🥇

dariakp (approved):

PR #2910, opened 2021-07-15
refactor(NODE-3421): Replace MongoDriverError instances with MongoRuntimeError children

W-A-James:
## Description

Replace appropriate MongoDriverError instances with instances of the following classes

* MongoBatchReExecutionError
* MongoCursorError
	* MongoCursorExhaustedError
	* MongoTailableCursorError
	* MongoCursorInUseError
* MongoNotConnectedError

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -1181,12 +1181,7 @@ export abstract class BulkOperationBase {
     options = options ?? {};
 
     if (this.s.executed) {
-      return handleEarlyError(
-        new MongoBatchReExecutionError(
-          'This batch has already been executed: Must create a new batch to execute'
-        ),
-        callback
-      );
+      return handleEarlyError(new MongoBatchReExecutionError(), callback);
     }
 
     const writeConcern = WriteConcern.fromOptions(options);
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -5,8 +5,8 @@ import {
   AnyError,
   MONGODB_ERROR_CODES,
   MongoServerError,
-  MongoDriverError,
-  MongoInvalidArgumentError
+  MongoInvalidArgumentError,
+  MongoBatchReExecutionError
 } from '../error';
 import {
   applyRetryableWrites,
@@ -1181,7 +1181,10 @@ export abstract class BulkOperationBase {
     options = options ?? {};
 
     if (this.s.executed) {
-      return handleEarlyError(new MongoDriverError('Batch cannot be re-executed'), callback);
+      return handleEarlyError(
+        new MongoBatchReExecutionError('Batch cannot be re-executed'),
+        callback
+      );
     }
 
     const writeConcern = WriteConcern.fromOptions(options);
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -1182,7 +1182,9 @@ export abstract class BulkOperationBase {
 
     if (this.s.executed) {
       return handleEarlyError(
-        new MongoBatchReExecutionError('Batch cannot be re-executed'),
+        new MongoBatchReExecutionError(
+          'This batch has already been executed: Must create a new batch to execute'
+        ),
         callback
       );
     }
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -1,7 +1,12 @@
 import { Callback, maybePromise, MongoDBNamespace, ns } from '../utils';
 import { Long, Document, BSONSerializeOptions, pluckBSONSerializeOptions } from '../bson';
 import { ClientSession } from '../sessions';
-import { MongoDriverError, MongoInvalidArgumentError } from '../error';
+import {
+  MongoDriverError,
+  MongoInvalidArgumentError,
+  MongoCursorExhaustedError,
+  MongoTailableCursorError
+} from '../error';
 import { ReadPreference, ReadPreferenceLike } from '../read_preference';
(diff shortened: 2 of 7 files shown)


andymina (approved):
LGTM 👍🏾

src/cursor/abstract_cursor.ts
```diff
@@ -284,7 +289,7 @@ export abstract class AbstractCursor<
   next<T = TSchema>(callback?: Callback<T | null>): Promise<T | null> | void {
     return maybePromise(callback, done => {
       if (this[kId] === Long.ZERO) {
-        return done(new MongoDriverError('Cursor is exhausted'));
+        return done(new MongoCursorExhaustedError('Cursor is exhausted'));
```
dariakp, line 292: Have we considered moving this generic message to the error constructor? Do we think there might be a different message related to the same error?

dariakp (commented):
Just some food for thought - the same comment applies to other specific-use errors as well

src/bulk/common.ts
```diff
@@ -1180,7 +1181,10 @@ export abstract class BulkOperationBase {
     options = options ?? {};
 
     if (this.s.executed) {
-      return handleEarlyError(new MongoDriverError('Batch cannot be re-executed'), callback);
+      return handleEarlyError(
+        new MongoBatchReExecutionError('Batch cannot be re-executed'),
```
nbbeeken, line 1185: Along the lines of Daria's comment: I think this is a good place to apply the 'what must be' style of message like, 'A new bulk operation must be created to in order to execute' as a rough suggestion, improvement needed of course

src/mongo_client.ts
```diff
@@ -528,7 +528,7 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
   startSession(options?: ClientSessionOptions): ClientSession {
     options = Object.assign({ explicit: true }, options);
     if (!this.topology) {
-      throw new MongoDriverError('Must connect to a server before calling this method');
+      throw new MongoNotConnectedError('Must connect to a server before calling this method');
```
nbbeeken, line 531: Let's take the chance to remove server from the message.
`MongoClient must be connected to start a session`

nbbeeken (changes requested):

src/cursor/abstract_cursor.ts
W-A-James: That's a really good point. For many of these errors they really only need one default message as far as I can see.

Do you think it would be a good idea to change those specific-use errors to have a default generic message, but still accept an optional message argument in case it may prove useful at some point in the future?

src/bulk/common.ts
W-A-James: Would something like this be appropriate:

`'This batch has already been executed: Must create new batch to execute'`

nbbeeken: 'This batch has already been executed, create new batch to execute' maybe?

src/cursor/abstract_cursor.ts
dariakp: Yes, I think that would be a good approach, to allow a message override, though I don't know if you need to even include it right now, it would be just as easy to add when we have an actual use case for it

W-A-James:
@andymina There is a TODO that refernces this ticket on src/db.ts: 752 that I probably added in NODE-3402. Feel free to point that to NODE-3484 instead

src/error.ts
```diff
@@ -206,8 +206,12 @@ export class MongoRuntimeError extends MongoDriverError {
  * @category Error
  */
 export class MongoBatchReExecutionError extends MongoRuntimeError {
-  constructor(message: string) {
-    super(message);
+  constructor(message?: string) {
+    if (message !== undefined) {
```
dariakp, line 210: I think it's unlikely that we'll have a use case for an empty string message, so, this can probably be shortened everywhere to just `super(message || '<default message>')`

dariakp (changes requested):

dariakp (commented):
I think we should make sure the errors in this ticket extend from the correct parent class - e.g., MongoNotConnectedError might be more of an incorrect use error than a runtime error (@nbbeeken thoughts?)

likewise, for the cursor methods, we really need to check whether those errors are thrown because of incorrect usage or if they should be "unreachable", and update which class they extend from accordingly

W-A-James:
@dariakp @nbbeeken Pulling the rebucketing of all error classes into this PR

test/functional/bulk.test.js
```diff
@@ -1892,7 +1892,7 @@ describe('Bulk', function () {
         expect(result).to.exist;
 
         bulk.execute(err => {
-          expect(err).to.match(/Batch cannot be re-executed/);
+          expect(err).to.match(/This batch has already been executed, create new batch to execute/);
```
nbbeeken, line 1895: I think we should change this to an instanceOf check for MongoBatchReExecutionError

nbbeeken (changes requested):

dariakp:
LGTM other than the suggestion Neal already pointed out

test/functional/bulk.test.js
W-A-James: Should I only change this one here and have a later PR handle changing all the error checks in the tests be instanceOf checks?

nbbeeken: This is the only one that was testing the message but it is preferred and its our API that its the instanceof that determines the error

nbbeeken (approved):

dariakp (approved):

PR #2912, opened 2021-07-16
refactor(NODE-3404): implement MongoRuntimeError children

andymina:
## Description
Replace instances of MongoDriverError with the appropriate children of MongoRuntimeError, listed below, as described in [docs/errors.md](https://github.com/mongodb/node-mongodb-native/blob/4.0/docs/errors.md?rgh-link-date=2021-07-07T20%3A54%3A43Z).

**Children of MongoRuntimeError used**
- MongoURIError
- MongoStreamError
- MongoChangeStreamError
- MongoGridFSStreamError
- MongoGridFSChunkError
- MongoCompressionError
- MongoDecompressionError

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -265,8 +265,8 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
     } else if (parent instanceof MongoClient) {
       this.type = CHANGE_DOMAIN_TYPES.CLUSTER;
     } else {
-      throw new MongoChangeStreamError(
-        'Argument "parent" provided to ChangeStream constructor must be an instance of Collection, Db, or MongoClient'
+      throw new MongoDriverError(
+        'parent provided to ChangeStream constructor is not an instance of Collection, Db, or MongoClient'
       );
     }
 
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -548,7 +548,7 @@ const CHANGE_STREAM_EVENTS = [
 
 function setIsEmitter<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'iterator') {
-    // TODO(NODE-3484): Replace with MongoChangeStreamModeError
+    // TODO(NODE-3485): Replace with MongoChangeStreamModeError
     throw new MongoAPIError(
       'ChangeStream cannot be used as an EventEmitter after being used as an iterator'
     );
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -1,38 +1,39 @@
 import Denque = require('denque');
-import type { Readable } from 'stream';
-import type { Document, Timestamp } from './bson';
+import {
+  MongoError,
+  AnyError,
+  isResumableError,
+  MongoDriverError,
+  MongoAPIError,
+  MongoChangeStreamError
+} from './error';
+import { AggregateOperation, AggregateOptions } from './operations/aggregate';
+import {
+  maxWireVersion,
+  calculateDurationInMs,
+  now,
+  maybePromise,
+  MongoDBNamespace,
+  Callback,
+  getTopology
+} from './utils';
+import type { ReadPreference } from './read_preference';
+import type { Timestamp, Document } from './bson';
+import type { Topology } from './sdam/topology';
+import type { OperationParent, CollationOptions } from './operations/command';
+import { MongoClient } from './mongo_client';
+import { Db } from './db';
 import { Collection } from './collection';
+import type { Readable } from 'stream';
 import {
   AbstractCursor,
   AbstractCursorEvents,
   AbstractCursorOptions,
   CursorStreamOptions
 } from './cursor/abstract_cursor';
-import { Db } from './db';
-import {
-  AnyError,
-  isResumableError,
-  MongoChangeStreamError,
-  MongoDriverError,
-  MongoError
-} from './error';
-import { MongoClient } from './mongo_client';
-import { InferIdType, Nullable, TypedEventEmitter } from './mongo_types';
-import { AggregateOperation, AggregateOptions } from './operations/aggregate';
-import type { CollationOptions, OperationParent } from './operations/command';
-import { executeOperation, ExecutionResult } from './operations/execute_operation';
-import type { ReadPreference } from './read_preference';
-import type { Topology } from './sdam/topology';
 import type { ClientSession } from './sessions';
-import {
-  calculateDurationInMs,
-  Callback,
-  getTopology,
-  maxWireVersion,
-  maybePromise,
-  MongoDBNamespace,
-  now
-} from './utils';
+import { executeOperation, ExecutionResult } from './operations/execute_operation';
+import { InferIdType, Nullable, TypedEventEmitter } from './mongo_types';
 
 /** @internal */
 const kResumeQueue = Symbol('resumeQueue');
@@ -265,8 +266,8 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
     } else if (parent instanceof MongoClient) {
       this.type = CHANGE_DOMAIN_TYPES.CLUSTER;
     } else {
-      throw new MongoDriverError(
-        'parent provided to ChangeStream constructor is not an instance of Collection, Db, or MongoClient'
+      throw new MongoChangeStreamError(
+        'Parent provided to ChangeStream constructor must be an instance of Collection, Db, or MongoClient'
       );
     }
 
@@ -547,7 +548,8 @@ const CHANGE_STREAM_EVENTS = [
 
 function setIsEmitter<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'iterator') {
-    throw new MongoChangeStreamError(
+    // TODO(NODE-3484): Replace with MongoChangeStreamModeError
+    throw new MongoAPIError(
       'ChangeStream cannot be used as an EventEmitter after being used as an iterator'
     );
   }
@@ -556,7 +558,7 @@ function setIsEmitter<TSchema>(changeStream: ChangeStream<TSchema>): void {
 
 function setIsIterator<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'emitter') {
-    throw new MongoChangeStreamError(
+    throw new MongoAPIError(
       'ChangeStream cannot be used as an EventEmitter after being used as an iterator'
     );
   }
@@ -634,7 +636,7 @@ function waitForTopologyConnected(
     }
 
     if (calculateDurationInMs(start) > timeout) {
-      // TODO: Replace with MongoNetworkTimeoutError
+      // TODO(NODE-3497): Replace with MongoNetworkTimeoutError
       return callback(new MongoDriverError('Timed out waiting for connection'));
     }
 
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -1,5 +1,11 @@
 import Denque = require('denque');
-import { MongoError, AnyError, isResumableError, MongoDriverError } from './error';
+import {
+  MongoError,
+  AnyError,
+  isResumableError,
+  MongoDriverError,
+  MongoChangeStreamError
+} from './error';
 import { AggregateOperation, AggregateOptions } from './operations/aggregate';
 import {
   maxWireVersion,
@@ -259,9 +265,8 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
     } else if (parent instanceof MongoClient) {
       this.type = CHANGE_DOMAIN_TYPES.CLUSTER;
     } else {
-      // TODO(NODE-3404): Replace this with MongoChangeStreamError
-      throw new MongoDriverError(
-        'Parent provided to ChangeStream constructor must an instance of Collection, Db, or MongoClient'
+      throw new MongoChangeStreamError(
+        'Argument "parent" provided to ChangeStream constructor must be an instance of Collection, Db, or MongoClient'
       );
     }
 
@@ -365,8 +370,7 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
    */
   stream(options?: CursorStreamOptions): Readable {
     this.streamOptions = options;
-    // TODO(NODE-3404): Replace this with MongoChangeStreamError
-    if (!this.cursor) throw new MongoDriverError(NO_CURSOR_ERROR);
+    if (!this.cursor) throw new MongoChangeStreamError(NO_CURSOR_ERROR);
     return this.cursor.stream(options);
   }
 
@@ -543,8 +547,8 @@ const CHANGE_STREAM_EVENTS = [
 
 function setIsEmitter<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'iterator') {
-    throw new MongoDriverError(
-      'Cannot use ChangeStream as an EventEmitter after using as an iterator'
+    throw new MongoChangeStreamError(
+      "ChangeStream can't be used as an EventEmitter after being used as an iterator"
     );
   }
   changeStream[kMode] = 'emitter';
@@ -552,8 +556,8 @@ function setIsEmitter<TSchema>(changeStream: ChangeStream<TSchema>): void {
 
 function setIsIterator<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'emitter') {
-    throw new MongoDriverError(
-      'Cannot use ChangeStream as iterator after using as an EventEmitter'
+    throw new MongoChangeStreamError(
+      "ChangeStream can't be used as an EventEmitter after being used as an iterator"
     );
   }
   changeStream[kMode] = 'iterator';
@@ -630,6 +634,7 @@ function waitForTopologyConnected(
     }
 
     if (calculateDurationInMs(start) > timeout) {
+      // TODO: Replace with MongoNetworkTimeoutError
       return callback(new MongoDriverError('Timed out waiting for connection'));
     }
 
@@ -676,17 +681,23 @@ function processNewChange<TSchema>(
   callback?: Callback<ChangeStreamDocument<TSchema>>
 ) {
   if (changeStream[kClosed]) {
+    // TODO(NODE-3405): Replace with MongoStreamClosedError
     if (callback) callback(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
 
   // a null change means the cursor has been notified, implicitly closing the change stream
   if (change == null) {
+    // TODO(NODE-3405): Replace with MongoStreamClosedError
     return closeWithError(changeStream, new MongoDriverError(CHANGESTREAM_CLOSED_ERROR), callback);
   }
 
   if (change && !change._id) {
-    return closeWithError(changeStream, new MongoDriverError(NO_RESUME_TOKEN_ERROR), callback);
+    return closeWithError(
+      changeStream,
+      new MongoChangeStreamError(NO_RESUME_TOKEN_ERROR),
+      callback
+    );
   }
 
   // cache the resume token
@@ -710,6 +721,7 @@ function processError<TSchema>(
 
   // If the change stream has been closed explicitly, do not process error.
   if (changeStream[kClosed]) {
+    // TODO(NODE-3405): Replace with MongoStreamClosedError
     if (callback) callback(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
@@ -770,6 +782,7 @@ function processError<TSchema>(
  */
 function getCursor<T>(changeStream: ChangeStream<T>, callback: Callback<ChangeStreamCursor<T>>) {
   if (changeStream[kClosed]) {
+    // TODO(NODE-3405): Replace with MongoStreamClosedError
     callback(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
@@ -795,11 +808,12 @@ function processResumeQueue<TSchema>(changeStream: ChangeStream<TSchema>, err?: 
     const request = changeStream[kResumeQueue].pop();
     if (!err) {
       if (changeStream[kClosed]) {
+        // TODO(NODE-3405): Replace with MongoStreamClosedError
         request(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
         return;
       }
       if (!changeStream.cursor) {
-        request(new MongoDriverError(NO_CURSOR_ERROR));
+        request(new MongoChangeStreamError(NO_CURSOR_ERROR));
         return;
       }
     }
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -1,38 +1,38 @@
 import Denque = require('denque');
-import {
-  MongoError,
-  AnyError,
-  isResumableError,
-  MongoDriverError,
-  MongoChangeStreamError
-} from './error';
-import { AggregateOperation, AggregateOptions } from './operations/aggregate';
-import {
-  maxWireVersion,
-  calculateDurationInMs,
-  now,
-  maybePromise,
-  MongoDBNamespace,
-  Callback,
-  getTopology
-} from './utils';
-import type { ReadPreference } from './read_preference';
-import type { Timestamp, Document } from './bson';
-import type { Topology } from './sdam/topology';
-import type { OperationParent, CollationOptions } from './operations/command';
-import { MongoClient } from './mongo_client';
-import { Db } from './db';
-import { Collection } from './collection';
 import type { Readable } from 'stream';
+import type { Document, Timestamp } from './bson';
+import { Collection } from './collection';
 import {
   AbstractCursor,
   AbstractCursorEvents,
   AbstractCursorOptions,
   CursorStreamOptions
 } from './cursor/abstract_cursor';
-import type { ClientSession } from './sessions';
-import { executeOperation, ExecutionResult } from './operations/execute_operation';
+import { Db } from './db';
+import {
+  AnyError,
+  isResumableError,
+  MongoChangeStreamError,
+  MongoDriverError,
+  MongoError
+} from './error';
+import { MongoClient } from './mongo_client';
 import { InferIdType, Nullable, TypedEventEmitter } from './mongo_types';
+import { AggregateOperation, AggregateOptions } from './operations/aggregate';
+import type { CollationOptions, OperationParent } from './operations/command';
+import { executeOperation, ExecutionResult } from './operations/execute_operation';
+import type { ReadPreference } from './read_preference';
+import type { Topology } from './sdam/topology';
+import type { ClientSession } from './sessions';
+import {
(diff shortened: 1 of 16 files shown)


W-A-James (changes requested):

W-A-James:
Apart from requested changes, LGTM

src/change_stream.ts
```diff
@@ -545,17 +551,17 @@ const CHANGE_STREAM_EVENTS = [
 
 function setIsEmitter<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'iterator') {
-    throw new MongoDriverError(
-      'Cannot use ChangeStream as an EventEmitter after using as an iterator'
+    throw new MongoChangeStreamError(
+      "ChangeStream can't be used as an EventEmitter after being used as an iterator"
```
W-A-James, line 555: I think it would be good to refrain from using contractions in error messages

src/change_stream.ts
```diff
@@ -545,17 +551,17 @@ const CHANGE_STREAM_EVENTS = [
...
   }
   changeStream[kMode] = 'emitter';
 }
 
 function setIsIterator<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'emitter') {
-    throw new MongoDriverError(
-      'Cannot use ChangeStream as iterator after using as an EventEmitter'
+    throw new MongoChangeStreamError(
+      "ChangeStream can't be used as an EventEmitter after being used as an iterator"
```
W-A-James, line 564:
```suggestion
      "ChangeStream cannot be used as an EventEmitter after being used as an iterator"
```

test/functional/gridfs_stream.test.js
```diff
@@ -473,7 +473,7 @@ describe('GridFS Stream', function () {
                     // Fail if user tries to abort an aborted stream
                     uploadStream.abort().then(null, function (error) {
                       expect(error.toString()).to.equal(
-                        'MongoDriverError: Cannot call abort() on a stream twice'
+                        'MongoGridFSStreamError: Cannot call abort() on a stream twice'
```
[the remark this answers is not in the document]
andymina: Initially I thought `MongoGridFSStreamError` would be good here, but upon second look `MongoStreamClosedError` looks/feels way better. Thanks for pointing this out.

src/change_stream.ts
```diff
@@ -259,9 +265,8 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
     } else if (parent instanceof MongoClient) {
       this.type = CHANGE_DOMAIN_TYPES.CLUSTER;
     } else {
-      // TODO(NODE-3404): Replace this with MongoChangeStreamError
       throw new MongoDriverError(
-        'Parent provided to ChangeStream constructor must an instance of Collection, Db, or MongoClient'
+        'parent provided to ChangeStream constructor is not an instance of Collection, Db, or MongoClient'
```
W-A-James, line 269:
```suggestion
        'Parent provided to ChangeStream constructor must be an instance of Collection, Db, or MongoClient'
```

test/functional/gridfs_stream.test.js
```diff
@@ -473,6 +473,7 @@ describe('GridFS Stream', function () {
                     // Fail if user tries to abort an aborted stream
                     uploadStream.abort().then(null, function (error) {
                       expect(error.toString()).to.equal(
+                        // TODO(NODE-3405): Replace with MongoStreamClosedError
```
W-A-James, line 476: You can remove these if they've been resolved in NODE-3405 and this isn't depending on it

src/change_stream.ts
```diff
@@ -634,6 +638,7 @@ function waitForTopologyConnected(
     }
 
     if (calculateDurationInMs(start) > timeout) {
+      // TODO: Replace with MongoNetworkTimeoutError
```
W-A-James, line 641: May want to flag with with a ticket number so we don't lose track of it or just resolve it here

W-A-James (changes requested):

W-A-James (approved):

src/change_stream.ts
```diff
@@ -259,9 +265,8 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
     } else if (parent instanceof MongoClient) {
       this.type = CHANGE_DOMAIN_TYPES.CLUSTER;
     } else {
-      // TODO(NODE-3404): Replace this with MongoChangeStreamError
       throw new MongoDriverError(
-        'Parent provided to ChangeStream constructor must an instance of Collection, Db, or MongoClient'
+        'Parent provided to ChangeStream constructor must be an instance of Collection, Db, or MongoClient'
```
dariakp, line 269: should the error have been updated to MongoChangeStreamError?

src/connection_string.ts
```diff
@@ -57,11 +57,11 @@ function matchesParentDomain(srvAddress: string, parentDomain: string): boolean
  */
 export function resolveSRVRecord(options: MongoOptions, callback: Callback<HostAddress[]>): void {
   if (typeof options.srvHost !== 'string') {
-    return callback(new MongoParseError('Cannot resolve empty srv string'));
+    return callback(new MongoURIError('Option "srvHost" must not be empty'));
```
dariakp, line 60: So, the mongo invalid argument error extends the mongo api error, but the mongo uri error extends the mongo runtime error, even though both are effectively validating arguments. Can we think of a better way to set up this hierarchy? It feels like the uri errors should subclass from the api errors, but maybe there's something better

src/connection_string.ts
```diff
@@ -569,7 +567,7 @@ export const OPTIONS = {
           if (['none', 'snappy', 'zlib'].includes(String(c))) {
             compressionList.add(String(c));
           } else {
-            throw new MongoParseError(`${c} is not a valid compression mechanism`);
+            throw new MongoCompressionError(`${c} is not a valid compression mechanism`);
```
dariakp, line 570: this particular compression error is an api error

src/db.ts
```diff
@@ -749,7 +749,6 @@ function validateDatabaseName(databaseName: string) {
   const invalidChars = [' ', '.', '$', '/', '\\'];
   for (let i = 0; i < invalidChars.length; i++) {
     if (databaseName.indexOf(invalidChars[i]) !== -1)
-      // TODO(NODE-3405): Change this out for a child of MongoParseError
       throw new MongoDriverError(
```
dariakp, line 752: why did we remove the todo without updating the error?

src/operations/execute_operation.ts
```diff
@@ -87,7 +87,6 @@ export function executeOperation<
         owner = Symbol();
         session = topology.startSession({ owner, explicit: false });
       } else if (session.hasEnded) {
-        // TODO(NODE-3405): Change this out for MongoExpiredSessionError
```
dariakp, line 90: the TODOs have been removed without updating the underlying errors

src/sdam/server.ts
```diff
@@ -270,7 +270,6 @@ export class Server extends TypedEventEmitter<ServerEvents> {
     }
 
     if (this.s.state === STATE_CLOSING || this.s.state === STATE_CLOSED) {
-      // TODO(NODE-3405): Change this out for MongoServerClosedError
```
dariakp, line 273: where is error update?

src/utils.ts
```diff
@@ -255,7 +255,6 @@ export function executeLegacyOperation(
       const optionsIndex = args.length - 2;
       args[optionsIndex] = Object.assign({}, args[optionsIndex], { session: session });
     } else if (opOptions.session && opOptions.session.hasEnded) {
-      // TODO(NODE-3405): Replace this with MongoExpiredSessionError
```
dariakp, line 258: missing actual error update

test/functional/gridfs_stream.test.js
dariakp: I think maybe this PR needs to get rebased after things are merged in so that we can keep pulling in latest changes and don't miss anything, let's talk about the order in which we do want to resolve the PRs in stand up

test/functional/unified-spec-runner/entities.ts
```diff
@@ -391,7 +392,7 @@ export class EntitiesMap<E = Entity> extends Map<string, E> {
 
         map.set(entity.bucket.id, new GridFSBucket(db, options));
       } else if ('stream' in entity) {
-        throw new Error(`Unsupported Entity ${JSON.stringify(entity)}`);
+        throw new MongoStreamError(`Unsupported Entity ${JSON.stringify(entity)}`);
```
dariakp, line 395: we don't need to add our custom errors to the test runner

test/functional/unified-spec-runner/operations.ts
```diff
@@ -202,7 +203,7 @@ operations.set('createChangeStream', async ({ entities, operation }) => {
 
   return new Promise((resolve, reject) => {
     const timeout = setTimeout(() => {
-      reject(new Error('Change stream never started'));
+      reject(new MongoChangeStreamError('Change stream never started'));
```
dariakp, line 206: this is fine to leave as a generic error

test/manual/ldap.test.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 const { MongoClient } = require('../../src');
 const { expect } = require('chai');
+const { MongoURIError } = require('../../src/error');
 
 describe('LDAP', function () {
   if (process.env.MONGODB_URI == null) {
-    throw new Error(`skipping SSL tests, MONGODB_URI environment variable is not defined`);
+    throw new MongoURIError(`skipping SSL tests, MONGODB_URI environment variable is not defined`);
```
dariakp, line 8: no need for custom errors in tests unless the tests are intended to mock a specific sort of error

dariakp (changes requested):

src/cmap/message_stream.ts
```diff
@@ -191,7 +191,7 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>)
 
     if (messageBody.length !== messageHeader.length) {
       callback(
-        new MongoDriverError(
+        new MongoDecompressionError(
           'Decompressing a compressed message from the server failed. The message is corrupt.'
```
nbbeeken, line 195: Seems like we could improve the message a bit here, a MongoDecompressError implies the first sentence maybe its worth being specific, we can workshop but something like: 'Message header size does not match message body size'

src/gridfs/download.ts
```diff
@@ -237,28 +243,24 @@ function doRead(stream: GridFSBucketReadStream): void {
...
       }
 
-      errmsg =
-        'ChunkIsWrongSize: Got unexpected length: ' +
-        buf.byteLength +
-        ', expected: ' +
-        expectedLength;
-      return __handleError(stream, new MongoDriverError(errmsg));
+      errmsg = `ChunkIsWrongSize: Got unexpected length: ${buf.byteLength}, expected: ${expectedLength}`;
+      return __handleError(stream, new MongoGridFSChunkError(errmsg));
```
nbbeeken, line 263: __handleError is a oneliner (maybe it contained more at some point) can we just inline it everywhere?

```js
      return stream.emit(
        GridFSBucketReadStream.ERROR,
        new MongoGridFSChunkError(
          `ChunkIsWrongSize: Got unexpected length: ${buf.byteLength}, expected: ${expectedLength}`
        )
      );
```

src/gridfs/upload.ts
```diff
@@ -143,7 +149,7 @@ export class GridFSBucketWriteStream extends Writable {
   abort(callback: Callback<void>): void;
   abort(callback?: Callback<void>): Promise<void> | void {
     const Promise = PromiseProvider.get();
-    let error: MongoDriverError;
+    let error: MongoGridFSStreamError;
```
nbbeeken, line 152: Can we refactor this to use `maybePromise`?

src/operations/execute_operation.ts
```diff
@@ -193,7 +191,6 @@ function executeWithServerSelection(
     session.inTransaction()
   ) {
     callback(
-      // TODO(NODE-3405): Change this out for MongoTransactionError
       new MongoDriverError(
```
nbbeeken, line 194: Are we going to make the changes that the comments mentioned?

src/sdam/server.ts
```diff
@@ -270,7 +270,6 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
       callback(new MongoDriverError('Server is closed'));
```
nbbeeken, line 273: SeverClosedErr?

src/gridfs/download.ts
```diff
@@ -1,14 +1,20 @@
...
+import type { Collection } from '../collection';
+import type { FindCursor } from '../cursor/find_cursor';
+import {
+  AnyError,
+  MongoDriverError,
+  MongoInvalidArgumentError,
+  MongoGridFSChunkError,
+  MongoGridFSStreamError
+} from '../error';
 import type { FindOptions } from '../operations/find';
```
nbbeeken, line 13: There's import changes here that aren't needed, I think you import sorting turned on in vscode? This is something we want to do as a lint fix down the line if you don't mind reverting these.

src/gridfs/download.ts
```diff
@@ -1,14 +1,20 @@
+import type { ObjectId } from 'bson';
```
nbbeeken, line 1: This should be `'../bson'` must've slipped through I'll look into linting for this.

nbbeeken (changes requested):

src/utils.ts
andymina: This TODO (along with previous ones mentioned [here](https://github.com/mongodb/node-mongodb-native/pull/2912#discussion_r676926837), [here](https://github.com/mongodb/node-mongodb-native/pull/2912#discussion_r676927348), and [here](https://github.com/mongodb/node-mongodb-native/pull/2912#discussion_r676927348)) have all been added in the [NODE-3405](https://github.com/mongodb/node-mongodb-native/pull/2913) so I removed them as I encountered them both on this PR and the NODE-3405 PR. If I messed up the order/practice of removing the TODO, I don't mind putting them back in.

src/db.ts
W-A-James: We went through and removed the TODOs here since they have already been replaced in NODE-3405

src/sdam/server.ts
andymina: This TODO and this one [here](https://github.com/mongodb/node-mongodb-native/pull/2912#discussion_r676945514) have both been changed in [NODE-3405](https://github.com/mongodb/node-mongodb-native/pull/2913) so I removed the TODOs here.

src/cmap/message_stream.ts
andymina: How does `'Message body and message header must be the same length'` sound?

src/gridfs/upload.ts
andymina: In hindsight, I think this whole thing is out of scope for this ticket. I `MongoStreamClosedError`, covered in [NODE-3405](https://github.com/mongodb/node-mongodb-native/pull/2913), better covers the errors thrown here. I'll leave a TODO comment for that PR and revert this change.

src/connection_string.ts
andymina: Recapping what we spoke about on Zoom:

For the time being, use the predefined clear-cut errors where appropriate. Otherwise default to MongoAPI and MongoRuntime to differentiate between "user's fault" and "driver's fault" errors. We'll sort out instances of MongoAPI and MongoRuntime in another PR.

src/connection_string.ts
andymina: In hindsight, I think this may be better suited as InvalidArgument because the it seems like the option provided _must_ be in the array. Unless the driver supports compression mechanisms other than snappy and zlib and I'm just missing something.

src/change_stream.ts
```diff
@@ -547,17 +551,17 @@ const CHANGE_STREAM_EVENTS = [
 
 function setIsEmitter<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'iterator') {
-    throw new MongoDriverError(
-      'Cannot use ChangeStream as an EventEmitter after using as an iterator'
+    throw new MongoChangeStreamError(
+      'ChangeStream cannot be used as an EventEmitter after being used as an iterator'
```
W-A-James, line 555: On further consideration, the error thrown here and in `setIsIterator` should both be `MongoAPIError` instances.

While these functions are internal, the errors result directly from the user violating the change stream API. When a change stream is created, it should either be used as an iterator _or_ as an event emitter (NODE-1797). The user has to make the decision to pick which API they want to use.

W-A-James (changes requested):

src/connection_string.ts
```diff
@@ -57,11 +57,11 @@ function matchesParentDomain(srvAddress: string, parentDomain: string): boolean
  */
 export function resolveSRVRecord(options: MongoOptions, callback: Callback<HostAddress[]>): void {
   if (typeof options.srvHost !== 'string') {
-    return callback(new MongoParseError('Cannot resolve empty srv string'));
+    return callback(new MongoAPIError('Option "srvHost" must not be empty'));
   }
 
   if (options.srvHost.split('.').length < 3) {
-    return callback(new MongoParseError('URI does not have hostname, domain name and tld'));
+    return callback(new MongoURIError('URI must include hostname, domain name, and tld'));
```
W-A-James, line 64: Replace this with MongoAPIError and mark for replacement with MongoConnectionStringError in NODE-3484.

src/db.ts
```diff
@@ -748,7 +749,6 @@ function validateDatabaseName(databaseName: string) {
   const invalidChars = [' ', '.', '$', '/', '\\'];
   for (let i = 0; i < invalidChars.length; i++) {
     if (databaseName.indexOf(invalidChars[i]) !== -1)
-      // TODO(NODE-3405): Change this out for a child of MongoParseError
```
W-A-James, line 751: Tag NODE-3484 here

W-A-James (changes requested):

src/cmap/wire_protocol/compression.ts
```diff
@@ -53,10 +53,8 @@ export function compress(
       zlib.deflate(dataToBeCompressed, zlibOptions, callback as zlib.CompressCallback);
       break;
     default:
-      throw new MongoDriverError(
-        'Attempt to compress message using unknown compressor "' +
-          self.options.agreedCompressor +
-          '".'
+      throw new MongoCompressionError(
```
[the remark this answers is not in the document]
andymina: Ooh, good catch. Replacing this now.

src/db.ts
andymina: NODE-3484 should be tagged at the top of this function since the whole thing has to be moved into MongoDBNamespace. Should I tag it here instead?

W-A-James (approved):

src/db.ts
W-A-James: If it's already tagged, then ignore this.

src/connection_string.ts
```diff
@@ -569,7 +568,9 @@ export const OPTIONS = {
           if (['none', 'snappy', 'zlib'].includes(String(c))) {
             compressionList.add(String(c));
           } else {
-            throw new MongoParseError(`${c} is not a valid compression mechanism`);
+            throw new MongoInvalidArgumentError(
+              `${c} is not a valid compression mechanism. Must be 'none', 'snappy', or 'zlib'. `
```
nbbeeken, line 572: We can use the enumToString helper in utils.ts here.

src/connection_string.ts
```diff
@@ -569,7 +568,9 @@ export const OPTIONS = {
           if (['none', 'snappy', 'zlib'].includes(String(c))) {
```
nbbeeken, line 568: related to my other comment, this should use the enum instead of this hardcoded list, lets take the the opportunity to make this a bit better

test/functional/change_stream.test.js
```diff
@@ -874,7 +874,7 @@ describe('Change Streams', function () {
           .next()
           .then(function () {
             // We should never execute this line because calling changeStream.next() should throw an error
-            throw new Error(
+            throw new MongoChangeStreamError(
```
nbbeeken, line 877: This should be left as a plain Error, no?

test/functional/change_stream.test.js
```diff
@@ -100,7 +100,7 @@ function triggerResumableError(changeStream, delay, onClose) {
  */
 function waitForStarted(changeStream, callback) {
   const timeout = setTimeout(() => {
-    throw new Error('Change stream never started');
+    throw new MongoChangeStreamError('Change stream never started');
```
nbbeeken, line 103: This can be left as a plain error

test/functional/change_stream.test.js
```diff
@@ -2007,7 +2007,7 @@ describe('Change Streams', function () {
           if (counter === 2) {
             changeStream.close(close);
           } else if (counter >= 3) {
-            close(new Error('should not have received more than 2 events'));
+            close(new MongoChangeStreamError('should not have received more than 2 events'));
```
nbbeeken, line 2010: This can also be a plain error, the errors here are all just making a test fail because they've reached a bad state, these should all probably be `expect.fail('...')` instead.

nbbeeken (changes requested):
Some small cleanup reqs :)

test/functional/change_stream.test.js
andymina: Are we not changing errors in tests too? Happy to make the change, but not too clear on why 😅

nbbeeken: [related](https://github.com/mongodb/node-mongodb-native/pull/2912#discussion_r682825112)
The usage of errors here is to fail the test when an unreachable spot is reached. They aren't being used to test "does the driver return such and such error in some given state" so there isn't a reason to change these. But also it is a little confusing to make them a specific error type because it implies there is a purpose for this error but when reading the test there isn't code that will catch and handle the error.

dariakp: The errors thrown by the tests are only for use by the tests themselves, they are the equivalent of an `expect.fail`. The errors here will never be user facing, so there is no need to specially type them, since they aren't coming from the driver, but rather from our own test setup.

The only time you have to update the types of the errors in the tests is when we are *checking* an error that *is* thrown by the driver itself.

andymina: Ahhh, makes sense. That's a lot clearer now, thank you guys! Will keep this in mind for future error PRs :)

src/connection_string.ts
```diff
@@ -613,11 +619,13 @@ export const OPTIONS = {
       const compressionList = new Set();
       for (const compVal of values as string[]) {
         for (const c of compVal.split(',')) {
-          if (['none', 'snappy', 'zlib'].includes(String(c))) {
+          if (Object.values(VALID_COMPRESSION_MECHS).includes(String(c))) {
```
nbbeeken, line 622:
```suggestion
          if (Object.keys(Compressor).includes(String(c))) {
```

We already have an enum `Compressor` lets use that here

src/connection_string.ts
```diff
@@ -613,11 +619,13 @@ export const OPTIONS = {
...
             compressionList.add(String(c));
           } else {
             throw new MongoInvalidArgumentError(
-              `${c} is not a valid compression mechanism. Must be 'none', 'snappy', or 'zlib'. `
+              `${c} is not a valid compression mechanism. Must be one of: ${enumToString(
```
nbbeeken, line 626: I guess that would just make this `${Object.keys(Compressor)}`

nbbeeken (changes requested):

src/connection_string.ts
andymina: Oops, didn't realize there was an existing enum already 😅

src/change_stream.ts
dariakp: The TODO in L551 references NODE-3484, which doesn't make sense for this ticket, please update to TODO-3485

src/cmap/connect.ts
```diff
@@ -284,7 +285,8 @@ function parseConnectOptions(options: ConnectionOptions): SocketConnectOpts {
   } else {
     // This should never happen since we set up HostAddresses
     // But if we don't throw here the socket could hang until timeout
-    throw new MongoDriverError(`Unexpected HostAddress ${JSON.stringify(hostAddress)}`);
+    // TODO(NODE-3484): Replace with a child of MongoParseError
```
dariakp, line 288: I don't think we need this TODO

test/functional/unified-spec-runner/operations.ts
dariakp: Still need to remove the custom error and the import here

test/manual/ldap.test.js
dariakp: Still needs to be updated to use the regular Error class

src/connection_string.ts
```diff
@@ -77,14 +79,12 @@ export function resolveSRVRecord(options: MongoOptions, callback: Callback<HostA
     if (err) return callback(err);
 
     if (addresses.length === 0) {
-      return callback(new MongoParseError('No addresses found at host'));
+      return callback(new MongoURIError('No addresses found at host'));
```
dariakp, line 82: So, the URI error now extends a runtime error, but this is misleading because its description still says it's the user's fault; moreover, if we are intending to introduce a MongoConnectionStringError, having both of these errors is going to be confusing; I think the best thing to do might be to remove the MongoURIError entirely, and just replace the errors here with MongoAPIError or MongoRuntimeError as appropriate (note, in L289 and L404, it is an API error)

src/db.ts
dariakp: Please leave the TODO in this PR, the PR that addresses the comment should be the one to remove the line

dariakp (changes requested):

src/db.ts
andymina: This TODO got moved up outside of the function and retagged with the right ticket. Refactoring `validateDatabaseName` falls under NODE-3484.

andymina: The TODO here got updated to point to the correct ticket, [NODE-3484](https://jira.mongodb.org/browse/NODE-3484), and moved to the top of this function since the ticket calls for the refactoring of the entire thing.

src/connection_string.ts
andymina: Oops, complete oversight on my part! Replacing this with MongoAPIError and removing the class definition. So sorry!!

test/functional/change_stream.test.js
```diff
@@ -2007,7 +2007,7 @@ describe('Change Streams', function () {
           if (counter === 2) {
             changeStream.close(close);
           } else if (counter >= 3) {
-            close(new Error('should not have received more than 2 events'));
+            expect.fail('Should not have received more than 2 events');
```
dariakp, line 2010: The original close function here served an important purpose: to prevent done from being called more than once; an expect.fail would throw an error, which would call done in whatever context is currently running, potentially causing unrelated tests to fail. tl;dr: passing the error explicitly to a wrapped done function (the `close` function in this context), the way it was written originally, is the safest way to handle event listener errors in tests

dariakp (changes requested):
Other than this small but important detail, LGTM

nbbeeken (approved):
once the fix daria mentions is in then this LGTM

dariakp (approved):
LGTM 🎉

PR #2913, opened 2021-07-20
refactor(NODE-3405): implement MongoRuntimeError children

andymina:
## Description
Replace instances of MongoDriverError with the appropriate children of MongoRuntimeError, listed below, as described in docs/errors.md.

Children of MongoRuntimeError used:
- MongoTransactionError
- MongoExpiredSessionError
- MongoKerberosError
- MongoResourceClosedError
- MongoServerClosedError
- MongoStreamClosedError
- MongoTopologyClosedError

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -58,7 +58,7 @@ const CHANGE_DOMAIN_TYPES = {
 const NO_RESUME_TOKEN_ERROR =
   'A change stream document has been received that lacks a resume token (_id).';
 const NO_CURSOR_ERROR = 'ChangeStream has no cursor';
-const CHANGESTREAM_CLOSED_ERROR = 'ChangeStream has been closed';
+const CHANGESTREAM_CLOSED_ERROR = 'ChangeStream is closed';
 
 /** @public */
 export interface ResumeOptions {
@@ -558,8 +558,9 @@ function setIsEmitter<TSchema>(changeStream: ChangeStream<TSchema>): void {
 
 function setIsIterator<TSchema>(changeStream: ChangeStream<TSchema>): void {
   if (changeStream[kMode] === 'emitter') {
+    // TODO(NODE-3485): Replace with MongoChangeStreamModeError
     throw new MongoAPIError(
-      'ChangeStream cannot be used as an EventEmitter after being used as an iterator'
+      'ChangeStream cannot be used as an iterator after being used as an EventEmitter'
     );
   }
   changeStream[kMode] = 'iterator';
@@ -683,7 +684,7 @@ function processNewChange<TSchema>(
   callback?: Callback<ChangeStreamDocument<TSchema>>
 ) {
   if (changeStream[kClosed]) {
-    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+    // TODO(NODE-3485): Replace with MongoChangeStreamClosedError
     if (callback) callback(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -5,7 +5,8 @@ import {
   isResumableError,
   MongoDriverError,
   MongoAPIError,
-  MongoChangeStreamError
+  MongoChangeStreamError,
+  MongoRuntimeError
 } from './error';
 import { AggregateOperation, AggregateOptions } from './operations/aggregate';
 import {
@@ -691,8 +692,8 @@ function processNewChange<TSchema>(
 
   // a null change means the cursor has been notified, implicitly closing the change stream
   if (change == null) {
-    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
-    return closeWithError(changeStream, new MongoAPIError(CHANGESTREAM_CLOSED_ERROR), callback);
+    // TODO(NODE-3485): Replace with MongoChangeStreamClosedError
+    return closeWithError(changeStream, new MongoRuntimeError(CHANGESTREAM_CLOSED_ERROR), callback);
   }
 
   if (change && !change._id) {
@@ -724,7 +725,7 @@ function processError<TSchema>(
 
   // If the change stream has been closed explicitly, do not process error.
   if (changeStream[kClosed]) {
-    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+    // TODO(NODE-3485): Replace with MongoChangeStreamClosedError
     if (callback) callback(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
@@ -785,7 +786,7 @@ function processError<TSchema>(
  */
 function getCursor<T>(changeStream: ChangeStream<T>, callback: Callback<ChangeStreamCursor<T>>) {
   if (changeStream[kClosed]) {
-    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+    // TODO(NODE-3485): Replace with MongoChangeStreamClosedError
     callback(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
@@ -811,7 +812,7 @@ function processResumeQueue<TSchema>(changeStream: ChangeStream<TSchema>, err?: 
     const request = changeStream[kResumeQueue].pop();
     if (!err) {
       if (changeStream[kClosed]) {
-        // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+        // TODO(NODE-3485): Replace with MongoChangeStreamClosedError
         request(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
         return;
       }
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -58,7 +58,7 @@ const CHANGE_DOMAIN_TYPES = {
 const NO_RESUME_TOKEN_ERROR =
   'A change stream document has been received that lacks a resume token (_id).';
 const NO_CURSOR_ERROR = 'ChangeStream has no cursor';
-const CHANGESTREAM_CLOSED_ERROR = 'ChangeStream is closed';
+const CHANGESTREAM_CLOSED_ERROR = 'ChangeStream has been closed';
 
 /** @public */
 export interface ResumeOptions {
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -683,15 +683,15 @@ function processNewChange<TSchema>(
   callback?: Callback<ChangeStreamDocument<TSchema>>
 ) {
   if (changeStream[kClosed]) {
-    // TODO(NODE-3405): Replace with MongoStreamClosedError
-    if (callback) callback(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
+    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+    if (callback) callback(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
 
   // a null change means the cursor has been notified, implicitly closing the change stream
   if (change == null) {
-    // TODO(NODE-3405): Replace with MongoStreamClosedError
-    return closeWithError(changeStream, new MongoDriverError(CHANGESTREAM_CLOSED_ERROR), callback);
+    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+    return closeWithError(changeStream, new MongoAPIError(CHANGESTREAM_CLOSED_ERROR), callback);
   }
 
   if (change && !change._id) {
@@ -723,8 +723,8 @@ function processError<TSchema>(
 
   // If the change stream has been closed explicitly, do not process error.
   if (changeStream[kClosed]) {
-    // TODO(NODE-3405): Replace with MongoStreamClosedError
-    if (callback) callback(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
+    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+    if (callback) callback(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
 
@@ -784,8 +784,8 @@ function processError<TSchema>(
  */
 function getCursor<T>(changeStream: ChangeStream<T>, callback: Callback<ChangeStreamCursor<T>>) {
   if (changeStream[kClosed]) {
-    // TODO(NODE-3405): Replace with MongoStreamClosedError
-    callback(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
+    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+    callback(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
 
@@ -810,8 +810,8 @@ function processResumeQueue<TSchema>(changeStream: ChangeStream<TSchema>, err?: 
     const request = changeStream[kResumeQueue].pop();
     if (!err) {
       if (changeStream[kClosed]) {
-        // TODO(NODE-3405): Replace with MongoStreamClosedError
-        request(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
+        // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+        request(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
         return;
       }
       if (!changeStream.cursor) {
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -58,7 +58,7 @@ const CHANGE_DOMAIN_TYPES = {
 const NO_RESUME_TOKEN_ERROR =
   'A change stream document has been received that lacks a resume token (_id).';
 const NO_CURSOR_ERROR = 'ChangeStream has no cursor';
-const CHANGESTREAM_CLOSED_ERROR = 'ChangeStream has been closed';
+const CHANGESTREAM_CLOSED_ERROR = 'ChangeStream is closed';
 
 /** @public */
 export interface ResumeOptions {
--- a/src/cmap/commands.ts
+++ b/src/cmap/commands.ts
@@ -85,7 +85,7 @@ export class Query {
     // Validate that we are not passing 0x00 in the collection name
     if (ns.indexOf('\x00') !== -1) {
       // TODO(NODE-3483): Use MongoNamespace static method
-      throw new MongoDriverError('namespace cannot contain a null character');
+      throw new MongoDriverError('Namespace cannot contain a null character');
     }
 
     // Basic options
--- a/src/cmap/commands.ts
+++ b/src/cmap/commands.ts
@@ -84,8 +84,8 @@ export class Query {
 
     // Validate that we are not passing 0x00 in the collection name
     if (ns.indexOf('\x00') !== -1) {
-      // TODO(NODE-3483): Replace with MongoCommandError
-      throw new MongoDriverError('Namespace cannot contain a null character');
+      // TODO(NODE-3483): Use MongoNamespace static method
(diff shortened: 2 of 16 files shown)


src/change_stream.ts
```diff
@@ -1,32 +1,38 @@
...
+import type { ClientSession } from './sessions';
+import {
+  calculateDurationInMs,
+  Callback,
+  getTopology,
+  maxWireVersion,
+  maybePromise,
+  MongoDBNamespace,
+  now
+} from './utils';
```
W-A-James, line 36: Just out of curiosity, why the rearrangement of the imports?

src/utils.ts
```diff
@@ -54,26 +54,31 @@ export function getSingleProperty(
  */
 export function checkCollectionName(collectionName: string): void {
   if ('string' !== typeof collectionName) {
+    // TODO(NODE-3405): Use MongoNamespace static method
```
W-A-James, line 57: We should probably make a ticket for this so we don't forget what exactly this refers to

src/cmap/connection_pool.ts
```diff
@@ -320,7 +320,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
           clearTimeout(waitQueueMember.timer);
         }
         if (!waitQueueMember[kCancelled]) {
-          waitQueueMember.callback(new MongoDriverError('connection pool closed'));
+          waitQueueMember.callback(new MongoResourceClosedError('Connection pool closed'));
```
W-A-James, line 323: Should we just go ahead and create a `MongoConnectionPoolClosedError`? It feels somewhat out of place having the more generic form of this error directly instantiated.

src/gridfs/upload.ts
```diff
@@ -1,14 +1,20 @@
...
+import { PromiseProvider } from '../promise_provider';
 import type { Callback } from '../utils';
-import type { Document } from '../bson';
-import type { GridFSBucket } from './index';
-import type { GridFSFile } from './download';
 import type { WriteConcernOptions } from '../write_concern';
-import type { Collection } from '../collection';
+import { WriteConcern } from './../write_concern';
+import type { GridFSFile } from './download';
+import type { GridFSBucket } from './index';
```
W-A-James, line 17: In the same vein as before, just curious as to what the intention was with reordering the imports.

src/sdam/server.ts
```diff
@@ -62,6 +63,8 @@ const stateTransition = makeStateMachine({
   [STATE_CLOSING]: [STATE_CLOSING, STATE_CLOSED]
 });
 
+const SERVER_CLOSED_ERROR = 'Server is closed';
```
W-A-James, line 66: In general, I think that when we have a resource that has been closed, the message should be of the form 

```javascript
'<resource type> has been closed'
```
That's a personal preference, but I think having it be more consistent would be better for error message readability. I'll flag other instances of this.

src/sdam/topology.ts
```diff
@@ -83,6 +83,8 @@ const stateTransition = makeStateMachine({
   [STATE_CLOSING]: [STATE_CLOSING, STATE_CLOSED]
 });
 
+const TOPOLOGY_CLOSED_ERROR = 'Topology is closed';
```
W-A-James, line 86:
```suggestion
const TOPOLOGY_CLOSED_ERROR = 'Topology has been closed';
```

src/sdam/topology.ts
```diff
@@ -956,7 +958,7 @@ function processWaitQueue(topology: Topology) {
   if (topology.s.state === STATE_CLOSED) {
     drainWaitQueue(
       topology[kWaitQueue],
-      new MongoDriverError('Topology is closed, please connect')
+      new MongoTopologyClosedError(TOPOLOGY_CLOSED_ERROR + ', please connect')
```
W-A-James, line 961: Change to use a template string

src/change_stream.ts
```diff
@@ -678,13 +684,17 @@ function processNewChange<TSchema>(
   callback?: Callback<ChangeStreamDocument<TSchema>>
 ) {
   if (changeStream[kClosed]) {
-    if (callback) callback(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
+    if (callback) callback(new MongoStreamClosedError(CHANGESTREAM_CLOSED_ERROR));
```
W-A-James, line 687: Tagging this here, but line 60 should read

```javascript
const CHANGESTREAM_CLOSED_ERROR = 'ChangeStream has been closed';
```

W-A-James (changes requested):

src/change_stream.ts
andymina: ESLint felt very strongly about having them in this order for some reason, lol.

src/sdam/server.ts
andymina: I'm a little hesitant to use the past tense in this error message because it might imply that the server has been closed as a result of this operation rather than conveying the current state of the server. Happy to see what everyone else thinks; maybe I'm being a little _too_ semantic.

src/cmap/connection_pool.ts
andymina: Sounds good! In general, I think it might be worth to make specific ClosedErrors for things that users will commonly run into (servers, streams, etc) and leaving `MongoResourceClosedError` as the general use case.

W-A-James: In that case, this might be worth discussing with the team to get a better idea of the more common cases

src/gridfs/upload.ts
andymina: The almighty ESLint reordered things here too.

src/gridfs/upload.ts
```diff
@@ -556,7 +562,7 @@ function writeRemnant(stream: GridFSBucketWriteStream, callback?: Callback): boo
 function checkAborted(stream: GridFSBucketWriteStream, callback?: Callback<void>): boolean {
   if (stream.state.aborted) {
     if (typeof callback === 'function') {
-      callback(new MongoDriverError('this stream has been aborted'));
+      callback(new MongoStreamClosedError('Stream has been aborted'));
```
W-A-James, line 565: In the same vein as [this comment on NODE-3421](https://github.com/mongodb/node-mongodb-native/pull/2910#pullrequestreview-712206125) we may want to consider adding a default message for MongoStreamClosedError with an optional message parameter for cases where more information is needed

src/cmap/commands.ts
```diff
@@ -84,8 +84,8 @@ export class Query {
 
     // Validate that we are not passing 0x00 in the collection name
     if (ns.indexOf('\x00') !== -1) {
-      // TODO(NODE-3483): Replace with MongoCommandError
-      throw new MongoDriverError('Namespace cannot contain a null character');
+      // TODO(NODE-3483): Use MongoNamespace static method
+      throw new MongoDriverError('namespace cannot contain a null character');
```
W-A-James, line 88: Why the change to start with a common letter here?

andymina: Oops, this got messed up in the rebase.

andymina:
> @andymina There is a TODO that points to this ticket in `src/db.ts: 752`. Feel free to point that to [NODE-3484](https://jira.mongodb.org/browse/NODE-3484) instead

I removed the TODOs here because I realized this is the `validateDatabaseName()` mentioned in [NODE-3483](https://jira.mongodb.org/browse/NODE-3483). I left a TODO pointing to NODE-3483 for this entire function.

W-A-James (approved):

src/change_stream.ts
```diff
@@ -1,5 +1,11 @@
 import Denque = require('denque');
-import { MongoError, AnyError, isResumableError, MongoDriverError } from './error';
+import {
+  MongoError,
+  AnyError,
+  isResumableError,
+  MongoDriverError,
+  MongoStreamClosedError
```
dariakp, line 7: As a general question to the team, do we want to conflate generic streams with change streams in our error reporting? Our change streams aren't really node streams in the sense that they don't extend from any of the base node streams. I could go either way, but my thought is that there could be some merit in differentiating "steam closed" vs "change stream closed" @nbbeeken @emadum Thoughts?

src/operations/execute_operation.ts
```diff
@@ -87,8 +89,7 @@ export function executeOperation<
         owner = Symbol();
         session = topology.startSession({ owner, explicit: false });
       } else if (session.hasEnded) {
-        // TODO(NODE-3405): Change this out for MongoExpiredSessionError
-        return cb(new MongoDriverError('Use of expired sessions is not permitted'));
+        return cb(new MongoExpiredSessionError('Use of expired sessions is not permitted'));
       } else if (session.snapshotEnabled && !topology.capabilities.supportsSnapshotReads) {
         return cb(new MongoDriverError('Snapshot reads require MongoDB 5.0 or later'));
```
dariakp, line 94: I think this might have been missed as an api error because snapshots were implemented while the other PR was being implemented, can we add the correct compatibility error since we are touching this file?

src/sessions.ts
```diff
@@ -286,12 +288,12 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    */
   startTransaction(options?: TransactionOptions): void {
     if (this[kSnapshotEnabled]) {
-      throw new MongoDriverError('Transactions are not allowed with snapshot sessions');
+      throw new MongoTransactionError('Transactions are not allowed with snapshot sessions');
```
dariakp, line 291: this is more of an API error - probably the compatibility one, because in order to get here, the user needs to intentionally start a snapshot session and then try to start a transaction on it

src/sessions.ts
```diff
@@ -286,12 +288,12 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
     }
 
     assertAlive(this);
     if (this.inTransaction()) {
-      throw new MongoDriverError('Transaction already in progress');
+      throw new MongoTransactionError('Transaction already in progress');
```
dariakp, line 296: This is also kind of an api error; we may need to rethink how we draw the line between runtime and api errors, specifically with respect to transactions: like, it's true that the driver is getting to an invalid state, but it's not "unexpected" because it's still the user making a choice; maybe the correct thing to do is subclass the transaction errors from API errors, or to subclass runtime errors into incorrect usage ones vs unexpected driver state, and then subclass these transaction errors according to the cause of the error

dariakp (changes requested):
I didn't leave a comment on every line regarding transaction errors, but see my comment regarding reconsidering how we subclass these so that we can differentiate between incorrect usage from the user's side vs unexpected driver state

src/change_stream.ts
nbbeeken: Seems worthy of differentiating, I would imagine we'd only need use of MongoChangeStreamError, but there may be instances of Streams related errors, unless they stop something from continuing in our code I would say node stream errors should be bubbled up as is.

andymina: The initial goal we had was to use `MongoStreamClosedError` between GridFSStream and ChangeStream to prevent creating two similar errors. If it's worth differentiating we could just split them down into `GridFSStreamClosed` and `ChangeStreamClosed`.

src/sessions.ts
```diff
@@ -286,12 +289,12 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
   startTransaction(options?: TransactionOptions): void {
     if (this[kSnapshotEnabled]) {
-      throw new MongoDriverError('Transactions are not allowed with snapshot sessions');
+      throw new MongoCompatibilityError('Transactions are not allowed with snapshot sessions');
     }
 
     assertAlive(this);
     if (this.inTransaction()) {
-      throw new MongoDriverError('Transaction already in progress');
+      throw new MongoAPIError('Transaction already in progress');
```
W-A-James, line 297: Since we are moving MongoTransactionError to be child of MongoAPIError, should we change this to MongoTransactionError and just go ahead and move MongoTransactionError to be a child of MongoAPIError in this PR?

W-A-James (changes requested):

src/cmap/connection_pool.ts
```diff
@@ -388,7 +388,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
           clearTimeout(waitQueueMember.timer);
         }
         if (!waitQueueMember[kCancelled]) {
-          waitQueueMember.callback(new MongoDriverError('connection pool closed'));
+          waitQueueMember.callback(new MongoAPIError('Connection pool closed'));
```
W-A-James, line 391: Tag this with a comment for replacement in NODE-3485

src/sessions.ts
W-A-James: If not, we should also mark this for replacement in NODE-3485

W-A-James (approved):
LGTM after addressing the remaining comments

W-A-James (approved):

src/gridfs/upload.ts
```diff
@@ -1,20 +1,13 @@
...
+import { MongoError, AnyError, MongoAPIError, MONGODB_ERROR_CODES } from '../error';
 import { WriteConcern } from './../write_concern';
-import type { GridFSFile } from './download';
+import { ObjectId } from '../bson';
+import { Callback, maybePromise } from '../utils';
+import type { Document } from '../bson';
 import type { GridFSBucket } from './index';
+import type { GridFSFile } from './download';
+import type { WriteConcernOptions } from '../write_concern';
+import type { Collection } from '../collection';
```
nbbeeken, line 10: revert the import sorting here

src/sdam/topology.ts
```diff
@@ -981,7 +988,7 @@ function processWaitQueue(topology: Topology) {
   if (topology.s.state === STATE_CLOSED) {
     drainWaitQueue(
       topology[kWaitQueue],
-      new MongoDriverError('Topology is closed, please connect')
+      new MongoTopologyClosedError(`${TOPOLOGY_CLOSED_ERROR}, please connect`)
```
nbbeeken, line 991: I think we can leave this error to be without message like the usage above

src/error.ts
```diff
@@ -409,7 +409,7 @@ export class MongoCursorInUseError extends MongoAPIError {
  * @category Error
  */
 export class MongoServerClosedError extends MongoAPIError {
-  constructor(message: string) {
+  constructor(message = 'Server has been closed') {
```
nbbeeken, line 412: Are we all in agreement on tense? super bikeshedding but present/imperative seems more ideal

src/sessions.ts
andymina: Capturing the rebucketing that happened in NODE-3421 where errors have been switched between MongoRuntimeError and MongoAPIError, with MongoTransactionError being one of them. MongoTransactionError is now a child of MongoAPIError.

src/sessions.ts
```diff
@@ -441,6 +443,7 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    * This is here to ensure that ClientSession is never serialized to BSON.
    */
   toBSON(): never {
+    // TODO(NODE-3484): Replace with MongoBSONParseError
```
nbbeeken, line 446: This is one of those unreachables, its not a BSON parse error its really any error just to stop BSON from attempting to serialize a session, fair to leave as a driver error or runtime.

src/utils.ts
```diff
@@ -258,8 +262,7 @@ export function executeLegacyOperation(
       const optionsIndex = args.length - 2;
       args[optionsIndex] = Object.assign({}, args[optionsIndex], { session: session });
     } else if (opOptions.session && opOptions.session.hasEnded) {
-      // TODO(NODE-3405): Replace this with MongoExpiredSessionError
-      throw new MongoDriverError('Use of expired sessions is not permitted');
+      throw new MongoExpiredSessionError('Use of expired sessions is not permitted');
```
nbbeeken, line 265: Let's make this error message the same has the one in applySession in sessions.ts
(maybe even make that the default message for expired session errors.)

nbbeeken (changes requested):

src/error.ts
andymina: @dariakp @emadum @W-A-James? Personally, I'm in favor of present tense but happy to defer.

src/sessions.ts
dariakp: The TODO can be removed

src/error.ts
emadum: My vote would also be for using present/imperative.

src/operations/common_functions.ts
```diff
@@ -42,7 +42,7 @@ export function indexInformation(
 
   // Did the user destroy the topology
   if (getTopology(db).isDestroyed())
-    return callback(new MongoDriverError('topology was destroyed'));
+    return callback(new MongoTopologyClosedError('Topology was destroyed'));
```
emadum, line 45:
```suggestion
    return callback(new MongoTopologyClosedError('Topology is destroyed'));
```
If we're going to update messages to present tense, can do this one as well.

emadum (approved):
LGTM 👍

andymina:
@dariakp re: [this comment](https://github.com/mongodb/node-mongodb-native/pull/2913#discussion_r686911454) You're totally right. I'm going to leave a TODO pointing toward [NODE-3483](https://jira.mongodb.org/browse/NODE-3483) which introduces MongoConnectionPoolClosedError as a child of RuntimeError. Also going to revert this to MongoRuntimeError for the time being.

nbbeeken (approved):

src/change_stream.ts
```diff
@@ -683,15 +683,15 @@ function processNewChange<TSchema>(
...
+    if (callback) callback(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
     return;
   }
 
   // a null change means the cursor has been notified, implicitly closing the change stream
   if (change == null) {
-    // TODO(NODE-3405): Replace with MongoStreamClosedError
-    return closeWithError(changeStream, new MongoDriverError(CHANGESTREAM_CLOSED_ERROR), callback);
+    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
+    return closeWithError(changeStream, new MongoAPIError(CHANGESTREAM_CLOSED_ERROR), callback);
```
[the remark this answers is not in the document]
andymina: Yes, they're all API errors. They catch the case where someone has closed their ChangeStream before we are able to execute the logic beneath these checks. As an example, processing the resume queue will fail with the API error because the ChangeStream has been closed. So it does seem like incorrect usage.

src/change_stream.ts
```diff
@@ -683,15 +683,15 @@ function processNewChange<TSchema>(
   callback?: Callback<ChangeStreamDocument<TSchema>>
 ) {
   if (changeStream[kClosed]) {
-    // TODO(NODE-3405): Replace with MongoStreamClosedError
-    if (callback) callback(new MongoDriverError(CHANGESTREAM_CLOSED_ERROR));
+    // TODO(NODE-3485): Replace with MongoChangeStreamStreamClosedError
```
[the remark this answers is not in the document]
dariakp: Can you update the other `StreamStream` comments in the file to match the spelling?

src/change_stream.ts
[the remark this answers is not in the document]
nbbeeken: Sorry Andy I overlooked this one when we were on the call, you're correct. Except for this one, this is the one case where the `change` variable is null because the underlying cursor gave us null. Since change streams are "forever" this would only happen if there was some server reason for shutting the change stream so this would not be the fault of usage but rather a runtime issue.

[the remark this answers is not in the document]
dariakp: This specific one isn't an api error, also, could update all the TODO comments in this file to remove the duplicate `StreamStream` spelling?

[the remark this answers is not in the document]
andymina: Oops! Going to change this to be MongoRuntimeError and leave it as such. I think creating an entirely new error for this might be overkill.

dariakp (changes requested):

dariakp (approved):

W-A-James (approved):

PR #2914, opened 2021-07-20
 test(NODE-3076): run unified change stream tests

emadum:
Note: also moved `test/spec/change-stream` to `test/spec/change-streams` so that it matches the specifications repository.


test/functional/change_stream_spec.test.js
```diff
@@ -2,13 +2,30 @@
...
+        it(String(test.description), {
+          metadata: { sessions: { skipLeakTests: true } },
+          test: async function () {
+            await runUnifiedTest(this, changeStreamTest, test);
+          }
+        });
+      }
+    });
+  }
+});
```
nbbeeken, line 26:
```suggestion
describe('Change Streams Spec - Unified', function () {
  runUnifiedSuite(loadSpecTests(path.join('change-streams', 'unified')));
});
```

We've been bringing this block of code to each new unified spec when we already have a shared way of doing it, IIRC it came in a merge after the original unified runner so we've been using this pattern as the template. Anyway! this works if we wanna pair down the diff a bit :)

test/functional/change_stream_spec.test.js
```diff
@@ -2,13 +2,30 @@
 
 const path = require('path');
 const chai = require('chai');
-const loadSpecTests = require('../spec').loadSpecTests;
+const { loadSpecTests } = require('../spec');
+const { runUnifiedTest } = require('./unified-spec-runner/runner');
```
nbbeeken, line 6:
```suggestion
const { runUnifiedSuite } = require('./unified-spec-runner/runner');
```

If we go with the other suggested change

nbbeeken (commented):
What do you think about using `runUnifiedSuite`?

test/functional/change_stream_spec.test.js
emadum: Reducing boilerplate is always good, thanks for pointing out the existence of this helper! 👍

nbbeeken (approved):
LGTM 🔥

dariakp (approved):
LGTM

PR #2915, opened 2021-07-20
fix(NODE-3441): fix typings for createIndexes

emadum:
This isn't super clear from the index management specification, but it seems like `IndexDescription` needs to extend `CreateIndexesOptions`, since each index description may also contain options.

### DIFF
--- a/test/types/index_options.test-d.ts
+++ b/test/types/index_options.test-d.ts
@@ -1,0 +1,25 @@
+import { expectAssignable, expectNotAssignable } from 'tsd';
+import type { IndexDescription } from '../../src';
+
+// test that all valid index options are allowed in IndexDescription
+expectAssignable<IndexDescription>({ key: {}, background: true });
+expectAssignable<IndexDescription>({ key: {}, expireAfterSeconds: 2400 });
+expectAssignable<IndexDescription>({ key: {}, name: 'index_1' });
+expectAssignable<IndexDescription>({ key: {}, sparse: true });
+expectAssignable<IndexDescription>({ key: {}, storageEngine: {} });
+expectAssignable<IndexDescription>({ key: {}, unique: true });
+expectAssignable<IndexDescription>({ key: {}, version: 1 });
+expectAssignable<IndexDescription>({ key: {}, default_language: 'english' });
+expectAssignable<IndexDescription>({ key: {}, language_override: 'english' });
+expectAssignable<IndexDescription>({ key: {}, textIndexVersion: 2 });
+expectAssignable<IndexDescription>({ key: {}, weights: {} });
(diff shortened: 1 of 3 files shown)


test/types/mongodb.test-d.ts
```diff
@@ -32,3 +32,8 @@ const composedMap = mappedAgg.map<string>(x => x.toString());
 expectType<AggregationCursor<string>>(composedMap);
 expectType<string | null>(await composedMap.next());
 expectType<string[]>(await composedMap.toArray());
+
+// test that VALID_INDEX_OPTIONS like expireAfterSeconds are allowed in IndexDefinitions
+coll.createIndexes([{ key: { event: 1 }, name: 'event', expireAfterSeconds: 2400 }]);
+// test that invalid options are not allowed
+expectError(coll.createIndexes([{ key: { event: 1 }, name: 'event', invalidOption: 2400 }]));
```
emadum, line 39: I added some minimal tests here to ensure the fix worked; would we want to actually test each option in [VALID_INDEX_OPTIONS](https://github.com/mongodb/node-mongodb-native/blob/9e48bbdc95149ca62fa21404624125b87c3c9d56/src/operations/indexes.ts#L28-L58), or does using `expireAfterSeconds` as a representative valid option seem like enough coverage for this scenario?

nbbeeken: I think we should add tests for them all:
You could use this form: `expectType<IndexDescription>({ key: {}, hidden: true });` not sure if hidden should be allowed here but it is in the VALID set

emadum: Good call, `version` was also missing as well as `hidden`. PTAL 🙏

test/types/index_options.test-d.ts
```diff
@@ -0,0 +1,28 @@
...
+  key: {},
+  background: true,
+  expireAfterSeconds: 2400,
+  name: 'index_1',
+  sparse: true,
+  storageEngine: {},
+  unique: true,
+  version: 1,
+  default_language: 'english',
+  language_override: 'english',
```
emadum, line 15: Note that these are camel case in the spec, too late to update them to conform for 4.0?

nbbeeken: I think so, we could add support for either one? Snake and camel variations are permitted in each driver but this is just unusual for us. 🤷  It's a small inconsistency we can live with if we want to just leave it alone.

nbbeeken (approved):
LGTM

dariakp (approved):
LGTM

test/types/index_options.test-d.ts
emadum: I'm fine with leaving as is, it's minor enough that I think it's better to avoid further scope creep in this PR.

PR #2916, opened 2021-07-21
fix(NODE-3442): AsyncIterator has incorrect return type

nbbeeken:
The AsyncIterator had TSchema or null as its return type the truth is that the for await loop ends when the next function returns null so you never get null in your loop.

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -224,10 +224,11 @@ export abstract class AbstractCursor<
   [Symbol.asyncIterator](): AsyncIterator<TSchema, void> {
     return {
       next: () =>
-        this.next().then(value => {
-          if (value !== null && value !== undefined) return { value, done: false };
-          return { value: undefined, done: true };
-        })
+        this.next().then(value =>
+          value !== null && value !== undefined
+            ? { value, done: false }
+            : { value: undefined, done: true }
+        )
     };
   }
 
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -221,9 +221,13 @@ export abstract class AbstractCursor<
     return this[kDocuments].splice(0, number ?? this[kDocuments].length);
   }
 
-  [Symbol.asyncIterator](): AsyncIterator<TSchema | null> {
+  [Symbol.asyncIterator](): AsyncIterator<TSchema, void> {
     return {
-      next: () => this.next<TSchema>().then(value => ({ value, done: value === null }))
+      next: () =>
+        this.next().then(value => {
+          if (value !== null && value !== undefined) return { value, done: false };
+          return { value: undefined, done: true };
+        })
     };
   }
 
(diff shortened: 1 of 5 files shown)


test/unit/execute_legacy_operation.test.js
```diff
@@ -39,18 +38,10 @@ describe('executeLegacyOperation', function () {
...
-
-    setTimeout(() => {
-      try {
-        expect(callbackError).to.equal(expectedError);
-        done();
-      } catch (e) {
-        done(e);
-      }
+    return executeLegacyOperation(topology, operation, [{}, null], options).then(null, err => {
+      expect(err).to.equal(expectedError);
```
nbbeeken, line 44: This was failing on the first commit I pushed, might have been flaky and I coulda tried a rerun but changing this to not rely on a setTimeout seems like a quick cleanup win.

test/functional/cursor_async_iterator.test.js
```diff
@@ -2,6 +2,7 @@
 
 const { expect } = require('chai');
 const Sinon = require('sinon');
+const { Promise: BluebirdPromise } = require('bluebird');
```
nbbeeken, line 5: I brought in bluebird because our current test just did `class P extends Promise` and I needed something that wouldn't extend the native promise.

src/cursor/abstract_cursor.ts
```diff
@@ -221,9 +221,13 @@ export abstract class AbstractCursor<
...
   }
 
-  [Symbol.asyncIterator](): AsyncIterator<TSchema | null> {
+  [Symbol.asyncIterator](): AsyncIterator<TSchema, void> {
     return {
-      next: () => this.next<TSchema>().then(value => ({ value, done: value === null }))
+      next: () =>
+        this.next().then(value => {
+          if (value !== null && value !== undefined) return { value, done: false };
+          return { value: undefined, done: true };
```
nbbeeken, line 229: Can totally ternary this, thoughts?

emadum (approved):
LGTM

src/cursor/abstract_cursor.ts
nbbeeken: I spread it out for debugging but just now thought I can clean it up, changed it, looks v neat now :)

dariakp (approved):
LGTM :)

adityapatadia:
I just upgraded driver from 3.6 to 4.1 and execution just stops at async iterator. Does not move ahead.

nbbeeken:
Hey @adityapatadia that sounds like a pretty bad bug, could you open a ticket on [our JIRA](https://jira.mongodb.org/projects/NODE) describing the steps to reproduce?

adityapatadia:
Apologies for wrong flag. When we use newly released v7.0 of snappy library, none of the queries work. I downgraded snappy and everything works fine.

nbbeeken:
Funny you mention it! I identified a bug with the new Snappy lib just hours ago, for now the driver can only use snappy 6.x but we plan to add a fix to support either soon. Keep an eye on the release notes for the fix or follow this ticket: https://jira.mongodb.org/browse/NODE-3528

adityapatadia:
This is now fixed: https://github.com/Brooooooklyn/snappy/pull/21

PR #2917, opened 2021-07-23
test: legacy spec runner improvements

emadum:
- Allow skipped tests to skip `beforeEach/afterEach` hooks, significantly speeding up test runs particularly against remote servers with higher latency.
- Fixes a bug in the URL returned by the legacy test runner when `isShardedEnvironment && !useMultipleMongoses`

### DIFF
--- a/test/functional/spec-runner/index.js
+++ b/test/functional/spec-runner/index.js
@@ -135,29 +135,8 @@ function generateTopologyTests(testSuites, testContext, filter) {
           beforeEach(() => prepareDatabaseForSuite(testSuite, testContext));
           afterEach(() => testContext.cleanupAfterSuite());
           testSuite.tests.forEach(spec => {
-            it(spec.description, function () {
-              if (requires.authEnabled && process.env.AUTH !== 'auth') {
-                // TODO: We do not have a way to determine if auth is enabled in our mocha metadata
-                // We need to do a admin.command({getCmdLineOpts: 1}) if it errors (code=13) auth is on
-                this.skip();
-              }
-
-              if (
-                spec.operations.some(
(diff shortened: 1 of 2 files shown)


test/functional/spec-runner/index.js
```diff
@@ -180,6 +159,31 @@ function generateTopologyTests(testSuites, testContext, filter) {
   });
 }
 
+function shouldRunSpecTest(requires, spec, filter) {
+  if (requires.authEnabled && process.env.AUTH !== 'auth') {
+    // TODO: We do not have a way to determine if auth is enabled in our mocha metadata
```
dariakp, line 164: Do you think it would be worthwhile to file a ticket so that we are keeping track of this TODO?

dariakp (approved):
Changes look good to me, though maybe we should file a ticket for that outstanding TODO

test/functional/spec-runner/index.js
```diff
@@ -135,29 +135,8 @@ function generateTopologyTests(testSuites, testContext, filter) {
...
-              }
-
-              if (
-                spec.skipReason ||
-                (filter && typeof filter === 'function' && !filter(spec, this.configuration))
-              ) {
-                return this.skip();
-              }
-
+            const maybeIt = shouldRunSpecTest.call(this, requires, spec, filter) ? it : it.skip;
```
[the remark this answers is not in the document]
emadum: No, passing it in does sound better, I'll update 👍

test/functional/spec-runner/index.js
emadum: NODE-3488

nbbeeken (approved):
🚀

PR #2918, opened 2021-07-26
fix(NODE-2883): Aggregate Operation should not require parent parameter

W-A-James:
## Description

Remove parent parameter from AggregateOperation and parent reference from AggregationCursor and change any instantiation of these internal classes to match their new signature

**What changed?**

* src/change_stream.ts
* src/collection.ts
* src/cursor/aggregation_cursor.ts
* src/db.ts
* src/operations/aggregate.ts
* src/operations/count_documents.ts

### DIFF
--- a/src/cursor/aggregation_cursor.ts
+++ b/src/cursor/aggregation_cursor.ts
@@ -7,7 +7,6 @@ import type { Sort } from '../sort';
 import type { Topology } from '../sdam/topology';
 import type { Callback, MongoDBNamespace } from '../utils';
 import type { ClientSession } from '../sessions';
-import type { OperationParent } from '../operations/command';
 import type { AbstractCursorOptions } from './abstract_cursor';
 import type { ExplainVerbosityLike } from '../explain';
 import type { Projection } from '../mongo_types';
@@ -15,8 +14,6 @@ import type { Projection } from '../mongo_types';
 /** @public */
 export interface AggregationCursorOptions extends AbstractCursorOptions, AggregateOptions {}
 
-/** @internal */
-const kParent = Symbol('parent');
 /** @internal */
 const kPipeline = Symbol('pipeline');
 /** @internal */
@@ -31,15 +28,13 @@ const kOptions = Symbol('options');
  */
 export class AggregationCursor<TSchema = Document> extends AbstractCursor<TSchema> {
   /** @internal */
-  [kParent]: OperationParent; // TODO: NODE-2883
-  /** @internal */
   [kPipeline]: Document[];
   /** @internal */
   [kOptions]: AggregateOptions;
 
   /** @internal */
   constructor(
-    parent: OperationParent,
+    //parent: OperationParent,
     topology: Topology,
     namespace: MongoDBNamespace,
     pipeline: Document[] = [],
@@ -47,7 +42,7 @@ export class AggregationCursor<TSchema = Document> extends AbstractCursor<TSchem
   ) {
     super(topology, namespace, options);
 
-    this[kParent] = parent;
+    //this[kParent] = parent;
     this[kPipeline] = pipeline;
     this[kOptions] = options;
   }
@@ -59,7 +54,7 @@ export class AggregationCursor<TSchema = Document> extends AbstractCursor<TSchem
   clone(): AggregationCursor<TSchema> {
     const clonedOptions = mergeOptions({}, this[kOptions]);
     delete clonedOptions.session;
-    return new AggregationCursor(this[kParent], this.topology, this.namespace, this[kPipeline], {
+    return new AggregationCursor(this.topology, this.namespace, this[kPipeline], {
       ...clonedOptions
     });
   }
@@ -70,7 +65,7 @@ export class AggregationCursor<TSchema = Document> extends AbstractCursor<TSchem
 
   /** @internal */
   _initialize(session: ClientSession | undefined, callback: Callback<ExecutionResult>): void {
-    const aggregateOperation = new AggregateOperation(this[kParent], this[kPipeline], {
+    const aggregateOperation = new AggregateOperation(this[kPipeline], this.namespace, {
       ...this[kOptions],
       ...this.cursorOptions,
       session
@@ -97,7 +92,7 @@ export class AggregationCursor<TSchema = Document> extends AbstractCursor<TSchem
 
     return executeOperation(
       this.topology,
-      new AggregateOperation(this[kParent], this[kPipeline], {
+      new AggregateOperation(this[kPipeline], this.namespace, {
         ...this[kOptions], // NOTE: order matters here, we may need to refine this
         ...this.cursorOptions,
         explain: verbosity
--- a/src/cursor/aggregation_cursor.ts
+++ b/src/cursor/aggregation_cursor.ts
@@ -41,7 +41,6 @@ export class AggregationCursor<TSchema = Document> extends AbstractCursor<TSchem
   ) {
     super(topology, namespace, options);
 
-    //this[kParent] = parent;
     this[kPipeline] = pipeline;
     this[kOptions] = options;
   }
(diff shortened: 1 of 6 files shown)


src/cursor/aggregation_cursor.ts
```diff
@@ -30,24 +27,22 @@ const kOptions = Symbol('options');
...
-  [kParent]: OperationParent; // TODO: NODE-2883
   /** @internal */
   [kPipeline]: Document[];
   /** @internal */
   [kOptions]: AggregateOptions;
 
   /** @internal */
   constructor(
-    parent: OperationParent,
+    //parent: OperationParent,
```
nbbeeken, line 37: Let's go ahead a remove this looks like we're passing right?

src/operations/aggregate.ts
```diff
@@ -47,14 +42,11 @@ export class AggregateOperation<T = Document> extends CommandOperation<T> {
   pipeline: Document[];
   hasWriteStage: boolean;
 
-  constructor(parent: OperationParent, pipeline: Document[], options?: AggregateOptions) {
-    super(parent, options);
+  constructor(pipeline: Document[], ns?: MongoDBNamespace, options?: AggregateOptions) {
+    super(undefined, Object.assign({ dbName: ns?.db }, options));
```
nbbeeken, line 46:
```suggestion
    super(undefined, { dbName: ns?.db, ...options });
```

I'm surprised the super class permits the first argument to be optional but if that's the case my only comment here is to use some modern syntax for the same thing.

src/operations/aggregate.ts
```diff
@@ -47,14 +42,11 @@ export class AggregateOperation<T = Document> extends CommandOperation<T> {
...
 
     this.options = options ?? {};
-    this.target =
-      parent.s.namespace && parent.s.namespace.collection
-        ? parent.s.namespace.collection
-        : DB_AGGREGATE_COLLECTION;
+    this.target = ns && ns.collection ? ns.collection : DB_AGGREGATE_COLLECTION;
```
nbbeeken, line 49:
```suggestion
    this.target = ns?.collection ?? DB_AGGREGATE_COLLECTION;
```

src/operations/aggregate.ts
```diff
@@ -47,14 +42,11 @@ export class AggregateOperation<T = Document> extends CommandOperation<T> {
   pipeline: Document[];
   hasWriteStage: boolean;
 
-  constructor(parent: OperationParent, pipeline: Document[], options?: AggregateOptions) {
-    super(parent, options);
+  constructor(pipeline: Document[], ns?: MongoDBNamespace, options?: AggregateOptions) {
```
nbbeeken, line 45: Don't change this if it turns out to be a lot of refactor but: AggOp takes pipeline then namespace, and AggCursor takes namespace then pipeline. Seems odd that those would be different unless one is consistent with the cursor and the other is consistent with operations. Any insight there and if its worth a little refactor?

nbbeeken (changes requested):

src/operations/aggregate.ts
W-A-James: I can go ahead and fix that actually just for dev friendliness. Shouldn't take too long. I did it that way to have the namespace be optional for the AggregateOperation, but since we're axing the parent parameter, it should just be mandatory and be the first argument.

src/cursor/aggregation_cursor.ts
```diff
@@ -30,24 +27,21 @@ const kOptions = Symbol('options');
...
-    parent: OperationParent,
     topology: Topology,
     namespace: MongoDBNamespace,
     pipeline: Document[] = [],
     options: AggregateOptions = {}
   ) {
     super(topology, namespace, options);
 
-    this[kParent] = parent;
+    //this[kParent] = parent;
```
nbbeeken, line 44: remove?

src/operations/aggregate.ts
```diff
@@ -47,14 +42,11 @@ export class AggregateOperation<T = Document> extends CommandOperation<T> {
...
-    super(parent, options);
+  constructor(ns: MongoDBNamespace, pipeline: Document[], options?: AggregateOptions) {
+    super(undefined, { ...options, dbName: ns.db });
 
     this.options = options ?? {};
-    this.target =
-      parent.s.namespace && parent.s.namespace.collection
-        ? parent.s.namespace.collection
-        : DB_AGGREGATE_COLLECTION;
+    this.target = ns.collection || DB_AGGREGATE_COLLECTION;
```
nbbeeken, line 49:
```suggestion
    this.target = ns.collection ?? DB_AGGREGATE_COLLECTION;
```
just a nit, optional

nbbeeken (changes requested):

src/operations/aggregate.ts
W-A-James: I changed that because when ns.collection is set to the empty string, we set the target to the empty string instead of `DB_AGGREGATE_COLLECTION`, which leads to a few test failures

nbbeeken: ah Fascinating, I feel like that makes this worthy of more explicit code.

This but with curly braces
```js
if (ns.collection === undefined || ns.collection === '') this.target = DB_AGGREGATE_COLLECTION;
else this.target = ns.collection;
```

Or we can bring back the ternary, Or leave it as is with a comment, `// covers undefined and empty string`, It's just easy to over look if the code itself or a comment doesn't call it out, what do you think?

W-A-James: yeah that sounds good. Makes sense to have the comment so the next person to see it and thinks to change it won't have to go bughunting like I did

nbbeeken (approved):
LGTM

src/operations/aggregate.ts
```diff
@@ -47,14 +42,16 @@ export class AggregateOperation<T = Document> extends CommandOperation<T> {
...
+  constructor(ns: MongoDBNamespace, pipeline: Document[], options?: AggregateOptions) {
+    super(undefined, { ...options, dbName: ns.db });
 
     this.options = options ?? {};
-    this.target =
-      parent.s.namespace && parent.s.namespace.collection
-        ? parent.s.namespace.collection
-        : DB_AGGREGATE_COLLECTION;
+
+    if (ns.collection === undefined || ns.collection === '') {
```
dariakp, line 50: I think we can capture the desired outcome with just `this.target = ns.collection || DB_AGGREGATE_COLLECTION` (it would preserve the previous logic, and it's different from existing logic in that it would still assign the target to `null` or `0` if those happened to get passed in as values; however, I don't see why we would want to start doing that)

dariakp (changes requested):
Just one small thing here

src/operations/aggregate.ts
dariakp: Lol I just noticed this after doing my review... I think a comment would be better; explicit undefined checks are somewhat weird and puts into question why we aren't validating for the other excluded cases... if we're concerned about nullish values but only these ones, why aren't we throwing errors on the other kinds?

src/operations/aggregate.ts
dariakp: (See my comment on the other thread)

src/operations/aggregate.ts
nbbeeken:
```
if (typeof ns.collection === 'string' && ns.collection !== '') this.target = ns.collection
else this.target = DB_AGGREGATE_COLLECTION;
```

Ok one more code suggestion ^ 
But also reverting to the ternary and using the same nullish that was there before along with a comment is a good solution to me as well. Your point is correct we should make sure that we are actually filtering for our intention and not leaving a gap that might suggest null is okay here.

dariakp: I'd recommend against the potential behavior change unless truly necessary; and if we do make that behavior change, we'd want to throw on ns.collection not being null/undefined so that we can guard against unexpected input sneaking through (one of those runtime errors)... feels like overkill for this ticket though, hence easier to just keep the original behavior (`this.target = ns.collection || DB_AGGREGATE_COLLECTION;`)

dariakp (approved):

PR #2920, opened 2021-07-28
feat(NODE-2843): implement sessions advanceClusterTime method

dariakp:
## Description

NODE-2843

**What changed?**

Implemented the advanceClusterTime session method, refactored internal helper function for naming consistency

### DIFF
--- a/src/sdam/common.ts
+++ b/src/sdam/common.ts
@@ -66,16 +66,16 @@ export interface ClusterTime {
   };
 }
 
-/** Shared function to determine clusterTime for a given topology */
-export function resolveClusterTime(
-  topology: Topology | ClientSession,
+/** Shared function to determine clusterTime for a given topology or session */
+export function _advanceClusterTime(
+  entity: Topology | ClientSession,
   $clusterTime: ClusterTime
 ): void {
-  if (topology.clusterTime == null) {
-    topology.clusterTime = $clusterTime;
+  if (entity.clusterTime == null) {
+    entity.clusterTime = $clusterTime;
   } else {
-    if ($clusterTime.clusterTime.greaterThan(topology.clusterTime.clusterTime)) {
-      topology.clusterTime = $clusterTime;
+    if ($clusterTime.clusterTime.greaterThan(entity.clusterTime.clusterTime)) {
+      entity.clusterTime = $clusterTime;
     }
   }
 }
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -266,7 +266,8 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
     if (
       !clusterTime.signature ||
       clusterTime.signature.hash?._bsontype !== 'Binary' ||
-      clusterTime.signature.keyId?._bsontype !== 'Long'
+      (typeof clusterTime.signature.keyId !== 'number' &&
+        clusterTime.signature.keyId?._bsontype !== 'Long') // apparently we decode the key to number?
     ) {
       throw new MongoInvalidArgumentError(
         'input cluster time must have a valid "signature" property with BSON Binary hash and BSON Long keyId'
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -265,8 +265,8 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
     }
     if (
       !clusterTime.signature ||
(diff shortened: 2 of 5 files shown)


test/unit/core/sessions.test.js
```diff
@@ -1,13 +1,22 @@
 'use strict';
 
 const mock = require('../../tools/mock');
+const BSON = require('bson');
```
nbbeeken, line 4:
```suggestion
const BSON = require('../../../src/bson');
```
Prefer importing bson from the driver so that the BSON-EXT logic (or anything else) is tested automatically.

src/sdam/common.ts
```diff
@@ -66,16 +66,16 @@ export interface ClusterTime {
   };
 }
 
-/** Shared function to determine clusterTime for a given topology */
-export function resolveClusterTime(
-  topology: Topology | ClientSession,
+/** Shared function to determine clusterTime for a given topology or session */
+export function _advanceClusterTime(
```
nbbeeken, line 70: Just a thought, we could just keep this without the underscore and put `@internal` in the tsdoc?

dariakp: I don't mind renaming, but this particular suggestion would make the name clash with the actual public method, so maybe we can come up with something else?

test/unit/core/sessions.test.js
dariakp: good call out, though all test files except for the bson import one require it directly - any thoughts on that?

nbbeeken: Hm, seems like something we should fix. In test/.eslintrc.json you can add bson to the forbidden modules, but there's 24 instances and at least two are intentionally testing bson vs bson-ext, so I guess lets defer to another ticket?

src/sdam/common.ts
nbbeeken: I think the clash is fine because the public one is a method so it can't be called without a session. But maybe: `advanceClusterTimeFor` to imply its for the topology or session... I mean I'm splitting hairs, I just don't love underscore names but I'm not going to hold us up on this 😆

dariakp: Yeah I understand that it's technically fine to use the same name, I just don't like to as a general practice because it makes it harder to tell at a glance (when doing a codebase wide search, for example) whether or not something is referencing the same thing. I don't have objections to something like `advanceClusterTimeFor` or `advanceClusterTimeOn` etc

test/unit/core/sessions.test.js
dariakp: yeah, feels a bit out of scope, can file a simple clean up ticket in the test epic; would you rather I do use the internal reference for this test though?

src/sdam/common.ts
nbbeeken: Both of those sound good to me, I leave the decision to you :)

test/unit/core/sessions.test.js
nbbeeken: Totally, I don't imagine there being any impact here so whichever for now and we'll clean them up in that ticket.

src/sdam/common.ts
nbbeeken: Decided to go with underscore

nbbeeken (approved):
LGTM

src/sessions.ts
```diff
@@ -249,6 +249,34 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
+    if (!clusterTime.clusterTime || clusterTime.clusterTime._bsontype !== 'Timestamp') {
+      throw new MongoInvalidArgumentError(
+        'input cluster time "clusterTime" property must be a valid BSON Timestamp'
+      );
+    }
+    if (
+      !clusterTime.signature ||
+      clusterTime.signature.hash?._bsontype !== 'Binary' ||
+      (typeof clusterTime.signature.keyId !== 'number' &&
+        clusterTime.signature.keyId?._bsontype !== 'Long') // apparently we decode the key to number?
```
dariakp, line 270: @nbbeeken Just bringing your attention to this change

nbbeeken: Just so I do not block this PR: looks correct to me.

Remind me to tell you the deeper issue here / file a ticket next week, has to do with BSON `promote*` options I think.

durran (approved):

PR #2922, opened 2021-07-30
fix(NODE-3510): omit incorrect `| void` in declaration of Promise overload of `rename()`

justingrant:
## Description

One of the TS overloaded declarations for `rename()` had a `| void` which AFAIK is incorrect. If a method returns a `Promise`, it should never be typed `| void`.  The problem doesn't show up in any other overload declarations, so I assume this was just a bug, not a statement about weird behavior of this method.

**What changed?**

I removed the `| void` for this overload's declaration. 

No.


nbbeeken (approved):
Thanks for catching this! it was certainly a mistake, only the callback overloads should return void. LGTM

dariakp (approved):

PR #2925, opened 2021-07-30
docs(NODE-3492): Deprecate CloseCursorOptions

W-A-James:
## Description
Deprecate CloseCursorOptions

**What changed?**

- src/cursor/abstract_cursor.ts

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -51,7 +51,8 @@ export const CURSOR_FLAGS = [
   'partial'
 ] as const;
 
-/** @public */
+/** @public
+ * @deprecated This interface is deprecated */
 export interface CursorCloseOptions {
   /** Bypass calling killCursors when closing the cursor. */
   /** @deprecated  the skipKillCursors option is deprecated*/
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -51,9 +51,7 @@ export const CURSOR_FLAGS = [
   'partial'
 ] as const;
 
-/** @public
- *  @deprecated Not used
- * */
+/** @public */
 export interface CursorCloseOptions {
   /** Bypass calling killCursors when closing the cursor. */
   skipKillCursors?: boolean;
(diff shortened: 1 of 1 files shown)


src/cursor/abstract_cursor.ts
```diff
@@ -43,7 +43,9 @@ export const CURSOR_FLAGS = [
   'partial'
 ] as const;
 
-/** @public */
+/** @public
+ *  @deprecated Not used
+ * */
 export interface CursorCloseOptions {
   /** Bypass calling killCursors when closing the cursor. */
   skipKillCursors?: boolean;
```
nbbeeken, line 51: I think we should deprecate just this option, I know the whole thing is unused but if we were to add options later then we wouldn't be un-deprecating but lets see what the team says.

nbbeeken (changes requested):

src/cursor/abstract_cursor.ts
dariakp: I agree with deprecating just the one option, which seemed to be the intent of the ticket anyway; I think if we deprecate the entire options object, it would make sense to also remove some of the overloads that use the options, which would be out of scope here, and would make it more work to add options in the future; my personal opinion is that if we manage to go an entire version without adding any options then we can consider removing the interface entirely for the next major release

dariakp (changes requested):

src/cursor/abstract_cursor.ts
nbbeeken: @W-A-James I think we still want the deprecation on the skipKillCursors as well as the overloads

W-A-James: @nbbeeken Ah I see what you and Daria mean now. In that case, should I go ahead and remove the deprecation from those function signatures?

nbbeeken (approved):

dariakp (changes requested):
For consistency's sake, we should either deprecate just the option, and leave the signatures alone, or deprecate the signatures and the interface. @nbbeeken and I are leaning toward the latter, but let's connect before making the final decision

nbbeeken (approved):

src/cursor/abstract_cursor.ts
```diff
@@ -43,9 +43,11 @@ export const CURSOR_FLAGS = [
   'partial'
 ] as const;
 
-/** @public */
+/** @public
+ * @deprecated This interface is deprecated */
 export interface CursorCloseOptions {
   /** Bypass calling killCursors when closing the cursor. */
+  /** @deprecated  the skipKillCursors option is deprecated*/
```
dariakp, line 50:
```suggestion
  /** @deprecated  the skipKillCursors option is deprecated */
```

dariakp (approved):

PR #2926, opened 2021-07-30
fix(NODE-3513): default command monitoring to off

nbbeeken:
Revert an accidental default enabling of command monitoring

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -724,7 +724,7 @@ export const OPTIONS = {
     type: 'uint'
   },
   monitorCommands: {
-    default: true,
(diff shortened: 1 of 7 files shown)


dariakp (changes requested):
Can we also add a sanity test to make sure we can set monitorCommands to `true`? I also don't see it in this file as part of `allURIOptions`, but I think it's allowed, so we should probably add a test for setting it both ways.

It looks like there are some tests failing in evergreen with "Cannot read property 'command' of undefined", so they may have been relying on command monitoring being turned on by default

dariakp (approved):

PR #2929, opened 2021-08-05
refactor(NODE-3519): remove force unpinning and simplify findOne

durran:
- Removes `FindOneOperation` to resolve cursor ownership issue (`toArray` recursively calls `next` so using `next` was fine here.)
- No longer forces connection unpinning.


src/collection.ts
```diff
@@ -710,16 +709,10 @@ export class Collection<TSchema extends Document = Document> {
...
-        this as TODO_NODE_3286,
-        filter,
-        resolveOptions(this, options)
-      ) as TODO_NODE_3286,
-      callback as TODO_NODE_3286
-    );
+    return this.find(filter, options)
+      .limit(-1)
+      .batchSize(1)
+      .next(callback as TODO_NODE_3286);
```
nbbeeken, line 715: (Just an fyi cus I was perusing the CI) Looks like the only issue is this used to return undefined, so we just need a small arrow function here to preserve that.

dariakp (approved):
LGTM

emadum (approved):

PR #2932, opened 2021-08-06
refactor(NODE-3337): unwrap WriteConcernError

andymina:
## Description

Unwrapped the nested errors seen in `handleMongoWriteConcernError` to be clearer and more readable.


dariakp:
@andymina we should be careful about removing the server error behavior of assigning all input object props to the error object; the function of the server error class is to wrap errors returned from the server, and I believe the original design was based around the idea that we might not always know what those properties are, so we just want to pass them on to the user so that they have that information from the server; removing this behavior is a potentially breaking change. The goal of the ticket was just to remove the explicit rewrapping of the write concern error, not necessarily change the actual implementation of the MongoServerError class.

andymina:
> @andymina we should be careful about removing the server error behavior of assigning all input object props to the error object; the function of the server error class is to wrap errors returned from the server, and I believe the original design was based around the idea that we might not always know what those properties are, so we just want to pass them on to the user so that they have that information from the server; removing this behavior is a potentially breaking change. The goal of the ticket was just to remove the explicit rewrapping of the write concern error, not necessarily change the actual implementation of the MongoServerError class.

Ahh, didn't realize that there were potentially more/new props coming in from the server. Initially I figured that the shape of MongoServerErrors that the driver needs would remain the same so I just explicitly retrieved them rather than relying on the loop. So sorry for any confusion! Will revert this change, ty :)

src/error.ts
```diff
@@ -136,23 +135,34 @@ export class MongoServerError extends MongoError {
...
+      this[kErrorLabels] = new Set(message.errorLabels);
+    }
+
+    this.code = message.code;
+    this.codeName = message.codeName;
+    this.writeConcernError = message.writeConcernError;
+    this.ok = message.ok;
+    this.topologyVersion = message.topologyVersion;
+
+    const currentProps = Object.getOwnPropertyNames(this);
```
dariakp, line 154: Is all this really necessary to simplify the multiple constructor calls in bulk/common? It looks like all this filtering is doing the same thing as the original for..in, but it's less efficient because of the multiple array function calls. Ideally, the server error class should not need to be touched at all to complete this ticket, and we definitely should not be adding new functionality to it unless absolutely necessary (referring to the kErrorLabels line above), so any logic that's specific to particular errors, and not all server errors, should live in those specific classes.

dariakp (changes requested):

src/error.ts
andymina: Neal and I talked things over and I removed the tons of array calls that were there. I opted to put prop names into a set for better efficiency and it'll likely be easier to read/maintain than loads of if conditions.

src/bulk/common.ts
```diff
@@ -334,9 +332,9 @@ export class BulkWriteResult {
  * @category Error
  */
 export class WriteConcernError {
-  err: MongoServerError;
+  err: { code: number; errmsg: string; errInfo?: Document };
```
nbbeeken, line 335: We technically have a breakage here, this is no longer `instanceof MongoServerError` but this was just relying on the key mapping logic of the MongoServerError constructor. 
I think its fair to make this change, all the properties are still accessible. Of course, up for discussion:

src/bulk/common.ts
```diff
@@ -346,16 +344,21 @@ export class WriteConcernError {
...
   /** Write concern error message. */
-  get errmsg(): string {
+  get errmsg(): string | undefined {
     return this.err.errmsg;
   }
 
-  toJSON(): { code?: number; errmsg: string } {
-    return { code: this.err.code, errmsg: this.err.errmsg };
+  /** Write concern error info. */
+  get errInfo(): Document | undefined {
```
nbbeeken, line 352: Why did we add this new getter again? There is spec related to this I believe, this makes our WriteConcernError "more" compliant/consistent.

src/bulk/common.ts
```diff
@@ -544,7 +547,8 @@ function mergeBatchResults(
   }
 
   if (result.writeConcernError) {
-    bulkResult.writeConcernErrors.push(new WriteConcernError(result.writeConcernError));
+    const { code, errmsg, errInfo } = result.writeConcernError;
+    bulkResult.writeConcernErrors.push(new WriteConcernError({ code, errmsg, errInfo }));
```
nbbeeken, line 551: I think we have to change this back to `result.writeConcernError`, we might be losing properties in this destructuring.
We can make the constructor accept a more flexible type, it's still good that we have a better typed `WriteConcernError` class, but incase there are more properties in certain server versions we should have them automatically pass through here.

src/bulk/common.ts
```diff
@@ -657,16 +661,12 @@ function handleMongoWriteConcernError(
   mergeBatchResults(batch, bulkResult, undefined, err.result);
 
   // TODO: Remove multiple levels of wrapping (NODE-3337)
```
nbbeeken, line 663: We can delete the TODO

src/bulk/common.ts
```diff
@@ -679,11 +679,22 @@ function handleMongoWriteConcernError(
...
   result: BulkWriteResult;
+  writeErrors: OneOrMore<WriteError> = [];
+  err?: WriteConcernError;
 
   /** Creates a new MongoBulkWriteError */
-  constructor(error: AnyError, result: BulkWriteResult) {
-    super(error as Error);
-    Object.assign(this, error);
+  constructor(
+    x: { message: string; code: number; writeErrors?: WriteError[] } | WriteConcernError | AnyError,
```
nbbeeken, line 687: We can name this something a little more descriptive than `x` now.

src/bulk/common.ts
```diff
@@ -679,11 +679,22 @@ function handleMongoWriteConcernError(
...
+    result: BulkWriteResult
+  ) {
+    super(x);
+    if (x instanceof Error) Object.assign(this, x);
```
nbbeeken, line 691: I believe we put the assign behind the gate of `instanceof Error` since that was filtering for MongoServerErrors while testing. 
We should move the assign to be the last call of the constructor ungated. (Or remove it all together and see if tests fail, but I defer to the team on seeing if that's acceptable).

src/operations/insert.ts
```diff
@@ -70,7 +70,7 @@ export class InsertOneOperation extends InsertOperation {
     super.execute(server, session, (err, res) => {
       if (err || res == null) return callback(err);
       if (res.code) return callback(new MongoServerError(res));
-      if (res.writeErrors) return callback(new MongoServerError(res.writeErrors[0]));
+      if (res.writeErrors) return callback(new MongoWriteConcernError(res.writeErrors[0]));
```
nbbeeken, line 73: Is this always correct? I don't think a writeError is always in the category of a WriteConcernError, but I could be wrong, lets sync to make sure this is correct

src/error.ts
```diff
@@ -682,13 +687,15 @@ function makeWriteConcernResultObject(input: any) {
...
+  errInfo?: Document;
 
-  constructor(message: ErrorDescription, result: Document) {
+  constructor(message: ErrorDescription, result?: Document) {
     if (result && Array.isArray(result.errorLabels)) {
       message.errorLabels = result.errorLabels;
     }
 
     super(message);
+    this.errInfo = message.errInfo;
```
nbbeeken, line 698: Related to another comment, I think I remember now this was automapped by the Object.assign in bulk right? and now its being done explicitly is that correct?

nbbeeken (changes requested):

src/bulk/common.ts
andymina: I didn't have any specific reason, just consistency between the getters of other props.

src/error.ts
andymina: Yup, that's correct!

src/operations/insert.ts
andymina: Capturing what spoke about in person: This has been reverted to MongoServerError since not every write error is a write concern error. The ideal solution would be to make this a `WriteError` class, but that would call for some error hierarchy refactoring since `WriteError` isn't a child of anything and that would be out of scope.

andymina:
Neal requested reviews from my laptop lol

src/error.ts
```diff
@@ -60,12 +60,12 @@ export const GET_MORE_RESUMABLE_CODES = new Set<number>([
...
 
 /** @public */
-export interface ErrorDescription {
+export interface ErrorDescription extends Document {
   message?: string;
   errmsg?: string;
   $err?: string;
   errorLabels?: string[];
-  [key: string]: any;
+  errInfo?: Document;
```
dariakp, line 68: I think removing `[key: string]: any` is lying about the type, because we still expect this description to contain arbitrary keys from the server (and we parse them in and assign them to the error in the MongoServerError class)

src/error.ts
```diff
@@ -136,23 +135,31 @@ export class MongoServerError extends MongoError {
...
-      }
-
-      for (const name in message) {
-        if (name === 'errorLabels' || name === 'errmsg' || name === 'message') {
-          continue;
-        }
-
-        (this as any)[name] = message[name];
-      }
+  constructor(message: ErrorDescription) {
```
dariakp, line 143: I honestly don't think that the changes below improve readability, and there are extra steps added that actually make it harder to follow the logic; originally, we were just iterating over all props returned from the server and assigning them to the error as its own properties; there is no complexity in maintaining that. 

The only exceptions where we don't directly read and assign from the server error are `message, errmsg, errorLabels`. So, what is the benefit from explicitly extracting, for example, `message.code` ahead of time, putting it into a set, and then excluding it from being set in the loop in L161? 

Moreover, setting these properties and then dynamically generating the props to filter is introducing unnecessary inefficiency: the set of properties to exclude doesn't change from error to error, we don't need to compute it each time.

If you feel that strongly about the triple `||` statement, you could create a constant regex declaration outside this class `const serverPropsToExclude = /^errorLabels$|^errmsg$|^message$/` and then set the condition to `if (serverPropsToExclude.test(name)) { continue; }`

dariakp (changes requested):

src/error.ts
dariakp: missed the extends above, resolving

src/bulk/common.ts
nbbeeken: You want to map these props on to this class? I think that sounds good removes a layer of indirection (and we keep a getter to prevent breakage)

src/error.ts
nbbeeken: > setting these properties and then dynamically generating the props to filter

I agree here, we could hoist this set of keys to a `KNOWN_SERVER_ERROR_PROPS` and it'll be a one time construction. I think that could go a long way to making the logic clearer around our intention here.

> So, what is the benefit from explicitly extracting, for example, message.code ahead of time

Nothing functionally, entirely from a dev perspective this hints to me what's expected on the MongoServerError class, maybe it is enough to keep the property declarations and their types at the top of the class now, but having explicit line by line assignment for these "well known" keys helps callout what is on the class. 

I think this might be the result of my bias perspective of, before this ticket, MongoServerError was a mystery box. It still is, but with intention, the idea is for it to inherit anything the server sends us as an error. I recommended we take this as far as we could in the direction of figuring what was actually defined on this class. 

I would be fine with removing the explicit assignment but keeping just the type defs at the top (along with `[key: string]: any` of course)

dariakp: I don't think we need a `KNOWN_SERVER_ERROR_PROPS` I think marking the corresponding block of the class prop type defs with a comment saying that they are known server error props would accomplish the same thing (regarding communicating intent) without having to explicitly write them out in actual code, too. I think it's only the 3 that are being excluded that are worth hoisting up, if we do this at all.

src/bulk/common.ts
dariakp: something along the lines of defining a `kErr` symbol (maybe more descriptive) set equal to the `{ code, errmsg, errInfo }` and then a public getter marked deprecated: `get err() { return this[kErr] }` so that we hopefully don't break any existing functionality in case someone is trying to read directly from the current err prop

src/bulk/common.ts
```diff
@@ -334,28 +335,44 @@ export class BulkWriteResult {
...
+  errCode: number;
+  errMsg: string;
+  errDetails?: Document;
+  [kErr]: { code: number; errmsg: string; errInfo?: Document };
 
-  constructor(err: MongoServerError) {
-    this.err = err;
+  constructor(error: { code: number; errmsg: string; errInfo?: Document }) {
+    this[kErr] = error;
+    this.errCode = error.code;
```
dariakp, line 345: Since this is a spec error, I don't think we should add new properties to the body (at least at this point), so I'd skip the explicit props on the class and just update the getters to read the code/errmsg/errInfo from [kErr] instead

dariakp (changes requested):

src/bulk/common.ts
andymina: Ahh, okay! That's super smart. Will make that change now.

src/bulk/common.ts
```diff
@@ -26,7 +26,10 @@ import type { Collection } from '../collection';
 import type { Topology } from '../sdam/topology';
 import type { CommandOperationOptions, CollationOptions } from '../operations/command';
 import type { Hint } from '../operations/operation';
-import type { Filter, OptionalId, UpdateFilter } from '../mongo_types';
+import type { Filter, OneOrMore, OptionalId, UpdateFilter } from '../mongo_types';
+
+/** @internal */
+const kErr = Symbol('err');
```
emadum, line 32:
```suggestion
const kServerError = Symbol('serverError');
```
Can we name this a bit more descriptively?

emadum (changes requested):
One small request, and there is a lint error to fix, otherwise LGTM though! 👍

src/bulk/common.ts
```diff
@@ -328,34 +329,61 @@ export class BulkWriteResult {
...
 
-  toJSON(): { code?: number; errmsg: string } {
-    return { code: this.err.code, errmsg: this.err.errmsg };
+  /** Write concern error info. */
+  get errInfo(): Document | undefined {
+    return this[kServerError].errInfo;
+  }
+
+  /**
+   * @internal
```
dariakp, line 369: does it make sense to mark this internal if it's explicitly intended as a backcompat option for users?

src/bulk/common.ts
```diff
@@ -328,34 +329,61 @@ export class BulkWriteResult {
...
+  get err(): WriteConcernErrorData {
+    return this[kServerError];
+  }
+
+  toJSON(): {
+    code: number;
+    errmsg: string;
+    errInfo?: Document;
+  } {
+    // WriteConcernErrorData has been inlined as the return to avoid making it public
```
dariakp, line 381: what's the reasoning against make it public?

nbbeeken: I figured it would be yet another type clouding the namespace relating to an unclearly layered WriteConcernError/WriteError/BulkError space. But it's an 🧅 that can be argued otherwise

src/bulk/common.ts
nbbeeken: I suppose my thinking was it avoids breaking runtime code but intentionally makes it inaccessible to TS but I realize that could be nearly equal in breakage. 

lets go with making this interface public and not introduce all these specialized exceptions. Sorry @andymina my advice steered you here!!

andymina: No worries!

emadum (approved):
LGTM 👍

dariakp (approved):
LGTM :)

nbbeeken (approved):
LGTM

PR #2933, opened 2021-08-06
test(NODE-3461): refactor lb manual to full test run

durran:
Removes the manual test and runs the entire test suite. The load balancer unified tests are moved into the `test/functional` directory and the test runner setup is changed to put the topology in the correct load balancer mode.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -319,18 +319,11 @@ export class Db {
    * @param name - the collection name we wish to access.
    * @returns return the new Collection instance
    */
-  collection<TSchema extends Document = Document>(name: string): Collection<TSchema>;
   collection<TSchema extends Document = Document>(
     name: string,
-    options: CollectionOptions
-  ): Collection<TSchema>;
-  collection<TSchema extends Document = Document>(
-    name: string,
-    options?: CollectionOptions
+    options: CollectionOptions = {}
   ): Collection<TSchema> {
-    if (!options) {
-      options = {};
-    } else if (typeof options === 'function') {
+    if (typeof options === 'function') {
       // TODO(NODE-3485): Replace this with MongoDeprecationError
       throw new MongoDriverError('The callback form of this helper has been removed.');
     }
--- a/src/db.ts
+++ b/src/db.ts
@@ -322,6 +322,10 @@ export class Db {
   collection<TSchema extends Document = Document>(name: string): Collection<TSchema>;
   collection<TSchema extends Document = Document>(
     name: string,
+    options: CollectionOptions
+  ): Collection<TSchema>;
+  collection<TSchema extends Document = Document>(
+    name: string,
     options?: CollectionOptions
   ): Collection<TSchema> {
     if (!options) {
--- a/test/functional/load-balancer-spec.test.js
+++ b/test/functional/load-balancer-spec.test.js
@@ -15,15 +15,7 @@ const SKIP = [
   //       in sub-optimal pinning.
   'pinned connections are not returned after an network error during getMore',
   'pinned connections are not returned to the pool after a non-network error on getMore',
-  'stale errors are ignored',
-
-  // NOTE: The driver correctly fails these 2 tests in non LB mode due to validation that
-  // loadBalanced=true is only valid as a URI option and not a regular driver option according
-  // to the spec. However the unified test runner implementation incorrectly passes URI options
-  // as regular driver options to the MongoClient. There are additional unit tests to cover
-  // these cases that demostrate this works as intended.
-  'operations against non-load balanced clusters fail if URI contains loadBalanced=true',
-  'operations against non-load balanced clusters succeed if URI contains loadBalanced=false'
+  'stale errors are ignored'
 ];
 
 describe('Load Balancer Unified Tests', function () {
--- a/test/functional/load-balancer-spec.test.js
+++ b/test/functional/load-balancer-spec.test.js
@@ -15,16 +15,16 @@ const SKIP = [
   //       in sub-optimal pinning.
   'pinned connections are not returned after an network error during getMore',
   'pinned connections are not returned to the pool after a non-network error on getMore',
-  'stale errors are ignored'
-];
+  'stale errors are ignored',
 
-require('../functional/retryable_reads.test');
-require('../functional/retryable_writes.test');
-require('../functional/uri_options_spec.test');
-require('../functional/change_stream_spec.test');
-require('../functional/versioned-api.test');
-require('../unit/core/mongodb_srv.test');
-require('../unit/sdam/server_selection/spec.test');
+  // NOTE: The driver correctly fails these 2 tests in non LB mode due to validation that
+  // loadBalanced=true is only valid as a URI option and not a regular driver option according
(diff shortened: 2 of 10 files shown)


nbbeeken (approved):

test/functional/load-balancer-spec.test.js
```diff
@@ -15,16 +15,16 @@ const SKIP = [
...
-require('../functional/retryable_reads.test');
-require('../functional/retryable_writes.test');
-require('../functional/uri_options_spec.test');
-require('../functional/change_stream_spec.test');
-require('../functional/versioned-api.test');
-require('../unit/core/mongodb_srv.test');
-require('../unit/sdam/server_selection/spec.test');
+  // NOTE: The driver correctly fails these 2 tests in non LB mode due to validation that
+  // loadBalanced=true is only valid as a URI option and not a regular driver option according
+  // to the spec. However the unified test runner implementation incorrectly passes URI options
```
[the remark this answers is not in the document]
nbbeeken: I think we can make the uri fix here, I can do that.

Summarizing from our slack:
- the subset was the required tests in the LB spec but we can expand to run everything for more coverage
- SDAM testing is passing by being unit tests that specify URIs without LB enabled

src/db.ts
```diff
@@ -320,6 +320,10 @@ export class Db {
    * @returns return the new Collection instance
    */
   collection<TSchema extends Document = Document>(name: string): Collection<TSchema>;
+  collection<TSchema extends Document = Document>(
```
dariakp, line 323: just curious why this was required

test/functional/unified-spec-runner/entities.ts
```diff
@@ -102,12 +104,21 @@ export class UnifiedMongoClient extends MongoClient {
...
+      connectionString.searchParams.set(name, String(value));
+    }
+    return connectionString.toString();
+  }
+
   constructor(url: string, description: ClientEntity) {
-    super(url, {
+    super(UnifiedMongoClient.connectionString(url, description.uriOptions), {
       monitorCommands: true,
       ...description.uriOptions,
```
dariakp, line 118: does it actually make sense to still pass in the uriOptions separately?

test/functional/load-balancer-spec.test.js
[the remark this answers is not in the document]
dariakp: With the URI options parsed correctly, can we uncomment these tests now?

test/functional/unified-spec-runner/entities.ts
dariakp: does the unified runner spec have a concept of non-uri client options?

src/db.ts
nbbeeken: This is a bug, when you have any overrides then you need to enumerate all the possibilities above the actual method declaration, so currently TS doesn't allow options to be passed in.

test/functional/unified-spec-runner/entities.ts
nbbeeken: Oop no that's an over sight we should just have the options in the uri now. There's nothing in the schema for unified runner for general options via the second param.

src/db.ts
dariakp: yeah in this case I'm not sure why we want to have overrides instead of just the optional options definition... imho fewer definitions is better, but not a hill I'm willing to die on if you have strong opinions otherwise

emadum (approved):

src/db.ts
nbbeeken: GH is not showing as outdated but I changed it to:
```ts
  /**
   * Returns a reference to a MongoDB Collection. If it does not exist it will be created implicitly.
   *
   * @param name - the collection name we wish to access.
   * @returns return the new Collection instance
   */
  collection<TSchema extends Document = Document>(
    name: string,
    options: CollectionOptions = {}
  ): Collection<TSchema> {
    if (typeof options === 'function') {
      // TODO(NODE-3485): Replace this with MongoDeprecationError
      throw new MongoDriverError('The callback form of this helper has been removed.');
    }
    const finalOptions = resolveOptions(this, options);
    return new Collection<TSchema>(this, name, finalOptions);
  }
 ```

File: test/functional/load-balancer-spec.test.js [hunk 4]
```diff
@@ -15,17 +15,15 @@ const SKIP = [
   //       in sub-optimal pinning.
   'pinned connections are not returned after an network error during getMore',
   'pinned connections are not returned to the pool after a non-network error on getMore',
-  'stale errors are ignored'
+  'stale errors are ignored',
+  // NOTE: The driver correctly fails these 2 tests in non LB mode due to validation that
+  // loadBalanced=true is only valid as a URI option and not a regular driver option according
+  // to the spec. In versions that are 3.4 or less an error still occurs but a different one (connection closes).
+  // TODO(NODE-3543): fix the path-ing that will produce errors for older servers
```
nbbeeken, line 22: Brought back the skip and made a ticket for the issue.

test/functional/unified-spec-runner/entities.ts
```diff
@@ -329,7 +330,10 @@ export class EntitiesMap<E = Entity> extends Map<string, E> {
         const useMultipleMongoses =
           (config.topologyType === 'LoadBalanced' || config.topologyType === 'Sharded') &&
           entity.client.useMultipleMongoses;
-        const uri = config.url({ useMultipleMongoses });
+        const uri = makeConnectionString(
```
nbbeeken, line 333: Moving this out of the constructor assisted with debugging, there was a bit too much inlined before

test/functional/unified-spec-runner/match.ts
```diff
@@ -326,29 +326,31 @@ export function expectErrorCheck(
     return;
   }
 
+  const expectMessage = `\n\nOriginal Error Stack:\n${error.stack}\n\n`;
```
nbbeeken, line 329: When getting some matching issue with any error I think it'll be handy to print out where the error came from, was using this to debug but seems good to keep.

test/functional/unified-spec-runner/schema.ts
```diff
@@ -96,7 +97,7 @@ export type EntityDescription =
   | { stream: StreamEntity }
   | { session: SessionEntity };
 export interface ServerApi {
-  version: string;
+  version: ServerApiVersion;
```
nbbeeken, line 100: This just fixes a TS red underline

test/tools/runner/index.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+require('source-map-support').install({
+  hookRequire: true
+});
```
nbbeeken, line 5: Makes the sourcemaps work with ts-node correctly

test/functional/load-balancer-spec.test.js
dariakp: can we remove the no longer true bit of the note concerning the URI option parsing?

dariakp (changes requested):

test/functional/load-balancer-spec.test.js
nbbeeken: done!

dariakp (approved):

PR #2934, opened 2021-08-10
docs(NODE-3481): deprecate unref in 4.1

andymina:
## Description

Unref is now deprecated and marked to be removed in the next major version.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -701,7 +701,10 @@ export class Db {
     );
   }
 
-  /** Unref all sockets */
+  /**
+   * Unref all sockets
+   * @deprecated This function is deprecated and will be removed in the 5.0 driver.
+   */
   unref(): void {
     getTopology(this).unref();
   }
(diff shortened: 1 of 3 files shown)


src/db.ts
```diff
@@ -701,7 +701,10 @@ export class Db {
     );
   }
 
-  /** Unref all sockets */
+  /**
+   * Unref all sockets
+   * @deprecated This function is deprecated and will be removed in the 5.0 driver.
```
nbbeeken, line 706:
```suggestion
   * @deprecated This function is deprecated and will be removed in the next major version.
```

src/sdam/topology.ts
```diff
@@ -770,8 +770,11 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
     return this.s.state === STATE_CLOSED;
   }
 
+  /**
+   * @deprecated This function is deprecated and will be removed in the 5.0 driver.
```
nbbeeken, line 774: same fix ^

nbbeeken (changes requested):

nbbeeken (approved):

PR #2935, opened 2021-08-11
fix(NODE-3468): remove generic overrides from find

nbbeeken:
Removing the generics from `find*` and `toArray/next` will make the result type more logical in scenarios where there is not schema or the filter specified is an incorrect type. 

Removed the same behavior from listCollections
Added a generic to db.aggregate where it should have been

Q: do we want to do something better with specifying a projection in your options argument to `find*` right now you can only "fix" the result with the `.project<>()` builder style, maybe we continue to accept generics on find*. but we no longer make it `Filter<T>`, the filter still has to be based on the schema (`Filter<TSchema>`) but the result can be changed to the specified type in the generic. thoughts?


src/cursor/abstract_cursor.ts
```diff
@@ -295,9 +294,9 @@ export abstract class AbstractCursor<
   }
 
   /** Get the next available document from the cursor, returns null if no more documents are available. */
-  next<T = TSchema>(): Promise<T | null>;
-  next<T = TSchema>(callback: Callback<T | null>): void;
-  next<T = TSchema>(callback?: Callback<T | null>): Promise<T | null> | void {
+  next(): Promise<TSchema | null>;
+  next(callback: Callback<TSchema | null>): void;
```
PaulEndri, line 298: Hi! I'm a bit confused as to the purpose of the explicit `| null` being used?

PaulEndri (commented):
Regarding your question re: the `.find` functionality: I personally would find it super helpful. Since the result of the `.find` can actually vary significantly, it makes sense to be able to type the final output while making sure the filter follows the schema. Just my .02

src/db.ts
```diff
@@ -289,7 +289,10 @@ export class Db {
    * @param pipeline - An array of aggregation stages to be executed
    * @param options - Optional settings for the command
    */
-  aggregate(pipeline: Document[] = [], options?: AggregateOptions): AggregationCursor {
+  aggregate<T = Document>(
```
PaulEndri, line 292: Should this default to Document? The result of this call could very well be grossly incompatible with an object, and even if it's not it may lead to some confusion since the result is more likely than not to not have all the functionality/expectations of a full document, if my understanding of what the Document type exposes is correct.

src/cursor/abstract_cursor.ts
```diff
@@ -413,7 +412,7 @@ export abstract class AbstractCursor<
           // these do need to be transformed since they are copying the rest of the batch
           const internalDocs = (transform
             ? this[kDocuments].splice(0, this[kDocuments].length).map(transform)
-            : this[kDocuments].splice(0, this[kDocuments].length)) as T[]; // TODO(NODE-3283): Improve transform typing
+            : this[kDocuments].splice(0, this[kDocuments].length)) as TSchema[]; // TODO(NODE-3283): Improve transform typing
```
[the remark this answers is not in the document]
nbbeeken: That is the type of `this[kDocuments]` although you did make me think of a potential improvement (really just internally for us) but if we instead type `this[kTransform]` to `(doc: TSchema) => TSchema` then we don't need these casts. 

When you pass the transform into `.map()` it will be `(doc: TSchema) => T` but map will return a new `Cursor<T>`. Where `T` is from the result type of the function passed to `.map`. So the argument that goes into `this[kTransform]` is the old schema and it returns the "current" one.

I suppose the most accurate would be `this[kTransform]: (doc: unknown) => TSchema` but then we need equivalent casting/whatever to narrow unknown

src/cursor/abstract_cursor.ts
nbbeeken: Our next functions explicitly return null when the end of the cursor is reached. Our callback definition only lets the second are be potentially undefined, so the `| null` is very technically necessary, if one is triple equals checking for null.

src/cursor/abstract_cursor.ts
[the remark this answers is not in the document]
PaulEndri: Well my original comment was because I didn't see a type on kDocument and was thinking about ways to get rid of the `as` because `this[kDocuments].splice` should automatically return something with the type of `<Type of<this[kDocuments]>>[]` so in this case if the source array was already typed, then this would automatically be TSchema[] so the typing here made me curious.

I actually played around a lot with prototyping a change to ktransform and others when thinking about ways to be able to type projections fully, and narrowing that down was like, a several hundred line TS object that existed only for being able to pass around the type accurately, and I think the `unknown` above might run into the same situation?  Although would you be able to use the `Partial` keyword here for some help? I like the idea of the different typing, just wondering if the effort required to type that unknown is worth it

src/cursor/abstract_cursor.ts
PaulEndri: Makes sense. The following is purely pedantry and feel free to ignore it, because I likely would if I were you lol.

For transparency, do think a helper type here would be useful? So instead of showing the | null it would be `Promise<EndOfCursor<TSchema>>` where EndOfCursor is defined as `T | null`? Again, super minute and what not. Marking this comment as resolved so it's out of mind lol. Thanks for answering my questions!

test/types/list_collections.test-d.ts
```diff
@@ -8,38 +8,30 @@ const db = new MongoClient('').db();
...
 
 // We default to the CollectionInfo result type
-expectType<ListCollectionsCursor<Pick<CollectionInfo, 'name' | 'type'> | CollectionInfo>>(
-  db.listCollections()
-);
+expectType<ListCollectionsCursor<CollectionInfo>>(db.listCollections());
 // By default it isn't narrowed to either type
 expectNotType<ListCollectionsCursor<Pick<CollectionInfo, 'name' | 'type'>>>(db.listCollections());
-expectNotType<ListCollectionsCursor<CollectionInfo>>(db.listCollections());
+expectType<ListCollectionsCursor<CollectionInfo>>(db.listCollections());
```
dariakp, line 14: I think these two tests should just be removed, because they are confusing (this one and the one above it), the comment in L12 is no longer relevant

test/types/list_collections.test-d.ts
```diff
@@ -8,38 +8,30 @@ const db = new MongoClient('').db();
...
 // note the nameOnly: false, yet strings are returned
-const overridden = await db
-  .listCollections<Pick<CollectionInfo, 'name' | 'type'>>({}, { nameOnly: false })
-  .toArray();
-expectType<Pick<CollectionInfo, 'name' | 'type'>[]>(overridden);
-const overriddenWithToArray = await db
-  .listCollections({}, { nameOnly: false })
-  .toArray<Pick<CollectionInfo, 'name' | 'type'>>();
-expectType<Pick<CollectionInfo, 'name' | 'type'>[]>(overriddenWithToArray);
+// NODE-3468: No longer permits overriding of return type via a generic
```
dariakp, line 37: Can we remove the 3 comment lines above since they make no sense without the corresponding tests?

test/types/list_collections.test-d.ts
```diff
@@ -8,38 +8,30 @@ const db = new MongoClient('').db();
...
+expectType<ListCollectionsCursor<CollectionInfo>>(db.listCollections());
 
 // Testing each argument variation
 db.listCollections();
 db.listCollections({ a: 2 });
 db.listCollections({ a: 2 }, { batchSize: 2 });
 
 const collections = await db.listCollections().toArray();
-expectType<EitherCollectionInfoResult[]>(collections);
+expectNotType<EitherCollectionInfoResult[]>(collections);
```
dariakp, line 22: I think adding an expectType ahead of this line would help make the intention more clear

test/types/list_collections.test-d.ts
```diff
@@ -8,38 +8,30 @@ const db = new MongoClient('').db();
...
 
 const nameOnly = await db.listCollections({}, { nameOnly: true }).toArray();
 expectType<Pick<CollectionInfo, 'name' | 'type'>[]>(nameOnly);
 
 const fullInfo = await db.listCollections({}, { nameOnly: false }).toArray();
 expectType<CollectionInfo[]>(fullInfo);
 
-const couldBeEither = await db.listCollections({}, { nameOnly: Math.random() > 0.5 }).toArray();
-expectType<EitherCollectionInfoResult[]>(couldBeEither);
+const cannotBeEither = await db.listCollections({}, { nameOnly: Math.random() > 0.5 }).toArray();
```
dariakp, line 30: hmm, so typescript can't automagically figure it out anymore?

dariakp (changes requested):
I need to review more of this, but I wanted to capture at least some feedback

test/types/community/collection/findX.test-d.ts
```diff
@@ -102,6 +102,23 @@ expectType<{ cost: number } | undefined>(
...
+type Pet = { type: string; age: number };
+const pets = db.collection<Pet>('pets');
+
+expectType<{ crazy: number }[]>(
+  await pets
+    .find<{ crazy: number }>({ type: 'dog', age: 1 })
+    .toArray()
+);
+
+expectType<{ 'dot.keys': number }[]>(
```
dariakp, line 116: I think we can remove this example and the corresponding comment in L106; however, I think it's worth adding a dot key case to the filterQuery  test file to show that it is assignable even when the referenced key is the wrong type

dariakp (changes requested):

test/types/community/collection/findX.test-d.ts
nbbeeken: Added a test to the bottom of filterQuery test file

src/cursor/abstract_cursor.ts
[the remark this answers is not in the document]
nbbeeken: Sorry for the lack of reply, been focused on getting this PR ready for fixing the user issues. I think accurately typing the transform is something we should look into, but it would only affect our internal TS code so at the moment we wouldn't need to prioritize it. but going forward more accuracy even if it requires TS helpers etc. could help catch any issues here.

src/cursor/abstract_cursor.ts
nbbeeken: It's true we have this repeated in a number of places, consolidating it to type would help us make sure we don't get one override different from the others, but I'll defer that to an enhancement effort down the line for now I want this PR to unblock the issue with the generics, but I do appreciate the suggestion! thanks!

emadum (approved):
Nice work! 🚀

dariakp (approved):

PR #2939, opened 2021-08-17
fix(NODE-3528): add support for snappy 7

nbbeeken:
Pulling this fix out of the larger deps upgrade PR.

Reworked the bson-ext and custom fle build variants to share the custom tasks. cleans up the EVG view a bit.

Test both snappy 6 and 7.

also upgrades our connection string dep

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,16 +31,10 @@
     "name": "The MongoDB NodeJS Team",
     "email": "dbx-node@mongodb.com"
   },
-  "peerOptionalDependencies": {
-    "kerberos": "^1.1.0",
-    "mongodb-client-encryption": "^1.0.0",
-    "snappy": "^6.1.1",
-    "bson-ext": "^2.0.0"
-  },
   "dependencies": {
     "bson": "^4.5.0",
     "denque": "^1.5.0",
-    "mongodb-connection-string-url": "^1.0.1"
+    "mongodb-connection-string-url": "^2.0.0"
   },
   "devDependencies": {
     "@istanbuljs/nyc-config-typescript": "^1.0.1",
@@ -82,7 +76,6 @@
     "semver": "^5.5.0",
     "sinon": "^4.3.0",
     "sinon-chai": "^3.2.0",
-    "snappy": "^6.3.0",
     "source-map-support": "^0.5.19",
     "standard-version": "^9.3.0",
     "through2": "^3.0.1",
--- a/src/cmap/wire_protocol/compression.ts
+++ b/src/cmap/wire_protocol/compression.ts
@@ -31,8 +31,6 @@ export const uncompressibleCommands = new Set([
   'copydb'
 ]);
 
-let SNAPPY_VERSION: number | undefined = undefined;
-
 // Facilitate compressing a message using an agreed compressor
 export function compress(
   self: { options: OperationDescription & zlib.ZlibOptions },
@@ -46,11 +44,7 @@ export function compress(
         return callback(Snappy['kModuleError']);
       }
 
-      if (!SNAPPY_VERSION) {
-        SNAPPY_VERSION = Number.parseInt(Snappy[PKG_VERSION].split('.')[0]);
-      }
-
-      if (SNAPPY_VERSION <= 6) {
+      if (Snappy[PKG_VERSION].major <= 6) {
         Snappy.compress(dataToBeCompressed, callback);
       } else {
         Snappy.compress(dataToBeCompressed)
@@ -91,11 +85,7 @@ export function decompress(
         return callback(Snappy['kModuleError']);
       }
 
-      if (!SNAPPY_VERSION) {
-        SNAPPY_VERSION = Number.parseInt(Snappy[PKG_VERSION].split('.')[0]);
-      }
-
-      if (SNAPPY_VERSION <= 6) {
+      if (Snappy[PKG_VERSION].major <= 6) {
         Snappy.uncompress(compressedData, { asBuffer: true }, callback);
       } else {
         Snappy.uncompress(compressedData, { asBuffer: true })
--- a/src/cmap/wire_protocol/compression.ts
+++ b/src/cmap/wire_protocol/compression.ts
@@ -85,7 +85,7 @@ export function decompress(
       const snappyResult = Snappy.uncompress(compressedData, { asBuffer: true }, callback);
 
       if (isPromiseLike(snappyResult)) {
-        // Using snappy 6.x
+        // Using snappy 7.x
         snappyResult.then(buffer => callback(undefined, buffer)).catch(error => callback(error));
       }
       break;
--- a/src/cmap/wire_protocol/compression.ts
+++ b/src/cmap/wire_protocol/compression.ts
@@ -1,8 +1,8 @@
 import * as zlib from 'zlib';
-import { Callback, isPromiseLike } from '../../utils';
+import type { Callback } from '../../utils';
 import type { OperationDescription } from '../message_stream';
 
-import { Snappy } from '../../deps';
+import { PKG_VERSION, Snappy } from '../../deps';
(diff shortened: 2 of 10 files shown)


src/cmap/wire_protocol/compression.ts
```diff
@@ -39,12 +39,18 @@ export function compress(
 ): void {
   const zlibOptions = {} as zlib.ZlibOptions;
   switch (self.options.agreedCompressor) {
-    case 'snappy':
+    case 'snappy': {
       if ('kModuleError' in Snappy) {
         return callback(Snappy['kModuleError']);
       }
-      Snappy.compress(dataToBeCompressed, callback);
+      const snappyResult = Snappy.compress(dataToBeCompressed, callback);
```
nbbeeken, line 46: Should we check function length here as well?

package.json
```diff
@@ -31,16 +31,10 @@
     "name": "The MongoDB NodeJS Team",
     "email": "dbx-node@mongodb.com"
   },
-  "peerOptionalDependencies": {
-    "kerberos": "^1.1.0",
-    "mongodb-client-encryption": "^1.0.0",
-    "snappy": "^6.1.1",
-    "bson-ext": "^2.0.0"
-  },
```
nbbeeken, line 39: This is unused IIUC

dariakp: I'm a bit confused, isn't this what specifies the compatible versions of these dependencies? e.g., for us it's snappy 6 or 7?

test/unit/snappy.test.js
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  after(async function () {
+    await mock.cleanup();
+    await client.close();
+  });
+
+  describe('Snappy', () => {
+    it(`should compress messages sent with snappy ${snappyVersion}`, async function () {
+      // This test should not take any longer
```
dariakp, line 32: missing the rest of the sentence?

test/unit/snappy.test.js
```diff
@@ -0,0 +1,52 @@
...
+      // The failure case is hanging because new snappy never calls the callback
```
dariakp, line 33: what failure case?

test/unit/snappy.test.js
```diff
@@ -0,0 +1,52 @@
...
+        if (doc.ismaster || doc.hello) {
+          return request.reply({ ...mock.DEFAULT_ISMASTER, compression: ['snappy'] });
+        }
+        if (doc.insert === 'snappy') {
+          return request.reply({ ok: 1 });
+        }
+      });
+      // The mock server uses snappy to decode messages so
+      // if this passes we implicitly test snappy is working
+      // TODO(NODE-XXXX): Add more comprehensive round trip testing
```
dariakp, line 47: insert ticket id

test/unit/snappy.test.js
dariakp: I think maybe these two comments mean that the timeout is being set because the test shouldn't take any longer than 5 seconds, and that if it doesn't complete, it will hang due to the callback never being called, is that right?

dariakp (changes requested):

src/cmap/wire_protocol/compression.ts
dariakp: not sure what you are asking

src/cmap/wire_protocol/compression.ts
```diff
@@ -72,12 +78,18 @@ export function decompress(
...
-    case Compressor.snappy:
+    case Compressor.snappy: {
       if ('kModuleError' in Snappy) {
         return callback(Snappy['kModuleError']);
       }
-      Snappy.uncompress(compressedData, { asBuffer: true }, callback as Callback);
+      const snappyResult = Snappy.uncompress(compressedData, { asBuffer: true }, callback);
+
+      if (isPromiseLike(snappyResult)) {
+        // Using snappy 6.x
```
dariakp, line 88: wouldn't this be using snappy 7?

src/cmap/wire_protocol/compression.ts
nbbeeken: We could assert the function length is what we expect it to be before assuming the last arg takes a callback, but might be brittle?

```js
(function(){}).length === 0
(function(a){}).length === 1
(function(a, b){}).length === 2
```

package.json
nbbeeken: only implicitly in the sense that it happens to have some versions numbers which are kinda accurate (bson-ext is wrong). The `peerOptionalDependencies` was used in 3.5 by our old [require_optional dependency](https://github.com/christkv/require_optional/blob/c11c4ec54c29d9f5b8a335db379a4670ceeb6cfb/index.js#L51). It's not a standard field and it isn't respected by any npm install logic, so I think dropping it, maybe in favor of a support table in our readme or docs would be better, thoughts?

test/unit/snappy.test.js
nbbeeken: Oh yea, should not take any longer than 5 sec, I didn't have the comment below at first so I figured this.timeout finished my thought, whoops!

test/unit/snappy.test.js
nbbeeken: Yep that's all correct, I can clarify all that.

src/cmap/wire_protocol/compression.ts
dariakp: so the danger I am seeing with assuming that the compress parameter takes a callback is that a future snappy might be updated to do something different with that second argument... it would be nice if we could determine the snappy version before calling the function

package.json
dariakp: do we have an optional dependencies file that's a single point of entry for all these? because that's where I'd want to specify these; alternatively, I suppose a table in the README for an optional dependencies section would not be out of place since I guess this is how we tell users these optional customizations? (let's not include bson-ext there at all though, it has worse performance than js-bson)

src/cmap/wire_protocol/compression.ts
nbbeeken: I added some logic to check the dependency version, let me know what you think.

src/deps.ts
```diff
@@ -52,6 +92,9 @@ export let Snappy:
 try {
   // Ensure you always wrap an optional require in the try block NODE-3199
   Snappy = require('snappy');
+  try {
+    (Snappy as any)[PKG_VERSION] = require('snappy/package.json').version;
```
dariakp, line 96: why not convert it to number here, so that we don't have to do the independent parsing logic up  in the other file? if the idea is to future proof it, you could parse the major/minor/patch right here and store in an object, but that's probably overkill, it's not like we're using it all over the repo, so parsing it directly to major for now is probably fine

dariakp (approved):

test/unit/snappy.test.js
```diff
@@ -0,0 +1,58 @@
...
+      // The mock server uses snappy to decode messages so
+      // if this passes we implicitly test snappy is working
+      // TODO(NODE-3560): Add more comprehensive round trip testing
+      await client.connect();
+      await client.db().collection('snappy').insertOne({ a: 1 });
+    });
+
+    it('should define a version number on the optional import', function () {
+      const { Snappy, PKG_VERSION } = require('../../src/deps');
+      expect(Snappy).to.have.property(PKG_VERSION, snappyVersion);
```
dariakp, line 55: is this still passing after making it an object?

nbbeeken: fixed!

dariakp (approved):

package.json
nbbeeken: I made NODE-3562 which recommends something more than a table in our README, but could be used to track that, sound good?

dariakp: yeah, looks good, tyvm!

emadum (approved):
LGTM

PR #2941, opened 2021-08-19
refactor(NODE-3554): Prune MongoDriverError instances

W-A-James:
## Description

Remove all instances of MongoDriverError and replace them with either MongoAPIError or MongoRuntimeError

### DIFF
--- a/src/cmap/auth/auth_provider.ts
+++ b/src/cmap/auth/auth_provider.ts
@@ -3,7 +3,7 @@ import type { Connection, ConnectionOptions } from '../connection';
 import type { MongoCredentials } from './mongo_credentials';
 import type { HandshakeDocument } from '../connect';
 import type { ClientMetadataOptions, Callback } from '../../utils';
-import { MongoDriverError } from '../../error';
+import { MongoAPIError } from '../../error';
 
 export type AuthContextOptions = ConnectionOptions & ClientMetadataOptions;
 
@@ -55,6 +55,6 @@ export class AuthProvider {
    */
   auth(context: AuthContext, callback: Callback): void {
     // TODO(NODE-3485): Replace this with MongoMethodOverrideError
-    callback(new MongoDriverError('`auth` method must be overridden by subclass'));
+    callback(new MongoAPIError('`auth` method must be overridden by subclass'));
   }
 }
--- a/src/cmap/auth/auth_provider.ts
+++ b/src/cmap/auth/auth_provider.ts
@@ -3,7 +3,7 @@ import type { Connection, ConnectionOptions } from '../connection';
 import type { MongoCredentials } from './mongo_credentials';
 import type { HandshakeDocument } from '../connect';
 import type { ClientMetadataOptions, Callback } from '../../utils';
-import { MongoAPIError } from '../../error';
+import { MongoRuntimeError } from '../../error';
 
 export type AuthContextOptions = ConnectionOptions & ClientMetadataOptions;
 
@@ -54,7 +54,7 @@ export class AuthProvider {
    * @param callback - The callback to return the result from the authentication
    */
   auth(context: AuthContext, callback: Callback): void {
-    // TODO(NODE-3485): Replace this with MongoMethodOverrideError
-    callback(new MongoAPIError('`auth` method must be overridden by subclass'));
+    // TODO(NODE-3483): Replace this with MongoMethodOverrideError
+    callback(new MongoRuntimeError('`auth` method must be overridden by subclass'));
   }
 }
--- a/src/cmap/auth/mongo_credentials.ts
+++ b/src/cmap/auth/mongo_credentials.ts
@@ -1,7 +1,7 @@
 // Resolves the default auth mechanism according to
 
 import type { Document } from '../../bson';
-import { MongoRuntimeError, MongoMissingCredentialsError } from '../../error';
+import { MongoAPIError, MongoMissingCredentialsError } from '../../error';
 import { AuthMechanism } from './defaultAuthProviders';
 
 // https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst
@@ -131,16 +131,16 @@ export class MongoCredentials {
       this.mechanism === AuthMechanism.MONGODB_X509
     ) {
       if (this.source != null && this.source !== '$external') {
-        // TODO(NODE-3483): Replace this with a MongoAuthValidationError
-        throw new MongoRuntimeError(
+        // TODO(NODE-3485): Replace this with a MongoAuthValidationError
+        throw new MongoAPIError(
           `Invalid source '${this.source}' for mechanism '${this.mechanism}' specified.`
         );
       }
     }
 
     if (this.mechanism === AuthMechanism.MONGODB_PLAIN && this.source == null) {
-      // TODO(NODE-3483): Replace this with a MongoAuthValidationError
-      throw new MongoRuntimeError('PLAIN Authentication Mechanism needs an auth source');
+      // TODO(NODE-3485): Replace this with a MongoAuthValidationError
+      throw new MongoAPIError('PLAIN Authentication Mechanism needs an auth source');
     }
 
     if (this.mechanism === AuthMechanism.MONGODB_X509 && this.password != null) {
@@ -148,8 +148,8 @@ export class MongoCredentials {
         Reflect.set(this, 'password', undefined);
         return;
       }
-      // TODO(NODE-3483): Replace this with a MongoAuthValidationError
-      throw new MongoRuntimeError(`Password not allowed for mechanism MONGODB-X509`);
+      // TODO(NODE-3485): Replace this with a MongoAuthValidationError
+      throw new MongoAPIError(`Password not allowed for mechanism MONGODB-X509`);
     }
   }
 
--- a/src/cmap/auth/mongo_credentials.ts
+++ b/src/cmap/auth/mongo_credentials.ts
@@ -1,7 +1,7 @@
 // Resolves the default auth mechanism according to
 
 import type { Document } from '../../bson';
-import { MongoDriverError, MongoMissingCredentialsError } from '../../error';
+import { MongoRuntimeError, MongoMissingCredentialsError } from '../../error';
 import { AuthMechanism } from './defaultAuthProviders';
 
 // https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst
@@ -132,7 +132,7 @@ export class MongoCredentials {
     ) {
       if (this.source != null && this.source !== '$external') {
         // TODO(NODE-3483): Replace this with a MongoAuthValidationError
-        throw new MongoDriverError(
+        throw new MongoRuntimeError(
           `Invalid source '${this.source}' for mechanism '${this.mechanism}' specified.`
         );
       }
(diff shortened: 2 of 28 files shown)


src/cmap/auth/mongo_credentials.ts
```diff
@@ -132,15 +132,15 @@ export class MongoCredentials {
     ) {
       if (this.source != null && this.source !== '$external') {
         // TODO(NODE-3483): Replace this with a MongoAuthValidationError
-        throw new MongoDriverError(
+        throw new MongoRuntimeError(
```
dariakp, line 135: Is the `validate()` function only used internally? Feels like the errors here could be API, not runtime

src/operations/connect.ts
```diff
@@ -41,7 +41,8 @@ export function connect(
       }
 
       // Return a more specific error message for MongoClient.connect
-      return callback(new MongoDriverError(warningMessage));
+      // TODO(NODE-3483)
+      return callback(new MongoRuntimeError(warningMessage));
```
dariakp, line 45: are we sure this is not an API error? (given the message is about the URI)

src/sdam/srv_polling.ts
```diff
@@ -67,8 +67,7 @@ export class SrvPoller extends TypedEventEmitter<SrvPollerEvents> {
     super();
 
     if (!options || !options.srvHost) {
-      // TODO(NODE-3483)
-      throw new MongoDriverError('Options for SrvPoller must exist and include srvHost');
+      throw new MongoInvalidArgumentError('Options for SrvPoller must exist and include srvHost');
```
dariakp, line 70: I think this one might be runtime, since the user can't directly instantiate SrvPoller

src/sdam/topology.ts
```diff
@@ -282,7 +283,7 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
         seedlist.push(seed);
       } else {
         // FIXME(NODE-3484): May need to be a MongoParseError
-        throw new MongoDriverError(`Topology cannot be constructed from ${JSON.stringify(seed)}`);
+        throw new MongoParseError(`Topology cannot be constructed from ${JSON.stringify(seed)}`);
```
dariakp, line 286: I think this one may be runtime

src/sdam/topology_description.ts
```diff
@@ -438,7 +438,7 @@ function updateRsWithPrimaryFromMember(
 ): TopologyType {
   if (setName == null) {
     // TODO(NODE-3483): should be an appropriate runtime error
-    throw new MongoDriverError('Argument "setName" is required if connected to a replica set');
+    throw new MongoRuntimeError('Argument "setName" is required if connected to a replica set');
```
dariakp, line 441: is setName user specified? if not, this would be a runtime error

src/sessions.ts
```diff
@@ -149,12 +150,12 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
 
     if (topology == null) {
       // TODO(NODE-3483)
-      throw new MongoDriverError('ClientSession requires a topology');
+      throw new MongoAPIError('ClientSession requires a topology');
```
dariakp, line 153: I think this is runtime

src/sessions.ts
```diff
@@ -842,7 +843,7 @@ export class ServerSessionPool {
 
   constructor(topology: Topology) {
     if (topology == null) {
-      throw new MongoDriverError('ServerSessionPool requires a topology');
+      throw new MongoRuntimeError('ServerSessionPool requires a topology');
```
dariakp, line 846: I think this is runtime

src/transactions.ts
```diff
@@ -164,7 +164,7 @@ export class Transaction {
       return;
     }
 
-    throw new MongoDriverError(
+    throw new MongoAPIError(
```
dariakp, line 167: this is almost definitely a runtime error

src/utils.ts
```diff
@@ -242,7 +243,7 @@ export function executeLegacyOperation(
 
   if (!Array.isArray(args)) {
     // TODO(NODE-3483)
-    throw new MongoDriverError('This method requires an array of arguments to apply');
+    throw new MongoAPIError('This method requires an array of arguments to apply');
```
dariakp, line 246: this function is for internal use, so this is runtime

src/utils.ts
```diff
@@ -304,7 +305,7 @@ export function executeLegacyOperation(
   // Return a Promise
   if (args[args.length - 1] != null) {
     // TODO(NODE-3483)
-    throw new MongoDriverError('Final argument to `executeLegacyOperation` must be a callback');
+    throw new MongoAPIError('Final argument to `executeLegacyOperation` must be a callback');
```
dariakp, line 308: I don't think executeLegacyOperation can be called directly, so this is runtime

src/utils.ts
```diff
@@ -590,7 +591,7 @@ export class MongoDBNamespace {
   static fromString(namespace?: string): MongoDBNamespace {
     if (!namespace) {
       // TODO(NODE-3483): Replace with MongoNamespaceError
-      throw new MongoDriverError(`Cannot parse namespace from "${namespace}"`);
+      throw new MongoAPIError(`Cannot parse namespace from "${namespace}"`);
```
dariakp, line 594: this is runtime

dariakp (changes requested):
I called out some errors that are definitely runtime, and then we need to double check some of the other cases to make sure we are applying the right type

src/cmap/auth/mongo_credentials.ts
nbbeeken: I think these are all API, this function (I believe) is only called during options parsing so it would be sometime identified at startup essentially

src/sdam/srv_polling.ts
W-A-James: For this one I called it a MongoInvalidArgumentError specifically because it gets instantiated in the constructor for `Topology` (src/sdam/topology.ts) which takes the options argument directly from the  `connect` function (src/operations/connect.ts:65).

I may be misunderstanding that `connect` function though, so totally open to fix that.

src/sdam/topology_description.ts
W-A-James: It seems that `setName` is created when `TopologyDescription` is instantiated and that class is exported in src/index.ts, so I'll go ahead and change this to `MongoAPIError`

src/sessions.ts
```diff
@@ -965,14 +966,14 @@ export function applySession(
...
-    return new MongoDriverError('Unable to acquire server session');
+    return new MongoRuntimeError('Unable to acquire server session');
   }
 
   // SPEC-1019: silently ignore explicit session with unacknowledged write for backwards compatibility
   // FIXME: NODE-2781, this check for write concern shouldn't be happening here, but instead during command construction
   if (options && options.writeConcern && (options.writeConcern as WriteConcern).w === 0) {
     if (session && session.explicit) {
-      return new MongoDriverError('Cannot have explicit session with unacknowledged writes');
+      return new MongoRuntimeError('Cannot have explicit session with unacknowledged writes');
```
nbbeeken, line 976: This is incorrect usage, A user providing their own session with WC = 0 isn't permitted.

src/sdam/srv_polling.ts
dariakp: Yeah it's a bit convoluted, but that connect function is actually invoked through the MongoClient constructor (it's imported in the mongo client file and then used by its public connect method), however the options passed to it are the ones parsed in when calling the MongoClient constructor, and that's where the user-facing validation error resides (see L71 in connection_string).. so, by the time we get to this code, it's our fault for not catching it because the user shouldn't really have to worry about what an SrvPoller is

src/utils.ts
nbbeeken: this will throw if namespace is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). I don't believe we currently pass in a user specified namespace directly to this function, but if we did this would be an API error. We have ticket where we want to roll the collection and Db's name validation into this class somehow.

src/cmap/auth/auth_provider.ts
```diff
@@ -55,6 +55,6 @@ export class AuthProvider {
    */
   auth(context: AuthContext, callback: Callback): void {
     // TODO(NODE-3485): Replace this with MongoMethodOverrideError
-    callback(new MongoDriverError('`auth` method must be overridden by subclass'));
+    callback(new MongoAPIError('`auth` method must be overridden by subclass'));
```
nbbeeken, line 58: This would be runtime only because we don't expose auth provider to be overridden by users. Ideally this should be an abstract class and this method is abstract making TS force us to implement it. but for now we can just keep this error, but make it Runtime for accuracy.

src/sdam/topology_description.ts
dariakp: huh, I think I was originally reading it as an API error, but I'm pretty sure it is a runtime error, which it currently is in this code, so no need to change it

src/cmap/auth/mongo_credentials.ts
```diff
@@ -132,15 +132,15 @@ export class MongoCredentials {
...
           `Invalid source '${this.source}' for mechanism '${this.mechanism}' specified.`
         );
       }
     }
 
     if (this.mechanism === AuthMechanism.MONGODB_PLAIN && this.source == null) {
       // TODO(NODE-3483): Replace this with a MongoAuthValidationError
-      throw new MongoDriverError('PLAIN Authentication Mechanism needs an auth source');
+      throw new MongoRuntimeError('PLAIN Authentication Mechanism needs an auth source');
```
nbbeeken, line 143: API error, someone would need to request PLAIN (LDAP) auth but not specify a db for the authSource

src/cmap/auth/mongo_credentials.ts
```diff
@@ -149,7 +149,7 @@ export class MongoCredentials {
         return;
       }
       // TODO(NODE-3483): Replace this with a MongoAuthValidationError
-      throw new MongoDriverError(`Password not allowed for mechanism MONGODB-X509`);
+      throw new MongoRuntimeError(`Password not allowed for mechanism MONGODB-X509`);
```
nbbeeken, line 152: API error, someone would need to request MONGODB-X509 auth and provide a password. x509 relies on the TLS cert you provide so we prevent confusion by disallowing a password.

nbbeeken (changes requested):

src/sessions.ts
```diff
@@ -149,12 +150,12 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
     if (topology == null) {
       // TODO(NODE-3483)
-      throw new MongoDriverError('ClientSession requires a topology');
+      throw new MongoAPIError('ClientSession requires a topology');
     }
 
     if (sessionPool == null || !(sessionPool instanceof ServerSessionPool)) {
       // TODO(NODE-3483)
-      throw new MongoDriverError('ClientSession requires a ServerSessionPool');
+      throw new MongoAPIError('ClientSession requires a ServerSessionPool');
```
dariakp, line 158: this one is also runtime

src/sessions.ts
dariakp: might have commented on the wrong error here, resolving

src/sessions.ts
dariakp: ^ I believe this still needs to be updated to MongoAPIError

dariakp (changes requested):
@W-A-James I resolved my comments but Neal's still need to be addressed here

src/operations/is_capped.ts
```diff
@@ -27,7 +27,7 @@ export class IsCappedOperation extends AbstractOperation<boolean> {
       .toArray((err, collections) => {
         if (err || !collections) return callback(err);
         if (collections.length === 0) {
-          return callback(new MongoDriverError(`collection ${coll.namespace} not found`));
+          return callback(new MongoRuntimeError(`collection ${coll.namespace} not found`));
```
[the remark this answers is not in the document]
W-A-James: Would that be MongoInvalidArgumentError then, or should I keep it as MongoAPIError?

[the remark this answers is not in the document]
dariakp: I think keep it API for simplicity

dariakp (approved):

PR #2943, opened 2021-08-19
fix(NODE-3511): deprecate fullResponse and remove associated buggy code paths

dariakp:
## Description
NODE-3511

**What changed?**
Removing the never properly implemented `fullResponse` option from the codebase, also marking `CommandOptions` internal, since they are, in fact, internal use only. Also, removing the ability to externally affect the `fullResult` option (TS compilation would break if `fullResult` is passed in as a command option, but nothing is stopping regular js users from setting it and overriding our internal settings - we should confirm with shell that they aren't using it for some niche functionality, but otherwise this should be safe to disable)

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -63,6 +63,8 @@ const kDescription = Symbol('description');
 const kIsMaster = Symbol('ismaster');
 /** @internal */
 const kAutoEncrypter = Symbol('autoEncrypter');
+/** @internal */
+const kFullResult = Symbol('fullResult');
 
 /** @internal */
 export interface QueryOptions extends BSONSerializeOptions {
@@ -89,7 +91,7 @@ export interface CommandOptions extends BSONSerializeOptions {
   readPreference?: ReadPreferenceLike;
   raw?: boolean;
   monitoring?: boolean;
-  fullResult?: boolean;
+  [kFullResult]?: boolean;
   socketTimeoutMS?: number;
   /** Session to use for the operation */
   session?: ClientSession;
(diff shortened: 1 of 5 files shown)


src/cmap/connection.ts
```diff
@@ -773,7 +775,7 @@ function write(
     requestId: command.requestId,
     cb: callback,
     session: options.session,
-    fullResult: typeof options.fullResult === 'boolean' ? options.fullResult : false,
+    fullResult: !!options[kFullResult],
```
nbbeeken, line 778: RE: my other comment should this just always be true?

src/cmap/connection.ts
```diff
@@ -526,7 +528,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
         Object.assign(options, { ...pluckBSONSerializeOptions(options) })
       );
 
-      queryOptions.fullResult = true;
+      queryOptions[kFullResult] = true;
       queryOptions.command = true;
       write(this, getMoreOp, queryOptions, (err, response) => {
         if (fullResult) return callback(err, response);
```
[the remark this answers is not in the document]
dariakp: yeah, write actually does switch between passing up true or false, I think for regular command methods (like insertMany), it ends up being false, but for some of our special ones where we have internal parsing set up for the full response, it sets it up with true

src/operations/command.ts
```diff
@@ -169,6 +166,6 @@ export abstract class CommandOperation<T> extends AbstractOperation<T> {
       }
     }
 
-    server.command(this.ns, cmd, { fullResult: !!this.fullResponse, ...options }, callback);
+    server.command(this.ns, cmd, { ...options }, callback);
```
nbbeeken, line 169: Is there a reason to spread this anymore?

dariakp: nope

nbbeeken (approved):
LGTM

emadum (approved):

PR #2945, opened 2021-08-24
fix(NODE-3546): revert findOne not found result type to null

nbbeeken:
findOne will now return null
ModifyResult correctly types value to be null or TSchema


dariakp (approved):

emadum (approved):
LGTM 👍

PR #2947, opened 2021-08-24
fix(NODE-3528): add support for snappy v7

nbbeeken:


lib/core/connection/utils.js
```diff
@@ -16,6 +19,14 @@ function retrieveBSON() {
   BSON.native = false;
 
   const optionalBSON = require_optional('bson-ext');
+  const bsonExtVersion = parsePackageVersion(
+    require_optional('bson-ext/package.json') || { version: '0.0.0' }
+  );
+  if (bsonExtVersion.major >= 4) {
+    emitWarningOnce(
+      'bson-ext version 4 and above does not work with the 3.x version of the mongodb driver'
+    );
```
nbbeeken, line 28: Since I added the version parsing logic I figured we could add a check here to warn folks, we could also throw here, but the driver will crash else where anyway "bson is not a constructor" is likely the first error to be hit.

lib/core/connection/utils.js
```diff
@@ -31,21 +42,43 @@ function noSnappyWarning() {
...
     };
   }
+
+  const snappyPkg = require_optional('snappy/package.json') || { version: '0.0.0' };
+  const version = parsePackageVersion(snappyPkg);
+  snappy[PKG_VERSION] = version;
+  if (version.major >= 7) {
+    const compressOriginal = snappy.compress;
+    const uncompressOriginal = snappy.uncompress;
+    snappy.compress = (data, callback) => {
```
nbbeeken, line 65: Slightly different approach taken, just wrap the functions here, the actual usage sight needs no changes now.

.evergreen/generate_evergreen_tasks.js
```diff
@@ -419,38 +418,48 @@ SINGLETON_TASKS.push({
   ]
 });
 
-SINGLETON_TASKS.push({
-  name: 'run-custom-csfle-tests',
-  tags: ['run-custom-csfle-tests'],
+const oneOffFuncs = [
```
nbbeeken, line 421: Ported over the CI changes, I actually took a stab at bringing in the bson-ext CI here too, but it started having too much overhead for this PR, but this sets us up to have easier one-off targets going forward.

lib/core/connection/utils.js
dariakp: wouldn't it be more helpful to throw here with the appropriate error instead of letting it fail in a weird way down the line? also, why not put the code doing the check in the `if (optionalBSON)` block?

nbbeeken: Changed it, and I tested this, it throws at the moment of `require('mongodb')` so it'll stop scripts right up front.

dariakp (approved):

emadum (approved):
LGTM with one stylistic suggestion.

.evergreen/generate_evergreen_tasks.js
```diff
@@ -464,7 +473,10 @@ BUILD_VARIANTS.push({
 });
 
 const fileData = yaml.safeLoad(fs.readFileSync(`${__dirname}/config.yml.in`, 'utf8'));
-fileData.tasks = (fileData.tasks || []).concat(BASE_TASKS).concat(TASKS).concat(SINGLETON_TASKS);
+fileData.tasks = (fileData.tasks || [])
+  .concat(BASE_TASKS)
+  .concat(TASKS)
+  .concat(SINGLETON_TASKS);
```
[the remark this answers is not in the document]
nbbeeken: done! did it locally along with prettier just to make sure the formatting is right

PR #2948, opened 2021-08-24
docs(NODE-3569): describe the error API based on instanceof check

nbbeeken:
Documentation for errors, and BSON bump

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,33 +126,33 @@ Add code to connect to the server and the database **myProject**:
 > if a callback is provided a Promise will not be returned.
 
 ```js
-const { MongoClient } = require('mongodb')
+const { MongoClient } = require('mongodb');
 // or as an es module:
 // import { MongoClient } from 'mongodb'
 
 // Connection URL
-const url = 'mongodb://localhost:27017'
-const client = new MongoClient(url)
+const url = 'mongodb://localhost:27017';
+const client = new MongoClient(url);
 
 // Database Name
-const dbName = 'myProject'
+const dbName = 'myProject';
 
 async function main() {
   // Use connect method to connect to the server
-  await client.connect()
-  console.log('Connected successfully to server')
-  const db = client.db(dbName)
-  const collection = db.collection('documents')
+  await client.connect();
+  console.log('Connected successfully to server');
+  const db = client.db(dbName);
+  const collection = db.collection('documents');
 
   // the following code examples can be pasted here...
 
-  return 'done.'
+  return 'done.';
 }
 
 main()
   .then(console.log)
   .catch(console.error)
-  .finally(() => client.close())
+  .finally(() => client.close());
 ```
 
 Run your app from the command line with:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -231,6 +231,30 @@ console.log('index name =', indexName)
 
 For more detailed information, see the [indexing strategies page](https://docs.mongodb.com/manual/applications/indexes/).
 
+## Error Handling
+
+If you need to filter certain errors from our driver we have a helpful tree of errors described in [docs/errors.md](https://github.com/mongodb/node-mongodb-native/blob/4.1/docs/errors.md).
+We expect `instanceof` checks to be maintained as our error API under semver as described in the hierarchy errors.md document.
```
dariakp, line 237: The "hierarchy errors.md document" bit reads a little strange to me, could we improve the phrasing there? Also "as our error API" -> "in our error API" ?

README.md
```diff
@@ -231,6 +231,30 @@ console.log('index name =', indexName)
...
+This means we can change the contents of error messages as we see fit to increase the helpfulness of a given message.
```
dariakp, line 238: Just to make it more obvious:
```suggestion
However, we can change the contents of error messages or further subclass errors as we see fit to increase the helpfulness of a given message.
```

README.md
```diff
@@ -231,6 +231,30 @@ console.log('index name =', indexName)
...
+Any new errors we add to the driver will directly extend or extend a child of `MongoError`.
```
dariakp, line 239:
```suggestion
Any new errors we add to the driver will directly extend an existing error class and no existing error will be moved to a different parent class outside of a major release.
```

README.md
```diff
@@ -231,6 +231,30 @@ console.log('index name =', indexName)
...
+This means `instanceof` will always be able to capture the errors that our driver throws (or returns in a callback).
+
+```typescript
+const client = new MongoClient(url);
+await client.connect()
+const collection = client.db().collection('collection')
```
dariakp, line 245:
```suggestion
await client.connect();
const collection = client.db().collection('collection');
```

dariakp (changes requested):

README.md
emadum, line 238: I think this bit could be more clear, maybe leading with the first part. Something along the lines of:

"It is our recommendation to avoid relying on parsing error message strings in your code, preferring `instanceof` checks. We guarantee the class names of our errors won't change according to semver guidelines, but error messages are subject to change at any time, even patch releases, as we see fit to increase the helpfulness of a given message."

emadum (changes requested):

README.md
dariakp: What do you think of my suggestion above regarding subclass phrasing? (since technically we did introduce subclasses outside a major release as the errors work went into a minor version)

README.md
```diff
@@ -225,12 +225,38 @@ performance. The following function creates an index on the **a** field in the
...
 ```
 
 For more detailed information, see the [indexing strategies page](https://docs.mongodb.com/manual/applications/indexes/).
 
+## Error Handling
+
+If you need to filter certain errors from our driver we have a helpful tree of errors described in [docs/errors.md](https://github.com/mongodb/node-mongodb-native/blob/4.1/docs/errors.md).
+
+It is our recommendation to avoid relying on parsing error message strings in your code, preferring instanceof checks.
+We guarantee the class names of our errors won't change according to semver guidelines, but error messages are subject to change at any time, as well as creating new subclass errors, even patch releases, as we see fit to increase the helpfulness of the errors.
```
emadum, line 239:
```suggestion
We guarantee `instanceof` checks will pass according to semver guidelines, but errors may be sub-classed or their messages may change at any time, even patch releases, as we see fit to increase the helpfulness of the errors.
```

nbbeeken: accepted 🚀

dariakp (approved):

emadum (approved):

PR #2949, opened 2021-08-24
fix(NODE-3463): pass explain error through to callback

emadum:


test/functional/explain.test.js
```diff
@@ -762,4 +763,41 @@ describe('Explain', function() {
...
+        mongodb: '>=3.4'
+      }
+    },
+    test: withClient(function(client, done) {
+      const db = client.db('shouldThrowCatchableError');
+      const collection = db.collection('test');
+      collection
+        .find({ a: 1 })
+        .explain('invalidExplain')
+        .catch(err => {
```
dariakp, line 779: can we add a `.then(() => done(new Error('expected explain to fail but it succeeded')))` or something along those lines here and in the test below, so that we fail early instead of a timeout in case this breaks?

also, for consistency, can we add a new line between these two `it` blocks?

lib/core/wireprotocol/query.js
```diff
@@ -37,9 +37,13 @@ function query(server, ns, cmd, cursorState, options, callback) {
 
   // If we have explain, we need to rewrite the find command
   // to wrap it in the explain command
-  const explain = Explain.fromOptions(options);
-  if (explain) {
-    findCmd = decorateWithExplain(findCmd, explain);
+  try {
+    const explain = Explain.fromOptions(options);
```
nbbeeken, line 41: NODE-3377
Totally expanding scope here but we're validating too strictly here, this could be a quick fix that goes in this PR, we just need to assert string or boolean.

test/functional/explain.test.js
emadum: I think the callback test should fail immediately since the error and success are handled in a single callback, right? I've added the suggested early failure into the promise test.

dariakp (approved):

test/functional/explain.test.js
dariakp: Yeah, you are right, I'm not sure what I was seeing originally, but the current version of the tests looks good

nbbeeken (approved):

PR #2951, opened 2021-08-25
fix(NODE-3487): check for nullish aws mechanism property

nbbeeken:
Checks for nullish aws mechanism property instead of falsy, but usage in the AWS mechanism propery checks for falsiness, this allows users to pass an empty string to avoid using the environment variable.

(needs 3.x port)


src/cmap/auth/mongodb_aws.ts
```diff
@@ -123,18 +136,15 @@ export class MongoDBAWS extends AuthProvider {
             path: '/',
             body
           },
-          {
-            accessKeyId: username,
-            secretAccessKey: password,
-            token
```
nbbeeken, line 129: This should've been `sessionToken`, I moved the whole object declaration to above.

src/cmap/auth/mongodb_aws.ts
```diff
@@ -123,18 +136,15 @@ export class MongoDBAWS extends AuthProvider {
...
-        const authorization = options.headers.Authorization;
-        const date = options.headers['X-Amz-Date'];
-        const payload: AWSSaslContinuePayload = { a: authorization, d: date };
-        if (token) {
-          payload.t = token;
+        const payload: AWSSaslContinuePayload = {
+          a: options.headers.Authorization,
+          d: options.headers['X-Amz-Date']
+        };
+        if (sessionToken != null && sessionToken !== '') {
```
nbbeeken, line 146: Should this just stay false-y?

src/deps.ts
```diff
@@ -111,9 +111,47 @@ try {
...
+          secretAccessKey: string;
+        }
+      | undefined
+  ): {
+    headers: {
+      Authorization: string;
+      'X-Amz-Date': string;
+    };
+  };
+}
```
nbbeeken, line 152: I inlined our own definitions of the types to hold us to our usage, also this catches the mistake of passing in `token` instead of `sessionToken`

PR #2956, opened 2021-08-25
test(NODE-3293): clean up change stream tests setImmediate usage

nbbeeken:
Change stream test clean up

### DIFF
--- a/test/functional/change_stream.test.js
+++ b/test/functional/change_stream.test.js
@@ -1806,7 +1806,7 @@ describe('Change Streams', function () {
           if (changeStream[kMode] === 'emitter') {
             // shutting down the client will end the session, if this happens before
             // the stream initialization aggregate operation is processed, it will throw
-            // sa ession ended error, which can't be caught if we end the stream, so
+            // a session ended error, which can't be caught if we end the stream, so
             // we need to wait for the stream to initialize before closing all the things
             await initPromise;
           }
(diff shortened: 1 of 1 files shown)


test/functional/change_stream.test.js
```diff
@@ -176,26 +171,25 @@ const pipeline = [
...
+    try {
+      await db.createCollection('test');
+    } catch {
+      // ns already exists, don't care
+    } finally {
+      await client.close();
+    }
   });
-  afterEach(() => mock.cleanup());
+  afterEach(async () => await mock.cleanup());
```
emadum, line 192: Just wondering, does switching to async/await in the hooks fix something or is this just stylistic?

test/functional/change_stream.test.js
```diff
@@ -1792,109 +1786,86 @@ describe('Change Streams', function () {
...
-        clearInterval(repeatInsert);
-      }
+      let err;
       if (changeStream) {
-        await changeStream.close();
+        try {
+          if (changeStream[kMode] === 'emitter') {
+            // shutting down the client will end the session, if this happens before
+            // the stream initialization aggregate operation is processed, it will throw
+            // sa ession ended error, which can't be caught if we end the stream, so
```
emadum, line 1809:
```suggestion
            // a session ended error, which can't be caught if we end the stream, so
```

emadum (approved):
Typo fix and a question but this LGTM 👍

test/functional/change_stream.test.js
nbbeeken: Stylistic mostly but when debugging I can step over this line without going into the mocha lib to see the result of the cleanup.

PR #2964, opened 2021-08-30
feat(NODE-3424): use hello for monitoring commands

nbbeeken:
makes use of the helloOk property in hello handshakes

the specs are brought in, in a seperate commit so should be able to filter on that.

Had to skip one test for the legacy topology state machine: `Repeated isWritablePrimary response must be processed`


lib/core/sdam/server_description.js
```diff
@@ -180,7 +180,7 @@ function parseServerType(ismaster) {
   if (ismaster.setName) {
     if (ismaster.hidden) {
       return ServerType.RSOther;
-    } else if (ismaster.ismaster) {
+    } else if (ismaster.ismaster || ismaster.isWritablePrimary) {
```
dariakp, line 183: does this fix a bug or implement a feature?

lib/core/topologies/replset_state.js
```diff
@@ -358,7 +359,8 @@ ReplSetState.prototype.update = function(server) {
   // Standalone server, destroy and return
   //
   if (ismaster && ismaster.ismaster && !ismaster.setName) {
-    this.topologyType = this.primary ? TopologyType.ReplicaSetWithPrimary : TopologyType.Unknown;
+    // We should not mark the topology as Unknown because of one standalone
```
dariakp, line 362: again just curious what this fixes

lib/core/uri_parser.js
```diff
@@ -106,6 +106,15 @@ function parseSrvConnectionString(uri, options, callback) {
         }
 
         record = qs.parse(record[0].join(''));
+
+        if (
+          Object.keys(record)
```
dariakp, line 111: below, in L118, it says that only two record keys are allowed, so does this just fail early in case the key is specifically "loadbalanced"? if so, would it not make more sense to put this logic in L119? just before the other error callback? or should this not be here at all since we already do the check in parsedOptions in L610?

lib/url_parser.js
```diff
@@ -101,6 +101,13 @@ module.exports = function(url, options, callback) {
         record = record[0].join('');
         const parsedRecord = qs.parse(record);
         const items = Object.keys(parsedRecord);
+        if (
```
dariakp, line 104: same question here as I have on the other file regarding doing this check here vs L112 vs not at all

test/unit/sdam/spec.test.js
```diff
@@ -164,7 +164,12 @@ function cloneForCompare(event) {
 
 function executeSDAMTest(testData, testDone) {
   parse(testData.uri, (err, parsedUri) => {
-    if (err) return done(err);
+    if (err) {
+      if (err.message.toLowerCase().indexOf('load balancer') !== -1) {
```
dariakp, line 168: should probably add a comment here with the justification

dariakp (changes requested):

lib/core/error.js
```diff
@@ -328,6 +333,10 @@ function isSDAMUnrecoverableError(error) {
     return true;
   }
 
+  if (typeof error.code !== 'undefined') {
```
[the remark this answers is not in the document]
nbbeeken: Good catch, lots of redundancy added here by mistake. Each of the isX functions checks for the code so there is no need for this fucntion to check for it.

lib/core/sdam/server_description.js
nbbeeken: I'll be honest this line was lifted from the other helloOk PR, I'll investigate, since we always patch in `ismaster.ismaster = ismaster.isWritablePrimary` whenever the hello comes in from the server my bet is that this was added to satify tests that bypass that section, but let me confirm and circle back

lib/core/topologies/replset_state.js
nbbeeken: It fixed an SDAM test where if a node that is within a replicaset declares itself as a standalone it should be removed from the topology but the topology should not become Unknown, simply remain ReplicaSetNoPrimary. This test: (test/spec/server-discovery-and-monitoring/rs/member_standalone.json)

lib/core/uri_parser.js
nbbeeken: This is the options that are specified in the DNS TXT record so we do need to keep the check here, I figure that this LB check if statement is kept exactly the same in the 4 places I had to add it, to remain consistent for this exception.

I can take the more refactor-y approach where I can make the `Object.keys(record).some` check work on an array of keys, that's what we've done in 4.x, up until now there's only been two options permitted in DNS TXT record, now there's a third, maybe we should make this section more future proof.

test/unit/sdam/spec.test.js
nbbeeken: Added a comment with a question, it would take some refactoring but I'm realizing now these tests will claim that LB passes on this version of the driver. WE know that its just skipped but maybe we should make the effort to properly skip it? I think I can just pass the mocha context down to here.

lib/core/sdam/server_description.js
nbbeeken: That's exactly what this was for, SDAM tests simulate the Topology directly so the patch is never made, I would move it to the testing files but I feel like such a small safety net should be kept in driver. However I think it should be patched in the ServerDescription constructor instead of an `||` check. I can open a ticket so we make sure 4.x is sync-ed if you think that change is worth it.

test/unit/sdam/spec.test.js
dariakp: Yes, I think if possible, it would be preferable to properly skip the test, however it may be beyond the scope of this PR - let's file a follow up ticket and add the number here?

lib/core/uri_parser.js
dariakp: As a suggestion, if you use `Object.keys(...).some(k => k.toLowerCase() === 'loadbalanced')` instead of the map/indexOf combination, it'll look a bit cleaner; I think I can live with it being where it currently is in all 4 places if it's less of a visual sprawl 😆

nbbeeken: turned out to be easy enough to do, and yea the .some is much cleaner so I've used that in each place.

test/unit/sdam/spec.test.js
nbbeeken: turned out to be easy enough to do so its now skipping correctly

nbbeeken:
Ok now this is ready if CI passes, sorry for all the rapid patches @dariakp

dariakp (approved):

PR #2965, opened 2021-08-30
fix(NODE-3574): reintroduce ObjectID export

nbbeeken:
Reintroduced the capitalized ObjectID export along with a deprecation tag.

### DIFF
--- a/test/types/mongodb.test-d.ts
+++ b/test/types/mongodb.test-d.ts
@@ -6,7 +6,7 @@ import type { FindCursor } from '../../src/cursor/find_cursor';
 import type { Document } from 'bson';
 import { Db } from '../../src';
 import { Topology } from '../../src/sdam/topology';
-import { ObjectId, ObjectID } from '../../src';
+import { ObjectId } from '../../src';
 
 // We wish to keep these APIs but continue to ensure they are marked as deprecated.
 expectDeprecated(Collection.prototype.insert);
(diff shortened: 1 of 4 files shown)


test/types/mongodb.test-d.ts
```diff
@@ -15,6 +16,8 @@ expectDeprecated(Collection.prototype.count);
 expectDeprecated(AggregationCursor.prototype.geoNear);
 expectDeprecated(Topology.prototype.unref);
 expectDeprecated(Db.prototype.unref);
+// expectDeprecated(ObjectID); // not sure why tsd can't assert the deprecation
```
emadum, line 19: I think this is because it's root level. I tried this out and it worked:

```
import * as AllExports from '../../src';
expectDeprecated(AllExports.ObjectID);
expectNotDeprecated(AllExports.ObjectId);
```

emadum (changes requested):
I'm not sure my suggested approach for tsd is the ideal one but it seems to work.

emadum (approved):

test/unit/bson_import.test.js
```diff
@@ -87,3 +87,13 @@ describe('When importing BSON', function () {
...
 });
+
+describe('MongoDB export', () => {
+  const mongodb = require('../../src');
+  it('should include ObjectId', () =>
+    expect(mongodb).to.have.property('ObjectId').that.is.a('function'));
+  it('should include ObjectID', () =>
+    expect(mongodb).to.have.property('ObjectID').that.is.a('function'));
+  it('should have ObjectID and ObjectId equal each other', () =>
+    expect(mongodb.ObjectId).to.deep.equal(mongodb.ObjectID));
```
dariakp, line 98: Should this be just regular equal since we're actually expecting reference equality?
```suggestion
    expect(mongodb.ObjectId).to.equal(mongodb.ObjectID));
```

dariakp: Also, I don't think there should be any package lock changes for this PR

nbbeeken: Lock file changes fixed, and no longer checking deep equality 🚀

dariakp (approved):

PR #2968, opened 2021-08-31
test(NODE-3409): support AWS temp credentials in CSFLE tests

emadum:
Backports #2880

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1239,6 +1239,17 @@ tasks:
           SSL: ssl
           TOPOLOGY: server
       - func: run tls tests
+  - name: test-tls-support-4.4
+    tags:
+      - tls-support
+    commands:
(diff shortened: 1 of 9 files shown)


nbbeeken (approved):
LGTM Thanks for fixing my test as well!

.evergreen/config.yml
```diff
@@ -549,39 +552,6 @@ tasks:
       - func: run tests
         vars:
           UNIFIED: 1
-  - name: test-5.0-server
```
dariakp, line 552: Are all these removed lines with 5.0 tests and tls-support-4.4 intentional?

dariakp: Looks like the automatic merge conflict resolution may have messed with the generation script and resulting file

dariakp (approved):

PR #2970, opened 2021-08-31
fix: versioned api low node compat fix

dariakp:
## Description

Updating versioned API not to use `Object.values` and unified runner not to use the node util for deep equality check.

### DIFF
--- a/test/functional/versioned-api.test.js
+++ b/test/functional/versioned-api.test.js
@@ -3,12 +3,13 @@
 const expect = require('chai').expect;
 const loadSpecTests = require('../spec/index').loadSpecTests;
 const runUnifiedTest = require('./unified-spec-runner/runner').runUnifiedTest;
-const ServerApiVersion = require('../../lib/core').ServerApiVersion;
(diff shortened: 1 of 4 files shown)


nbbeeken (approved):

test/functional/versioned-api.test.js
```diff
@@ -3,12 +3,13 @@
 const expect = require('chai').expect;
 const loadSpecTests = require('../spec/index').loadSpecTests;
 const runUnifiedTest = require('./unified-spec-runner/runner').runUnifiedTest;
-const ServerApiVersion = require('../../lib/core').ServerApiVersion;
+const ValidServerApiVersions = require('../../lib/core').ValidServerApiVersions;
```
emadum, line 6:
```suggestion
const validVersions = require('../../lib/core').ValidServerApiVersions;
```
Nit, but if you rename this here there's no need for the second declaration further down.

emadum (approved):
LGTM with one nit 👍

PR #2977, opened 2021-09-08
fix(NODE-3434): errInfo should be exposed on bulk write

dariakp:
## Description

**What changed?**
`insertMany` now exposes the `errInfo` returned from the server on error the same way `insertOne` does


nbbeeken (approved):
LGTM!

PR #2978, opened 2021-09-10
fix(ConnectionString): missing name for getBoolean

gitawego:
## Description
in follow example:
```js
const newSet = new Set(['false',false].map(getBoolean));

console.log(newSet.size === 1) // false

const newSetB = new Set(['false',false].map(getBoolean.bind(null,'tls')));

console.log(newSetB.size === 1) // true

```

the newSet would contain two values `false, true`, because `getBoolean` is checking index 0 and 1 instead of values.

this bug makes  mongodb connection is always failed.

**What changed?**

set name to getBoolean

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -385,13 +385,13 @@ export function parseOptions(
   }
 
   if (allOptions.has('tls')) {
-    if (new Set(allOptions.get('tls')?.map(getBoolean)).size !== 1) {
+    if (new Set(allOptions.get('tls')?.map(getBoolean.bind(null,'tls'))).size !== 1) {
       throw new MongoParseError('All values of tls must be the same.');
     }
   }
 
   if (allOptions.has('ssl')) {
-    if (new Set(allOptions.get('ssl')?.map(getBoolean)).size !== 1) {
+    if (new Set(allOptions.get('ssl')?.map(getBoolean.bind(null,'ssl'))).size !== 1) {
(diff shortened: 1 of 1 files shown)


dariakp (approved):
Hi, thank you so much for submitting this fix. I actually just ran into this myself in some related work, so I'll go ahead and get your changes in and then add some additional testing around this.

dariakp (changes requested):
Looks like some lint changes are needed here:

> [2021/09/10 21:00:59.051] /data/mci/daf772d9d5576d9bbcedc8af64e8b216/src/src/connection_string.ts
[2021/09/10 21:00:59.051]   388:60  error  Do not strictly check null  no-restricted-syntax
[2021/09/10 21:00:59.051]   388:65  error  Insert `·`                  prettier/prettier
[2021/09/10 21:00:59.051]   394:60  error  Do not strictly check null  no-restricted-syntax
[2021/09/10 21:00:59.051]   394:65  error  Insert `·`                  prettier/prettier
[2021/09/10 21:00:59.051] ✖ 4 problems (4 errors, 0 warnings)
[2021/09/10 21:00:59.051]   2 errors and 0 warnings potentially fixable with the `--fix` option.

dariakp:
I incorporated the fix in a slightly different form in the linked PR #2979 to make the driver validation of these options more spec compliant. Thank you again for calling this out.

PR #2979, opened 2021-09-10
fix(NODE-3591): tlsCertificateKeyFile option does not default cert

dariakp:
## Description

**What changed?**
Clarified wording in API docs around `tlsCertificateKeyFile` option, fixed the behavior to be as written; additionally added spec-compliant validation of multiple ssl/tls options set in the URI.

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -335,6 +335,10 @@ export function parseOptions(
     allOptions.set(key, values);
   }
 
+  if (allOptions.has('tlsCertificateKeyFile') && !allOptions.has('tlsCertificateFile')) {
+    allOptions.set('tlsCertificateFile', allOptions.get('tlsCertificateKeyFile'));
+  }
+
   const unsupportedOptions = setDifference(
     allKeys,
     Array.from(Object.keys(OPTIONS)).map(s => s.toLowerCase())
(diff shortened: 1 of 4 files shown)


src/connection_string.ts
```diff
@@ -335,6 +335,19 @@ export function parseOptions(
...
   }
 
+  if (allOptions.has('tlsCertificateKeyFile') && !allOptions.has('tlsCertificateFile')) {
+    allOptions.set('tlsCertificateFile', allOptions.get('tlsCertificateKeyFile'));
+  }
+
+  if (allOptions.has('tls') || allOptions.has('ssl')) {
+    const tlsAndSslOpts = (allOptions.get('tls') || [])
+      .concat(allOptions.get('ssl') || [])
+      .map(getBoolean.bind(null, 'tls/ssl'));
```
nbbeeken, line 345:
```suggestion
    const tlsAndSslOpts = [...(allOptions.get('tls') ?? []), ...(allOptions.get('ssl') ?? [])].map(
      (value: boolean) => getBoolean('tls/ssl', value)
    );
```

It's totally preference, but this seems to avoid the bind. 
The other syntax is just 🧅 no need to use it, .concat works fine too.

nbbeeken (approved):

PR #2981, opened 2021-09-13
fix(NODE-3559): incorrect GridFS stream type

dariakp:
## Description

NODE-3559 Fix GridFS writeable stream type


nbbeeken (approved):
LGTM

PR #2985, opened 2021-09-20
docs: update driver documentation link in readme

pensierinmusica:
## Description

**What changed?**
The previous link was not working anymore


dariakp:
@pensierinmusica Hi there, thank you for submitting a fix! However, it seems that the reason the link wasn't working was because of a broken redirect in our hosted documentation, which appears to be working correctly now, so no changes should be needed.

nbbeeken:
The links have been fixed, so it works with and without `/current`. Thanks again!

PR #2986, opened 2021-09-20
fix(NODE-3609): correct listDatabases return type

nbbeeken:
I took a look at the other listX() methods, we have strong type information for `listCollections()` that switches based on the nameOnly option, I figured we don't need the strict typing here but I'm open to changing that. `listIndexes()` has no type information, it returns an unannotated cursor so .toArray() will return `Promise<Document[]>`, which at least the array of objects part is correct.

### DIFF
--- a/src/operations/list_databases.ts
+++ b/src/operations/list_databases.ts
@@ -7,7 +7,12 @@ import type { Db } from '../db';
 import type { ClientSession } from '../sessions';
 
 /** @public */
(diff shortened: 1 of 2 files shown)


dariakp (changes requested):
Looks fine to me and I agree with the reasoning for the rest, but can we add a TS test?

dariakp (approved):

PR #2988, opened 2021-09-21
feat(NODE-3625): add the addFields function to AggregationCursor

massivefermion:
**What changed?**
The addFields function was added to AggregationCursor class for adding an addFields stage to the pipeline
no

### DIFF
--- a/src/cursor/aggregation_cursor.ts
+++ b/src/cursor/aggregation_cursor.ts
@@ -209,6 +209,13 @@ export class AggregationCursor<TSchema = Document> extends AbstractCursor<TSchem
     return this;
   }
 
+  /** Add an addFields stage to the aggregation pipeline */
+  addFields($addFields: Document): this {
+    assertUninitialized(this);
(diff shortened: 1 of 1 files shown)


dariakp:
Hi @massivefermion, thank you for submitting this suggestion. I created a jira ticket to track it here: https://jira.mongodb.org/browse/NODE-3625.

dariakp:
@massivefermion Hi, just to give you an update: following a discussion of this issue, the team feels that it would be better to provide a general builder helper method for aggregation stages (something along the lines of `addStage(<stageName>, <props>)`) instead of implementing a special one for every stage. We believe that approach would be more flexible for future aggregation stages and, in general, more maintainable from a test coverage perspective. Currently, we intend to support explicit helpers only for the most frequently used stages. I am sorry that we are unable to accept the PR at this time, but we appreciate your suggestion and will consider the general stage helper feature ([NODE-3639](https://jira.mongodb.org/browse/NODE-3639)) for our next minor release. Thank you for understanding!

PR #2999, opened 2021-10-13
test(NODE-3606): legacy and new versions of the CSFLE library

nbbeeken:
Our "custom-csfle" has been expanded to run the 1.x version
and our normal tests will use the Node-API version.


dariakp (approved):

dariakp:
@nbbeeken The description on the jira ticket says: 

> When we release the next major version of mongodb-client-encryption our 3.x branch which tests on older versions of node will need to install csfle@1 while Node.js 12+ variants can install csfle@2
> 
> The 4.x branch can be left alone to install the latest csfle always.

so is there an update that needs to happen for 3.x? and I guess 4.x doesn't actually install latest, it's still pinned to a major version

PR #3002, opened 2021-10-13
test(NODE-3606): legacy and new versions of the CSFLE library

nbbeeken:
3.x port

- we'll just switch fle versions based on nodejs version


.evergreen/run-kerberos-tests.sh
```diff
@@ -22,8 +22,13 @@ fi
 echo "Running kinit"
 kinit -k -t "$(pwd)/.evergreen/drivers.keytab" -p ${KRB5_PRINCIPAL}
 
-npm install kerberos
+set -o xtrace
+npm install kerberos@">=2.0.0-beta.0"
```
dariakp, line 26: does kerberos not require a particular node version?

PR #3006, opened 2021-10-18
fix(NODE-3116): reschedule unreliable async interval first

durran:
### Description

When there is an unreliable clock we force async intervals to reschedule immediately.

#### What is changing?

The async interval check for negative clock times, like potentially AWS lambda, is moved to the first check so that we can potentially reschedule instead of just returning as was done previously.

No.

### DIFF
--- a/test/unit/utils.test.js
+++ b/test/unit/utils.test.js
@@ -140,8 +140,11 @@ describe('utils', function () {
 
             // needs to happen on the third call because `wake` checks
             // the `currentTime` at the beginning of the function
+            // The value of now() is not actually negative in the case of
+            // the unreliable check so we force to a negative value now
+            // for this test.
             if (clockCalled === 3) {
-              return now() - 100000;
+              return -1;
             }
 
             return now();
--- a/test/unit/utils.test.js
+++ b/test/unit/utils.test.js
@@ -2,7 +2,6 @@
 const {
   eachAsync,
   executeLegacyOperation,
-  now,
   makeInterruptibleAsyncInterval,
   BufferPool
 } = require('../../src/utils');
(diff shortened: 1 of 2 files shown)


test/unit/utils.test.js
```diff
@@ -140,8 +140,11 @@ describe('utils', function () {
 
             // needs to happen on the third call because `wake` checks
             // the `currentTime` at the beginning of the function
+            // The value of now() is not actually negative in the case of
```
dariakp, line 143: Could you explain how this test works to test the "immediately schedule if the clock is unreliable"? I am having trouble following what the marks array is tracking. Also, the original test had an unexpected return time (i.e., a time in the past), why is that no longer a legitimate test case? Which clause in the utils code was it hitting originally? And does it definitely hit the new clause now? How do we differentiate between "executeAndReschedule" and just "reschedule"?

dariakp: My concern is that none of these tests are meaningfully asserting on the actual intent of the code: for example, if you comment out the entire block under the logic that is supposed to debounce multiple calls, the test above this one still passes, which tells me that we can't be confident that the current fix doesn't actually break some other desired behavior. I am worried that the current test structure is based on a flawed approach: we need to make sure that these tests actually isolate the test conditions, which means that if we are using just a sequence tracking the callback calls, we need to either make sure the sequence is unique to the expected code path - and make it clear *why* that sequence is unique to the expect path - or we need to make sure we are also asserting on what we expect to *not* happen.

dariakp:
Note: we should decide if we still want to target 4.1 for this or update to 4.2

dariakp (approved):

nbbeeken (approved):

PR #3008, opened 2021-10-18
chore(NODE-3550): unskip serverless tests

nbbeeken:
#### What is changing?

Numerous serverless tests can be unskipped with some test runner patching.

No

Test all the things!

### DIFF
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -375,7 +375,7 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    */
   startTransaction(options?: TransactionOptions): void {
     if (this[kSnapshotEnabled]) {
-      throw new MongoCompatibilityError('Transactions are not allowed with snapshot sessions');
+      throw new MongoCompatibilityError('Transactions are not supported in snapshot sessions');
     }
 
     assertAlive(this);
--- a/test/functional/spec-runner/index.js
+++ b/test/functional/spec-runner/index.js
@@ -121,7 +121,7 @@ function parseRunOn(runOn) {
 }
 
 function generateTopologyTests(testSuites, testContext, filter) {
-  testSuites.forEach(testSuite => {
+  for (const testSuite of testSuites) {
     // TODO: remove this when SPEC-1255 is completed
     let runOn = testSuite.runOn;
     if (!testSuite.runOn) {
@@ -132,38 +132,36 @@ function generateTopologyTests(testSuites, testContext, filter) {
     }
     const environmentRequirementList = parseRunOn(runOn);
 
-    environmentRequirementList.forEach(requires => {
+    for (const requires of environmentRequirementList) {
       const suiteName = `${testSuite.name} - ${requires.topology.join()}`;
       describe(suiteName, {
         metadata: { requires },
-        test: function () {
+        test() {
           beforeEach(() => prepareDatabaseForSuite(testSuite, testContext));
           afterEach(() => testContext.cleanupAfterSuite());
-          testSuite.tests.forEach(spec => {
+
+          for (const spec of testSuite.tests) {
             const maybeIt = shouldRunSpecTest(this.configuration, requires, spec, filter)
               ? it
               : it.skip;
-            maybeIt(spec.description, function () {
-              let testPromise = Promise.resolve();
+
+            maybeIt(spec.description, async function () {
               if (spec.failPoint) {
-                testPromise = testPromise.then(() => testContext.enableFailPoint(spec.failPoint));
+                await testContext.enableFailPoint(spec.failPoint);
               }
 
-              // run the actual test
-              testPromise = testPromise.then(() =>
-                runTestSuiteTest(this.configuration, spec, testContext)
-              );
+              await runTestSuiteTest(this.configuration, spec, testContext);
 
               if (spec.failPoint) {
-                testPromise = testPromise.then(() => testContext.disableFailPoint(spec.failPoint));
+                await testContext.disableFailPoint(spec.failPoint);
               }
-              return testPromise.then(() => validateOutcome(spec, testContext));
+              return await validateOutcome(spec, testContext);
             });
-          });
+          }
         }
       });
-    });
-  });
+    }
+  }
 }
 
 function shouldRunSpecTest(configuration, requires, spec, filter) {
--- a/test/functional/spec-runner/index.js
+++ b/test/functional/spec-runner/index.js
@@ -99,6 +99,10 @@ function parseTopologies(topologies) {
 }
 
 function parseRunOn(runOn) {
+  if (runOn.length === 0) {
+    throw new Error('An empty run on requirement will skip all tests!');
+  }
+
   return runOn.map(config => {
     const topology = parseTopologies(config.topology);
     const version = [];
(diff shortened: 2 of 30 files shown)


src/sessions.ts
```diff
@@ -374,7 +374,7 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    */
   startTransaction(options?: TransactionOptions): void {
     if (this[kSnapshotEnabled]) {
-      throw new MongoCompatibilityError('Transactions are not allowed with snapshot sessions');
```
nbbeeken, line 377: After fixing `isClientError` this started failing because of [this test](https://github.com/mongodb/specifications/blob/master/source/sessions/tests/unified/snapshot-sessions.yml#L482). This feels low impact to fix here.

test/functional/spec-runner/index.js
```diff
@@ -121,7 +121,7 @@ function parseRunOn(runOn) {
 }
 
 function generateTopologyTests(testSuites, testContext, filter) {
-  testSuites.forEach(testSuite => {
+  for (const testSuite of testSuites) {
```
nbbeeken, line 124: Sorry about the code churn here, reworking this use normal loops and async await made debugging far more straight forward. For example, if configuring the fail point fails you find out when stepping over that line vs having to look for the ultimate catch clause that handled it.

test/tools/unified-spec-runner/match.ts
```diff
@@ -321,9 +330,12 @@ export function expectErrorCheck(
...
 ): boolean {
-  if (Object.keys(expected)[0] === 'isClientError' || Object.keys(expected)[0] === 'isError') {
-    // FIXME: We cannot tell if Error arose from driver and not from server
-    return;
+  if (expected.isClientError === true) {
+    expect(error).to.satisfy(
+      (error: unknown) => error instanceof MongoDriverError || error instanceof MongoNetworkError
+    );
+  } else if (expected.isClientError === false) {
+    expect(error).to.be.instanceOf(MongoServerError);
```
nbbeeken, line 338: This was possible to fix since we merged the last of the better errors project. But now we can differentiate clearly `isClientError`s and server errors. Even though MongoNetworkError aren't descendent of MongoDriverError we consider it to be a client error here since we are the one to capture and construct it, as opposed to MongoServerError being made of message and code directly from MongoDB.

test/tools/unified-spec-runner/runner.ts
```diff
@@ -199,6 +199,21 @@ export async function runUnifiedTest(
...
+        if (
+          wasProvidedCredentials &&
+          actualEvents.length > 4 &&
+          actualEvents[0].constructor.name.toLowerCase() === 'ConnectionReadyEvent'.toLowerCase()
+        ) {
+          // This is a set of CMAP events, remove the out event at index 1 and the in event at index 2
+          expect(actualEvents[1].constructor.name).to.equal('ConnectionCheckedOutEvent');
+          expect(actualEvents[2].constructor.name).to.equal('ConnectionCheckedInEvent');
+          actualEvents.splice(1, 2);
+        }
```
nbbeeken, line 215: Here's the logic to handle the case where we have an extra ping command that runs upon connecting our clients. I put it here (instead of in matchEvents) since determining if credentials were provided requires access to the `clientId`, happy to adjust.

test/unit/sessions.test.js
```diff
@@ -80,7 +80,7 @@ describe('Sessions - unit/core', function () {
           session = new ClientSession(client, sessionPool, { snapshot: true });
           expect(session.snapshotEnabled).to.equal(true);
           expect(() => session.startTransaction()).to.throw(
-            'Transactions are not allowed with snapshot sessions'
+            'Transactions are not supported in snapshot sessions'
```
nbbeeken, line 83: same deal with the change in error string to match spec test that's now running.

test/functional/transactions.test.js
```diff
@@ -122,25 +124,12 @@ const SKIP_TESTS = [
...
-      'abortTransaction does not retry after WriteConcernError Interrupted',
-      'commitTransaction does not retry error without RetryableWriteError label',
-      'commitTransaction is not retried after UnsatisfiableWriteConcern error',
-      'commitTransaction fails after Interrupted'
-    );
-  }
-  suitesToRun.forEach(suiteSpec => {
+  const suitesToRun = [
+    { name: 'spec tests', specPath: path.join('transactions', 'legacy') },
+    { name: 'withTransaction spec tests', specPath: path.join('transactions', 'convenient-api') }
```
dariakp, line 129: I'm confused, were the convenient-api tests just not ran at all before?

test/tools/unified-spec-runner/runner.ts
```diff
@@ -199,6 +199,21 @@ export async function runUnifiedTest(
           actualEvents = clientCommandEvents.get(clientId);
         }
         expect(actualEvents, `No client entity found with id ${clientId}`).to.exist;
+        const wasProvidedCredentials = !!entities.getEntity('client', clientId).options.credentials;
+
+        // TODO(NODE-2471): Currently the node driver executes a ping command when credentials are provided
+        // Here, we need to filter out two extra events from a set of CMAP events.
+        if (
+          wasProvidedCredentials &&
+          actualEvents.length > 4 &&
```
dariakp, line 208: why specifically > 4? even with the expected events being at index 1 & 2, don't we need just `>= 3` to apply this logic?

test/tools/unified-spec-runner/runner.ts
dariakp: the location makes sense to me

test/tools/unified-spec-runner/runner.ts
nbbeeken: I added this condition before also checking for credentials I may be able to drop it. But it seemed like there were tests (failure cases) that expected to see connectionPoolReady and connectionCheckOut, but not followed by a checkIn. In that case even though it is the ping command being the operation that caused the failure, it still proves the failure occurs. (I'll get back to you on the test name).

dariakp: I'm not following, will be interested to see what fails and why

test/functional/transactions.test.js
dariakp: Is it because of this: 

> Note: the killAllSessions command is not supported on Serverless, so the transactions tests may hang if an individual test leaves a transaction open when it finishes (CLOUDP-84298)

I mean it specifically tells us not to test them, so I would follow up with the drivers group to make sure we understand the implications; I don't want to run into a situation where future changes cause things to fail and then we have to spend time investigating why it's happening; at a minimum, if we do keep them, I would make sure to put a comment above explicitly saying that "these are not required to run for serverless, we are just choosing to because they work"

test/tools/unified-spec-runner/match.ts
dariakp: Not sure I follow, `isClientError` is also optional. Are you saying that in situations where only `isError` is specified, and `isClientError` is not specified, we can't assert one of the above specific types?

test/tools/unified-spec-runner/runner.ts
nbbeeken: whoops [added back in a check to make sure the array has at least one item in it](https://spruce.mongodb.com/version/617074eb562343589588413e/tasks)

test/functional/transactions.test.js
nbbeeken: Well grepping the logs for "withTransaction" I'm seeing nothing so somehow they are still skipped 🤔

test/tools/unified-spec-runner/match.ts
nbbeeken: Yes, I think the test doesn't care about the origin of the error if `isClientError` is omitted, so we shouldn't assert anything.

PR #3011, opened 2021-10-19
fix(NODE-3515): do proper opTime merging in bulk results

durran:
### Description

Fixes the merging of results in a bulk write inside a transaction to properly compare and set the `lastOp` value in the result.

#### What is changing?

Per various specifications, `lastOp` and `opTime` are optional fields in results that can be either a `Timestamp`, or an object with a `ts` and `t` property of `Timestamp` and `Long` respectively. In some cases they are both `Longs`. Our comparison logic on these values could get into a hot mess when trying to compare values of different types. This changes the `lastOp` value in the bulk write result to always be an object with the `ts` and `t` fields and never sets it as a `Timestamp` value anymore.

No

This was reported as a bug in HELP-28460 and related to NODE-3515, although the user provided patch did not solve all potential cases.

### DIFF
--- a/lib/bulk/common.js
+++ b/lib/bulk/common.js
@@ -455,7 +455,7 @@ function mergeBatchResults(batch, bulkResult, err, result) {
   }
 
   // The server write command specification states that lastOp is an optional
-  // mongod only field that has a type of timestamp. Across various scare specs
+  // mongod only field that has a type of timestamp. Across various scarce specs
   // where opTime is mentioned, it is an "opaque" object that can have a "ts" and
   // "t" field with Timestamp and Long as their types respectively.
   // The "lastOp" field of the bulk write result is never mentioned in the driver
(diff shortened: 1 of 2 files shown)


test/unit/bulk_write.test.js
```diff
@@ -131,4 +135,48 @@ describe('Bulk Writes', function() {
...
+              hash: 'AAAAAAAAAAAAAAAAAAAAAAAAAAA=',
+              keyId: 0
+            }
+          },
+          operationTime: '7020546605669417498'
+        };
+
+        it('replaces the lastOp with the properly formatted timestamp', function() {
+          mergeBatchResults(batch, bulkResult, null, result);
+          expect(bulkResult.lastOp.t).to.equal(Long.ZERO);
```
nbbeeken, line 177: Can we add few more assertions here to catch us. lastOp will now always be transformed to have the shape of {`t` and `ts`} right?

```suggestion
          expect(bulkResult.lastOp).to.be.an('object');
          expect(bulkResult.lastOp).to.have.property('t', Long.ZERO);
	  expect(bulkResult.lastOp).to.have.property('ts');

```

lib/bulk/common.js
```diff
@@ -445,44 +454,41 @@ function mergeBatchResults(batch, bulkResult, err, result) {
     return;
   }
 
-  // Deal with opTime if available
+  // The server write command specification states that lastOp is an optional
+  // mongod only field that has a type of timestamp. Across various scare specs
```
nbbeeken, line 458: scare specs? 👻 🎃 (typo?)

lib/bulk/common.js
```diff
@@ -445,44 +454,41 @@ function mergeBatchResults(batch, bulkResult, err, result) {
...
-          typeof bulkResult.lastOp.ts === 'number'
-            ? Long.fromNumber(bulkResult.lastOp.ts)
-            : bulkResult.lastOp.ts;
-        lastOpT =
-          typeof bulkResult.lastOp.t === 'number'
-            ? Long.fromNumber(bulkResult.lastOp.t)
-            : bulkResult.lastOp.t;
+    let opTime = result.lastOp || result.opTime;
+
+    if (opTime) {
```
dariakp, line 468: I'm confused - isn't L465 already ensuring that opTime is defined?

test/unit/bulk_write.test.js
```diff
@@ -131,4 +135,48 @@ describe('Bulk Writes', function() {
 
     expect(() => result.insertedIds).to.not.throw();
   });
+
+  describe('#mergeBatchResults', function() {
+    context('when opTime is an object', function() {
+      context('when the lastOp is a Timestamp', function() {
```
dariakp, line 141: should there be additional tests for all the expected lastOp shapes?

test/unit/bulk_write.test.js
dariakp: Can we not also assert on the `ts` property?

lib/bulk/common.js
```diff
@@ -445,44 +454,41 @@ function mergeBatchResults(batch, bulkResult, err, result) {
...
+      // If the opTime is a Timestamp, convert it to a consistent format to be
+      // able to compare easily. Converting to the object from a timestamp is
+      // much more straightforward than the other direction.
+      if (opTime._bsontype === 'Timestamp') {
```
dariakp, line 472: should we have an `else` with an assertion about the shape to make sure that if it's unexpected we fail in a predictable manner?

lib/bulk/common.js
durran: Yes, the check is now removed.

PR #3014, opened 2021-10-24
feat(NODE-3469,NODE-3615,NODE-3507): update min and max wire versions

durran:
### Description

Updates the minimum wire version for the driver to 6 (MongoDB 3.6) and max wire version to 14 (MongoDB 5.1). Also includes NODE-3615 for bumping max wire version and NODE-3507 for updating SDAM spec tests to raise the max wire version.

- NODE-3469
- NODE-3615
- NODE-3507
- DRIVERS-1315

#### What is changing?

The minimum wire version in the wire protocol has been updated to 6, so MongoDB 3.6, as well as the maximum version number to 5.1 and wire version to 14 for the upcoming 5.1 release. (instead of 5.0). Tests have been added to check the relevant constants.

Perhaps the minimum server version that the driver supports is now 3.6 and not 2.6

DRIVERS-1315 to deprecate support for EOL MongoDB versions is the motivation.


test/functional/collations.test.js
```diff
@@ -21,7 +21,7 @@ describe('Collation', function () {
     test: function () {
       const configuration = this.configuration;
       const client = configuration.newClient(`mongodb://${testContext.server.uri()}/test`);
-      const primary = [Object.assign({}, mock.DEFAULT_ISMASTER)];
+      const primary = [Object.assign({}, mock.DEFAULT_ISMASTER_36)];
```
nbbeeken, line 24: Just a QoL request, since this seems to be a search and replace, could we improve the naming while we're here.
If you head to `test/tools/mock.js` you can stop importing `DEFAULT_ISMASTER` altogether. And we can rename this one to be the new `DEFAULT_HELLO` or `DEFAULT_LEGACY_HELLO` or maybe just `HELLO` or `LEGACY_HELLO`, it's kind of implicitly the default.

src/operations/list_collections.ts
```diff
@@ -90,14 +90,17 @@ export class ListCollectionsOperation extends CommandOperation<string[]> {
       return;
     }
 
-    const command = {
+    return super.executeCommand(server, session, this.generateCommand(), callback);
+  }
+
+  /** @internal */
+  generateCommand(): Document {
```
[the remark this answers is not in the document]
nbbeeken: Sounds good, can you put that in the docs string comment, you can drop `@internal` since the whole class is internal, but it can't hurt to say this was written for unit testing. 
(My thinking: I can imagine a future where generateCommand is defined on every operation class and that would just be intentional design as opposed to just help test.)

nbbeeken (changes requested):
Seems pretty much good to go! Thanks!

src/operations/list_collections.ts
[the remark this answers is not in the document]
durran: Updated the comment to note the method was added for testing. I agree - I was thinking the same in the future that all the commands could have it to starting easing the ability to unit test them.

test/functional/collations.test.js
durran: I've updated to use `HELLO` and `LEGACY_HELLO` now.

durran: (There were a few cases where the default was still used)

nbbeeken (approved):

test/functional/unit-sdam/srv_polling.test.js
```diff
@@ -299,7 +299,7 @@ describe('Mongos SRV Polling', function () {
         return records.map(r => `${r.name}:${r.port}`);
       }
 
-      const MONGOS_DEFAULT_ISMASTER = Object.assign({}, mock.DEFAULT_ISMASTER_36, {
+      const MONGOS_LEGACY_HELLO = Object.assign({}, mock.HELLO, {
```
dariakp, line 302: Is it really LEGACY_HELLO if we are using the most up to date HELLO? Perhaps `MONGOS_HELLO` would work here?

test/functional/unit_bypass_validation.test.js
```diff
@@ -29,17 +29,18 @@ describe('bypass document validation', function () {
             ok: 1,
             cursor: {
               firstBatch: [{}],
-              id: 23,
+              id: 0,
               ns: 'test.test'
             }
           });
         } catch (e) {
+          console.log('error', e);
```
dariakp, line 37: do we need to keep the console logs here and in L91?

test/functional/unit_bypass_validation.test.js
```diff
@@ -29,17 +29,18 @@ describe('bypass document validation', function () {
             ok: 1,
             cursor: {
               firstBatch: [{}],
-              id: 23,
+              id: 0,
```
dariakp, line 32: could you remind me why this needed to be updated (and why it was working before)?

test/tools/mock.js
```diff
@@ -70,6 +70,6 @@ function createServer(port, host, options) {
 module.exports = {
   createServer,
   cleanup,
-  DEFAULT_ISMASTER,
-  DEFAULT_ISMASTER_36
+  LEGACY_HELLO: DEFAULT_ISMASTER,
```
dariakp, line 73: Do we actually need the LEGACY_HELLO at all if we no longer support pre-3.6? I saw most tests got updated to use `HELLO` regardless of whether they were previously `DEFAULT_ISMASTER` or `DEFAULT_ISMASTER_36`. There are only 3 files (if I counted right) that still use `LEGACY_HELLO`: find_cursor, the skipped buffering_proxy test, and the helper MongosFixture (which isn't even used in the code base). Is there any harm in eliminating these references entirely and replacing with `HELLO`?

dariakp (changes requested):
Just a few questions here and a suggestion. Also, since we are removing tests but not the tested codepaths, can we file a follow up ticket to clean those up? (i.e., remove anything that checks an out of bound version; in particular, perhaps we can update the ServerCapabilities to always return true for the corresponding functionalities). On that note, there's also a couple of skipped change_stream tests that still have the maxWireVersion set to 4, which should probably be bumped.

test/functional/unit-sdam/srv_polling.test.js
durran: Updated to be `MONGOS_HELLO` now.

test/functional/unit_bypass_validation.test.js
durran: Removed.

test/functional/unit_bypass_validation.test.js
durran: An exhausted cursor always returns id `0`, so the driver knows there is nothing left to get more and can do its cleanup. Part of this cleanup is ending the session that the cursors owns, but if there is no session this does not need to happen. This test previously was executing on a mocked wire version that had no session support, so the test would not be leaking a session. Once the mocked server was updated to a wire version that supported sessions, this cleanup code was not running and the client was getting closed while the cursor still expected more results from the mocked server.

test/tools/mock.js
durran: I've completely removed the use of `LEGACY_HELLO` now in the tests as well as the unused `MongosFixture`.

test/functional/unit_bypass_validation.test.js
dariakp: Missed the one in L91, but I took the liberty of deleting it myself since it's the only outstanding thing here to merge

dariakp: I also filed a follow up [NODE-3469](https://jira.mongodb.org/browse/NODE-3469) for the other clean up

matthewmayer:
Shouldn't this a semver-major breaking change, since the minimum wire version is increased to 6, which drops support for MongoDB3.2

dariakp:
Hi @matthewmayer, thanks for reaching out: this change was intentional in accordance with our [MongoDB Software Lifecycle Schedules](https://www.mongodb.com/support-policy/lifecycles). Given the server versions 3.4 and earlier are outside of our support policy range, and have been since January 2020, we no longer test our driver against them and therefore can no longer guarantee compatibility. In the interest of avoiding undefined behavior or silent failures we opted for a thrown error to make sure users are made fully aware of the updated support.

We understand that this could be an unexpected change for some users; however, given the EOL status of the platforms, we expect the real impact to be minimal. We are working on a strategy to better communicate server version support changes in the future that aligns with our configurable [Versioned API feature](https://docs.mongodb.com/manual/reference/versioned-api/) and, of course, we are always listening to the feedback from our community.

matthewmayer:
I understand it is intentional, but it seems this should have been released as library version 5.0.0 not 4.2.0 following correct semver principles to minimize possible breakage.

PR #3022, opened 2021-11-02
feat(NODE-3083): support aggregate writes on secondaries

durran:
### Description

Allows $merge and $out aggregate operations to execute on secondaries. The logic here is:

1. If the server version is < 5.0, these operations always execute on the primary.
2. If the server version is >= 5.0, these operations operate using the provided read preference or the primary if none is provided.

The spec specifically avoids multi-version clusters, so we can handle this at server selection by implementing our own server selection function for this case that does not need to scan twice.

Includes NODE-3684 to skip $out/$merge tests on serverless

#### What is changing?

Operations can be flagged with `trySecondaryWrite` to use the new server selector function. This function looks at the read preference and common wire version of the topology to potentially change the read preference used to select the server.

Potentially if we want to let users know about this new feature in the node docs.

DRIVERS-823

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -97,6 +97,7 @@ export interface CommandOptions extends BSONSerializeOptions {
   session?: ClientSession;
   documentsReturnedIn?: string;
   noResponse?: boolean;
+  omitReadPreference?: boolean;
 
   // FIXME: NODE-2802
   willRetryWrite?: boolean;
--- a/src/operations/aggregate.ts
+++ b/src/operations/aggregate.ts
@@ -1,5 +1,4 @@
 import { CommandOperation, CommandOperationOptions, CollationOptions } from './command';
-import { ReadPreference } from '../read_preference';
 import { MongoInvalidArgumentError } from '../error';
 import { maxWireVersion, MongoDBNamespace } from '../utils';
 import { Aspect, defineAspects, Hint } from './operation';
@@ -65,7 +64,7 @@ export class AggregateOperation<T = Document> extends CommandOperation<T> {
     }
 
     if (this.hasWriteStage) {
-      this.readPreference = ReadPreference.primary;
+      this.trySecondaryWrite = true;
     }
 
     if (this.explain && this.writeConcern) {
--- a/src/operations/command.ts
+++ b/src/operations/command.ts
@@ -10,6 +10,7 @@ import type { Server } from '../sdam/server';
 import type { BSONSerializeOptions, Document } from '../bson';
 import type { ReadConcernLike } from './../read_concern';
 import { Explain, ExplainOptions } from '../explain';
+import { MIN_SECONDARY_WRITE_WIRE_VERSION } from '../sdam/server_selection';
 
 const SUPPORTS_WRITE_CONCERN_AND_COLLATION = 5;
 
@@ -124,6 +125,10 @@ export abstract class CommandOperation<T> extends AbstractOperation<T> {
 
     if (this.readConcern && commandSupportsReadConcern(cmd) && !inTransaction) {
       Object.assign(cmd, { readConcern: this.readConcern });
+    }
+
+    if (this.trySecondaryWrite && serverWireVersion < MIN_SECONDARY_WRITE_WIRE_VERSION) {
+      options.omitReadPreference = true;
     }
 
     if (options.collation && serverWireVersion < SUPPORTS_WRITE_CONCERN_AND_COLLATION) {
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -17,7 +17,7 @@ import type { Topology } from '../sdam/topology';
 import type { ClientSession } from '../sessions';
 import type { Document } from '../bson';
 import { supportsRetryableWrites } from '../utils';
-import { secondaryWritableServerSelector } from '../sdam/server_selection';
+import { secondaryWritableServerSelector, ServerSelector } from '../sdam/server_selection';
 
 const MMAPv1_RETRY_WRITES_ERROR_CODE = MONGODB_ERROR_CODES.IllegalOperation;
 const MMAPv1_RETRY_WRITES_ERROR_MESSAGE =
@@ -151,6 +151,16 @@ function executeWithServerSelection(
     session.unpin();
   }
 
+  let selector: ReadPreference | ServerSelector;
+
+  // If operation should try to write to secondary use the custom server selector
+  // otherwise provide the read preference.
+  if (operation.trySecondaryWrite) {
+    selector = secondaryWritableServerSelector(topology.commonWireVersion, readPreference);
+  } else {
+    selector = readPreference;
+  }
+
   const serverSelectionOptions = { session };
   function callbackWithRetry(err?: any, result?: any) {
     if (err == null) {
@@ -180,16 +190,6 @@ function executeWithServerSelection(
       );
 
       return;
-    }
-
-    let selector;
-
-    // If operation should try to write to secondary use the custom server selector
-    // otherwise provide the read preference.
-    if (operation.trySecondaryWrite) {
-      selector = secondaryWritableServerSelector(topology.commonWireVersion, readPreference);
-    } else {
-      selector = readPreference;
     }
 
     // select a new server, and attempt to retry the operation
@@ -238,7 +238,7 @@ function executeWithServerSelection(
   }
 
   // select a server, and execute the operation against it
-  topology.selectServer(readPreference, serverSelectionOptions, (err?: any, server?: any) => {
+  topology.selectServer(selector, serverSelectionOptions, (err?: any, server?: any) => {
     if (err) {
       callback(err);
       return;
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -17,6 +17,7 @@ import type { Topology } from '../sdam/topology';
 import type { ClientSession } from '../sessions';
 import type { Document } from '../bson';
 import { supportsRetryableWrites } from '../utils';
+import { secondaryWritableServerSelector } from '../sdam/server_selection';
 
 const MMAPv1_RETRY_WRITES_ERROR_CODE = MONGODB_ERROR_CODES.IllegalOperation;
 const MMAPv1_RETRY_WRITES_ERROR_MESSAGE =
@@ -181,8 +182,18 @@ function executeWithServerSelection(
       return;
     }
 
+    let selector;
+
+    // If operation should try to write to secondary use the custom server selector
+    // otherwise provide the read preference.
+    if (operation.trySecondaryWrite) {
+      selector = secondaryWritableServerSelector(topology.commonWireVersion, readPreference);
+    } else {
+      selector = readPreference;
+    }
+
     // select a new server, and attempt to retry the operation
-    topology.selectServer(readPreference, serverSelectionOptions, (e?: any, server?: any) => {
+    topology.selectServer(selector, serverSelectionOptions, (e?: any, server?: any) => {
       if (
         e ||
         (operation.hasAspect(Aspect.READ_OPERATION) && !supportsRetryableReads(server)) ||
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -299,6 +299,14 @@ export class Server extends TypedEventEmitter<ServerEvents> {
     // Clone the options
     const finalOptions = Object.assign({}, options, { wireProtocolCommand: false });
 
+    // There are cases where we need to flag the read preference not to get sent in
+    // the command, such as pre-5.0 servers attempting to perform an aggregate write
+    // with a non-primary read preference. In this case the effective read preference
+    // (primary) is not the same as the provided and must be removed completely.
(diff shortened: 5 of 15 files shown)


src/cmap/connection.ts
```diff
@@ -97,6 +97,7 @@ export interface CommandOptions extends BSONSerializeOptions {
   session?: ClientSession;
   documentsReturnedIn?: string;
   noResponse?: boolean;
+  omitReadPreference?: boolean;
```
durran, line 100: Added this option not to tightly couple the operation layer to others.

src/operations/aggregate.ts
```diff
@@ -65,7 +64,7 @@ export class AggregateOperation<T = Document> extends CommandOperation<T> {
     }
 
     if (this.hasWriteStage) {
-      this.readPreference = ReadPreference.primary;
+      this.trySecondaryWrite = true;
```
durran, line 67: We now flag the operation to try executing the write on a secondary instead of setting a primary read preference. This is not specific to $out/$merge as I wanted a bit more generic name in case more appear in the future.

src/operations/command.ts
```diff
@@ -126,6 +127,10 @@ export abstract class CommandOperation<T> extends AbstractOperation<T> {
       Object.assign(cmd, { readConcern: this.readConcern });
     }
 
+    if (this.trySecondaryWrite && serverWireVersion < MIN_SECONDARY_WRITE_WIRE_VERSION) {
```
durran, line 130: The spec states that if we have a write operation that can execute on a secondary, but the server doesn't satisfy the requirements, there must be no read preference provided. This is backed by the crud spec unified tests.

src/operations/execute_operation.ts
```diff
@@ -150,6 +151,16 @@ function executeWithServerSelection(
     session.unpin();
   }
 
+  let selector: ReadPreference | ServerSelector;
+
+  // If operation should try to write to secondary use the custom server selector
+  // otherwise provide the read preference.
+  if (operation.trySecondaryWrite) {
```
durran, line 158: In the case of a potential write to a secondary, we use our new server selector instead of passing the read preference directly to `selectServer`.

src/operations/execute_operation.ts
```diff
@@ -182,7 +193,7 @@ function executeWithServerSelection(
     }
 
     // select a new server, and attempt to retry the operation
-    topology.selectServer(readPreference, serverSelectionOptions, (e?: any, server?: any) => {
+    topology.selectServer(selector, serverSelectionOptions, (e?: any, server?: any) => {
```
durran, line 196: `selectServer` takes either a `ReadPreference` or a `ServerSelector` function as its first argument.

src/sdam/server.ts
```diff
@@ -299,6 +299,14 @@ export class Server extends TypedEventEmitter<ServerEvents> {
     // Clone the options
     const finalOptions = Object.assign({}, options, { wireProtocolCommand: false });
 
+    // There are cases where we need to flag the read preference not to get sent in
+    // the command, such as pre-5.0 servers attempting to perform an aggregate write
+    // with a non-primary read preference. In this case the effective read preference
+    // (primary) is not the same as the provided and must be removed completely.
+    if (finalOptions.omitReadPreference) {
```
durran, line 306: Comment explains the reasoning here from the spec.

src/sdam/server_selection.ts
```diff
@@ -8,6 +8,9 @@ import type { ServerDescription, TagSet } from './server_description';
 const IDLE_WRITE_PERIOD = 10000;
 const SMALLEST_MAX_STALENESS_SECONDS = 90;
 
+//  Minimum version to try writes on secondaries.
+export const MIN_SECONDARY_WRITE_WIRE_VERSION = 13;
```
durran, line 12: Wire version `13`, or server `5.0` is the minimum this functionality can happen on.

src/sdam/server_selection.ts
```diff
@@ -28,6 +31,24 @@ export function writableServerSelector(): ServerSelector {
     );
 }
 
+/**
+ * Returns a server selector that uses a read preference to select a
+ * server potentially for a write on a secondary.
+ */
+export function secondaryWritableServerSelector(
```
durran, line 38: Our new custom server selector that determines the effective read preference for server selection and delegates to the existing `readPreferenceServerSelector`. Implementing a new one here and delegating allows us not to mess with the existing logic in the other selector. While it is backed by spec tests, the lack of unit tests was a red flag for me to approach it by modifying the other.

test/functional/crud_spec.test.js
```diff
@@ -424,15 +424,11 @@ describe('CRUD spec v1', function () {
...
-// TODO: Unskip when implementing NODE-3083.
-const SKIP = ['aggregate-write-readPreference', 'db-aggregate-write-readPreference'];
-
 describe('CRUD unified', function () {
   for (const crudSpecTest of loadSpecTests('crud/unified')) {
     expect(crudSpecTest).to.exist;
     const testDescription = String(crudSpecTest.description);
-    const spec = SKIP.includes(testDescription) ? context.skip : context;
-    spec(testDescription, function () {
+    context(testDescription, function () {
```
durran, line 431: The new spec tests are no longer skipped now.

test/spec/crud/unified/aggregate-write-readPreference.json
```diff
@@ -90,7 +90,8 @@
       "description": "Aggregate with $out includes read preference for 5.0+ server",
       "runOnRequirements": [
         {
-          "minServerVersion": "5.0"
+          "minServerVersion": "5.0",
+          "serverless": "forbid"
```
durran, line 94: Synced for NODE-3684

test/unit/operations/aggregate.test.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+const { expect } = require('chai');
+const { AggregateOperation } = require('../../../src/operations/aggregate');
+
+describe('AggregateOperation', function () {
+  const db = 'test';
+
+  describe('#constructor', function () {
```
durran, line 9: New unit tests to check that the operation gets flagged properly when the aggregation is a write.

test/unit/sdam/server_selection.test.js
```diff
@@ -0,0 +1,381 @@
...
+const { ReadPreference } = require('../../../src/read_preference');
+const {
+  secondaryWritableServerSelector,
+  MIN_SECONDARY_WRITE_WIRE_VERSION
+} = require('../../../src/sdam/server_selection');
+const { ServerDescription } = require('../../../src/sdam/server_description');
+const { TopologyDescription } = require('../../../src/sdam/topology_description');
+const { TopologyType } = require('../../../src/sdam/common');
+
+describe('server selection', function () {
```
durran, line 14: New unit tests for our new server selector.

src/operations/command.ts
dariakp: Since this is the only place where "omitReadPreference" is set, is there a benefit to having an `omitReadPreference` defined as a distinct command option as opposed to just performing the option removal right here? (I'm not saying we definitely shouldn't, I just want to understand the reasoning better). And if we do want to keep the option, would `ignoreReadPreference` to be more clear about what we want to make sure happens? (Or is it just that we don't want to send a mismatched preference to, if I understand correctly, an already guaranteed primary server selected?)

dariakp (commented):
Just a question to make sure I understand the implementation (I'll take a second pass at the newly added tests once I better understand the context, but I'll move it to team review)

src/operations/command.ts
durran: I used “omit” instead of “ignore” because in many cases there is an explicit read preference set that needs to be removed from the command. Using “ignore” to me would imply no action taken but here we remove it. There is some vagueness going on in the specs, such as https://jira.mongodb.org/plugins/servlet/mobile#issue/DRIVERS-1918 which I figured we would address later.

nbbeeken (approved):
LGTM

test/unit/sdam/server_selection.test.js
```diff
@@ -0,0 +1,381 @@
...
+      context('when a common wire version is not provided', function () {
+        const topologyDescription = new TopologyDescription(
+          TopologyType.ReplicaSetWithPrimary,
+          serverDescriptions,
+          'test',
+          MIN_SECONDARY_WRITE_WIRE_VERSION,
+          new ObjectId(),
+          MIN_SECONDARY_WRITE_WIRE_VERSION
+        );
+        const selector = secondaryWritableServerSelector();
```
dariakp, line 114: this does not differentiate between the wire version not being provided and the read preference not being provided, so I think doing `(undefined, ReadPreference.secondary)` would work better

test/unit/sdam/server_selection.test.js
```diff
@@ -0,0 +1,381 @@
...
+          });
+        });
+      });
+
+      context('when the common server version is < 5.0', function () {
+        const topologyDescription = new TopologyDescription(
+          TopologyType.ReplicaSetWithPrimary,
+          serverDescriptions,
+          'test',
+          MIN_SECONDARY_WRITE_WIRE_VERSION - 1,
```
dariakp, line 78: Would it be better to keep the maxSetVersion at 5.0 here so that we're narrowing it down to specifically common wire version being the determining factor? (To account for the case where we have a mix and match of the available servers)

dariakp (changes requested):
Just a few small requests regarding the tests; I think we're all happy with the actual code approach, and thank you so much for adding the new unit tests!

test/unit/sdam/server_selection.test.js
durran: Updated per suggestion.

test/unit/sdam/server_selection.test.js
durran: I've updated to this now. Nice suggestion - this actually caught a bug.

test/unit/operations/aggregate.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+    context('when $merge is the last stage', function () {
+      const operation = new AggregateOperation(db, [{ $merge: { into: 'test' } }], { dbName: db });
+
+      it('sets trySecondaryWrite to true', function () {
+        expect(operation.trySecondaryWrite).to.be.true;
+      });
+    });
+
+    context('when no writable stages', function () {
```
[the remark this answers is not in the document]
durran: Added another context block for the case where the pipeline is not empty, but has no write stages.

test/unit/operations/aggregate.test.js
```diff
@@ -0,0 +1,42 @@
...
+    context('when out is in the options', function () {
+      const operation = new AggregateOperation(db, [], { out: 'test', dbName: db });
+
+      it('sets trySecondaryWrite to true', function () {
+        expect(operation.trySecondaryWrite).to.be.true;
+      });
+    });
+
+    context('when $out is the last stage', function () {
+      const operation = new AggregateOperation(db, [{ $out: 'test' }], { dbName: db });
```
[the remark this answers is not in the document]
durran: Added 2 extra context blocks for the cases where $out and $merge are not the last stage in the pipeline. (I prefer one assertion per `it` block for better documentation for the tests - hence the extra contexts added).

test/unit/sdam/server_selection.test.js
```diff
@@ -0,0 +1,381 @@
...
+const { ReadPreference } = require('../../../src/read_preference');
+const {
+  secondaryWritableServerSelector,
+  MIN_SECONDARY_WRITE_WIRE_VERSION
+} = require('../../../src/sdam/server_selection');
+const { ServerDescription } = require('../../../src/sdam/server_description');
+const { TopologyDescription } = require('../../../src/sdam/topology_description');
+const { TopologyType } = require('../../../src/sdam/common');
+
+describe.only('server selection', function () {
```
dariakp, line 14:
```suggestion
describe('server selection', function () {
```

test/unit/sdam/server_selection.test.js
```diff
@@ -0,0 +1,381 @@
...
+          });
+        });
+      });
+
+      context('when the common server version is < 5.0', function () {
+        const topologyDescription = new TopologyDescription(
+          TopologyType.Sharded,
+          serverDescriptions,
+          'test',
+          MIN_SECONDARY_WRITE_WIRE_VERSION - 1,
```
dariakp, line 164: can we make the same change here and in the 2 other similar cases as we did for the replica set?

dariakp (changes requested):

test/unit/sdam/server_selection.test.js
```diff
@@ -0,0 +1,381 @@
...
+          });
+        });
+      });
+
+      context('when the common server version is < 5.0', function () {
+        const topologyDescription = new TopologyDescription(
+          TopologyType.Single,
+          serverDescriptions,
+          'test',
+          MIN_SECONDARY_WRITE_WIRE_VERSION - 1,
```
dariakp, line 336:
```suggestion
          MIN_SECONDARY_WRITE_WIRE_VERSION,
```

dariakp (changes requested):

dariakp (approved):

PR #3030, opened 2021-11-08
fix(NODE-3648): run get more ops through server selection

durran:
### Description

Runs `getmore` command operations through server selection, selecting the same server.

#### What is changing?

A new server selector has been added, `sameServerSelector`, which takes a `ServerDescription` and returns a `ServerDescription[]`. This will always select the same server as the description provided, unless the server has transitioned into an unknown state, where it will return an empty array. In this case, server selection will force monitor checks to potentially update the server and select the same one again.

A new `GetMoreOperation` has been added in order for the cursors on `getMore` to execute this operation with server selection. The pinned server's description on the operation will be used in conjunction with the new server selector.

No

NODE-3648

### DIFF
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -17,7 +17,11 @@ import type { Topology } from '../sdam/topology';
 import type { ClientSession } from '../sessions';
 import type { Document } from '../bson';
 import { supportsRetryableWrites } from '../utils';
-import { secondaryWritableServerSelector, ServerSelector } from '../sdam/server_selection';
+import {
+  sameServerSelector,
+  secondaryWritableServerSelector,
+  ServerSelector
+} from '../sdam/server_selection';
 
 const MMAPv1_RETRY_WRITES_ERROR_CODE = MONGODB_ERROR_CODES.IllegalOperation;
 const MMAPv1_RETRY_WRITES_ERROR_MESSAGE =
@@ -153,9 +157,14 @@ function executeWithServerSelection(
 
   let selector: ReadPreference | ServerSelector;
 
-  // If operation should try to write to secondary use the custom server selector
-  // otherwise provide the read preference.
-  if (operation.trySecondaryWrite) {
+  if (operation.hasAspect(Aspect.CURSOR_ITERATING)) {
+    // Get more operations must always select the same server, but run through
+    // server selection to potentially force monitor checks if the server is
+    // in an unknown state.
+    selector = sameServerSelector(operation.server?.description);
+  } else if (operation.trySecondaryWrite) {
+    // If operation should try to write to secondary use the custom server selector
+    // otherwise provide the read preference.
     selector = secondaryWritableServerSelector(topology.commonWireVersion, readPreference);
   } else {
     selector = readPreference;
--- a/src/operations/get_more.ts
+++ b/src/operations/get_more.ts
@@ -1,0 +1,48 @@
+import type { Document, Long } from '../bson';
+import type { Callback, MongoDBNamespace } from '../utils';
+import type { Server } from '../sdam/server';
+import { Aspect, AbstractOperation, OperationOptions, defineAspects } from './operation';
+import type { ClientSession } from '../sessions';
+
+/**
+ * @public
+ * @typeParam TSchema - Unused schema definition, deprecated usage, only specify `FindOptions` with no generic
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export interface GetMoreOptions<TSchema extends Document = Document> extends OperationOptions {
+  /** Set the batchSize for the getMoreCommand when iterating over the query results. */
+  batchSize?: number;
+  /** You can put a $comment field on a query to make looking in the profiler logs simpler. */
+  comment?: string | Document;
+  /** Number of milliseconds to wait before aborting the query. */
+  maxTimeMS?: number;
+}
+
+/** @internal */
+export class GetMoreOperation extends AbstractOperation {
+  cursorId: Long;
+  options: GetMoreOptions;
+
+  constructor(ns: MongoDBNamespace, cursorId: Long, options: GetMoreOptions = {}) {
+    super(options);
+    this.options = options;
+    this.ns = ns;
+    this.cursorId = cursorId;
+  }
+
+  execute(server: Server, session: ClientSession, callback: Callback<Document>): void {
+    this.server = server;
+
+    server.getMore(
+      this.ns,
+      this.cursorId,
+      {
+        ...this.options,
+        session: session
+      },
+      callback
+    );
+  }
+}
+
+defineAspects(GetMoreOperation, [Aspect.READ_OPERATION, Aspect.RETRYABLE, Aspect.CURSOR_ITERATING]);
--- a/src/operations/get_more.ts
+++ b/src/operations/get_more.ts
@@ -22,17 +22,21 @@ export interface GetMoreOptions<TSchema extends Document = Document> extends Ope
 export class GetMoreOperation extends AbstractOperation {
   cursorId: Long;
   options: GetMoreOptions;
+  server: Server;
 
-  constructor(ns: MongoDBNamespace, cursorId: Long, options: GetMoreOptions = {}) {
+  constructor(ns: MongoDBNamespace, cursorId: Long, server: Server, options: GetMoreOptions = {}) {
     super(options);
     this.options = options;
     this.ns = ns;
     this.cursorId = cursorId;
+    this.server = server;
   }
 
+  /**
+   * Although there is a server already associated with the get more operation, the signature
+   * for execute passes a server so we will just use that one.
+   */
   execute(server: Server, session: ClientSession, callback: Callback<Document>): void {
-    this.server = server;
-
     server.getMore(
       this.ns,
       this.cursorId,
--- a/src/sdam/server_selection.ts
+++ b/src/sdam/server_selection.ts
@@ -35,11 +35,12 @@ export function writableServerSelector(): ServerSelector {
  * The purpose of this selector is to select the same server, only
  * if it is in a state that it can have commands sent to it.
  */
-export function sameServerSelector(description: ServerDescription): ServerSelector {
+export function sameServerSelector(description?: ServerDescription): ServerSelector {
   return (
     topologyDescription: TopologyDescription,
     servers: ServerDescription[]
   ): ServerDescription[] => {
+    if (!description) return [];
     // Filter the servers to match the provided description only if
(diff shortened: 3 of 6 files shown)


src/operations/execute_operation.ts
```diff
@@ -153,9 +157,14 @@ function executeWithServerSelection(
 
   let selector: ReadPreference | ServerSelector;
 
-  // If operation should try to write to secondary use the custom server selector
-  // otherwise provide the read preference.
-  if (operation.trySecondaryWrite) {
+  if (operation.hasAspect(Aspect.CURSOR_ITERATING)) {
```
durran, line 160: I called this aspect "cursor iterating" in case there are future commands that behave similar to `getmore`.

src/operations/get_more.ts
```diff
@@ -0,0 +1,44 @@
...
+import type { Server } from '../sdam/server';
+import { Aspect, AbstractOperation, OperationOptions, defineAspects } from './operation';
+import type { ClientSession } from '../sessions';
+
+/**
+ * @public
+ * @typeParam TSchema - Unused schema definition, deprecated usage, only specify `FindOptions` with no generic
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export interface GetMoreOptions<TSchema extends Document = Document> extends OperationOptions {
```
durran, line 12: Options passed for the command are `batchSize`, `comment`, `maxTimeMS`, and the inherited options from `OperationOptions`.

src/sdam/server_selection.ts
```diff
@@ -31,6 +31,24 @@ export function writableServerSelector(): ServerSelector {
     );
 }
 
+/**
+ * The purpose of this selector is to select the same server, only
+ * if it is in a state that it can have commands sent to it.
+ */
+export function sameServerSelector(description?: ServerDescription): ServerSelector {
```
durran, line 38: Selects the same server unless it has transitioned to an unknown state.

test/unit/operations/get_more.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+const sinon = require('sinon');
+const { expect } = require('chai');
+const { Long } = require('../../../src/bson');
+const { GetMoreOperation } = require('../../../src/operations/get_more');
+const { Server } = require('../../../src/sdam/server');
+const { ClientSession } = require('../../../src/sessions');
+const { ReadPreference } = require('../../../src/read_preference');
+
+describe('GetMoreOperation', function () {
```
durran, line 11: Unit tests for the new operation.

test/unit/sdam/server_selection.test.js
```diff
@@ -31,6 +32,67 @@ describe('server selection', function () {
     isWritablePrimary: true,
     ok: 1
   });
+  const unknown = new ServerDescription('127.0.0.1:27022', {
+    ok: 0
+  });
+
+  describe('#sameServerSelector', function () {
```
durran, line 39: Unit tests for the new server selector.

test/unit/sdam/server_selection.test.js
```diff
@@ -53,19 +115,19 @@ describe('server selection', function () {
             MIN_SECONDARY_WRITE_WIRE_VERSION,
             ReadPreference.secondary
           );
-          const server = selector(topologyDescription, Array.from(serverDescriptions.values()));
+          const servers = selector(topologyDescription, Array.from(serverDescriptions.values()));
```
durran, line 118: Server selectors return an array of server descriptions so I fixed the variable name in these tests to be plural.

src/sdam/server_selection.ts
```diff
@@ -31,6 +31,24 @@ export function writableServerSelector(): ServerSelector {
...
+  return (
+    topologyDescription: TopologyDescription,
+    servers: ServerDescription[]
+  ): ServerDescription[] => {
+    if (!description) return [];
+    // Filter the servers to match the provided description only if
+    // the type is not unknown.
+    return servers.filter((s: ServerDescription) => {
```
nbbeeken, line 46:
```suggestion
    return servers.filter(sd => {
      return sd.address === description.address && sd.type !== ServerType.Unknown;
```

Mostly stylistic but our well typed stuff can make use of less annotations.

src/operations/get_more.ts
nbbeeken:
```suggestion
export interface GetMoreOptions extends OperationOptions {
```
and you can drop the eslint exception comment.

src/operations/get_more.ts
```diff
@@ -0,0 +1,43 @@
...
+    super(options);
+    this.options = options;
+    this.ns = ns;
+    this.cursorId = cursorId;
+    this.server = server;
+  }
+
+  /**
+   * Although there is a server already associated with the get more operation, the signature
+   * for execute passes a server so we will just use that one.
```
nbbeeken, line 36: Should we double check here that server is the same? I think it should always even be exactly the same reference, so we just make use of a plain equals:
```ts
if (server !== this.server) {
    return callback(new MongoRuntimeError('Getmore must run on the same server operation began on'))
}
// ...
```

nbbeeken (changes requested):
Some initial thoughts. This is very close to ready as is, but there's another component to fixing the original bug that we could consider to include here. Only in this version of the driver have we now dropped support for versions of MongoDB that don't support sessions. Part of the issue was that our hasSessionSupport logic was omitting the session because the monitors downgraded it. Now that we run getMores through serverSelection we shouldn't encounter this issue. 

But should additionally remove the checks altogether? and in fact maybe we should consider throwing Connection#getMore if we don't have a session at that point. (possibly too far) but IIUC of all operations getMores 100% need to have a session defined. (with our currently supported server vers)

src/operations/execute_operation.ts
durran: I called this `CURSOR_ITERATING` since I did not want to confuse it with transaction server pinning, which is not operation specific and thought it might cause confusion.

src/operations/get_more.ts
durran: Updated with this check and added new test.

src/operations/execute_operation.ts
nbbeeken: Yea the potential confusion is more than its worth. This is fine.

nbbeeken (approved):
LGTM

test/unit/sdam/server_selection.test.js
```diff
@@ -31,6 +32,67 @@ describe('server selection', function () {
...
+    ok: 0
+  });
+
+  describe('#sameServerSelector', function () {
+    const serverDescriptions = new Map();
+    serverDescriptions.set('127.0.0.1:27017', primary);
+    serverDescriptions.set('127.0.0.1:27018', unknown);
+
+    context('when the server is unknown', function () {
+      const topologyDescription = new TopologyDescription(
```
dariakp, line 45: could we move the `topologyDescription` and the list of servers (`serverList = Array.from(serverDescriptions.values())`) into a `beforeEach`? Since the same descriptions are used for all the test cases and only the input to sameServerSelector varies, it would make sense to isolate the rest of the shared setup, too

test/unit/sdam/server_selection.test.js
```diff
@@ -31,6 +32,67 @@ describe('server selection', function () {
...
+      );
+      const selector = sameServerSelector(unknown);
+      const servers = selector(topologyDescription, Array.from(serverDescriptions.values()));
+
+      it('returns an empty array', function () {
+        expect(servers).to.be.empty;
+      });
+    });
+
+    context('when the server is not unknown', function () {
```
dariakp, line 61: Could we also add a test case for when there is no server that matches the address? e.g., `selector = sameServerSelector(secondary)` ?

test/unit/operations/get_more.test.js
```diff
@@ -0,0 +1,118 @@
...
+const { Server } = require('../../../src/sdam/server');
+const { ClientSession } = require('../../../src/sessions');
+const { ReadPreference } = require('../../../src/read_preference');
+const { Aspect } = require('../../../src/operations/operation');
+const { MongoRuntimeError } = require('../../../src/error');
+
+describe('GetMoreOperation', function () {
+  const ns = 'db.coll';
+  const cursorId = Long.fromNumber(1);
+  const options = {
```
dariakp, line 16: can we freeze these so that we can make sure the options aren't mutated? (otherwise that deep equal check below doesn't guarantee as much)

dariakp (changes requested):
Just a few small suggestions around the testing

test/unit/sdam/server_selection.test.js
```diff
@@ -31,6 +32,67 @@ describe('server selection', function () {
...
     ok: 1
   });
+  const unknown = new ServerDescription('127.0.0.1:27022', {
+    ok: 0
+  });
+
+  describe('#sameServerSelector', function () {
+    const serverDescriptions = new Map();
+    serverDescriptions.set('127.0.0.1:27017', primary);
+    serverDescriptions.set('127.0.0.1:27018', unknown);
```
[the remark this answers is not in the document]
durran: I've changed all the tests in this test module to follow this pattern now.

test/unit/sdam/server_selection.test.js
durran: Moved into `beforeEach` and `before` blocks for setup now.

test/unit/sdam/server_selection.test.js
durran: Extra test case added.

test/unit/operations/get_more.test.js
durran: Now all params are frozen.

test/unit/sdam/server_selection.test.js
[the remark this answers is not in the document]
dariakp: Thank you!

test/unit/operations/get_more.test.js
```diff
@@ -0,0 +1,118 @@
...
+const { Long } = require('../../../src/bson');
+const { GetMoreOperation } = require('../../../src/operations/get_more');
+const { Server } = require('../../../src/sdam/server');
+const { ClientSession } = require('../../../src/sessions');
+const { ReadPreference } = require('../../../src/read_preference');
+const { Aspect } = require('../../../src/operations/operation');
+const { MongoRuntimeError } = require('../../../src/error');
+
+describe('GetMoreOperation', function () {
+  const ns = Object.freeze('db.coll');
```
dariakp, line 14: This one is just a string, so `const` has the immutability covered
```suggestion
  const ns = 'db.coll';
```

dariakp (changes requested):
One small nit

PR #3031, opened 2021-11-08
feat(NODE-3467): implement srvMaxHosts, srvServiceName options

nbbeeken:
#### What is changing?

A new option `srvMaxHosts` is now implemented that enables a user to limit the number of mongoses the driver will connect to. The driver will select from the list of hosts returned by a DNS lookup at random using the Fisher-Yates algorithm.

The `srvServiceName` allows the srv address prefix to be changed from the default of `mongodb`

There is new TSDoc, I don't think further documentation will be needed.

The number of mongos hosts can be large and each gets its own connection pool, it allows for tuning of how many query routers a driver connects to, primarily in order to manage networking resources.


src/utils.ts
```diff
@@ -1319,6 +1321,23 @@ export class HostAddress {
     Object.freeze(this);
   }
 
+  equals(other: HostAddress): boolean {
+    return (
```
dariakp, line 1325: just double checking that we definitely want to be strict even in cases of null/undefined with the triple equals here

dariakp (changes requested):
Some feedback just on the helpers

src/utils.ts
```diff
@@ -1405,3 +1428,32 @@ export function parsePackageVersion({ version }: { version: string }): {
...
+ *
+ * Reference: https://bost.ocks.org/mike/shuffle/
+ * @param sequence - items to be shuffled
+ * @param limit - the number of ite
+ */
+export function shuffle<T>(sequence: Iterable<T>, limit = 0): Array<T> {
+  const items = Array.from(sequence); // shallow copy in order to never shuffle the input
+
+  if (limit > items.length) {
+    throw new MongoInvalidArgumentError('Limit must be less than the number of items');
```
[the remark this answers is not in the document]
nbbeeken: There isn't a good specific one, I think staying general is fine here since `ShuffleError` would be too generic, and `SrvShuffleError` would be too specific.

test/unit/utils.test.js
```diff
@@ -297,4 +299,31 @@ describe('utils', function () {
...
+      const output = shuffle(input);
+      expect(Array.isArray(output)).to.be.true;
+    });
+    it('should give a random subset if limit is less than the input length', () => {
+      const input = ['a', 'b', 'c', 'd', 'e'];
+      const output = shuffle(input, 3);
+      expect(output).to.have.lengthOf(3);
+    });
+
+    it('should give a random shuffling of the input', () => {
```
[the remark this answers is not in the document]
nbbeeken: done.

test/unit/utils.test.js
```diff
@@ -297,4 +299,31 @@ describe('utils', function () {
...
   });
+
+  describe('shuffler function', () => {
+    it('should support iterables', function () {
+      // Kind of an implicit test, we should not throw/crash here.
+      const input = new Set(['a', 'b', 'c']);
+      const output = shuffle(input);
+      expect(Array.isArray(output)).to.be.true;
+    });
+    it('should give a random subset if limit is less than the input length', () => {
```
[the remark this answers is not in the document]
nbbeeken: Nice idea

test/unit/utils.test.js
```diff
@@ -297,4 +299,31 @@ describe('utils', function () {
...
+      const input = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i'];
+      const output = shuffle(input);
+      // Of course it is possible a shuffle returns exactly the same as the input
+      // but it is so improbable it is worth the flakiness in my opinion
+      expect(output).to.not.deep.equal(input);
+      expect(output).to.have.lengthOf(input.length);
+    });
+
+    it('should throw if limit is larger than input size', () => {
+      expect(() => shuffle([], 100)).to.throw(MongoInvalidArgumentError);
```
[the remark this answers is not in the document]
nbbeeken: Done. I think I got it all!

src/sdam/srv_polling.ts
```diff
@@ -114,7 +137,11 @@ export class SrvPoller extends TypedEventEmitter<SrvPollerEvents> {
   success(srvRecords: dns.SrvRecord[]): void {
     this.haMode = false;
     this.schedule();
-    this.emit(SrvPoller.SRV_RECORD_DISCOVERY, new SrvPollingEvent(srvRecords));
+    const event = new SrvPollingEvent(srvRecords);
+    if (!this.lastSrvPollingEvent?.equals(event)) {
+      this.emit(SrvPoller.SRV_RECORD_DISCOVERY, event);
+    }
```
nbbeeken, line 143: @dariakp 
This is my test of a current discussion around if hosts should be "sticky" or each srv polling event should cause a reshuffling. 
I POC-ed this and didn't mean to commit it whoops! I'll revert this but keep the tests fixes I made on your current comments

src/utils.ts
```diff
@@ -1405,3 +1419,32 @@ export function parsePackageVersion({ version }: { version: string }): {
   const [major, minor, patch] = version.split('.').map((n: string) => Number.parseInt(n, 10));
   return { major, minor, patch };
 }
+
+/**
+ * Fisher–Yates Shuffle
+ *
+ * Reference: https://bost.ocks.org/mike/shuffle/
+ * @param sequence - items to be shuffled
+ * @param limit - the number of ite
```
dariakp, line 1428: cut off sentence, needs a note about the special significance of `0` (i.e. that it means return all, not return none)

test/unit/utils.test.js
```diff
@@ -297,4 +299,66 @@ describe('utils', function () {
...
+      expect(output).to.not.deep.equal(input);
+      expect(output).to.have.lengthOf(input.length);
+    });
+
+    it('should handle empty array', function () {
+      expect(shuffle([])).to.deep.equal([]);
+      expect(shuffle([], 0)).to.deep.equal([]);
+    });
+
+    it('should handle limit set to 0', function () {
```
dariakp, line 351: Since this is already partly covered by the test case in L337, can we get rid of this and just make a copy of that test case with the wording `should give a random shuffling of the entire input when limit is explicitly set to 0`

test/unit/utils.test.js
```diff
@@ -297,4 +299,66 @@ describe('utils', function () {
...
+      expect(shuffle(['a', 'b'])).to.have.lengthOf(2);
+      expect(shuffle(['a', 'b'], 0)).to.have.lengthOf(2);
+    });
+
+    it('should throw if limit is greater than zero and empty array', function () {
+      expect(() => shuffle([], 2)).to.throw(MongoRuntimeError);
```
dariakp, line 357: to test the boundary:
```suggestion
      expect(() => shuffle([], 1)).to.throw(MongoRuntimeError);
```

test/unit/utils.test.js
```diff
@@ -297,4 +299,66 @@ describe('utils', function () {
...
+
+    it('should give a random subset if limit is less than the input length', () => {
+      const input = ['a', 'b', 'c', 'd', 'e'];
+      const output = shuffle(input, 3);
+      expect(output).to.have.lengthOf(3);
+    });
+
+    const input = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i'];
+    for (let n = 1; n <= input.length; n++) {
+      it(`should give a random subset of length ${n}`, function () {
```
dariakp, line 326: I appreciate the idea here, but the resulting tests are going to read as "should give a random subset of length 5", for example, which doesn't clearly convey the intent; the important things to test for any given input are the boundaries (where the behavior changes): default, `0`, `1`, `n-1`, and `n`. We have default and 0 covered, so, I recommend taking the logic here using it to replace the test case in L318, and just test the `input.length - 1`. To cover `1`, we can actually have two test cases, one where the input is only 1 element, to make sure we return that element, and one where the input is greater (feel free to keep using the 9 element array), to prevent excessive flakiness, run the shuffle 5 times with limit 1 and then make sure that the set of all returned elements has size > 1. And then reuse a copy of the L337 test case to test `limit === input.length`.

dariakp (changes requested):
More feedback on just the shuffle portion

src/utils.ts
nbbeeken: done

test/unit/utils.test.js
```diff
@@ -297,4 +299,66 @@ describe('utils', function () {
...
+    it('should give a random shuffling of the entire input when no limit provided', () => {
+      const input = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i'];
+      const output = shuffle(input);
+      // Of course it is possible a shuffle returns exactly the same as the input
+      // but it is so improbable it is worth the flakiness in my opinion
+      expect(output).to.not.deep.equal(input);
+      expect(output).to.have.lengthOf(input.length);
+    });
+
+    it('should handle empty array', function () {
```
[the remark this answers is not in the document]
nbbeeken: :+1:

test/unit/utils.test.js
nbbeeken: done!

test/unit/utils.test.js
```diff
@@ -456,4 +458,92 @@ describe('utils', function () {
...
+    });
+
+    it('should not mutate the original input', function () {
+      const input = Object.freeze(['a', 'b', 'c', 'd', 'e']);
+      const output = shuffle(input); // This will throw if shuffle tries to edit the input
+      expect(output).to.not.deep.equal(input);
+      expect(output).to.have.lengthOf(input.length);
+    });
+
+    it(`should give a random subset of length input.length - 1`, function () {
```
dariakp, line 477:
```suggestion
    it(`should give a random subset of length equal to limit when limit is less than the input length`, function () {
```

test/unit/utils.test.js
```diff
@@ -456,4 +458,92 @@ describe('utils', function () {
...
+      const input = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i'];
+      const output = shuffle(input, input.length - 1);
+      expect(output).to.not.deep.equal(input);
+      expect(output).to.have.lengthOf(input.length - 1);
+    });
+
+    it(`should give a random subset of length input.length`, function () {
```
dariakp, line 484:
```suggestion
    it(`should give a random shuffling of the entire input when limit is equal to input length`, function () {
```

test/unit/utils.test.js
```diff
@@ -456,4 +458,92 @@ describe('utils', function () {
...
+        const output = shuffle(input);
+        expect(output).to.deep.equal(input);
+      }
+      for (let i = 0; i < 10; i++) {
+        const output = shuffle(input, 1); // and with limit
+        expect(output).to.deep.equal(input);
+      }
+    });
+
+    it(`should return a different item on every call of limit 1`, function () {
```
dariakp, line 503: small technicality, since we can't guarantee it'll be "different" on any given call
```suggestion
    it(`should return a random item on every call of limit 1`, function () {
```

test/unit/utils.test.js
dariakp: you added the new test cases but didn't delete this one

src/utils.ts
```diff
@@ -1405,3 +1419,32 @@ export function parsePackageVersion({ version }: { version: string }): {
   const [major, minor, patch] = version.split('.').map((n: string) => Number.parseInt(n, 10));
   return { major, minor, patch };
 }
+
+/**
+ * Fisher–Yates Shuffle
+ *
+ * Reference: https://bost.ocks.org/mike/shuffle/
+ * @param sequence - items to be shuffled
+ * @param limit - if nonzero shuffle will slice the randomized array e.g, `.slice(0, limit)`
```
dariakp, line 1428: add something along the lines of "otherwise will return the entire randomized array, same as in the case where the limit is not specified" ?

dariakp (changes requested):
just some wording clean up on the shuffle

src/sdam/srv_polling.ts
```diff
@@ -29,18 +29,16 @@ export class SrvPollingEvent {
...
-      })
-    );
+  hostnames(): Set<string> {
+    return new Set(this.srvRecords.map(r => HostAddress.fromSrvRecord(r).toString()));
   }
 }
 
 /** @internal */
 export interface SrvPollerOptions extends LoggerOptions {
+  rescanSrvIntervalMS?: number;
```
dariakp, line 39: I think we can remove this now

dariakp (changes requested):
Just capturing some thoughts before I move on to looking at more of this

src/connection_string.ts
```diff
@@ -136,6 +136,14 @@ export function resolveSRVRecord(options: MongoOptions, callback: Callback<HostA
           options.loadBalanced = true;
         }
 
+        if (
+          options.replicaSet &&
+          typeof options.srvMaxHosts === 'number' &&
```
dariakp, line 141: I think we should make the `srvMaxHosts` check consistent here and in the `validateLoadBalancedOptions` - i.e., either use "truthiness", or use the nonzero number check; I think at this point, the srvMaxHosts is always going to be a number, so we *can* use truthy for "non-zero", but we can also use `> 0` in both places to be more precise

dariakp (changes requested):
Thoughts on srvMaxHosts specifically

src/connection_string.ts
```diff
@@ -315,6 +320,16 @@ export function parseOptions(
 
   const allOptions = new CaseInsensitiveMap();
 
+  if (isSRV) {
```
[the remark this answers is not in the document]
dariakp: validateLoadBalancedOptions

src/sdam/topology_description.ts
```diff
@@ -139,23 +140,51 @@ export class TopologyDescription {
...
+    const incomingHostnames = ev.hostnames();
+    const currentHostnames = new Set(this.servers.keys());
+
+    const hostnamesToRemove = new Set<string>();
+    for (const hostname of currentHostnames) {
+      if (!incomingHostnames.has(hostname)) {
+        // If the SRV Records no longer include this hostname
+        // we have to stop using it
+        hostnamesToRemove.add(hostname);
+      }
```
dariakp, line 154: I think if we seed `hostnamesToAdd` with the incomingHostnames and add an else loop to remove the matching names here, you'll be left with the correct set of `hostnamesToAdd` without needing the second loop below

src/utils.ts
```diff
@@ -1405,3 +1419,32 @@ export function parsePackageVersion({ version }: { version: string }): {
...
+ */
+export function shuffle<T>(sequence: Iterable<T>, limit = 0): Array<T> {
+  const items = Array.from(sequence); // shallow copy in order to never shuffle the input
+
+  if (limit > items.length) {
+    throw new MongoRuntimeError('Limit must be less than the number of items');
+  }
+
+  let remainingItemsToShuffle = items.length;
+  while (remainingItemsToShuffle !== 0) {
```
dariakp, line 1438: can we optimize this very slightly by changing this to `> 1` ? Technically, on that last pass, we have already shuffled all we can, right? So the `randomIndex` can only be `0`, and remaining placement can only be `0`

dariakp (changes requested):
More notes

src/utils.ts
dariakp: follow up question, if this does in fact generate a random sorting of the array, wouldn't the backward sorting be as random as the forward sorting? since after we go though `x` iterations of the loop, the last `x` elements will never be changed, so if we have a limit, say `n`, couldn't we stop after `n` iterations and return the corresponding slice from the end of the array?

test/unit/utils.test.js
```diff
@@ -3,13 +3,15 @@ const {
...
+  shuffle
 } = require('../../src/utils');
 const { expect } = require('chai');
 const sinon = require('sinon');
+const { MongoRuntimeError } = require('../../src/error');
 
-describe('utils', function () {
-  context('eachAsync', function () {
+describe('driver utils', function () {
+  context('eachAsync function', function () {
```
dariakp, line 14: purely preference, but have you considered using parentheses to indicate that a method is being tested? e.g., `describe('eachAsync()')`

src/connection_string.ts
```diff
@@ -360,6 +375,18 @@ export function parseOptions(
     );
   }
 
+  if (
```
dariakp, line 378: can this check be in the else clause of the `if (isSRV)` check above that? would be nice to have srv-related stuff grouped together, and this isn't relying on the processed option values anyway

src/connection_string.ts
```diff
@@ -924,6 +955,13 @@ export const OPTIONS = {
     default: 0,
     type: 'uint'
   },
+  srvMaxHosts: {
+    type: 'uint'
```
dariakp, line 959: why don't we default this to 0 here?

dariakp (changes requested):
Still have tests to review, but here's more feedback

src/connection_string.ts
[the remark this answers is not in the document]
nbbeeken: I'm confused on the second point here. The modification I made to validateLoadBalancedOptions will error if it is an SRV connection string and LB + srvMaxHosts are combined. If you don't have an srv connection string then srvMaxHosts is already incorrect usage.

src/utils.ts
nbbeeken: We can do the `> 1` change but I think we run into an issue if we try and shuffle some subset. Our test that does a limit of 1, 5 times over, and expects at least 2 unique values started failing. Because now, among the "large" input array it's very unlikely that we get the chance to shuffle the first item out of place, such that the concluding slice returns something different.

test/unit/utils.test.js
nbbeeken: It's maybe a bad reason but if we don't name our tests something different than the variable that gets imported at the top, the explorer puts the run/debug buttons up at the require, instead of above the test. So I'm up for any naming, its just a small QoL.

src/connection_string.ts
```diff
@@ -309,10 +300,58 @@ export function parseOptions(
...
+    const noUserSpecifiedTLS = !objectOptions.has('tls') && !urlOptions.has('tls');
+    const noUserSpecifiedSSL = !objectOptions.has('ssl') && !urlOptions.has('ssl');
+    if (noUserSpecifiedTLS && noUserSpecifiedSSL) {
+      objectOptions.set('tls', true);
+      objectOptions.set('ssl', true);
+    }
+  } else {
+    if (srvSpecificOptionsSpecified) {
+      throw new MongoParseError(
+        'Cannot use maxSrvHosts nor srvServiceName with a non-srv connection string'
```
dariakp, line 339:
```suggestion
        'Cannot use srvMaxHosts or srvServiceName with a non-srv connection string'
```

src/connection_string.ts
```diff
@@ -309,10 +300,58 @@ export function parseOptions(
...
+      );
+    }
+  }
+
+  if (nonZeroSrvMaxHosts) {
+    if (urlOptions.get('loadBalanced') === true) {
+      throw new MongoParseError('Cannot use srvMaxHosts option with loadBalanced');
```
dariakp, line 346: this duplicates the logic in validateLoadBalancedOptions, and it throws a different error than the one thrown by validateLoadBalancedOptions - we probably don't want that

src/connection_string.ts
```diff
@@ -309,10 +300,58 @@ export function parseOptions(
...
+  }
+
   if (objectOptions.has('loadBalanced')) {
     throw new MongoParseError('loadBalanced is only a valid option in the URI');
   }
 
+  // SRV connection string validations
+
+  const nonZeroSrvMaxHosts =
+    objectOptions.get('srvMaxHosts') > 0 || urlOptions.get('srvMaxHosts') > 0;
```
dariakp, line 318: I think this is better to do after the options are parsed because this is really only for checking against the replicaSet (see my other comment on the redundant loadbalanced check), if we do this just before the loadBalanced validation below, we'll be able to read the srvMaxHosts from the already parsed options (instead of trying to manually read them and bypassing the validation on the option definitions themselves; this would also ensure the correct precedence is maintained if option is specified in more than one place: in the above logic, if srvMaxHosts is set to 0 in the object but not in the uri, we would throw an error, even tough the objectOptions should override it and not throw).

src/connection_string.ts
```diff
@@ -289,12 +286,6 @@ export function parseOptions(
...
 
-    if (key.toLowerCase() === 'serverapi') {
-      throw new MongoParseError(
-        'URI cannot contain `serverApi`, it can only be passed to the client'
-      );
-    }
-
     if (key.toLowerCase() === 'authsource' && urlOptions.has('authSource')) {
       // If authSource is an explicit key in the urlOptions we need to remove the implicit dbName
       urlOptions.delete('authSource');
```
dariakp, line 291: Is this a bug? Just noticed it says dbName needs to be removed, and that's also the only thing actually set in the code above, so deleting authSource does nothing

dariakp (changes requested):
Some feedback on just the code, maybe hopefully the last round, still need to look at the tests

test/functional/uri_options_spec.test.js
```diff
@@ -24,10 +26,46 @@ describe('URI Options (spec)', function () {
...
+        try {
+          driverOptions = parseOptions(test.uri);
+          hosts = await promisify(resolveSRVRecord)(driverOptions);
+        } catch (error) {
+          thrownError = error;
+        }
+
+        if (test.valid === false || test.warning === true) {
+          // We implement warnings as errors
+          expect(thrownError).to.exist;
```
dariakp, line 57: can we assert on the error type? (MongoParseError?)

dariakp (changes requested):
URI spec test portion comments

test/unit/connection_string.test.js
```diff
@@ -224,81 +225,96 @@ describe('Connection String', function () {
...
+              }
+              if (test.parsed_options && test.parsed_options.user && test.parsed_options.password) {
+                expect(options.credentials.username).to.equal(test.parsed_options.user);
+                expect(options.credentials.password).to.equal(test.parsed_options.password);
+              }
+
+              // TODO(): srvMaxHost limiting happens in the callback given to resolveSRVRecord inside the driver
+              // How can we test this in unit test?
+
+              // if (options.srvHost && comment.includes('srvMaxHosts')) {
```
durran, line 293: Just a note but it's potentially one way to add a check... Could be that we test that the list of hosts after the randomisation is equal to the list of server description host names in the created topology? Just thinking aloud.

src/utils.ts
```diff
@@ -1405,3 +1419,33 @@ export function parsePackageVersion({ version }: { version: string }): {
...
+ */
+export function shuffle<T>(sequence: Iterable<T>, limit = 0): Array<T> {
+  const items = Array.from(sequence); // shallow copy in order to never shuffle the input
+
+  if (limit > items.length) {
+    throw new MongoRuntimeError('Limit must be less than the number of items');
+  }
+
+  let remainingItemsToShuffle = items.length;
+  const lowerBound = limit === 0 ? 1 : items.length - limit;
```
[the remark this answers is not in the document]
nbbeeken: Nice catch on that edge case! I have to adjust the slice condition as well, to account for lowerBound always being at least 1. That's my way around overwriting the parameters which I prefer not to do (just an onion tho)

src/connection_string.ts
nbbeeken: Hmm this code is strange but I'm also not sure what it is trying to accomplish, I reviewed the specs and I don't see where dbName and authSource would conflict with one another. and implicit `dbName` implies to me refering to how dbName defaults to `test` but that doesn't happen before this point. 

I'm going to try deleting it, lets see what the waterfall says, but we can keep it and file a ticket.

dariakp: Deleting it shouldn't do anything, but I meant it should maybe be updated to delete 'dbName' instead, because of the code above that does:
` if (url.pathname !== '/' && url.pathname !== '') {
    const dbName = decodeURIComponent(
      url.pathname[0] === '/' ? url.pathname.slice(1) : url.pathname
    );
    if (dbName) {
      urlOptions.set('dbName', [dbName]);
    }
  }`
Clearly the behavior isn't tested or there would be failures.

src/utils.ts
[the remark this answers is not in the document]
dariakp: that's why I said do the same check in L1450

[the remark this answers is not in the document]
nbbeeken: My mistake! overlooked that mention 😅

nbbeeken:
@dariakp I went through and made sure I addresses all your current comments.
@durran I've rewritten the commented out testing in connection_string.test.js, its close to working, I just need to do so much manual address lookups in the tests to make sure we get the same results expected by the tests. So everything expected is there, we just have slightly different shaped data.

test/unit/mongo_client.test.js
```diff
@@ -719,4 +720,59 @@ describe('MongoOptions', function () {
...
+    try {
+      options = parseOptions('mongodb+srv://localhost.a.com', { srvServiceName: 'a'.repeat(255) });
+      await promisify(resolveSRVRecord)(options);
+    } catch (error) {
+      thrownError = error;
+    }
+    expect(thrownError).to.have.property('code', 'EBADNAME');
+  });
+
+  it('srvServiceName should not error if it greater than 15 characters', async () => {
```
dariakp, line 766:
```suggestion
  it('srvServiceName should not error if it is greater than 15 characters as long as the DNS query limit is not surpassed', async () => {
```

test/unit/mongo_client.test.js
```diff
@@ -719,4 +720,59 @@ describe('MongoOptions', function () {
...
+    }).to.throw(MongoParseError, 'Cannot limit srv hosts with loadBalanced enabled');
+    expect(() => {
+      new MongoClient('mongodb+srv://localhost', { srvMaxHosts: 2, replicaSet: 'blah' });
+    }).to.throw(MongoParseError, 'Cannot use srvMaxHosts option with replicaSet');
+    expect(() => {
+      new MongoClient('mongodb+srv://localhost?loadBalanced=true', { srvMaxHosts: 2 });
+    }).to.throw(MongoParseError, 'Cannot limit srv hosts with loadBalanced enabled');
+  });
+
+  it('srvServiceName and srvMaxHosts cannot be used on a non-srv connection string', () => {
```
dariakp, line 739: we should add conditions for `srvMaxHosts` set to 0, since for non-srv we have to throw regardless of whether it's specifically a positive integer

test/unit/mongo_client.test.js
```diff
@@ -719,4 +720,59 @@ describe('MongoOptions', function () {
       expect(parse).to.throw(/URI/);
     });
   });
+
+  it('srvMaxHosts cannot be combined with LB or ReplicaSet', () => {
```
dariakp, line 724:
```suggestion
  it('srvMaxHosts > 0 cannot be combined with LB or ReplicaSet', () => {
```

test/unit/mongo_client.test.js
```diff
@@ -719,4 +720,59 @@ describe('MongoOptions', function () {
...
+      new MongoClient('mongodb+srv://localhost?srvMaxHosts=2&loadBalanced=true');
+    }).to.throw(MongoParseError, 'Cannot limit srv hosts with loadBalanced enabled');
+    expect(() => {
+      new MongoClient('mongodb+srv://localhost', { srvMaxHosts: 2, replicaSet: 'blah' });
+    }).to.throw(MongoParseError, 'Cannot use srvMaxHosts option with replicaSet');
+    expect(() => {
+      new MongoClient('mongodb+srv://localhost?loadBalanced=true', { srvMaxHosts: 2 });
+    }).to.throw(MongoParseError, 'Cannot limit srv hosts with loadBalanced enabled');
+  });
+
```
dariakp, line 738: Should we also add tests that we don't throw on `srvMaxHosts === 0` even with lb or replicaset?

src/operations/connect.ts
```diff
@@ -51,8 +51,9 @@ export function connect(
   if (typeof options.srvHost === 'string') {
     return resolveSRVRecord(options, (err, hosts) => {
       if (err || !hosts) return callback(err);
-      for (const [index, host] of hosts.entries()) {
-        options.hosts[index] = host;
+
+      for (let index = 0; index < hosts.length; index++) {
```
dariakp, line 55: was there something wrong with the original for..of loop? if not, let's put it back exactly how it was so that it doesn't show in the changed files list

src/sdam/topology.ts
```diff
@@ -290,8 +293,13 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
     const topologyType = topologyTypeFromOptions(options);
     const topologyId = globalTopologyCounter++;
 
+    const selectedHosts =
+      options.srvMaxHosts === 0 || options.srvMaxHosts >= seedlist.length
+        ? seedlist
+        : shuffle(seedlist, options.srvMaxHosts);
```
nbbeeken, line 299: I moved the initial shuffling to here, in order to unit test since the callback to resolveSRVRecord is inaccessible to the tests.

dariakp (changes requested):
Some comments while I look at the tests

src/sdam/topology.ts
dariakp: seems like that works too

dariakp (changes requested):
Just to summarize what we discussed - we want to pull in the spec changes associated with test 13, which involve NODE-3450 along with the changes that would have been done in NODE-3680; we also want to make the mock tests more closely resemble the prose tests as written, and we want to break out the seed discovery tests into their own file and see if they can run as integration tests.

test/unit/polling_srv_records_for_mongos_discovery.prose.test.ts
```diff
@@ -0,0 +1,258 @@
...
+  });
+
+  it('13 - DNS record with custom service name can be found', async () => {
+    client = new MongoClient(SRV_CONNECTION_STRING, {
+      tls: false,
+      srvServiceName: 'myFancySrvServiceName',
+      serverSelectionTimeoutMS: 5000
+    });
+
+    makeStubs({ srvServiceName: 'myFancySrvServiceName' });
```
dariakp, line 246: I think it would be easy enough to make this match the actual spec test case - specify one initial record on port 2017, then return two records for ports 2019 and 2020, so why not do that?

dariakp (changes requested):

dariakp (approved):

test/unit/connection_string.test.js
nbbeeken: Moved these tests to [initial_dns_seedlist_discovery.spec.test.ts](https://github.com/mongodb/node-mongodb-native/blob/NODE-3467/limit-mongos/test/integration/initial-dns-seedlist-discovery/initial_dns_seedlist_discovery.spec.test.ts#L20) to go along with our nice naming convention but also skipped them due to needing more concrete assertions around the error cases. And it is tracked here: NODE-3757

PR #3032, opened 2021-11-09
fix(NODE-3344): allow setting `defaultTransactionOptions` with POJO rather than ReadConcern instance

durran:
Re: https://github.com/Automattic/mongoose/issues/10805

### Description

Right now, `startSession({ defaultTransactionOptions: { readConcern: { level: 'snapshot' } } })` fails to compile with a `Property 'toJSON' is missing in type '{ level: string; }' but required in type 'ReadConcern'.` error. This seems like something that should change.

#### What is changing?

`TransactionOptions.readConcern` is now a `ReadConcernLike`, which allows using strings or `{ level }` object.

Make `TransactionOptions` more user-friendly and line up with `Cursor#withReadConcern()` and `CollectionOptions`.

### DIFF
--- a/test/types/mongodb.test-d.ts
+++ b/test/types/mongodb.test-d.ts
@@ -6,6 +6,8 @@ import type { FindCursor } from '../../src/cursor/find_cursor';
 import type { Document } from 'bson';
 import { Db } from '../../src';
 import { Topology } from '../../src/sdam/topology';
+import type { ClientSession } from '../../src/sessions';
+import { ReadConcern, ReadConcernLevel } from '../../src/read_concern';
 import * as MongoDBDriver from '../../src';
 
 // We wish to keep these APIs but continue to ensure they are marked as deprecated.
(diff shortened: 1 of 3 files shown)


test/types/mongodb.test-d.ts
```diff
@@ -38,6 +40,14 @@ const composedMap = mappedAgg.map<string>(x => x.toString());
 expectType<AggregationCursor<string>>(composedMap);
 expectType<string | null>(await composedMap.next());
 expectType<string[]>(await composedMap.toArray());
+expectType<ClientSession>(
```
dariakp, line 43: Could we move these into a root level `test/types/sessions.test-d.ts`, adding a comment above along the lines of `// should allow ReadConcern or ReadConcernLike as readConcern in defaultTransactionOptions`, and then also add a sanity test like `expectError(client.startSession({ ... { readConcern: <some invalid parameter> } }))`

dariakp (changes requested):

test/types/mongodb.test-d.ts
durran: Moved to new root test in `test/types/sessions.test-d.ts` and added error check.

dariakp (approved):

nbbeeken (approved):
LGTM!

PR #3033, opened 2021-11-09
refactor(NODE-3273): remove undefined error param from topology open event

ljhaywar:
### Description

[NODE-3273](https://jira.mongodb.org/browse/NODE-3273) Remove undefined error parameter from topology "open" event

#### What is changing?

The error parameter is no longer being passed on the topology "open" event

No, this is an internal change

The error is always undefined based on the flow of the code, so there is no need to pass it.

### DIFF
--- a/test/functional/connection.test.js
+++ b/test/functional/connection.test.js
@@ -72,8 +72,8 @@ describe('Connection - functional', function () {
   });
 
   it('should only pass one argument (topology and not error) for topology "open" events', function (done) {
-    var configuration = this.configuration;
-    var client = configuration.newClient({ w: 1 }, { maxPoolSize: 1 });
+    const configuration = this.configuration;
+    const client = configuration.newClient({ w: 1 }, { maxPoolSize: 1 });
 
     client.on('topologyOpening', () => {
       client.topology.on('open', (...args) => {
--- a/test/functional/connection.test.js
+++ b/test/functional/connection.test.js
@@ -13,14 +13,25 @@ describe('Connection - functional', function () {
     return setupDatabase(this.configuration);
   });
 
-  afterEach(done => {
+  afterEach(async () => {
+    let savedError;
     if (client) {
-      client.close();
+      try {
+        await client.close();
+      } catch (err) {
+        savedError = err;
+      }
     }
     if (testClient) {
-      testClient.close();
+      try {
+        await testClient.close();
+      } catch (err) {
+        savedError = err;
+      }
     }
-    done();
+    if (savedError) {
+      throw savedError;
+    }
   });
 
   it('should correctly start monitoring for single server connection', {
@@ -41,12 +52,10 @@ describe('Connection - functional', function () {
         isMonitoring = event instanceof ServerHeartbeatStartedEvent;
       });
 
-      client
-        .connect()
-        .then(() => {
-          expect(isMonitoring);
-        })
-        .finally(done);
+      client.connect().then(() => {
+        expect(isMonitoring);
+        done();
+      });
     }
   });
 
@@ -162,7 +171,7 @@ describe('Connection - functional', function () {
       client = configuration.newClient();
 
       client.connect(
-        connectionTester(configuration, 'testConnectNoOptions', function (client) {
+        connectionTester(configuration, 'testConnectNoOptions', function () {
           done();
         })
       );
@@ -227,7 +236,7 @@ describe('Connection - functional', function () {
         testClient = configuration.newClient(opts);
 
         testClient.connect(
-          connectionTester(configuration, 'testConnectGoodAuthAsOption', function (client) {
+          connectionTester(configuration, 'testConnectGoodAuthAsOption', function () {
             done();
           })
         );
--- a/test/functional/connection.test.js
+++ b/test/functional/connection.test.js
@@ -6,8 +6,16 @@ const { ServerHeartbeatStartedEvent, MongoClient } = require('../../src');
 const { Topology } = require('../../src/sdam/topology');
 
 describe('Connection - functional', function () {
+  let client;
+
   before(function () {
     return setupDatabase(this.configuration);
+  });
+
+  afterEach(done => {
+    if (client) {
+      client.close(done);
+    }
   });
 
   it('should correctly start monitoring for single server connection', {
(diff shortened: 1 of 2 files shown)


test/functional/connection.test.js
```diff
@@ -70,6 +71,21 @@ describe('Connection - functional', function () {
     }
   });
 
+  it('topology "open" event should only pass topology (and not error)', function (done) {
```
gjchong25, line 74: Usually the test description reads like a sentence (ex. "it should correctly create X" or "it should throw an error if Y happens", you can see more examples within this connection.test file). I would try to stay consistent with that wording. Maybe you could say something like `'should only pass one topology argument for topology "open" events'`, or you could create another `describe` block if you think what you're testing should fall under a different category than the `Connection - functional` descriptor.

gjchong25 (commented):
Make sure to add a description of your PR in the Conversation tab (look at other PRs for reference). You also don't need to make your PR a draft, I generally just make a PR and you can add more commits and changes to address PR feedback as you go. There's also an evergreen test that seems to be failing in `test/functional/connections_stepdown.test.js` which I would take a look at (you can look at the specific error message in the Logs tab.

ljhaywar:
@gjchong25 Thanks for reviewing! I hadn't had a chance to clean up the PR yet, which is why I put it in Draft. Is there another way to indicate the PR is not yet ready for review?

I've added a commit to address the test wording suggestion.  Do you all typically squash and merge, or do I need to squash these myself?

I'll figure out what's going on with that failing test.

ljhaywar:
Looks like the tests are all passing after I pushed the commit with your wording suggestion change. It must have been a flakey failure.

gjchong25 (approved):

gjchong25:
I usually assume that a PR isn't ready for review unless you request review from someone on the team (maybe this differs between teammates, though). In terms of squashing and merging, the implementer doesn't merge but instead the primary reviewer, once resolving all the comments, will then request team review and then merge all your changes once the second reviews are in. [Here](https://wiki.corp.mongodb.com/display/DRIVERS/Node+Team+Practices) is a more comprehensive guide of the dev process!

test/functional/connection.test.js
```diff
@@ -70,6 +71,21 @@ describe('Connection - functional', function () {
     }
   });
 
+  it('should only pass one argument (topology and not error) for topology "open" events', function (done) {
+    var configuration = this.configuration;
+    var client = configuration.newClient({ w: 1 }, { maxPoolSize: 1 });
```
nbbeeken, line 76:
```suggestion
    const configuration = this.configuration;
    const client = configuration.newClient({ w: 1 }, { maxPoolSize: 1 });
```

We tend to modernize code as we come across it, I know you were going off the nearby test but its nice to try and update things as we get to them.

nbbeeken (changes requested):
LGTM, and nice test! Thanks!

test/functional/connection.test.js
ljhaywar: Makes sense. Thanks!

nbbeeken (approved):

test/functional/connection.test.js
```diff
@@ -70,6 +71,21 @@ describe('Connection - functional', function () {
     }
   });
 
+  it('should only pass one argument (topology and not error) for topology "open" events', function (done) {
+    const configuration = this.configuration;
+    const client = configuration.newClient({ w: 1 }, { maxPoolSize: 1 });
+
+    client.on('topologyOpening', () => {
+      client.topology.on('open', (...args) => {
+        expect(args.length).to.equal(1);
```
dariakp, line 80: using a more precise assertion will leave a better error message if this fails
```suggestion
        expect(args).to.have.lengthOf(1);
```

test/functional/connection.test.js
```diff
@@ -70,6 +71,21 @@ describe('Connection - functional', function () {
...
+        expect(args[0]).to.be.instanceOf(Topology);
+        client.close(done);
```
dariakp, line 82: as written, this won't execute if the assertions above fail; for that reason, we want to write our clean up steps in an `afterEach` block; for this test, given that the other tests in this file suffer from a similar flaw, I recommend declaring a `let client;` at the top of the describe block, and then assigning to that variable in L76; in this line, you can just call `done()`, and then you need to add an `afterEach(done => {...})` block where you do `if (client) { client.close(done) }`

dariakp (changes requested):

ljhaywar:
@dariakp Thanks for the feedback!  I made the changes you suggested. It felt odd to not have every test in that file use the afterEach(), so I updated them to use it.  I put those changes in a separate commit in case you don't think I should be doing that as part of this PR.  

There is some strange test structure in some of the tests that use a restOfTest() function, but I left that alone as it wasn't directly related to this work.

ljhaywar:
Just saw these failures. Will fix this up this afternoon.

test/functional/connection.test.js
```diff
@@ -27,12 +41,12 @@ describe('Connection - functional', function () {
...
       });
 
-      return client
+      client
         .connect()
         .then(() => {
           expect(isMonitoring);
         })
-        .finally(() => client.close());
+        .finally(done);
```
dariakp, line 49: since you are rewriting this section to use a `done()` callback instead of the async return syntax, you don't need a finally anymore, you can delete this line and move the `done()` to the `.then` clause, to run after expect

test/functional/connection.test.js
```diff
@@ -3,17 +3,31 @@ const { withClient, setupDatabase } = require('./shared');
...
   });
 
+  afterEach(done => {
+    if (client) {
+      client.close();
+    }
+    if (testClient) {
+      testClient.close();
+    }
+    done();
```
dariakp, line 23: This is not quite right: the `.close()` api is asynchronous, so we can't know that it succeeded without awaiting it or passing in a callback (note the syntax I had in my original code). If you want to handle proper cleanup for multiple async calls, it gets a bit messy, since you want as much of the cleanup to happen as possible. You'll basically want to rewrite your afterEach to use the async form: `afterEach(async () => { ... });`, inside, you'll want to do something like:
```
let savedError;
if (client) {
  try { await client.close(); } catch (err) { savedError = err; }
}
if (testClient) {
  try { await testClient.close(); } catch (err) { savedError = err; }
}
if (savedError) { throw savedError; }
```

dariakp (changes requested):
I don't mind the slight scope increase for the sake of test clean up, however, the situations in the different test cases here are all somewhat different from each other, and require careful handling of the async code

test/functional/connection.test.js
ljhaywar: Thank you for this! I had tried a few different options that didn't work this morning.

dariakp (approved):

ljhaywar:
@dariakp Thanks for the detailed feedback!

PR #3034, opened 2021-11-09
feat(NODE-3692): make change stream events typing more generic

durran:
Added typescript generic typing to ChangeStreamEvents

## Description

**What changed?**
The typing for `ChangeStreamEvents` was not generic, and did not contain `<TSchema>` so it was always assumed to be of type `Document`. This would affect code usage like this:
```typescript
interface MyDoc extends Document{
   name: string
}
const client = new MongoClient('');
const coll = client.db('test').collection<MyDoc>('test');
coll.watch().on('change', c => c.fullDocument) 
// c.fullDocument above is typed as Document, but should be MyDoc
```
`ChangeStreamDocument` already had generic types, so I just passed through `<TSchema>`.

I tried it out locally and the typing showed correctly in my editor.

Could someone confirm that the `init` and `more` events do indeed return the same structure as `TSchema`, as I do not know what these events are?

### DIFF
--- a/test/types/mongodb.test-d.ts
+++ b/test/types/mongodb.test-d.ts
@@ -21,11 +21,13 @@ expectDeprecated(MongoDBDriver.ObjectID);
 expectNotDeprecated(MongoDBDriver.ObjectId);
 
 interface TSchema extends Document {
+  name?: string;
 }
 
 // test mapped cursor types
 const client = new MongoClient('');
-const coll = client.db('test').collection('test');
+const db = client.db('test');
+const coll = db.collection('test');
 const findCursor = coll.find();
 expectType<Document | null>(await findCursor.next());
 const mappedFind = findCursor.map<number>(obj => Object.keys(obj).length);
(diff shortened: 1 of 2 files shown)


test/types/mongodb.test-d.ts
```diff
@@ -38,6 +39,16 @@ const composedMap = mappedAgg.map<string>(x => x.toString());
...
 expectType<string[]>(await composedMap.toArray());
+const changeStream = coll.watch();
+changeStream.on('init', (doc) => {
+  expectType<Document>(doc);
+});
+changeStream.on('more', (doc) => {
+  expectType<Document | undefined>(doc);
+});
+changeStream.on('change', (doc) => {
+  expectType<ChangeStreamDocument<Document>>(doc);
```
nbbeeken, line 50: I think we need to test with a schema here to see if the types propagate correctly.

durran: Updated.

test/types/mongodb.test-d.ts
```diff
@@ -19,9 +20,14 @@ expectDeprecated(Db.prototype.unref);
 expectDeprecated(MongoDBDriver.ObjectID);
 expectNotDeprecated(MongoDBDriver.ObjectId);
 
+interface TSchema extends Document {
+  name?: string;
```
dariakp, line 24: I think this needs to have at least one non-optional parameter, because otherwise a generic document type might still match the TSchema

dariakp (changes requested):

test/types/mongodb.test-d.ts
dariakp: (This is more of an issue for something like expectAssignable and isn't a concern here where we use expectType, since the type check is strict, but better safe than sorry)

durran: Updated to make it required.

dariakp (approved):

nbbeeken (approved):

PR #3035, opened 2021-11-09
chore: script populating integration spec topics

nbbeeken:
#### What is changing?

A helper script to assist with test clean up work.

I want the integration folder to remain closely in sync with the spec repo. A script to assist with that seems like a good way to keep with the practice. There are folders here which I believe may remain empty for sometime, but I think at somepoint we will either move in or have some case where it makes sense to put a test under a given topic.

-  Ran `npm run check:lint` script
- Changes are covered by tests
- New TODOs have a related JIRA ticket

### DIFF
--- a/etc/populate_spec_topics.sh
+++ b/etc/populate_spec_topics.sh
@@ -1,0 +1,23 @@
+#! /usr/bin/env bash
+
+if [[ -z $MONGODB_SPECIFICATIONS_DIRECTORY ]]; then
+    tmpdir=$(mktemp -d -t spec_testsXXXX)
+    curl -sL "https://github.com/mongodb/specifications/archive/master.zip" -o "$tmpdir/specs.zip"
+    unzip -d "$tmpdir" "$tmpdir/specs.zip" > /dev/null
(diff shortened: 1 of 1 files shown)


dariakp:
@nbbeeken Do we actually want folders for top-level .rst files?

nbbeeken:
Fair point I may be over prescribing us here. I think creating the missing directories might be worth it though? Which would be:
```
benchmarking
bson-corpus
bson-decimal128
causal-consistency
collation
compression
connections-survive-step-down
message
mongodb-handshake
ocsp-support
polling-srv-records-for-mongos-discovery
serverless-testing
transactions-convenient-api
```

I think each one of these would get used by our own collection of additional tests, (in addition to the spec runners). Also I don't think there would be harm in at some point running the bson corpus test with our driver. As well as making the spec benchmarks a regular part of the test runs.

I could easily extend this script to do similar directory creation or just file existence check in the unit directory. So we could have something that recreates the driver's structure. Worth doing?

dariakp (approved):

PR #3039, opened 2021-11-16
feat(NODE-3729): add withId to default return type for collection.find and collection.findOne

PaulEndri:
#### What is changing?
Adding a `WithId` wrapper to the return type of collection.findOne and collection.find

No, should be self documented by ts output and is otherwise (probably) expected behavior

Ran into a scenario similar to [NODE-3729](https://jira.mongodb.org/browse/NODE-3729) and after seeing [PR 3020](https://github.com/mongodb/node-mongodb-native/pull/3020) I wanted to suggest a more targeted solution that only impacts the find methods directly.

By changing only the return type of `.find()`; the `_id` property is correctly returned on all find operations. I limited the change to only instances where the return type for `.find()` isn't overridden because at that point the user is stating they want manual control over the resulting type and are responsible for their own `_id` definition (if one exists).

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -676,12 +676,16 @@ export class Collection<TSchema extends Document = Document> {
    * @param options - Optional settings for the command
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
-  findOne(): Promise<TSchema | null>;
-  findOne(callback: Callback<TSchema | null>): void;
-  findOne(filter: Filter<TSchema>): Promise<TSchema | null>;
-  findOne(filter: Filter<TSchema>, callback: Callback<TSchema | null>): void;
-  findOne(filter: Filter<TSchema>, options: FindOptions): Promise<TSchema | null>;
-  findOne(filter: Filter<TSchema>, options: FindOptions, callback: Callback<TSchema | null>): void;
+  findOne(): Promise<WithId<TSchema> | null>;
+  findOne(callback: Callback<WithId<TSchema> | null>): void;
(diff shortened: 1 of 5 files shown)


dariakp (approved):

nbbeeken (approved):
Thanks! LGTM

src/collection.ts
```diff
@@ -725,10 +729,10 @@ export class Collection<TSchema extends Document = Document> {
...
    * @param filter - The filter predicate. If unspecified, then all documents in the collection will match the predicate
    */
-  find(): FindCursor<TSchema>;
-  find(filter: Filter<TSchema>, options?: FindOptions): FindCursor<TSchema>;
-  find<T>(filter: Filter<TSchema>, options?: FindOptions): FindCursor<T>;
-  find(filter?: Filter<TSchema>, options?: FindOptions): FindCursor<TSchema> {
+  find(): FindCursor<WithId<TSchema>>;
+  find(filter: Filter<WithId<TSchema>>, options?: FindOptions): FindCursor<WithId<TSchema>>;
+  find<T>(filter: Filter<WithId<TSchema>>, options?: FindOptions): FindCursor<T>;
+  find(filter?: Filter<WithId<TSchema>>, options?: FindOptions): FindCursor<WithId<TSchema>> {
```
avaly, line 735: @PaulEndri I think adding `WithId` to the `Filter` type here was a mistake. The other `Filter` types in this file are not using it.

LonelessCodes:
Should `ModifyResult<TSchema>.value` also be `WithId<TSchema>`? Because I believe this might be the same case.

PR #3040, opened 2021-11-17
fix(NODE-3765): make replacement for replaceOne operations without _id

tusbar:
#### What is changing?

`collection.replaceOne`, `collection.findOneAndReplace` and the `replaceOne` bulk operation now take a `WithoutId<TModel>` type for the replacement document.

I don’t think so. It should be self-documented from the tsdoc.

- The `_id` field is immutable, so it shouldn’t be required as it has to be the same value when passed.
- This makes `replaceOne` more in line with `insertOne`.

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -93,6 +93,7 @@ import type {
   TODO_NODE_3286,
   UpdateFilter,
   WithId,
+  WithoutId,
   OptionalId,
   Flatten
 } from './mongo_types';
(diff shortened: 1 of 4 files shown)


test/types/community/collection/replaceX.test-d.ts
```diff
@@ -0,0 +1,22 @@
...
+import { MongoClient, ObjectId } from '../../../../src';
+
+// test collection.replaceX functions
+const client = new MongoClient('');
+const db = client.db('test');
+
+interface TestModel {
+  _id: ObjectId;
+  stringField: string;
+}
```
nbbeeken, line 11: Can basically take a clone of this schema, except try a Model that has an _id that is a custom type like `_id: number`

src/collection.ts
```diff
@@ -459,26 +459,29 @@ export class Collection<TSchema extends Document = Document> {
    * @param options - Optional settings for the command
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
-  replaceOne(filter: Filter<TSchema>, replacement: TSchema): Promise<UpdateResult | Document>;
   replaceOne(
     filter: Filter<TSchema>,
-    replacement: TSchema,
+    replacement: OptionalId<TSchema>
```
nbbeeken, line 464: I think we should instead change these to `WithoutId` whether or not you are using ObjectId the _id field is always immutable, so if its ever specified here you get an error from the server. Might as well catch that in TS here!

src/collection.ts
```diff
@@ -459,26 +459,29 @@ export class Collection<TSchema extends Document = Document> {
...
     filter: Filter<TSchema>,
-    replacement: TSchema,
+    replacement: OptionalId<TSchema>,
     options: ReplaceOptions,
     callback: Callback<UpdateResult | Document>
   ): void;
   replaceOne(
     filter: Filter<TSchema>,
-    replacement: TSchema,
+    replacement: OptionalId<TSchema>,
```
nbbeeken, line 484: We should also update the `findOneAndReplace` APIs (around line 1280)

nbbeeken (changes requested):
Thanks so much for your help here! I think we can take this in a bit more of a strict direction

src/collection.ts
tusbar: Yeah, fair enough.

src/collection.ts
tusbar: Done

tusbar: These didn’t even use the generic type.

nbbeeken (approved):
LGTM, (test failures are unrelated and already are fixed on main)

test/types/community/collection/replaceX.test-d.ts
dariakp: @nbbeeken Just wanted to follow up on this - could you clarify what you mean? Were you asking for a new test case?

dariakp (commented):
LGTM, but wanted to follow up on this orphaned comment

test/types/community/collection/replaceX.test-d.ts
nbbeeken: Yea my comment is no longer relevant, when using `OptionalId` we'd have to test for `ObjectId` type and non-`ObjectId` type. But `WithoutId` omits the `_id` altogether so we don’t need the dupe test for other types.

dariakp (approved):

baileympearson (approved):
LGTM!

PR #3041, opened 2021-11-17
feat(NODE-3633): add Socks5 support

addaleax:

Yes. This adds connection string parameters and MongoClient options.

See NODE-3633 and adjacent tickets

- [x] Include changes from https://github.com/mongodb/specifications/commit/81a39b97c8bdf97a5835a4922e86bef20699ef6c
- [x] Wait for spec PR to be merged (https://github.com/mongodb/specifications/pull/1103)
- [x] Merge libmongocrypt PR (https://github.com/mongodb/libmongocrypt/pull/220)
- [x] Move helper python script to drivers-evergreen-tools (https://github.com/mongodb-labs/drivers-evergreen-tools/pull/181)


src/cmap/connect.ts
```diff
@@ -381,10 +415,81 @@ function makeConnection(options: ConnectionOptions, _callback: CallbackWithType<
...
+}
+
+function makeSocks5Connection(
+  options: MakeConnectionOptions & { proxyOptions: undefined },
+  proxyOptions: ProxyOptions,
+  callback: Callback<Stream>
+) {
+  const proxyHost = proxyOptions.host.includes(':') ? `[${proxyOptions.host}]` : proxyOptions.host;
+  const proxyPort = proxyOptions.port ?? 1080;
+  const hostAddress = HostAddress.fromString(`${proxyHost}:${proxyPort}`);
```
nbbeeken, line 432: Might make sense to move this logic to a helper on HostAddress, either a bespoke one or maybe add flag to fromString to do this `.includes(':')` logic.

src/cmap/connect.ts
```diff
@@ -381,10 +415,81 @@ function makeConnection(options: ConnectionOptions, _callback: CallbackWithType<
...
+          existing_socket: rawSocket,
+          timeout: options.connectTimeoutMS,
+          command: 'connect',
+          destination: {
+            host: destination.host,
+            port: destination.port
+          },
+          proxy: {
+            // host and port are ignored because we pass existing_socket
+            host: 'localhost',
```
nbbeeken, line 465: Weird that we have to defined this, but an opportunity to put any random string here to make it clear its ignored like `'iLoveJavaScript'`! 
We can totally leave this alone.. but why not ship an easter egg? :)

.evergreen/run-socks5-tests.sh
```diff
@@ -0,0 +1,42 @@
...
+PID=$!
+env SOCKS5_CONFIG='["localhost",1080,"username","p4ssw0rd"]' npm run check:socks5
+kill $PID
+
+# Second, test with Socks5 + no authentication
+"$PYTHON_BINARY" "$SOCKS5_SERVER_SCRIPT" --port 1081 --map "localhost:12345 to $FIRST_HOST" &
+PID=$!
+env SOCKS5_CONFIG='["localhost",1081]' npm run check:socks5
+kill $PID
+
```
nbbeeken, line 40: I think we could add `npx mocha --file test/tools/runner/index.js test/functional/client_side_encryption` here and check that FLE is working over the proxy. I think you'll need to pull in a few more variables in the EVG config, you can take a look at the top of run-custom-csfle script.

We could also totally make this follow up work, would it be possible to put together a `MONGODB_URI` variable that would just "work" with the existing tests? (They shouldn't have any special requirements for the URI)

nbbeeken (changes requested):
Got some initial feedback for ya, its looking good, mostly nits. Also, feel free to rebase to pull in the test fix on main so we can get the ✅  I don't mind the files viewed getting messed up for that.

src/cmap/connect.ts
addaleax: @nbbeeken `HostAddress` is public API – is this okay? If so, does `HostAddress.fromHostPort(host, port)` sound good?

src/deps.ts
```diff
@@ -262,6 +262,12 @@ export interface AutoEncryptionOptions {
     /** Command line arguments to use when auto-spawning a mongocryptd */
     mongocryptdSpawnArgs?: string[];
   };
+  proxyOptions?: {
```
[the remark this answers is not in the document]
addaleax: Done!

.evergreen/socks5srv.py
```diff
@@ -0,0 +1,237 @@
...
+import argparse
+
+# Usage: python3 socks5srv.py --port port [--auth username:password] [--map 'host:port to host:port' ...]
+# TODO: Move this script into the shared drivers-evergreen-tools repository
+
+class AddressRemapper:
+  """A helper for remapping (host, port) tuples to new (host, port) tuples
+
+  This is useful for Socks5 servers used in testing environments,
+  because the succesful use of the Socks5 proxy can be demonstrated
```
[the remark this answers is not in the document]
addaleax: Done!

src/cmap/connect.ts
```diff
@@ -331,12 +350,27 @@ function makeConnection(options: ConnectionOptions, _callback: CallbackWithType<
     _callback(err, ret);
   };
 
+  if (proxyOptions?.host != null) {
+    // Currently, only Socks5 is supported.
+    return makeSocks5Connection(
+      {
+        ...options,
+        connectTimeoutMS: connectionTimeout, // Should always be present for Socks5
```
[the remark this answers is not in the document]
addaleax: Done, spelled it `connectTimeoutMS` consistently across the file :+1:

src/cmap/connect.ts
```diff
@@ -331,12 +350,27 @@ function makeConnection(options: ConnectionOptions, _callback: CallbackWithType<
...
+    );
+  }
+
   if (useTLS) {
     const tlsSocket = tls.connect(parseSslOptions(options));
     if (typeof tlsSocket.disableRenegotiation === 'function') {
       tlsSocket.disableRenegotiation();
     }
     socket = tlsSocket;
+  } else if (existingSocket) {
```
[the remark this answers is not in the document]
addaleax: That wouldn’t quite work, because we also use `existingSocket` in the TLS case, it just isn’t quite obvious because that happens inside of `parseSslOptions` – added a comment about that :)

.evergreen/run-socks5-tests.sh
addaleax: Yeah, that’s a great idea. I’ve added an invocation of `bash "${PROJECT_DIRECTORY}/.evergreen/run-custom-csfle-tests.sh"` here and modified the test runner so that it forwards proxy options in its configuration where appropriate, please take a look and let me know what I messed up :slightly_smiling_face:

.evergreen/run-custom-csfle-tests.sh
```diff
@@ -34,10 +36,12 @@ pushd ../csfle-deps-tmp
...
 rm -rf libmongocrypt mongo-c-driver
 
-git clone https://github.com/mongodb/libmongocrypt.git
+#git clone https://github.com/mongodb/libmongocrypt.git
+git clone https://github.com/mongodb-js/libmongocrypt.git
 pushd libmongocrypt
 git fetch --tags
-git checkout "$CSFLE_GIT_REF" -b csfle-custom
+#git checkout "$CSFLE_GIT_REF" -b csfle-custom
+git checkout "ea3f6252b55cf235e8adf3a04930479d6e5ed11f" -b csfle-custom
```
addaleax, line 44: I mentioned it in the PR description, but I’m also leaving a comment here just so I don’t forget about it – https://github.com/mongodb/libmongocrypt/pull/220 should probably be merged before this so that this “just works” afterwards

.evergreen/run-socks5-tests.sh
addaleax: @nbbeeken Actually, it looks like some more options forwarding is needed in order for the TLS version of these tests to pass … is that worth it or should the CSFLE tests only be run when the server does not require TLS? I’m happy to go either way.

src/cmap/connect.ts
nbbeeken: HostAddress is public type API since its accessible on some events I think but its not exported for use at runtime so changing existing behavior is free game, so I think enhancing fromString would work. but fromHostPort sounds good too!

.evergreen/run-socks5-tests.sh
nbbeeken: I think you can gate it on running only when TLS is disabled. The important part to me of checking FLE over the proxy is that when we create an internal separate client, that it is handed the same config options and works as expected. The FLE deadlock tests will set up and run the scenario where a separate MongoClient instance is needed, and it _should_ inherit everything but that's what I want to see pass.

src/cmap/connect.ts
addaleax: Sure, done :) I do think it should be `HostAddress.fromHostPort(host, port)` since doing something like `HostAddress.fromString(`${host}:${port}`, true /* lastColonSeparatesPortFromHost */)` feels like it’s a bit more convoluted than it needs to be.

`HostAddress.fromSrvRecord()` was something that was added in the meantime as well. I’ve made it call `HostAddress.fromHostPort()`, because that felt a bit cleaner – I don’t think it makes much of a difference because we require SRV records to point to hostnames.

.evergreen/run-socks5-tests.sh
addaleax: @nbbeeken Sure, done – tests pass :) I noticed that the xunit.xml files from different mocha runs overwrite each other, not sure what to do about that but it also feels like more of a cosmetic issue (unless you feel strongly about it).

.evergreen/socks5srv.py
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env python3
```
addaleax, line 1: Also leaving a comment for context, the PR to add this to drivers-evergreen-tools is in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/181

.evergreen/run-socks5-tests.sh
addaleax: Actually, it doesn’t pass on Windows either: 

```
[2021/12/03 11:09:25.960] + '[' nossl == nossl ']'
[2021/12/03 11:09:25.960] + env 'MONGODB_URI=mongodb://localhost:12345/?proxyHost=localhost&proxyUsername=username&proxyPassword=p4ssw0rd' bash /cygdrive/c/data/mci/72615ac373d7b45a5862ded1dd3d86af/src/.evergreen/run-custom-csfle-tests.sh
[2021/12/03 11:09:26.009] + set -o errexit
[2021/12/03 11:09:26.009] + echo 'adding temporary AWS credentials to environment'
[2021/12/03 11:09:26.009] + . Z:/data/mci/72615ac373d7b45a5862ded1dd3d86af/drivers-tools/.evergreen/csfle/set-temp-creds.sh
[2021/12/03 11:09:26.069] ++ set +o xtrace
[2021/12/03 11:09:26.069] Traceback (most recent call last):
[2021/12/03 11:09:26.069]   File "<stdin>", line 1, in <module>
[2021/12/03 11:09:26.069] ImportError: No module named boto3
[2021/12/03 11:26:37.398] Command stopped early: context canceled
```

Not sure what’s up with that, but I assume it’s fine because right now the csfle tests only run on Ubuntu to begin with?

nbbeeken: Yea we've decided its enough coverage to just try FLE on ubuntu for now. So we could gate running the fle script to only linux. And yea its fine that the xunit gets over written, we'll still see failures so I agree mostly cosmetic.

.evergreen/run-custom-csfle-tests.sh
```diff
@@ -34,10 +36,12 @@ pushd ../csfle-deps-tmp
...
 rm -rf libmongocrypt mongo-c-driver
 
-git clone https://github.com/mongodb/libmongocrypt.git
+#git clone https://github.com/mongodb/libmongocrypt.git
+git clone https://github.com/mongodb-js/libmongocrypt.git
 pushd libmongocrypt
 git fetch --tags
-git checkout "$CSFLE_GIT_REF" -b csfle-custom
+#git checkout "$CSFLE_GIT_REF" -b csfle-custom
+git checkout "7896064885e05ad494362b383cbbb9eade98c7ae" -b csfle-custom
```
nbbeeken, line 44: We can reset this now we've merged the FLE bit

.evergreen/config.yml
```diff
@@ -293,6 +293,54 @@ functions:
...
+          export AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}"
+          export AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}"
+          export CSFLE_GIT_REF="${CSFLE_GIT_REF}"
+          export CDRIVER_GIT_REF="${CDRIVER_GIT_REF}"
+          EOT
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: src
+        script: >
```
nbbeeken, line 316: TIL about this character "folding newlines", so when parsed this string will won't have the empty lines? 😎

test/manual/socks5.test.js
```diff
@@ -0,0 +1,325 @@
+'use strict';
+const { MongoClient } = require('../../src');
```
nbbeeken, line 2: We've been moving tests to TS files, I tried it locally, and it looks like you would just have to change these to es imports if you wanna make that improvement.

test/spec/uri-options/proxy-options.json
```diff
@@ -0,0 +1,139 @@
+{
```
nbbeeken, line 1: Unfortunately our uri-options spec runner is a little under the weather at the moment (https://jira.mongodb.org/browse/NODE-3738). It might be worth taking a look at `test/integration/uri-options/uri_options.spec.test.js` to see if these tests are falling under the skip condition. I'm taking a look too, just waiting on the current run to get there.

If it is skipped we might be okay with tolerating that, but you can see in that same file I broke out tests for the recent new srv options.

src/mongo_client.ts
```diff
@@ -248,6 +248,8 @@ export interface MongoClientOptions extends BSONSerializeOptions, SupportedNodeC
   autoEncryption?: AutoEncryptionOptions;
   /** Allows a wrapping driver to amend the client metadata generated by the driver to include information about the wrapping driver */
   driverInfo?: DriverInfo;
+  /** Configures a Socks5 proxy used for creating TCP connections. */
+  proxyOptions?: ProxyOptions;
```
nbbeeken, line 252: I think instead of a sub object we should have each `proxy*` option defined directly on the MongoClientOptions interface, since that's how the uri options work. IIUC

src/connection_string.ts
```diff
@@ -860,6 +878,33 @@ export const OPTIONS = {
...
+  } as OptionDescriptor,
+  proxyOptions: {
+    type: 'record'
+  },
+  proxyPassword: {
+    target: 'proxyOptions',
+    transform({ values: [value], options }): ProxyOptions {
+      return { host: '', ...options.proxyOptions, password: String(value) || undefined };
+    }
+  } as OptionDescriptor,
```
nbbeeken, line 895: If we do flatten these options on to the MongoClientOptions interface we shouldn't need to cast here anymore

test/spec/uri-options/proxy-options.json
addaleax: Yeah, some of the tests are being skipped here. Is it worth special-casing these tests? My understanding is that they would pass after NODE-3738.

Just to be sure – is it actually correct that the invalid connection strings have `warning: false` and `valid: false` both set? If I remove the `warning: false`, then all tests just pass here. (I based this on other spec tests which also have both set for invalid connection strings, but now I’m not sure if this was actually something I should have done or not.)

nbbeeken: I believe this is okay just because invalid likely takes precedence, if we're throwing then that's our warning, so we're not checking for both. Our driver started by implementing these spec tests such that connection strings that should only warn were actually throwing errors and that was considered the same. We'll try to correct this to what ever extent isn't breaking when we get to that ticket.

So good to know they pass, I think you're right we don't need to special case here then. Thanks for checking!

src/mongo_client.ts
nbbeeken: ty!
@durran Just want to you to catch this thread as to why we're making the changes over in libmongocrypt

nbbeeken (approved):
LGTM thanks for all the effort on this!!

durran (approved):

PR #3044, opened 2021-11-18
fix: remove WithId from find filter type

avaly:
#### What is changing?

Fixes a regression in the `find()` `filter` TypeScript type.

No.

I believe there was a bug introduced in the #3039 PR: https://github.com/mongodb/node-mongodb-native/pull/3039/files#r752006554

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -730,9 +730,9 @@ export class Collection<TSchema extends Document = Document> {
    * @param filter - The filter predicate. If unspecified, then all documents in the collection will match the predicate
    */
   find(): FindCursor<WithId<TSchema>>;
-  find(filter: Filter<WithId<TSchema>>, options?: FindOptions): FindCursor<WithId<TSchema>>;
-  find<T>(filter: Filter<WithId<TSchema>>, options?: FindOptions): FindCursor<T>;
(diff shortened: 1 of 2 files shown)


PaulEndri:
`db.collection.find({_id: 'something'})` is a valid find operation in mongo itself and it's why it was added to the filter definition. This probably needs to get fixed by making whatever downstream incompatibility was accidentally caused by this change, compatible with the new definition.

dariakp:
@avaly Thanks for catching the inconsistency! I think the correct approach would be to add the `WithId<TSchema>` handling to the Filter type definition itself, so that it automatically propagates to the operations that should support it. As @PaulEndri mentioned, the `_id` is a valid filter option regardless of whether you explicitly specify it in your schema, since you can search on it even if it's auto-generated by the server. I think we didn't catch the omission earlier because we currently have the `TSchema` permitting all string keys, but it would be more correct to be explicit about the `_id` type (via `WithId`) than just letting it fall through. Do you have any concerns about taking that path forward?

avaly:
@dariakp @PaulEndri you both make sense! I agree that maybe this should be solved inside of the `Filter` type.

Feel free to close this PR.

dariakp:
@avaly Alright, I created [NODE-3770](https://jira.mongodb.org/browse/NODE-3770) to track the work to add consistent handling, closing this PR.

PR #3045, opened 2021-11-18
fix(NODE-3711): retry txn end on retryable write

durran:
### Description

The driver was retrying `commitTransaction` and `abortTransaction` but previously failing 4 spec tests that were tricking the driver to base retries on these commands on error code. The retryable writes spec defines the specific errors that are retryable, and the transactions spec states these 2 commands must adhere to that specification. The fact that an operation is retryable is stated as:

_When connected to a MongoDB instance that supports retryable writes (versions 3.6+), the driver MUST treat all errors with the RetryableWriteError label as retryable._

Note that the ticket was scoped to these two commands only, so the full gambit of retryable writes was not touched. There is a ticket to address that in NODE-3769.

#### What is changing?

This adds a new check for for if a transaction command can be retried, `isRetryableEndTransactionError`, which bases this solely on if the error contains a "RetryableWriteError" and does not look at the error code.

No

NODE-3711

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -697,6 +697,10 @@ const RETRYABLE_WRITE_ERROR_CODES = new Set<number>([
   MONGODB_ERROR_CODES.ExceededTimeLimit
 ]);
 
+export function isRetryableEndTransactionError(error: MongoError): boolean {
+  return error.hasErrorLabel('RetryableWriteError');
+}
+
 export function isRetryableWriteError(error: MongoError): boolean {
   if (error instanceof MongoWriteConcernError) {
     return RETRYABLE_WRITE_ERROR_CODES.has(error.result?.code ?? error.code ?? 0);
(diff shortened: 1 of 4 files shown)


src/error.ts
```diff
@@ -697,6 +697,10 @@ const RETRYABLE_WRITE_ERROR_CODES = new Set<number>([
   MONGODB_ERROR_CODES.ExceededTimeLimit
 ]);
 
+export function isRetryableEndTransactionError(error: MongoError): boolean {
```
dariakp, line 700: do we expect to rename this function when we implement the next ticket or add a new function that returns the same thing to address the missing cases? i.e., will this become the new `isRetryableError` check?

dariakp (commented):
Just have a question to clarify the intent of the implementation, otherwise LGTM

src/error.ts
durran: I believe this will go away as you mention and become `isRetryable` error, barring no other complications.

nbbeeken (approved):

PR #3054, opened 2021-11-23
fix(NODE-3627): Enable flexible BSON validation for server error key containing invalid utf-8

gjchong25:
### Description

This change allows for validation toggling on the `writeErrors` key, which can contain a server error message which has invalid utf-8, thus causing the user to get an invalid utf-8 error rather than the corresponding server error. Unit tests using a real server error response are added to make sure the invalid utf-8 error is not being thrown when validation on the `writeErrors` key is turned off.


src/cmap/commands.ts
```diff
@@ -837,13 +838,18 @@ export class BinMsg {
     const promoteValues = options.promoteValues ?? this.opts.promoteValues;
     const promoteBuffers = options.promoteBuffers ?? this.opts.promoteBuffers;
     const bsonRegExp = options.bsonRegExp ?? this.opts.bsonRegExp;
+    if (options.validation && Object.keys(options.validation.utf8)[0] !== 'writeErrors') {
+      throw new Error('Can only toggle validation settings for writeErrors key');
```
durran, line 842: I think we should probably use one of our custom error classes here, like `MongoInvalidArgumentError` in this case.

test/unit/commands.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      msgHeader,
+      msgBodyInvalidUtf8WriteErrors
+    );
+    expect(() => binMsgInvalidUtf8ErrorMsg.parse({ validation: { utf8: { n: false } } })).to.throw(
+      Error,
+      'Can only toggle validation settings for writeErrors key'
+    );
+  });
+
+  it('should not throw invalid utf8 error when validation disabled for writeErrors', function () {
```
durran, line 86: I generally prefer the approach to writing tests thinking of tests as documentation, and the test output should read in a manner that is clear to someone just reading it without looking at the code itself. With this in mind it's always good to put one assertion per `it` block when possible, wrapped in a `context`. This produces some really easy to read output for each exact assertion... So an example for this would to change it to:

```javascript
context('when validation is disabled for writeErrors', function () {
  const binMsgInvalidUtf8ErrorMsg = new BinMsg(
    Buffer.alloc(0),
    msgHeader,
    msgBodyInvalidUtf8WriteErrors
  );
  const options = { validation: { utf8: { writeErrors: false } } };

  it('does not raise an error', function () {
    expect(() =>
      binMsgInvalidUtf8ErrorMsg.parse(options)
    ).to.not.throw();
  });

  it('contains the invalid utf8 in the writeError', function () {
    expect(
      BSON.deserialize(badutf8inputtodeserialize, options)
    ).to.deep.equals(invalidUtf8InWriteErrorsJSON);
  });
});
```

durran: What do you think about changing the tests with multiple assertions to follow this format?

test/unit/commands.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  'hex'
+);
+
+const badutf8inputtodeserialize = Buffer.from(
```
durran, line 19: nit: change variable name to camel case.

durran (changes requested):

durran (dismissed):
Nice work, thanks!

src/cmap/commands.ts
```diff
@@ -837,6 +838,12 @@ export class BinMsg {
     const promoteValues = options.promoteValues ?? this.opts.promoteValues;
     const promoteBuffers = options.promoteBuffers ?? this.opts.promoteBuffers;
     const bsonRegExp = options.bsonRegExp ?? this.opts.bsonRegExp;
+    if (options.validation && Object.keys(options.validation.utf8)[0] !== 'writeErrors') {
```
baileympearson, line 841: Are we not allowing general validation options at this time?  The type on line 472 implies that ` { validation: true / false } ` would be allowed, but this logic would throw an error if that was passed in.

gjchong25: This is an internal workaround to deal with server errors containing invalid utf8, so we're not really allowing keys besides writeErrors to have validation enabled/disabled. But you're right, I should narrow this down more (but I'm getting type errors if I try to change the type from `Record<string, true> | Record<string, false>` to `Record<string, boolean>`, since its saying the type isn't assignable to type 'DeserializeOptions' even though I omitted validation from the BSONSerializeOptions type?).

baileympearson: Oh ok - so this is an internal change and not a user facing interface?  And it's building off of your work in the BSON library?  

As long as it's not user-facing, I would probably type it the way you had it originally so that it matches the types in the bson repo [here](https://github.com/mongodb/js-bson/blob/522182577733a3b7a4d70d3c509cc7e6621d0470/src/parser/deserializer.ts#L63).  But I'm also still new so I'm interested to hear what the team thinks

dariakp: @baileympearson good points here, let me try to address some of them: this validation and the things inside the class are 100% internal, we could probably make do without that validation at all, given that it's internally driven at the moment; however, we have a follow-up ticket to allow users to disable validation entirely, so I think this was Grace trying to set up the stage for the follow up work. That said, I don't think we'd want to try to validate this here, instead, we'd probably validate the user options in `resolveBSONOptions`. @gjchong25 I think this is safe to leave for the follow up work. Can we remove the validation check and the passing in of the options? Just hardcode the `const validation = { utf8: { writeErrors: false } }` for now?

src/cmap/commands.ts
```diff
@@ -469,6 +469,7 @@ export interface MessageHeader {
 export interface OpResponseOptions extends BSONSerializeOptions {
   raw?: boolean;
   documentsReturnedIn?: string | null;
+  validation?: { utf8: Record<string, true> | Record<string, false> };
```
dariakp, line 472: we can remove this type if we keep the validation hardcoded for this ticket

src/cmap/commands.ts
```diff
@@ -851,8 +858,9 @@ export class BinMsg {
       if (payloadType === 0) {
         const bsonSize = this.data.readUInt32LE(this.index);
         const bin = this.data.slice(this.index, this.index + bsonSize);
-        this.documents.push(raw ? bin : BSON.deserialize(bin, _options));
-
+        this.documents.push(
+          raw ? bin : BSON.deserialize(bin, Object.assign({ validation }, _options))
```
dariakp, line 862: instead of doing this in two places - here and in L877, can we instead add it to the BSONSerializeOptions `_options` variable above? And to avoid touching the `BSONSerializeOptions` type, you can probably do `_options: BSONSerializeOptions & { validation: <type desrciption> }`

src/cmap/commands.ts
```diff
@@ -837,6 +838,12 @@ export class BinMsg {
     const promoteValues = options.promoteValues ?? this.opts.promoteValues;
     const promoteBuffers = options.promoteBuffers ?? this.opts.promoteBuffers;
     const bsonRegExp = options.bsonRegExp ?? this.opts.bsonRegExp;
```
dariakp, line 840: @durran @nbbeeken Just to double check, are we concerned at all with the `parse` method in the Response class or just the BinMsg? When do we hit the Response logic?

dariakp (changes requested):

src/cmap/commands.ts
nbbeeken: We don't! Fortunately, we dropped the server versions that do not use OP_MSG so this condition in message_stream.ts:
```ts
ResponseType = messageHeader.opCode === OP_MSG ? BinMsg : Response;
```
always returns BinMsg, we could add the logic for completeness but that section of code is destined for deletion.

src/cmap/commands.ts
nbbeeken: I dropped the throw / and checks

src/cmap/commands.ts
nbbeeken: done!

dariakp (approved):
Thanks @gjchong25 and @nbbeeken both!

PR #3055, opened 2021-11-26
fix(NODE-3767): don't delete dbName if authSource is provided

avlan:
#### What is changing?
`dbName` option is not going to be deleted when `authSource` is present in the connection string.

I don't think there is unless this is intentional behaviour.

I have noticed that the `dbName` option is deleted when `authSource` is being passed in the connection string since version **4.2.0** (after [this commit](https://github.com/mongodb/node-mongodb-native/pull/3031#discussion_r748391108)) but I couldn't find any documentation for this neither the [MongoDB documentation](https://docs.mongodb.com/manual/reference/connection-string/#mongodb-urioption-urioption.authSource) nor in the [MongoDB driver documentation](https://mongodb.github.io/node-mongodb-native/4.2/classes/MongoClient.html#connect). I thought that this might be a bug since according to those documentations mentioned above, there's no colliding between `dbName` and `authSource` and currently both options are allowed.

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -291,11 +291,6 @@ export function parseOptions(
     }
   }
 
-  if (urlOptions.has('authSource')) {
-    // If authSource is an explicit key in the urlOptions we need to remove the dbName
-    urlOptions.delete('dbName');
-  }
-
   const objectOptions = new CaseInsensitiveMap(
     Object.entries(options).filter(([, v]) => v != null)
   );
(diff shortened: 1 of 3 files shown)


lmachens:
Good job, I faced this issue too when upgrading to 4.2.0.

nbbeeken (dismissed):
Thanks so much for the help, I brought over some additional test cases from a more or less duplicate PR so we can get the best of both merged in together.

PR #3058, opened 2021-11-29
fix(NODE-2370): correct a return type of hasNext()

wegylexy:
### Description

`hasNext()` returns `Promise<boolean>` instead of `Promise<void>`. The type definition is wrong.

This initial pull request fixes one instance. There are more.

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -310,7 +310,7 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
   }
 
   /** Check if there is any document still available in the Change Stream */
-  hasNext(callback?: Callback): Promise<void> | void {
+  hasNext(callback?: Callback): Promise<boolean> | void {
     setIsIterator(this);
     return maybePromise(callback, cb => {
       getCursor(this, (err, cursor) => {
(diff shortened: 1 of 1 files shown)


nbbeeken (changes requested):
Thanks for the help, always appreciated! I have a small suggestion on how we can make the typing a bit more useful.

src/change_stream.ts
```diff
@@ -310,7 +310,7 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
   }
 
   /** Check if there is any document still available in the Change Stream */
-  hasNext(callback?: Callback): Promise<void> | void {
+  hasNext(callback?: Callback): Promise<boolean> | void {
```
nbbeeken, line 313:
```suggestion
  hasNext(): Promise<boolean>;
  hasNext(callback: Callback<boolean>): void;
  hasNext(callback?: Callback): Promise<boolean> | void {
```

I think this would increase usability even more, dividing up the return type depending on whether or not a callback was provided.

nbbeeken (changes requested):

wegylexy:
If somebody has time, please fix other hasNext() too.

nbbeeken (approved):
Look like the hasNext on cursor is already in the correct format, so we're almost good to go here. Thanks again!

dariakp (approved):
LGTM

PR #3060, opened 2021-11-30
fix(NODE-3705): ReadPreference.fromOptions omitting hedge and maxStalenessSeconds when readPreference is a string

baileympearson:
### Description
Prior to these changes, when `ReadPreference.fromOptions` was passed a `string` for the read preference, hedge configuration and maxStalenessSeconds were not set on the `ReadPreference` object.  This PR passes hedge configuration and maxStalenessSeconds configuration through to the read preference object when it is created using `ReadPreference.fromOptions` with a string as the read preference.

Additionally, this PR moves unit tests for the ReadPreference class into the unit test suite from the functional test suite.


src/read_preference.ts
```diff
@@ -156,7 +156,9 @@ export class ReadPreference {
     }
 
     if (typeof readPreference === 'string') {
-      return new ReadPreference(readPreference as ReadPreferenceMode, readPreferenceTags);
+      return new ReadPreference(readPreference as ReadPreferenceMode, readPreferenceTags, {
+        hedge: options.hedge
```
nbbeeken, line 160: It wasn't mentioned in this ticket but I would imagine we also have a gap here if options has a `maxStalenessSeconds` defined on it. It's used in the other contstruction on L165, so I think it should be here too.

test/functional/readpreference.test.js
```diff
@@ -78,6 +78,50 @@ describe('ReadPreference', function () {
...
+      });
+
+      it('should accept { readPreference, readPreferenceTags }', function () {
+        const readPreference = ReadPreference.fromOptions({
+          readPreference: SECONDARY,
+          readPreferenceTags: TAGS
+        });
+        expect(readPreference).to.be.an.instanceOf(ReadPreference);
+        expect(readPreference.mode).to.equal(SECONDARY);
+        expect(readPreference.tags).to.eql(TAGS);
```
nbbeeken, line 106: Let's make the assertions consistent, with a preference for `equal`
In fact, its more of a stylistic opinion but I usually go with property checks:
```js
expect(readPreference).to.have.property('mode', SECONDARY);
```
this way if it fails it prints out a diff of the whole object. But sticking with `equal` is ok too!

edit:
Ok TIL that `eql` is deep equal :) facinating I made it this far without knowing, we've just used `.to.be.deep.equal(xx)`. I don't have a strong opinion on using one or the other.

test/functional/readpreference.test.js
```diff
@@ -78,6 +78,50 @@ describe('ReadPreference', function () {
...
+          hedge: {
+            enabled: true
+          }
+        });
+        expect(readPreference).to.be.an.instanceOf(ReadPreference);
+        expect(readPreference.mode).to.equal(SECONDARY);
+        expect(readPreference.tags).to.eql(TAGS);
+        expect(readPreference.hedge).to.eql({ enabled: true });
+      });
+    });
```
nbbeeken, line 122: It is possible that `primary` and `hedge` options were being combined and since hedge was ignored there was no issue. Now its possible to get a thrown error from the constructor: `'Primary read preference cannot be combined with hedge'`. Can we add a test expecting that thrown error.

Also if you make the other change with the maxStalenessSeconds, we would want a check for that thrown error as well.

nbbeeken (changes requested):
Nice! this looks close to ready

src/read_preference.ts
baileympearson: Yup, I forgot to mention that in the description.  I wasn't sure if I should include that as a part of my PR but that's an easy change to make as well, so I'll add it!

test/functional/readpreference.test.js
```diff
@@ -78,6 +78,50 @@ describe('ReadPreference', function () {
     });
   });
 
+  describe('fromOptions factory method', () => {
```
nbbeeken, line 81: Some meta test organizing, we are gradually improving the placement of tests so I think it would make sense to take this block and the `::constructor` one out of here and move them to a new file: `test/unit/read_preference.test.ts`. They don't depend on being connected to a MongoDB server like the other tests.

test/functional/readpreference.test.js
baileympearson: I wasn't sure how thoroughly to test the factory method.  We have a comprehensive suite of tests for the constructor above the tests I added, and specifically on line 32 we do check for the `'Primary read preference cannot be combined with hedge'` error.  I can add more thorough testing around valid/invalid combinations of options, but I wasn't sure if that was necessary since we basically just reformat the options passed in and defer to the constructor (which is already tested).

nbbeeken: Yea its a bit of duplication but since now this method throws it where it didn't before I don't want the behavior to go unnoticed so to speak. If we test for it, future changes to the method will be sure to either maintain the behavior or require a intentional choice to modify a test.

I don't think we need any more combinations beyond:
```
expect(() => ReadPreference.fromOptions({ readPreference: 'primary', hedge: { enabled: true } }).to.throw('...')
```

Also [L32](https://github.com/mongodb/node-mongodb-native/blob/NODE-3705/enable-hedge-options-when-readPreference-is-string/test/functional/readpreference.test.js#L32) is a test header for me, I'm actually not seeing a test case for hedge and primary in the constructor section either 🤔

test/functional/readpreference.test.js
baileympearson: Done!

test/functional/readpreference.test.js
baileympearson: I think you were right, I didn't notice that test.  I added this particular test scenario to both the constructor tests and the factory method tests.

test/unit/read_preference.test.js
```diff
@@ -0,0 +1,138 @@
+'use strict';
+
+const { ReadPreference } = require('../../src');
+const chai = require('chai');
+chai.use(require('chai-subset'));
+const expect = chai.expect;
```
nbbeeken, line 6:
```suggestion
import { ReadPreference } from '../../src';
import { expect } from 'chai';
```

If it doesn't give you too much trouble we can use TS here, just more drive by improvements.

test/unit/read_preference.test.js
```diff
@@ -0,0 +1,138 @@
...
+
+describe('ReadPreference', function () {
```
nbbeeken, line 8:
```suggestion
describe('class ReadPreference', function () {
```
If you use the mocha explorer then you may notice that the buttons appear on the import at the top of the file instead of here, I have worked around this by changing up the naming a bit. Take a look at `test/unit/utils.test.ts` for some examples. It's a silly-ish work around but I think its worth it for the developer UX 😄

nbbeeken (changes requested):
Just some more small improvements, but I'll open this up to the team now

test/unit/read_preference.test.js
baileympearson: Done!

nbbeeken (approved):

dariakp (approved):

PR #3061, opened 2021-12-01
feat(NODE-3784):  Add `enableUtf8Validation` option

baileympearson:
#### What is changing?

A new option was added to the db, client, operation and collection classes to allow users to enable/disable utf8 validation.

The legacy response class is not modified - deserialization changes are only present in the `BinMsg` class.

Yes

UTF8 validation is now enabled by default.  Disabling utf8 validation can result in a performance benefit for users.


src/cmap/commands.ts
```diff
@@ -876,4 +876,15 @@ export class BinMsg {
 
     this.parsed = true;
   }
+
+  private _parseBsonSerializationOptions({ enableUtf8Validation }: BSONSerializeOptions): {
+    utf8: { writeErrors: false } | false;
+  } {
+    // eslint-disable-next-line no-restricted-syntax
+    if (enableUtf8Validation === undefined || enableUtf8Validation) {
```
baileympearson, line 884: I _think_ this should never be undefined, because I set the default value to `true` in both `connection_string.ts` and `resolveBSONOptions`.  But the interface does allow for an undefined, so I decided to handle it.  Checking for `undefined` results in a linting error so I have to use an `eslint-disable`

src/cmap/commands.ts
```diff
@@ -876,4 +875,14 @@ export class BinMsg {
 
     this.parsed = true;
   }
+
+  private _parseBsonSerializationOptions({ enableUtf8Validation }: BSONSerializeOptions): {
```
nbbeeken, line 879:
```suggestion
  parseBsonSerializationOptions({ enableUtf8Validation }: BSONSerializeOptions): {
```

Since we're working on an internal class here, everything's inherently private, its stylistic but the access modifier likely would just get in the way of unit tests written in typescript if that ever came about.

src/cmap/commands.ts
```diff
@@ -856,7 +855,7 @@ export class BinMsg {
       if (payloadType === 0) {
         const bsonSize = this.data.readUInt32LE(this.index);
         const bin = this.data.slice(this.index, this.index + bsonSize);
-        this.documents.push(raw ? bin : BSON.deserialize(bin, bsonOptions));
+        this.documents.push(raw ? bin : deserialize(bin, bsonOptions));
```
nbbeeken, line 858: Is this change (`BSON.des` vs just `des`) something to make sinon work? I think that's fine let's just make it the same for all usages and serialize usages.

src/cmap/commands.ts
```diff
@@ -6,6 +6,7 @@ import type { Long, Document, BSONSerializeOptions } from '../bson';
 import type { ClientSession } from '../sessions';
 import type { CommandOptions } from './connection';
 import { MongoRuntimeError, MongoInvalidArgumentError } from '../error';
+import { deserialize } from 'bson';
```
nbbeeken, line 9: You have to import BSON things from `src/bson.ts` to continue supporting our optional dependency logic.

test/functional/cmap/commands.test.ts
```diff
@@ -0,0 +1,122 @@
...
+            { name: 'John Doe' },
+            passOptionTo === 'operation' ? option : {}
+          );
+
+          expect(deserializeSpy.called).to.be.true;
+          const validationArgument = deserializeSpy.lastCall.lastArg.validation;
+          expect(validationArgument).to.deep.equal(EXPECTED_VALIDATION_DISABLED_ARGUMENT);
+        } finally {
+          await client.close();
+        }
```
nbbeeken, line 51: Nothing to change here, just calling out that normally we would like to use  an `afterEach` hook to ensure the client closes when each test finishes, unfortunately our session leak checker is currently implemented as an afterEach hook that runs before the one you would define here. So your hook wouldn't get the chance to clean up the sessions/client before the leak checker would throw.

nbbeeken (changes requested):

src/cmap/commands.ts
baileympearson: Fixed!

src/cmap/commands.ts
baileympearson: ok yeah good catch.  This was to make Sinon work, but I had to do this because I wasn't importing from `src/bson`.

src/cmap/commands.ts
```diff
@@ -6,6 +6,7 @@ import type { Long, Document, BSONSerializeOptions } from '../bson';
 import type { ClientSession } from '../sessions';
 import type { CommandOptions } from './connection';
 import { MongoRuntimeError, MongoInvalidArgumentError } from '../error';
+// import { deserialize } from '../bson';
```
durran, line 9: I'd say after the revert we can just remove this now.

durran (changes requested):

src/cmap/commands.ts
baileympearson: Oh, good catch!

src/cmap/commands.ts
baileympearson: Yup, this was just to sinon work.  But importing from `src/bson` instead of `'bson'` fixed the issue that I was seeing earlier, and I can revert it.

nbbeeken (approved):

src/bson.ts
```diff
@@ -55,6 +55,9 @@ export interface BSONSerializeOptions
     > {
   /** Return BSON filled buffers from operations */
   raw?: boolean;
+
+  /** Enable utf8 validation when deserializing BSON documents.  Defaults to true. */
+  enableUtf8Validation?: boolean;
```
dariakp, line 60: Leaving this comment on the first file for visibility: let's wait to merge this PR until we decide whether or not we're going to do another patch release before the holidays

dariakp (changes requested):

durran (dismissed):

test/unit/commands.test.ts
```diff
@@ -37,45 +38,64 @@ const invalidUtf8InWriteErrorsJSON = {
...
-      errmsg:
-        'E11000 duplicate key error collection: bigdata.test index: text_1 dup key: { text: "d☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃�..." }'
-    }
-  ],
-  ok: 1
-};
 
 describe('BinMsg BSON utf8 validation', () => {
-  context('when validation is disabled for writeErrors', () => {
+  test('invalid utf8 data throws as expected when deserialized', () => {
```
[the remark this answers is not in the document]
baileympearson: This is a poorly written test description, I'll update it (whether or not we keep the test is another question).

Grace and @nbbeeken wrote this test to confirm that the mock data they manually created for these unit tests was correctly malformed (it did contain invalid utf8).  I think Neal wanted to keep this, so I left it.  Do you have a preference whether or not we keep it?

test/unit/commands.test.ts
```diff
@@ -37,45 +38,64 @@ const invalidUtf8InWriteErrorsJSON = {
...
-      errmsg:
-        'E11000 duplicate key error collection: bigdata.test index: text_1 dup key: { text: "d☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃☃�..." }'
-    }
-  ],
-  ok: 1
-};
 
 describe('BinMsg BSON utf8 validation', () => {
-  context('when validation is disabled for writeErrors', () => {
+  test('mock data contains replacement characters for invalid utf8 in writeError object', () => {
```
dariakp, line 44:
```suggestion
  test('bson correctly deserializes data with replacement characters for invalid utf8 in writeErrors object', () => {
  // this is a sanity check to make sure nothing unexpected is happening in the deserialize method itself
```

dariakp (changes requested):
Just a small suggestion for phrasing on the test case

test/unit/commands.test.ts
baileympearson: Done!

dariakp (approved):

durran (approved):

PR #3063, opened 2021-12-02
refactor(NODE-3788): update names of offensive error codes

ljhaywar:
### Description

NODE-3788

#### What is changing?

- Error codes with offensive names are being renamed
- Other offensive wording in error.ts is being renamed

The tests that reference these error codes are spec tests. Updates to those tests are covered in NODE-3791.

No

Following the guidance documented in https://tools.ietf.org/id/draft-knodel-terminology-00.html, we will remove all oppressive and unnecessarily gendered language in driver documentation, code, tests, specs, and spec tests.

- [ ] ~~New TODOs have a related JIRA ticket~~

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -9,9 +9,11 @@ export type AnyError = MongoError | Error;
 const kErrorLabels = Symbol('errorLabels');
 
 /** @internal */
+// Abstract the offensive terminology to a constant
 const LEGACY_NOT_PRIMARY_ERROR_MESSAGE_REGEX = /not master/;
 
 /** @internal */
+// Abstract the offensive terminology to a constant
 const LEGACY_NOT_PRIMARY_OR_SECONDARY_ERROR_MESSAGE_REGEX = /not master or secondary/;
 
 /** @internal MongoDB Error Codes */
--- a/src/error.ts
+++ b/src/error.ts
@@ -8,13 +8,26 @@ export type AnyError = MongoError | Error;
 /** @internal */
 const kErrorLabels = Symbol('errorLabels');
 
-/** @internal */
-// Abstract the offensive terminology to a constant
-const LEGACY_NOT_PRIMARY_ERROR_MESSAGE_REGEX = /not master/;
+/**
+ * @internal
+ * The legacy error message from the server that indicates the node is not a writable primary
+ * https://github.com/mongodb/specifications/blob/b07c26dc40d04ac20349f989db531c9845fdd755/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-writable-primary-and-node-is-recovering
+ */
+export const LEGACY_NOT_WRITABLE_PRIMARY_ERROR_MESSAGE = 'not master';
 
-/** @internal */
-// Abstract the offensive terminology to a constant
-const LEGACY_NOT_PRIMARY_OR_SECONDARY_ERROR_MESSAGE_REGEX = /not master or secondary/;
+/**
+ * @internal
+ * The legacy error message from the server that indicates the node is not a primary or secondary
+ * https://github.com/mongodb/specifications/blob/b07c26dc40d04ac20349f989db531c9845fdd755/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-writable-primary-and-node-is-recovering
+ */
+export const LEGACY_NOT_PRIMARY_OR_SECONDARY_ERROR_MESSAGE = 'not master or secondary';
+
+/**
+ * @internal
+ * The error message from the server that indicates the node is recovering
+ * https://github.com/mongodb/specifications/blob/b07c26dc40d04ac20349f989db531c9845fdd755/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-writable-primary-and-node-is-recovering
+ */
+export const NODE_IS_RECOVERING_ERROR_MESSAGE = 'node is recovering';
 
 /** @internal MongoDB Error Codes */
 export const MONGODB_ERROR_CODES = Object.freeze({
@@ -722,8 +735,8 @@ export function isRetryableError(error: MongoError): boolean {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     (typeof error.code === 'number' && RETRYABLE_ERROR_CODES.has(error.code!)) ||
     error instanceof MongoNetworkError ||
-    !!error.message.match(LEGACY_NOT_PRIMARY_ERROR_MESSAGE_REGEX) ||
-    !!error.message.match(/node is recovering/)
+    !!error.message.match(new RegExp(LEGACY_NOT_WRITABLE_PRIMARY_ERROR_MESSAGE)) ||
+    !!error.message.match(new RegExp(NODE_IS_RECOVERING_ERROR_MESSAGE))
   );
 }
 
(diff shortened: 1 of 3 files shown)


ljhaywar:
Questions:
- I renamed a constant and a function. Neither are exported, so I *think* that means I can make the changes without any deprecation notices. Is that the case?
- I updated the return message for two functions. From what I can tell, those messages are only being used inside of this file. Are there any other places the messages are being used?

baileympearson:
> Questions:
> 
> * I renamed a constant and a function. Neither are exported, so I _think_ that means I can make the changes without any deprecation notices. Is that the case?
> * I updated the return message for two functions. From what I can tell, those messages are only being used inside of this file. Are there any other places the messages are being used? Is it OK to change the error message (is anyone relying on that specific error message)?

- Yup, renaming the constant and the function should be fine.  It's a purely internal change and has no functionality change.
- The return messages you updated are actually only returning a boolean.  Those functions are testing an error message against the strings up updated to determine if the error message passed in to the function matches those error strings.  I think this is the scenario Daria mentioned and is not safe to update, since these error messages are being returned from the server.

src/error.ts
```diff
@@ -714,7 +714,7 @@ export function isRetryableError(error: MongoError): boolean {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     (typeof error.code === 'number' && RETRYABLE_ERROR_CODES.has(error.code!)) ||
     error instanceof MongoNetworkError ||
-    !!error.message.match(/not master/) ||
+    !!error.message.match(/not primary/) ||
```
durran, line 717: I would create a constant for the regex `(not primary|not master)` and use it here.

src/error.ts
```diff
@@ -744,20 +744,20 @@ function isRecoveringError(err: MongoError) {
...
-    return SDAM_NOTMASTER_CODES.has(err.code);
+    return SDAM_NOTPRIMARY_CODES.has(err.code);
   }
 
   if (isRecoveringError(err)) {
     return false;
   }
 
-  return /not master/.test(err.message);
+  return /not primary/.test(err.message);
```
durran, line 760: I would use the constant mentioned in the other comment here to test for `(not primary|not master)`

durran (changes requested):

src/error.ts
durran: I would also add unit tests to test `isSDAMUnrecoverableError` in test/unit/error.test.js for both cases in the error message.

test/functional/unit-sdam/topology.test.js
```diff
@@ -227,7 +227,7 @@ describe('Topology (unit)', function () {
         if (doc.ismaster || doc.hello) {
           request.reply(Object.assign({}, mock.HELLO, { maxWireVersion: 9 }));
         } else if (doc.insert) {
-          request.reply({ ok: 0, message: 'not master' });
+          request.reply({ ok: 0, message: 'not primary' });
```
durran, line 230: "not writable primary"

src/error.ts
ljhaywar: I've added unit tests

ljhaywar:
We're seeing the same test failures in more than one PR, so I don't think these failures are related to the code changes.

@baileympearson this is ready for your review

src/error.ts
```diff
@@ -744,20 +752,23 @@ function isRecoveringError(err: MongoError) {
     return SDAM_RECOVERING_CODES.has(err.code);
   }
 
-  return /not master or secondary/.test(err.message) || /node is recovering/.test(err.message);
+  return (
+    LEGACY_NOT_PRIMARY_OR_SECONDARY_ERROR_MESSAGE_REGEX.test(err.message) ||
+    /node is recovering/.test(err.message)
```
baileympearson, line 757: Can we make a constant for this error message as well?

src/error.ts
```diff
@@ -8,6 +8,14 @@ export type AnyError = MongoError | Error;
 /** @internal */
 const kErrorLabels = Symbol('errorLabels');
 
+/** @internal */
+// Abstract the offensive terminology to a constant
+const LEGACY_NOT_PRIMARY_ERROR_MESSAGE_REGEX = /not master/;
```
baileympearson, line 13: Can we store the strings here as constants, not the regexs?  Then we can export them, and use the constant error messages in the unit tests as well.  

The functions below that use these constants will need to be updated to create a regex from the string 
``` typescript
new RegExp(LEGACY_NOT_PRIMARY_ERROR_MESSAGE)
```

test/unit/error.test.js
```diff
@@ -8,12 +8,20 @@ const { ReplSetFixture } = require('../tools/common');
 const { ns } = require('../../src/utils');
 const { Topology } = require('../../src/sdam/topology');
 const { MongoNetworkError, MongoWriteConcernError } = require('../../src/index');
-const { isRetryableEndTransactionError } = require('../../src/error');
+const {
```
baileympearson, line 11: I have a suspicion it won't be super straightforward, but if it is - can we convert this file to Typescript?

ljhaywar: Not sure what you mean. Which file do you think should be converted to TypeScript?

src/error.ts
ljhaywar: I waffled about this initially. Updated

src/error.ts
ljhaywar: I like it!

test/unit/error.test.js
baileympearson: `error.test.js` (this test file).  It should be as simple as updating the file extension to `.ts`, and then fixing the issues that pop up in VSCode.  But it's not necessary

src/error.ts
```diff
@@ -8,6 +8,27 @@ export type AnyError = MongoError | Error;
...
+ * https://github.com/mongodb/specifications/blob/b07c26dc40d04ac20349f989db531c9845fdd755/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-writable-primary-and-node-is-recovering
+ */
+export const LEGACY_NOT_PRIMARY_OR_SECONDARY_ERROR_MESSAGE = 'not master or secondary';
+
+/**
+ * @internal
+ * The error message from the server that indicates the node is recovering
+ * https://github.com/mongodb/specifications/blob/b07c26dc40d04ac20349f989db531c9845fdd755/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-writable-primary-and-node-is-recovering
+ */
+export const NODE_IS_RECOVERING_ERROR_MESSAGE = 'node is recovering';
```
nbbeeken, line 30:
```suggestion
export const NODE_IS_RECOVERING_ERROR_MESSAGE = /node is recovering/;
```
Since these are always used as RegExp, can we just declare them as that here? (Totally correct me if I missed something)

nbbeeken (changes requested):

test/unit/error.test.js
ljhaywar: Oh, got it! My hunch is that changing the file type seems out of scope for this ticket, but I'm happy to take it on if others agree. I don't want to tack in too many changes to this PR. I'm not sure how the team is handling the conversion (change a file as you touch it or changing the files in a batch).

src/error.ts
nbbeeken: Nevermind just saw bailey's comment, fair point!

nbbeeken (approved):

src/error.ts
ljhaywar: I had these as regular expressions initially. @baileympearson suggested making them strings, so they can be used in the unit tests as well.

test/unit/error.test.js
nbbeeken: I took a glance at the effort, for this file I think it's too many changes needed, others might be more straight forward but the sync require and automagic error importing is something we can do later I think

baileympearson (approved):

test/unit/error.test.js
baileympearson: Sounds good, we can hold off on this for now.

PR #3064, opened 2021-12-02
refactor(NODE-3717): reorganize tests part 2

dariakp:
### Description
NODE-3717

#### What is changing?
- crud and bulk tests were moved to `integration/crud`; one prose test was separated into a prose file from bulk
- max staleness test were moved to `integration/max-staleness` - they use a mock server but I think it's better to make them into proper integration tests than convert to a unit test (see my comment on this PR and new ticket)
- load balancer, retryable reads, and retryable writes tests have all been moved into a corresponding integration directory

NOTE: The changes in this PR affect the configuration for `run-serverless-tests.sh` we should double check to make sure all desired tests are still running (see my comment on that file).

N/A

Better test organization

### DIFF
--- a/.evergreen/run-serverless-tests.sh
+++ b/.evergreen/run-serverless-tests.sh
@@ -12,10 +12,10 @@ if [ -z ${SERVERLESS_ATLAS_USER+omitted} ]; then echo "SERVERLESS_ATLAS_USER is 
 if [ -z ${SERVERLESS_ATLAS_PASSWORD+omitted} ]; then echo "SERVERLESS_ATLAS_PASSWORD is unset" && exit 1; fi
 
 npx mocha --file test/tools/runner/index.js \
-  test/functional/crud_spec.test.js \
+  test/integration/crud/crud.spec.test.js \
   test/functional/retryable_reads.test.js \
   test/functional/retryable_writes.test.js \
   test/functional/sessions.test.js \
   test/functional/transactions.test.js \
   test/functional/versioned-api.test.js \
-  test/functional/load-balancer-spec.test.js
+  test/integration/load-balancers/load_balancers.spec.test.js
--- a/.evergreen/run-serverless-tests.sh
+++ b/.evergreen/run-serverless-tests.sh
@@ -14,7 +14,7 @@ if [ -z ${SERVERLESS_ATLAS_PASSWORD+omitted} ]; then echo "SERVERLESS_ATLAS_PASS
 npx mocha --file test/tools/runner/index.js \
   test/integration/crud/crud.spec.test.js \
   test/integration/retryable-reads/retryable_reads.spec.test.js \
-  test/functional/retryable_writes.test.js \
+  test/integration/retryable-writes/retryable_writes.spec.test.js \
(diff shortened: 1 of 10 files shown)


test/integration/crud/crud.prose.test.js
```diff
@@ -0,0 +1,43 @@
...
+    await client.connect();
+  });
+
+  afterEach(async () => {
+    if (client) {
+      await client.close();
+    }
+  });
+
+  // Note: This test does not fully implement the described test, it's missing command monitoring assertions
```
dariakp, line 18: Not adding a TODO ticket here because this is just getting a jump start on the spec compliance analysis work that will be done in NODE-3719, I only moved it at all because I knew it was intended to be a spec test (since I wrote it)

.evergreen/run-serverless-tests.sh
```diff
@@ -12,10 +12,10 @@ if [ -z ${SERVERLESS_ATLAS_USER+omitted} ]; then echo "SERVERLESS_ATLAS_USER is
 if [ -z ${SERVERLESS_ATLAS_PASSWORD+omitted} ]; then echo "SERVERLESS_ATLAS_PASSWORD is unset" && exit 1; fi
 
 npx mocha --file test/tools/runner/index.js \
```
dariakp, line 14: For every line that has been updated here, we need to make sure all the tests still run and pass as part of the `test-test-serverless` task in evergreen (reviewers: please compare the tests that run on this PR against the tests in waterfall, for this task, which runs under the "Serverless Test" configuration in evergreen)

.evergreen/run-serverless-tests.sh
```diff
@@ -12,10 +12,10 @@ if [ -z ${SERVERLESS_ATLAS_USER+omitted} ]; then echo "SERVERLESS_ATLAS_USER is
...
-  test/functional/crud_spec.test.js \
-  test/functional/retryable_reads.test.js \
-  test/functional/retryable_writes.test.js \
+  test/integration/crud/crud.spec.test.js \
+  test/integration/retryable-reads/retryable_reads.spec.test.js \
+  test/integration/retryable-writes/retryable_writes.spec.test.js \
   test/functional/sessions.test.js \
```
dariakp, line 18: Transaction and session tests both have tests that aren't the automated spec tests, so would end up getting separated upon moving, and since the serverless test spec requires us to run the prose tests, too (which aren't clearly identified in these combined test files), I think I'd rather do it in a separate PR from this one

.evergreen/run-serverless-tests.sh
durran: Even better - this PR has serverless running more tests than main. Main was missing retryable reads and writes it looks like.

durran: Looks like the same 400 in main and this PR, so good there.

durran (approved):

durran:
I'm leaving the conversations unresolved so anyone else who wants a look can see them easily, but feel free to resolve them yourself when you want to merge. LGTM.

PR #3065, opened 2021-12-03
test(NODE-3801): fix explain tests to work with latest server

dariakp:
### Description
NODE-3801

#### What is changing?
Updated the currently failing tests to consistently check for indicator properties anywhere in the response object, regardless of nesting level. The property "queryPlanner" should be present in all cases. The property "executionStats" should only be present when additional verbosity is requested.

No
Latest server changed the shape of the explain output, so we need to update our tests to make our CI green again.


nbbeeken (approved):
Clever trick checking the JSON string! 🚀

PR #3067, opened 2021-12-06
chore(NODE-3798): update dependencies; pull in mock server

nbbeeken:
#### What is changing?

Pulled in mock server, removed 'co' dependency, upgraded others.

Pulling in the mock server allows us to better debug and make changes as we phase out its use.


dariakp:
@nbbeeken does this partially cover NODE-3798?

nbbeeken:
Partially indeed, I can see about covering all of those points, or updating the ticket to list what's left to do? Any preference? 🙂

dariakp:
@nbbeeken It's your hackathon, so up to you, I would tag the ticket in the PR and then add a comment to the Jira saying that the moving part is done, then the ticket is just going to be for clean up (readme/deprecation stuff)

nbbeeken:
Updated the ticket, I'm happy to be the one to do the rest of the work later, but not during this week. If we want to close NODE-3798 along with this PR we can move the rest of the AC to a follow up.

dariakp (approved):
LGTM, I'd rather merge this now because I'm pretty sure it could potentially conflict with your circular deps work

baileympearson (approved):
Changes look good to me - nice work!  

Question: It was out of the scope of this work to rewrite the mock server to use BinMsg instead of the legacy op codes.  I can't find anything on Jira - do we have a ticket somewhere to update the mock server?

nbbeeken:
> do we have a ticket somewhere to update the mock server?
@baileympearson 
Our long term goal is to phase out use of it altogether, we now work only with server versions that have a wide array of failpoint features that should make the need of a mock server no longer necessary. So I'll actually make a ticket to track its removal 🙂

PR #3079, opened 2021-12-13
fix(NODE-3675): SRV option bug correctly defaults authSource to $external 

syz99:
#### What is changing?
Previously, for external mechanisms, it was not always the case that `authSource` defaulted to `$external` even though that was the intended behavior / what the spec outlines. External auth source mechanisms are now explicitly checked for in `connection_string` to avoid errant overrides (e.g. authSource = admin over $external). Unit tests added as well.

n/a

[ [jira](https://jira.mongodb.org/browse/NODE-3675) ]


test/unit/connection_string.test.js
```diff
@@ -94,6 +94,13 @@ describe('Connection String', function () {
     expect(options.credentials.mechanism).to.equal(AuthMechanism.MONGODB_GSSAPI);
   });
 
+  it('should provide default authSource when valid AuthMechanism provided', function () {
+    const options = parseOptions(
+      'mongodb+srv://jira-sync.pw0q4.mongodb.net/testDB?authMechanism=MONGODB-AWS&retryWrites=true&w=majority'
+    );
+    expect(options.credentials.source).to.equal('$external');
```
addaleax, line 101: Am I understanding correctly that an explicit `authSource` in the TXT record is supposed to be ignored if the auth mechanism is one that always uses a `$external` auth source?

I can see how that makes sense, but it’s also somewhat surprising and doesn’t match my understanding of the SRV spec, because generally options from the TXT record are in between the built-in defaults and the user-specified connection string in terms of priority. (Or the other way around: If we ignore `authSource` from the TXT record for certain auth mechanisms, why is `authSource` not also ignored if it is specified in the connection string?)

syz99:
I've chosen to replace any checks for external auth mechanisms (e.g.` if (GSSAP || AWS || X509)`) with a check for the newly introduced ($EXTERNAL_AUTH_SOURCE_MECHANISMS) set "typing".

This is mainly a choice of grouping these three as the external auth source mechanisms explicitly. Does this introduce too much confusion / make things less readable?

example diffs: ([1](https://github.com/mongodb/node-mongodb-native/pull/3079/files#:~:text=%24EXTERNAL_AUTH_SOURCE_MECHANISMS.has(-,mechanism,-))), ([2](https://github.com/mongodb/node-mongodb-native/pull/3079/files#:~:text=if%20(-,%24,-EXTERNAL_AUTH_SOURCE_MECHANISMS.has(this))

test/unit/connection_string.test.js
nbbeeken: I took a look at the spec and it does seem that this case isn't explicit spoken on, but the overriding rule is that these mechanisms MUST use $external. 

As for our normal URI parsing, I suppose it should ignore the option, fortunately by not ignoring it this bug does have a workaround by explicitly passing `authSource=$external` the driver will pull it in and the credentials will be approved by the `credentials.validate()`.

Currently our options are not parsed in a particular order, so the source might get set before the mechanism is determined. It might be better if we just coerce to `$external` in the credentials if the mechanism is one of the special ones instead of throwing, maybe? Because of the options parsing strangeness I'm kind of inclined to leave in the "feature" of having the option in the URI override the `$external` rule even if it leads to an error.

I think the real fix is that DNS records should be accurate but it can take time to update, but user's shouldn't have to manually work around that.

addaleax: > I took a look at the spec and it does seem that this case isn't explicit spoken on, but the overriding rule is that these mechanisms MUST use $external.

Then that sounds a bit like this should be clarified in the spec, both for authSource from TXT and from the user-passed connection string? :)

> Because of the options parsing strangeness I'm kind of inclined to leave in the "feature" of having the option in the URI override the `$external` rule even if it leads to an error.

Agreed, it's generally a good idea to error out if conflicting options were explicitly passed as input.

nbbeeken: DRIVERS-1036 Looks like there's something in the backlog for this, I left a comment describing what we're looking for :)

nbbeeken (dismissed):
LGTM, thanks!

test/unit/srv_option_handling.test.ts
```diff
@@ -0,0 +1,119 @@
+import { expect } from 'chai';
```
dariakp, line 1: The tests in the test/unit directory should follow the directory structure of the src folder (see the tests directory [README](https://github.com/mongodb/node-mongodb-native/blob/main/test/readme.md)): it looks like what this set of unit tests is trying to cover is the `resolveSRVRecord` method from the connection_string file, so we want to put this in the `connection_string.test.js` inside a `describe('resolveSRVRecord()' => { // })` block

test/unit/srv_option_handling.test.ts
```diff
@@ -0,0 +1,119 @@
...
+const resolveSRVRecordAsync = promisify(resolveSRVRecord);
+
+describe('Srv Option Handling', () => {
+  let resolveSrvStub: sinon.SinonStub;
+  let resolveTxtStub: sinon.SinonStub;
+  let resolveSRVRecordStub: sinon.SinonStub;
+  let lookupStub: sinon.SinonStub;
+
+  afterEach(async () => {
+    if (resolveSrvStub) {
```
dariakp, line 19: does `sinon.restore()` not work here?

test/unit/srv_option_handling.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      return process.nextTick(callback, null, mockAddress);
+    });
+
+    resolveTxtStub = sinon.stub(dns, 'resolveTxt').callsFake((address, thisIsWhatWeAreTesting) => {
+      thisIsWhatWeAreTesting(null, mockRecord);
+    });
+  }
+
+  for (const mechanism of $EXTERNAL_AUTH_SOURCE_MECHANISMS) {
+    it('should set authSource to $external for ${mechanism} external mechanism:', async function () {
```
dariakp, line 61: I think you meant this to be a template string

test/unit/srv_option_handling.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      makeStub('authSource=thisShouldNotBeAuthSource');
+      const options = {
+        credentials: new MongoCredentials({
+          source: '$external',
```
dariakp, line 65: since source is explicitly set as `$external` here, does it actually matter that a real external mechanism is used in the test? i.e., what would happen if we run this test with `source: '$external'` and `mechanism: 'potato'`? Also, what would happen if we use an external mechanism but set `source: 'admin'`?

dariakp (changes requested):

src/cmap/auth/providers.ts
```diff
@@ -10,5 +10,12 @@ export const AuthMechanism = Object.freeze({
   MONGODB_X509: 'MONGODB-X509'
 } as const);
 
+/** @public */
+export const $EXTERNAL_AUTH_SOURCE_MECHANISMS = new Set<AuthMechanism | undefined>([
```
baileympearson, line 14: Since all three auth providers are defined above, they'll never be `undefined`.  We should be able to remove the `| undefined` from the type of the set

baileympearson: Also, can we remove the `$` from the name of the variable?

test/unit/srv_option_handling.test.ts
syz99: yep, good catch!

src/cmap/auth/providers.ts
baileympearson: Oh I see, because when we use it, we potentially pass in an undefined value.  That's not something I had considered! 
 I'm okay with leaving the undefined then 😄

syz99: The reason we had the $ was to distinguish these mechanisms related to the `$external` authSource -- not sure if there's a better naming to include that context

test/unit/srv_option_handling.test.ts
syz99: 1) It matters since _if_ a non-external mechanism is used, an entirely new MongoCredentials instance would be called as found in `connection_string.ts`. So these cases check the `options.credentials` of a potentially different MongoCredentials instance not just whatever is in the options variable. Just added new assertions to test/show this! 

2) also above all we only want to test the SRV options handling case, not necessarily MongoCredentials in this scope

src/cmap/auth/providers.ts
syz99: 100% agree. Basically in this case, external does not mean like "outside of MongoDB" -- we're brainstorming other namings :)

test/unit/connection_string.test.ts
```diff
@@ -312,4 +325,107 @@ describe('Connection String', function () {
...
+
+        const options = {
+          credentials,
+          srvHost: 'test.mock.test.build.10gen.cc',
+          srvServiceName: 'mongodb',
+          userSpecifiedAuthSource: false
+        } as MongoOptions;
+
+        await resolveSRVRecordAsync(options);
+        expect(options.credentials === credentials).to.be.true;
```
dariakp, line 379: why not just check `expect(options.credentials).to.equal(credentials)`?

test/unit/connection_string.test.ts
```diff
@@ -312,4 +325,107 @@ describe('Connection String', function () {
...
+
+      const options = {
+        credentials,
+        srvHost: 'test.mock.test.build.10gen.cc',
+        srvServiceName: 'mongodb',
+        userSpecifiedAuthSource: false
+      } as MongoOptions;
+
+      await resolveSRVRecordAsync(options);
+      expect(options.credentials !== credentials).to.be.true;
```
dariakp, line 404: similar comment as in L379

dariakp (changes requested):

src/cmap/auth/providers.ts
dariakp: I second Bailey's suggestion here, is there really a chance of this causing confusion? @nbbeeken

test/unit/connection_string.test.ts
nbbeeken: This was my onion 🧅 , with my recent revelation that `.eql` means deep, and `.equal` means shallow. I have a preference to make clear our intention. We prefer `expect(x).to.deep.equal(y);` since it makes it obvious deepness is desired but there's no equivalent `.shallow`. (maybe we could add it!)

It's a rare case that shallow equality is important to a test that's checking something with objects so it feels like something that's easily overlooked, or something that when read later is thought to be a mistake. If someone changes this to be deep, it would also pass but it would lose the meaning originally intended. 

Maybe this is what code comments where invented for? 🤔

dariakp: yeah, overthinking it here a bit; `to.equal` is the correct assertion, and yes, code comments is where it's at if this is a concern, in fact, a code comment would be appropriate either way because it took me a minute to try to figure out why this assertion was there in the first place, so something like `// check reference identity because ...` would be ideal. Of course I'm sure @durran would tell you to just use a separate `it` assertion altogether to clarify the intention of your test cases; I don't have equally strong feelings about one assertion per `it` block, but I do want clarity, and using the most appropriate assertion for the job (in this case, it would actually be something like `expect(options).property('credentials').to.equal(credentials)`)

src/cmap/auth/providers.ts
```diff
@@ -10,5 +10,12 @@ export const AuthMechanism = Object.freeze({
   MONGODB_X509: 'MONGODB-X509'
 } as const);
 
+/** @public */
```
nbbeeken, line 13:
```suggestion
/** @internal */
```

nbbeeken (changes requested):

src/cmap/auth/providers.ts
```diff
@@ -10,5 +10,12 @@ export const AuthMechanism = Object.freeze({
...
+export const AUTH_MECHS_AUTH_SRC_EXTERNAL = new Set<AuthMechanism>([
+  AuthMechanism.MONGODB_GSSAPI,
+  AuthMechanism.MONGODB_AWS,
+  AuthMechanism.MONGODB_X509
+]);
+
 /** @public */
 export type AuthMechanism = typeof AuthMechanism[keyof typeof AuthMechanism];
```
nbbeeken, line 21: Can you move this type above the Set we're adding, since its associated with the AuthMechanism object

nbbeeken (changes requested):

nbbeeken (approved):

dariakp (approved):

baileympearson (approved):

PR #3082, opened 2021-12-14
feat(NODE-3793): Remove offensive language from code and tests

ljhaywar:
### Description

NODE-3793

#### What is changing?

Remove and deprecate (when removal is not possible) all references to "slaveOk" in the source code and tests. Note that spec test updates are being tracked in NODE-3791

No

Following the guidance documented in tools.ietf.org/id/draft-knodel-terminology-00.html, we will remove all oppressive and unnecessarily gendered language in driver documentation, code, tests, specs, and spec tests.

- [x] Changes are covered by tests. I'm unsure and have commented below

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -112,7 +112,11 @@ export interface CollectionOptions
   extends BSONSerializeOptions,
     WriteConcernOptions,
     LoggerOptions {
+  /**
+   * @deprecated Use secondaryOk instead
+   */
   slaveOk?: boolean;
+  secondaryOk?: boolean;
   /** Specify a read concern for the collection. (only MongoDB 3.2 or higher supported) */
   readConcern?: ReadConcernLike;
   /** The preferred read preference (ReadPreference.PRIMARY, ReadPreference.PRIMARY_PREFERRED, ReadPreference.SECONDARY, ReadPreference.SECONDARY_PREFERRED, ReadPreference.NEAREST). */
@@ -127,7 +131,7 @@ export interface CollectionPrivate {
   namespace: MongoDBNamespace;
   readPreference?: ReadPreference;
   bsonOptions: BSONSerializeOptions;
-  slaveOk?: boolean;
+  secondaryOk?: boolean;
   collectionHint?: Hint;
   readConcern?: ReadConcern;
   writeConcern?: WriteConcern;
@@ -182,7 +186,8 @@ export class Collection<TSchema extends Document = Document> {
       bsonOptions: resolveBSONOptions(options, db),
       readConcern: ReadConcern.fromOptions(options),
       writeConcern: WriteConcern.fromOptions(options),
-      slaveOk: options == null || options.slaveOk == null ? db.slaveOk : options.slaveOk
+      secondaryOk:
+        options == null || options.secondaryOk == null ? db.secondaryOk : options.secondaryOk
     };
   }
 
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -113,10 +113,9 @@ export interface CollectionOptions
     WriteConcernOptions,
     LoggerOptions {
   /**
-   * @deprecated Use secondaryOk instead
+   * @deprecated Use readPreference instead
    */
   slaveOk?: boolean;
-  secondaryOk?: boolean;
   /** Specify a read concern for the collection. (only MongoDB 3.2 or higher supported) */
   readConcern?: ReadConcernLike;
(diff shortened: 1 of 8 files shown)


ljhaywar:
General question: Are these changes covered by tests?  I tried changing the return value of slaveOk() to both true and false (instead of `this.secondaryOk;`) and all of the tests passed.

test/functional/cursor.test.js
```diff
@@ -3752,7 +3752,7 @@ describe('Cursor', function () {
           db.s.topology,
           db.s.namespace,
           {},
-          { limit: 0, skip: 0, slaveOk: false, readPreference: 42 }
+          { limit: 0, skip: 0, secondaryOk: false, readPreference: 42 }
```
ljhaywar, line 3755: This test is being skipped. I unskipped it, and it failed in the same way for both `slaveOk: false` and `secondaryOk: false`. I'm not sure how else to test this broken test.

dariakp: the comment above says that this test is questionable in value; and secondaryOk is not a valid command option for any command (slaveOk wasn't part of the public command options object); I think we can just delete the test

src/collection.ts
```diff
@@ -112,7 +112,11 @@ export interface CollectionOptions
   extends BSONSerializeOptions,
     WriteConcernOptions,
     LoggerOptions {
+  /**
+   * @deprecated Use secondaryOk instead
+   */
   slaveOk?: boolean;
+  secondaryOk?: boolean;
```
dariakp, line 119: We should see if this flag is ever actually parsed or used anywhere distinct from readPreference (and why? what should it control that readPreference doesn't?), otherwise, if readPreference is sufficient to cover the use case, we can just say "use readPreference" in the deprecation message. @nbbeeken @durran can you check my logic here?

dariakp (changes requested):
Just capturing this particular thought for now, not a full review

src/db.ts
```diff
@@ -185,8 +185,18 @@ export class Db {
     return this.s.options;
   }
 
-  // slaveOk specified
+  /**
+   * slaveOk specified
```
baileympearson, line 189: Is this method deprecated because it's a part of the `Db` class and removing it would be considered a breaking change?

baileympearson: It's only used in one place by us and that's for internal use

ljhaywar: I deprecated this one, because it's part of the public Db class. Please let me know if I didn't get that right. :-)

baileympearson: I think you're right!  Just unfortunate we're not able to remove it entirely because it's part of a public api

src/collection.ts
durran: From my understanding `secondaryOk` is required still when communicating with a mongos in conjunction with the read preference, but the driver should be doing this automatically and the user isn't required to do anything here. So I would say your logic is correct that we can say "use read preference" in the deprecation message.

src/db.ts
ljhaywar: I opened NODE-3818 to track removing this option in the next major release

src/collection.ts
ljhaywar: I've updated the deprecation note to instruct users to use the readPreference option instead.

I also deleted the slaveOk option and its associated logic in the internal CollectionPrivate interface.  The tests are still passing and the slaveOk option doesn't appear to be used anywhere else. I think this is what Daria and Neal were discussing in Slack yesterday.

test/unit/db.test.ts
```diff
@@ -0,0 +1,50 @@
+import { expect } from 'chai';
+
+import { MongoClient } from '../../src';
+import { Db, DbOptions } from '../../src/db';
+import { ReadPreference } from '../../src/read_preference';
+
+describe('Db - test secondaryOk matches readPreference', function () {
```
dariakp, line 7: Just a few points of feedback on the tests:

1) Can we call this describe something like "Db" or "Db instance", and then have all these tests nested in a second describe block "secondaryOk"?

2) Let's make sure we don't keep the .only on the tests. @nbbeeken  recently added a lint rule for this, but it might not be in your branch yet (could be worth a rebase)

3) Let's update the expect assertions to `expect(mydb).property('secondaryOk').to.be.<true/false>` and ```expect(mydb).property(`${legacy_secondary_ok}\`).to.be.<true/false>```

dariakp (changes requested):

ljhaywar:
Going to add a few more tests. Will comment here when I'm done

ljhaywar:
I've finished adding tests. This is ready for review.

dariakp (approved):
Looks good to me :)

baileympearson (approved):

PR #3083, opened 2021-12-16
refactor(NODE-3717): test reorg part 3 - sessions and transactions

dariakp:
### Description
NODE-3717

#### What is changing?
Moving session and transaction tests into the integration directory and separating spec tests from non-spec tests. The main thing we want to make sure is that the same tests that ran from those sessions and transactions files still run now. 

Note that for serverless tests, I left the "non-spec" test files in the list so that we don't lose coverage; once we get the prose tests identified and separated from that set (it's difficult to tell at a glance whether or not there are prose tests in the mix), we can update the serverless run to include just the prose files.

N/A

More manageable test organization and ease of audit


.evergreen/run-serverless-tests.sh
```diff
@@ -13,9 +13,12 @@ if [ -z ${SERVERLESS_ATLAS_PASSWORD+omitted} ]; then echo "SERVERLESS_ATLAS_PASS
 
 npx mocha --file test/tools/runner/index.js \
   test/integration/crud/crud.spec.test.js \
+  test/integration/crud/crud.prose.test.js \
   test/integration/retryable-reads/retryable_reads.spec.test.js \
   test/integration/retryable-writes/retryable_writes.spec.test.js \
-  test/functional/sessions.test.js \
-  test/functional/transactions.test.js \
+  test/integration/sessions/sessions.spec.test.js \
```
dariakp, line 19: Just like last time, for every line that has been updated here, we need to make sure all the tests still run and pass as part of the test-serverless task in evergreen (reviewers: please compare the tests that run on this PR against the tests in waterfall, for this task, which runs under the "Serverless Test" configuration in evergreen)

durran: I've done an Evergreen patch since the serverless tests don't run automatically on PRs. https://spruce.mongodb.com/version/61bb1112e3c3313ac9f9acc8/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

durran: Interesting is 1 more test seems to have run. Main has 400 passing/250 pending and this branch has 401 passing/250 pending. Same tests are skipped in both it looks like so struggling to see where the extra test came from, but I guess 1 more test is better than less. :)

durran (approved):

nbbeeken (approved):

.evergreen/run-serverless-tests.sh
dariakp: @durran I added the crud prose test file to the run (forgot to in the last PR), so that's where the extra test came from

PR #3086, opened 2021-12-16
refactor(NODE-3717): test reorg part 4

dariakp:
### Description
NODE-3717

#### What is changing?
Separating spec tests from integration tests and moving the related tests from functional into integration test directory.

This PR covers almost all files that actually reference existing spec files.

Note: I noticed that the connection string spec tests also include auth spec tests, but I think we can clean that up more when we do the full spec audit.

N/A

Consistent test organization


test/unit/cmap/connection_monitoring_and_pooling.spec.test.js
```diff
@@ -0,0 +1,281 @@
+'use strict';
```
durran, line 1: Should this be in `test/integration` instead of `test/unit` along with the connection string spec test? They are the only 2 in `test/unit` and even though they are closer to unit tests than integration tests with have other similar ones like uri options that's moved to `test/integration`. Maybe better to not have these 2 stragglers?

durran:
Test numbers on evg patch vs main look good. Just the one comment for clarification.

durran (approved):

PR #3088, opened 2022-01-03
fix(NODE-3843): make `GridFSBucketWriteStream.prototype.end()` return `this` for compat with community nodejs types

vkarpov15:
### Description

Currently `GridFSBucketWriteStream.prototype.end()` returns `void`. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/57473/files changed it so that `end()` now must return `this`, which means Mongoose and the Node driver are now incompatible with @types/node@17.0.6. First noticed this [here](https://github.com/Automattic/mongoose/runs/4686998575?check_suite_focus=true)

#### What is changing?

See above

No

Compilation fails with @types/node@17.0.6. I took a look and it looks like this change compiles fine with @types/node@16.x as well.

### DIFF
--- a/src/gridfs/upload.ts
+++ b/src/gridfs/upload.ts
@@ -172,21 +172,21 @@ export class GridFSBucketWriteStream extends Writable implements NodeJS.Writable
    * @param encoding - Optional encoding for the buffer
    * @param callback - Function to call when all files and chunks have been persisted to MongoDB
    */
-  end(): void;
-  end(chunk: Buffer): void;
-  end(callback: Callback<GridFSFile | void>): void;
-  end(chunk: Buffer, callback: Callback<GridFSFile | void>): void;
-  end(chunk: Buffer, encoding: BufferEncoding): void;
+  end(): this;
+  end(chunk: Buffer): this;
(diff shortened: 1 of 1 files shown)


marek-benes:
Please do review it and publish a new version. Affected projects are broken.

baileympearson (approved):
Thanks for submitting this change!  

This looks good to me.  I'll bring it to the team's attention for a second review and hopefully we can get a solution out in a timely manner.

durran (approved):

baileympearson:
Our plan is to release these changes in our next minor release, which we're targeting for Thursday.  I'm going to go ahead and merge this into main but we can continue the conversation here as needed.

In the meantime, would pinning the version of `@node/types` to an older version work as a temporary solution?  
@vkarpov15 @marek-benes

vkarpov15:
@baileympearson we pinned @types/node to < 17.0.6 in Mongoose, so I think it should work if this project pins in the meantime as well

Fazatholomew:
@baileympearson I can confirm that on my NestJS project which uses Mongoose 6.1.4, only <16.11.7 works. @types/node 17x has the problem

PR #3091, opened 2022-01-05
fix(NODE-3792): remove offensive language throughout the codebase

ljhaywar:
### Description

Remove references to "master" throughout the codebase

#### What is changing?

- Any references to master that can be updated are updated or abstracted to a constant
- Any references to master that cannot be updated are deprecated
- Exceptions: 
   - CSFLE references to masterKey are out of scope
   - Historical documentation (e.g. history.md and changes_x.md) can remain as-is
   - References to GitHub repos with master branches
   - Spec tests are covered in NODE-3791

No

Following the guidance documented in tools.ietf.org/id/draft-knodel-terminology-00.html, we will remove all oppressive and unnecessarily gendered language in driver documentation, code, tests, specs, and spec tests.

### DIFF
--- a/src/cmap/connect.ts
+++ b/src/cmap/connect.ts
@@ -6,6 +6,7 @@ import * as tls from 'tls';
 
 import type { Document } from '../bson';
 import { Int32 } from '../bson';
+import { LEGACY_HELLO_COMMAND } from '../constants';
 import {
   AnyError,
   MongoCompatibilityError,
@@ -70,15 +71,15 @@ export function connect(options: ConnectionOptions, callback: Callback<Connectio
   });
 }
 
-function checkSupportedServer(ismaster: Document, options: ConnectionOptions) {
+function checkSupportedServer(hello: Document, options: ConnectionOptions) {
   const serverVersionHighEnough =
-    ismaster &&
-    (typeof ismaster.maxWireVersion === 'number' || ismaster.maxWireVersion instanceof Int32) &&
-    ismaster.maxWireVersion >= MIN_SUPPORTED_WIRE_VERSION;
+    hello &&
+    (typeof hello.maxWireVersion === 'number' || hello.maxWireVersion instanceof Int32) &&
+    hello.maxWireVersion >= MIN_SUPPORTED_WIRE_VERSION;
   const serverVersionLowEnough =
-    ismaster &&
-    (typeof ismaster.minWireVersion === 'number' || ismaster.minWireVersion instanceof Int32) &&
-    ismaster.minWireVersion <= MAX_SUPPORTED_WIRE_VERSION;
+    hello &&
+    (typeof hello.minWireVersion === 'number' || hello.minWireVersion instanceof Int32) &&
+    hello.minWireVersion <= MAX_SUPPORTED_WIRE_VERSION;
 
   if (serverVersionHighEnough) {
     if (serverVersionLowEnough) {
@@ -86,13 +87,13 @@ function checkSupportedServer(ismaster: Document, options: ConnectionOptions) {
     }
 
     const message = `Server at ${options.hostAddress} reports minimum wire version ${JSON.stringify(
-      ismaster.minWireVersion
+      hello.minWireVersion
     )}, but this version of the Node.js Driver requires at most ${MAX_SUPPORTED_WIRE_VERSION} (MongoDB ${MAX_SUPPORTED_SERVER_VERSION})`;
     return new MongoCompatibilityError(message);
   }
 
   const message = `Server at ${options.hostAddress} reports maximum wire version ${
-    JSON.stringify(ismaster.maxWireVersion) ?? 0
+    JSON.stringify(hello.maxWireVersion) ?? 0
   }, but this version of the Node.js Driver requires at least ${MIN_SUPPORTED_WIRE_VERSION} (MongoDB ${MIN_SUPPORTED_SERVER_VERSION})`;
   return new MongoCompatibilityError(message);
 }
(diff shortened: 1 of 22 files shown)


src/cmap/connect.ts
```diff
@@ -148,7 +149,7 @@ function performInitialHandshake(
 
       if ('isWritablePrimary' in response) {
         // Provide pre-hello-style response document.
-        response.ismaster = response.isWritablePrimary;
+        response[LEGACY_HELLO_COMMAND] = response.isWritablePrimary;
```
ljhaywar, line 152: I think this is what @nbbeeken noted when we were refining this ticket. I have this note: "Current code ismaster.ismaster = ismaster.isWritablePrimary. Need to flip this."

Can someone provide more context on why we want to flip this?

nbbeeken: Whether the driver is connected to an old server or new server, we want to be able to write code that works uniformly across new / old servers. So instead of having this code that patches on a legacy key when communicating with a server that uses the new isWritablePrimary key, we can patch always patch in the new key if its not present. 

example:
```ts
if (!('isWritablePrimary' in response)) {
  // Provide hello-style response document.
  response.isWritablePrimary = response[LEGACY_HELLO_COMMAND];
}
```
With the above, if we're talking to a new server, isWritablePrimary is defined, so this doesn't run, but on an old server is we will attach it. Now the rest of the driver can always assume that the new key is present.

src/sdam/monitor.ts
```diff
@@ -256,14 +257,14 @@ function checkServer(monitor: Monitor, callback: Callback<Document>) {
...
       if (err) {
         failureHandler(err);
         return;
       }
-      if ('isWritablePrimary' in isMaster) {
+      if ('isWritablePrimary' in hello) {
         // Provide pre-hello-style response document.
-        isMaster.ismaster = isMaster.isWritablePrimary;
+        hello[LEGACY_HELLO_COMMAND] = hello.isWritablePrimary;
       }
```
nbbeeken, line 268: We'd want to do that flip we're discussing here too

src/cmap/connect.ts
ljhaywar: Gotcha. Thanks! 

Does this have a cascading effect?  Meaning, are their places in the driver code where it's looking for ismaster and not isWritablePrimary? The tests are passing on my machine, so I'm thinking this is probably OK...just wanted to verify with you.

nbbeeken: It's actually not used much, look for where you've changed to `hello[LEGACY_HELLO_COMMAND]`, I think I see two instances. Like the one in server description where we check for both, we could now just check for the new key.

ljhaywar: Thanks for the info! I made the changes, and the tests are failing (without even tryin the delete response you recommended). I've started debugging but haven't figured it out yet. I'll do more debugging on Monday.

ljhaywar: I started with fresh eyes today. Here is what I've learned.
- The check in `server_description` is checking if the server is a primary (`ismaster` or `isWritablePrimary`), so this one can be changed. In order to get the tests to pass with this change, I had to update line 171 of `test/functional/unit-sdam/server_selection/spec.test.js` to set `isWritablePrimary` instead of `ismaster`. I *think* this is OK, since this is a fake hello, and I'm not manipulating something coming back from a real server. Can you double check?
- I think the other one you were referring to is `isHello()` in `utils.ts`. `isHello()` is checking if the request is a hello command -- it's checking for `hello` or `ismaster` NOT `isWritablePrimary` or `ismaster`. I think we need to leave this one alone. Let me know if there is another file that I'm overlooking.

I added `delete response[LEGACY_HELLO_COMMAND];` in the two places you suggested:
- Patch with 3.6-replica_set: https://spruce.mongodb.com/version/61dc56b6c9ec4475aa752c58/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC. The tests are passing
- Patch with 5.0-sharded cluster: https://spruce.mongodb.com/version/61dc5ec161837d18207b9984/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC. The tests are passing but timing out with Context: Canceled.

src/sdam/monitor.ts
```diff
@@ -452,7 +450,7 @@ function measureRoundTripTime(rttPinger: RTTPinger, options: RTTPingerOptions) {
     return;
   }
 
-  connection.command(ns('admin.$cmd'), { ismaster: 1 }, undefined, err => {
+  connection.command(ns('admin.$cmd'), { LEGACY_HELLO_COMMAND: 1 }, undefined, err => {
```
nbbeeken, line 453:
```suggestion
  connection.command(ns('admin.$cmd'), { [LEGACY_HELLO_COMMAND]: 1 }, undefined, err => {
```

src/cmap/connect.ts
nbbeeken: We'll have to take a closer look at the hangs, 5.0 is the server version that speaks in hello so I wonder why it would behave differently, but everything else lgtm, one comment below tho, maybe it relates to the hangs.

src/sdam/monitor.ts
ljhaywar: OMG! Great catch! Thank you!

dariakp (commented):
The code all LGTM, really great job here @ljhaywar ! Just waiting on the tests to get sorted out :)

ljhaywar:
I reran the local patch from my machine with `delete response[LEGACY_HELLO_COMMAND];` in the two places Neal suggested, and this time the build passed:
https://spruce.mongodb.com/version/61dec74b2a60ed5bf60f7889/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Maybe it was the mistake I made that Neal caught? Or maybe it was unrelated as I've been seeing the Context Cancelled errors sporadically in my builds and others'? Either way, I think we're good to go.

nbbeeken (approved):

PR #3092, opened 2022-01-05
fix(NODE-3795): unexpected No auth provider for DEFAULT defined error

syz99:
### Description

Adding test case and fix for a specified authSource but no further credentials provided.
The spec states that authSource can be ignored when no other credentials are given, we have a check that is throwing in this specific scenario because we still have an empty credentials object making it down to the connection creation layer.

#### What is changing?

This PR removes the credentials from the options object in this specific scenario and tests for it.

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -394,6 +394,16 @@ export function parseOptions(
     }
 
     mongoOptions.credentials.validate();
+
+    // Check if the only auth related option provided was authSource, if so we can remove credentials
+    if (
+      mongoOptions.credentials.password === '' &&
+      mongoOptions.credentials.username === '' &&
+      mongoOptions.credentials.mechanism === AuthMechanism.MONGODB_DEFAULT &&
+      Object.keys(mongoOptions.credentials.mechanismProperties).length === 0
+    ) {
+      delete mongoOptions.credentials;
+    }
   }
 
   if (!mongoOptions.dbName) {
--- a/test/unit/connection_string.test.ts
+++ b/test/unit/connection_string.test.ts
@@ -89,6 +89,15 @@ describe('Connection String', function () {
     expect(options.credentials.source).to.equal('$external');
   });
 
+  it('should retain user-specified authSource with no credentials for non-srv URI', function () {
+    const mockAuthSource = 'thisShouldBeAuthSource';
+    const options = parseOptions(`mongodb://localhost/?authSource=${mockAuthSource}`);
+
+    expect(options).to.have.nested.property('credentials.username', '');
+    expect(options).to.have.nested.property('credentials.mechanism', 'DEFAULT');
+    expect(options).to.have.nested.property('credentials.source', mockAuthSource);
+  });
+
   it('should parse a numeric authSource with variable width', function () {
     const options = parseOptions('mongodb://test@localhost/?authSource=0001');
     expect(options.credentials.source).to.equal('0001');
@@ -346,6 +355,8 @@ describe('Connection String', function () {
       } as MongoOptions;
 
       await resolveSRVRecordAsync(options as any);
+      expect(options).to.have.nested.property('credentials.username', '');
+      expect(options).to.have.nested.property('credentials.mechanism', 'DEFAULT');
       expect(options).to.have.nested.property('credentials.source', 'thisShouldBeAuthSource');
     });
   });
--- a/test/unit/connection_string.test.ts
+++ b/test/unit/connection_string.test.ts
@@ -94,7 +94,7 @@ describe('Connection String', function () {
     expect(options.credentials.source).to.equal('$external');
   });
 
-  it('should retain user-specified authSource with no credentials', function () {
+  it('should omit credentials option when the only authSource is provided', function () {
     let options = parseOptions(`mongodb://a/?authSource=someDb`);
     expect(options).to.not.have.property('credentials');
     options = parseOptions(`mongodb+srv://a/?authSource=someDb`);
--- a/test/unit/connection_string.test.ts
+++ b/test/unit/connection_string.test.ts
@@ -334,5 +334,20 @@ describe('Connection String', function () {
       expect(options).property('credentials').to.equal(credentials);
       expect(options).to.have.nested.property('credentials.source', 'admin');
     });
+
+    it('should retain specified authSource with no provided credentials', async function () {
+      makeStub('authSource=thisShouldBeAuthSource');
+      const credentials = {};
+
+      const options = {
+        credentials,
+        srvHost: 'test.mock.test.build.10gen.cc',
+        srvServiceName: 'mongodb',
+        userSpecifiedAuthSource: false
+      } as MongoOptions;
+
+      await resolveSRVRecordAsync(options as any);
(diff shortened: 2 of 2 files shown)


test/unit/connection_string.test.ts
```diff
@@ -334,5 +334,19 @@ describe('Connection String', function () {
       expect(options).property('credentials').to.equal(credentials);
       expect(options).to.have.nested.property('credentials.source', 'admin');
     });
+
+    it('should retain specified authSource with no provided credentials', async function () {
+      makeStub('authSource=thisShouldBeAuthSource');
```
nbbeeken, line 339: Do we have a similar test for a non-srv connection string?

nbbeeken (changes requested):

test/unit/connection_string.test.ts
syz99: just added + added extra assertions on the `DEFAULT` mechanism to clearly test the specified error scenario in the ticket

test/unit/connection_string.test.ts
```diff
@@ -89,6 +89,15 @@ describe('Connection String', function () {
     expect(options.credentials.source).to.equal('$external');
   });
 
+  it('should retain user-specified authSource with no credentials for non-srv URI', function () {
+    const mockAuthSource = 'thisShouldBeAuthSource';
+    const options = parseOptions(`mongodb://localhost/?authSource=${mockAuthSource}`);
```
nbbeeken, line 94: Just commenting so we don't lose track, we're going to want to add an integration test for this, because the error that gets thrown is all the way down at the connection creation layer.

src/connection_string.ts
```diff
@@ -394,6 +394,16 @@ export function parseOptions(
     }
 
     mongoOptions.credentials.validate();
+
+    // Check if the only auth related option provided was authSource, if so we can remove credentials
```
dariakp, line 398: The PR description and the PR title both say that only test additions are included; is this supposed to be a `fix:`? If so, can we include more context in the PR desciption?

test/unit/connection_string.test.ts
```diff
@@ -89,6 +94,33 @@ describe('Connection String', function () {
...
+
+    expect(options).to.have.nested.property('credentials.username', '');
+    expect(options).to.have.nested.property('credentials.mechanism', 'DEFAULT');
+    expect(options).to.have.nested.property('credentials.source', mockAuthSource);
+  });
+
+  it('should omit credentials if the only auth related option is authSource', async () => {
+    const client = new MongoClient('mongodb://localhost:123/?authSource=someDb', {
```
dariakp, line 107: Why is this test connecting to a client while the other test above is using `parseOptions`? It's kind of confusing for a unit test, since, in general, if we are testing different functions then they shouldn't all be under the "Connection string" bucket, each helper should have its own describe block (like `resolveSrvRecordAsync` does).

For this specific test case, it's not clear why we have to run it through connect before checking the options.

dariakp (changes requested):

test/unit/connection_string.test.ts
nbbeeken: The error we're looking to **not** see is `new MongoInvalidArgumentError('No AuthProvider for ${credentials.mechanism} defined.')` in `prepareHandshakeDocument` and/or `performInitialHandshake`. Neither function is exported currently but if I did export them becuase of the inlined callbacks I think I would need to mock quite a bit of internals to get down to that layer. My thinking is I can lean on server selection failing for th unit tests to assert we at least don't get an error related to auth. I'm happy to comment something to this effect and move the test to a better spot, thoughts?

src/connection_string.ts
nbbeeken: done!

test/unit/connection_string.test.ts
dariakp: I think it's fine to do what you are saying with the comments, no need for convoluted mocks, but you could even kind of document the intention in the assertions and the test title by asserting that the thrown error is not a MongoAPIError derivative (and we can also make the test title reflect that expectation); obligatory suggestion to split into 2 separate test cases - one that it should not throw, and the other about what the actual result is expected to be

nbbeeken: The success case requires me to have an actual server to connect to, I think it could be done in an int test but I think its overkill for this case. Additionally because this is covered by the uri-options spec tests, so we'll have a more complete coverage when those are fixed up.

test/unit/connection_string.test.ts
```diff
@@ -89,6 +94,39 @@ describe('Connection String', function () {
...
+  it('should retain user-specified authSource with no credentials for non-srv URI', function () {
+    const mockAuthSource = 'thisShouldBeAuthSource';
+    const options = parseOptions(`mongodb://localhost/?authSource=${mockAuthSource}`);
+
+    expect(options).to.have.nested.property('credentials.username', '');
+    expect(options).to.have.nested.property('credentials.mechanism', 'DEFAULT');
+    expect(options).to.have.nested.property('credentials.source', mockAuthSource);
+  });
+
+  it('should omit credentials and throw a non-MongoAPIError if the only auth related option is authSource', async () => {
```
dariakp, line 106:
```suggestion
  it('should omit credentials and not throw a MongoAPIError if the only auth related option is authSource', async () => {
```

dariakp (changes requested):
Just a small suggestion because I think the goal of the test, if I understand it, is to make sure that we don't fail in the wrong place, but there's nothing invalid about the setup - we only fail because there isn't a proper server, not because we specified an authSource, right?

test/unit/connection_string.test.ts
nbbeeken: Yea I'm relying on no server running on localhost:123, and then checking we're not throwing the InvalidArg error, so I agree with the phrasing update here.

dariakp (approved):

test/unit/connection_string.test.ts
```diff
@@ -89,6 +94,39 @@ describe('Connection String', function () {
     expect(options.credentials.source).to.equal('$external');
   });
 
+  it('should retain user-specified authSource with no credentials for non-srv URI', function () {
```
durran, line 97: This test is failing on evg as it expects an empty username but the property does not exist.

durran (changes requested):

test/unit/connection_string.test.ts
```diff
@@ -89,6 +94,37 @@ describe('Connection String', function () {
     expect(options.credentials.source).to.equal('$external');
   });
 
+  it('should retain user-specified authSource with no credentials', function () {
```
dariakp, line 97: in what way are the assertions below "retaining" the authSource? the previous assertions were about having the property present, but now it's actually omitting it, so I'm confused about the expected behavior here. What's supposed to happen to authSource=someDb?

dariakp (changes requested):

test/unit/connection_string.test.ts
nbbeeken: I changed the functionality after the test was written whoops! the credentials object should be discarded if the only auth related option specified is authSource. Updated!

dariakp (approved):

PR #3095, opened 2022-01-06
chore: improve test filtering to use mocha hooks

nbbeeken:
#### What is changing?

Use a beforeEach hook to run test filters.

Also: remove docker stuff, move the sdam_viz tool to etc with out other helper scripts, corrected small type issue in unified test schema, enabled test sorting

- xuint files will now all have the same test names
- tests are always searchable in logs
- `LOG_FILTER_REASON` can be enabled to view which filter made a test skip

### Follow up work

Our legacy spec runner still "multiplies" the tests by runOn requirement and then relies on the filter to skip the ones that don't apply. Fixing this would drastically reduce the test count without actually removing tests.
Now that we have a standardized runOn requirement documented in the unified format, it would be nice to refactor our filtering metadata to match that, and reuse the logic in topologySatisfies in the unified runner.


durran (changes requested):
These tests https://github.com/mongodb/node-mongodb-native/blob/d407b9f92b628fd33cff393bcab8fefe8f5bfeaf/test/integration/client-side-encryption/client_side_encryption.prose.test.js#L541 need to be filtered to run on only server version 4.2 and higher to not fail with "Unknown $jsonSchema keyword: encrypt"

etc/sdam_viz.js
```diff
@@ -1,9 +1,9 @@
-#!/usr/bin/env node
-'use strict';
+/* eslint-disable no-console */
+/* eslint-disable @typescript-eslint/no-var-requires */
```
baileympearson, line 3: Did you mean to move this file into the `etc` directory?  I don't see this directory in the codebase as it is now, so I'd be curious to know what the purpose of this directory is if this change was intentional

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -552,14 +552,16 @@ describe('Client Side Encryption Prose Tests', function () {
     const limitsKey = loadLimits('limits-key.json');
     const limitsDoc = loadLimits('limits-doc.json');
 
-    before(function () {
-      // First, perform the setup.
+    let firstTimeSetup = true;
+    beforeEach(async function () {
```
baileympearson, line 556: could we just use a `before` hook here to run this code once before the tests start?

test/tools/runner/filters/client_encryption_filter.js
```diff
@@ -40,11 +39,7 @@ class ClientSideEncryptionFilter {
     const clientSideEncryption =
       test.metadata && test.metadata.requires && test.metadata.requires.clientSideEncryption;
 
-    // CSFLE is only supported on LTS versions of node
```
baileympearson, line 43: Question: Are we removing the node version check because we always test on versions of node >4?

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -552,14 +552,16 @@ describe('Client Side Encryption Prose Tests', function () {
...
+      if (firstTimeSetup) {
+        firstTimeSetup = false;
```
nbbeeken, line 558: So `before` hooks will still run if tests are skipped which was causing an issue here, the before hook was doing logic that also depended on a certain mongo version. So introducing this change would add a kind of hazzard. This is similar to how our session leak checker is the first afterEach that runs so you can't use afterEach to close clients currently. 
I still think the change is worth it and we can follow up with some better work around potentially later. Thoughts? 

I will add notes about both of these special scenarios to the test/readme

baileympearson (changes requested):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
baileympearson: disregard this comment - I didn't see Neal's comment when I reviewed the PR.

etc/sdam_viz.js
nbbeeken: Yes I did, it should be there, there's a prepare.js script and crawfish.mjs in there. Since this is a stand alone helper script and doesn't get imported by any tests it would better belong in here.  Also the imports were broken

test/tools/runner/filters/client_encryption_filter.js
nbbeeken: Yep!

baileympearson (dismissed):
LGTM.  Nice work Neal! 

This will also nice to be have for the test reorg work, because now we can clearly see that the correct number of tests are being considered with each test run 😄

test/readme.md
```diff
@@ -143,6 +144,18 @@ modify the steps to work with existing Node projects.
...
+
+> TODO: flesh this section out more
+
+We use mocha to construct our test suites and chai to assert expectations.
+
+Some special notes on how mocha works with our testing setup:
+
+- `before` hooks will run even if a test is skipped by the environment it runs on.
+  - So, for example, if your before hook does logic that can only run on a certain server version you can't depend on your test block metadata to filter for that.
+- `after` hooks cannot be used to clean up clients because the session leak checker currently runs its afterEach hook first.
```
dariakp, line 157:
```suggestion
- `after` hooks cannot be used to clean up clients because the session leak checker currently runs in an `afterEach` hook, which would be executed before any `after` hook has a chance to run
```

test/readme.md
```diff
@@ -143,6 +144,18 @@ modify the steps to work with existing Node projects.
 
 > **Note:** When making driver changes, you will need to run `npm run build:ts` with each change in order for it to take effect.
 
+## Writing Tests
+
+> TODO: flesh this section out more
```
dariakp, line 149: are you intentionally leaving the TODO here?

test/tools/runner/index.js
```diff
@@ -18,92 +17,84 @@ const MONGODB_API_VERSION = process.env.MONGODB_API_VERSION;
...
 
-function filterOutTests(suite) {
-  suite.tests = suite.tests.filter(test => filters.every(f => f.filter(test)));
-  suite.suites.forEach(suite => filterOutTests(suite));
-}
+beforeEach(async function () {
+  if (Object.keys(this.currentTest.metadata).length > 0) {
+    let ok = true;
+    for (const filter of filters) {
+      ok = ok && filter.filter(this.currentTest);
```
dariakp, line 60: just a sanity check - is the metadata prop always defined?

since the only way for the test to be skipped is for a filter to get applied, we can simplify the nesting here, I think, no need to keep track of an "ok", just do something like 
```js
const failedFilter = filters.find(filter => !filter.filter(this.currentTest));
if (failedFilter) { 
  // LOG_FILTER_REASON logic
  this.skip();
}
```

dariakp (changes requested):
Just a few thoughts here

dariakp:
@nbbeeken One more question, you mention the legacy spec runner follow up work - do we already have a ticket for that?

test/readme.md
nbbeeken: There's a lot more that could be written here about how to write good tests / how our to go about fitting into the way we write them, I could file follow up but I'd want to be specific if I made tickets. Like reccomending hooks for client creation and cleanup, or table tests when possible, or minimize assertions to keep a test clear. Does that all sounds like something that would fit here? 

My short term goal was to capture these hazards at least somewhere

test/tools/runner/index.js
nbbeeken: Cleaned this up, much more streamlined thanks!
cc: @baileympearson pinging you for another look cus of the changes I mention below:
Annnnd, modified the filter logging logic, I was using a conditional hack here of modifying the test name so that you could see it for debugging purposes but its useful information in our daily workflow I think. Apparently Go's built in test framework has a way of printing the skip reason for a test so I gave it a shot for us, here's what it looks like:

```
Retryable Reads
    aggregate-serverErrors - single,replicaset
      ✔ Aggregate fails after NotWritablePrimary when retryReads is false
    aggregate-serverErrors - sharded,load-balanced
      - Aggregate fails after NotWritablePrimary when retryReads is false
        ↬ filtered by MongoDBTopologyFilter - { requires: { topology: [ 'sharded', 'load-balanced' ], mongodb: '>= 4.1.7', authEnabled: false, serverless: undefined } }
``` 

We could even use this to attach reasons for our manual skips, checkout `monitoring.test.js` The ticket number will now get logged next to the test. I think now that mocha will log all test names on all runs, its an easy way for folks to find common spec tests and quickly locate the reason for a skip! And our xunit output file will still have only the test name so the test history is preserved in evergreen! 

shout out to @divjotarora for giving me the idea! 🚀

nbbeeken:
> `@nbbeeken` One more question, you mention the legacy spec runner follow up work - do we already have a ticket for that?

Yep! here: https://jira.mongodb.org/browse/NODE-3817

test/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "../tsconfig.json",
+    "compilerOptions": {
+        "strict": false,
+        "allowJs": true,
+        "checkJs": false
+    },
+    "include": ["../node_modules/@types/mocha/index.d.ts", "./mongodb-mocha.d.ts",]
```
baileympearson, line 8: I don't think the trailing comma is meant to be here.  In fact - doesn't JSON not allow trailing commas at all?  I'm surprised VSCode didn't complain about this

nbbeeken: tsconfig files can be written in JSONC, comments and trailling commas are allowed. I don't think I need this file anymore anyway though

test/tools/runner/index.js
```diff
@@ -9,101 +9,92 @@ const fs = require('fs');
...
-  // NOTE: if we first parse the connection string and redact auth, then we can reenable this
-  // const usingUnifiedTopology = !!process.env.MONGODB_UNIFIED_TOPOLOGY;
-  // console.log(
-  //   `connecting to: ${chalk.bold(MONGODB_URI)} using ${chalk.bold(
-  //     usingUnifiedTopology ? 'unified' : 'legacy'
-  //   )} topology`
-  // );
+beforeEach(async function () {
+  // `metadata` always exists, `requires` is optional
+  const requires = this.currentTest.metadata.requires;
```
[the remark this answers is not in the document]
nbbeeken: It's attached to all tests regardless of if it is specified because metadata_ui.js always at least returns an empty object from `_parseArgs`.

test/readme.md
dariakp: I think it's a good idea to do in the future, but this is fine for now

dariakp (approved):

PR #3098, opened 2022-01-07
chore(NODE-3717): move tests p-z

baileympearson:
#### What is changing?

This PR moves tests from the functional test folder into integration.  This PR handles files named with letters p-z.  

This PR also improves the view test to make it a true integration test, as well as convert it to typescript.

No


test/functional/promises_collection.test.js
```diff
@@ -1,318 +0,0 @@
-'use strict';
```
dariakp, line 1: Great job breaking this up!

dariakp (changes requested):
Just a few suggestions/discussion points here, but overall great job with this refactor!

test/integration/collection-management/view.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    db = client.db(configuration.db);
+  });
+
+  afterEach(async function () {
+    await db.dropCollection('test');
+    await client.close();
+  });
+
+  it('should successfully create a view on a collection', {
+    metadata: { requires: { topology: 'single' } },
```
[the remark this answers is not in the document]
baileympearson: Not that I know of.  I actually just copied this from the original test file but even then, it was mocked so it didn't really matter.  I'll remove the restriction to just this topology unless Neal can think of a reason we shouldn't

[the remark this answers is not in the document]
nbbeeken: All good to remove, unless the test uses a feature that is specific to certain deployment we should try testing on all

test/integration/node-specific/mongo_client.test.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+const { setupDatabase } = require('../shared');
+const f = require('util').format;
+
+class CustomPromise extends Promise {}
+CustomPromise.prototype.isCustomMongo = true;
+
+describe('MongoClient', function () {
```
nbbeeken, line 8: There's another describe block that uses this same title, currently it prints out as a warning, cus I think when I added the code there was many duplicates but since then we've cleaned them up. We have to keep each title unique or the xunit file won't output every test. 

Maybe you can change this line to a throw? or just fix this name for now
https://github.com/mongodb/node-mongodb-native/blob/main/test/tools/reporter/mongodb_reporter.js#L179

test/integration/connection-monitoring-and-pooling/multiple_db.test.js
```diff
@@ -0,0 +1,20 @@
...
+    var configuration = this.configuration;
+    const client = configuration.newClient({}, { sslValidate: false });
+    client.connect(function (err, client) {
+      for (var i = 0; i < 100; i++) {
+        client.db('test');
+      }
+
+      client.close(done);
+    });
+  });
```
nbbeeken, line 19: I think this test can be deleted, no idea what listeners its referring to here

dariakp: Yeah, the multiple_db test should have kept the first test "shouldCorrectlyUseSameConnectionsForTwoDifferentDbs" (maybe update the name) per my original comment and deleted the other two

test/integration/collection-management/promise_collection_db_management.test.js
```diff
@@ -0,0 +1,109 @@
+'use strict';
+const { assert: test, setupDatabase } = require('../shared');
+const f = require('util').format;
+
+class CustomPromise extends Promise {}
+CustomPromise.prototype.isCustomMongo = true;
+
+describe('Collection Management (promise tests)', function () {
```
dariakp, line 8:
```suggestion
describe('Collection and Db Management (promise tests)', function () {
```

test/integration/collection-management/promise_collection_db_management.test.js
```diff
@@ -0,0 +1,109 @@
...
+  before(function () {
+    return setupDatabase(this.configuration);
+  });
+
+  it('Should correctly createCollection using Promise', function (done) {
+    var configuration = this.configuration;
```
dariakp, line 14: could we update all the vars to const in this file?

test/integration/crud/bulk.test.js
```diff
@@ -21,6 +22,116 @@ describe('Bulk', function () {
     return setupDatabase(this.configuration);
   });
 
+  context('promise tests', () => {
+    it('Should correctly execute unordered bulk operation in promise form', {
+      metadata: {
+        requires: {
+          topology: ['single', 'replicaset', 'sharded']
```
dariakp, line 29: Do we actually want to explicitly exclude these from running against anything that returns "load-balanced" (same question on the other files where we have kept the metadata specified)?

test/integration/crud/find_and_modify.test.js
```diff
@@ -9,6 +9,157 @@ describe('Find and Modify', function () {
...
+
+        test: function (done) {
+          var configuration = this.configuration;
+          var client = configuration.newClient(configuration.writeConcernMax(), {
+            maxPoolSize: 1
+          });
+
+          client.connect().then(function (client) {
+            var db = client.db(configuration.db);
+            // LINE var MongoClient = require('mongodb').MongoClient,
```
dariakp, line 30: let's not keep these commented lines; it's a hold over of some sort of autogenerated docs that's definitely no longer relevant

test/integration/crud/insert.test.js
```diff
@@ -0,0 +1,70 @@
...
+describe('crud - insert', function () {
+  before(function () {
+    return setupDatabase(this.configuration);
+  });
+
+  context('promise tests', () => {
+    it('Should correctly execute Collection.prototype.insertOne', {
+      metadata: {
+        requires: {
+          topology: ['single']
```
dariakp, line 15: why just single here?

test/integration/node-specific/mongo_client.test.js
dariakp: My vote is to do both; as far as renaming, just something like "MongoClient integration" would work fine

dariakp (changes requested):

test/integration/crud/bulk.test.js
baileympearson: I wouldn't think so but I also wasn't sure.  I decided not to add the topology `load balanced` because even though I didn't think it would break anything, I wasn't sure.

This does make me think though - it seems like instead of opting-in to topologies, we should be explicitly blocking topologies.  I would guess that when Durran added `load-balanced` as an option, he didn't update all the tests to include `load-balanced` as a valid topology, which is why it's not included in this list.  Flipping the logic in our filter is definitely out of scope for this PR but I think it might be worthwhile.  Thoughts?

baileympearson: But for now, for this particular test, I can remove the restrictions unless there's a reason not to.

dariakp: I think we should expect all our tests to work on all topologies by default, so the idea of opt out is sound, but out of scope here; we'd probably want to do something like "requires: { topology: { not: [ ... ] } }" for the tests that do need special filtering and think further about the implications; but yes, just removing the requirement for now on these tests is a good first step

test/integration/node-specific/mongo_client.test.js
baileympearson: When I tried to modify the reporter to throw instead of warn, I noticed that the afterEach session checker hooks were what failed.  For now, I'd rather just change the name of the test to fix the duplicate test warning and we can look at this later

test/integration/crud/insert.test.js
baileympearson: Same thoughts here as above.  I'll remove the topology restrictions here too

dariakp (approved):
Looking good now, I don't think we need to worry about hunting down all the instances of `var` at this point, so it should be good as is, provided the pipeline passes

dariakp (approved):

PR #3101, opened 2022-01-11
test(NODE-2556): add integration testing for listDatabases' authorizedDatabases flag

syz99:
### Description
added int tests for the `authorizedDatabases` option of `listDatabases`

#### What is changing?
a new `enumerating-databases` int test folder with a `list_databases` test file has been added with appropriate tests

drive-by deletion/cleanup of gitkeep files

n/a 


test/integration/enumerating-databases/list_databases.test.ts
```diff
@@ -0,0 +1,77 @@
+import { expect } from 'chai';
+
+import { AddUserOptions, MongoClient, MongoServerError } from '../../../src';
+
+describe('listDatabases', function () {
+  beforeEach(function () {
+    if (process.env.AUTH !== 'auth') {
```
nbbeeken, line 7: Let's add a todo here connected to the NODE-3860 ticket

test/integration/enumerating-databases/list_databases.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    it('should error for authorizedDatabases set to false', async function () {
+      let thrownError;
+      try {
+        await authorizedClient.db().admin().listDatabases({ authorizedDatabases: false });
+      } catch (error) {
+        thrownError = error;
+      }
+
+      if (thrownError) {
```
nbbeeken, line 71: If thrownError is false-y then we aren't asserting anything here, I think we should expect the error to always exist here? did you see something else in manual testing

test/integration/enumerating-databases/list_databases.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      this.skip();
+    }
+  });
+
+  describe('authorizedDatabases flag', function () {
+    const username = 'a';
+    const password = 'b';
+    const mockAuthorizedDb = 'mockAuthorizedDb';
+
+    let client, authorizedClient: MongoClient;
```
nbbeeken, line 17: It's preferable to me at least that we have each variable on its own line, but a little more than just style reasons we're missing the annotation for client.

test/integration/enumerating-databases/list_databases.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      const adminListDbs = await client.db().admin().listDatabases();
+      const authorizedListDbs = await authorizedClient
+        .db()
+        .admin()
+        .listDatabases({ authorizedDatabases: true });
+      const adminDbs = adminListDbs.databases;
+      const authorizedDbs = authorizedListDbs.databases;
+
+      expect(adminDbs).to.have.length.greaterThan(1);
+      expect(authorizedDbs).to.have.length(1);
```
nbbeeken, line 56: I think you have to use `lengthOf` here
https://www.chaijs.com/api/bdd/#method_lengthof

test/integration/enumerating-databases/list_databases.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    const mockAuthorizedDb = 'mockAuthorizedDb';
+
+    let client, authorizedClient: MongoClient;
+
+    const authorizedUserOptions: AddUserOptions = {
+      roles: [{ role: 'read', db: mockAuthorizedDb }]
+    };
+
+    beforeEach(async function () {
+      client = this.configuration.newClient({ auth: { username: 'user', password: 'password' } });
```
nbbeeken, line 24: Can we add a coment that these username and pass comes form mlaunch's defaults. 
When we fix the CI issue we'll prob have to change these to fetched from the environment

nbbeeken (changes requested):
Nice work!

test/integration/enumerating-databases/list_databases.test.ts
syz99: No you're right I was being explicit in case for some reason that isn't hit, but I guess if it isn't hit then the test assertion itself would catch that "wrong" behavior

test/integration/enumerating-databases/list_databases.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import { AddUserOptions, MongoClient, MongoServerError } from '../../../src';
+
+describe.only('listDatabases', function () {
+  // TODO(NODE3860): Create driver test variants that run with AUTH enabled
+  // beforeEach(function () {
+  //   if (process.env.AUTH !== 'auth') {
+  //     this.skip();
+  //   }
+  // });
```
nbbeeken, line 11:
```suggestion
  beforeEach(function () {
    if (process.env.AUTH !== 'auth') {
      this.skip();
    }
  });
```

We still need the skip, but the comment LGTM

test/integration/enumerating-databases/list_databases.test.ts
```diff
@@ -0,0 +1,79 @@
+import { expect } from 'chai';
+
+import { AddUserOptions, MongoClient, MongoServerError } from '../../../src';
+
+describe.only('listDatabases', function () {
```
nbbeeken, line 5:
```suggestion
describe('listDatabases', function () {
```

Let's give the other tests a chance to run! 🚀  😁

nbbeeken (changes requested):

test/integration/enumerating-databases/list_databases.test.ts
```diff
@@ -0,0 +1,79 @@
+import { expect } from 'chai';
```
baileympearson, line 1: My understanding of the `integration` directory is that the directory structure should roughly match the `spec` repo.  I don't think there's a spec for `enumerating-databases`.  I'm not sure exactly where these tests belong, maybe @nbbeeken or @dariakp  would have thoughts.  

I was thinking perhaps in `auth` because they're specifically testing auth functionality right now.

dariakp: There is actually a top level enumerate-databases spec, which probably matches this the closest. I would name the file enumerate-databases.test.ts and put it in top level integration directory

baileympearson: Ah, my bad.  I missed it because I only looked at the specs that were in sub-directories

test/integration/enumerating-databases/list_databases.test.ts
syz99: sorry uncommented to test, putting back.

nbbeeken (dismissed):

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,79 @@
+import { expect } from 'chai';
+
+import { AddUserOptions, MongoClient, MongoServerError } from '../../src';
+
+describe('listDatabases', function () {
+  // TODO(NODE3860): Create driver test variants that run with AUTH enabled
```
baileympearson, line 6: Am I reading this comment backwards?  It looks like these tests only run when auth is enabled.

dariakp: To add on to the question, which variants do these run under right now?

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  beforeEach(function () {
+    if (process.env.AUTH !== 'auth') {
+      this.skip();
```
dariakp, line 9: Can we make use of the skipReason that Neal implemented in his recent PR to append the reason to the skip?

test/integration/enumerate_databases.test.ts
dariakp, line 6:
```suggestion
  // TODO(NODE-3860): Create driver test variants that run with AUTH enabled
```

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    let client: MongoClient;
+    let authorizedClient: MongoClient;
+
+    const authorizedUserOptions: AddUserOptions = {
+      roles: [{ role: 'read', db: mockAuthorizedDb }]
+    };
+
+    beforeEach(async function () {
+      // pass credentials from cluster_setup's mlaunch defaults
+      // TODO(NODE3860): pass credentials instead based on environment variable
```
dariakp, line 27:
```suggestion
      // TODO(NODE-3860): pass credentials instead based on environment variable
```

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      expect(adminDbs).to.have.length.greaterThan(1);
+      expect(authorizedDbs).to.have.lengthOf(1);
+
+      expect(adminDbs.filter(db => db.name === mockAuthorizedDb).length).equals(1);
+      expect(adminDbs.filter(db => db.name !== mockAuthorizedDb).length).greaterThan(0);
+      expect(authorizedDbs.filter(db => db.name === mockAuthorizedDb).length).equals(1);
+    });
+
+    it('should error for authorizedDatabases set to false', async function () {
```
dariakp, line 67: is this test supposed to be for the case where there are no authorized databases? is so, that should be part of the test case description, because setting the flag itself to false should not cause an error; we should have tests here that show a successful use of the authorizedDatabases flag set to false as well as a case for the flag being omitted entirely

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+describe('listDatabases', function () {
+  // TODO(NODE3860): Create driver test variants that run with AUTH enabled
+  beforeEach(function () {
+    if (process.env.AUTH !== 'auth') {
+      this.skip();
+    }
+  });
+
+  describe('authorizedDatabases flag', function () {
```
dariakp, line 13: general question - do we not need to cover the different server versions? is 4.0.5 server not in our pipeline?

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      await authorizedClient.connect();
+    });
+
+    afterEach(async function () {
+      await client.db('admin').removeUser(username);
+      await client.close();
+      await authorizedClient.close();
+    });
+
+    it('should list authorized databases for authorizedDatabases set to true', async function () {
```
dariakp, line 50: I am confused by the test setup here - can we clearly state what different cases are being tested and what the expected outcome is? Right now the assertions are a mix of what I assume are sanity checks on the setup and actual assertions pertaining to the behavior under test - it would be good to clearly separate that in the test, with clarifying comments if needed. Also, can we consistently use the lengthOf assertion to assert on the length?

dariakp (changes requested):
In addition to the comments regarding the test structure, please do not delete the .gitkeep files. The idea behind them is to make sure that the spec-related folders exist even if there are no files in them, so even if for the folders with files present we don't want to accidentally delete the folder by moving its contents elsewhere.

test/integration/enumerate_databases.test.ts
nbbeeken:
```suggestion
      this.currentTest.skipReason = 'TODO(NODE-3860): requires AUTH enabled';
      this.skip();
```

There's not really docs on this so just fyi this is how you would do it!

test/integration/enumerate_databases.test.ts
nbbeeken: capturing convo from slack - we shouldn't need to break out tests for before and after that version, our choice of making a user with a `read` role should give the same behavior across versions.

test/integration/enumerate_databases.test.ts
syz99: currently using mlaunch's default auth which is username: 'user', password: 'password' (so to run that locally I add an --auth flag in the cluster_setup script for server and comment out the skip test)

test/integration/enumerate_databases.test.ts
syz99: yeah this was for the case of existing authorized databases, with the flag set to false -- sorry updated the test descriptions to be more specific

syz99:
Offloading ticket now. PR could be paused since a few new TODO's needed:  testing duplicates and verifying listDatabases return type is an array (i.e. verifying an "Iterable of Document types" as on the spec.

Scope for this PR is to complete Test Case 1 [here](https://github.com/mongodb/specifications/blob/master/source/enumerate-databases.rst#test-plan).

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,119 @@
...
+        this.currentTest.skipReason =
+          'TODO(NODE-3860): Create driver test variants that require AUTH enabled';
+        this.skip();
+      }
+    });
+
+    beforeEach(async function () {
+      // pass credentials from cluster_setup's mlaunch defaults
+      // TODO(NODE-3860): pass credentials instead based on environment variable
+      client = this.configuration.newClient({ auth: { username: 'user', password: 'password' } });
```
baileympearson, line 53: question @nbbeeken  - based on our conversation this morning, this test will need to be adjusted when I make changes to the cluster-setup script to use bob/pwd123 instead of the default user and password?

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      await client.db('admin').addUser(username, password, authorizedUserOptions);
+
+      authorizedClient = this.configuration.newClient({
+        auth: { username: username, password: password }
+      });
+      await authorizedClient.connect();
+    });
+
+    afterEach(async function () {
+      await client?.db('admin').removeUser(username);
```
baileympearson, line 70: Do we need to remove the collection `a` created in the `beforeEach` block here as well?

test/integration/enumerate_databases.test.ts
nbbeeken: It should probably just switch to using `newClient()` with no args, and the configuration class will handle using the right connection string or setting the auth settings appropriately.

test/integration/enumerate_databases.test.ts
nbbeeken: Sure, let me also give it a less common name

baileympearson (approved):

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    await client.connect();
+  });
+
+  afterEach(async function () {
+    await client?.close();
+  });
+
+  it('should return an Array', async () => {
+    const dbInfo = await client.db().admin().listDatabases();
+    expect(Array.isArray(dbInfo.databases)).to.be.true;
```
dariakp, line 19: does `to.be.an('array')` not work here?

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  });
+
+  it('should contain no duplicates', async () => {
```
dariakp, line 22: So, this setup doesn't really create any databases, right? If we really want to ensure we are testing the no duplicates assertion, I think we have to first create the conditions that could conceivably cause duplicates (e.g., try to create more than one database with the same name). Testing that an empty set is empty doesn't really accomplish that.

test/integration/enumerate_databases.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    afterEach(async function () {
+      await client?.db('admin').removeUser(username);
+      await client?.db(mockAuthorizedDb).dropDatabase();
+      await client?.close();
+      await authorizedClient?.close();
+    });
+
+    it('should list authorized databases with authorizedDatabases set to true', async function () {
```
dariakp, line 77: can we add scenarios for passing true/false flag explicitly into listDatabases for the case of the user that has the full listDatabases permissions and show that the returned list is not affected?

dariakp (changes requested):

test/integration/enumerate_databases.test.ts
nbbeeken: TIL! I thought this only took `typeof` results.

test/integration/enumerate_databases.test.ts
nbbeeken: Since it's just not possible to create a duplicate named DB I feel like that would be testing MongoDB more than our driver. The spec doesn't require us to return an array, it says we could return an iterable, I think the test's goal is to make sure that one's iterable doesn't mistakenly loop over a value twice. If the server somehow handed us back two DB's with the same name, currently our logic wouldn't do anything to stop that from reaching the user, which I think is "more" correct in that its passing along the server's mistake.  

This test and the  one above are described in the spec's test plan, so pulling them  in even though they're not really significant for our API feels right for completion sake.

test/integration/enumerate_databases.test.ts
nbbeeken: done!

test/integration/enumerate_databases.test.ts
dariakp: can we at least assert that the list is not empty?

dariakp: Also, if there are prose tests here, can we separate them into a prose test file and explicitly include the test plan instructions?

dariakp (changes requested):

test/integration/enumerate_databases.test.ts
nbbeeken: Reorganized and added length assertions as well as a check that it results in an array of objects. I pulled in the test plan in a comment at the top of the new file

baileympearson (approved):

test/integration/enumerate_databases/enumerate_databases.prose.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    const dbInfo = await client.db().admin().listDatabases();
+    expect(dbInfo).to.have.property('databases');
+    expect(dbInfo.databases).to.be.an('array');
+    expect(dbInfo.databases).to.have.lengthOf.at.least(1);
+    for (const db of dbInfo.databases) {
+      expect(db).to.be.a('object');
+    }
+  });
+
+  it('Verify that all databases on the server are present in the result set', async () => {
```
dariakp, line 33: We should try to make sure that we have a clean environment with a known list of dbs for this test, and then assert on the expected list

dariakp (changes requested):

test/integration/enumerate_databases/enumerate_databases.prose.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      .map(({ name }) => name)
+      .filter(name => !REQUIRED_DBS.includes(name));
+
+    for (const dbToDrop of databasesToDrop) {
+      await client.db(dbToDrop).dropDatabase();
+    }
+
+    await client.db(DB_NAME).createCollection(DB_NAME);
+
+    ENTIRE_DB_LIST = (await client.db('admin').command({ listDatabases: 1 })).databases.map(
```
baileympearson, line 33: non-blocking: we could use `nameOnly` here so we don't have to `map ({ name }) => name` later on

baileympearson (dismissed):

baileympearson:
I left one small comment but it's non-blocking.  LGTM

test/integration/enumerate_databases/enumerate_databases.prose.test.ts
nbbeeken: `nameOnly: true` returns in the shape of:
```js
{
  databases: [ { name: 'admin' }, { name: 'config' }, { name: 'local' } ],
  ok: 1
}
```

So nothing would change, it omits the metadata from each db description object reducing the BSON transmitted

test/integration/enumerate_databases/enumerate_databases.prose.test.ts
nbbeeken: done!

test/integration/enumerate_databases/enumerate_databases.prose.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      ({ name }) => name
+    );
+    ENTIRE_DB_LIST.sort();
```
dariakp, line 36: Should we add an expectations here around the ENTIRE_DB_LIST not being empty?

dariakp (changes requested):
Just one small thing

test/integration/enumerate_databases/enumerate_databases.prose.test.ts
nbbeeken: sure!

dariakp (approved):
LGTM, tyvm :)

PR #3103, opened 2022-01-12
chore(NODE-3717): test reorg penultimate part

dariakp:
### Description
NODE-3717

#### What is changing?
Many things - basically, everything except functional/core: notably, the aws test has been moved, so we should sanity check to make sure the configuration still runs ok.

A number of tests have been consolidated with existing integration and unit test files. Things that aren't explicitly spec-related were moved more or less wholesale into the node-specific directory: this includes operation examples and the general examples folder. We can clean that up further later, right now though the goal is to make it easier to determine spec compliance, and for that purpose these files can be safely ignored.

N/A

Get tests that test the same things next to each other so that it's easier to tell what we have.


test/integration/node-specific/mongo_client.test.js
```diff
@@ -1,18 +1,240 @@
...
+
+  //////////////////////////////////////////////////////////////////////////////////////////
+  //
+  // new MongoClient connection tests
+  //
+  //////////////////////////////////////////////////////////////////////////////////////////
+  it('Should open a new MongoClient connection', {
+    metadata: {
+      requires: {
+        topology: ['single']
```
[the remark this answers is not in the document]
dariakp: Yes, if you look at the commit history, I did try to remove the topology restrictions resulting in them all failing against unsupported topologies: the tests would have to updated to correctly initialize against different topologies, and, given that I'm not certain that these tests are even worth keeping in their current form, I didn't want to go through the effort of figuring out exactly which changes are needed to work with which topologies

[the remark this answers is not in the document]
dariakp: By the way, the changes in the file are primarily in copying existing tests from elsewhere over into the file and indentation, not changing the tests themselves

test/integration/node-specific/examples/versioned_api.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-const { MongoClient } = require('../../../src');
+const { MongoClient } = require('../../../../src');
```
nbbeeken, line 2: Can we rename this file to end in `.test.js`?

test/integration/node-specific/mapreduce.test.js
```diff
@@ -14,7 +13,7 @@ describe('MapReduce', function () {
    */
   it('shouldPerformMapReduceWithStringFunctions', {
     metadata: {
-      requires: { topology: ['single', 'replicaset', 'sharded', 'ssl', 'heap', 'wiredtiger'] }
+      requires: { topology: ['single', 'replicaset', 'sharded'] }
```
nbbeeken, line 16: I see we removed the extraneous ones here, any reason to not remove them all?

test/unit/sdam/max_staleness.spec.test.js
```diff
@@ -0,0 +1,59 @@
...
+    serverConnect = sinon.stub(Server.prototype, 'connect').callsFake(function () {
+      this.s.state = 'connected';
+    });
+  });
+
+  after(() => {
+    serverConnect.restore();
+  });
+
+  const specTests = collectStalenessTests(maxStalenessDir);
```
nbbeeken, line 46: Nothing to change here, just interesting that this doesn't use the shared `loadSpecTests` function. Maybe its worth filing a TODO?

test/integration/node-specific/mapreduce.test.js
dariakp: My experience with removing the topology requirement in mongo client wasn't great, see my answer to Bailey's question; I mean we can try it here, but we may end up taking all day to figure out what is or isn't safe to remove

test/unit/sdam/server_selection_spec_helper.js
```diff
@@ -1,147 +1,17 @@
...
-const { ReadPreference } = require('../../../../src/read_preference');
-const { MongoServerSelectionError } = require('../../../../src/error');
-const ServerSelectors = require('../../../../src/sdam/server_selection');
-
-const { EJSON } = require('bson');
+const { Topology } = require('../../../src/sdam/topology');
+const { ServerType, TopologyType } = require('../../../src/sdam/common');
+const { ServerDescription } = require('../../../src/sdam/server_description');
+const { ReadPreference } = require('../../../src/read_preference');
+const { MongoServerSelectionError } = require('../../../src/error');
```
nbbeeken, line 6: Does it make sense to move this to `tools/spec-runner`? And/or rename it from `helper` -> `runner`?

test/unit/sdam/topology.test.js
```diff
@@ -25,7 +39,7 @@ describe('Topology (unit)', function () {
 
       test: function (done) {
         // Attempt to connect
-        var server = new Topology([`${this.configuration.host}:${this.configuration.port}`], {
+        var server = new Topology([`localhost:27017`], {
```
nbbeeken, line 42: What's the change here for? shouldn't it fetch from the environment?

test/integration/auth/mongodb_aws.test.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-const expect = require('chai').expect;
+const { expect } = require('chai');
 
 describe('MONGODB-AWS', function () {
   before(function () {
```
nbbeeken, line 5: A slight enhancement while we're here, now that these will show up in the other variants can we add a skipReason to this hook?

test/unit/sdam/server_selection_spec_helper.js
dariakp: possibly, but there's some standardization overall that we can make for things like this which will be more evident when we do the full spec audit (up next after finishing the general sorting phase)

test/unit/sdam/max_staleness.spec.test.js
dariakp: same comment regarding spec standardization as on the helper file

test/integration/node-specific/examples/versioned_api.js
nbbeeken: True it would need to skip based on `MONGODB_API_VERSION` being defined, we don't need to do that here.

test/integration/auth/mongodb_aws.test.js
dariakp: Any particular suggestion for the skip reason?

test/unit/sdam/server_selection_spec_helper.js
nbbeeken: Yea makes sense, kinda goes along with my other comment about the custom loader functions, we'll defer to later

test/integration/auth/mongodb_aws.test.js
nbbeeken: cus of this: `MONGODB_URI.indexOf('MONGODB-AWS')` it could be `test requires MONGODB_URI to contain 'MONGODB-AWS' auth mechanism` I don't think you need to attach the value of MONGODB_URI cus its logged elsewhere

nbbeeken (approved):

test/integration/auth/mongodb_aws.test.js
dariakp: Will include in the next refactor

baileympearson (approved):

PR #3105, opened 2022-01-13
chore(NODE-3717): test reorg - the conclusion

dariakp:
### Description
NODE-3717

#### What is changing?
Some final clean up of the functional/core tests and removal of remaining dead files in the functional directory.

Of note, our snappy tests have been silently failing for over a month due to the directory path not having been updated in package.json, so this PR also enables the mocha `failZero` flag, which prevents test runs from showing as passing if no tests actually ran.

No, although some test-related readme cleanup is included as part of this PR.

Our tests are finally sorted! 🎉 


test/unit/error.test.js
```diff
@@ -47,6 +49,46 @@ describe('MongoErrors', () => {
     });
   }
 
+  describe('MongoError#constructor', () => {
```
nbbeeken, line 52: Would these assertions be better added to the loop over all the errors here: https://github.com/mongodb/node-mongodb-native/blob/main/test/unit/error.test.js#L34-L48

test/tools/reporter/mongodb_reporter.js
```diff
@@ -135,10 +135,7 @@ class MongoDBMochaReporter extends mocha.reporters.Spec {
         timestamp = timestamp ? timestamp.toISOString().split('.')[0] : '';
 
         output.testSuites.push({
-          package:
-            suite.file.includes('functional') || suite.file.includes('integration')
-              ? 'Functional'
-              : 'Unit',
+          package: suite.file.includes('integration') ? 'Functional' : 'Unit',
```
nbbeeken, line 138: This is a bit of a throwaway field in the xunit output, it doesn't impact anything on EVG, should we just name it `integration` now?

package.json
```diff
@@ -104,7 +104,7 @@
     "check:eslint": "eslint -v && eslint --max-warnings=0 --ext '.js,.ts' src test",
     "check:tsd": "tsd --version && tsd",
     "check:dts": "./node_modules/typescript/bin/tsc --noEmit mongodb.d.ts && tsd",
-    "check:test": "mocha --file test/tools/runner --recursive test/functional test/integration",
+    "check:test": "mocha --file test/tools/runner --recursive test/integration",
```
nbbeeken, line 107: We can drop the recursive flag from here, I think at some point it was needed for some runs and not others but its enabled in our mocharc

test/unit/sdam/server_discovery_and_monitoring.spec.test.js
```diff
@@ -19,7 +19,7 @@ const { eachAsyncSeries, ns } = require('../../../src/utils');
 
 const chai = require('chai');
 chai.use(require('chai-subset'));
-chai.use(require('../../functional/spec-runner/matcher').default);
+chai.use(require('../../tools/spec-runner/matcher').default);
```
nbbeeken, line 22: Should we use the shared chai-addon.js file here? Actually we should consolidate any `chai.use` calls, I see 20 instances, quick win?

nbbeeken (changes requested):

test/unit/error.test.js
dariakp: I thought about this, but I feel like it's kind of awkward in there, because we can't really recursively define error constructor behavior outside the one shared name property, so then if we do it in the loop, we have to check for the error class name, which feels super hacky.

test/tools/reporter/mongodb_reporter.js
dariakp: oh I have no idea what the output goes to... does the reporter only run on the integration directory now and isn't used for our actual unit tests, or do you just mean `Functional -> Integration`?

test/unit/sdam/server_discovery_and_monitoring.spec.test.js
dariakp: sure, good call; I assume you're just suggesting "require(</path/to/chai-addons.js>)?

test/tools/reporter/mongodb_reporter.js
nbbeeken: The reporter works for both unit and integration cus its pointed at by the mocharc. And yea, I just mean the string "Integration". It's just a prop on the very root node of the xml output:

```xml
<testsuite package="Unit" id="0" name="..." timestamp="..." hostname=".." tests="0" failures="0" errors="0" time="0">
```

test/unit/error.test.js
nbbeeken: Fair enough, let's keep this then!

test/unit/sdam/server_discovery_and_monitoring.spec.test.js
nbbeeken: yep!

global.d.ts
```diff
@@ -58,6 +58,16 @@ declare global {
        * - my test
        *   - TODO(NODE-XXXX): Feature implementation impending!
        * ```
+       *
```
baileympearson, line 61: Should this documentation live in our testing readme instead of the global types file?

baileympearson: Nevermind, I see you added it there as well

dariakp: Yeah, Neal reminded me that we have it in the types even though the types aren't quire working yet for test

test/unit/sdam/server_discovery_and_monitoring.spec.test.js
dariakp: turns out this is already in the mocharc file, so no need for those lines at all

nbbeeken (approved):

baileympearson (approved):

PR #3112, opened 2022-01-20
chore(NODE-3719,NODE-3883): identify causal consistency spec tests

dariakp:
### Description
NODE-3719, NODE-3883

#### What is changing?
Identifying implemented and missing causal consistency spec tests; spec also says we should run on sharded topologies, not just replica set, so I expanded the required topology list.

N/A

Spec compliance audit


test/integration/causal-consistency/causal_consistency.prose.test.js
```diff
@@ -0,0 +1,285 @@
...
+          .then(() => {
+            const command = test.commands.started[1].command;
+            expect(command).to.have.any.key('readConcern');
+            expect(command.readConcern).to.not.have.any.key('afterClusterTime');
+          });
+      }
+    }
+  );
+
+  // TODO[CHECK WITH TEAM]: This may be an invalid spec test case in light of SPEC-1019
```
dariakp, line 258: Open question here, kind of a confusing situation since it seems that SPEC-1019 has removed this from being relevant, not sure if there's a drivers ticket to update this case or if it's just an oversight

dariakp (commented):
Open question

test/integration/causal-consistency/causal_consistency.prose.test.js
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * session = client.startSession(causalConsistency = true)
+     * document = collection.anyReadOperation(session, ...)
+     * capture the command sent to the server (using APM or other mechanism)
+     * assert that the command does not have an afterClusterTime
+     */
+    {
+      metadata: {
+        requires: { topology: ['replicaset', 'sharded'] },
+        // Skipping session leak tests b/c these are explicit sessions
```
nbbeeken, line 46: This is interesting, I would think this test still clean up the sessions it creates

nbbeeken (approved):
LGTM, looks like we captured everything. One Q that we could add to the follow up ticket

PR #3102, opened 2022-01-11
fix(NODE-3852,NODE-3854,NODE-3856): Misc typescript fixes for 4.3.1

baileympearson:
### Description

This PR includes three Typescript fixes / improvements.

#### What is changing?

1. NODE-3856 - allows users to user a document of their collection schema type as a filter in `findX` operations
2. NODE-3854 - update the return types of `findOneAnd<op>` to include the `_id` field.  The behavior of these functions is now aligned with the general `findX` methods
3. NODE-3852  - Adds support for _some_ recursive schemas

##### Is there new documentation needed for these changes?

Probably.  I think we should document exactly what is and is not supported somewhere once we flesh out exactly what we're supporting and we're not supporting.

### Double check the following

- [x] Ran `npm run check:lint` script
- [x] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [x] PR title follows the correct format: `<type>(NODE-xxxx)<!>: <description>`
- [x] Changes are covered by tests
- [x] New TODOs have a related JIRA ticket

PR #3115, opened 2022-01-25
fix(NODE-3895): Emit TypeScript errors when filtering on fields not in collection schema

noahsilas:
### Description
As a developer, I'd like the type system to catch errors I make, like trying to
query on a field that doesn't exist, or when I have made a typo in an existing
field name.

For example:

```typescript
interface Person = {
  name: string;
  age: number;
}
 

db.collection<Person>('people').find(
  { nam: 'Alice' } // should be a type error, as `nam` is not an attribute of the schema
);
```

Related JIRA: https://jira.mongodb.org/browse/NODE-3895

#### What is changing?
The TypeScript `Filter<TSchema>` definition is becoming more type safe.

Previously, querying for these fields outside the schema would pass type
checks because the `Filter<TSchema>` type included a union with
`RootFilterOperators<WithId<TSchema>>`, and `RootFilterOperators` was
defined as `extends Document`. The BSON `Document` type itself is an
object with the index property `{ [key: string]: any }`, which seemed to match
any keys present in the document that didn't have stricter rules applied.

By removing `extends Document` from that definition, we gain significant
type safety; several issues in the tests turned up. This also showed some
difficulties with the recent typing additions in #3102 - some tests started
failing with this change because the `NestedPath<>` generation was bailing
out early even on non-recursive structures (when a nested object happened
to have a subset of keys of the parent). The final commit in this series
improves how recursive unions are detected.

As one final note, folks who have started using that new recursive object
support will find that unsupported queries (those reaching through a
recursive relation) will now need to be decorated with a `@ts-expect-error`
or similar annotation. This should call attention to these queries lack of
type safety (instead of the current behavior where the filter is implicitly
cast to an `any`).

Good question! I didn't find any very meaningful documentation about the
existing TypeScript experience of this library, but if that exists somewhere
I'd be happy to update it given a pointer to the page.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -284,6 +284,7 @@ export type {
   InferIdType,
   IntegerType,
   IsAny,
+  IsInUnion,
   Join,
   KeysOfAType,
   KeysOfOtherType,
@@ -306,6 +307,7 @@ export type {
   RootFilterOperators,
   SchemaMember,
   SetFields,
+  TypeEquals,
   UpdateFilter,
   WithId,
   WithoutId
--- a/src/mongo_types.ts
+++ b/src/mongo_types.ts
@@ -91,8 +91,23 @@ export type AlternativeType<T> = T extends ReadonlyArray<infer U>
 /** @public */
 export type RegExpOrString<T> = T extends string ? BSONRegExp | RegExp | T : T;
 
+/**
+ * This is a type that allows any `$` prefixed keys and makes no assumptions
+ * about their value types. This stems from a design decision that newly added
+ * filter operators should be accepted without needing to upgrade this package.
+ *
+ * This has the unfortunate side effect of preventing type errors on unknown
+ * operator keys, so we should prefer not to extend this type whenever possible.
+ *
+ * @see https://github.com/mongodb/node-mongodb-native/pull/3115#issuecomment-1021303302
+ * @public
+ */
+export type OpenOperatorQuery = {
+  [key: `$${string}`]: unknown;
+};
+
 /** @public */
-export interface RootFilterOperators<TSchema> {
+export interface RootFilterOperators<TSchema> extends OpenOperatorQuery {
   $and?: Filter<TSchema>[];
   $nor?: Filter<TSchema>[];
   $or?: Filter<TSchema>[];
--- a/src/mongo_types.ts
+++ b/src/mongo_types.ts
@@ -90,7 +90,7 @@ export type AlternativeType<T> = T extends ReadonlyArray<infer U>
 export type RegExpOrString<T> = T extends string ? BSONRegExp | RegExp | T : T;
 
 /** @public */
-export interface RootFilterOperators<TSchema> extends Document {
+export interface RootFilterOperators<TSchema> {
   $and?: Filter<TSchema>[];
   $nor?: Filter<TSchema>[];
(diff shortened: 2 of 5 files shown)


dariakp:
@noahsilas Thank you for your thoughtful PR and suggestions here: the reason we have RootFilterOperators extend Document is so that any driver version can be forwards compatible with new query operators that are continually being added to the server. There is an improvement that we would love to make to only extend $-prefixed keys instead of the generic Document (since operators are always $-prefixed), but we haven't had a chance to look into making that work.

noahsilas:
@dariakp Can I encourage a reversal of that thinking? While I appreciate the idea of future compatibility, any usage of those operators would still be type-unsafe until the user upgrades (as the value would implicitly and silently be `any`).

In the meantime, folks trying to use the existing operators also lose some of their type safety. For instance, I personally often try to use `$eql` instead of `$eq` - this would be a fantastic error for the type system to catch early!

Finally, it's possible for people writing queries with the latest-and-greatest operators to explicitly opt-out of the type checker if they know that their server supports a query operator that isn't on this explicit list of known operators in the TypeScript definition: they can annotate that query with `@ts-expect-error`. This even has the added benefit of flagging that line for review when they eventually do upgrade this package to a version that supports the new operator! On the other hand, it's _not_ possible for me as a consumer of this library to opt in to a stricter subset of known operators if all keys starting with a `$` prefix are allowed with an `any` type.

Thank you for considering this alternate point of view! 😄

dariakp:
@noahsilas We are always looking to improve the user experience, so we'll discuss this further at our next team meeting to see if we can come up with some less breaking way to allow for this frequently requested feature. Thanks again for reaching out!

noahsilas:
@dariakp Thank you for the context! For now I've implemented the improvement you asked for where the `RootFilterOperators<TSchema>` type allows arbitrary `$`-prefixed keys: see commit 6974918 at the end of this PR. I've filed [NODE-3904](https://jira.mongodb.org/browse/NODE-3904) so that I could link to it from that commit and capture some of the details there.

baileympearson:
Hey @noahsilas, thanks for taking the time with this PR. After discussing this with the team, we decided that it would be best to defer strengthening the Filter type to a future major version release and ensure that we have a proper technical design for the functionality going forward. Some of our concerns were:
 
### Recursive Types
At this point in time, we believe the solution we have is preferrable to the solution in this PR. We do not believe it's best practice to make a change that causes users to annotate type errors with @ts-expect-error or a similar compiler directive. Recursive schemas are permitted in both Typescript and the Node driver, and with the changes to recursive types included in this PR valid recursive type queries will show up as compiler errors. Our philosophy is that in the scenario where the driver can't provide strict type safety, we should default to any so we don't cause any unnecessary errors in user's code.
 
### Limiting the extension to $-keys instead of Document
While this is something we would like to see merged eventually, the current implementation breaks recursive types and dot notation. For a recursive type, the dot-notation key will not be generated as a part of the NestedPaths type (the core component of our recursive types implementation). In this scenario, the dot-notation key is permitted by the extends Document. By limiting the extends Document to only $-prefixed keys, we will cause errors for valid dot notation keys for recursive types. Similar to the previous section, we will not merge changes that cause type errors for a valid filter.
 
This is something we'd like to address in the future but don't have time for at the moment. Feel free to take a stab at it as well if you'd like!

PR #3119, opened 2022-01-28
test(NODE-3817): refactor RunOn filter in legacy spec runner to use mocha hooks

nbbeeken:
#### What is changing?

- Removes the logic that duplicates tests in the legacy spec runner based on runOn spec
  - Legacy spec runner will now reshape the legacy runOn requirement to match the unified format and use the same function from the unified runner to calculate test skip
  - `topologySatisfies` can be called from a test or beforeEach hook
- Moves our start up hooks, to Mocha's require format (`--file` flag is slated for removal)
  - The configuration.js hook is our old tools/runner/index.js file, it will detect if the run is for unit tests
    - ~Is the unit test detection logic strong enough? it uses the `process.argv[2]` string~ 
    - No unit test detection, we have a config file that is used for integration runs, and unit tests use the default `.mocharc.json` file
  - prints a clear specification of the current detected environment
- Retryable Writes had custom runOn parsing logic, inherited similiar changes to the legacy spec runner
- SDAM integration tests explcitly skip on LB topologies
- ADL - patches in an empty runOn array, they're the only legacy specs that don't have them
- Converted the TestConfiguration to TS, corrected small issue with auth option making it on to the searchParams
- Unified runner attaches skipReason found in test files to our mocha skipReason now
- Snappy the "unit" test was hitting the session leak checker for using an after hook instead of an afterEach

This makes our test run more clear since there is not longer duplicates of a test. For example a countDocuments spec test is the same regardless of what topology it is running against, so we shouldn't duplicate the test name and function based on all possible topology we'd want to run that test against.

Migrating our hooks to the require format is to help us keep up to date on mocha's expectations. Writing them in this format is a precurser to potentially running tests in parallel.


test/tools/spec-runner/index.js
```diff
@@ -92,109 +93,123 @@ function gatherTestSuites(specPath) {
...
-    spec.operations.some(
-      op => op.name === 'waitForEvent' && op.arguments.event === 'PoolReadyEvent'
-    )
-  ) {
-    // TODO(NODE-2994): Connection storms work will add new events to connection pool
-    return false;
-  }
+      const allRequirements = runOn.map(legacyRunOnToRunOnRequirement);
+
+      let shouldRun = true;
```
baileympearson, line 156: If you want to be unnecessarily fancy here (which I don't recommend, I just had fun writing this):

```typescript
const id = x => x;
let shoundRun = await Promise.all(
  allRequirements.map((req) => topologySatisfies(this.currentTest.ctx, requirement, utilClient))
).then(satisfied => satisfied.every(id))
```

Is it more readable?  Not sure.  Is it fun?  I think so 😄

nbbeeken: Clever approach! I think it would work the same way but we would get a nondeterministic "reason" depending on which runOn came back with false last. Something to consider 🤔

test/tools/runner/hooks/configuration.js
```diff
@@ -6,12 +6,14 @@ require('source-map-support').install({
...
-const { TestConfiguration } = require('./config');
-const { getEnvironmentalOptions } = require('../utils');
-const mock = require('../mongodb-mock/index');
+const { MongoClient } = require('../../../../src');
+const { TestConfiguration } = require('../config');
+const { getEnvironmentalOptions } = require('../../utils');
+const mock = require('../../mongodb-mock/index');
 const { inspect } = require('util');
 
+const IS_UNIT = process.argv.length >= 3 && process.argv[2].includes('unit');
```
baileympearson, line 15: Maybe we could create an environment variable and feed it into the tests, like `IS_UNIT=true`, instead of relying on parsing the test directory structure?

test/tools/runner/hooks/configuration.js
```diff
@@ -95,15 +102,42 @@ before(async function () {
...
+    version: this.configuration.buildInfo.version,
+    node: process.version,
+    os: process.platform,
+    serverless: process.env.SERVERLESS === '1',
+    auth: process.env.AUTH === 'auth',
+    tls: process.env.SSL === 'ssl',
+    csfle: this.configuration.clientSideEncryption.enabled,
+    serverApi: MONGODB_API_VERSION,
+    atlas: process.env.ATLAS_CONNECTIVITY != null,
+    adl: 'unknown', // TODO
```
baileympearson, line 117: Should we create follow up tickets for these three todos or are these something you will tackle in this PR?

test/tools/spec-runner/context.js
```diff
@@ -87,11 +87,13 @@ class TestRunnerContext {
...
-    return Promise.all([
-      this.runForAllClients(client => client.close()),
-      this.sharedClient.close()
-    ]);
+  async teardown() {
+    await this.runForAllClients(async client => {
+      if (client) await client.close();
+    });
+    if (this.sharedClient) {
+      this.sharedClient.close();
```
baileympearson, line 95: Do we need to `await` here?

test/integration/retryable-writes/retryable_writes.spec.test.js
```diff
@@ -2,47 +2,64 @@
...
+
+        let shouldRun = true;
+        for (const requirement of allRequirements) {
+          shouldRun =
+            shouldRun && (await topologySatisfies(this.currentTest.ctx, requirement, utilClient));
+        }
+
+        await utilClient.close();
+
+        if (!shouldRun) this.skip();
```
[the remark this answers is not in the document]
baileympearson: After reviewing the rest of this PR, I think I answered this myself.  The runOnRequirement checker already adds a skip reason if there is one so I think adding one here isn't necessary 😄

test/tools/spec-runner/index.js
```diff
@@ -92,109 +93,123 @@ function gatherTestSuites(specPath) {
...
 
-    const mongodb = version.join(' ');
-    return {
-      topology,
-      mongodb,
-      authEnabled: !!config.authEnabled,
-      serverless: config.serverless
-    };
-  });
+  if (typeof runOn.serverless === 'string') {
```
baileympearson, line 119: How strict should we be here?  

We could also check that the `serverless` parameter is one of `'forbid' | 'allow' | 'require'` in addition to checking that it's a string.

baileympearson (changes requested):

test/integration/retryable-writes/retryable_writes.spec.test.js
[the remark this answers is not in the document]
nbbeeken: Right, these specs have a slightly different setup so they needed this "almost" copy of the beforeEach hook but it will still call topologySatisfies which calculates a reason.

test/tools/spec-runner/index.js
nbbeeken: Topology satisfies will throw if its an unknown string. My intention here is actually that runOn.serverless is not undefined, I could update the if to that.

In fact this translation function could be simplified to only making the authEnabled -> auth translation, since I think that's the only difference. The code was written like this to avoid undefined to be printed by the metamocha_ui filters. but the unified filter doesn't print out the raw spec, it prints a calculated string

test/tools/spec-runner/context.js
nbbeeken: Yes, good catch!

test/tools/runner/hooks/configuration.js
nbbeeken: Follow ups if we want to keep these. Currently those special environments are kept separate and run on their own variant, there isn't much use to having them stated here, unless we feel otherwise. For example it obvious when we're running against ADL, so is there a need for this?

test/tools/runner/config.ts
```diff
@@ -1,41 +1,66 @@
...
+import { getEnvironmentalOptions } from '../utils';
+
+interface ProxyParams {
+  proxyHost?: string;
+  proxyPort?: number;
+  proxyUsername?: string;
+  proxyPassword?: string;
+}
+
+interface UrlOptions {
```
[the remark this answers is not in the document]
nbbeeken: done!

test/tools/runner/hooks/configuration.js
nbbeeken: This is something we'll want everyone's eyes on, I'm cool with that approach, maybe more clarity like `NO_SERVER=true`. But also folks shouldn't have to do this locally so maybe there's some heuristic we can figure out here.

- In the case of **no** server running: 
  - I want to be able to run `mocha test/unit` without issue so this before hook has to return early.
  - I expect `mocha test/integration` to fail server selection
  - I expect `mocha test/unit test/integration` to either fail server selection, or crash tests that access `this.configuration.x`, since it will be undefined
- In the case of a server running: 
  - I want to be able to run `mocha test/unit test/integration` without issue. <- currently this would fail, cus my logic makes the hook return early, leaving configuration undefined
  - I want to run `mocha test/unit` without issue, in this case it will still connect to the server and define this.configuration it's just that the unit tests don't use it, extra work but no biggie
  - ofc int tests work as expected

baileympearson (approved):

test/tools/runner/hooks/configuration.js
nbbeeken: @dariakp @durran, now that this is in team review could use a meeting of minds here for best approach.
I haven't added an env var like bailey asked yet, that could be done in CI for our lint task, but I wouldn't want to folks to have to do that locally necessarily to run unit tests without a server up, but then again it isn't a huge hurdle.

dariakp: can we instead set up a different mocharc file for the integration directory vs the unit directory?

test/tools/spec-runner/context.js
```diff
@@ -87,11 +87,13 @@ class TestRunnerContext {
...
-  teardown() {
-    return Promise.all([
-      this.runForAllClients(client => client.close()),
-      this.sharedClient.close()
-    ]);
+  async teardown() {
+    await this.runForAllClients(async client => {
+      if (client) await client.close();
+    });
+    if (this.sharedClient) {
```
dariakp, line 94: just a note that with the Promise.all, the first thing failing doesn't prevent the second from executing, but with this setup, if the runForAllClients fails, this sharedClient block won't execute

dariakp (changes requested):
Capturing some feedback that I started in a separate tab

test/tools/unified-spec-runner/runner.ts
```diff
@@ -57,7 +53,7 @@ export async function runUnifiedTest(
   // If test.skipReason is specified, the test runner MUST skip this
   // test and MAY use the string value to log a message.
   if (test.skipReason) {
-    console.warn(`Skipping test ${test.description}: ${test.skipReason}.`);
+    ctx.skipReason = test.skipReason;
```
dariakp, line 56: Does the printout work the same? Or is there logic elsewhere to handle it now?

package.json
```diff
@@ -104,18 +104,18 @@
     "check:eslint": "eslint -v && eslint --max-warnings=0 --ext '.js,.ts' src test",
     "check:tsd": "tsd --version && tsd",
     "check:dts": "./node_modules/typescript/bin/tsc --noEmit mongodb.d.ts && tsd",
-    "check:test": "mocha --file test/tools/runner test/integration",
-    "check:unit": "mocha test/unit/",
+    "check:test": "mocha --config test/mocharc_int.jsonc test/integration",
```
dariakp, line 107: the file seems to be `mocha_mongodb.json`

package.json
```diff
@@ -104,18 +104,18 @@
...
-    "check:csfle": "mocha --file test/tools/runner test/integration/client-side-encryption",
+    "check:atlas": "mocha --config test/manual/mocharc.json test/manual/atlas_connectivity.test.js",
+    "check:adl": "mocha test/manual/atlas-data-lake-testing",
+    "check:aws": "mocha test/integration/auth/mongodb_aws.test.js",
+    "check:ocsp": "mocha --config test/manual/mocharc.json test/manual/ocsp_support.test.js",
+    "check:kerberos": "mocha --config test/manual/mocharc.json test/manual/kerberos.test.js",
+    "check:tls": "mocha --config test/manual/mocharc.json test/manual/tls_support.test.js",
+    "check:ldap": "mocha --config test/manual/mocharc.json test/manual/ldap.test.js",
+    "check:socks5": "mocha --config test/manual/mocharc.json test/manual/socks5.test.ts",
+    "check:csfle": "mocha test/integration/client-side-encryption",
```
dariakp, line 118: does this need a config file flag?

package.json
```diff
@@ -104,18 +104,18 @@
...
-    "check:aws": "mocha --file test/tools/runner test/integration/auth/mongodb_aws.test.js",
-    "check:ocsp": "mocha --config \"test/manual/mocharc.json\" test/manual/ocsp_support.test.js",
-    "check:kerberos": "mocha --config \"test/manual/mocharc.json\" test/manual/kerberos.test.js",
-    "check:tls": "mocha --config \"test/manual/mocharc.json\" test/manual/tls_support.test.js",
-    "check:ldap": "mocha --config \"test/manual/mocharc.json\" test/manual/ldap.test.js",
-    "check:socks5": "mocha --config \"test/manual/mocharc.json\" test/manual/socks5.test.ts",
-    "check:csfle": "mocha --file test/tools/runner test/integration/client-side-encryption",
+    "check:atlas": "mocha --config test/manual/mocharc.json test/manual/atlas_connectivity.test.js",
+    "check:adl": "mocha test/manual/atlas-data-lake-testing",
+    "check:aws": "mocha test/integration/auth/mongodb_aws.test.js",
```
dariakp, line 112: does this need a config file flag?

package.json
```diff
@@ -104,18 +104,18 @@
...
-    "check:adl": "mocha --file test/tools/runner test/manual/atlas-data-lake-testing",
-    "check:aws": "mocha --file test/tools/runner test/integration/auth/mongodb_aws.test.js",
-    "check:ocsp": "mocha --config \"test/manual/mocharc.json\" test/manual/ocsp_support.test.js",
-    "check:kerberos": "mocha --config \"test/manual/mocharc.json\" test/manual/kerberos.test.js",
-    "check:tls": "mocha --config \"test/manual/mocharc.json\" test/manual/tls_support.test.js",
-    "check:ldap": "mocha --config \"test/manual/mocharc.json\" test/manual/ldap.test.js",
-    "check:socks5": "mocha --config \"test/manual/mocharc.json\" test/manual/socks5.test.ts",
-    "check:csfle": "mocha --file test/tools/runner test/integration/client-side-encryption",
+    "check:atlas": "mocha --config test/manual/mocharc.json test/manual/atlas_connectivity.test.js",
+    "check:adl": "mocha test/manual/atlas-data-lake-testing",
```
dariakp, line 111: does this need a config file flag?

.evergreen/run-serverless-tests.sh
```diff
@@ -11,7 +11,7 @@ if [ -z ${MONGODB_URI+omitted} ]; then echo "MONGODB_URI is unset" && exit 1; fi
 if [ -z ${SERVERLESS_ATLAS_USER+omitted} ]; then echo "SERVERLESS_ATLAS_USER is unset" && exit 1; fi
 if [ -z ${SERVERLESS_ATLAS_PASSWORD+omitted} ]; then echo "SERVERLESS_ATLAS_PASSWORD is unset" && exit 1; fi
 
-npx mocha --file test/tools/runner/index.js \
+npx mocha \
```
dariakp, line 14: needs config file flag

dariakp (changes requested):

test/tools/runner/hooks/configuration.js
nbbeeken: Went with the bespoke config option for int tests.

test/manual/atlas-data-lake-testing/atlas_data_lake_testing.spec.js
```diff
@@ -24,5 +24,7 @@ describe('Atlas Data Lake - spec', function () {
     return testContext.setup(this.configuration);
   });
 
+  for (const suite of testSuites) suite.runOn = []; // patched in for the spec runner
```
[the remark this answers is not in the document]
nbbeeken: ADL is the only legacy specs that don't define this array, I added an if to error if the prop wasn't the right type, but I'd rather keep the check and just patch ADL tests that deviate from the norm. But open to the alternative!

test/tools/spec-runner/context.js
nbbeeken: We should be able to change this back with the new way the config file works. This was happening where all the int tests are skipped because there's no mongodb running but some of the after hooks didn't gate their calls the way before checks did

test/tools/unified-spec-runner/runner.ts
nbbeeken: It prints out correctly after the test along with our other skipReasons, printing inside the test like this made it appear before the test title, a little confusing.

.evergreen/run-serverless-tests.sh
nbbeeken: fixed

package.json
nbbeeken: fixed

package.json
nbbeeken: fixed

package.json
nbbeeken: fixed

package.json
nbbeeken: fixed

test/tools/spec-runner/context.js
nbbeeken: Ok actually I simplified it, reading runForAllClients we were calling close on the sharedClient twice

test/manual/atlas-data-lake-testing/atlas_data_lake_testing.spec.js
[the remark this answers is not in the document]
dariakp: I think that's fine actually

dariakp (dismissed):
LGTM :)

PR #3121, opened 2022-01-31
ci(NODE-3860): Enable auth on in CI

baileympearson:
### Description

This PR enables auth for tests running in CI and locally.

#### What is changing?

A few key changes

- Auth is enabled by default locally and in CI
- Many tests were failing when auth was enabled - these have been skipped and a follow up ticket has been created
- A new filter, AuthFilter, was created to filter out tests that require or do not require auth to run
- A handful of tests were fixed instead of skipped.  With these tests, it was either easy to identify the problem or it was easier to fix the issue than to skip the tests (sometimes hundreds of tests would fail for the same reason).

No.


test/tools/runner/index.js
```diff
@@ -12,7 +12,17 @@ const { getEnvironmentalOptions } = require('../utils');
 const mock = require('../mongodb-mock/index');
 const { inspect } = require('util');
 
-const MONGODB_URI = process.env.MONGODB_URI || 'mongodb://localhost:27017';
+// Default our tests to have auth enabled
+process.env.AUTH = process.env.AUTH === 'noauth' ? 'noauth' : 'auth';
```
baileympearson, line 16: I'm open to thoughts here.  With this change I wanted to

1. default tests to running with auth when running locally (or in CI)
2. not require changes to the scripts in package.json or require passing in environment variables

If anyone has a better thought about how to accomplish this, I'm open to suggestions

test/integration/client-side-encryption/client_side_encryption.spec.test.js
```diff
@@ -7,6 +7,52 @@ const {
...
+  'unset works with an encrypted field',
+  'updateOne with deterministic encryption',
+  'updateMany with deterministic encryption',
+  'type=date',
+  'type=regex',
+  'type=timestamp',
+  'type=javascript',
+  'type=binData',
+  'type=int',
+  'type=objectId',
```
nbbeeken, line 42: We wanted to go back and reassess if these tests would get fixed by the sharded uri fix we had to make

test/integration/connection-monitoring-and-pooling/connection.test.js
```diff
@@ -87,7 +87,7 @@ describe('Connection', function () {
       }
     });
 
-    it('should support calling back multiple times on exhaust commands', {
+    it.skip('should support calling back multiple times on exhaust commands', {
       metadata: { requires: { apiVersion: false, mongodb: '>=4.2.0', topology: ['single'] } },
```
nbbeeken, line 91: Should be an auth disabled requirement instead right?

test/integration/connection-monitoring-and-pooling/connection.test.js
```diff
@@ -195,7 +195,7 @@ describe('Connection', function () {
       }
     });
 
-    it('should correctly connect to server using domain socket', {
+    it.skip('should correctly connect to server using domain socket', {
       metadata: { requires: { topology: 'single', os: '!win32' } },
```
nbbeeken, line 199: Should be an auth disabled requirement instead right?

test/integration/enumerate_databases/enumerate_databases.test.ts
```diff
@@ -2,8 +2,15 @@ import { expect } from 'chai';
 
 import { AddUserOptions, MongoClient, MongoServerError } from '../../../src';
 
+const metadata = {
+  requires: {
+    mongodb: '>=4.0.6',
+    auth: 'enabled',
+    topology: ['single', 'replicaset', 'sharded']
+  }
+} as any;
```
nbbeeken, line 11: Let's add MongoDBMetadataUI to the global namespace, and you can `as MongoDBMetadataUI` to get some hinting

test/integration/sessions/sessions.test.js
```diff
@@ -6,13 +6,20 @@ const { setupDatabase, withMonitoredClient } = require('../shared');
...
     this.commands = { started: [], succeeded: [] };
     this.client = config.newClient({ w: 1 }, { maxPoolSize: 1, monitorCommands: true });
 
+    const auth = config.options.auth;
+    const isAuthEnabled = !!(auth && auth.username && auth.password);
     this.client.on('commandStarted', event => {
+      if (event.commandName === 'ping' && isAuthEnabled && !hasInitialPingOccurred) {
+        hasInitialPingOccurred = true;
+        return;
+      }
```
nbbeeken, line 22: Can we leave a comment here referencing why we have to do this, and reference NODE-2149

test/manual/socks5.test.ts
```diff
@@ -17,6 +17,9 @@ describe('Socks5 Connectivity', function () {
     return;
   }
 
+  // 'TODO: NODE-3891 - fix tests broken when AUTH enabled'
+  const maybeIt = process.env.AUTH === 'auth' ? it.skip : it;
```
nbbeeken, line 21: These tests are worth taking a gander at why they're an issue, a socks connection will always be used in an auth scenario, I'm confident that the failure we're experiencing here are a miss-config as opposed to a real issue, any chance you have reference to an error message?

test/tools/cluster_setup.sh
```diff
@@ -10,19 +10,38 @@ DATA_DIR=${DATA_DIR:-data}
...
-elif [[ $1 == "sharded_cluster" ]]; then
-    mkdir -p $SHARDED_DIR
-    mlaunch init --dir $SHARDED_DIR --replicaset --nodes 3 --arbiter --name rs --port 51000 --enableMajorityReadConcern --setParameter enableTestCommands=1 --sharded 1 --mongos 2
-    echo "mongodb://localhost:51000,localhost:51001"
-elif [[ $1 == "server" ]]; then
-    mkdir -p $SINGLE_DIR
-    mlaunch init --dir $SINGLE_DIR --single --setParameter enableTestCommands=1
-    echo "mongodb://localhost:27017"
+if [[ "$AUTH" == "auth" ]]; then
+    if [[ $1 == "replica_set" ]]; then
```
nbbeeken, line 16: I would be okay with the script always turning on auth unconditionally, if I need to go out of my way to disable it then writing out the command is fine. I'm thinking keeping this wrapper script simple is ideal, and it keeps us all mostly on the same page with local testing. thoughts?

test/tools/reporter/mongodb_reporter.js
```diff
@@ -229,10 +229,15 @@ class MongoDBMochaReporter extends mocha.reporters.Spec {
   }
 
   /**
-   * @param {MongoMochaTest & {skipReason?: string}} test
+   * @param {MongoMochaTest & {skipReason?: string} & { metadata?: { skipReason?: string }} } test
```
nbbeeken, line 232: super minor, we could edit the type defined at the top of this file instead, but this just convenience hints, no worries.

test/tools/runner/index.js
nbbeeken: It makes sense why you had to do this to make the auth setting globally available, I think we can defer to a better solution in NODE-3714, if you could reference that in a todo here.

test/tools/spec-runner/index.js
```diff
@@ -169,10 +169,16 @@ function generateTopologyTests(testSuites, testContext, filter) {
 }
 
 function shouldRunSpecTest(configuration, requires, spec, filter) {
-  if (requires.authEnabled && process.env.AUTH !== 'auth') {
+  if (requires.authEnabled) {
     // TODO(NODE-3488): We do not have a way to determine if auth is enabled in our mocha metadata
     // We need to do a admin.command({getCmdLineOpts: 1}) if it errors (code=13) auth is on
```
nbbeeken, line 174: This todo is no longer relvant?

test/tools/spec-runner/index.js
```diff
@@ -169,10 +169,16 @@ function generateTopologyTests(testSuites, testContext, filter) {
...
-    return false;
+    if (process.env.AUTH !== 'auth') {
+      return false;
+    }
+
+    if (spec.failPoint.configureFailPoint === 'failCommand') {
+      return false;
+    }
```
nbbeeken, line 181: Why is this also gated within authEnabled? you should be able to run failCommands even when auth is turned on

test/tools/spec-runner/context.js
```diff
@@ -30,6 +30,29 @@ class Thread {
...
+}
+
+function extractAuthString(connectionString) {
+  const indices = getIndicesOfAuthInUrl(connectionString);
+  if (!indices) {
+    return null;
+  }
+
+  return connectionString.slice(indices.start, indices.end);
+}
```
nbbeeken, line 54: Can we import these from utils now?

test/tools/unified-spec-runner/unified-utils.ts
```diff
@@ -78,7 +78,7 @@ export async function topologySatisfies(
       // but in node you'll run into auth errors / successes at client.connect() time.
       // so we cannot run into saslContinue failPoints that get configured for an operation to fail with
       // Ex. 'errors during authentication are processed' in test/spec/load-balancers/sdam-error-handling.yml
-      ok &&= false; // process.env.AUTH === 'auth';
+      ok &&= process.env.AUTH === 'auth';
```
nbbeeken, line 81: The proceeding TODO is no longer relevant 🙂  yahoo!!

nbbeeken (changes requested):

test/tools/cluster_setup.sh
baileympearson: Yup, I think that's reasonable!

test/tools/runner/index.js
baileympearson: Done!

test/tools/reporter/mongodb_reporter.js
baileympearson: done done!

test/manual/socks5.test.ts
baileympearson: I definitely was seeing failures before but now I don't, for some reason.  Maybe I fixed it somewhere else?  Regardless, I re-enabled them and ran a patch [here](https://spruce.mongodb.com/version/61facc97d6d80a4fb31284ad/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and they're passing so I reverted this change.

test/integration/connection-monitoring-and-pooling/connection.test.js
baileympearson: yes good catch!

test/integration/client-side-encryption/client_side_encryption.spec.test.js
baileympearson: Screen Shot 2022-02-02 at 2 37 04 PM

Our fix to the test runner didn't fix these tests.  These tests seem to be failing with an "extra ping command".  I can dig into these and see if I can find a fix but I'd rather wait and fix them as a part of the follow up ticket

nbbeeken: Ah ok yea at least we confirmed its not about the content of the test, I'm fine with defering

nbbeeken (approved):

test/tools/runner/filters/auth_filter.js
```diff
@@ -0,0 +1,36 @@
...
+    const auth = test.metadata.requires.auth;
+
+    if (auth === 'enabled') {
+      return this.isAuthEnabled;
+    } else if (auth === 'disabled') {
+      return !this.isAuthEnabled;
+    }
+
+    // defaults to 'disabled' because it's gotta default to something.
+    return true;
```
dariakp, line 32: could we throw instead so that we catch typos/unsupported values?

dariakp (changes requested):

test/integration/enumerate_databases/enumerate_databases.test.ts
```diff
@@ -2,8 +2,15 @@ import { expect } from 'chai';
 
 import { AddUserOptions, MongoClient, MongoServerError } from '../../../src';
 
+const metadata: MongoDBMetadataUI = {
+  requires: {
+    mongodb: '>=4.0.6',
+    auth: 'enabled',
+    topology: ['single', 'replicaset', 'sharded']
```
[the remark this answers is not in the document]
nbbeeken: +1 to Daria's comment I think we can remove the limiting altogether

But if you don't mind looks like I didn't get the types quite right for the topology, you could throw this in global.d.ts to improve the hinting  
```ts
import { OneOrMore } from './src/mongo_types';
type TopologyTypeRequirement = OneOrMore<TopologyTypes> | OneOrMore<WithExclusion<TopologyTypes>>
```

test/integration/node-specific/mongo_client.test.js
```diff
@@ -37,12 +37,13 @@ describe('MongoClient integration', function () {
               return 1;
             }
           },
-          serializeFunctions: true,
-          raw: true
+          serializeFunctions: true
```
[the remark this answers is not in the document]
nbbeeken: The raw option has issues when passed to the MongoClient, enabling auth revealed it, but this test also doesn't need to use it, so there's not really a TODO to leave behind. related: NODE-3946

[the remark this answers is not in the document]
dariakp: Ok, sure

test/integration/connection-monitoring-and-pooling/connection.test.js
```diff
@@ -88,7 +88,10 @@ describe('Connection', function () {
     });
 
     it('should support calling back multiple times on exhaust commands', {
-      metadata: { requires: { apiVersion: false, mongodb: '>=4.2.0', topology: ['single'] } },
+      metadata: {
+        requires: { apiVersion: false, mongodb: '>=4.2.0', topology: ['single'], auth: 'disabled' },
```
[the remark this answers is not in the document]
baileympearson: Yeah, it's skipped right now because it's failing when auth is enabled.  Would you want me to handle it differently?

test/tools/spec-runner/utils.js
```diff
@@ -1,18 +1,42 @@
...
+}
+
 function resolveConnectionString(configuration, spec, context) {
   const isShardedEnvironment = configuration.topologyType === 'Sharded';
   const useMultipleMongoses = spec && !!spec.useMultipleMongoses;
   const username = context && context.user;
   const password = context && context.password;
-  const authSource = context && context.authSource;
+  const authSource =
+    context && context.authSource && context.authSource !== 'admin' ? context.authSource : 'admin';
```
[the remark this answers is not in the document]
baileympearson: Because some tests specifically do pass in a value for `authSource` and if that value is not `admin`, we don't want to override it.  We just want to default to `admin`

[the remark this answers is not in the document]
dariakp: is that line equivalent to `context?.authSource || 'admin'`?

test/tools/reporter/mongodb_reporter.js
```diff
@@ -229,10 +229,15 @@ class MongoDBMochaReporter extends mocha.reporters.Spec {
   }
 
   /**
-   * @param {MongoMochaTest & {skipReason?: string}} test
+   * @param {MongoMochaTest} test
    */
   pending(test) {
     if (REPORT_TO_STDIO) console.log(chalk.cyan(`↬ ${test.fullTitle()}`));
+    if (test.metadata && test.metadata.skipReason && typeof test.metadata.skipReason === 'string') {
```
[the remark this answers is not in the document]
baileympearson: I added this specifically for the deadlock tests, which all share a metadata object but all needed to be skipped.  Rather than adding the skipReason for each test individually, I skipped the whole suite by attaching it to the metadata. 

I actually prefer the metadata approach because it keeps all data related to why a test may be skipped together but don't feel strongly about this.  If you'd prefer to have a single method for attaching skip reasons to tests, I can revert this change

test/tools/spec-runner/utils.js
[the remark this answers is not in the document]
baileympearson: Yeah, you could also use an `||`.  I can make that change.  

Unfortunately we can't use optional chaining in our raw `.js` tests because we still test on Node v12 and optional chaining wasn't added until v13

test/tools/spec-runner/index.js
```diff
@@ -151,6 +151,16 @@ function generateTopologyTests(testSuites, testContext, filter) {
         shouldRun = false;
       }
 
+      // check auth requirements
+      for (const requirement of allRequirements) {
+        const { auth } = requirement;
+        if (auth) {
+          if (process.env.AUTH !== 'auth') {
+            shouldRun = false;
```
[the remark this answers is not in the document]
baileympearson: I think this is actually unnecessary, since `isAnyRequirementSatisfied` already checks for `auth` requirements.

.evergreen/config.yml
```diff
@@ -2020,3 +2149,13 @@ buildvariants:
       NODE_LTS_NAME: erbium
     tasks:
       - serverless_task_group
+  - name: no-auth-tests
```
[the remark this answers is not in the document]
nbbeeken: Slack convo capture: we want to expand the noauth variants to one of each topology, only 5.0 version

test/integration/connection-monitoring-and-pooling/connection.test.js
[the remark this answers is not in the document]
nbbeeken: I'm resolving the comments pointing out tests that would get skipped, our slack convo covered what topologies we want to see run w/ noauth.

test/integration/transactions/transactions.spec.test.js
```diff
@@ -93,7 +93,10 @@ const SKIP_TESTS = [
 
   // TODO(NODE-2034): Will be implemented as part of NODE-2034
   'Client side error in command starting transaction',
-  'Client side error when transaction is in progress'
+  'Client side error when transaction is in progress',
+
+  // 'TODO: NODE-3891 - fix tests broken when AUTH enabled'
+  'implicit abort'
```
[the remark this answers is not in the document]
dariakp: I'm not sure I understand the comment regarding "with auth is enabled" - do you mean it's failing on both auth AND noauth? If it's only failing on one or the other, we should still let it run on whichever configuration option is supported by adding it conditionally

test/tools/spec-runner/utils.js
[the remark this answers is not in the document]
dariakp: ah yes, good call out on the chaining

test/integration/transactions/transactions.spec.test.js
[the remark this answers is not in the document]
baileympearson: Yes, it fails when auth is enabled and when it is disabled

test/integration/transactions/transactions.spec.test.js
```diff
@@ -90,7 +90,14 @@ const SKIP_TESTS = [
 
   // TODO(NODE-2034): Will be implemented as part of NODE-2034
   'Client side error in command starting transaction',
-  'Client side error when transaction is in progress'
+  'Client side error when transaction is in progress',
+
+  // 'TODO: NODE-3891 - fix tests broken when AUTH enabled'
+  //    Unlike most of the failing auth tests, this is failing with auth is enabled.
```
dariakp, line 96:
```suggestion
  //    Unlike most of the failing auth tests, this is failing with and without auth enabled.
```

test/tools/spec-runner/index.js
```diff
@@ -390,18 +390,13 @@ function runTestSuiteTest(configuration, spec, context) {
     let testPromise = Promise.resolve();
     return testPromise
       .then(() => testOperations(spec, operationContext))
-      .catch(err => {
-        // If the driver throws an exception / returns an error while executing this series
-        // of operations, store the error message.
-        throw err;
-      })
-      .then(() => {
+      .then(() => validateExpectations(context.commandEvents, spec, savedSessionData))
```
dariakp, line 393: we probably want to reverse this order, since that's what is making the spec tests here fail; does reversing the order back to what it was cause other tests to fail?

dariakp (changes requested):

test/tools/reporter/mongodb_reporter.js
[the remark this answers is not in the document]
dariakp: I think having multiple ways to use a feature makes that feature more brittle (i.e., higher chance of breaking without noticing in one place or the other), so my personal preference would be to avoid it; but I appreciate the convenience and the visibility afforded by having it in metadata, so if you don't find the risk of the brittleness compelling here, I'm fine with leaving it as is

test/manual/tls_support.test.js
```diff
@@ -21,7 +22,9 @@ describe('TLS Support', function () {
     makeConnectionTest(connectionString, tlsSettings)
   );
 
-  it(
+  // 'TODO: NODE-3891 - fix tests broken when AUTH enabled'
+  const maybeIt = AUTH_ENABLED ? it.skip : it;
```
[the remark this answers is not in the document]
dariakp: Did we confirm that this test runs fine in CI now?

test/tools/reporter/mongodb_reporter.js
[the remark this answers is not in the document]
nbbeeken: What about keeping the `skipReason` made by the metadata filters and appending additional context. "skipped because of an auth disabled requirement" is still useful to see, but then we could have a `skipContext` or `metaSkipReason` (or maybe keep it simple `note: 'xxx'`) that if defined appends more to the message so that the logs reflect both, skipped cus auth and cus of a TODO.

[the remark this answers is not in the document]
dariakp: The funny thing about this logic is that it's going to append that reason right now no matter what the cause of the skip is, so, for example, when they don't run on single topologies, we'd still get the same metadata skipReason, if I am reading this correctly, since the skipReason isn't attached to a specific filter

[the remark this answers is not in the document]
nbbeeken: Okay I think that's a valid point and that could lead to confusion, we have the skipReason telling us the reason was auth and we'll keep the TODO to track the unskipping for the particular environment with no auth enabled. I'm okay with that.

nbbeeken (approved):

durran (approved):

dariakp (approved):
This is fine to merge, as long as we have a follow up PR on the same ticket to clean up the skipReason logic

PR #3122, opened 2022-02-01
feat(NODE-3351): use hostname canonicalization 

durran:
### Description

Sets the driver to use the proper `CANONICALIZE_HOST_NAME` option when using Kerberos.

#### What is changing?

Removes the old check of `gssapiCanonicalizeHostName` to use the proper option. This was dead code as it was not a public member of our `AuthMechanismProperties` and was not parsed by the URI.

No

NODE-3351

### DIFF
--- a/src/cmap/auth/gssapi.ts
+++ b/src/cmap/auth/gssapi.ts
@@ -13,7 +13,9 @@ import { Callback, ns } from '../../utils';
 import { AuthContext, AuthProvider } from './auth_provider';
 
 type MechanismProperties = {
+  // TODO: Remove in 5.0
   gssapiCanonicalizeHostName?: boolean;
+  CANONICALIZE_HOST_NAME?: boolean;
   SERVICE_NAME?: string;
   SERVICE_REALM?: string;
 };
@@ -174,7 +176,10 @@ function performGssapiCanonicalizeHostName(
   mechanismProperties: MechanismProperties,
   callback: Callback<string>
 ): void {
-  if (!mechanismProperties.gssapiCanonicalizeHostName) return callback(undefined, host);
+  if (
+    !mechanismProperties.gssapiCanonicalizeHostName &&
+    !mechanismProperties.CANONICALIZE_HOST_NAME
+  ) return callback(undefined, host);
 
   // Attempt to resolve the host name
   dns.resolveCname(host, (err, r) => {
--- a/src/cmap/auth/gssapi.ts
+++ b/src/cmap/auth/gssapi.ts
@@ -13,6 +13,8 @@ import { Callback, ns } from '../../utils';
 import { AuthContext, AuthProvider } from './auth_provider';
 
 type MechanismProperties = {
+  /** @deprecated use `CANONICALIZE_HOST_NAME` instead */
+  gssapiCanonicalizeHostName?: boolean;
   CANONICALIZE_HOST_NAME?: boolean;
   SERVICE_NAME?: string;
   SERVICE_REALM?: string;
--- a/src/cmap/auth/gssapi.ts
+++ b/src/cmap/auth/gssapi.ts
@@ -13,8 +13,6 @@ import { Callback, ns } from '../../utils';
 import { AuthContext, AuthProvider } from './auth_provider';
 
 type MechanismProperties = {
-  // TODO: Remove in 5.0
-  gssapiCanonicalizeHostName?: boolean;
   CANONICALIZE_HOST_NAME?: boolean;
   SERVICE_NAME?: string;
   SERVICE_REALM?: string;
@@ -176,10 +174,7 @@ function performGssapiCanonicalizeHostName(
   mechanismProperties: MechanismProperties,
   callback: Callback<string>
 ): void {
-  if (
-    !mechanismProperties.gssapiCanonicalizeHostName &&
-    !mechanismProperties.CANONICALIZE_HOST_NAME
-  ) return callback(undefined, host);
+  if (!mechanismProperties.CANONICALIZE_HOST_NAME) return callback(undefined, host);
 
   // Attempt to resolve the host name
   dns.resolveCname(host, (err, r) => {
--- a/src/cmap/auth/mongo_credentials.ts
+++ b/src/cmap/auth/mongo_credentials.ts
@@ -1,7 +1,7 @@
 // Resolves the default auth mechanism according to
-
 import type { Document } from '../../bson';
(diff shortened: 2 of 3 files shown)


test/manual/kerberos.test.js
```diff
@@ -51,7 +51,16 @@ describe('Kerberos', function () {
...
+    const client = new MongoClient(
+      `${krb5Uri}&authMechanismProperties=SERVICE_NAME:mongodb,CANONICALIZE_HOST_NAME:true&maxPoolSize=1`
+    );
+    client.connect(function (err, client) {
+      expect(err).to.not.exist;
+      verifyKerberosAuthentication(client, done);
+    });
+  });
+
   it.skip('validate that SERVICE_REALM and CANONICALIZE_HOST_NAME can be passed in', function (done) {
```
dariakp, line 64: I think the skip comment was accidentally deleted, we should keep it (or better yet put it in a `skipReason`)

test/manual/kerberos.test.js
```diff
@@ -51,7 +51,16 @@ describe('Kerberos', function () {
     });
   });
 
-  // Unskip this test when a proper setup is available - see NODE-3060
+  it('validate that CANONICALIZE_HOST_NAME can be passed in', function (done) {
+    const client = new MongoClient(
+      `${krb5Uri}&authMechanismProperties=SERVICE_NAME:mongodb,CANONICALIZE_HOST_NAME:true&maxPoolSize=1`
+    );
+    client.connect(function (err, client) {
+      expect(err).to.not.exist;
```
dariakp, line 59: I know this is in line with the rest of the tests, but I'm wondering if it would be better to do `if (err) { return done(err) }` instead, or even better, specify this as an async function and avoid the callback nesting altogether

dariakp (changes requested):
Just a couple minor things here

test/manual/kerberos.test.js
durran: I've brought the comment back now.

test/manual/kerberos.test.js
dariakp: we may need to keep the done callback in the `it` function definition, since right now the verifyKerberosAuthentication execution won't call done and it's not written as a promise; if that's too much of a headache or too weird to rewrite, maybe we just keep the old syntax for the test

dariakp (approved):

src/cmap/auth/gssapi.ts
```diff
@@ -174,7 +174,7 @@ function performGssapiCanonicalizeHostName(
   mechanismProperties: MechanismProperties,
   callback: Callback<string>
 ): void {
-  if (!mechanismProperties.gssapiCanonicalizeHostName) return callback(undefined, host);
+  if (!mechanismProperties.CANONICALIZE_HOST_NAME) return callback(undefined, host);
```
nbbeeken, line 177: I did a little test script (below) and this line is reachable with the `gssapiCanonicalizeHostName` option specified. So this isn't dead code, I think we need to continue supporting the old option but we could assign it to the new one inside the credentials class or something? Even though it wasn't in our types it did work in 3.x and still works in 4.x.

```ts
import { MongoClient } from '.';

const client = new MongoClient(
  'mongodb://u@localhost?authMechanism=GSSAPI&authMechanismProperties=SERVICE_NAME:mongodb,gssapiCanonicalizeHostName:true&maxPoolSize=1'
);

async function main(args) {
  await client.connect();
}

main(process.argv)
  .then(console.log)
  .catch(console.error)
  .finally(() => client.close());
```

test/manual/kerberos.test.js
```diff
@@ -51,6 +51,16 @@ describe('Kerberos', function () {
...
   });
 
+  it('validate that CANONICALIZE_HOST_NAME can be passed in', function (done) {
+    const client = new MongoClient(
+      `${krb5Uri}&authMechanismProperties=SERVICE_NAME:mongodb,CANONICALIZE_HOST_NAME:true&maxPoolSize=1`
+    );
+    client.connect(function (err, client) {
+      if (err) return done(err);
+      verifyKerberosAuthentication(client, done);
+    });
```
nbbeeken, line 61: `verifyKerberosAuthentication` just asserts that we can connect and run an operation, is there more we could do to check that canonicalization was performed? sinon should be able to listen to `dns.resolveCname` maybe?

nbbeeken (changes requested):

src/cmap/auth/gssapi.ts
dariakp: Maybe we should at least mark it deprecated then? We're gearing up for a minor release anyway, shouldn't be a big deal

durran: I've put it back and marked as deprecated and added a test to verify it still works when set.

test/manual/kerberos.test.js
durran: I've added a spy in these tests now to verify the method was called.

src/cmap/auth/mongo_credentials.ts
```diff
@@ -89,6 +89,11 @@ export class MongoCredentials {
       }
     }
 
+    if ('gssapiCanonicalizeHostName' in this.mechanismProperties) {
+      this.mechanismProperties.CANONICALIZE_HOST_NAME =
```
nbbeeken, line 93: Let's add an `emitWarningOnce` here to maximize awareness

nbbeeken (changes requested):

src/cmap/auth/mongo_credentials.ts
durran: Added.

nbbeeken (dismissed):

dariakp (approved):
LGTM!

test/manual/kerberos.test.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 const { MongoClient } = require('../../src');
 const chai = require('chai');
+const sinon = require('sinon');
```
dariakp, line 4: Since the manual tests don't use the same config as the general integration tests, we need to include this line: `chai.use(require('sinon-chai'));` to be able to use the assertions that are currently failing in the CI

dariakp (changes requested):

test/manual/kerberos.test.js
durran: Updated.

dariakp (approved):

PR #3124, opened 2022-02-01
test(NODE-3719): spec review cleanup

dariakp:
### Description
NODE-3719

#### What is changing?
- Reorganized auth tests to separate out prose
- Separated convenient api tests from the transactions tests (both specs and test files)
- Updated TODO references to point to the correct tickets
- Updated unified and legacy runner to check correctly against multiple requirements ("Dirty explicit session is discarded" test used to be skipped on sharded AND replicaset topologies even though it's supposed to run on both)
- Removed the last vestiges of metadata in unit tests
- Replaced all instances of `runUnifiedTest` with `runUnifiedSuite`

**ATTENTION: Reviewers, please verify test counts against main and make sure we are skipping fewer tests than we were before, not more**

N/A
Spec compliance review


test/integration/command-monitoring/command_monitoring.spec.test.js
```diff
@@ -259,6 +260,7 @@ describe('Command Monitoring spec tests', function () {
             metadata: { requires: requirements },
             test: function () {
               if (
+                // TODO(NODE-3308): update spec file
```
nbbeeken, line 263: Wanna do `this.skipReason = xx` here?

test/integration/command-monitoring/command_monitoring.spec.test.js
```diff
@@ -239,7 +240,7 @@ describe('Command Monitoring spec tests', function () {
     }
 
     loadSpecTests('command-monitoring/legacy').forEach(scenario => {
-      if (scenario.name === 'command') return; // FIXME(NODE-3074): remove when `count` spec tests have been fixed
+      if (scenario.name === 'command') return; // TODO(NODE-3369): remove when `count` spec tests have been fixed
```
nbbeeken, line 243: @baileympearson, this might be relevant to your count PR

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -23,7 +23,7 @@ describe('Retryable Reads', function () {
       spec.description.match(/listCollections/i) ||
       spec.description.match(/listCollectionNames/i) ||
       spec.description.match(/estimatedDocumentCount/i) ||
-      // FIXME(NODE-3074): uncomment when `count` spec tests have been fixed
+      // FIXME(NODE-3369): uncomment when `count` spec tests have been fixed
       // spec.description.match(/count/i) ||
```
nbbeeken, line 27: Do these run successfully now?, or is this an inclusion list?

test/integration/sessions/sessions.spec.test.js
```diff
@@ -44,9 +44,7 @@ describe('Sessions', function () {
 
     function testFilter(spec) {
```
nbbeeken, line 45: I say we just remove this, since its unused

test/unit/assorted/optional_require.test.js
```diff
@@ -53,24 +53,17 @@ describe('optionalRequire', function () {
     });
   });
 
-  context('aws4', {
-    metadata: {
-      requires: {
-        mongodb: '>=4.4'
+  context('aws4', function () {
+    it('should error if not installed', function () {
```
nbbeeken, line 57: Thanks for removing the metadata stuff from the unit tests! Is this all of them? it should be safe to remove the `"ui"` setting from the root `.mocharc.json` if so, then future tests will error if metadata is accidentally reintroduced.

nbbeeken (changes requested):
Some clean up nits, looking good

test/integration/retryable-reads/retryable_reads.spec.test.js
dariakp: inclusion

test/unit/assorted/optional_require.test.js
dariakp: I think so

test/integration/command-monitoring/command_monitoring.spec.test.js
```diff
@@ -278,18 +280,6 @@ describe('Command Monitoring spec tests', function () {
...
-          it(String(test.description), {
-            metadata: { sessions: { skipLeakTests: true } },
-            test: async function () {
-              await runUnifiedTest(this, loadedSpec, test);
-            }
-          });
-        }
-      });
-    }
+    runUnifiedSuite(loadSpecTests(path.join('command-monitoring', 'unified')));
```
baileympearson, line 283: just curious, is there any benefit to using `path.join` in place of `command-monitoring/unified` as a string?

test/integration/retryable-reads/retryable_reads.spec.test.js
baileympearson: I was _just_ looking at this exact line on my other PR and wondering why NODE-3074 was linked here 😄

test/unit/sdam/topology.test.js
```diff
@@ -34,20 +34,16 @@ describe('Topology (unit)', function () {
...
-
-      test: function (done) {
-        // Attempt to connect
-        var server = new Topology([`localhost:27017`], {
-          metadata: makeClientMetadata({
-            appName: 'My application name'
-          })
-        });
+    it('should correctly pass appname', function (done) {
+      // Attempt to connect
```
baileympearson, line 38: can we remove this useless comment?

test/tools/spec-runner/index.js
```diff
@@ -132,12 +132,17 @@ function generateTopologyTests(testSuites, testContext, filter) {
...
-          shouldRun && (await topologySatisfies(this.currentTest.ctx, requirement, utilClient));
+        const met = await topologySatisfies(this.currentTest.ctx, requirement, utilClient);
+        if (met) {
+          someRequirementMet = true;
+          break;
+        }
       }
 
+      let shouldRun = someRequirementMet;
+
```
baileympearson, line 145: same comment as in `unified-spec-runner/runner` - abstracting this to a function would make this a lot easier to understand.  If we moved the loop into a function, we could write it like this

```typescript
const someRequirementMet = allRequirements.length > 0 && await isAnyRequirementMet(....) 
```

test/integration/command-monitoring/command_monitoring.spec.test.js
dariakp: this used to be best practice for specifying paths because of how paths resolved on windows vs linux systems, I honestly don't know if the underlying logic here works automagically nowadays to make them resolve fine on windows even with the forward slash

dariakp: it looks like loadspectests feeds it into a path.resolve, so that also normalizes it, which means we could get away with not using path.join, but I feel like this is one of those better safe than sorry things

test/integration/sessions/sessions.spec.test.js
```diff
@@ -3,11 +3,11 @@
...
 const { expect } = require('chai');
 const { TestRunnerContext, generateTopologyTests } = require('../../tools/spec-runner');
-const { runUnifiedTest } = require('../../tools/unified-spec-runner/runner');
+const { runUnifiedSuite } = require('../../tools/unified-spec-runner/runner');
 const { loadSpecTests } = require('../../spec');
 
-describe('Sessions', function () {
-  describe('legacy spec tests', function () {
+describe('Sessions spec tests', function () {
+  describe('Sessions legacy spec tests', function () {
```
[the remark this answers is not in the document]
dariakp: I think this goes back to what @nbbeeken said about having repeated blocks with the same names even if they are in different files, if it doesn't cause an issue for the runner, then yeah, we can remove it

test/integration/command-monitoring/command_monitoring.spec.test.js
nbbeeken: Windows supports both directions of slash these days, but you never know! I appreciate a cross platform abstraction when I see one.

test/integration/sessions/sessions.spec.test.js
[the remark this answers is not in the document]
nbbeeken: All our top level suites need to have unique names in order to generate the final xunit file. (I think this is actually just a limitation I introduced by using a Map in the reporter, so we could circle back on that) It uses `suite.fullTitle()` to identify each suite, so every nested suite will have its parent's name prefixed, so even if there's multiple `legacy spec tests` strings, as long as their parent's are different we're fine. 

In fact, (don't wanna bloat more sorry, but) https://github.com/mongodb/node-mongodb-native/blob/main/test/tools/reporter/mongodb_reporter.js#L178 we could change this warn to throw now, when I first wrote the reporter we had many dupes, but we've since cleared them up in follow up work

test/tools/unified-spec-runner/runner.ts
```diff
@@ -89,19 +89,35 @@ export async function runUnifiedTest(
...
+      for (const requirement of requirements) {
+        const met = await topologySatisfies(ctx, requirement, utilClient);
+        if (met) {
+          if (reqType === 'suite') {
+            someSuiteRequirementMet = true;
+          } else {
+            someTestRequirementMet = true;
+          }
+          break;
+        }
```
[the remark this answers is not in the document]
nbbeeken: We actually have one more instance of incorrect and-ing: https://github.com/mongodb/node-mongodb-native/blob/NODE-3719/spec-review-cleanup/test/integration/retryable-writes/retryable_writes.spec.test.js#L25-L31

The legacy spec tests don't have a per test runOn, only one file wide would it be too much logic to combine the difference between the two?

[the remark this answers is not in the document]
dariakp: Sure, never say no to cleaner, more reusable logic :)

nbbeeken (approved):

baileympearson (approved):

PR #3126, opened 2022-02-03
fix(NODE-3878): use legacy count operation on collection.count

baileympearson:
#### What is changing?

This PR updates `Collection.count` to use the legacy count operation.  It also unskips tests that were failing when `Collection.count` was modified to use the new count operation.


nbbeeken (approved):
LGTM thanks!

cc: @addaleax Just want to this change on your radar, it's an (overdue) revert. The deprecated count method should've continued to use the deprecated count operation, there is no visible user facing changes, only performance and accuracy depending on deployment. It may be worth surfacing the change in shell change notes, but not essential

PR #3129, opened 2022-02-04
feat(NODE-2939): update hostname canonicalization opts

durran:
### Description

Updates `CANONICALIZE_HOST_NAME` options to accept `true`, `false`, "none", "forward", "forwardAndReverse".

#### What is changing?

When canonicalizing the host name when using GSSAPI, the driver will now behave as follows for these values:
- `true` or "forwardAndReverse": Performs a forward DNS lookup of the host and a reverse lookup of the IP address to obtain the hostname. If the reverse lookup fails the driver falls back to a cname lookup.
- "forward": Performs a cname lookup of the host.
- `false` or "none": Does no hostname canonicalization.

When syncing the spec tests some other auth spec tests we hadn't synced yet were there, so I decided to leave them and skip the single failure.

This PR is branched off https://github.com/mongodb/node-mongodb-native/pull/3123 so needs to go in/be reviewed after that.

None

DRIVERS-1803 / NODE-2939

### DIFF
--- a/src/cmap/auth/gssapi.ts
+++ b/src/cmap/auth/gssapi.ts
@@ -13,7 +13,9 @@ import { Callback, ns } from '../../utils';
 import { AuthContext, AuthProvider } from './auth_provider';
 
 type MechanismProperties = {
+  // TODO: Remove in 5.0
   gssapiCanonicalizeHostName?: boolean;
+  CANONICALIZE_HOST_NAME?: boolean;
   SERVICE_NAME?: string;
   SERVICE_REALM?: string;
 };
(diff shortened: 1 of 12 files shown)


src/cmap/auth/gssapi.ts
```diff
@@ -174,14 +179,41 @@ function performGssapiCanonicalizeHostName(
...
+  }
+  // The case for forward is just to resolve the cname as dns.lookup()
+  // will not return it.
+  resolveCname(host, callback);
+}
 
+function resolveCname(host: string, callback: Callback<string>): void {
   // Attempt to resolve the host name
   dns.resolveCname(host, (err, r) => {
     if (err) return callback(err);
```
addaleax, line 213: I think this should also be

```suggestion
    if (err) return callback(undefined, host);
```

because Node.js gives you `ENODATA` errors for domains which exist but have no CNAME records attached

PR #3130, opened 2022-02-04
feat(NODE-2938): add service host mechanism property

durran:
### Description

Adds a `SERVICE_HOST` property to auth mechanisms for the kerberos auth case where a different host in the service principal name should be provided instead of the resolved host.

#### What is changing?

Adds a `SERVICE_HOST` property and updates the tests to include it.

This branch is rebased against https://github.com/mongodb/node-mongodb-native/pull/3122

None

NODE-2938

### DIFF
--- a/test/manual/kerberos.test.js
+++ b/test/manual/kerberos.test.js
@@ -28,6 +28,7 @@ describe('Kerberos', function () {
     return;
   }
   let krb5Uri = process.env.MONGODB_URI;
+  const parts = krb5Uri.split('@', 2);
 
   if (!process.env.KRB5_PRINCIPAL) {
     console.error('skipping Kerberos tests, KRB5_PRINCIPAL environment variable is not defined');
@@ -39,7 +40,6 @@ describe('Kerberos', function () {
     if (process.env.LDAPTEST_PASSWORD == null) {
       throw new Error('The env parameter LDAPTEST_PASSWORD must be set');
     }
-    const parts = krb5Uri.split('@', 2);
     krb5Uri = `${parts[0]}:${process.env.LDAPTEST_PASSWORD}@${parts[1]}`;
   }
 
@@ -59,6 +59,37 @@ describe('Kerberos', function () {
     client.connect(function (err, client) {
       expect(err).to.not.exist;
       verifyKerberosAuthentication(client, done);
+    });
+  });
+
+  context('when passsing SERVICE_HOST', function () {
+    context('when the SERVICE_HOST is invalid', function () {
+      const client = new MongoClient(`${krb5Uri}&maxPoolSize=1`, {
+        authMechanismProperties: {
+          SERVICE_HOST: 'example.com'
+        }
+      });
+
+      it('fails to authenticate', function () {
+        return client.connect().catch(e => {
+          expect(e).to.exist;
+        });
+      });
+    });
+
+    context('when the SERVICE_HOST is valid', function () {
+      const client = new MongoClient(`${krb5Uri}&maxPoolSize=1`, {
+        authMechanismProperties: {
+          SERVICE_HOST: 'ldaptest.10gen.cc'
+        }
+      });
+
+      it('authenticates', function (done) {
+        client.connect(function (err, client) {
+          expect(err).to.not.exist;
+          verifyKerberosAuthentication(client, done);
+        });
+      });
     });
   });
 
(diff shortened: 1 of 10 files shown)


test/manual/kerberos.test.js
```diff
@@ -97,6 +97,37 @@ describe('Kerberos', function () {
     });
   });
 
+  context('when passsing SERVICE_HOST', function () {
```
dariakp, line 100: was this supposed to be "when parsing"?

test/manual/kerberos.test.js
```diff
@@ -97,6 +97,37 @@ describe('Kerberos', function () {
...
+    context('when the SERVICE_HOST is invalid', function () {
+      const client = new MongoClient(`${krb5Uri}&maxPoolSize=1`, {
+        authMechanismProperties: {
+          SERVICE_HOST: 'example.com'
+        }
+      });
+
+      it('fails to authenticate', function () {
+        return client.connect().catch(e => {
+          expect(e).to.exist;
```
dariakp, line 110: should we check for a more specific error here? also, this logic would actually pass if the connect is successful and the catch never executes - we probably want to do something like 
```
let expectedError;
await client.connect().catch(e => { expectedError = e });
if (!e) {
  expect.fail('Expected connect to fail');
}
expect(e).property(message).to.. // some assertion on the error message or error class
```

dariakp (changes requested):
Just a small test adjustment here

test/manual/kerberos.test.js
nbbeeken: Nice suggestion I like how clean this is, if we can assert on the property existence at least, (message match would be great) then this is pretty short and sweet, and it should catch the case where there is no error.
```js
let expectedError; 
await client.connect().catch(e => { expectedError = e });
expect(expectedError).to.have.property('message', /x/);
```

nbbeeken (approved):
I agree with Daria's nits, but LGTM otherwise

test/manual/kerberos.test.js
nbbeeken: I think this pattern is worth documenting in our `test/readme.md`! 💡

durran: Typo and test updated.

test/manual/kerberos.test.js
durran: Changed to "when passing SERVICE_HOST as an auth mech option"

dariakp (dismissed):

dariakp (approved):

baileympearson (dismissed):

test/manual/kerberos.test.js
```diff
@@ -97,6 +97,42 @@ describe('Kerberos', function () {
...
+
+      it('fails to authenticate', async function () {
+        let expectedError;
+        await client.connect().catch(e => {
+          expectedError = e;
+        });
+        if (!expectedError) {
+          expect.fail('Expected connect with invalid SERVICE_HOST to fail');
+        }
+        expect(expectedError).property('message').to.include('Authentication failed');
```
dariakp, line 116: This test failed with a different error in CI - does it fail with this error locally?

dariakp (changes requested):

test/manual/kerberos.test.js
durran: Oh I had my local env setup wrong. I've updated now to check regex as the macos and linux error messages are slightly different.

dariakp (approved):

test/manual/kerberos.test.js
```diff
@@ -97,6 +101,43 @@ describe('Kerberos', function () {
...
+
+      it('fails to authenticate', async function () {
+        let expectedError;
+        await client.connect().catch(e => {
+          expectedError = e;
+        });
+        if (!expectedError) {
+          expect.fail('Expected connect with invalid SERVICE_HOST to fail');
+        }
+        console.log(expectedError);
```
dariakp, line 120: stray console log here

test/manual/kerberos.test.js
```diff
@@ -65,6 +65,8 @@ describe('Kerberos', function () {
   });
 
   it('validate that gssapiCanonicalizeHostName can be passed in', function (done) {
+    // CNAME resolution on evg macos boxes resolves to a ec2 CNAME which is not associated with the KDC.
+    if (process.platform === 'darwin') this.skip();
```
dariakp, line 69: can we do a skipReason for both of the new skips here? also, just for my benefit, does this mean this feature does not apply on macos?

dariakp (changes requested):

test/manual/kerberos.test.js
durran: Removed.

dariakp (approved):

PR #3131, opened 2022-02-04
feat(NODE-2939): add new hostname canonicalization opts

durran:
### Description
NODE-2939
Also syncs NODE-2443 and NODE-2563

Updates `CANONICALIZE_HOST_NAME` options to accept `true`, `false`, "none", "forward", "forwardAndReverse".

#### What is changing?

When canonicalizing the host name when using GSSAPI, the driver will now behave as follows for these values:
- `true` or "forwardAndReverse": Performs a forward DNS lookup of the host and a reverse lookup of the IP address to obtain the hostname. If the reverse lookup fails the driver falls back to a cname lookup.
- "forward": Performs a cname lookup of the host.
- `false` or "none": Does no hostname canonicalization.

When syncing the spec tests some other auth spec tests we hadn't synced yet were there, so I decided to leave them and skip the single failure.

None

DRIVERS-1803 / NODE-2939


src/cmap/auth/gssapi.ts
```diff
@@ -174,14 +177,41 @@ function performGssapiCanonicalizeHostName(
...
+  }
+  // The case for forward is just to resolve the cname as dns.lookup()
+  // will not return it.
+  resolveCname(host, callback);
+}
 
+function resolveCname(host: string, callback: Callback<string>): void {
   // Attempt to resolve the host name
   dns.resolveCname(host, (err, r) => {
     if (err) return callback(err);
```
addaleax, line 211: > I think this should also be

```suggestion
    if (err) return callback(undefined, host);
```

> because Node.js gives you `ENODATA` errors for domains which exist but have no CNAME records attached

(https://github.com/mongodb/node-mongodb-native/pull/3129#discussion_r799450206)

durran: Updated to suggestion.

nbbeeken (changes requested):

src/cmap/auth/mongo_credentials.ts
```diff
@@ -158,6 +163,11 @@ export class MongoCredentials {
       // TODO(NODE-3485): Replace this with a MongoAuthValidationError
       throw new MongoAPIError(`Password not allowed for mechanism MONGODB-X509`);
     }
+
+    const canonicalization = this.mechanismProperties.CANONICALIZE_HOST_NAME ?? false;
+    if (!CANONICALIZATION_VALUES.includes(canonicalization)) {
+      throw new MongoAPIError(`Invalid CANONICALIZE_HOST_NAME value: ${canonicalization}`);
+    }
```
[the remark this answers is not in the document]
durran: We did not say to deprecate in the spec - we just changed the wording that drivers MAY support `true` and `false`.

src/cmap/auth/gssapi.ts
```diff
@@ -174,14 +177,48 @@ function performGssapiCanonicalizeHostName(
...
+    resolveCname(host, callback);
+  }
+}
 
+function resolveCname(host: string, callback: Callback<string>): void {
   // Attempt to resolve the host name
   dns.resolveCname(host, (err, r) => {
-    if (err) return callback(err);
+    /* eslint no-console: 0 */
+    console.log('resolveCname', err, r);
```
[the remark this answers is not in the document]
durran: Removed.

test/unit/assorted/auth.spec.test.ts
```diff
@@ -1,13 +1,16 @@
...
+const SKIP = ['should throw an exception if username and no password (MONGODB-AWS)'];
+
 describe('Auth option spec tests', function () {
   const suites = loadSpecTests('auth');
 
   for (const suite of suites) {
     describe(suite.name, function () {
       for (const test of suite.tests) {
-        it(`${test.description}`, function () {
+        const maybeIt = SKIP.includes(test.description) ? it.skip : it;
```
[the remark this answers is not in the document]
durran: Changed to hook and added skip reason with new NODE-3986 note.

nbbeeken (dismissed):

test/spec/auth/connection-string.json
```diff
@@ -108,6 +143,16 @@
         }
       }
     },
+    {
+      "description": "must raise an error when the authSource is empty",
```
dariakp, line 147: This test and the one immediately below implement NODE-2443

test/spec/auth/connection-string.json
```diff
@@ -207,6 +252,18 @@
         "mechanism_properties": null
       }
     },
+    {
+      "description": "should recognize the mechanism with no username when auth source is explicitly specified (MONGODB-X509)",
```
dariakp, line 256: This is NODE-2563

test/spec/auth/connection-string.json
```diff
@@ -367,6 +424,62 @@
...
+      "credential": {
+        "username": null,
+        "password": null,
+        "source": "$external",
+        "mechanism": "MONGODB-AWS",
+        "mechanism_properties": null
+      }
+    },
+    {
+      "description": "should recognise the mechanism when auth source is explicitly specified (MONGODB-AWS)",
```
dariakp, line 441: This is NODE-2563

test/unit/assorted/auth.spec.test.ts
```diff
@@ -1,13 +1,19 @@
 import { loadSpecTests } from '../../spec';
 import { executeUriValidationTest } from '../../tools/uri_spec_runner';
 
+const SKIP = ['should throw an exception if username and no password (MONGODB-AWS)'];
```
dariakp, line 4: This should have been implemented as part of the closed NODE-2288 ticket but wasn't

dariakp (commented):
I'm leaving some comments here because we pulled in more specs than just this ticket, so we should tag the corresponding tickets on this PR and then close them when this merges

src/cmap/auth/gssapi.ts
```diff
@@ -12,10 +12,22 @@ import {
 import { Callback, ns } from '../../utils';
 import { AuthContext, AuthProvider } from './auth_provider';
 
+/** @public */
+export const CANONICALIZATION_VALUES = [
```
dariakp, line 16: I'm wondering if we could take the same approach here as we do elsewhere with enums, (e.g., `ReadPreferenceMode`) with an `export const ... Object.freeze` and `export type ... typeof ... keyof` for consistency, that way we don't have use actual strings in our code

test/manual/kerberos.test.js
```diff
@@ -79,18 +79,73 @@ describe('Kerberos', function () {
...
+  context('when passing in CANONICALIZE_HOST_NAME', function () {
+    beforeEach(function () {
+      if (process.platform === 'darwin') {
+        this.currentTest.skipReason =
+          'DNS does not resolve with proper CNAME record on evergreen MacOS';
+        this.skip();
+      }
+    });
+
+    context('when the value is true', function () {
```
dariakp, line 91: For this entire set of tests, the actual test logic seems to be identical - is there more we can test for each specific test case here to make sure that the code is actually doing something different in each case?

dariakp (changes requested):

src/cmap/auth/gssapi.ts
durran: I've updated to use the same pattern.

```
export const CanonicalizationProperties = Object.freeze({
  on: true,
  off: false,
  none: 'none',
  forward: 'forward',
  forwardAndReverse: 'forwardAndReverse'
} as const);
```

test/manual/kerberos.test.js
durran: I've updated to assert on the dns spy for each of the calls.

test/spec/auth/connection-string.json
durran: Linked in JIRA ticket.

test/spec/auth/connection-string.json
durran: Linked in JIRA ticket.

test/unit/assorted/auth.spec.test.ts
durran: https://jira.mongodb.org/browse/NODE-3986

src/cmap/auth/gssapi.ts
```diff
@@ -179,14 +192,47 @@ function performGssapiCanonicalizeHostName(
...
+    // will not return it.
+    resolveCname(host, callback);
+  }
+}
 
+function resolveCname(host: string, callback: Callback<string>): void {
   // Attempt to resolve the host name
   dns.resolveCname(host, (err, r) => {
-    if (err) return callback(err);
+    if (err) return callback(undefined, host);
```
baileympearson, line 232: Is this change intentional?  This seems to change the error handling behavior to always just fallback to returning the `host` instead of propagating the error.

src/cmap/auth/gssapi.ts
```diff
@@ -12,10 +12,23 @@ import {
 import { Callback, ns } from '../../utils';
 import { AuthContext, AuthProvider } from './auth_provider';
 
+/** @public */
+export const CanonicalizationProperties = Object.freeze({
```
dariakp, line 16: Would keeping these as `CanonicalizationValues` instead of "properties" (or perhaps `CanonicalizationSettings`) be better? Properties could imply keys that can be set on an object, but this could just be me

test/manual/kerberos.test.js
```diff
@@ -79,18 +79,81 @@ describe('Kerberos', function () {
...
+
+    context('when the value is false', function () {
+      it('successfully authenticates', function (done) {
+        const client = new MongoClient(
+          `${krb5Uri}&authMechanismProperties=SERVICE_NAME:mongodb,CANONICALIZE_HOST_NAME:false&maxPoolSize=1`
+        );
+        client.connect(function (err, client) {
+          if (err) return done(err);
+          expect(dns.resolveCname).to.not.be.calledOnce;
+          expect(dns.lookup).to.not.be.calledOnce;
```
dariakp, line 139: Would this and the assertions in the following case be stronger if we expect `to.not.be.called` instead of `to.not.be.calledOnce`?

src/cmap/auth/gssapi.ts
durran: Yes this was intentional to always fallback to the provided host if the dns lookups error.

src/cmap/auth/gssapi.ts
```diff
@@ -179,14 +192,47 @@ function performGssapiCanonicalizeHostName(
...
+      // No ip found, return the error.
+      if (error) return callback(error);
+
+      // Perform a reverse ptr lookup on the ip address.
+      dns.resolvePtr(address, (err, results) => {
+        // This can error as ptr records may not exist for all ips. In this case
+        // fallback to a cname lookup as dns.lookup() does not return the
+        // cname.
+        if (err) {
+          return resolveCname(host, callback);
```
dariakp, line 216: Does this and the logic in L219 contradict the spec directive: "The driver MUST error if **any** lookup errors or returns no results."

src/cmap/auth/gssapi.ts
```diff
@@ -179,14 +192,47 @@ function performGssapiCanonicalizeHostName(
...
+        }
+        // If the ptr did not error but had no results, return the host.
+        callback(undefined, results.length > 0 ? results[0] : host);
+      });
+    });
+  } else {
+    // The case for forward is just to resolve the cname as dns.lookup()
+    // will not return it.
```
dariakp, line 224: what's the difference between dns.lookup and dns.resolveCname?

test/manual/kerberos.test.js
dariakp: Could we make the test cases a bit more descriptive in terms of what the spec says should happen? (i.e., for `'none'` and `false`, `"should authenticate without performing any lookups"`, for `'forwardAndReverse'` and `true`, `"should authenticate by performing a forward and reverse lookup"`, for `'forward'`, `"should authenticate by performing a forward DNS lookup only"`). And then probably also add cases for what's expected on error if that's not too difficult to do (if sinon lets us also stub the response from the functions, we could simulate errors there). For the cases that should be identical, we could make sure the logic stays the same by doing something like 
```js
for const option of [true, 'forwardAndReverse'] { 
  context(`when the value is ${option}`, function() {
    it('should authenticate by performing a forward and reverse lookup', ... );
    it('should <error behavior> if the forward lookup fails with an error', ...);
    it('should <error behavior> if the forward lookup fails with no results', ...);
    it('should <error behavior> if the reverse lookup fails with an error', ...);
    it('should <error behavior> if the reverse lookup fails with no results', ...);
  })
}
``` 
and similarly for `[false, 'none']`, with a separate block for `'forward'` only

dariakp (changes requested):

src/cmap/auth/gssapi.ts
durran: `dns.lookup()` returns an ip address and `dns.resolveCname()` returns a host name.

dariakp (commented):
Let's make sure we rebase after https://github.com/mongodb/node-mongodb-native/pull/3138 goes in, because the credential option parsing will be affected

src/cmap/auth/gssapi.ts
durran: Changes to `CanonicalizationValues`

test/manual/kerberos.test.js
durran: Updated to `to.not.be.called`

durran:
@dariakp This is now rebased and uses the test refactor suggestions.

durran:
Investigating last 2 failures.

src/cmap/auth/gssapi.ts
durran: Spec changes merged and verified against the way the current shell works.

dariakp (commented):
@durran Thanks for updating the tests! Would it be worthwhile/possible to add the expectations on the fallback behavior at each point in each case? Just to make sure we don't accidentally break that in the future?

nbbeeken (changes requested):

durran:
> @durran Thanks for updating the tests! Would it be worthwhile/possible to add the expectations on the fallback behavior at each point in each case? Just to make sure we don't accidentally break that in the future? **Edit**: if possible, we'd also want to confirm that the final response correctly uses the reply of the respective functions (like, we'd want to stub the reply from the reverse lookup to return something different than the forward, and then make sure we use the reverse response)

We actually didn't want to stub the later case, which is why I opened https://jira.mongodb.org/browse/BUILD-14646 to get an integration test for it afterwards. For the others I'll add expectations on the spies for the fallback behaviour.

src/cmap/auth/gssapi.ts
```diff
@@ -12,10 +12,23 @@ import {
 import { Callback, ns } from '../../utils';
 import { AuthContext, AuthProvider } from './auth_provider';
 
+/** @public */
+export const CanonicalizationValues = Object.freeze({
```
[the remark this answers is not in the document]
durran: Changed to `GSSAPICanonicalizationValues`.

src/index.ts
```diff
@@ -176,6 +176,7 @@ export type {
   ResumeToken,
   UpdateDescription
 } from './change_stream';
+export type { GSSAPICanonicalizationValues } from './cmap/auth/gssapi';
```
nbbeeken, line 179: The enum and the type need to both be exported, we've collected the enums we export on line 87

durran: Updated.

test/manual/kerberos.test.js
durran: I updated the tests now to test the further error/empty array/success cases on the reverse lookup and the cname lookup. The general case I left as-is since stubbing dns.lookup would affect the other lookup calls and not test this in isolation.

src/index.ts
```diff
@@ -86,6 +86,7 @@ export {
 
 // enums
 export { BatchType } from './bulk/common';
+export { GSSAPICanonicalizationValues } from './cmap/auth/gssapi';
```
nbbeeken, line 89: Sorry, I had edited my comment after I posted so probably was missed, I think we should not make this plural, and we need **both** the enum export here (thanks!) and type export that was here below.

Here's an example of the code that ensuring both exports supports
```ts
import { GSSAPICanonicalizationValue } from 'mongodb';

function processKerberosOption(x: GSSAPICanonicalizationValue) { 
  // The same symbol from the one import above is both a type annotation
  // And a value that can be used as an enum
  x = GSSAPICanonicalizationValue.forwardAndReverse;
}
```

nbbeeken (changes requested):

src/index.ts
durran: Updated to `GSSAPICanonicalizationValue` and verified the single export works as-is.

nbbeeken (approved):

test/manual/kerberos.test.js
```diff
@@ -79,18 +80,179 @@ describe('Kerberos', function () {
     });
   });
 
-  it('validate that CANONICALIZE_HOST_NAME can be passed in', function (done) {
-    if (process.platform === 'darwin') {
-      this.test.skipReason = 'DNS does not resolve with proper CNAME record on evergreen MacOS';
-      this.skip();
+  context('when passing in CANONICALIZE_HOST_NAME', function () {
+    beforeEach(function () {
+      //if (process.platform === 'darwin') {
```
dariakp: Looks like this is still WIP - but we probably want to either uncomment these or delete them before we merge.

Also, to summarize discussions from the earlier calls, for both success and failure cases, wherever possible, it would be good to assert the result from the correct function is returned (not just counting function calls, because, among other things, function call counts don't ensure any particular order of calls) or what those functions are called with. We can do that by making sure the functions resolve with different values and then checking that the callback is called with the one from the expected function call.

dariakp (changes requested):

test/manual/kerberos.test.js
nbbeeken: Just scanning I think that the 'when the cname lookup is empty' test section could test the value of the hostname returned, it would be some mock value, but the empty case at least hands back what was passed in. 

Separate point: 
Would it help to move the testing of this to a unit file and export the `performGSSAPICanonicalizeHostName` function? I think we have a good idea of the very explicit order of events that need be tested, and putting it in unit would just mean not needing what ever env variables are needed to spin up the actual int tests.

durran: I've uncommented that. With respect to asserting the number of function calls vs what they were called with, it's a bit tricky in this case. Firstly, these tests never stub `dns.lookup` as this is not the only place it is used and the stubs are on the entire module. So these tests are stubbing the reverse ptr lookup and the cname lookups since they are only used in the canonicalization case. With the reverse ptr lookup, that would need to assert on the stub that it was passed an IP address, which I didn't want to hard code in the test and attempting the lookup would be yet another call in the test to `dns.lookup`, which we already had a spy on. (I felt like it was going in circles). :) For the cname assertions I've changed this from `calledOnce` to `calledOnceWith` to ensure the host was being passed in.

dariakp: Could the stub/real method concern be worked around with the sinon callThrough method on stubs, and the conditional onFirstCall/onSecondCall type definitions for what the stub returns? (Like in the last couple of comments in this [thread](https://stackoverflow.com/questions/37343232/sinon-js-only-stub-a-method-once))

There's also Neal's suggestion to make it more of a unit test. If we have to, I don't think we absolutely must hold up the feature on this, but at the same time I am hesitant to have a feature where we don't actually fully test that the intended behavior is executed for each of the specified options.

durran: My preference would be to create a new ticket as a follow up for testing, and separate these into unit tests as @nbbeeken suggested by exporting the `performGSSAPICanonicalizeHostName` function. Then when the BUILD ticket is done for the DNS changes I can update the auth spec with prose tests for Kerberos and ensure our manual tests follow the prose tests.

nbbeeken (approved):
Thanks! LGTM

dariakp (approved):
LGTM - great job with the new tests!

PR #3133, opened 2022-02-07
test(NODE-3719,NODE-3543): fix unified runner to not pass uri opts in object and unskip lb tests

dariakp:
### Description
NODE-3719, NODE-3543

#### What is changing?
Fixed the unified runner to not pass in URI options as an object, since that behavior is incorrect for URI-only options (like `loadBalanced`, and since they are already included in the generated URI string anyway.

N/A

Spec compliance review


nbbeeken (approved):
LGTM, good catch

PR #3134, opened 2022-02-07
chore(NODE-3971) options as the first argument of `collection.watch()`

alecgibson:

We're currently allowed to pass either a `pipeline` or `options` to
`collection.watch()` as the first argument.

However, the TypeScript definitions don't reflect this. This change adds
method overloads that allow for passing the options as the only argument
as well as passing it as a second argument.

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -1422,8 +1422,13 @@ export class Collection<TSchema extends Document = Document> {
    * @param pipeline - An array of {@link https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/|aggregation pipeline stages} through which to pass change stream documents. This allows for filtering (using $match) and manipulating the change stream documents.
    * @param options - Optional settings for the command
    */
+  watch<TLocal = TSchema>(options?: ChangeStreamOptions): ChangeStream<TLocal>;
   watch<TLocal = TSchema>(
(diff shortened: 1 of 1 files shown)


baileympearson:
Hey @alecgibson - thanks for taking the time to submit a PR!  At this time, this change is not something we'd like to merge into our driver.  Although our code does allow users to omit the `pipeline` parameter, we would like to keep the types as they are.  The workaround for this particular issue would be to just pass in an empty array for the pipeline, which isn't a significant amount of overhead (and is arguably easier to understand):

```typescript
collection.watch([], options)
```

In the future, the Node team would like to move away from using overloads as much as we can for the following reasons:

- TS overloading causes issues with our TSDocs because only first overload is documented.
- Overloading has both a runtime cost (extra logic to correctly assign parameters) and a development time cost (extra logic and reasoning necessary to worry about correctly assigning parameters).
- VSCode only shows the first overload by default, which makes it less helpful when there are many overloads.
- Many of our overloads are required because we support both promises and callbacks in the driver.  In the next major version, we hope to remove callback support.  Once callback support is removed, many overloads will be unnecessary.

Although there will be scenarios where overloads will be necessary (such as when the return type of the function is dependent on the parameters of the function), this particular method is not an exception.

PR #3136, opened 2022-02-08
fix: Filter query root operators support for partials

avaly:
### Description

A change was introduced in https://github.com/mongodb/node-mongodb-native/pull/3102 in v4.3.1 regarding the new `Filter` type, which throws new errors for filter objects with root operators created dynamically.

#### What is changing?

I moved the `RootFilterOperators<WithId<TSchema>>` outside of the union inside the `Filter<TSchema>`.

This fixes the use case below, but it does break 3 lines in the find recursive test. I would love some help from @baileympearson, since you were the one to work on the recursive support.

No.

```ts
// this works in v4.3.1
const fooFilterBuiltNow: Filter<Foo> = {
  a: 'foo',
  $or: [{ a: 'bar' }]
};

const fooFilterBuiltLater: Filter<Foo> = {};
fooFilterBuiltLater.a = 'foo';
// this throws a TS error in v4.3.1
fooFilterBuiltLater.$or = [{ a: 'bar' }];
```

### DIFF
--- a/src/mongo_types.ts
+++ b/src/mongo_types.ts
@@ -65,13 +65,15 @@ export type EnhancedOmit<TRecordOrUnion, KeyUnion> = string extends keyof TRecor
 export type WithoutId<TSchema> = Omit<TSchema, '_id'>;
 
 /** A MongoDB filter can be some portion of the schema or a set of operators @public */
-export type Filter<TSchema> =
+export type Filter<TSchema> = (
   | Partial<TSchema>
-  | ({
+  | {
       [Property in Join<NestedPaths<WithId<TSchema>>, '.'>]?: Condition<
         PropertyType<WithId<TSchema>, Property>
       >;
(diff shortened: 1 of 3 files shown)


avaly:
@baileympearson Understood. It's not pressing, I just stumbled upon this today while trying to upgrade our codebase and hit this error a lot of times (due to how our code is structured).

avaly:
@baileympearson Any update on this one? 

We're running into this issue when using computed properties in a filter query:

```ts
await MetadataItem.updateMany({
  _id: { $nin: ids },
 [dynamicKey]: { $gt: 0 }
}, update);
```

TypeScript is complaining that `Types of property '_id' are incompatible.` because the filter is only tested against the `Partial<TSchema>` of the union type in `Filter<TSchema>`, most likely due to the dynamic key.

PR #3137, opened 2022-02-08
test(NODE-3860): improve skipReason reporting for disabled 'auth' tests

baileympearson:
#### What is changing?

This PR cleans up the `skipReason` logic added in https://github.com/mongodb/node-mongodb-native/pull/3121.  By including the skipReason and the filter status `auth: disabled` in the metadata for tests that fail because `auth` was enabled, in cases where the test would not have run due to other requirements, we report the skip reason being that `auth` was enabled rather than the actual skip reason for the test.

This PR also fixes the broken AWS tests.

No.


dariakp (commented):
LGTM in general, but what do we think of dropping the leading instance of `test` in `testSkipBrokenAuthTestBeforeEachHook` and just making it `skipBrokenAuthTestBeforeEachHook `?

nbbeeken (approved):
LGTM ty!

PR #3138, opened 2022-02-08
fix(NODE-3813): unexpected type conversion of read preference tags

baileympearson:
#### What is changing?

This PR does two things:
- fixes options parsing for arrays of options.
- refactors how we parse key-value pairs of options for a particular option.  In particular, it removes logic from parsing `readPreference` that was converting string values to native JS types, causing spec tests to fail

No.

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -236,25 +236,6 @@ function* entriesFromString(value: string) {
   }
 }
 
-function* parseAuthMechanismParameters(stringValue: string) {
-  const validKeys = [
-    'SERVICE_NAME',
-    'SERVICE_REALM',
-    'CANONICALIZE_HOST_NAME',
-    'AWS_SESSION_TOKEN'
-  ];
-
-  for (const [key, value] of entriesFromString(stringValue)) {
-    if (validKeys.includes(key)) {
-      if (key === 'CANONICALIZE_HOST_NAME') {
-        yield [key, getBoolean(key, value)];
-      } else {
-        yield [key, value];
-      }
-    }
-  }
-}
-
 class CaseInsensitiveMap<Value = any> extends Map<string, Value> {
   constructor(entries: Array<[string, any]> = []) {
     super(entries.map(([k, v]) => [k.toLowerCase(), v]));
@@ -652,7 +633,26 @@ export const OPTIONS = {
     target: 'credentials',
     transform({ options, values: [value] }): MongoCredentials {
       if (typeof value === 'string') {
-        value = Object.fromEntries(parseAuthMechanismParameters(value));
+        const validKeys = [
+          'SERVICE_NAME',
+          'SERVICE_REALM',
+          'CANONICALIZE_HOST_NAME',
+          'AWS_SESSION_TOKEN'
+        ];
+
+        const properties = Object.create(null);
+
+        for (const [key, _value] of entriesFromString(value)) {
+          if (validKeys.includes(key)) {
+            if (key === 'CANONICALIZE_HOST_NAME') {
+              properties[key] = getBoolean(key, _value);
+            } else {
+              properties[key] = _value;
+            }
+          }
+        }
+
+        value = properties;
       }
       if (!isRecord(value)) {
         throw new MongoParseError('AuthMechanismProperties must be an object');
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -204,11 +204,22 @@ function getUint(name: string, value: unknown): number {
   return parsedValue;
 }
 
-
 function toArray<T>(value: T): T[];
 function toArray<T>(value: T[]): T[];
 function toArray<T>(value: T | T[]): T[] {
   return Array.isArray(value) ? value : [value];
+}
+
+function parseType(value: string): number | boolean | string {
+  try {
+    return getBoolean('', value);
+  } catch {
+    try {
+      return getInt('', value);
+    } catch {
+      return value;
+    }
+  }
 }
 
 function toRecord(value: string): Record<string, any> {
@@ -219,18 +230,7 @@ function toRecord(value: string): Record<string, any> {
     if (value == null) {
       throw new MongoParseError('Cannot have undefined values in key value pairs');
     }
-    try {
-      // try to get a boolean
-      record[key] = getBoolean('', value);
-    } catch {
-      try {
-        // try to get a number
-        record[key] = getInt('', value);
-      } catch {
-        // keep value as a string
-        record[key] = value;
-      }
-    }
+    record[key] = value;
   }
   return record;
 }
@@ -632,7 +632,9 @@ export const OPTIONS = {
     target: 'credentials',
     transform({ options, values: [value] }): MongoCredentials {
       if (typeof value === 'string') {
-        value = toRecord(value);
+        value = Object.fromEntries(
+          Object.entries(toRecord(value)).map(([key, value]) => [key, parseType(value)])
+        );
       }
       if (!isRecord(value)) {
         throw new MongoParseError('AuthMechanismProperties must be an object');
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -224,7 +224,7 @@ function toArray<T>(value: OneOrMore<T>): T[] {
   return Array.isArray(value) ? value : [value];
 }
 
-function* parseObjectString(value: string) {
+function* entriesFromString(value: string) {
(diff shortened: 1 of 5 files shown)


src/connection_string.ts
```diff
@@ -228,7 +228,7 @@ function toRecord(value: string): Record<string, any> {
   return record;
 }
 
-class CaseInsensitiveMap extends Map<string, any> {
+class CaseInsensitiveMap<Value = any> extends Map<string, Value> {
```
baileympearson, line 231: I added a type parameter here because I wanted stronger typing on the `urlOptions` object.  Since the urlOptions object is always `Map<string, any[]>`, I wanted Typescript to help out so that it's clear to future developers on line 332 why we can immediately spread the object (unlike the other two if cases).

src/connection_string.ts
```diff
@@ -478,12 +485,11 @@ export function parseOptions(
     throw new MongoParseError('Can only specify both of proxy username/password or neither');
   }
 
-  if (
-    urlOptions.get('proxyHost')?.length > 1 ||
-    urlOptions.get('proxyPort')?.length > 1 ||
-    urlOptions.get('proxyUsername')?.length > 1 ||
-    urlOptions.get('proxyPassword')?.length > 1
-  ) {
+  const proxyOptions = ['proxyHost', 'proxyPort', 'proxyUsername', 'proxyPassword'].map(
```
baileympearson, line 488: Previously, urlOptions was of type `Map<string, any>`, so `.get()` returned a value of the type `any | undefined` which gets simplified to `any`.  So no type errors.

When I added stricter typing to the `urlOptions` object, `.get()` now returns `any[] | undefined`, Typescript complains that `.get(...)?.length` could be undefined.

durran:
CSFLE failures unrelated and fixed in https://github.com/mongodb/node-mongodb-native/pull/3070

src/connection_string.ts
```diff
@@ -326,13 +326,20 @@ export function parseOptions(
   for (const key of allKeys) {
     const values = [];
     if (objectOptions.has(key)) {
-      values.push(objectOptions.get(key));
+      const options = Array.isArray(objectOptions.get(key))
```
durran, line 329: nit: this could be simplified by doing `[].concat(objectOptions.get(key))` in these 3 places. It looks a little odd but does the same without accessing the options twice or the need for the ternary. Could also make it a util function called `wrapInArray` or something similar.

durran (changes requested):

dariakp (changes requested):
Please make sure to either unskip or update the TODO tagged to this ticket in uri_options.spec.test

durran:
There's an eslint error on latest: https://spruce.mongodb.com/task/mongo_node_driver_next_lint_run_checks_patch_44bbd6e94d58fac9eb07a07e5437662eb6409dc5_62041da20305b97e46960d87_22_02_09_20_01_39/logs?execution=0

durran (approved):

src/connection_string.ts
```diff
@@ -204,6 +204,24 @@ function getUint(name: string, value: unknown): number {
   return parsedValue;
 }
 
+function toArray<T>(value: T): T[];
+function toArray<T>(value: T[]): T[];
+function toArray<T>(value: T | T[]): T[] {
+  return Array.isArray(value) ? value : [value];
+}
```
nbbeeken, line 211:
```suggestion
/** Wrap single value in an array if the value isn't an array */
function orMore<T>(value: OneOrMore<T>): T[] {
  return Array.isArray(value) ? value : [value];
}
```
Just some clever lingo, optional suggestion 😎

src/connection_string.ts
```diff
@@ -204,6 +204,24 @@ function getUint(name: string, value: unknown): number {
...
+  try {
+    return getBoolean('', value);
+  } catch {
+    try {
+      return getInt('', value);
+    } catch {
+      return value;
+    }
+  }
+}
```
nbbeeken, line 223: I know I authored this but I'm thinking this is a 🦶 🔫  waiting to happen. `getBoolean` will give you true for `1` and false for `0` and `-1` making those numbers impossible to reach `getInt`. 
I think we should remove this, and instead inspect `authMechanismProperties` for the properties we know about, and translate them to their types and leave the rest as strings.

nbbeeken (changes requested):

src/connection_string.ts
```diff
@@ -184,8 +185,22 @@ const FALSEHOODS = new Set(['false', 'f', '0', 'n', 'no', '-1']);
 function getBoolean(name: string, value: unknown): boolean {
   if (typeof value === 'boolean') return value;
   const valueString = String(value).toLowerCase();
-  if (TRUTHS.has(valueString)) return true;
-  if (FALSEHOODS.has(valueString)) return false;
+  if (TRUTHS.has(valueString)) {
```
baileympearson, line 188: The spec for parsing booleans says that while drivers are encouraged to support values other than `true/false`, a warning is supposed to be emitted: https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#values

src/connection_string.ts
```diff
@@ -631,7 +652,7 @@ export const OPTIONS = {
     target: 'credentials',
     transform({ options, values: [value] }): MongoCredentials {
       if (typeof value === 'string') {
-        value = toRecord(value);
+        value = Object.fromEntries(parseAuthMechanismParameters(value));
```
nbbeeken, line 655: In order to maintain the previous behavior we can do this, or reevaluate the need

```js
        value = Object.assign(Object.create(null), Object.fromEntries(parseAuthMechanismParameters(value)));
```

nbbeeken (changes requested):

src/connection_string.ts
```diff
@@ -631,7 +633,26 @@ export const OPTIONS = {
...
+          if (validKeys.includes(key)) {
+            if (key === 'CANONICALIZE_HOST_NAME') {
+              properties[key] = getBoolean(key, _value);
+            } else {
+              properties[key] = _value;
+            }
+          }
+        }
+
+        value = properties;
```
nbbeeken, line 655: Could we hoist properties and pass it into MongoCreds?

nbbeeken (changes requested):

src/connection_string.ts
```diff
@@ -3,6 +3,7 @@ import * as fs from 'fs';
 import ConnectionString from 'mongodb-connection-string-url';
 import { URLSearchParams } from 'url';
 
+import type { OneOrMore } from '.';
```
nbbeeken, line 6:
```suggestion
import type { OneOrMore } from './mongo_types';
```

I'm not sure if we could get a lint rule to help us check this? Maybe the forbidden imports thing? but we could defer that

nbbeeken (changes requested):

nbbeeken (dismissed):
LGTM

src/connection_string.ts
```diff
@@ -631,7 +634,28 @@ export const OPTIONS = {
...
+          'CANONICALIZE_HOST_NAME',
+          'AWS_SESSION_TOKEN'
+        ];
+
+        const mechanismProperties = Object.create(null);
+
+        for (const [key, _value] of entriesFromString(value)) {
+          if (validKeys.includes(key)) {
+            if (key === 'CANONICALIZE_HOST_NAME') {
+              mechanismProperties[key] = getBoolean(key, _value);
```
dariakp, line 649: Just a note that this is probably going to conflict with Durran's PR to expand the CANONICALIZE_HOST_NAME options

src/connection_string.ts
```diff
@@ -631,7 +634,28 @@ export const OPTIONS = {
     target: 'credentials',
     transform({ options, values: [value] }): MongoCredentials {
       if (typeof value === 'string') {
-        value = toRecord(value);
+        const validKeys = [
```
dariakp, line 637: Are we going to intentionally silently ignore invalid keys here? Does this actually produce the same behavior as before?

nbbeeken: This array might be unnecessary, the idea was to only translate the type of the keys we know about and leave whatever remains as is (strings). I spoke generally about the options, so that might be where the inspiration came from but I think `CANONICALIZE_HOST_NAME`, and `gssapiCannoncializeHostname` (which we need to add here, whoops!) are the only ones that need translating to boolean. the rest can remain as strings

dariakp (approved):

PR #3143, opened 2022-02-11
fix(NODE-3917): Throw an error when directConnection is set with multiple hosts

baileympearson:
#### What is changing?

This PR updates the URI parsing logic to throw an error when the `directConnection` option is set with more than one host.

No, I don't think so.

### DIFF
--- a/test/spec/uri-options/connection-options.json
+++ b/test/spec/uri-options/connection-options.json
@@ -135,7 +135,8 @@
       "valid": false,
       "warning": false,
       "hosts": null,
-      "auth": null
+      "auth": null,
+      "options": {}
     },
     {
       "description": "directConnection=false",
@@ -180,6 +181,18 @@
       }
     },
     {
+      "description": "loadBalanced=true with directConnection=false",
+      "uri": "mongodb://example.com/?loadBalanced=true&directConnection=false",
+      "valid": true,
+      "warning": false,
+      "hosts": null,
+      "auth": null,
+      "options": {
+        "loadBalanced": true,
+        "directConnection": false
+      }
+    },
+    {
       "description": "loadBalanced=false",
       "uri": "mongodb://example.com/?loadBalanced=false",
       "valid": true,
@@ -211,15 +224,6 @@
     {
       "description": "loadBalanced=true with directConnection=true causes an error",
       "uri": "mongodb://example.com/?loadBalanced=true&directConnection=true",
-      "valid": false,
-      "warning": false,
-      "hosts": null,
-      "auth": null,
-      "options": {}
-    },
-    {
-      "description": "loadBalanced=true with directConnection=false causes an error",
-      "uri": "mongodb://example.com/?loadBalanced=true&directConnection=false",
       "valid": false,
       "warning": false,
       "hosts": null,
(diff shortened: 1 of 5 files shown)


test/unit/connection_string.test.ts
```diff
@@ -148,6 +148,19 @@ describe('Connection String', function () {
...
+      const options = parseOptions('mongodb://localhost:27027/?directConnection=true');
+      expect(options.directConnection).to.be.true;
+    });
+
+    it('throws when directConnection is true and there is more than one host', () => {
+      expect(() =>
+        parseOptions('mongodb://localhost:27027,localhost:27018/?directConnection=true')
+      ).to.throw(MongoParseError, 'directConnection option requires exactly one host');
+    });
+  });
```
addaleax, line 162: Is there some spec test for this as well? I think @dariakp uncovered this at the same time as me because she was working on those

dariakp: Yes, there are two tests tagged to this ticket in uri_options.spec.test

baileympearson: Yeah, you're right.  Good catch!  

If we have this covered in a spec test do you think it's worth keeping the tests I added?

dariakp: I don't think it hurts to keep our tests, but, strictly speaking, it's not really necessary since the existing spec tests do cover both of the above scenarios

test/spec/uri-options/connection-options.json
```diff
@@ -179,6 +180,18 @@
         "loadBalanced": true
       }
     },
+    {
```
dariakp, line 183: which spec repo commit did this spec change come from?

test/unit/assorted/uri_options.spec.test.ts
```diff
@@ -7,10 +7,6 @@ describe('URI option spec tests', function () {
   const suites = loadSpecTests('uri-options');
 
   const skipTests = [
-    // TODO(NODE-3917): Fix directConnection and loadBalanced option validation
```
dariakp, line 10: Be sure to also address the "Invalid loadBalanced value" scenario in this file below

dariakp (changes requested):

test/spec/uri-options/connection-options.json
baileympearson: https://github.com/mongodb/specifications/commit/33916af7c70397732c203b4c3f55c7bbf8e2aebc

and 

https://github.com/mongodb/specifications/commit/cb541a15ec3dcaa91d1ba3867330b45ddb22ce93

Adding the `options: {}` seemed like a small change, and our test still passed so it seemed like a nice update.

wrt these tests - we were actually testing old behavior that was not spec compliant.  We were failing the test because we are spec compliant, so updating the test fixes the issue 😄

baileympearson: These tests are now completely in sync with the specs repo too.

dariakp: Yup, thanks for getting those squared away! I just wanted to make sure we weren't keeping any related NODE tickets open and that the sync was complete if we're pulling in the changes. Crossing t's and dotting i's 😄

dariakp (approved):

nbbeeken (approved):

PR #3144, opened 2022-02-12
fix(NODE-3769): retryable writes are not compliant with specification

nbbeeken:
### Description
Fix NODE-3769, test NODE-2111

#### What is changing?

- Add a new `MongoUnexpectedServerResponseError` which is a subclass of `MongoRuntimeError` I could continue using `MongoRuntimeError` to avoid adding a new class, but the refactor/fix to execute operation called for it.
- Refactoring / correcting of executeOperation's retry logic
- isRetryableWrite now **only** checks the label if we're connected to a topology above a certain wire version, as required by the spec.
- Refactored the retryable write runner, no logical changes, just async/await, I ran my changes against the unmodified runner on main, so I'm confident that my test runner changes are safe
- Moved the labels to an enum `MONGODB_ERROR_LABELS`, This allows us to type check our labels, (mostly could catch future spelling mistakes if we miss actual test coverage) felt like it was still relevant to this work to keep in.
- Deleted the unused `clearAndRemoveTimerFrom` function. entirely internal, never reached.
- Deleted `connectServers` called only from one place and simply created a level of indirection that is frustrating to debug 
- Deleted `isRetryableEndTransactionError`, it had one condition, checking for the retryWriteLabel, inlining the condition has better readability, there could be reason to bring it back if there is more conditions.
- Expanded the conditionals in `supportsRetryableWrites` and made it accept an optional server (removes an existence gate from a caller if stmt)
- Removed the constrained timeout from LB testing, causes a lot of timeout failures when you debug locally
- `async`-ify `testOperations` in the legacy spec runner, debugging.
- Added to our error unit testing, we had a public error that wasn't exported at the top level, I think my test should catch all those cases now.
- New instructions for launching LB cluster
- ran prettier on `global.d.ts`, it doesn't run automagically currently 😿 

Our retryable logic was falling back on error codes, the spec explicitly states that on newer server versions the label should be the ONLY source of truth. So we were still retrying errors, and we will continue to do so, but the rules were incorrect.

  - TODO can we break the type of hasErrorLabel, addErrorLabel, narrowing from string to known labels
  - TODO file ticket about try catch in execOperation
  - TODO inside `commandHandler` of `endTransaction` is it okay that we only check the retry read condition before unpinning / adding the unknown commit label? I don't think so, but also the retry has already occurred at this stage, so I don't think the session gets reused, unpinning might be correct.


global.d.ts
```diff
@@ -42,6 +50,9 @@ interface MetadataAndTest<Fn> {
 
     interface Test {
       metadata: MongoDBMetadataUI;
+
+      /** @deprecated Attach spec to a test if you need access to it in a beforeEach hook */
```
baileympearson, line 54: Just curious - how come this is deprecated?  And what's the intended replacement?

src/cmap/connection.ts
```diff
@@ -109,8 +109,10 @@ export interface CommandOptions extends BSONSerializeOptions {
   noResponse?: boolean;
   omitReadPreference?: boolean;
 
-  // FIXME: NODE-2802
-  willRetryWrite?: boolean;
+  // FIXME(NODE-2802): Currently the CommandOptions take a property willRetryWrite which is a hint from executeOperation that the txnNum should be applied to this command.
```
baileympearson, line 112: incredibly unimportant nit: can we use `TODO` instead of `FIXME`?  Seems like we've been using `TODO` in recent work

src/utils.ts
```diff
@@ -1419,13 +1419,28 @@ export function enumToString(en: Record<string, unknown>): string {
...
+  if (!server) {
+    return false;
+  }
+
+  if (server.loadBalanced) {
+    // Loadbalanced topologies will always support retry writes
+    return true;
+  }
+
+  if (server.description.maxWireVersion >= 6) {
```
dariakp, line 1432: do we really need this check considering our min supported wire version is 6?

test/tools/unified-spec-runner/entities.ts
```diff
@@ -242,7 +242,7 @@ export class FailPointMap extends Map<string, Document> {
         try {
           await client.connect();
         } catch (error) {
-          console.error(`failed to connect disableFailPoints ${hostAddress}`);
+          console.error(`failed to connect disableFailPoints ${hostAddress} - ${error}`);
```
dariakp, line 245: Just curious - does that actually work well? in my experience the error objects don't do well inside template strings, and also we throw the error right away, so hypothetically, shouldn't that make the error show up in the logs immediately after that console statement?

test/unit/error.test.js
```diff
@@ -89,34 +110,6 @@ describe('MongoErrors', () => {
     });
   });
 
-  describe('#isRetryableEndTransactionError', function () {
```
dariakp, line 92: I'm not saying we should definitely test the new error helpers (isRetryableRead/Write), but what were your reasons for not including the unit tests for them?

test/types/community/client.test-d.ts
```diff
@@ -55,6 +55,7 @@ export async function testFunc(): Promise<MongoClient> {
 
 MongoClient.connect(connectionString, err => {
   if (err instanceof MongoError) {
+    //@ts-expect-error // TODO? can we break?
```
dariakp, line 58: context for this?

dariakp (changes requested):
Partial feedback so that I don't lose it, still got the main files left to go through

test/types/community/client.test-d.ts
nbbeeken: With the error label enum, typescript now can enforce you pass in one of the known labels, so narrower than string. This is a public API (although doubtful its useful outside of a driver) but this kind of type check seems like something desirable

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    expect(result).to.deep.include(expected);
+  }
+
+  if (test.outcome.collection) {
+    const collectionResults = await ctx.collection.find({}).toArray();
+
+    expect(collectionResults).to.deep.equal(test.outcome.collection.data);
+  }
+
+  return result;
```
dariakp, line 162: I also noticed that we aren't asserting on command construction here [like the spec says](https://github.com/mongodb/specifications/tree/master/source/retryable-writes/tests#command-construction-tests) even though we do turn on command monitoring. It may be out of scope for this ticket, but can you check my sanity here and maybe we can file a follow up?

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -0,0 +1,209 @@
...
+import { isAnyRequirementSatisfied } from '../../tools/unified-spec-runner/unified-utils';
+
+interface RetryableWriteTestContext {
+  client?: MongoClient;
+  db?: Db;
+  collection?: Collection;
+  failPointName?: any;
+}
+
+describe('Legacy Retryable Writes Specs', function () {
```
dariakp, line 15: The ticket NODE-3769 mentions that skipped tests should be unskipped for this. Do you know what was being referenced there? I wasn't able to find anything in the code base that was tagged to this ticket or any skipped retryable writes tests for that matter.

src/sessions.ts
```diff
@@ -729,25 +733,25 @@ function endTransaction(session: ClientSession, commandName: string, callback: C
...
-    if (e) {
+    if (error instanceof MongoError) {
       if (
-        e instanceof MongoNetworkError ||
-        e instanceof MongoWriteConcernError ||
-        isRetryableError(e) ||
-        isMaxTimeMSExpiredError(e)
+        error instanceof MongoNetworkError ||
+        error instanceof MongoWriteConcernError ||
+        isRetryableReadError(error) || // TODO is commit a write? probably, double check its retry rules
```
dariakp, line 740: any context on this TODO? do we need to figure out if this check also needs to check for retryable write now that they're split up?

src/operations/execute_operation.ts
```diff
@@ -94,63 +96,56 @@ export function executeOperation<
...
-      return cb(new MongoCompatibilityError('Current topology does not support sessions'));
+      return callback(new MongoCompatibilityError('Current topology does not support sessions'));
     }
 
     try {
-      executeWithServerSelection(topology, session, operation, (err, result) => {
-        if (session && session.owner && session.owner === owner) {
-          return session.endSession(err2 => cb(err2 || err, result));
+      executeWithServerSelection<TResult>(topology, session, operation, (error, result) => {
+        if (session && session.owner != null && session.owner === owner) {
```
dariakp, line 111:
```suggestion
        if (session?.owner != null && session.owner === owner) {
```

src/operations/execute_operation.ts
```diff
@@ -94,63 +96,56 @@ export function executeOperation<
...
         }
 
-        cb(err, result);
+        callback(error, result);
       });
-    } catch (e) {
-      if (session && session.owner && session.owner === owner) {
+    } catch (error) {
+      // TODO: we shouldn't catch here.
+      // We catch and NOT finally, cus its only in case of an error that we want to end the session
```
dariakp, line 119: I'm not following this logic - and why do we no longer want to throw the error or pass it to the callback? Looks like it's meant to catch sync errors from executeWithServerSelection?

src/operations/execute_operation.ts
```diff
@@ -94,63 +96,56 @@ export function executeOperation<
...
         session.endSession();
       }
-
-      throw e;
     }
   });
 }
 
-function supportsRetryableReads(server: Server) {
+function supportsRetryableReads(server?: Server) {
```
dariakp, line 127: could we remove this now or just change to return true now that we don't support servers below 6?

src/operations/execute_operation.ts
```diff
@@ -246,21 +248,20 @@ function executeWithServerSelection(
...
-      return;
+  topology.selectServer(selector, serverSelectionOptions, (error, server) => {
+    if (error || !server) {
+      return callback(error);
     }
 
     if (session && operation.hasAspect(Aspect.RETRYABLE)) {
       const willRetryRead =
-        topology.s.options.retryReads !== false &&
+        topology.s.options.retryReads !== false && // why is this not false
```
dariakp, line 258: the world wants to know; they both get set to true by default, and they are both supposed to be boolean as defined on the MongoClient, so I say we should update both places to match; in fact, we should probably just use `.retryReads` (no `===` or `!==`)

src/operations/execute_operation.ts
```diff
@@ -170,60 +165,68 @@ function executeWithServerSelection(
...
       if (
-        err &&
-        err instanceof MongoNetworkError &&
-        server.loadBalanced &&
-        session &&
-        session.isPinned &&
-        !session.inTransaction() &&
-        operation.hasAspect(Aspect.CURSOR_CREATING)
+        !error &&
+        operation.hasAspect(Aspect.WRITE_OPERATION) &&
```
dariakp, line 216: should this just be `isWriteOperation`?

src/error.ts
```diff
@@ -688,8 +725,8 @@ export class MongoWriteConcernError extends MongoServerError {
   }
 }
 
-// see: https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#terms
-const RETRYABLE_ERROR_CODES = new Set<number>([
+// https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#retryable-error
+const RETRYABLE_READS_ERROR_CODES = new Set<number>([
```
dariakp, line 729: nit: can we use consistent singular/plural form for both reads and writes? i.e., rename this one to `read` or the one below to `writes`

src/operations/execute_operation.ts
```diff
@@ -170,60 +165,68 @@ function executeWithServerSelection(
...
-      hasWriteAspect &&
-      itShouldRetryWrite &&
-      err.code === MMAPv1_RETRY_WRITES_ERROR_CODE &&
-      err.errmsg.match(/Transaction numbers/)
+      isWriteOperation &&
+      originalError.code === MMAPv1_RETRY_WRITES_ERROR_CODE &&
+      /Transaction numbers/.test(originalError.errmsg)
     ) {
-      callback(
+      return callback(
```
dariakp, line 185: This code happens to not currently be tested - I think we should pull in NODE-2111 as part of this work, since we're trying to ensure spec compliance (it's just one prose test)

dariakp (changes requested):

src/error.ts
```diff
@@ -705,41 +742,57 @@ const RETRYABLE_ERROR_CODES = new Set<number>([
...
-  MONGODB_ERROR_CODES.SocketException,
+  ...RETRYABLE_READS_ERROR_CODES,
   MONGODB_ERROR_CODES.ExceededTimeLimit
 ]);
 
-export function isRetryableEndTransactionError(error: MongoError): boolean {
-  return error.hasErrorLabel('RetryableWriteError');
-}
+export function isRetryableWriteError(error: MongoError, maxWireVersion: number): boolean {
+  if (!(error instanceof MongoError)) {
```
baileympearson, line 752: thought: What if we put a static method on the `MongoError` class to determine whether something was a `MongoError`?  We could add a type assertion too to help Typescript, if we ever cared about that

```typescript
class MongoError {
    ...

    static isMongoError(err: any) : err is MongoError { error instanceof MongoError }
}
```

baileympearson (changes requested):

src/operations/execute_operation.ts
nbbeeken: Good catch, I didn't mean to change the logic here, and yea that's what I'm getting at with my quick TODO, this relates to what we found in NODE-3422. This is just something I think we'll address with async/await, having multiple ways to recieve errors is tough to reason about.

src/cmap/connection.ts
nbbeeken: done

nbbeeken: done

src/operations/execute_operation.ts
nbbeeken: done

src/operations/execute_operation.ts
nbbeeken: done!

src/utils.ts
```diff
@@ -1419,13 +1419,28 @@ export function enumToString(en: Record<string, unknown>): string {
...
+    // Talking to a 3.6+ server
+    if (server.description.logicalSessionTimeoutMinutes != null) {
+      // that supports sessions
+      if (server.description.type !== ServerType.Standalone) {
+        // and that is not a standalone
+        return true;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
nbbeeken: We totally can, I wrote it like this because it was easier to parse (like with my 👀), but happy to change

test/integration/retryable-writes/retryable_writes.spec.test.ts
nbbeeken: @durran Do you have context on the tests you mentioned? I didn't come across skipping logic

test/tools/unified-spec-runner/entities.ts
nbbeeken: For some reason nothing was showing up, but this is easily re-added if debugging comes up again

test/types/community/client.test-d.ts
nbbeeken: This causes other headaches, lets stick with string

global.d.ts
nbbeeken: Deprecated is the the wrong word, let me leave a comment, we shouldn't need to attach the spec to a test object to get access to it inside a beforeEach I feel like its hack-y, but it's also the way we can get the legacy spec tests to use our new way of skipping tests easily. So I wanted the strikethough to discourage use, a very VSCode solution, but then again maybe this is the way we should do things. Right now the unified runner does everything from within the `it`, if we wanted to refactor it to use hooks we'd need to get the currentTest's spec description to the hook somehow

src/error.ts
nbbeeken: I think it's a good idea! While fixing this I got some real power out of our many specific errors, I'm wondering if we can do something that's generic, and available on all error classes as an improvement

src/sdam/topology.ts
```diff
@@ -435,7 +434,12 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
 
     // connect all known servers, then attempt server selection to connect
     const serverDescriptions = Array.from(this.s.description.servers.values());
-    connectServers(this, serverDescriptions);
+    this.s.servers = new Map(
+      serverDescriptions.map(serverDescription => [
+        serverDescription.address,
+        createAndConnectServer(this, serverDescription)
```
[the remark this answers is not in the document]
nbbeeken: We are doing async actions, but not an awaitable one, the monitor time interval is kicked off to begin sending heartbeats, (or streaming them). I don't believe any code in the remaining steps after this will reference something that's generated by the SDAM machinery. We send subsequent operations through server selection, that's how we await the data that comes out of SDAM. What were you thinking of?

src/error.ts
```diff
@@ -80,6 +80,17 @@ export const GET_MORE_RESUMABLE_CODES = new Set<number>([
...
+/** @public */
+export const MONGODB_ERROR_LABELS = Object.freeze({
+  RetryableWriteError: 'RetryableWriteError',
+  TransientTransactionError: 'TransientTransactionError',
+  UnknownTransactionCommitResult: 'UnknownTransactionCommitResult',
+  ResumableChangeStreamError: 'ResumableChangeStreamError'
+} as const);
+
+/** @public */
+export type MONGODB_ERROR_LABELS = typeof MONGODB_ERROR_LABELS[keyof typeof MONGODB_ERROR_LABELS];
```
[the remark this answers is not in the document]
nbbeeken: It might be confusing to read but it does give the expected behavior of an enum, which is a container for a set of values, and also enumerates the possible values. I do thing the all caps style doesn't fit with our other enums, udpated!

src/utils.ts
nbbeeken: Nope good point! removed!

test/unit/error.test.js
nbbeeken: I can write some unit tests, no reason against them

test/integration/retryable-writes/retryable_writes.spec.test.ts
nbbeeken: Yea I don't see any command collection happening, I think I can do it without much effort but I'll post a ticket if it's not simple.

src/operations/execute_operation.ts
```diff
@@ -170,60 +160,60 @@ function executeWithServerSelection(
...
-      ) {
-        session.unpin({ force: true, forceClear: true });
+      session.unpin({ force: true, forceClear: true });
+    }
+
+    // select a new server, and attempt to retry the operation
+    topology.selectServer(selector, serverSelectionOptions, (error?: Error, server?: Server) => {
+      if (!error && operation.hasAspect(Aspect.READ_OPERATION)) {
+        return callback(
+          new MongoUnexpectedServerResponseError('Selected server does not support retryable reads')
```
dariakp, line 201: This block should be deleted: I think you removed the `!supportsRetryableReads` check which would always return false, so L199 would have been equivalent to an `... && false`, i.e., never happens; as is, the current logic will always throw on retrying reads

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    if (errorLabelsOmit) {
+      for (const label of errorLabelsOmit) {
+        expect(thrownError.errorLabels).to.not.contain(label);
+      }
+    }
+  } else if (test.outcome.result) {
+    expect(thrownError, thrownError?.stack).to.not.exist;
+    const expected = test.outcome.result;
+    result = transformToResultValue(result);
+    expect(result).to.deep.include(expected);
```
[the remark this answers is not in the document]
nbbeeken: I'll file a ticket here I think, the chai errors all look similiar to:
```
{ 
  -  "acknowledged": true
  -  "insertedCount": 2
  "insertedIds": {
    "0": 2
    "1": 3
}
```
So we seem to have extra keys

src/operations/execute_operation.ts
```diff
@@ -246,24 +235,17 @@ function executeWithServerSelection(
...
-        !inTransaction &&
-        supportsRetryableReads(server) &&
-        operation.canRetryRead;
-
-      const willRetryWrite =
-        topology.s.options.retryWrites === true &&
-        !inTransaction &&
-        supportsRetryableWrites(server) &&
-        operation.canRetryWrite;
+      const willRetryRead = topology.s.options.retryReads;
```
dariakp, line 244: this and the willRetryWrite blocks should not have semicolons

dariakp (changes requested):
Some partial feedback on the revisions

src/error.ts
dariakp: I think we should be careful about double wrapping things in functions that already have straightforward assertions (like `instanceof`), for performance reasons; I like the basic idea for readability and such, I just think we should consider it more carefully

src/operations/execute_operation.ts
nbbeeken: I suppose we should keep that feature

dariakp (commented):
Just the one additional comment here

src/operations/execute_operation.ts
dariakp: 😂

test/tools/unified-spec-runner/runner.ts
```diff
@@ -187,7 +187,7 @@ export async function runUnifiedTest(
       for (const expectedEventList of test.expectEvents) {
         const clientId = expectedEventList.client;
         const eventType = expectedEventList.eventType;
-        let actualEvents;
+        let actualEvents: CmapEvent[] | CommandEvent[];
```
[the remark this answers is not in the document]
nbbeeken: Yea the unified topology just has an event type flag to determine which, but they are distinct lists. A ternary is warrented here to clearly show it.

src/operations/execute_operation.ts
nbbeeken: Fixed!

test/tools/unified-spec-runner/match.ts
```diff
@@ -270,9 +270,24 @@ function validEmptyCmapEvent(
...
 
+/**
+ * Needed when you write an assertion like this, but TS doesn't know
+ * to narrow the type.
+ * `expect(actualEvent).to.be.instanceOf(CommandFailedEvent);`
+ *
+ * Could we extend chai to do this?
+ */
+function isCommandEvent(ev: any): asserts ev is CommandEvent {
+  console.assert(
```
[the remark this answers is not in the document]
nbbeeken: It just dodged writing an if stmt but it's gone in favor of better chai assertions

test/tools/unified-spec-runner/match.ts
```diff
@@ -300,6 +315,7 @@ export function matchesEvents(
       ]);
     } else if (expectedEvent.commandFailedEvent) {
       expect(actualEvent).to.be.instanceOf(CommandFailedEvent);
+      isCommandEvent(actualEvent);
```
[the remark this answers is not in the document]
nbbeeken: Good idea, done!

src/cmap/connection.ts
```diff
@@ -109,11 +109,12 @@ export interface CommandOptions extends BSONSerializeOptions {
...
-  // FIXME: NODE-2802
-  willRetryWrite?: boolean;
+  // TODO(NODE-2802): Currently the CommandOptions take a property willRetryWrite which is a hint from executeOperation that the txnNum should be applied to this command.
+  // Applying a session to a command should happen as part of command construction, most likely in the CommandOperation#executeCommand method,
+  // where we have access to the details we need to determine if a txnNum should also be applied.
+  willRetryWrite?: true;
 
-  // FIXME: NODE-2781
-  writeConcern?: WriteConcernOptions | WriteConcern | W;
+  writeConcern?: WriteConcern;
```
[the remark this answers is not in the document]
nbbeeken: replied in another thread above but:
I fixed the issue that NODE-2781 talks about kinda, well we fixed it at some other stage, we had a mix of WriteConcern proper or the properties that make up a WriteConcern being passed in, but we fixed that.

test/integration/retryable-writes/retryable_writes.spec.test.ts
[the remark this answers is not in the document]
dariakp: Leaving comment open for relevant TODO tag

test/integration/retryable-writes/retryable_writes.spec.test.ts
dariakp: still planning to do this?

dariakp (changes requested):

src/sessions.ts
nbbeeken: Removed the TODO and simplified the logic here

src/error.ts
nbbeeken: done!

test/unit/error.test.ts
```diff
@@ -424,4 +397,94 @@ describe('MongoErrors', () => {
...
+          error: new MongoError(LEGACY_NOT_WRITABLE_PRIMARY_ERROR_MESSAGE)
+        },
+        {
+          result: true,
+          error: new MongoError('node is recovering')
+        }
+      ];
+
+      for (const { result, error } of tests) {
+        it(`should return ${result} for isRetryableWriteError(${error})`, () => {
```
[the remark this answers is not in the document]
nbbeeken: Added a description prop

test/unit/error.test.ts
```diff
@@ -424,4 +397,94 @@ describe('MongoErrors', () => {
...
+          expect(result).to.be.equal(isRetryableReadError(error));
```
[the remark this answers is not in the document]
nbbeeken: Whoops, yes that would be better, flipped them!

test/unit/error.test.ts
```diff
@@ -424,4 +397,94 @@ describe('MongoErrors', () => {
...
+          error: new MongoWriteConcernError(
+            { message: 'RetryableWriteError error label attached wire version 9+' },
+            { errorLabels: ['RetryableWriteError'] }
+          ),
+          maxWireVersion: 9
+        }
+      ];
+      for (const { result, error, maxWireVersion } of tests) {
+        it(`should return ${result} for isRetryableWriteError(${error}, ${maxWireVersion})`, () => {
+          expect(result).to.be.equal(isRetryableWriteError(error, maxWireVersion));
```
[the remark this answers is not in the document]
nbbeeken: 👍

test/unit/error.test.ts
```diff
@@ -424,4 +397,94 @@ describe('MongoErrors', () => {
...
+          result: true,
+          error: new MongoWriteConcernError(
+            { message: 'RetryableWriteError error label attached wire version 9+' },
+            { errorLabels: ['RetryableWriteError'] }
+          ),
+          maxWireVersion: 9
+        }
+      ];
+      for (const { result, error, maxWireVersion } of tests) {
+        it(`should return ${result} for isRetryableWriteError(${error}, ${maxWireVersion})`, () => {
```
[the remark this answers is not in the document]
nbbeeken: I switched up the language so the it's read like:
`a MongoWriteConcernError with a RetryableWriteError label above server 4.4 is a retryable write`
so it indicates what values are relevant to the tests, I can drop the 2.3's I was just using a low double to indicate it has no relevancy but that could be put in a comment for further clarity.

test/integration/retryable-writes/retryable_writes.spec.test.ts
nbbeeken: Filed a ticket: NODE-4033, the complexity of bulkWrites and findAndModify operations should be done in a follow up

test/integration/retryable-writes/retryable_writes.spec.test.ts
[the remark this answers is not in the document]
nbbeeken: Filed: NODE-4034

test/integration/retryable-writes/retryable_writes.spec.test.ts
durran: I think I just added that in case tests were skipped but I don't see anything currently in retryable writes actually getting skipped.

test/unit/error.test.ts
```diff
@@ -424,4 +397,147 @@ describe('MongoErrors', () => {
...
+          // @ts-expect-error: passing in a plain error to test false case
+          error: new Error('do not retry me!')
+        },
+        {
+          description: 'error code 1',
+          result: false,
+          error: new MongoServerError({ message: '', code: 1 })
+        },
+        {
+          description: 'error code ShutdownInProgress',
```
dariakp, line 495: what's special about ShutdownInProgress? there is no special block to handle this error in the code, so it's strange to call it out

test/unit/error.test.ts
```diff
@@ -424,4 +397,147 @@ describe('MongoErrors', () => {
...
+    describe('#isRetryableReadError', () => {
+      const tests: { description: string; result: boolean; error: MongoError }[] = [
+        {
+          description: 'plain error',
+          result: false,
+          // @ts-expect-error: passing in a plain error to test false case
+          error: new Error('do not retry me!')
+        },
+        {
+          description: 'error code 1',
```
dariakp, line 490: what's special about  error code 1? there is no special block to handle this error in the code, so it's strange to call it out, should this say something about the list of retryable read error codes?

test/unit/error.test.ts
```diff
@@ -424,4 +397,147 @@ describe('MongoErrors', () => {
...
+          result: true,
+          error: new MongoNetworkError('socket bad, try again')
+        },
+        {
+          description: 'error with legacy not writable primary error message',
+          result: true,
+          error: new MongoError(LEGACY_NOT_WRITABLE_PRIMARY_ERROR_MESSAGE)
+        },
+        {
+          description: 'error with node is recovering',
```
dariakp, line 510: should this be "...with node is recovering error message"?

test/unit/error.test.ts
```diff
@@ -424,4 +397,147 @@ describe('MongoErrors', () => {
...
+        }
+      ];
+
+      for (const { description, result, error } of tests) {
+        it(`${description} is${result ? '' : ' not'} a retryable read`, () => {
+          expect(isRetryableReadError(error)).to.be.equal(result);
+        });
+      }
+
+      for (const { description, result, error } of tests) {
```
dariakp, line 522: I don't think we should have retryable write tests inside the `isRetryableReadError` block, that just makes it confusing

test/unit/error.test.ts
```diff
@@ -424,4 +397,147 @@ describe('MongoErrors', () => {
...
+          description: 'a MongoWriteConcernError with an ExceededTimeLimit code above server 4.4',
+          result: false,
+          error: new MongoWriteConcernError({}, { code: 262 }),
+          maxWireVersion: 9
+        },
+        {
+          description: 'a MongoWriteConcernError with an ExceededTimeLimit code below server 4.4',
+          result: true,
+          error: new MongoWriteConcernError({}, { code: 262 }),
+          maxWireVersion: 2.3
```
dariakp, line 447: "below server 4.4" should test the boundary, so use maxWireVersion: 8, otherwise these would all still pass even if we changed the condition in the errors.ts code to test for, say `maxWireVersion: 6`

this goes for all of the examples above (and varying 2.3 and 0 just makes it more confusing to understand the intent); I am also wondering if there should be cases for both above and below the maxWireVersion: 9 for each case where the maxWireVersion should not be relevant, just to prove that it isn't

test/unit/error.test.ts
```diff
@@ -424,4 +397,147 @@ describe('MongoErrors', () => {
...
+          maxWireVersion: 0
+        },
+        {
+          description: 'a MongoWriteConcernError with an ExceededTimeLimit code above server 4.4',
+          result: false,
+          error: new MongoWriteConcernError({}, { code: 262 }),
+          maxWireVersion: 9
+        },
+        {
+          description: 'a MongoWriteConcernError with an ExceededTimeLimit code below server 4.4',
```
dariakp, line 444: If ExceededTimeLimit code is supposed to represent a retryable write code, the test description should reflect that instead of calling attention to a specific label

dariakp (changes requested):

src/error.ts
```diff
@@ -403,6 +414,32 @@ export class MongoGridFSChunkError extends MongoRuntimeError {
...
+ * When an operation fails, it is the driver's job to retry it. It must perform serverSelection
+ * again to make sure that it attempts the operation against a server in a good state. If server
+ * selection returns a server that does not support retryable operations, this error is used.
+ * This scenario is unlikely as retryable support would also have been determined on the first attempt
+ * but it is possible the state change could report a selectable server that does not support retries.
+ *
+ * @public
+ * @category Error
+ */
+export class MongoUnexpectedServerResponseError extends MongoRuntimeError {
```
baileympearson, line 433: I haven't looked closely,  but I would guess that even before this PR our generated "error tree" svg used in our error documentation is out of date (see https://github.com/mongodb/node-mongodb-native/tree/main/docs/charts).  

It's out of scope to in your work to get them back into sync but would you mind adding your new error to the generate script, generating a new SVG and including those as changes in your PR?

src/operations/execute_operation.ts
```diff
@@ -94,63 +96,51 @@ export function executeOperation<
...
-    return;
   }
 
-  if (
-    session &&
-    session.isPinned &&
-    session.transaction.isCommitted &&
-    !operation.bypassPinningCheck
-  ) {
+  if (session?.isPinned && session?.transaction.isCommitted && !operation.bypassPinningCheck) {
```
baileympearson, line 143: How come the optional chaining is necessary here?  Session is of ClientSession and shouldn't be undefined, so we shouldn't need to check for its existence.

nbbeeken: Apparently this was the source of the CI difficulties I've been having, despite TS claiming the session is always present removing the checks for existence fail a ton of the CSFLE tests but they only fail on evergreen. 🤔

test/unit/error.test.ts
nbbeeken: Fixed

test/unit/error.test.ts
nbbeeken: Fixed, used 8/9

test/unit/error.test.ts
nbbeeken: Fixed

test/unit/error.test.ts
nbbeeken: Yep fixed!

test/unit/error.test.ts
nbbeeken: fixed, it's just a code that isn't retryable, description says so now

test/unit/error.test.ts
nbbeeken: fixed.

test/integration/retryable-writes/retryable_writes.spec.prose.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const error = await client
+      .db('test')
+      .collection('test')
+      .insertOne({ a: 1 })
+      .catch(error => error);
+
+    expect(error).to.exist;
+    expect(error).to.be.instanceOf(MongoServerError);
+    expect(error).to.have.property('originalError').that.instanceOf(MongoError);
```
dariakp, line 57: nit: can we make it `that.is.instanceOf` for the sake of grammar?

dariakp (dismissed):
Putting an approve because these are minor:

test/unit/error.test.ts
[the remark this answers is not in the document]
dariakp: is it "above 4.4" or "at least 4.4"?

src/utils.ts
```diff
@@ -35,8 +35,6 @@ import { W, WriteConcern, WriteConcernOptions } from './write_concern';
  * @public
  */
 export type Callback<T = any> = (error?: AnyError, result?: T) => void;
```
baileympearson, line 37: If this type is exported, is this technically a breaking change?

baileympearson (changes requested):

global.d.ts
```diff
@@ -42,6 +50,9 @@ interface MetadataAndTest<Fn> {
 
     interface Test {
       metadata: MongoDBMetadataUI;
+
+      /** @deprecated Attach spec to a test if you need access to it in a beforeEach hook, not recommended?? */
```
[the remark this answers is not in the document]
nbbeeken: I actually think we may adopt this pattern for other programatic tests, so I'm going to remove the comment altogether.

test/unit/error.test.ts
```diff
@@ -424,4 +399,136 @@ describe('MongoErrors', () => {
       });
     });
   });
+
+  describe('retryable errors', () => {
+    describe('#needsRetryableWriteLabel', () => {
+      // Note the wireVersions used below are used to represent
+      // 8 - below server version 4.4
```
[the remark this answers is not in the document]
nbbeeken: Good idea

test/integration/retryable-writes/retryable_writes.spec.prose.test.ts
nbbeeken: sure :)

src/utils.ts
nbbeeken: It was never used on any of our public APIs but yes, if we feel like users may have imported this type we should revert this. Should I deprecate it perhaps?

src/operations/execute_operation.ts
```diff
@@ -94,63 +96,51 @@ export function executeOperation<
...
-    return;
   }
 
-  if (
-    session &&
-    session.isPinned &&
-    session.transaction.isCommitted &&
-    !operation.bypassPinningCheck
-  ) {
+  if (session?.isPinned && session.transaction.isCommitted && !operation.bypassPinningCheck) {
```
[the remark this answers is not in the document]
nbbeeken: Interesting the typescript still wasn't fixed for this function would that help? the current typing still claimed it always existed, I'm fixing that now.

src/cmap/connection.ts
```diff
@@ -109,11 +109,12 @@ export interface CommandOptions extends BSONSerializeOptions {
   noResponse?: boolean;
   omitReadPreference?: boolean;
 
-  // FIXME: NODE-2802
-  willRetryWrite?: boolean;
+  // TODO(NODE-2802): Currently the CommandOptions take a property willRetryWrite which is a hint from executeOperation that the txnNum should be applied to this command.
```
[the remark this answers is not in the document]
nbbeeken: done!

PR #3149, opened 2022-02-22
fix(NODE-3948): Add error code to MongoSystemError

baileympearson:
#### What is changing?

This PR adds an error code property to the `MongoSystemError` class.

This PR also converts the error tests to Typescript.

No.


test/unit/error.test.ts
```diff
@@ -89,6 +91,30 @@ describe('MongoErrors', () => {
...
+        } as any as EverythingFromDriver.TopologyDescription;
+
+        const error = new MongoSystemError('something went wrong', topologyDescription);
+        expect(error).to.haveOwnProperty('code', 123);
+      });
+    });
+
+    context('when the topology description does not contain a code', () => {
+      it('contains the code as a top level property', () => {
+        const topologyDescription = {} as any as EverythingFromDriver.TopologyDescription;
```
dariakp, line 110: I think it might be worthwhile to also test case where the topologyDescription contains an error prop that does not have a code, just to cover all our bases

test/unit/error.test.ts
```diff
@@ -1,48 +1,50 @@
...
+import { ReplSetFixture } from '../tools/common';
+import { cleanup } from '../tools/mongodb-mock/index';
+import { getSymbolFrom } from '../tools/utils';
 
 describe('MongoErrors', () => {
-  // import errors as object
   let errorClasses = Object.fromEntries(
-    Object.entries(require('../../src/index')).filter(([key]) => key.endsWith('Error'))
-  );
+    Object.entries(EverythingFromDriver).filter(([key]) => key.endsWith('Error'))
```
dariakp, line 33: let's try to square this approach with #3144 to minimize merge conflicts

dariakp (changes requested):
just a couple of things

test/unit/error.test.ts
```diff
@@ -89,6 +115,39 @@ describe('MongoErrors', () => {
...
+          }
+        } as any as TopologyDescription;
+
+        const error = new MongoSystemError('something went wrong', topologyDescription);
+        expect(error).to.haveOwnProperty('code', 123);
+      });
+    });
+
+    context('when the topology description does not contain a code', () => {
+      it('contains the code as a top level property', () => {
```
dariakp, line 133:
```suggestion
      it('contains code as a top level property that is undefined', () => {
```

test/unit/error.test.ts
```diff
@@ -89,6 +115,39 @@ describe('MongoErrors', () => {
...
+        const topologyDescription = { error: {} } as any as TopologyDescription;
+
+        const error = new MongoSystemError('something went wrong', topologyDescription);
+        expect(error).to.haveOwnProperty('code', undefined);
+      });
+    });
+
+    context('when the topology description does not contain an error property', () => {
+      it('contains the code as a top level property', () => {
```
dariakp, line 142:
```suggestion
      it('contains code as a top level property that is undefined', () => {
```

test/unit/error.test.ts
```diff
@@ -89,6 +115,39 @@ describe('MongoErrors', () => {
     });
   });
 
+  describe('MongoSystemError#constructor', () => {
+    context('when the topology description contains a code', () => {
+      it('contains the code as a top level property', () => {
```
dariakp, line 120:
```suggestion
      it('contains the specified code as a top level property', () => {
```

dariakp (changes requested):
just a couple of nits, otherwise lgtm

test/unit/error.test.ts
```diff
@@ -1,53 +1,79 @@
...
-        let error = new errorClasses[errorName]('generated by test');
+        const error = new (errorClass as any)('generated by test', {});
         // expect name property to be class name
         expect(error).to.have.property('name', errorName);
 
         try {
           error.name = 'renamed by test';
+          // eslint-disable-next-line no-empty
         } catch (err) {}
         expect(error).to.have.property('name', errorName);
```
nbbeeken, line 73: Some drive by improvement we could consider, is being more direct about the assertions? I notice we have the eslint disable down here but the one at the top of the file is commented out. Maybe we remove the need for either.

```suggestion
        expect(errorClass.prototype).to.have.ownPropertyDescriptor('name')
          .that.has.property('set', undefined);
        expect(errorClass.prototype).to.have.ownPropertyDescriptor('name')
          .that.has.property('get').that.is.a('function');
```

nbbeeken (changes requested):

test/unit/error.test.ts
dariakp: This assertion suggestion reads a bit weird to me, I feel like expecting a name property of a specified value is easier to understand than indirect references to getters and setters... the important thing, after all, is that the error has a name, not that it's specifically the outcome of a getter (imho)

nbbeeken: I feel like we had this exact convo when we first merged these tests 😅 , I'm fine with it as is too, if so then the ask would be to remove the eslint comment at the top of the file

baileympearson: Done!

dariakp (dismissed):
LGTM, thanks!

test/unit/error.test.ts
```diff
@@ -254,14 +314,14 @@ describe('MongoErrors', () => {
...
-    afterEach(() => mock.cleanup());
+    afterEach(() => cleanup());
     beforeEach(() => test.setup());
 
     function makeAndConnectReplSet(cb) {
       let invoked = false;
       const replSet = new Topology(
         [test.primaryServer.hostAddress(), test.firstSecondaryServer.hostAddress()],
-        { replicaSet: 'rs' }
+        { replicaSet: 'rs' } as any
```
nbbeeken, line 324: `//@ts-expect-error` goes here too right?

nbbeeken (changes requested):

test/unit/error.test.ts
baileympearson: Actually, no because `rs` is the correct type.  The better option is to directly cast as `TopologyOptions`, which I fixed 😄

nbbeeken (approved):

dariakp (approved):

PR #3151, opened 2022-02-24
fix(NODE-3521): update session support checks

durran:
### Description

With the increase of the minimum wire version, server versions below 3.6 are no longer supported and thus the need to check if a topology supports sessions is no longer needed. This check also caused session mismatch errors in rare cases when a topology went unknown before a cursor's initial `find` command executed and came back before the subsequent `getMore` command was executed.

#### What is changing?

Removes all `hasSessionSupport` checks and tests that are no longer relevant.

No

NODE-3521 / NODE-4009

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -653,8 +653,15 @@ function next<T>(cursor: AbstractCursor, blocking: boolean, callback: Callback<T
 
   if (cursorId == null) {
     // All cursors must operate within a session, one must be made implicitly if not explicitly provided
-    if (cursor[kSession] == null && cursor[kTopology].hasSessionSupport()) {
-      cursor[kSession] = cursor[kTopology].startSession({ owner: cursor, explicit: false });
+    if (cursor[kSession] == null) {
+      if (cursor[kTopology].shouldCheckForSessionSupport()) {
+        return cursor[kTopology].selectServer(ReadPreference.primaryPreferred, err => {
+          if (err) return callback(err);
+          return next(cursor, blocking, callback);
+        });
+      } else if (cursor[kTopology].hasSessionSupport()) {
+        cursor[kSession] = cursor[kTopology].startSession({ owner: cursor, explicit: false });
+      }
     }
 
     cursor._initialize(cursor[kSession], (err, state) => {
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -653,7 +653,7 @@ function next<T>(cursor: AbstractCursor, blocking: boolean, callback: Callback<T
 
   if (cursorId == null) {
     // All cursors must operate within a session, one must be made implicitly if not explicitly provided
-    if (cursor[kSession] == null && cursor[kTopology].hasSessionSupport()) {
+    if (cursor[kSession] == null) {
       cursor[kSession] = cursor[kTopology].startSession({ owner: cursor, explicit: false });
     }
 
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -653,7 +653,7 @@ function next<T>(cursor: AbstractCursor, blocking: boolean, callback: Callback<T
 
   if (cursorId == null) {
     // All cursors must operate within a session, one must be made implicitly if not explicitly provided
-    if (cursor[kSession] == null) {
+    if (cursor[kSession] == null && cursor[kTopology].hasSessionSupport()) {
       cursor[kSession] = cursor[kTopology].startSession({ owner: cursor, explicit: false });
     }
 
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -89,13 +89,19 @@ export function executeOperation<
     // that are not explicitly provided with a session.
     let session: ClientSession | undefined = operation.session;
     let owner: symbol | undefined;
-    if (session == null) {
-      owner = Symbol();
-      session = topology.startSession({ owner, explicit: false });
-    } else if (session.hasEnded) {
-      return cb(new MongoExpiredSessionError('Use of expired sessions is not permitted'));
-    } else if (session.snapshotEnabled && !topology.capabilities.supportsSnapshotReads) {
-      return cb(new MongoCompatibilityError('Snapshot reads require MongoDB 5.0 or later'));
+    if (topology.hasSessionSupport()) {
+      if (session == null) {
+        owner = Symbol();
+        session = topology.startSession({ owner, explicit: false });
+      } else if (session.hasEnded) {
+        return cb(new MongoExpiredSessionError('Use of expired sessions is not permitted'));
+      } else if (session.snapshotEnabled && !topology.capabilities.supportsSnapshotReads) {
+        return cb(new MongoCompatibilityError('Snapshot reads require MongoDB 5.0 or later'));
+      }
+    } else if (session) {
+      // If the user passed an explicit session and we are still, after server selection,
+      // trying to run against a topology that doesn't support sessions we error out.
+      return cb(new MongoCompatibilityError('Current topology does not support sessions'));
     }
 
     try {
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -77,6 +77,14 @@ export function executeOperation<
   }
 
   return maybePromise(callback, cb => {
+    if (topology.shouldCheckForSessionSupport()) {
+      return topology.selectServer(ReadPreference.primaryPreferred, err => {
+        if (err) return cb(err);
+
+        executeOperation<T, TResult>(topology, operation, cb);
+      });
+    }
+
     // The driver sessions spec mandates that we implicitly create sessions for operations
     // that are not explicitly provided with a session.
     let session: ClientSession | undefined = operation.session;
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -89,19 +89,13 @@ export function executeOperation<
     // that are not explicitly provided with a session.
     let session: ClientSession | undefined = operation.session;
     let owner: symbol | undefined;
-    if (topology.hasSessionSupport()) {
-      if (session == null) {
-        owner = Symbol();
-        session = topology.startSession({ owner, explicit: false });
-      } else if (session.hasEnded) {
-        return cb(new MongoExpiredSessionError('Use of expired sessions is not permitted'));
-      } else if (session.snapshotEnabled && !topology.capabilities.supportsSnapshotReads) {
-        return cb(new MongoCompatibilityError('Snapshot reads require MongoDB 5.0 or later'));
-      }
-    } else if (session) {
-      // If the user passed an explicit session and we are still, after server selection,
-      // trying to run against a topology that doesn't support sessions we error out.
-      return cb(new MongoCompatibilityError('Current topology does not support sessions'));
+    if (session == null) {
+      owner = Symbol();
+      session = topology.startSession({ owner, explicit: false });
+    } else if (session.hasEnded) {
+      return cb(new MongoExpiredSessionError('Use of expired sessions is not permitted'));
+    } else if (session.snapshotEnabled && !topology.capabilities.supportsSnapshotReads) {
+      return cb(new MongoCompatibilityError('Snapshot reads require MongoDB 5.0 or later'));
     }
 
     try {
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -77,14 +77,6 @@ export function executeOperation<
   }
 
   return maybePromise(callback, cb => {
-    if (topology.shouldCheckForSessionSupport()) {
-      return topology.selectServer(ReadPreference.primaryPreferred, err => {
-        if (err) return cb(err);
-
-        executeOperation<T, TResult>(topology, operation, cb);
-      });
-    }
-
     // The driver sessions spec mandates that we implicitly create sessions for operations
     // that are not explicitly provided with a session.
     let session: ClientSession | undefined = operation.session;
--- a/test/unit/cursor/aggregation_cursor.test.js
+++ b/test/unit/cursor/aggregation_cursor.test.js
@@ -1,15 +1,14 @@
 'use strict';
 
 const expect = require('chai').expect;
-const { MongoError } = require('../../../src/error');
 const mock = require('../../tools/mongodb-mock/index');
 const { Topology } = require('../../../src/sdam/topology');
 const { Long } = require('bson');
 const { MongoDBNamespace, isHello } = require('../../../src/utils');
-const { FindCursor } = require('../../../src/cursor/find_cursor');
+const { AggregationCursor } = require('../../../src/cursor/aggregation_cursor');
 
 const test = {};
-describe('Find Cursor', function () {
+describe('Aggregation Cursor', function () {
   describe('#next', function () {
     afterEach(function () {
       mock.cleanup();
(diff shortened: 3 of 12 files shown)


nbbeeken (commented):
Some open Qs

src/operations/execute_operation.ts
```diff
@@ -89,19 +89,13 @@ export function executeOperation<
     // that are not explicitly provided with a session.
     let session: ClientSession | undefined = operation.session;
     let owner: symbol | undefined;
```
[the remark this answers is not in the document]
durran: Actually that's a good catch. I think that extra run through server selection should maybe be renamed - as if we don't have any data bearing servers or known servers in the description we want to force the monitor check. I think it's not to check for session support but to check at the beginning if we have something to actually execute the operation against.

src/operations/execute_operation.ts
```diff
@@ -89,19 +89,13 @@ export function executeOperation<
...
-    } else if (session) {
-      // If the user passed an explicit session and we are still, after server selection,
-      // trying to run against a topology that doesn't support sessions we error out.
-      return cb(new MongoCompatibilityError('Current topology does not support sessions'));
+    if (session == null) {
+      owner = Symbol();
+      session = topology.startSession({ owner, explicit: false });
+    } else if (session.hasEnded) {
+      return cb(new MongoExpiredSessionError('Use of expired sessions is not permitted'));
+    } else if (session.snapshotEnabled && !topology.capabilities.supportsSnapshotReads) {
```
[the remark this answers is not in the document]
durran: If any of those mongolike services did give us a `logicalSessionTimeoutMinutes` in the previous code we would have still attached a session so the behaviour in this case would be the same with or without this change.

src/operations/execute_operation.ts
```diff
@@ -89,19 +89,13 @@ export function executeOperation<
...
-        return cb(new MongoCompatibilityError('Snapshot reads require MongoDB 5.0 or later'));
-      }
-    } else if (session) {
-      // If the user passed an explicit session and we are still, after server selection,
-      // trying to run against a topology that doesn't support sessions we error out.
-      return cb(new MongoCompatibilityError('Current topology does not support sessions'));
+    if (session == null) {
+      owner = Symbol();
+      session = topology.startSession({ owner, explicit: false });
+    } else if (session.hasEnded) {
```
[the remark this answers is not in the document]
durran: With these changes we assume `logicalSessionTimeoutMinutes` is always there and a number, and if it is not that's a server bug. With respect to the default of 10 minutes - I'm not sure where that came from and it's not in the spec. The spec only states a 30 minute timeout on the server side which is configurable on the server for when the session is idle - so if the session is not idle I'd imagine for long running operations like ETL jobs they could be valid for very long durations (especially with explicit sessions). I looked in the Java driver and they default it to 30, so I'm not sure why we default to 10. Maybe a question for next week when Jeff is back.

src/operations/execute_operation.ts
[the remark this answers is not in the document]
nbbeeken: But if they did not give us a `logicalSessionTimeoutMinutes` then we wouldn't have attached one but now we will regardless, usually "extra" data is harmless but it could have implications where folks were relying on sessions being specifically omitted with that case

src/operations/execute_operation.ts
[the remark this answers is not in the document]
nbbeeken: That's not an assumption that is safe to make I think, the rule of determining `logicalSessionTimeoutMinutes` is that there must be data bearing servers, and you pick the minimum between all nodes, unless _any_ node has a null value than you set it to null for the whole topology. 

Using our mlaunch helper script I can get an undefined logicalSessionTimeoutMinutes like so:
```
> var client = await mdb.MongoClient.connect('mongodb://bob:pwd123@localhost:31003?directConnection=true')
> client.topology.logicalSessionTimeoutMinutes
undefined
```
The node at 31003 is an arbiter

src/operations/execute_operation.ts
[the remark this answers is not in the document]
nbbeeken: See my comment on the other thread, but to expand, as you said one of the conditions of `shouldCheckForSessionSupport` is that there are data bearing servers, I think that we should still both be checking for that and not sending a session if there are no data bearing servers in the topology.

src/operations/execute_operation.ts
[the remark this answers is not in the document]
durran: My thought here was that if there weren't any data bearing servers then server selection would fail when executing the command and there would be no need for the extra server selection to happen ahead of time, since its purpose was only to determine session support. We would not select an arbiter, unknown, rsghost , or rsother to run a command against - the only thing getting sent there is the `hello` from the monitor which does not send a session. If any data bearing server in the cluster had a null `logicalSessionTimeoutMinutes` that would indicate to me that it was a mixed version cluster and that node was pre 3.6 which would error now with the wire version bump. But it seems there a lot of potential unknowns here after this was all brought up. So I'm thinking I just roll back this PR to the single spot to fix this issue - the check in the cursor. (And just leave the remaining checks in for the time being). Thoughts?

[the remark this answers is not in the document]
nbbeeken: Yea that sounds good to me, and actually another thing we could try is instead of removing the check what about reusing your "reselect same server" logic from the getMore fix awhile back inside next before we call hasSessionSupport. The problem I think originates from the fact that `hasSessionSupport` is a value that changes depending on the current state of the topology, running a server selection will make sure that value is updated before the code proceeds, the other option is that hasSessionSupport is something that we might want to refactor to be async, but I think the selection change would be easier for now. Want to give that a try?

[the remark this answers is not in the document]
nbbeeken: When direct connecting to an arbiter on this PR vs main, there is an LSID being attached even when logicalSessionTimeoutMinutes is null, "server selection would fail when executing the command" isn't going to be true for all commands, the point of connecting to an arbiter is to do management commands on it, So I think we need to only change the cursor logic to not impact all commands / topology scenarios
```
NODE-3521 {
  ping: 1,
  lsid: {
    id: new Binary(Buffer.from("80990b6ec8d4451082d3c3361e901479", "hex"), 4)
  },
  '$db': 'test'
}
```

```
main branch { ping: 1, '$db': 'test' }
```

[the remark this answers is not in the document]
durran: I liked the idea of doing server selection in this case so I reverted all the changes and introduced the `shouldCheckForSessionSupport` logic into the cursor's `next` method. Selecting the same server wouldn't be valid in this edge case as no server has actually been selected yet (the initial find command hasn't been run) but I think before we check for session support we always need to run server selection if the topology is in a state that it cannot be determined. I also think this brings us in line with the spec - we always should do this when checking.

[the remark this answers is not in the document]
durran: I've also just verified that this latest change also addresses the issue on lambda.

src/cursor/abstract_cursor.ts
```diff
@@ -653,8 +653,15 @@ function next<T>(cursor: AbstractCursor, blocking: boolean, callback: Callback<T
 
   if (cursorId == null) {
     // All cursors must operate within a session, one must be made implicitly if not explicitly provided
-    if (cursor[kSession] == null && cursor[kTopology].hasSessionSupport()) {
-      cursor[kSession] = cursor[kTopology].startSession({ owner: cursor, explicit: false });
+    if (cursor[kSession] == null) {
+      if (cursor[kTopology].shouldCheckForSessionSupport()) {
+        return cursor[kTopology].selectServer(ReadPreference.primaryPreferred, err => {
```
nbbeeken, line 658: I'm not sure about primaryPreferred here, I think we should make use of `this[kServer]`, and do the self selection thing we did for getMore, if selecting our saved server fails then we can't proceed anyway, correct?

nbbeeken (commented):
Q about the selection

src/cursor/abstract_cursor.ts
```diff
@@ -653,8 +653,15 @@ function next<T>(cursor: AbstractCursor, blocking: boolean, callback: Callback<T
 
   if (cursorId == null) {
     // All cursors must operate within a session, one must be made implicitly if not explicitly provided
-    if (cursor[kSession] == null && cursor[kTopology].hasSessionSupport()) {
-      cursor[kSession] = cursor[kTopology].startSession({ owner: cursor, explicit: false });
+    if (cursor[kSession] == null) {
+      if (cursor[kTopology].shouldCheckForSessionSupport()) {
```
nbbeeken, line 657: I think this is potentially not relevant, I think we always want to do the "no-op" self server selection to make sure the logicalSessionTimeoutMinutes property is correctly set by the monitor. 

Can we add some testing, maybe mock the conditions, if we start by stating the scenario it could help this code from regressing, the cursor has to be constructed without an implicit session, we could just do that part by manually constructing a cursor. Then calling next should enter serverSelection because kSession is null. And on the  recursive call of next we'll attach a session, we could just assert the session exists at the end I don't think we need to listen to the recursive call or anything that fancy.

src/cursor/abstract_cursor.ts
jyemin: Will `this[kServer]` be non-null when `cursorId` is null?

jyemin: primaryPreferred is correct when connected to a replica set, but it's not clear to me how selectServer will behave when the cluster type is sharded or single.  Will a mongos or standalone satisfy this read preference?

nbbeeken: Ah if the cursorId is null then we haven't contacted the server yet, so `this[kServer]` will be null, good catch, then I suppose we should just respect the readPreference that's in the cursor options. It actually seems like this is not the place we should be starting the session, maybe that's the core of the bug here, executeOperation that's called in FindCursor's _initialize should handle implicit session logic, but we're creating it here in the cursor in an "explicit" style.

jyemin: The cursor's read preference is not the right one to use (it's likely to be `primary`, but any primary or secondary will do.

nbbeeken: Okay yea I see now, then this is the right approach, ty for chiming in!

src/cursor/abstract_cursor.ts
```diff
@@ -653,8 +653,15 @@ function next<T>(cursor: AbstractCursor, blocking: boolean, callback: Callback<T
...
+        return cursor[kTopology].selectServer(ReadPreference.primaryPreferred, err => {
+          if (err) return callback(err);
+          return next(cursor, blocking, callback);
+        });
+      } else if (cursor[kTopology].hasSessionSupport()) {
+        cursor[kSession] = cursor[kTopology].startSession({ owner: cursor, explicit: false });
+      }
     }
 
     cursor._initialize(cursor[kSession], (err, state) => {
```
nbbeeken, line 667: Is it true that cursor[kSession] will ever be null / undefined here? I noticed some spread logic in FindCursor that might be overriding a session from the options:

```ts
   const findOperation = new FindOperation(undefined, this.namespace, this[kFilter], {
      ...this[kBuiltOptions], // NOTE: order matters here, we may need to refine this
      ...this.cursorOptions,
      session
    });
```

I think if we are able to change this function to always take a defined session then we can be sure that this ^ isn't a bug.

src/cursor/abstract_cursor.ts
durran: When `cursorId` is null, `this[kServer]` is null. The problem with using `this[kServer]` is this issue happens *before* the `find` operation executes - we have not selected a server yet.

src/cursor/abstract_cursor.ts
durran: I've added 3 unit tests specifically on `next()`:

1. A session is set if there is a data bearing server in the cluster
2. A session is set if a data bearing server becomes available as a result of the server selection
3. No session is set if there is no data bearing server in the cluster.

As for the point about the check for session support not necessarily being relevant - I do agree that if we removed that then this would work in theory but if would produce an infinite loop on a successful server selection as `selectServer` takes a callback so we recurse on `next` when that finishes.

src/cursor/abstract_cursor.ts
durran: Good catch. With these changes we would never get to this code with an undefined session, so I've changed it both in  `FindCursor` and `AggregateCursor` to never accept `undefined`.

jyemin: This confuses me.  If the session is never undefined, why is the code in abstract_cursor that handles a null session still necessary?

durran: It handles both cases (null and undefined) as we are checking loose equality with `==`.

nbbeeken (approved):

test/unit/cursor/find_cursor.test.js
```diff
@@ -10,6 +10,115 @@ const { FindCursor } = require('../../../src/cursor/find_cursor');
...
+      });
+    });
+
+    context('when a data bearing server becomes available', function () {
+      beforeEach(function () {
+        let helloCalls = 0;
+        test.server.setMessageHandler(request => {
+          const doc = request.document;
+          if (isHello(doc)) {
+            request.reply(helloCalls > 0 ? { errmsg: 'network error' } : mock.HELLO);
```
dariakp, line 93: if I'm reading this correctly, the server is responding with the right response on the very first request, and then with an error on subsequent ones... how does that simulate a data bearing server becoming available? I thought we'd expect to first see an error and then a success for that case. It could be nice to have something here asserting that the expected requests are being made in the expected order or at least some indication of what the expected requests are for each scenario

test/unit/cursor/aggregation_cursor.test.js
```diff
@@ -0,0 +1,135 @@
...
+      });
+    });
+
+    context('when a data bearing server becomes available', function () {
+      beforeEach(function () {
+        let helloCalls = 0;
+        test.server.setMessageHandler(request => {
+          const doc = request.document;
+          if (isHello(doc)) {
+            request.reply(helloCalls > 0 ? { errmsg: 'network error' } : mock.HELLO);
```
dariakp, line 102: same question here as in the other file and same comment about clarifying the expected order of events, and, ideally asserting on it so that we know that the test is actually creating the expected conditions

dariakp (changes requested):

test/unit/cursor/aggregation_cursor.test.js
```diff
@@ -0,0 +1,140 @@
+'use strict';
+
+const expect = require('chai').expect;
```
baileympearson, line 3:
```suggestion
const { expect } = require('chai');
const { createServer, cleanup, isHello } = require('../../tools/mongodb-mock/index')
```

test/unit/cursor/aggregation_cursor.test.js
```diff
@@ -0,0 +1,140 @@
...
+const { MongoDBNamespace, isHello } = require('../../../src/utils');
+const { AggregationCursor } = require('../../../src/cursor/aggregation_cursor');
+
+const test = {};
+describe('Aggregation Cursor', function () {
+  describe('#next', function () {
+    afterEach(function () {
+      mock.cleanup();
+    });
+    beforeEach(function () {
```
baileympearson, line 16:
```suggestion
beforeEach(async function() {
    test.server = await mock.createServer();
})
```

test/unit/cursor/find_cursor.test.js
```diff
@@ -10,6 +10,120 @@ const { FindCursor } = require('../../../src/cursor/find_cursor');
 
 const test = {};
 describe('Find Cursor', function () {
+  describe('#next', function () {
+    afterEach(function () {
+      mock.cleanup();
+    });
+    beforeEach(function () {
```
baileympearson, line 17:
```suggestion
beforeEach(async function() {
    test.server = await mock.createServer();
})
```

baileympearson (changes requested):

dariakp (approved):
I agree with Bailey's cleanup suggestions but otherwise LGTM

dariakp (approved):

baileympearson (approved):

PR #3153, opened 2022-02-25
chore(NODE-4018): add doc generation scripts and template to main

baileympearson:
#### What is changing?

This PR adds scripts needed to generate API docs into the main branch.

- The Hugo static site template used to host the docs is included in this PR.  The only file that was changed in the site template is `etc/docs/template/data/releases.toml`.  Everything else in `etc/docs/template` can be ignored.
- Two new packages were installed as dev dependencies.  I installed typedoc (because we already used it in our build:docs script, but it wasn't a dependency) and I installed a toml parser / serializer, used to update the static site template data when generating docs.
- The contents of `docs` was moved.
  - The images and charts were moved into the `etc` folder.
  - The `.md` files were moved into the top level of the repo.

Once this PR merges, a second PR will copy the contents of `gh-pages` to the `docs` folder.

Link to sample docs: https://baileympearson.github.io/node-mongodb-native/.

No.


etc/docs/template/data/releases.toml
```diff
@@ -0,0 +1,107 @@
+current = "4.4 Driver"
+mongoDBManual = "https://docs.mongodb.com/drivers/node/"
+
+[[versions]]
+version = "4.4 Driver"
+status = "current"
+api = "./4.4"
+usesMongoDBManual = true
+semvarVersion = "4.4"
```
dariakp, line 9: I noticed that this is `semvar` instead of `semver` is that how it was originally? what were the changes that were made to this file vs the original one?

package.json
```diff
@@ -77,6 +78,7 @@
     "standard-version": "^9.3.2",
     "ts-node": "^10.4.0",
     "tsd": "^0.19.1",
+    "typedoc": "^0.22.12",
```
dariakp, line 81: I recall Neal mentioning that this didn't play well with npm due to strict TS version requirements, is this no longer the case?

etc/docs/template/README.md
```diff
@@ -0,0 +1,28 @@
+# MongoDB Java Driver Front page
```
dariakp, line 1: why is there stuff about the java driver here?

etc/docs/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+- `generate-docs.ts` is used to generate API docs for a major or minor release.
+
+### Dependencies
+
+`generate-docs.ts` requires the following in addition to dependencies installed with `npm i`:
+
+* Hugo static web generator `v0.30.2`
+  * You can download the right version [here](https://github.com/gohugoio/hugo/releases/tag/v0.30.2)
+* typedoc
```
dariakp, line 19: this was added to dev dependencies, will it need to be installed separately, too, or should it not be in dev dependencies?

etc/docs/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+* jsdoc v3
+* node (v6.x or v8.x) and npm (>= 3.x). Note: worked for me with 8.17.0, but not 8.0.0.
+* python sphinx
+
+### Usage
+
+To generate API documentation for a new major or minor version:
+
+#### Generate Documentation for the new Version
```
dariakp, line 32:
```suggestion
#### Generate Documentation for the New Version
```

etc/docs/generate-docs.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  // api - the name of the folder for the generated documentation to live.  typically should be version_id prefixed with `./`
+  //  ex: ./4.5
+  api: './4.5',
+
+  // usesMongoDBManual - if true, includes a link to the mongodb documentation for the new version
+  usesMongoDBManual: true,
+
+  // the 
+  semvarVersion: '4.5'
```
dariakp, line 51: should this be semver?

etc/docs/README.md
```diff
@@ -0,0 +1,54 @@
...
+#### Push the changes to Github
+
+Finally, push the new documentation to Github and confirm that the site builds as expected.
+
+### Doc Generation Overview
+
+API docs are stored in the top level `docs` folder in the `node-mongodb-driver` Github repository and are hosted on Github Pages.  Each major or minor release,
+a new set of API docs are generated and added to our Hugo static site.  This is a three step process:
+
+1. Generate the new API docs using type doc
```
dariakp, line 52:
```suggestion
1. Generate the new API docs using typedoc
```

etc/docs/README.md
```diff
@@ -0,0 +1,54 @@
...
+#### Update the Node Driver Docs
+
+After generating the API docs for a particular version, the next step is to update the static site to include the new docs.  Open the `generate-docs.ts` 
+script and update the `NEW_VERSION` object with information for the new version.
+
+Run the `generate-docs.ts` script from the `docs_utils` directory.
+
+#### Push the changes to Github
+
+Finally, push the new documentation to Github and confirm that the site builds as expected.
```
dariakp, line 45: Should this just say "commit the changes", create a PR, and once it's merged into main, that will update the docs?

etc/docs/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Doc Generation Overview
+
+API docs are stored in the top level `docs` folder in the `node-mongodb-driver` Github repository and are hosted on Github Pages.  Each major or minor release,
+a new set of API docs are generated and added to our Hugo static site.  This is a three step process:
+
+1. Generate the new API docs using type doc
+2. Run the generate docs script to create the new version of the Hugo site, copy in the new documentation and move the newly generated site into the `docs` folder.
+3. Push the changes to main
```
dariakp, line 54: I don't think we should skip the PR step here

etc/docs/generate-docs.ts
```diff
@@ -0,0 +1,121 @@
...
+  versions: VersionSchema[]
+}
+
+/**
+ * ATTENTION - EDIT THIS BEFORE RELEASING
+ */
+const NEW_VERSION: VersionSchema = {
+  // version - the display name for the version of the driver on the docs website
+  //  ex: Driver 4.5
+  version: 'Driver 4.5',
```
dariakp, line 38: Why are we switching this from `<version> Driver` to `Driver <version>`? Won't it look a bit jarring on the index page? Also, if we want to default to generating for the "next" driver version `Next Driver` would probably read better than `Driver Next`, so if we end up taking an argument for this ``` `${semVer} Driver` ``` would probably be better to use as the standard than the other way around

etc/docs/README.md
```diff
@@ -0,0 +1,54 @@
...
+To generate API documentation for a new major or minor version:
+
+#### Generate Documentation for the new Version
+
+First, generate the API docs for the new version.  To do this, run `npm run build:docs` on main.  This will output the documentation in the docs_utils/build folder.
+
+#### Update the Node Driver Docs
+
+After generating the API docs for a particular version, the next step is to update the static site to include the new docs.  Open the `generate-docs.ts` 
+script and update the `NEW_VERSION` object with information for the new version.
```
dariakp, line 39: can we make this not be a manual step but instead a command line arg to an npm script (preferably in package.json, too)? that would make it easier to eventually generate docs for automatic checking in the CI

etc/docs/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Finally, push the new documentation to Github and confirm that the site builds as expected.
+
+### Doc Generation Overview
+
+API docs are stored in the top level `docs` folder in the `node-mongodb-driver` Github repository and are hosted on Github Pages.  Each major or minor release,
+a new set of API docs are generated and added to our Hugo static site.  This is a three step process:
+
+1. Generate the new API docs using type doc
+2. Run the generate docs script to create the new version of the Hugo site, copy in the new documentation and move the newly generated site into the `docs` folder.
```
dariakp, line 53: the script already does the copying into docs

etc/docs/generate-docs.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  // status - a status for the version.  typically would probably just be 'latest'
+  status: 'current',
```
dariakp, line 41: what does this affect? I didn't see a difference in the TOC based on this

etc/docs/generate-docs.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  // api - the name of the folder for the generated documentation to live.  typically should be version_id prefixed with `./`
+  //  ex: ./4.5
+  api: './4.5',
+
+  // usesMongoDBManual - if true, includes a link to the mongodb documentation for the new version
+  usesMongoDBManual: true,
```
dariakp, line 48: I noticed this always links to latest docs instead of that version's specific docs, is that intentional? might be a question for @nbbeeken

etc/docs/generate-docs.ts
```diff
@@ -0,0 +1,121 @@
...
+  const command = `cp -R ${pathToBuiltDocs} ${outputDirectory}`;
+  return await exec(command);
+}
+
+async function updateSiteTemplateForNewVersion(tomlData: TomlVersionSchema, jsonData: JsonVersionSchema[]) {
+  tomlData.versions.unshift(NEW_VERSION);
+  tomlData.current = NEW_VERSION.version;
+  writeFileSync(RELEASES_TOML_FILE, stringify(tomlData as any));
+
+  jsonData.unshift({ version: NEW_VERSION.semvarVersion });
```
dariakp, line 87: Should this allow for updating the docs that were already generated for a given version? I think at least for main/"next", it would be nice to regenerate and update docs continuously instead of just the one time

dariakp (changes requested):

etc/docs/template/data/releases.toml
baileympearson: no that's just a typo.  I added semverVersion to each line so that each entry matched the schema in my script but that's the only change in this file

package.json
baileympearson: I thought that was jsdoc - used to generate the legacy documentation.  @nbbeeken do you remember which package caused problems?

etc/docs/generate-docs.ts
baileympearson: It's used in the html template - https://github.com/mongodb-js/learn-mongodb-docs/blob/master/site/layouts/partials/releases.html

etc/docs/generate-docs.ts
baileympearson: I'm not sure.  My schema in the script simply matches the existing toml data used to generate our static site.  If looking into this is something we want to include in this work, I can investigate

package.json
nbbeeken: It was [typedoc](https://github.com/TypeStrong/typedoc/blob/master/package.json#L30-L32). When the next version of typescript drops we get install errors until typedoc adopts it, which we usually don't need to be blocked on

etc/docs/generate-docs.ts
baileympearson: It could - I'm happy to update the script to support that if we wanted.  FYI though, although the script supports "main/next" simply because you can name your version whatever you like, it would take a bit more effort to make it work well.  If that's something we want to support in this PR, I can also make those changes

etc/docs/generate-docs.ts
baileympearson: This was unintentional.  I'll flip the wording

etc/docs/generate-docs.ts
dariakp: oh, it looks like it gets turned into a css class, probably worth mentioning or enumerating what classes are available/meaningful; I feel like maybe @nbbeeken was using it to control whether the older docs are displayed or collapsed

etc/docs/generate-docs.ts
dariakp: I was hoping Neal had a quick answer there

etc/docs/generate-docs.ts
dariakp: Just curious - what would be the issues we'd need to work around with using something like `Next` instead of a real version? Mainly I'm just concerned with always being able to properly preview the current docs and to generate them in a consistent spot so that they *can* be included in PRs

etc/docs/template/README.md
baileympearson: I'm not sure.  I removed it because the readme didn't seem necessary anymore

etc/docs/generate-docs.ts
baileympearson: I assume we'd always want `main` at the top of the list of driver versions.  So basically just add logic to detect when `main` is passed as the version, and make sure to place it at the top of the list driver versions on the generated static site

etc/docs/template/README.md
nbbeeken: Isn't Java a strictly typed version of JavaScript? 🤪  I think our docs was forked from Java docs long ago, our [404](https://mongodb.github.io/node-mongodb-native/ada) page still has Duke on it

etc/docs/template/static/s/css/frontpage.css
```diff
@@ -0,0 +1,399 @@
...
+}
+
+#nodelogo {
+  background-image: url("../img/nodejs.png");
+  background-position: left top;
+  background-repeat: no-repeat;
+}
+
+#mongodbJVMlogo {
+  background-image: url("../img/mongoJVMlogo.png");
```
nbbeeken, line 235: Let's change this to: https://wiki.corp.mongodb.com/pages/viewpage.action?pageId=79305748&preview=/80089640/133270877/node.png :)

etc/docs/template/layouts/partials/assets/analytics.html
```diff
@@ -0,0 +1,9 @@
+<script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+
+  ga('create', 'UA-29229787-1', 'auto');
```
nbbeeken, line 7: Can we find out who has access to this reporting, we should double check this goes somewhere useful.

etc/docs/template/layouts/partials/assets/javascripts.html
```diff
@@ -0,0 +1,7 @@
+<script type="text/javascript" src="{{.Site.BaseURL}}/s/js/jquery.js"></script>
+<script type="text/javascript" src="{{.Site.BaseURL}}/s/lib/bootstrap.js"></script>
+<script type="text/javascript" src="{{.Site.BaseURL}}/s/lib/highlight/highlight.pack.js"></script>
```
nbbeeken, line 3: Do these get created by the generation of the site? I'm not seeing any of the `/lib/` scripts in this PR

etc/docs/template/layouts/partials/header/main.html
```diff
@@ -0,0 +1,12 @@
+<header id="header-db" class="row" role="navigation">
+  <div class="header-content">
+    <!--logo start-->
+    <div class="logo pull-left">
+      <a href="{{.Site.BaseURL}}">
+        <img src="{{.Site.BaseURL}}/s/img/logo-mongodb-header.png", alt="MongoDB.org" />
```
nbbeeken, line 6: We should probably change this:
https://www.mongodb.com/brand-resources

etc/docs/template/layouts/partials/footer.html
```diff
@@ -0,0 +1,8 @@
+<div class="footer">
+  <div class="copyright">
+    <p>Copyright © 2015 <a class="smalltext" href="http://www.mongodb.com">MongoDB, Inc</a>.
```
nbbeeken, line 3: I think we still hold the copyright :) so this could be 2022 (or maybe we don't need this?)

etc/docs/template/layouts/404.html
```diff
@@ -0,0 +1,45 @@
...
+  <!--header start-->
+  {{ partial "header/main.html" . }}
+  <!--header end-->
+
+  <div id="promo" class="jumbotron">
+    <div class="container">
+    <div class="row">
+      <div class="col-md-2"></div>
+      <div class="col-md-6">
+        <h1>404: Page not found</h1>
```
nbbeeken, line 22: Just a quick win idea, a link back to the homepage or mongodb manual could be good here, if we end up having to break legacy version links down the road this page may get an uptick in visits.

etc/docs/template/static/s/js/jquery.js
```diff
@@ -0,0 +1,9789 @@
+/*!
+ * jQuery JavaScript Library v1.10.2
```
nbbeeken, line 2: Should we be concerned about this? If we could just try the latest jquery still works, it might be a good idea to update

etc/docs/template/layouts/index.html
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    {{ partial "meta.html"}}
+    <link rel="shortcut icon" href="{{.Site.BaseURL}}/s/img/favicon.png">
```
nbbeeken, line 5: The icon should be updated: https://www.mongodb.com/assets/images/global/favicon.ico

etc/docs/template/.gitignore
```diff
@@ -0,0 +1 @@
+public
```
nbbeeken, line 1: Is this relevant to this directory?

etc/docs/generate-docs.ts
```diff
@@ -0,0 +1,149 @@
...
+    const existingVersionIndex = tomlData.versions.findIndex(({ semverVersion }) => semverVersion === newVersion.semverVersion);
+    tomlData.versions[existingVersionIndex] = newVersion;
+  } else {
+    tomlData.versions.unshift(newVersion);
+    tomlData.current = newVersion.version;
+
+    jsonVersions.unshift({ version: newVersion.semverVersion })
+  }
+
+  writeFileSync(RELEASES_TOML_FILE, stringify(tomlData as any));
```
nbbeeken, line 83: `await writeFile` instead?

etc/docs/legacy-generate.sh
```diff
@@ -0,0 +1,138 @@
...
+        ;;
+        "4")
+            generate_4x "$branch"
+            cp -R "checkout/$branch/docs/public" "./public/$branch"
+        ;;
+
+        # if 'main' is specified as a branch, we're releasing a new version
+        # TODO: figure out a way to pass in the version
+        "m")
+            echo "HEY YOU BETTER EDIT ME!!!"
```
nbbeeken, line 101: We decided to do this in a follow up right? should we ticket the TODO?

package.json
```diff
@@ -119,6 +121,7 @@
     "check:csfle": "mocha --config test/mocha_mongodb.json test/integration/client-side-encryption",
     "check:snappy": "mocha test/unit/assorted/snappy.test.js",
     "prepare": "node etc/prepare.js",
+    "preview:docs": "python  -m http.server --directory etc/docs/build",
```
nbbeeken, line 124: Totally fine if you think this is not the right direction but: https://github.com/nbbeeken/nbbeeken.github.io/blob/main/scripts/server.js we could have a little server that works just on node to avoid the need for python here. I guess we all need python for mlaunch so not a huge deal.

typedoc.json
```diff
@@ -3,5 +3,6 @@
     "src/index.ts"
   ],
   "excludeTags": ["internal"],
-  "out": "docs/public"
+  "excludeInternal": false,
```
nbbeeken, line 6: We want to exculdeInternal, so this should be true right?

nbbeeken (changes requested):

etc/docs/template/layouts/partials/assets/javascripts.html
baileympearson: No.  These scripts are present in the gh-pages branch (https://github.com/mongodb/node-mongodb-native/tree/gh-pages/s/lib).  They'll be present after part 2 of this PR, when I move the contents of `gh-pages` to main.

etc/docs/legacy-generate.sh
baileympearson: What exactly are you referring to here?  

I included this script only so that, if needed, we could generate docs for legacy versions (3.x).  The other script I wrote should support generating docs for any 4x version

etc/docs/template/static/s/css/frontpage.css
baileympearson: Let's update this in my second PR - the `img` files don't exist in this PR since there are only changes to the site template (which in our case, doesn't include the images).

etc/docs/template/layouts/partials/assets/analytics.html
baileympearson: Yeah, I can check

etc/docs/README.md
```diff
@@ -0,0 +1,38 @@
...
+* node (v6.x or v8.x) and npm (>= 3.x). Note: worked for me with 8.17.0, but not 8.0.0.
+* python sphinx
+
+### Usage
+
+To generate API documentation for a new major or minor version:
+
+`npm run docs:build -- <version> <optional status for the version>`
+
+The generated docs can be previewed using `npm run docs:preview`.
```
baileympearson, line 35: This step hasn't been fully tested because the actual documentation site won't be ported over until a follow up PR.  I'll test this and ensure that the preview step works as expected then.

etc/docs/generate-docs.ts
baileympearson: I can't find any css classes that correspond to the statuses we use now and I confirmed on the current hosted version of our docs that these css classes don't actually apply any styles.  I'll remove the code in the template that applies the styles.  

This value is also used to filter out legacy versions of the driver when building the template though.

etc/docs/template/layouts/partials/assets/analytics.html
baileympearson: Mary confirmed that we do intentionally collect these analytics

etc/docs/template/layouts/partials/header/main.html
baileympearson: I'll address this with the other `img` related comments in the follow up PR!

etc/docs/template/layouts/index.html
baileympearson: same comment as other `img` changes

etc/docs/template/layouts/404.html
baileympearson: Do you mind if I make this change in my follow up PR?  There's no way to test that the link I add works with the changes in this PR (since we don't actually have the whole site included).

etc/docs/template/static/s/js/jquery.js
baileympearson: Do you mind if I make this change in my follow up PR? There's no way to test that the link I add works with the changes in this PR (since we don't actually have the whole site included).

etc/docs/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Once everything looks correct, open a PR against `main`.  Once the PR is merged, the docs will automatically re-build via a Github action.  Once this
```
dariakp, line 37: Just a question - is there an actual github action that *we* configure to rebuild the site or does it just automatically get updated once we point the github pages setting in the repo to the main docs branch? If there is no action for us to manually manage, it may be better just to say here that our github pages config points to the /docs directory on main and is kept up to date by github itself

typedoc.json
```diff
@@ -3,5 +3,6 @@
     "src/index.ts"
   ],
   "excludeTags": ["internal"],
-  "out": "docs/public"
+  "excludeInternal": true,
```
dariakp, line 6: with this being the right setting, I don't think we should keep the line above this one to `"excludeTags": ["internal"]`, since presumably they wouldn't be present in the first place, and if we somehow accidentally lose this flag, excluding the internal tag from being displayed would just make it more confusing to our users

etc/docs/generate-docs.ts
dariakp: In that case, should we make it an enum of `'current' | 'not-supported'` and/or add validation that it's not an unknown value (to prevent typos)?

etc/docs/generate-docs.ts
```diff
@@ -0,0 +1,150 @@
...
+async function updateSiteTemplateForNewVersion(newVersion: VersionSchema, tomlData: TomlVersionSchema, jsonVersions: JsonVersionSchema[]) {
+  const versionExists = jsonVersions.some(({ version }) => version === newVersion.semverVersion);
+  if (versionExists) {
+    const existingVersionIndex = tomlData.versions.findIndex(({ semverVersion }) => semverVersion === newVersion.semverVersion);
+    tomlData.versions[existingVersionIndex] = newVersion;
+  } else {
+    tomlData.versions.unshift(newVersion);
+    tomlData.current = newVersion.version;
+
+    jsonVersions.unshift({ version: newVersion.semverVersion })
```
dariakp, line 81: I think we could make this process a bit more resilient to generating versions out of order by just writing a custom semver sort function and passing it to both tomlData and and jsonVersions (should be straightforward since it's internal and also we only do major.minor, no patch or pre-release flags); we could also have it check for our special keyword `main` or `current`, something like:
```js
const versionSort = (a, b) => {
  const vA = a.version;
  const vB = b.version;
  if (vA === 'current') { return -1; }
  if (vB === 'current') { return 1; }
//  < logic to return -1 if vA > vB and 1 if vA < bB >
// since we want the most recent version first
}

<unshift new version>
tomlData.versions.sort(versionSort);
tomlData.current = tomlData.versions[0].versions;

<unshift new version>
jsonVersions.sort(versionSort);

```

dariakp (changes requested):

etc/docs/generate-docs.ts
dariakp: I have a suggestion in a new comment to handle this with a sort, let me know what you think

package.json
nbbeeken: this is fine to leave as is, can we fix the spacing on the args? `python  -m` -> `python -m`

etc/docs/generate-docs.ts
nbbeeken: We don't currently use the supported class but my quick fix to make the version list table get too long isn't ideal, I think making use of these CSS classes to better indicate the status of each version is a future improvement. Is it okay if we keep this and consider making use of it in the next PR or just future work?

I'm thinking background color like [wikipedia uses](https://en.wikipedia.org/wiki/OpenSSL#Major_version_releases)

package.json
baileympearson: Why not two spaces everywhere instead of one?  `python  -m  http.server  --directory  etc/docs/build`?  🤔 

jk, good catch! I'll remove the space

etc/docs/README.md
baileympearson: The Github action is managed by Github and it should update as soon as we switch the location of the static site hosting from `gh-pages` to `main/docs`.  I'll update the wording here

etc/docs/generate-docs.ts
nbbeeken: It's always linked to the latest, we could add versioning but not here I think.

dariakp (dismissed):
Approved with the expectation of a follow up with the full docs and another with the template/generating script updates

nbbeeken (approved):

PR #3154, opened 2022-02-25
refactor(NODE-4042): remove execute legacy operation

nbbeeken:
#### What is changing?

Remove executeLegacyOperation by creating a shim class that `extends AbstractOperation` inline.

Remove debt with different more maintainable debt.

## Open Qs
- Can we declare this class outside of this code section for reuse or does it need to be inlined?
  - Yes, done.
- Are there Aspects that should be set on this shim class? (Retryability is still passing so... maybe no?)
  - It would only be cosmetic, a Bulk operation isn't real, in that it is broken up into insertMany/updateMany/deleteManys, those are retried respectively, it's a "fake" operation
- We certainly want to file for follow up work, what would we define as a better solution? One tricky problem is that .execute is a public API function, so subclassing the internal AbstractOperation that forces you to implement execute its way is not possible.


src/bulk/common.ts
```diff
@@ -923,6 +923,32 @@ export interface BulkWriteOptions extends CommandOperationOptions {
...
+/**
+ * TODO(NODE-XXX)
+ * BulkWrites merge complexity is implemented in executeCommands
+ * This provides a vehicle to treat bulkOperations like any other operation (hence "shim")
+ * We would like this logic to simply live inside the BulkWriteOperation class
+ * @internal
+ */
+class BulkWriteShimOperation extends AbstractOperation {
+  bulkOperation: BulkOperationBase;
+  constructor(options: BulkWriteOptions, bulkOperation: BulkOperationBase) {
```
dariakp, line 935: usually "bucket" options are the last parameter passed to a constructor, any particular reason to have this flipped here?

test/integration/crud/bulk_execute_operation.test.ts
```diff
@@ -0,0 +1,29 @@
...
+describe('Bulk executeOperation', () => {
+  let client;
+  beforeEach(async function () {
+    client = await this.configuration.newClient({ monitorCommands: true }).connect();
+  });
+  afterEach(async () => {
+    await client?.close();
+  });
+
+  it('should use the same session for every operation', async () => {
```
dariakp, line 12: should we do 2 cases here, for completeness, both for ordered and unordered ops? since the change was to common?

dariakp (changes requested):
I actually don't have a lot of feedback here as the change is relatively straightforward. Do we have sufficient testing on gridfs operations to ensure that the updated code works the same as the code previously funneled through execute legacy operation?

test/integration/crud/bulk_execute_operation.test.ts
nbbeeken: The difference between ordered and unordered is the ability to affect grouping of the operations, I don't think it adds any additional coverage, the shim class isn't a "real" operation that get's executed it's for us to pass the options an properties along in the right shape for executeCommands to build each actual operation

src/bulk/common.ts
nbbeeken: No reason, I'll flip!

test/integration/crud/bulk_execute_operation.test.ts
dariakp: I think it's low risk enough to be fine in this case, but I think generally as a best practice, whenever we are using an integration test and not a unit test to make sure the class works correctly, and whenever there are two or more distinct ways to access the code under test in an integration setting, we should try to test all the pathways (in this case, it would be to ensure that the overrides from the specific bulk class don't affect the session attachment from the base class)

src/bulk/common.ts
```diff
@@ -923,6 +923,32 @@ export interface BulkWriteOptions extends CommandOperationOptions {
   forceServerObjectId?: boolean;
 }
 
+/**
+ * TODO(NODE-XXX)
```
dariakp, line 927: marking this to make sure we don't leave a placeholder TODO

dariakp (changes requested):

src/bulk/common.ts
nbbeeken: done!

baileympearson (approved):

dariakp (approved):

PR #3157, opened 2022-03-01
docs(NODE-4018): Doc generation script improvements and site enhancements

baileympearson:
#### What is changing?

- Updated some images on our docs site
- Added a docs preview server (see comment)
- Enhanced the docs generation by adding a custom semver sort function per Daria's suggestion (great idea! works like a charm)
- Fixes broken links on the docs homepage
- Upgrades jquery to the latest version (I noticed no change in functionality when testing locally)

No.


etc/docs/preview.ts
```diff
@@ -0,0 +1,16 @@
+import * as express from 'express';
+
```
baileympearson, line 2: The python http server didn't support redirecting endpoints from `node-mongodb-native` to `/`, so I added express as a dev dependency and wrote a simple server to preview docs.

etc/docs/build.ts
```diff
@@ -2,13 +2,12 @@
 
 import { parse, stringify } from '@iarna/toml';
 import { exec as execCb } from 'child_process';
-import { readFileSync, writeFile as writeFileCb } from 'fs';
+import { writeFile, readFile } from 'fs/promises';
```
dariakp, line 5: since the file was renamed, we should also update the README that talks about generate-docs.ts

etc/docs/build.ts
```diff
@@ -69,18 +68,40 @@ async function copyNewDocsToGeneratedSite({ semverVersion }: VersionSchema) {
...
+  }
+  // put legacy 3x driver version at the bottom of the list
+  if (a === 'core') {
+    return 1;
+  }
+  if (b === 'core') {
+    return -1;
+  }
+
+  return b.localeCompare(a);
```
dariakp, line 86: Unfortunately the cheat method won't work here:

```js
['5.9', '5.10','5.11','10.5'].sort((a,b)=>b.localeCompare(a)) 
// [ '5.9', '5.11', '5.10', '10.5' ]
```

dariakp (changes requested):

etc/docs/build.ts
nbbeeken: We have `semver` as a dev dep, could we use that?

dariakp: The plain `compare` is not meant to work with invalid npm versions but there might be some special function functionality you could reuse? a string split and numeric compare should be pretty trivial though

baileympearson: I think `semver.greaterThan` would work.  But now I'm not sure without testing it some 😅

dariakp: something like '5.1' (as opposed to something like '5.1.1' is not valid semver, so you can't put it directly into the comparators. Doing a map `split` chained with map `Number` on the inputs would be one line of code, a conditional comparing them piece by piece another line, so either which way, it's two lines of code. 😉

etc/docs/build.ts
```diff
@@ -122,48 +60,33 @@ async function main() {
...
-  const versionAlreadyExists = jsonVersions.some(({version }) => version === semverVersion)
+  const versionAlreadyExists = jsonVersions.some(({ version }) => version === semverVersion)
 
   if (versionAlreadyExists) {
-    const response = await prompt(`Version ${semverVersion} already exists.  Do you want to override the existing docs? [y/n] `);
-    if (response !== 'y') {
-      console.error("something went wrong.  Exiting...");
-      process.exit(1);
-    }
+    await confirm(`Version ${semverVersion} already exists.  Do you want to override the existing docs? [y/n] `);
```
dariakp, line 83: I like the idea here, but would it prevent the script from being able to run in the CI? Since we're probably only to run the main generation in CI to ensure docs are kept in sync, maybe we can add a condition to only prompt for the override if the version is not "current".

Also, food for thought, should we give the docs from main some other name, like 
"next" instead of "current" to avoid confusion with the most recently released version?

dariakp (changes requested):

etc/docs/preview.ts
```diff
@@ -0,0 +1,36 @@
...
+        if (await canAccess(filePath)) {
+          response.end(await readFile(filePath))
+        } else {
+          response.end(await readFile(fromDir('404.html')))
+        };
+      }
+    } catch (error) {
+      response.statusCode = 500;
+      response.end(`Bad Request: ${error}`, 'utf8');
+    }
```
nbbeeken, line 33:
```
} finally {
      console.log(request.method, response.statusCode, pathname);
}
```

We can log the status code to make it a bit more obvious when something is missing.

etc/docs/preview.ts
```diff
@@ -0,0 +1,36 @@
...
+async function main() {
+  for await (const [request, response] of on(server, 'request')) {
+    const url = new URL(request.url, `http://${request.headers.host}`);
+    console.log(request.method, url.pathname);
+    const pathname = url.pathname.replace('/node-mongodb-native', '');
+    console.log(pathname)
+    try {
+      if (pathname === '/') {
+        response.end(await readFile(fromDir('index.html')));
+        continue;
```
nbbeeken, line 21:
```suggestion
```

etc/docs/preview.ts
```diff
@@ -0,0 +1,36 @@
...
+      } else {
+        const filePath = fromDir(pathname.slice(1));
+        if (await canAccess(filePath)) {
+          response.end(await readFile(filePath))
+        } else {
+          response.end(await readFile(fromDir('404.html')))
```
nbbeeken, line 27:
```suggestion
          response.statusCode = 404
          response.end(await readFile(fromDir('404.html')))
```

etc/docs/preview.ts
```diff
@@ -0,0 +1,36 @@
...
+const server = createServer();
+server.listen(8080, 'localhost', 128, () => console.log('Listening on http://localhost:8080'));
+const canAccess = async (pathlike: PathLike) => await access(pathlike).then(() => true,() => false);
+
+async function main() {
+  for await (const [request, response] of on(server, 'request')) {
+    const url = new URL(request.url, `http://${request.headers.host}`);
+    console.log(request.method, url.pathname);
+    const pathname = url.pathname.replace('/node-mongodb-native', '');
+    console.log(pathname)
```
nbbeeken, line 17:
```suggestion
```

I think we should instead fix the paths in the html, we shouldn't have to do this request editing, but we need to double check on your test fork. All our src/img/css tags should start after the `/node-mongodb-native` part. 

This:
```html
<link rel="stylesheet" href="/node-mongodb-native/s/lib/bootstrap.min.css" type="text/css" />
```
Should be:
```html
<link rel="stylesheet" href="s/lib/bootstrap.min.css" type="text/css" />
```

docs/index.html
```diff
@@ -8,7 +8,7 @@
 <meta name="author" content="">
 <meta name="keyword" content="">
 
-    <link rel="shortcut icon" href="/node-mongodb-native/s/img/favicon.png">
```
nbbeeken, line 11: Commenting here, but referring to the img code below:
- `logo-mongodb-header.png`
- `mongodb-university-logo.png`
- `node.png`

Were we replacing these?

nbbeeken (changes requested):

etc/docs/preview.ts
```diff
@@ -0,0 +1,36 @@
+import { on } from 'events';
```
nbbeeken, line 1: I retract my request for this, apologies! Let's go with the python solution or we can use `npx http-server docs` which will install it as needed.

etc/docs/build.ts
baileympearson: Since Neal kindly pointed out we already have yargs installed, I switched to using yargs and added some extra options. 
 Now it feels like a real command line script 😄 .

- `--yes` will skip any prompts for the user (similar to `npm init -y`)
- `--versionName` will set the display name on the docs page.  i.e., `npm run build:docs -- --versionName next` will display `Next Driver` on the docs 

Personally I prefer this, because then we're not making assumptions in the script and it's more flexible.  What do you think?

etc/docs/preview.ts
nbbeeken: Yes!

baileympearson: Filed a ticket here - https://jira.mongodb.org/browse/NODE-4056

etc/docs/build.ts
```diff
@@ -0,0 +1,96 @@
...
+    const existingVersionIndex = tomlData.versions.findIndex(({ tag }) => tag === newVersion.tag);
+    tomlData.versions[existingVersionIndex] = newVersion;
+  } else {
+    tomlData.versions.unshift(newVersion);
+
+    jsonVersions.unshift({ version: newVersion.tag })
+  }
+
+  tomlData.versions.sort((a, b) => customSemverCompare(a.tag, b.tag))
+  tomlData.current = tomlData.versions[0].version;
```
dariakp, line 39: Just noticed this part - will it be an issue when we update the docs with the `next` version? Maybe we should just create the `next` docs in this PR, and then update this to be `tomlData.versions[1].version` (or we can keep it [0] but filter out "next")

dariakp (changes requested):

etc/docs/utils.ts
```diff
@@ -0,0 +1,95 @@
+import { createInterface } from "readline";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+
+export interface JsonVersionSchema {
```
durran, line 5: nit: we have the common pattern of indenting 2 spaces but this file has 4 spaces. Other ts in this PR are also 2.

durran (changes requested):

etc/docs/utils.ts
```diff
@@ -0,0 +1,95 @@
+import { createInterface } from "readline";
```
dariakp, line 1: similar to Durran's comment, we should probably stick to single quotes for strings, there is a mix and match of them here and in L43

dariakp (changes requested):

dariakp (approved):

nbbeeken (approved):

PR #3159, opened 2022-03-03
test(NODE-4057): update dns discovery tests

durran:
### Description

Updates the spec tests to verify DRIVERS-2224, which corrects the test cases for testing srv records in a load balancer environment.

None

DRIVERS-2224/NODE-4057

### DIFF
--- a/test/integration/initial-dns-seedlist-discovery/initial_dns_seedlist_discovery.spec.test.ts
+++ b/test/integration/initial-dns-seedlist-discovery/initial_dns_seedlist_discovery.spec.test.ts
@@ -18,7 +18,9 @@ function makeTest(test, topology) {
   });
 
   it(test.comment, async function () {
-    return this.skip(); // FIXME(NODE-3757): These tests require specific environment setups, also the error cases need better assertions
+    this.test.skipReason =
+      'TODO(NODE-3757): These tests require specific environment setups, also the error cases need better assertions';
+    return this.skip();
     if (topology === 'replica-set' && this.configuration.topologyType !== 'ReplicaSetWithPrimary') {
       return this.skip();
     }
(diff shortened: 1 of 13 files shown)


nbbeeken:
Taking a closer look at the CI the tests are being [skipped](https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_next_ubuntu_18.04_erbium_test_load_balancer_patch_ade5ec26d57e9ddd158f0c9dd550925b9548e099_6220c8e57742ae36580632a2_22_03_03_13_55_50/0?type=T#L2672), if they still need to be after this change let's try adding a skipReason.

nbbeeken (dismissed):

dariakp (commented):
@nbbeeken your last comment was asking regarding the skips and if we should be adding a skip reason - did that get resolved?

nbbeeken:
Oh good catch! this ticket was blocked for too long 😅  @durran can we double check about the skipping here, is that intentional?

durran:
> Oh good catch! this ticket was blocked for too long 😅 @durran can we double check about the skipping here, is that intentional?

@nbbeeken The tests are skipped due to https://jira.mongodb.org/browse/NODE-3757 - should I go ahead and address that as part of this ticket?

nbbeeken:
Oh nice, as long as we have it tracked, could you add it as a skip reason in `test/integration/initial-dns-seedlist-discovery/initial_dns_seedlist_discovery.spec.test.ts` I see there's a comment from before we had that feature. I feel like there is value in making it so if/when a related ticket comes up and we first check the logs it's clear the skip was intentional.
```js
  this.currentTest.skipReason =
      'TODO(NODE-3757): These tests require specific environment setups, also the error cases need better assertions';
    return this.skip();
```

durran:
@nbbeeken I've updated with the skip reason and rebased fresh against main here now.

nbbeeken (approved):
Most excellent thank you

baileympearson (approved):

PR #3164, opened 2022-03-08
refactor: correct types for AbstractOperation.execute

nbbeeken:
#### What is changing?

The session can sometimes be undefined in execOperation, we luckily didn't have any code that relied on its existence :)

* made explicit use of override
* made explicit use of return, caught bug in withSession
* Updated eslint typescript plugin so the parser stops mentioning TS version mismatch

Correct internal types, should have no impact on users, helps devs reason about code.


src/bulk/common.ts
```diff
@@ -289,6 +289,7 @@ export class BulkWriteResult {
     if (index < this.result.writeErrors.length) {
       return this.result.writeErrors[index];
     }
+    return;
```
baileympearson, line 292:
```suggestion
return index < this.result.writeErrors.length ? this.result.writeErrors[index] : undefined
```

nbbeeken: 🧑‍💻  ⛳ ?
I'm okay with the suggestion but I don't think we need to be writing `return undefined;` anywhere, which I know this isn't but it's implicitly similar. Does that make sense?

package.json
```diff
@@ -52,15 +52,15 @@
     "@types/semver": "^7.3.9",
     "@types/sinon": "^10.0.11",
     "@types/whatwg-url": "^8.2.1",
-    "@typescript-eslint/eslint-plugin": "^5.13.0",
-    "@typescript-eslint/parser": "^5.13.0",
+    "@typescript-eslint/eslint-plugin": "^5.14.0",
```
baileympearson, line 55: sneaking in some version bumps I see, classic classic

src/mongo_client.ts
```diff
@@ -556,52 +549,34 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
    * NOTE: presently the operation MUST return a Promise (either explicit or implicitly as an async function)
    *
    * @param options - Optional settings for the command
-   * @param callback - An callback to execute with an implicitly created session
+   * @param withSessionCallback - An callback to execute with an implicitly created session
```
baileympearson, line 552: The name of the parameter is still callback.  Let's either rename the parameter or revert this change

src/bulk/common.ts
nbbeeken: What about `this.result.writeErrors[index]`?

src/error.ts
```diff
@@ -670,7 +670,6 @@ function makeWriteConcernResultObject(input: any) {
 export class MongoWriteConcernError extends MongoServerError {
   /** The result document (provided if ok: 1) */
   result?: Document;
```
baileympearson, line 672: Did you mean to remove `errInfo` from the class?

baileympearson (changes requested):

src/mongo_client.ts
nbbeeken: Good catch I say we leave it as is

src/bulk/common.ts
baileympearson: For some reason I thought that threw an error if index was out of range.  I guess not - I'd prefer this.

I actually prefer `return undefined` but that's just preference.  I like that it's explicit 🤷 but I don't feel terribly strongly about this

src/error.ts
nbbeeken: Yes it's already defined in the parent, you'll notice a few other removals, if the override doesn't actually narrow the type then there's no purpose for it.

src/bulk/common.ts
dariakp: return undefined would read weird to me personally, and in general, and especially in holding with TS conception of types, if your function is supposed to return something but you don't have a result, the appropriate default is `null` not undefined

src/connection_string.ts
```diff
@@ -955,16 +957,14 @@ export const OPTIONS = {
           readPreference: { ...options.readPreference, ...value },
           ...value
         } as any);
-      }
-      if (isRecord(value, ['mode'] as const)) {
+      } else if (isRecord(value, ['mode'] as const)) {
```
dariakp, line 960: why the additions of `else` here all the returns are already definitely typed, so this seems like unnecessary logic, minus the last throw; if anything, this syntax implies that the block above might *not* properly return

src/cursor/find_cursor.ts
```diff
@@ -329,21 +329,6 @@ export class FindCursor<TSchema = Document> extends AbstractCursor<TSchema> {
     return this;
   }
 
-  /**
-   * Set a maxTimeMS on the cursor query, allowing for hard timeout limits on queries (Only supported on MongoDB 2.6 or higher)
-   *
-   * @param value - Number of milliseconds to wait before aborting the query.
-   */
-  maxTimeMS(value: number): this {
```
dariakp, line 337: does it matter that this used to be defined on `kBuiltOptions`, which gets spread during initialize on L73, and now it's deeper on the `cursorOptions`, which second in the spread? the comment there explicitly says that the order matters, i.e. that those cursorOptions take priority over the find cursor options, and, while that seems backwards, it's a little bit ominous, so I just wanted to know if you explicitly looked into that or have any context on the comment

src/bulk/common.ts
nbbeeken: I agree, I went with the ternary though to preserve the API. Filed NODE-4068

src/mongo_client.ts
```diff
@@ -564,44 +557,26 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
-      const result = callback(session);
-      return Promise.resolve(result).then(
-        result => cleanupHandler(undefined, result, undefined),
-        err => cleanupHandler(err, null, { throw: true })
-      );
-    } catch (err) {
-      return cleanupHandler(err, null, { throw: false }) as Promise<void>;
-    }
+    return Promise.resolve()
+      .then(() => withSessionCallback(session))
```
dariakp, line 578: technically, because the result is returned, the declared return type is incorrect, because the type returned would be the type of the return type of `withSessionCallback` and not guaranteed to be `Promise<void>`

dariakp (changes requested):
Some questions comments

src/connection_string.ts
nbbeeken: Good point, there was no need for else(s), reverted, kept the throws.

src/cursor/find_cursor.ts
nbbeeken: Ah kBuiltOptions vs kOptions, really good catch I did not catch the difference in the super class. I've reverted

src/mongo_client.ts
nbbeeken: fixed!

test/integration/sessions/sessions.test.js
```diff
@@ -72,84 +72,59 @@ describe('Sessions Spec', function () {
...
-
-          return promise
-            .then(() => expect(client.topology.s.sessionPool.sessions).to.have.length(1))
+          return client
+            .withSession(testCase.operation(client))
+            .then(
+              () => expect(client.topology.s.sessionPool.sessions).to.have.length(1),
+              () => expect(client.topology.s.sessionPool.sessions).to.have.length(1)
+            )
             .then(() => client.close())
```
dariakp, line 128: the client close here and in the test case below should probably be in an afterEach; the test cases should also differentiate `resolve` from `reject` cases in their description and corresponding assertions; the test case descriptions should indicate that we care that sessions are cleaned up in both cases

dariakp (changes requested):

dariakp (dismissed):
LGTM!

src/connection_string.ts
```diff
@@ -975,6 +977,7 @@ export const OPTIONS = {
           rpOpts
         );
       }
+      throw new MongoParseError(`Unknown ReadPreference value: ${value}`);
```
baileympearson, line 980: Could we add a test for this new error message if it's not covered in our spec tests already?

test/integration/sessions/sessions.test.ts
```diff
@@ -72,93 +73,108 @@ describe('Sessions Spec', function () {
...
-          },
-          {
-            description: 'should support operations that throw exceptions',
-            operation: (/* client */) => (/* session */) => {
-              throw new Error('something went wrong!');
-            }
+      const tests = [
+        {
+          description: 'should resolve non-async callbacks that return promises',
+          operation: client => session => {
```
baileympearson, line 89: Is there any reason we couldn't have a single function with two parameters?  We're not currying anywhere so I don't think this is necessary

test/integration/sessions/sessions.test.ts
```diff
@@ -72,93 +73,108 @@ describe('Sessions Spec', function () {
...
+            throw new Error('something went wrong!');
+          }
+        }
+      ];
 
-        it('supports passing options to ClientSession', function () {
-          const client = test.client;
+      for (const testCase of tests) {
+        it(testCase.description, async function () {
+          const shouldResolve = testCase.description.startsWith('should resolve');
```
baileympearson, line 129: This isn't blocking - but could we just add a property to the tests instead to indicate whether or not the test should resolve or reject?

```typescript
{
      description: 'should reject callbacks that throw synchronous exceptions',
      operation: (/* client */) => (/* session */) => {
        throw new Error('something went wrong!');
      },
      result: 'resolve' // or something like this
}

...

const shoudResolve = result === 'resolve';
```

baileympearson (changes requested):

test/integration/sessions/sessions.test.ts
nbbeeken: The function returned from the first call has to be correctly shaped for withSession, withSession only passes in the session as a single argument. That being said after updating this test to the before/after style the client is accessible from the outer scope so we don't need the factory style anymore

src/connection_string.ts
nbbeeken: Yes should be a quick unit test, good point

dariakp (approved):

PR #3165, opened 2022-03-09
fix(NODE-3688): make handshake errors retryable

durran:
### Description

Makes operations retryable when the handshake fails.

#### What is changing?

- Updates spec tests for retryable reads & writes and transactions.
- Verifies handshakes errors are retryable via new integration tests that mimic the unified tests.
- Skips unified retryable handshake tests. (see comments in the spec test runner)

No

NODE-3688

#### Notes

After syncing tests there were some extras that sneaked in:

This PR also syncs retryable writes unified tests from NODE-3733/DRIVERS-1385 that were missed in that work.
This PR also syncs transaction unified tests from DRIVERS-1539 that did not get individual language tickets.


test/integration/retryable-reads/retryable_reads.prose.spec.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { TopologyType } from '../../../src';
+
+const VALID_TOPOLOGIES = [
+  TopologyType.ReplicaSetWithPrimary,
+  TopologyType.Sharded,
+  TopologyType.LoadBalanced
+];
+
+describe('Retryable Reads (prose)', function () {
```
dariakp, line 12: See my comment on the retryable_reads.spec.test.js file - I think we should reserve this file for the lone prose test that will be implemented as part of avoiding connection storms

test/integration/transactions/transactions.spec.test.js
```diff
@@ -78,8 +78,16 @@ class TransactionsRunnerContext extends TestRunnerContext {
...
+// an authenticated server and 2) command monitoring is at the connection level so
+// when the handshake fails no command started event is emitted.
+const SKIP = [
+  'AbortTransaction succeeds after handshake network error',
+  'CommitTransaction succeeds after handshake network error'
+];
+
 describe('Transactions Spec Unified Tests', function () {
-  runUnifiedSuite(loadSpecTests(path.join('transactions', 'unified')));
+  runUnifiedSuite(loadSpecTests(path.join('transactions', 'unified')), SKIP);
```
dariakp, line 90: Similar comment here as on the retryable_reads spec test file regarding moving the manual implementations from prose to this file; it would also be good to name the test cases to match the skipped descriptions

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -196,3 +198,15 @@ async function turnOffFailPoint(client, name) {
     mode: 'off'
   });
 }
+
+// These tests are skipped because the driver 1) executes a ping when connecting to
+// an authenticated server and 2) command monitoring is at the connection level so
+// when the handshake fails no command started event is emitted.
+const SKIP = [
```
dariakp, line 205: same comment here about moving the manual spec test implementations from prose to this file and matching test case names

test/integration/retryable-writes/retryable_writes.spec.prose.test.ts
```diff
@@ -1,64 +1,139 @@
...
 
-  it('retryable writes raise an exception when using the MMAPv1 storage engine', async () => {
-    const failPoint = await client.db('admin').command({
-      configureFailPoint: 'failCommand',
-      mode: { times: 1 },
-      data: {
-        failCommands: ['insert'],
-        errorCode: 20, // MMAP Error code,
-        closeConnection: false
+  context('when errors occur in the handshake', function () {
```
dariakp, line 74: see comment in spec.test.js file

test/integration/transactions/transactions.prose.spec.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { TopologyType } from '../../../src';
+
+const VALID_TOPOLOGIES = [
+  TopologyType.ReplicaSetWithPrimary,
+  TopologyType.Sharded,
+  TopologyType.LoadBalanced
+];
+
+describe('Transactions (prose)', function () {
```
dariakp, line 12: see comment in spec.test.js file

dariakp (changes requested):
Just a couple of organizational notes; I verified the synced spec files but I haven't fully reviewed the actual manual implementations of the spec tests yet, I wanted to hear your thoughts on the potential alternate approach in case it actually makes things easier

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -28,3 +30,10 @@ describe('Retryable Reads', function () {
     );
   });
 });
+
+// These tests are skipped because the driver 1) executes a ping when connecting to
+// an authenticated server and 2) command monitoring is at the connection level so
+// when the handshake fails no command started event is emitted.
+describe.skip('Retryable Reads (unified)', function () {
```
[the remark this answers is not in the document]
durran: Good point. I'll update to skip the individual tests for future proofing.

test/integration/transactions/transactions.spec.test.js
```diff
@@ -119,3 +135,79 @@ describe('Transactions Spec Legacy Tests', function () {
...
+describe('Transactions', function () {
+  const dbName = 'retryable-handshake-tests';
+  const collName = 'coll';
+  const docs = [{ _id: 1, x: 11 }];
+  let client;
+  let db;
+  let coll;
+
+  beforeEach(function () {
+    if (
```
dariakp, line 148: should there be an auth requirement and a skip on serverless requirement in addition to the ones below?

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -28,3 +37,87 @@ describe('Retryable Reads', function () {
...
+    { _id: 2, x: 22 },
+    { _id: 3, x: 33 }
+  ];
+  let client;
+  let db;
+  let coll;
+
+  beforeEach(function () {
+    if (
+      semver.lt(this.configuration.buildInfo.version, '4.2.0') ||
```
dariakp, line 67: should there be an auth requirement here?

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -28,3 +37,87 @@ describe('Retryable Reads', function () {
...
+        configureFailPoint: 'failCommand',
+        mode: 'off'
+      });
+      await coll.drop();
+      await client.close();
+    }
+  });
+
+  context('when the handshake fails with a network error', function () {
+    it('retries the read', async function () {
```
dariakp, line 91: can we add a comment or update both test cases to clearly refer to the skipped tests that they are meant to implement?

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -28,3 +37,87 @@ describe('Retryable Reads', function () {
...
+      await client.connect();
+      await coll.insertMany(docs);
```
dariakp, line 93: can we move the setup to the a beforeEach in both test cases?

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -28,3 +37,87 @@ describe('Retryable Reads', function () {
...
+      await db.admin().command({
+        configureFailPoint: 'failCommand',
+        mode: { times: 2 },
+        data: {
+          failCommands: ['saslContinue', 'ping'],
+          closeConnection: true
+        }
+      });
+      const documents = await coll.find().toArray();
```
dariakp, line 102: in the original test, the find is with an `_id: 2`, in both test cases, I think we can keep that for consistency

dariakp (changes requested):
Looked through all the tests, just a few notes here:

test/integration/transactions/transactions.spec.test.js
```diff
@@ -119,3 +135,79 @@ describe('Transactions Spec Legacy Tests', function () {
...
+      await coll.drop();
+      await client.close();
+    }
+  });
+
+  context('when the handshake fails with a network error', function () {
+    it('retries the abort', async function () {
+      await client.connect();
+      await coll.insertMany(docs);
+      const session = client.startSession();
```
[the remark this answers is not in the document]
durran: Moved into the beforeEach.

test/integration/transactions/transactions.spec.test.js
```diff
@@ -119,3 +135,79 @@ describe('Transactions Spec Legacy Tests', function () {
...
+      session.startTransaction();
+      await db.admin().command({
```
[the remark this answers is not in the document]
durran: Moved the insert one to before the fail point now.

test/integration/transactions/transactions.spec.test.js
```diff
@@ -119,3 +135,79 @@ describe('Transactions Spec Legacy Tests', function () {
...
+      await session.abortTransaction();
+      await session.endSession();
+      const doc = await coll.findOne({ _id: 2 });
+      expect(doc).to.not.exist;
+    });
+
+    it('retries the commit', async function () {
+      await client.connect();
+      await coll.insertMany(docs);
+      const session = client.startSession();
```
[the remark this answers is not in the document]
durran: Moved up to the beforeEach.

test/integration/transactions/transactions.spec.test.js
```diff
@@ -119,3 +135,79 @@ describe('Transactions Spec Legacy Tests', function () {
...
+        configureFailPoint: 'failCommand',
+        mode: 'off'
+      });
+      await coll.drop();
+      await client.close();
+    }
+  });
+
+  context('when the handshake fails with a network error', function () {
+    it('retries the abort', async function () {
```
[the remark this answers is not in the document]
durran: Comment with test description added.

test/integration/transactions/transactions.spec.test.js
```diff
@@ -119,3 +135,79 @@ describe('Transactions Spec Legacy Tests', function () {
 
   generateTopologyTests(testSuites, testContext, testFilter);
 });
+
+describe('Transactions', function () {
```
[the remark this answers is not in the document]
durran: Changed description to indicate manual.

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -196,3 +206,81 @@ async function turnOffFailPoint(client, name) {
...
+  const dbName = 'retryable-handshake-tests';
+  const collName = 'coll';
+  const docs = [{ _id: 1, x: 11 }];
+  let client;
+  let db;
+  let coll;
+
+  beforeEach(function () {
+    if (
+      semver.lt(this.configuration.buildInfo.version, '4.2.0') ||
```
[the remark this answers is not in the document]
durran: I made it 4.2 since that is the minimum server version required in the actual spec test itself. Added auth requirement and no serverless requirement.

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -196,3 +206,81 @@ async function turnOffFailPoint(client, name) {
...
+const SKIP = [
+  'InsertOne succeeds after retryable handshake error',
+  'InsertOne succeeds after retryable handshake error ShutdownInProgress'
+];
+
+describe('Retryable Writes (unified)', function () {
+  runUnifiedSuite(loadSpecTests(path.join('retryable-writes', 'unified')), SKIP);
+});
+
+describe('Retryable Writes', function () {
```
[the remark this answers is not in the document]
durran: Updated description to indicate manual test.

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -196,3 +206,81 @@ async function turnOffFailPoint(client, name) {
...
+    await db?.admin().command({
+      configureFailPoint: 'failCommand',
+      mode: 'off'
+    });
+    await coll?.drop();
+    await client?.close();
+  });
+
+  context('when the handshake fails with a network error', function () {
+    it('retries the write', async function () {
```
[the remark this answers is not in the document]
durran: Added spec test description in comment.

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -196,3 +206,81 @@ async function turnOffFailPoint(client, name) {
...
+      await client.connect();
+      await coll.insertMany(docs);
```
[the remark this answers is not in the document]
durran: Moved setup to beforeEach.

test/integration/transactions/transactions.spec.test.js
durran: Added auth and serverless checks.

test/integration/retryable-reads/retryable_reads.spec.test.js
durran: Added auth and serverless checks.

test/integration/retryable-reads/retryable_reads.spec.test.js
durran: Added comment with skipped spec test description.

test/integration/retryable-reads/retryable_reads.spec.test.js
durran: Moved setup to beforeEach

test/integration/retryable-reads/retryable_reads.spec.test.js
durran: Updated to find one with _id of 2.

test/integration/transactions/transactions.spec.test.js
```diff
@@ -119,3 +135,79 @@ describe('Transactions Spec Legacy Tests', function () {
...
+      await db.admin().command({
+        configureFailPoint: 'failCommand',
+        mode: 'off'
+      });
+      await coll.drop();
+      await client.close();
+    }
+  });
+
+  context('when the handshake fails with a network error', function () {
```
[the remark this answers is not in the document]
durran: That ticket is a bit odd to me as it's mentioning heartbeat failures which I don't see happening in this scenario. We are failing on `saslContinue` and `ping` which are not being called from the monitor as the monitors don't auth.

test/integration/transactions/transactions.spec.test.js
dariakp: I think we should be able to include the runOn checks for server version, topology, and auth in the metadata on the `it` blocks (similar to how we do in `enumerate_databases.test.ts` for example), which would allow our runner to automatically attach the correct skip reason, then we can update the check below to only check for SERVERLESS explicitly and update the skipReason text accordingly

test/integration/transactions/transactions.spec.test.js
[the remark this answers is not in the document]
dariakp: I see, I think it may be worth adding a comment anyway to both of the manual blocks flagging DRIVERS-2032 until the corresponding spec tests are unskipped in the spec files themselves, just in case we do run into issues here

test/integration/retryable-writes/retryable_writes.spec.test.ts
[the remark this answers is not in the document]
dariakp: I think they need to be flipped

test/integration/retryable-writes/retryable_writes.spec.test.ts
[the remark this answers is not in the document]
dariakp: oops, I think I may have been looking at the wrong file for the server version requirements, disregard!

but, this test doesn't need a serverless check actually (or at least it doesn't have one in the spec file); also I think we can capture all the requirements in a metadata tag, which should automate the skipReason generation

test/integration/transactions/transactions.spec.test.js
```diff
@@ -119,3 +135,79 @@ describe('Transactions Spec Legacy Tests', function () {
...
+      await db.admin().command({
+        configureFailPoint: 'failCommand',
+        mode: { times: 2 },
+        data: {
+          failCommands: ['saslContinue', 'ping'],
+          closeConnection: true
+        }
+      });
+      await session.abortTransaction();
+      await session.endSession();
```
dariakp, line 192: I think this should go in the afterEach for both cases

test/integration/retryable-reads/retryable_reads.spec.test.js
dariakp: This one shouldn't need a serverless check according to the unified spec test file; and we can probably do the other checks via the test metadata so that we don't have to manually do the skipReason

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -28,3 +37,90 @@ describe('Retryable Reads', function () {
...
+    it('retries the read', async function () {
+      await db.admin().command({
+        configureFailPoint: 'failCommand',
+        mode: { times: 2 },
+        data: {
+          failCommands: ['saslContinue', 'ping'],
+          closeConnection: true
+        }
+      });
+      const doc = await coll.findOne({ _id: 2 });
```
dariakp, line 105: this is a `find` in the original test

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -28,3 +37,90 @@ describe('Retryable Reads', function () {
...
+    it('retries the read', async function () {
+      await db.admin().command({
+        configureFailPoint: 'failCommand',
+        mode: { times: 2 },
+        data: {
+          failCommands: ['saslContinue', 'ping'],
+          errorCode: 91 // ShutdownInProgress
+        }
+      });
+      const documents = await coll.find().toArray();
```
dariakp, line 122: -> `find({ _id: 2 })` for consistency with the original test (and update the assertion accordingly)

dariakp (changes requested):
A few more cleanup notes, but since @durran you'll be on vacation, I think the rest of the team can easily wrap this up ourselves, don't worry about it

test/integration/transactions/transactions.spec.test.js
dariakp: updated this

test/integration/retryable-reads/retryable_reads.spec.test.js
dariakp: updated this

test/integration/retryable-reads/retryable_reads.spec.test.js
dariakp: updated this

test/integration/retryable-reads/retryable_reads.spec.test.js
dariakp: updated this logic

test/integration/transactions/transactions.spec.test.js
dariakp: updated this

test/integration/retryable-writes/retryable_writes.spec.test.ts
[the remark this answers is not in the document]
dariakp: updated this

dariakp (approved):
I made some modifications to the tests addressing my previous comments, passing on the review to the rest of the team now

src/cmap/connect.ts
```diff
@@ -182,7 +184,12 @@ function performInitialHandshake(
           );
         }
         provider.auth(authContext, err => {
-          if (err) return callback(err);
+          if (err) {
+            if (err instanceof MongoError) {
```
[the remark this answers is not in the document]
dariakp: Good question, but I don't think that's the case. If you read the actual spec change, it references "any of the above retryable errors", such as "ShutdownInProgress" which would actually make their way into the driver as MongoServerError. Although it does make me wonder if we should be specific here, or if the fact that we are doing this during authentication is sufficient to ensure that any errors would qualify as retryable. @nbbeeken Do you happen to have any insight into this off the top of your head?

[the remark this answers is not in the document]
baileympearson: Hm, I'm not sure.  That's listed specifically for the retryable reads section.  The retryable writes section makes it seem like it's only network errors: 

> When the driver encounters a network error establishing an initial connection to a server,
>   it MUST add a RetryableWriteError label to that error if the MongoClient performing
>   the operation has the retryWrites configuration option set to true.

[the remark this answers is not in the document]
dariakp: No, if you look at https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#determining-retryable-errors
The retryable write error codes are similar to the reads, so this is to handle the server-side retryable errors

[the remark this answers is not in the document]
dariakp: It does bring up an interesting point though regarding whether handshake errors that don't have the retryable write label for new servers should be excluded here like we would do in the operation handler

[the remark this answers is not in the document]
dariakp: resolving in favor of follow up comment

test/integration/retryable-reads/retryable_reads.spec.test.js
```diff
@@ -28,3 +29,88 @@ describe('Retryable Reads', function () {
...
+    afterEach(async function () {
+      await db.admin().command({
+        configureFailPoint: 'failCommand',
+        mode: 'off'
+      });
+      await coll.drop();
+      await client.close();
+    });
+
+    context('when the handshake fails with a network error', function () {
```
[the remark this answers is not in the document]
dariakp: @durran just to double check - have you tried running *all* the currently skipped tests in all 3 files via the unified runner? do they all fail? I thought we had filter logic for the ping commands, so I'm curious to know what exactly is failing each set of expected monitoring events

dariakp (changes requested):

test/integration/retryable-reads/retryable_reads.spec.test.js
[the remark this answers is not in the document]
durran: I've updated the unified runner to detect now whether the tests explicitly ping and add the additional cmap events into the expected events list. A todo has been added into the runner code as a comment on what can be removed when NODE-2149 is finished.

src/cmap/connect.ts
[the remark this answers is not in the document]
durran: This has been updated to use `needsRetryableWriteLabel` and that code has been refactored to not take the `maxWireVersion` into account anymore, since the configure fail point error does not automatically add the label in 4.4+ servers. So the code basically just looks to see if the label is there or not and ignores the wire version.

src/error.ts
```diff
@@ -744,21 +744,18 @@ const RETRYABLE_WRITE_ERROR_CODES = new Set<number>([
 ]);
 
 export function needsRetryableWriteLabel(error: Error, maxWireVersion: number): boolean {
-  if (maxWireVersion >= 9) {
-    // 4.4+ servers attach their own retryable write error
-    return false;
-  }
   // pre-4.4 server, then the driver adds an error label for every valid case
```
nbbeeken, line 747: Discussed in standup/tech chat we should bring this logic back and find out why our driver has a wire version value when handling a handshake error.

nbbeeken (changes requested):

durran:
@nbbeeken What I've ended up doing here (since there are a few old stale comments) is adding a label to the handshake error, and then ensuring that if we are 4.4 and higher we don't add any additional labels except in the case where that label exists. This covers all the cases and all spec tests are passing. My changes to the spec tests are now removed.

nbbeeken (approved):
Sorry this took me a bit to get to, I think we're all set. I reran the one failure, but it looks transient

baileympearson (approved):

src/cmap/connect.ts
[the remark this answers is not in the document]
nbbeeken: Sorry I LGTM-ed but this looks outstanding still
@durran What're your thoughts on point 4? Copying the existing tests and changing the error code seems like it could "just work" 🤞  Maybe we merge this and open a follow up PR for the sake of ease of review, thoughts?

test/integration/retryable-reads/retryable_reads.spec.test.js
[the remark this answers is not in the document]
nbbeeken: We were able to fix the runner to run the unified tests, woohoo!

PR #3167, opened 2022-03-09
feat(NODE-3699): add support for `comment` field

baileympearson:
#### What is changing?

This PR adds support for the `comment` field.  It includes changes from three drivers tickets:

- [NODE-3699](https://jira.mongodb.org/browse/NODE-3699)
- [NODE-4035](https://jira.mongodb.org/browse/NODE-4035)
- [NODE-3973](https://jira.mongodb.org/browse/NODE-3973)

No.

#### Notes

Not modified:

- the legacy count comment specified a `$comment` field of type string.  The new comment field is supported in MongoDB 4.4+ and our legacy count command only runs for server versions <=3.2.
- There is pre-existing logic in `CommandOperation.executeCommand` to attach a `comment` to a command when executing the command if the comment is of type string.  All the new operations in the drivers spec attach the `comment` field to the command prior to calling `super.executeOperation`, so the logic in `CommandOperation.executeCommand` does not need to be updated.  I also decided to leave it as-it-is (instead of making the change to accept non-string comments in this location specifically) because I didn't want to change the behavior for commands which only support string comments.


src/operations/delete.ts
```diff
@@ -80,6 +94,10 @@ export class DeleteOperation extends CommandOperation<Document> {
       command.let = options.let;
     }
 
+    if (options.comment) {
```
baileympearson, line 97: We were applying the `comment` option in the delete statement portion of the command, and not at the top level.  See https://docs.mongodb.com/manual/reference/command/delete/

nbbeeken: Does it work in both places? the linked example only shows it in the top level document

src/operations/find_and_modify.ts
```diff
@@ -140,6 +149,10 @@ class FindAndModifyOperation extends CommandOperation<Document> {
       this.cmdBase.let = options.let;
     }
 
+    if (options.comment) {
```
nbbeeken, line 152: I think our if stmt filters for comment should be more specific given the wide array of types that can be passed in, I don't see why number `0` shouldn't be allowed. But maybe even null should be permitted as well, this might be one of the few cases where checking explicitly for undefined is warrented. Thoughts?

src/operations/update.ts
```diff
@@ -107,6 +107,10 @@ export class UpdateOperation extends CommandOperation<Document> {
       command.let = options.let;
     }
 
+    if (options.comment) {
```
nbbeeken, line 110: Falesy discussion in the other thread.

src/change_stream.ts
```diff
@@ -46,9 +46,13 @@ const kClosed = Symbol('closed');
...
-const CURSOR_OPTIONS = ['batchSize', 'maxAwaitTimeMS', 'collation', 'readPreference'].concat(
-  CHANGE_STREAM_OPTIONS
-);
+const CURSOR_OPTIONS = [
+  'batchSize',
+  'maxAwaitTimeMS',
+  'collation',
+  'readPreference',
+  'comment'
+].concat(CHANGE_STREAM_OPTIONS);
```
nbbeeken, line 55: We'll have a merge conflict to resolve here when #3169 goes in. (no action, just so we catch any issue.)

src/operations/delete.ts
nbbeeken: It's still in the DeleteStatement type, should it just be removed from there? I see its public but if it never worked then maybe it was bug that it was defined there

test/unit/connection_string.test.ts
```diff
@@ -291,9 +291,10 @@ describe('Connection String', function () {
 
   describe('resolveSRVRecord()', () => {
     const resolveSRVRecordAsync = promisify(resolveSRVRecord);
+    const sandbox = sinon.createSandbox();
```
baileympearson, line 294: The unit tests I added in get_more.test.ts started failing when I ran the full test suite.  It seems like these tests were clear the mock I was setting in `it('executes a getMore on the provided server')`.  The solution is to use a sinon sandbox.

src/operations/get_more.ts
```diff
@@ -12,20 +12,35 @@ import { AbstractOperation, Aspect, defineAspects, OperationOptions } from './op
...
+   * Comment to apply to the operation.
+   *
+   * getMore only supports 'comment' in server versions 4.4 and above.
+   */
+  comment?: any;
   /** Number of milliseconds to wait before aborting the query. */
   maxTimeMS?: number;
 }
 
+function filterOptions(
```
nbbeeken, line 25: We have a filterOptions in utils.ts, maybe we can reuse / update that one?

src/cursor/abstract_cursor.ts
```diff
@@ -620,94 +629,111 @@ export abstract class AbstractCursor<
...
 
           if (response.cursor.ns) {
-            cursor[kNamespace] = ns(response.cursor.ns);
+            this[kNamespace] = ns(response.cursor.ns);
           }
 
-          cursor[kDocuments] = response.cursor.firstBatch;
+          this[kDocuments] = response.cursor.firstBatch;
         } else {
           // NOTE: This is for support of older servers (<3.2) which do not use commands
```
nbbeeken, line 670: If this comment is true we can delete this `else` block

src/cursor/abstract_cursor.ts
```diff
@@ -620,94 +629,111 @@ export abstract class AbstractCursor<
...
+        return this[kTopology].selectServer(ReadPreference.primaryPreferred, {}, err => {
           if (err) return callback(err);
-          return next(cursor, blocking, callback);
+          return this.__initialize(callback);
         });
-      } else if (cursor[kTopology].hasSessionSupport()) {
-        cursor[kSession] = cursor[kTopology].startSession({ owner: cursor, explicit: false });
+      } else if (this[kTopology].hasSessionSupport()) {
+        this[kSession] = this[kTopology].startSession({ owner: this, explicit: false });
       }
```
nbbeeken, line 649: A bit of scope creep but I think it's an easy win, if we make ExecutionResult generic now, and then also _initialize generic, we should be able to pass through and interface that has the minimal properties used in this code here, keeping it correct.

```
this._initialize<
      | { cursor: { id: Long | number; ns?: string; firstBatch: TSchema[] } }
      | { cursorId: number | Long; documents: TSchema[] } // old way, can be removed
    >(this[kSession], (err, state) => {
```

test/unit/operations/get_more.test.ts
```diff
@@ -0,0 +1,146 @@
...
+describe('GetMoreOperation', function () {
+  const ns = new MongoDBNamespace('db.coll');
+  const cursorId = Object.freeze(Long.fromNumber(1));
+  const options = {
+    batchSize: 100,
+    maxTimeMS: 500,
+    readPreference: ReadPreference.primary
+  };
+
+  describe('#constructor', function () {
```
nbbeeken, line 24: Doesn't have to land in this test file, is there a way we can check that sending falsey comments works since we went out of our way to check only for undefined?

nbbeeken (changes requested):

src/operations/get_more.ts
```diff
@@ -12,20 +12,35 @@ import { AbstractOperation, Aspect, defineAspects, OperationOptions } from './op
...
 export class GetMoreOperation extends AbstractOperation {
   cursorId: Long;
   override options: GetMoreOptions;
 
   constructor(ns: MongoDBNamespace, cursorId: Long, server: Server, options: GetMoreOptions = {}) {
     super(options);
-    this.options = options;
+
+    // comment on getMore is only supported for server versions 4.4 and above
+    const invalidOptions = maxWireVersion(server) < 9 ? new Set(['comment']) : new Set<string>();
```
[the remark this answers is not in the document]
baileympearson: Done!

src/operations/get_more.ts
baileympearson: Done!  Let me know what you think of my approach.  

The existing utility filtered based on an allowlist, whereas I needed a denylist.  I did think that it might be worthwhile to have since we'll probably filter objects like this in the future so I just added a second utility.

test/unit/operations/get_more.test.ts
baileympearson: Yeah, good catch.

I added an integration test to cover this scenario.  Rather than testing every option, I centralized the logic that applies the comment into a utility function and just tested a single scenario.  If you'd prefer not to have a function call (I think Daria has in the past expressed concerns about using function calls for simple checks like this due to performance considerations), I can write a test for each operation that supports `comment`.

src/operations/find_and_modify.ts
baileympearson: Addressed! d

src/change_stream.ts
baileympearson: Done!

src/cursor/abstract_cursor.ts
```diff
@@ -42,6 +42,8 @@ const kInitialized = Symbol('initialized');
 const kClosed = Symbol('closed');
 /** @internal */
 const kKilled = Symbol('killed');
+/** @internal */
```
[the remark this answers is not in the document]
baileympearson: Yeah, good call.  Done

src/change_stream.ts
```diff
@@ -411,6 +412,15 @@ export interface ChangeStreamCursorOptions extends AbstractCursorOptions {
...
   startAfter?: boolean;
+  /**
+   * Comment to apply to the operation.
+   *
+   * In server versions pre-4.4, 'comment' must be string.  A server
+   * error will be thrown if any other type is provided.
+   *
+   * In server versions 4.4 and above, 'comment' can be any valid BSON type.
+   */
+  comment?: any;
```
dariakp, line 423: why are we adding this explicitly here if it already extends `AbstractCursorOptions` which has it defined?

src/cursor/abstract_cursor.ts
dariakp: seconding the follow up thought

src/operations/aggregate.ts
```diff
@@ -31,6 +30,16 @@ export interface AggregateOptions extends CommandOperationOptions {
...
   let?: Document;
+  /**
+   * Comment to apply to the operation.
+   *
+   * In server versions pre-4.4, 'comment' must be string.  A server
+   * error will be thrown if any other type is provided.
+   *
+   * In server versions 4.4 and above, 'comment' can be any valid BSON type.
+   */
+  comment?: any;
```
dariakp, line 41: why do we add it explicitly here if it's already defined on `CommandOperationOptions`?

src/operations/delete.ts
```diff
@@ -12,8 +12,15 @@ import { Aspect, defineAspects, Hint } from './operation';
...
-  comment?: string | Document;
+  /**
+   * Comment to apply to the operation.
+   *
+   * In server versions pre-4.4, 'comment' must be string.  A server
+   * error will be thrown if any other type is provided.
+   *
+   * In server versions 4.4 and above, 'comment' can be any valid BSON type.
+   */
+  comment?: any;
```
dariakp, line 23: this is already defined on command operation options, do we need it here?

src/operations/find.ts
```diff
@@ -46,8 +46,15 @@ export interface FindOptions<TSchema extends Document = Document> extends Comman
...
-  comment?: string | Document;
+  /**
+   * Comment to apply to the operation.
+   *
+   * In server versions pre-4.4, 'comment' must be string.  A server
+   * error will be thrown if any other type is provided.
+   *
+   * In server versions 4.4 and above, 'comment' can be any valid BSON type.
+   */
+  comment?: any;
```
dariakp, line 57: do we need this if it's already on command operation options?

src/operations/get_more.ts
```diff
@@ -25,7 +29,10 @@ export class GetMoreOperation extends AbstractOperation {
 
   constructor(ns: MongoDBNamespace, cursorId: Long, server: Server, options: GetMoreOptions = {}) {
     super(options);
-    this.options = options;
+
+    // comment on getMore is only supported for server versions 4.4 and above
+    this.options = maxWireVersion(server) < 9 ? disallowOptions(options, ['comment']) : options;
```
dariakp, line 34: does the spec say that we have to silently ignore the option on servers below 4.4?

src/utils.ts
```diff
@@ -189,6 +189,19 @@ export function filterOptions(options: AnyOptions, names: string[]): AnyOptions
   return filterOptions;
 }
 
+/** @internal */
+export function disallowOptions(options: AnyOptions, denylist: string[]): AnyOptions {
+  const accumulatedOptions: AnyOptions = {};
```
dariakp, line 194: Do we worry about the performance cost of an extra iteration over the keys and also creating a new object here? I like the idea of not mutating input as much as anyone, but have we considered deleting the keys or throwing an error on the forbidden options? I think ideally we'd want to create a new set of options from the user-provided settings exactly once, and then modify them as needed; as it happens, the only place we use this helper - the get more operation - uses the object generated via destructuring in the AbstractOperation definition, so we are already working with a generated list of options.

Have we considered replacing the use of this helper in the getMore operation with a simple `delete options.comment` (or, again, just throwing a validation error)?

dariakp (changes requested):
Leaving feedback on everything except the tests.

test/tools/unified-spec-runner/match.ts
```diff
@@ -184,6 +184,9 @@ export function resultCheck(
     } else if (Long.isLong(expected) && typeof actual === 'number') {
       // Long requires special equality check
       expect(expected.equals(actual)).to.be.true;
+    } else if (Number.isNaN(actual) && Number.isNaN(expected)) {
```
dariakp, line 187: which test added a case where this is applicable?

test/tools/unified-spec-runner/schema.ts
```diff
@@ -20,9 +20,10 @@ export interface UnifiedSuite {
   description: string;
   schemaVersion: string;
   runOnRequirements?: RunOnRequirement[];
-  createEntities?: [EntityDescription, ...EntityDescription[]];
-  initialData?: [CollectionData, ...CollectionData[]];
-  tests: [Test, ...Test[]];
+  createEntities?: EntityDescription[];
+  /** Data inserted before **all tests */
+  initialData?: CollectionData[];
+  tests: Test[];
```
dariakp, line 26: I think the previous typings were trying to assert that the arrays have to have at least one member, was that incorrect?

test/spec/change-streams/unified/change-streams.json
```diff
@@ -1,10 +1,21 @@
 {
   "description": "change-streams",
   "schemaVersion": "1.0",
+  "runOnRequirements": [
```
dariakp, line 4: @nbbeeken Please double check to make sure all the tests in the spec directory that have been added match to the expected spec updates

dariakp (changes requested):
I still need to review the builder and the falsy values test, but here's feedback on the rest of the tests

test/tools/unified-spec-runner/operations.ts
```diff
@@ -206,17 +206,16 @@ operations.set('createChangeStream', async ({ entities, operation }) => {
...
+    batchSize: operation.arguments.batchSize,
+    comment: operation.arguments.comment
   });
-  changeStream.eventCollector = new EventCollector(changeStream, ['init', 'change', 'error']);
 
-  return new Promise((resolve, reject) => {
-    const timeout = setTimeout(() => {
-      reject(new Error('Change stream never started'));
-    }, 2000);
+  changeStream.eventCollector = new EventCollector(changeStream, ['init', 'error']);
```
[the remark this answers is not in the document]
baileympearson: I'm going to respond here about this comment as well as for your question about why I skipped the `'Test consecutive resume'` test.

Our unified spec runner was not spec compliant for change streams.  The `createChangeStream` operation is only supposed to perform the initial `aggregate` call.  Our stream-based implementation of change streams is lazy - we don't perform any calls until a `change` listener is added and once one is, we immediately perform the aggregate call and _and_ start sending getMores.  One solution would be to refactor our change streams implementation to perform the `aggregate` call on creation, and `getMores` when a `change` listener is added but that would be a decent refactor.  Instead, our unified tests now use the cursor-based implementation.

`'Test consecutive resume'` now fails because our iterator-based change stream implementation doesn't resume properly (see https://jira.mongodb.org/browse/NODE-4126).  I'll update the jira ticket to reflect this.

I think we're safe to actually remove the event collector entirely though.

src/operations/find.ts
baileympearson: See my comment in change_stream.ts

src/operations/delete.ts
baileympearson: See my comment in change_stream.ts

src/cmap/connection.ts
```diff
@@ -124,7 +124,15 @@ export interface GetMoreOptions extends CommandOptions {
...
-  comment?: Document | string;
+  /**
+   * Comment to apply to the operation.
+   *
+   * In server versions pre-4.4, 'comment' must be string.  A server
+   * error will be thrown if any other type is provided.
+   *
+   * In server versions 4.4 and above, 'comment' can be any valid BSON type.
+   */
+  comment?: any;
```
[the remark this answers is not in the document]
baileympearson: I'll tackle the second question first.  `undefined` is not permitted (well, it'll be permitted by typescript but not included as a comment) and null is allowed.  `null` is the preferred nullish value in bson and `undefined` is deprecated as a bson type.  We have to draw the line somewhere though, since `comment` is an optional property we have to have some way to let users not specify a comment.  I think permitting any value but `undefined` is reasonable.  I could add a note in the docstring though to make it explicit.  

The second question is more interesting and Neal might know better than I do.  I considered creating a `BSONLike` type that encompassed all possible BSON values that we support.  But in my experimentation, I couldn't find a type or construct that BSON wouldn't serialze.

src/cmap/connection.ts
```diff
@@ -574,6 +582,11 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     if (typeof options.maxAwaitTimeMS === 'number') {
       getMoreCmd.maxTimeMS = options.maxAwaitTimeMS;
     }
+    // we check for undefined specifically here to allow falsy values
+    // eslint-disable-next-line no-restricted-syntax
+    if (options.comment !== undefined) {
```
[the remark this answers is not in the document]
nbbeeken: Yes, I guess I didn't capture this in an earlier review comment. `comment` can be any BSON type, of which null is an option. We could instead change this code to be `if('comment' in options)` or some similar key existence check to permit undefined to be mapped on to the command. But by default our driver has ignoreUndefined = true so it won't actually make it to the server. If ignoreUndefined = false, undefined gets converted to a BSON `null` since BSON undefined is deprecated. Since you'd end up with null anyway in this edge case usage I think it fair for our driver to only accept/handle null. Does all that track?

test/tools/unified-spec-runner/match.ts
nbbeeken: In addition to the new constructed tests, our existing tests are in EJSON and could give us NaN values like so:
```ts
{"$numberDouble": "Infinity" | "-Infinity" | "NaN"> }
```

While we're here can we add a proper equality check for `0` and `-0`. 
- if expected and actual are both typeof number, and both `=== 0`
- then assert `Object.is(actual, expected)` is true.

```js
Object.is(0, -0) // false
```

test/tools/unified-spec-runner/schema.ts
nbbeeken: Yea can confirm, I generated the definitions from the schema and the schema has an enforcement of "at least 1" which caused this `[type, ...type[]]` to be generated, but it has more friction in practice that doesn't help us. I think I ran into issues with array methods like `.map` not typing easily.

test/spec/change-streams/unified/change-streams.json
nbbeeken: Confirmed, everything is synced and up to date. I caught the listCollection case that we may want to consider testing.

src/change_stream.ts
```diff
@@ -411,6 +412,15 @@ export interface ChangeStreamCursorOptions extends AbstractCursorOptions {
   startAtOperationTime?: OperationTime;
   resumeAfter?: ResumeToken;
   startAfter?: boolean;
```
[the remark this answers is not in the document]
dariakp: sure, but is there any harm in updating just the typescript type here to be correct? since neither of the other tickets would directly address this, and it *is* a one line change

src/change_stream.ts
dariakp: I'm not sure but the reason I'm pointing it out here is that this makes it inconsistent with most of the other places we have declared types, and it also defeats the purpose of "extends" in the way we've used it so far in all these interfaces (i.e., as the shared base, so that we don't have to copy paste exactly the same description in 50 different files). If we're going to declare all properties on all types individually, I'd rather we just have separate types. But I think we could also have a documentation solution - i.e., have something that follows the tree and compiles the complete types for doc purposes. Either way, the less consistency we have in how we do things now, the more headache it will be to change everything to be uniform in one way or another later on, so my vote is to keep our convention.

test/tools/unified-spec-runner/schema.ts
dariakp: sounds good to me

test/spec/change-streams/unified/change-streams.json
dariakp: Let's leave this comment open to track that additional test

test/tools/utils.ts
```diff
@@ -87,8 +96,13 @@ export class EventCollector {
    * Will only return one event at a time from the front of the list
    * Useful for iterating over the events in the order they occurred
    */
-  waitAndShiftEvent(eventName: string): Promise<Record<string, any>> {
-    return new Promise<Record<string, any>>((resolve, reject) => {
+  waitAndShiftEvent(eventName) {
```
dariakp, line 99: I believe this is now no longer used with the Promise.race having been removed

dariakp (changes requested):

src/cmap/connection.ts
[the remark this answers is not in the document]
dariakp: per slack conversation, let's update to `unknown`

src/operations/get_more.ts
dariakp: the spec is kind of vague on this account, it just says "do not send comment", not "warn" or anything like that; so I think in this case it's fine to just delete

src/utils.ts
dariakp: Yeah, I agree in general, hence the point about making sure we create the new set of options exactly once. I think we should ensure that the top level input options are never mutated, but use the most performant way of manipulating the options as they are passed around internally. What I was saying though is that we are safe in getMore because the `super` class also spreads the options, so the input options in `AbstractCursor` get spread before they are passed to `GetMoreOperation`, and then those shallow copied options are spread *again* in the `AbstractOperation` constructor, and this helper then makes yet another copy of the copy of a copy.

src/change_stream.ts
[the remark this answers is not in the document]
nbbeeken: Oh yea, in favor of correcting `startAfter`. I was only answering the question on why it's unknown

src/cursor/abstract_cursor.ts
nbbeeken: Good catch on the "raw" case, yea let's do this later.

dariakp: just to confirm, are we thinking of doing this in NODE-3286? if so, can we capture both of your comments there? (both Neal's suggestion and bailey's note about raw)

src/utils.ts
baileympearson: `AbstractOperation` does copy bson options, but not other user set options.  You're correct that we're safe here because we spread in `AbstractCursor` but it seems like we could easily forget that the copying is necessary.  I could write a regression test to prevent this but I think, unless there is a demonstratable performance degradation, copying is simpler.  Maybe @nbbeeken could weigh in too?

test/unit/connection_string.test.ts
baileympearson: I refactored the get_more unit tests so these changes are no longer necessary

test/tools/utils.ts
baileympearson: Removed!

test/spec/change-streams/unified/change-streams.json
baileympearson: I added tests for listCollection, listDatabases and listIndexes!

test/tools/unified-spec-runner/match.ts
baileympearson: Added the case for +/- zero!

test/unit/operations/get_more.test.ts
```diff
@@ -0,0 +1,146 @@
...
+        const operation = new GetMoreOperation(ns, cursorId, server, optionsWithComment);
+        expect(operation.options).to.deep.equal(optionsWithComment);
+      });
+    });
+  });
+
+  describe('#execute', function () {
+    context('when the server is the same as the instance', function () {
+      const server = new Server(new Topology([], {} as any), new ServerDescription(''), {} as any);
+      const session = sinon.createStubInstance(ClientSession);
```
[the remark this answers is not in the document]
baileympearson: I refactored these two tests so that the stubs are created inside the test, rather than in the contexts.  I chose to just create them in the tests because there wasn't much that was used between the two tests.

src/utils.ts
baileympearson: Removed in favor of deleting `options.comment`

src/change_stream.ts
baileympearson: I removed the redundant type declarations.

src/change_stream.ts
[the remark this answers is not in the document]
baileympearson: Fixed!

src/cmap/connection.ts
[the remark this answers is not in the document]
baileympearson: Done!

test/tools/unified-spec-runner/match.ts
```diff
@@ -184,6 +184,18 @@ export function resultCheck(
...
+      return;
+    } else if (
+      typeof expected === 'number' &&
+      typeof actual === 'number' &&
+      expected === 0 &&
+      actual === 0
+    ) {
+      // case to handle +0 and -0
+      expect(Object.is(expected, actual)).to.be.true;
+      return;
```
nbbeeken, line 198: Thanks for the fix! No need to return here, just want to keep it consistent with the other blocks

src/cursor/abstract_cursor.ts
nbbeeken: Well so maybe my code block over did it, I was just thinking right now the `state.response` type is Document, if we type it to be:
```ts
type x = { cursor: { id: Long | number; ns?: string; firstBatch: unknown[] } }
```
We can at least get assertions on `id` and `ns` but the schema isn't actually relevant to this function at all. So my goal wasn't to make the schema go lower but instead type our expectation of the response shape (for example the potential Long or number for id). Seems like a small win

nbbeeken: Or another option, could be to type put the annotation on the variable we make:
```
const response: Record<string, unknown> = state.response;
// This would make sure our assertions about each of the properties `id` and `ns` is type safe
```

test/unit/operations/get_more.test.ts
nbbeeken: Absolute fire test builder 🔥 thank you!

test/spec/change-streams/unified/change-streams.json
nbbeeken: Test LGTM, some organizational requests discussed on slack

src/cursor/abstract_cursor.ts
baileympearson: Unless you feel strongly about this particular fix, I'm going leave it as-is.  I think that rather than retroactively typing this function with a type that matches how its used, we should try and fix this generally in the follow up ticket.

test/spec/change-streams/unified/change-streams.json
baileympearson: Done!

test/tools/unified-spec-runner/match.ts
baileympearson: I removed the return from both of the new blocks.

nbbeeken (approved):
LGTM

test/integration/enumerate_collections.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    databaseName: '',
+    documents: [
+      { _id: 1, x: 11 },
+      { _id: 2, toBeDeleted: true } // This should only be used by the delete test
+    ]
+  })
+  .databaseName('listCollections-with-falsy-values')
+  .test(
+    new TestBuilder('listCollections should not send comment for server versions < 4.4')
+      .runOnRequirement({ maxServerVersion: '4.4.0' })
```
dariakp, line 16: technically, the bound is supposed to be inclusive, so this test would, in fact, run and fail on `4.4.0`, it's just that our `4.4` server in the pipeline is the latest patch, which causes this test not to run there; if you look at the drivers spec, you'll see `maxServerVersion` is defined with `x.y.99`, where `y` is the highest minor to which the test would be applicable

dariakp (changes requested):
just a couple of minor but important notes

dariakp (approved):

PR #3168, opened 2022-03-10
docs: build docs from latest main

baileympearson:
#### What is changing?

Docs built from @nbbeeken 's changes so we can update other PRs and get them merged.

No.


nbbeeken (approved):
If we haven't already, let's see if we can disable evergreen based on changes occurring only in this subdirectory.

Thanks for the docs!

PR #3169, opened 2022-03-10
fix(NODE-4069): remove 'default' from options for fullDocument field in change stream options

baileympearson:
#### What is changing?

This PR removes the default value of `'default'` for the `fullDocument` field in the ChangeStreamOptions.

This PR also adds tests to options that are set in the `createChangeStreamCursor` method.  It seemed too out of scope to write tests to fully test the constructor, but at the very least I could add tests for options that were set in the ChangeStreamCursor constructor from this method.

No.

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -46,9 +46,14 @@ const kClosed = Symbol('closed');
 const kMode = Symbol('mode');
 
 const CHANGE_STREAM_OPTIONS = ['resumeAfter', 'startAfter', 'startAtOperationTime', 'fullDocument'];
-const CURSOR_OPTIONS = ['batchSize', 'maxAwaitTimeMS', 'collation', 'readPreference'].concat(
-  CHANGE_STREAM_OPTIONS
-);
+
+const CURSOR_OPTIONS = [
+  'batchSize',
+  'maxAwaitTimeMS',
+  'collation',
+  'readPreference',
+  ...CHANGE_STREAM_OPTIONS
+];
 
 const CHANGE_DOMAIN_TYPES = {
   COLLECTION: Symbol('Collection'),
@@ -94,8 +99,8 @@ export interface PipeOptions {
  * @public
  */
 export interface ChangeStreamOptions extends AggregateOptions {
-  /** Allowed values: ‘default’, ‘updateLookup’. When set to ‘updateLookup’, the change stream will include both a delta describing the changes to the document, as well as a copy of the entire document that was changed from some time after the change occurred. */
-  fullDocument?: string;
+  /** Allowed values: ‘updateLookup’. When set to ‘updateLookup’, the change stream will include both a delta describing the changes to the document, as well as a copy of the entire document that was changed from some time after the change occurred. */
+  fullDocument?: 'updateLookup';
   /** The maximum amount of time for the server to wait on new documents to satisfy a change stream query. */
   maxAwaitTimeMS?: number;
   /** Allows you to start a changeStream after a specified event. See {@link https://docs.mongodb.com/manual/changeStreams/#resumeafter-for-change-streams|ChangeStream documentation}. */
@@ -576,8 +581,14 @@ function createChangeStreamCursor<TSchema>(
   changeStream: ChangeStream<TSchema>,
   options: ChangeStreamOptions
 ): ChangeStreamCursor<TSchema> {
-  const changeStreamStageOptions: Document = { fullDocument: options.fullDocument || 'default' };
-  applyKnownOptions(changeStreamStageOptions, options, CHANGE_STREAM_OPTIONS);
+  const changeStreamStageOptions: Document = applyKnownOptions(
+    {
+      fullDocument: options.fullDocument
+    },
+    options,
+    CHANGE_STREAM_OPTIONS
+  );
+
   if (changeStream.type === CHANGE_DOMAIN_TYPES.CLUSTER) {
     changeStreamStageOptions.allChangesForCluster = true;
   }
@@ -606,11 +617,11 @@ function createChangeStreamCursor<TSchema>(
 }
 
 function applyKnownOptions(target: Document, source: Document, optionNames: string[]) {
-  optionNames.forEach(name => {
-    if (source[name]) {
-      target[name] = source[name];
+  for (const option of optionNames) {
+    if (source[option]) {
+      target[option] = source[option];
     }
-  });
+  }
 
   return target;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -166,7 +166,6 @@ export type {
 export type { OrderedBulkOperation } from './bulk/ordered';
 export type { UnorderedBulkOperation } from './bulk/unordered';
 export type {
-  ChangeStream,
   ChangeStreamCursor,
   ChangeStreamCursorOptions,
   ChangeStreamDocument,
@@ -178,6 +177,7 @@ export type {
   ResumeToken,
   UpdateDescription
 } from './change_stream';
+export { ChangeStream } from './change_stream';
 export type {
   AuthMechanismProperties,
   MongoCredentials,
--- a/test/integration/change-streams/change_stream.test.js
+++ b/test/integration/change-streams/change_stream.test.js
@@ -7,7 +7,7 @@ const { EventCollector, getSymbolFrom } = require('../../tools/utils');
 const { expect } = require('chai');
 
 const sinon = require('sinon');
-const { Long, ReadPreference, MongoNetworkError } = require('../../../src');
+const { Long, ReadPreference, MongoNetworkError, ChangeStream } = require('../../../src');
 
 const crypto = require('crypto');
 const { isHello } = require('../../../src/utils');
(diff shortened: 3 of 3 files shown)


src/change_stream.ts
```diff
@@ -94,8 +99,8 @@ export interface PipeOptions {
  * @public
  */
 export interface ChangeStreamOptions extends AggregateOptions {
-  /** Allowed values: ‘default’, ‘updateLookup’. When set to ‘updateLookup’, the change stream will include both a delta describing the changes to the document, as well as a copy of the entire document that was changed from some time after the change occurred. */
-  fullDocument?: string;
+  /** Allowed values: ‘updateLookup’. When set to ‘updateLookup’, the change stream will include both a delta describing the changes to the document, as well as a copy of the entire document that was changed from some time after the change occurred. */
+  fullDocument?: 'updateLookup';
```
nbbeeken, line 103: I think we may have to keep this as string type to avoid breakage. Maybe let's add a type test and a TODO so we can change this in v5?

src/change_stream.ts
```diff
@@ -576,8 +581,14 @@ function createChangeStreamCursor<TSchema>(
   changeStream: ChangeStream<TSchema>,
   options: ChangeStreamOptions
 ): ChangeStreamCursor<TSchema> {
-  const changeStreamStageOptions: Document = { fullDocument: options.fullDocument || 'default' };
-  applyKnownOptions(changeStreamStageOptions, options, CHANGE_STREAM_OPTIONS);
+  const changeStreamStageOptions: Document = applyKnownOptions(
```
nbbeeken, line 584: Up to you, but maybe this helper isn't worth having I'm thinking that the array we're passing in here is exactly 4 keys it might be worth it to just hardcode the work since it's reletively small.

src/index.ts
```diff
@@ -178,6 +177,7 @@ export type {
   ResumeToken,
   UpdateDescription
 } from './change_stream';
+export { ChangeStream } from './change_stream';
```
nbbeeken, line 180: Why does this need to change? A ChangeStream shouldn't be directly constructed so I think its okay that we only make it accessible via the `watch` calls.

test/integration/change-streams/change_stream.test.js
```diff
@@ -187,6 +187,76 @@ describe('Change Streams', function () {
   });
   afterEach(async () => await mock.cleanup());
 
+  context('ChangeStreamCursor options', function () {
+    let client, db, collection;
+
+    beforeEach(async function () {
+      client = this.configuration.newClient();
```
nbbeeken, line 194: nit `client = await this.configuration.newClient().connect();`

test/integration/change-streams/change_stream.test.js
```diff
@@ -187,6 +187,76 @@ describe('Change Streams', function () {
...
+      collection = undefined;
+    });
+
+    context('fullDocument', () => {
+      it('sets fullDocument to `undefined` if no value is passed', function () {
+        const changeStream = new ChangeStream(client);
+
+        expect(changeStream.cursor).to.haveOwnProperty('pipeline');
+        const pipelineOptions = changeStream.cursor.pipeline[0].$changeStream;
+        expect(pipelineOptions).to.haveOwnProperty('fullDocument').to.be.undefined;
```
nbbeeken, line 213: Tests look good to me as is, so no need to change, but I've tried to get in the habbit of using the nested feature when splunking deep into an object, at least when there are questions of nullishness

`expect(changeStream).to.have.nested.property('cursor.pipeline[0].$changeStream.fullDocument', undefined);`

If anything in the chain is nullish we crash without a useful assertion printout, YMMV depending, maybe this case is overkill.

package.json
```diff
@@ -122,6 +122,7 @@
     "check:socks5": "mocha --config test/manual/mocharc.json test/manual/socks5.test.ts",
     "check:csfle": "mocha --config test/mocha_mongodb.json test/integration/client-side-encryption",
     "check:snappy": "mocha test/unit/assorted/snappy.test.js",
+    "eslint:fix": "npm run check:eslint -- --fix",
```
nbbeeken, line 125: I don't mind the addition but any chance we could use the check prefix? `check:fix`,`check:fixlint`, `check:fixeslint`

test/integration/change-streams/change_stream.test.js
```diff
@@ -187,6 +187,76 @@ describe('Change Streams', function () {
...
+    afterEach(async function () {
+      await client.close();
+      client = undefined;
+      db = undefined;
+      collection = undefined;
+    });
+
+    context('fullDocument', () => {
+      it('sets fullDocument to `undefined` if no value is passed', function () {
+        const changeStream = new ChangeStream(client);
```
nbbeeken, line 209: Why not use `.watch()` maybe I missed something?

nbbeeken (changes requested):

package.json
baileympearson: what about `fix:eslint`?  None of our other `check` prefixes actually modify any code, so it doesn't feel right to use `check` as the prefix.

test/integration/change-streams/change_stream.test.js
baileympearson: :smart: Yes, good call!

package.json
nbbeeken: Sure! that seems good to me I could see other fixes being added, like maybe a fix to run prettier on all our readme(s)

test/integration/change-streams/change_stream.test.js
baileympearson: I went with this approach - the only concern I have is that when checking for `undefined`, any typo in the string will cause a false negative in the test

src/change_stream.ts
```diff
@@ -94,7 +104,7 @@ export interface PipeOptions {
  * @public
  */
 export interface ChangeStreamOptions extends AggregateOptions {
-  /** Allowed values: ‘default’, ‘updateLookup’. When set to ‘updateLookup’, the change stream will include both a delta describing the changes to the document, as well as a copy of the entire document that was changed from some time after the change occurred. */
+  /** Allowed values: ‘updateLookup’. When set to ‘updateLookup’, the change stream will include both a delta describing the changes to the document, as well as a copy of the entire document that was changed from some time after the change occurred. */
```
nbbeeken, line 107: Looks like smart quotes got used here, not necessarily not allowed but I think ascii single quote is more common.

src/change_stream.ts
```diff
@@ -568,25 +578,50 @@ function setIsIterator<TSchema>(changeStream: ChangeStream<TSchema>): void {
...
-  const changeStreamStageOptions: Document = { fullDocument: options.fullDocument || 'default' };
-  applyKnownOptions(changeStreamStageOptions, options, CHANGE_STREAM_OPTIONS);
-  if (changeStream.type === CHANGE_DOMAIN_TYPES.CLUSTER) {
-    changeStreamStageOptions.allChangesForCluster = true;
-  }
-
+  const changeStreamStageOptions = createChangeStreamStageOptions(
+    changeStream,
+    changeStreamOptions
+  );
```
nbbeeken, line 613:
```suggestion
  const changeStreamStageOptions = {
    resumeAfter: options.resumeAfter != null ? options.resumeAfter : undefined,
    startAfter: Boolean(options.startAfter),
    startAtOperationTime:
      options.startAtOperationTime != null ? options.startAtOperationTime : undefined,
    fullDocument: options.fullDocument != null ? options.fullDocument : undefined,
    allChangesForCluster: changeStream.type === CHANGE_DOMAIN_TYPES.CLUSTER
  };
  const cursorOptions = {
    ...changeStreamStageOptions,
    batchSize: options.batchSize != null ? options.batchSize : undefined,
    maxAwaitTimeMS: options.maxAwaitTimeMS != null ? options.maxAwaitTimeMS : undefined,
    collation: options.collation != null ? options.collation : undefined,
    readPreference: options.readPreference != null ? options.readPreference : undefined
  };
```

I was thinking more very bespoke, and delete the list of options at the top of this file, doing things this way is verbose but it catches more errors. When I did this I found we have a small bug in the `get resumeOptions` logic

nbbeeken (changes requested):

src/change_stream.ts
baileympearson: Done!

src/change_stream.ts
baileympearson: Per our conversation on slack, I re-added applyKnownOptions.

test/integration/change-streams/change_stream.test.js
nbbeeken: It shouldn't be the case, I tried the following:
```js
expect({a: {}}).to.have.nested.property('a.b', undefined); // throws
expect({a: {b: undefined}}).to.have.nested.property('a.b', undefined); // passes
```

so actually I think two of your tests need to be changed to say "expect to NOT have nested property" distinct from a defined property that is undefined

src/change_stream.ts
```diff
@@ -568,25 +576,37 @@ function setIsIterator<TSchema>(changeStream: ChangeStream<TSchema>): void {
   }
   changeStream[kMode] = 'iterator';
 }
+
+function applyKnownOptions(source: Document, options: ReadonlyArray<string>) {
```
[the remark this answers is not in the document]
baileympearson: Not at all, done!

nbbeeken (dismissed):
LGTM!

test/integration/change-streams/change_stream.test.js
```diff
@@ -187,6 +187,77 @@ describe('Change Streams', function () {
...
+
+    afterEach(async function () {
+      await client.close();
+      client = undefined;
+      db = undefined;
+      collection = undefined;
+    });
+
+    context('fullDocument', () => {
+      it('sets fullDocument to `undefined` if no value is passed', function () {
```
dariakp, line 207: the test case says it sets fullDocument to `undefined`, but the assertion is for the property not being set at all on the object. Should it say "does not set the fullDocument property if no value is passed"?

test/integration/change-streams/change_stream.test.js
```diff
@@ -187,6 +187,77 @@ describe('Change Streams', function () {
...
+        const changeStream = client.watch();
+
+        expect(changeStream).not.to.have.nested.property(
+          'cursor.pipeline[0].$changeStream.fullDocument'
+        );
+      });
+
+      it('assigns `fullDocument` to the correct value if it is passed as an option', function () {
```
dariakp, line 215: Should there be another test case that uses a random string and/or "default" so that if we add code to start validating the options we'll get it flagged as a breaking change?

test/integration/change-streams/change_stream.test.js
```diff
@@ -187,6 +187,77 @@ describe('Change Streams', function () {
...
+      it('assigns `allChangesForCluster` to `true` if the ChangeStream.type is Cluster', function () {
+        const changeStream = client.watch();
+
+        expect(changeStream).to.have.nested.property(
+          'cursor.pipeline[0].$changeStream.allChangesForCluster',
+          true
+        );
+      });
+
+      it('does not assigns `allChangesForCluster` if the ChangeStream.type is Db', function () {
```
dariakp, line 235:
```suggestion
      it('does not assign `allChangesForCluster` if the ChangeStream.type is Db', function () {
```

test/integration/change-streams/change_stream.test.js
```diff
@@ -187,6 +187,77 @@ describe('Change Streams', function () {
...
+        const changeStream = db.watch();
+
+        expect(changeStream).not.to.have.nested.property(
+          'cursor.pipeline[0].$changeStream.allChangesForCluster'
+        );
+      });
+
+      it('does not assign `allChangesForCluster` if the ChangeStream.type is Db', function () {
```
dariakp, line 243:
```suggestion
      it('does not assign `allChangesForCluster` if the ChangeStream.type is Collection', function () {
```

dariakp (changes requested):

test/types/change_stream.test-d.ts
```diff
@@ -0,0 +1,9 @@
+import { expectError } from 'tsd';
+
+import type { ChangeStreamOptions } from '../../src';
+
+declare const changeStreamOptions: ChangeStreamOptions;
+
+// The change stream spec says that we cannot throw an error for invalid values to `fullDocument`
+// for future compatability.  This means we must leave `fullDocument` as type string.
+expectError<'updateLookup' | undefined>(changeStreamOptions.fullDocument);
```
dariakp, line 9: would an `expectType<string>(changeStreamOptions.fullDocument)` be better here? my concern is that the `expectError` would still pass if we restricted `fullDocument` as long as we didn't restrict it to explicitly `'updateLookup' | undefined`, if I understand the tsd assertions correctly

baileympearson: Yeah, you're right.  We should instead test for `string` exactly

dariakp (approved):

PR #3170, opened 2022-03-11
fix: ensure getTopology doens't throw

robertsLando:
#### What is changing?

See https://www.mongodb.com/community/forums/t/nodejs-driver-v4-4-1-doesnt-handle-errors-correctly/151675

##### Is there new documentation needed for these changes?

No

#### What is the motivation for this change?

Fixes https://www.mongodb.com/community/forums/t/nodejs-driver-v4-4-1-doesnt-handle-errors-correctly/151675

### Double check the following

- [ ] Ran `npm run check:lint` script
- [ ] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [ ] PR title follows the correct format: `<type>(NODE-xxxx)<!>: <description>`
- [ ] Changes are covered by tests
- [ ] New TODOs have a related JIRA ticket

PR #3172, opened 2022-03-15
fix(NODE-4074): ensure getTopology doesn't throw synchronously

baileympearson:
### Description

This PR is a re-do of the work done by Daniel Lando (original PR: https://github.com/mongodb/node-mongodb-native/pull/3170).  I messed up the rebase and accidentally removed all changes.

#### What is changing?

All uses of `getTopology` that are used in an asynchronous manner now catch any errors thrown and pass the error along to the appropriate callback.

- All calls to `getTopology` that were used in conjunction with `executeOperation` were moved into `executeOperation`
- Tests were added to ensure we're handling the errors appropriately with both callbacks and promises
- Two other uses of `getTopology` (`indexInformation` and `AddUserOperation.execute`) were calling getTopology in an asynchronous context, so these two uses now handle the potential error thrown from `getTopology`.
- `getTopology` was enhanced to take other topology providers that are used throughout the driver

Nope! 

### DIFF
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -65,29 +65,25 @@ export interface ExecutionResult {
  */
 export function executeOperation<
   T extends AbstractOperation<TResult>,
-  TResult = ResultTypeFromOperation<T>,
-  TSchema = Document
->(topology: Topology | TopologyProvider<TSchema>, operation: T): Promise<TResult>;
+  TResult = ResultTypeFromOperation<T>
+>(topologyProvider: Topology | TopologyProvider, operation: T): Promise<TResult>;
 export function executeOperation<
   T extends AbstractOperation<TResult>,
-  TResult = ResultTypeFromOperation<T>,
-  TSchema = Document
->(topology: Topology | TopologyProvider<TSchema>, operation: T, callback: Callback<TResult>): void;
+  TResult = ResultTypeFromOperation<T>
+>(topologyProvider: Topology | TopologyProvider, operation: T, callback: Callback<TResult>): void;
 export function executeOperation<
   T extends AbstractOperation<TResult>,
-  TResult = ResultTypeFromOperation<T>,
-  TSchema = Document
+  TResult = ResultTypeFromOperation<T>
 >(
-  topology: Topology | TopologyProvider<TSchema>,
+  topologyProvider: Topology | TopologyProvider,
   operation: T,
   callback?: Callback<TResult>
 ): Promise<TResult> | void;
 export function executeOperation<
   T extends AbstractOperation<TResult>,
-  TResult = ResultTypeFromOperation<T>,
-  TSchema = Document
+  TResult = ResultTypeFromOperation<T>
 >(
-  topology: Topology | TopologyProvider<TSchema>,
+  topologyProvider: Topology | TopologyProvider,
   operation: T,
   callback?: Callback<TResult>
 ): Promise<TResult> | void {
@@ -97,8 +93,10 @@ export function executeOperation<
   }
 
   return maybePromise(callback, callback => {
+    let topology: Topology | TopologyProvider;
     try {
-      topology = topology instanceof Topology ? topology : getTopology(topology);
+      topology =
+        topologyProvider instanceof Topology ? topologyProvider : getTopology(topologyProvider);
     } catch (error) {
       return callback(error);
     }
@@ -106,7 +104,7 @@ export function executeOperation<
       return topology.selectServer(ReadPreference.primaryPreferred, {}, err => {
         if (err) return callback(err);
 
-        executeOperation<T, TResult, TSchema>(topology, operation, callback);
+        executeOperation<T, TResult>(topology, operation, callback);
       });
     }
 
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -19,9 +19,15 @@ import {
   secondaryWritableServerSelector,
   ServerSelector
 } from '../sdam/server_selection';
-import type { Topology } from '../sdam/topology';
+import { Topology } from '../sdam/topology';
 import type { ClientSession } from '../sessions';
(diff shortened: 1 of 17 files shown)


test/integration/connection-monitoring-and-pooling/connection.test.js
```diff
@@ -452,6 +452,26 @@ describe('Connection', function () {
       })
     );
 
+    it('throws when attempting an operation if the client is not connected', function (done) {
+      const client = this.configuration.newClient();
+      const collection = client.db('shouldCorrectlyFailOnRetry').collection('test');
+      collection.insertOne({ a: 2 }, err => {
+        expect(err).to.match(/must be connected/);
```
nbbeeken, line 459: Can we check for instanceof `MongoNotConnectedError` class instead, error messages are subject to change.

test/integration/connection-monitoring-and-pooling/connection.test.js
```diff
@@ -452,6 +452,26 @@ describe('Connection', function () {
...
+        done();
+      });
+    });
+
+    it('throws when attempting an operation if the client is not connected (promises)', async function () {
+      const client = this.configuration.newClient();
+      const collection = client.db('shouldCorrectlyFailOnRetry').collection('test');
+
+      try {
+        await collection.insertOne({ a: 2 });
```
nbbeeken, line 469: We need an expect.fail() below the insertOne to make sure the error doesn't disappear on us.

An alternative:
```ts
const error = await collection.insertOne({a:2}).catch(error => error)
expect(error).to.be.instanceOf(MongoNotConnectedError)
```
You can keep it all a oneliner and it catches both cases here, if it succeeds the instanceOf will fail, if it throws we get our error returned to us.

test/integration/connection-monitoring-and-pooling/connection.test.js
```diff
@@ -463,13 +483,26 @@ describe('Connection', function () {
...
     );
+
+    it('should correctly fail on retry when client has been closed (promises)', async function () {
+      const client = await this.configuration.newClient().connect();
+      const collection = client.db('shouldCorrectlyFailOnRetry').collection('test');
+      await collection.insertOne({ a: 1 });
+      await client.close(true);
+
+      try {
+        await collection.insertOne({ a: 2 });
```
nbbeeken, line 502: Same issue from other thread, need expect.fail or alternative approach.

src/utils.ts
```diff
@@ -326,12 +326,17 @@ export function decorateWithExplain(command: Document, explain: Explain): Docume
...
 }
 
+/**
+ * @internal
+ */
+export type TopologyProvider<T> = MongoClient | Db | Collection<T>;
+
 /**
  * A helper function to get the topology from a given provider. Throws
  * if the topology cannot be found.
```
nbbeeken, line 336: Let's add an `@throws` to the tsdoc here, might help the next person catch themselves if more usage of the helper is needed in the future.

nbbeeken (changes requested):

nbbeeken (approved):

nbbeeken (changes requested):
Just noticed 😅  We've got some lint errors to sort out

nbbeeken (approved):

robertsLando:
Any news on this?

src/operations/execute_operation.ts
```diff
@@ -60,30 +66,44 @@ export interface ExecutionResult {
...
+  callback?: Callback<TResult>
+): Promise<TResult> | void {
   if (!(operation instanceof AbstractOperation)) {
     // TODO(NODE-3483): Extend MongoRuntimeError
     throw new MongoRuntimeError('This method requires a valid operation instance');
   }
 
   return maybePromise(callback, callback => {
+    let topology: Topology;
+    try {
```
dariakp, line 97: Let's run some perf tests as a sanity check to make sure we aren't incurring a perf hit with this added logic

dariakp (approved):

robertsLando:
Seems that there are still some collections methods that could throw:

https://github.com/mongodb/node-mongodb-native/blob/main/src/collection.ts#L762

https://github.com/mongodb/node-mongodb-native/blob/main/src/collection.ts#L1411

baileympearson:
Hey @robertsLando!  These methods can still throw.  However, they're actually synchronous methods so throwing is the expected behavior (any method that returns a cursor is synchronous because cursors are lazy and don't fetch data until they're consumed).

The other methods in this PR needed to be fixed because they took callbacks.  Previously, if we didn't catch the error and the user passed a callback, we could crash the user's process because we weren't propagating the error up the callback chain.

robertsLando:
Yeah I know that (I'm the user who opened the issue and PR for this). I just saw the new release today and when I tested it I got this issue. While I understand what you mean I don't think this behaviour is correct as it's not working as before, in my opinion the error should be thrown when `next` is called (like it was before). Wrapping every find with a try catch is much less user friendly

Before:
```js
collection.find({}).next( (err, doc) => {
// err contains the error connection is not opened
})
```

Now:
```js
try {
collection.find({}).next( (err, doc) => {
 		// never reached
})
} catch(error) {
 // error contains the error connection is not opened

}
```

baileympearson:
Just to clarify - you're saying that the behavior changed in v4.5.0?  I've tried in all versions 4.3.0 -> 4.5.0 and in all cases I see the error being caught in the `catch` block, not in the `next` function.

```typescript
import { MongoClient } from "mongodb";

async function run() {
  const client = new MongoClient(< uri >);

  const db = client.db('nodeTest');
  const collection = db.collection('nodeTest');

  try {
    collection.find({}).next(( )=> {
      console.log('caught error inside next');
    });
  } catch (err) {
    console.log('error caught in catch')
  }
}

run();
```

robertsLando:
> Just to clarify - you're saying that the behavior changed in v4.5.0

Nope this has changed in 4+ version but I didn't see it reported as breaking change. Also as you can see in my example the code looks a bit less user friendly

PR #3175, opened 2022-03-16
fix(NODE-3810): fix async interval checks for streaming protocol cases

durran:
### Description

Fix async interval executing during streaming protocol.

#### What is changing?

Moves the async executor check for a negative time until next call to the end of the wake function, in order for a negative time until next wake to return immediately without executing, which is the case of the sdam monitor check while using the streaming protocol. (In that case the monitor should never initiate the check after the first call since the server is "streaming" responses)

No

NODE-3810


src/utils.ts
```diff
@@ -862,34 +862,34 @@ export function makeInterruptibleAsyncInterval(
...
     if (timeUntilNextCall > minInterval) {
       reschedule(minInterval);
-      cannotBeExpedited = true;
+    }
+
+    // This is possible in virtualized environments like AWS Lambda where our
+    // clock is unreliable. In these cases the timer is "running" but never
+    // actually completes, so we want to execute immediately and then attempt
+    // to reschedule.
+    if (timeUntilNextCall < 0) {
```
dariakp, line 891: I'm curious that no tests failed/needed to be updated as a result of this change - were none of the existing tests testing this behavior?

src/utils.ts
```diff
@@ -914,7 +914,7 @@ export function makeInterruptibleAsyncInterval(
   }
 
   function executeAndReschedule() {
-    cannotBeExpedited = false;
+    lastWakeTime = 0;
```
dariakp, line 917: Are we sure that this is the right reset behavior? Why do we need to reset the lastWakeTime at all? It's set to currentTime by the wake function, which is what calls the execute function here, so it should be correct. The previous code was manually setting a flag to allow or forbid expediting the wake, so it needed to be reset. What is the reasoning for resetting it here?

Actually, looking over the code some more, it seems that a cleaner approach might be to leave the `cannotBeExpedited` logic as is, but add a line at the top of `wake` to override the value to `true` if the lastWakeTime is negative, i.e. `cannotBeExpedited = cannotBeExpedited ||  lastWakeTime < 0`. I think it would make it easier to reason about what exactly we are resetting and why.

src/utils.ts
```diff
@@ -862,34 +862,34 @@ export function makeInterruptibleAsyncInterval(
 
   function wake() {
     const currentTime = clock();
-    const nextScheduledCallTime = lastCallTime + interval;
-    const timeUntilNextCall = nextScheduledCallTime - currentTime;
+    const timeSinceLastWake = currentTime - lastWakeTime;
```
dariakp, line 865: If I am reading the code correctly, the very first time this function is called, lastWakeTime is going to be `undefined`, so this timeSinceLastWake is going to be `NaN`. It may be best to default it to 0 in L854.

src/utils.ts
```diff
@@ -862,34 +862,34 @@ export function makeInterruptibleAsyncInterval(
...
+    const timeSinceLastCall = currentTime - lastCallTime;
+    const timeUntilNextCall = interval - timeSinceLastCall;
```
dariakp, line 867: why did the logic for `timeUntilNextCall` need a refactor? it's equivalent to the logic that was already in place, just using `nextScheduledCallTime` instead of `timeSinceLastCall` as far as I can tell, i.e., the previous logic had `(lastCallTime + interval) - currentTime` and this logic has `interval - (currentTime - lastCallTime) = interval - currentTime + lastCallTime`

test/unit/utils.test.js
```diff
@@ -94,6 +94,28 @@ describe('driver utils', function () {
...
+      context('when the time since last wake is negative', () => {
+        it('does not execute and returns immediately', () => {
+          executor = makeInterruptibleAsyncInterval(fnSpy, {
+            minInterval: 10,
+            interval: 30,
+            clock: () => {
+              return -50;
+            }
+          });
+          // stop the executor
```
dariakp, line 106: can we explain more about why it has to be stopped first? I think this step might not be necessary if we initialize the variable instead of leaving it undefined (see my comment on the other file regarding NaN)

dariakp (changes requested):

src/utils.ts
durran: Updated to default to zero.

src/utils.ts
durran: I agree the `cannotBeExpedited` logic was easier to reason with but I am hesitant to bring it back for 2 reasons: 1) I could not replicate the exact same behaviour in my lambda testing that the customer was seeing and 2) I believe that `cannotBeExpedited` logic was potentially wrong. In the previous implementation it looked possible to continuously "wake" with a negative `timeUntilNextCall` (some lambda situations) that would never debounce the calls. This revert to the old time based logic would at least ensure it could not be woken continuously sooner than the minimum interval.

src/utils.ts
durran: There was no test covering the negative clock case which was now added. Moving the block did not fail any existing tests.

PR #3180, opened 2022-03-25
fix(NODE-3810): delay timeout errors by one event loop tick

nbbeeken:
#### What is changing?

Delay timeout errors from a socket by one event loop tick.

In FAAS environments timers will expire between invocations but will only execute upon waking up. Node.js always runs timeouts first in the event loop, this means our monitoring socket will give us a timeout error before it reads the already pending hello response. By putting the timeout error logic inside another (small as possible) timeout callback it will enforce timeout errors to be handled on the next event loop iteration, giving the driver the opportunity to first read from the socket. If it is able to read, it will clear the timeout preventing the error from being raised to our monitoring logic. If nothing is ready pending to be read from the socket, the (small as possible) timeout will fire making the timeout error occur as usual.


durran (approved):
LGTM if we're ok with my one question. Nice work.

test/unit/cmap/connection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    // @ts-expect-error: This subset of options is all that is needed
+    connect({ connectionType: Connection, hostAddress: server.hostAddress() }, (err, conn) => {
+      expect(err).to.not.exist;
+      expect(conn).to.exist;
+
+      conn.command(ns('$admin.cmd'), { ping: 1 }, { socketTimeoutMS: 50 }, (err, result) => {
+        expect(err).to.be.instanceOf(MongoNetworkTimeoutError);
+        expect(result).to.not.exist;
+
+        expect(conn).property('stream').property('destroyed').to.be.true;
```
dariakp, line 62:
```suggestion
        expect(conn).property('stream').property('destroyed', true);
```

test/unit/cmap/connection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+        done();
+      });
+    });
+  });
+
+  it('should throw a network error with kBeforeHandshake set to false on timeout after hand shake', function (done) {
```
dariakp, line 69:
```suggestion
  it('should throw a network error with kBeforeHandshake set to false on timeout after handshake', function (done) {
```

test/unit/cmap/connection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+        const beforeHandshakeSymbol = getSymbolFrom(err, 'beforeHandshake', false);
+        expect(beforeHandshakeSymbol).to.be.a('symbol');
+        expect(err).to.have.property(beforeHandshakeSymbol, false);
+
+        done();
+      });
+    });
+  });
+
+  it('should throw a network error with kBeforeHandshake set to true on timeout before hand shake', function (done) {
```
dariakp, line 98:
```suggestion
  it('should throw a network error with kBeforeHandshake set to true on timeout before handshake', function (done) {
```

test/unit/cmap/connection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      clock.restore();
+    });
+
+    it('should delay timeout errors by one tick', async () => {
+      expect(connection).to.have.property(kDelayedTimeoutId, null);
+
+      driverSocket.emit('timeout');
+      expect(connection.onTimeout).to.have.been.calledOnce;
+      expect(connection).to.have.property(kDelayedTimeoutId).that.is.instanceOf(NodeJSTimeoutClass);
+
```
dariakp, line 169: can we add an expect connection `closed` to be `false` and expect `driverSocket.destroy` to not have been called to explicitly verify that it doesn't fire sooner?

test/unit/cmap/connection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      clock.tick(1);
+
+      expect(driverSocket.destroy).to.have.been.calledOnce;
+      expect(connection).to.have.property('closed', true);
+      // timeout callback should clear it's own reference
```
dariakp, line 174: => its
but also, see my question in the code regarding why

test/unit/cmap/connection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      // onMessage ignores unknown 'responseTo' value
+      messageStream.emit('message', { responseTo: null });
+
+      // New message before clock ticks 1 will clear the timeout
+      expect(connection).to.have.property(kDelayedTimeoutId, null);
+
+      // ticking the clock should do nothing, there is no timeout anymore
+      clock.tick(1);
+
+      expect(driverSocket.destroy).to.not.have.been.calledOnce;
```
dariakp, line 195: the correct assertion here is to `not.have.been.called` (at all), because not being called once doesn't rule out have been called more than once

dariakp (changes requested):

src/cmap/connection.ts
```diff
@@ -309,40 +316,139 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
-  destroy(options: DestroyOptions, callback: Callback): void;
-  destroy(options?: DestroyOptions | Callback, callback?: Callback): void {
+  onTimeout() {
+    if (this.closed) {
+      return;
+    }
+
+    this[kDelayedTimeoutId] = setTimeout(() => {
+      const delayedTimeoutId = this[kDelayedTimeoutId];
+      if (delayedTimeoutId != null) {
```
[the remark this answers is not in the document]
nbbeeken: I think this was contrived from overly defensive programming, it should not be necessary to clear here. Or check for nullish. If one (somehow, since it should not be possible) already exists, either it expired, or we want a new one anyway for the new timeout error that caused us to reach this point.

test/unit/cmap/connection.test.ts
```diff
@@ -108,12 +115,95 @@ describe('Connection - unit/cmap', function () {
...
+      const messageStreamSymbol = getSymbolFrom(connection, 'messageStream');
+      kDelayedTimeoutId = getSymbolFrom(connection, 'delayedTimeoutId');
+      messageStream = connection[messageStreamSymbol];
+    });
+
+    afterEach(() => {
+      clock.restore();
+    });
+
+    it('should delay timeout errors by one tick', async () => {
```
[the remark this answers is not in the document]
nbbeeken: The test is `should destroy streams which time out` btw the mock server just doesn't return a reply to the driver

test/tools/runner/hooks/configuration.js
```diff
@@ -66,15 +66,14 @@ async function initializeFilters(client) {
 }
 
 const testSkipBeforeEachHook = async function () {
-  // `metadata` always exists, `requires` is optional
-  const requires = this.currentTest.metadata.requires;
+  const metadata = this.currentTest.metadata;
```
baileympearson, line 69: Is there a reason for this change or it is purely cosmetic?

src/cmap/connection.ts
```diff
@@ -309,40 +316,133 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+          callback(new MongoWriteConcernError(document.writeConcernError, document));
+          return;
+        }
+
+        if (document.ok === 0 || document.$err || document.errmsg || document.code) {
+          callback(new MongoServerError(document));
+          return;
+        }
+      } else {
+        // Pre 3.2 support
```
baileympearson, line 434: It's okay if you don't want to - but any desire to remove this legacy code?

test/tools/runner/hooks/configuration.js
nbbeeken: When I first started trying to test this I had a more simulated approach that would pass sometimes so I tried using mocha's retries feature. It looks like the retried test doesn't get the metadata attached, so this crashed. I can revert, but we can't use retries with a metadata test currently.

src/cmap/connection.ts
nbbeeken: I'm happy to do so in another PR, I normally would but I'm thinking that the refactor has already increased the surface area enough for this change.

baileympearson (changes requested):

src/cmap/connection.ts
baileympearson: Understandable, either way (or not addressing it now at all) is okay.

test/tools/runner/hooks/configuration.js
baileympearson: Ah makes sense.  Sounds good

dariakp (commented):
agree with Bailey's suggestion but otherwise LGTM

PR #3182, opened 2022-03-28
test(NODE-4157): SDAM unit test typescript and improvements

nbbeeken:
#### What is changing?

This is a breakout of the changes from #3174

Link to commit that shows diff: [88e5473](https://github.com/mongodb/node-mongodb-native/pull/3182/commits/88e5473f8d57f1ea94c2dca1ddf947ffdfa0a2b5)

Brings in partial spec sync, description update in: `secondary_ignore_ok_0.json` (https://github.com/mongodb/specifications/commit/5bd06a81d97850a1365d73f08bbb7dea562e72b9)

During debugging the new tests these fixups were natural improvements, using chai property assertions, typescript, easier tracking of the current phase to see the change our SDAM machine will perform to the topology.


test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -157,129 +167,109 @@ function cloneMap(map) {
...
-                withConnectionStub.restore();
-
-                phaseCb();
-              });
-            },
-            err => {
-              expect(err).to.not.exist;
-              phaseDone();
-            }
+      topology.removeAllListeners('error');
```
dariakp, line 237: is this reset logic potentially problematic if some of the expects above fail? I'm not sure how the phases are supposed to work, but feels like there should be some kind of before/after cleanup logic, maybe a try/finally

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -157,129 +167,109 @@ function cloneMap(map) {
...
+      events = [];
+    } else if (phase.applicationErrors) {
+      for (const appError of phase.applicationErrors) {
+        const withConnectionStub = sinon
+          .stub(ConnectionPool.prototype, 'withConnection')
+          .callsFake(withConnectionStubImpl(appError));
+
+        const server = topology.s.servers.get(appError.address);
+        const res = promisify(server.command.bind(server))(ns('admin.$cmd'), { ping: 1 }, {});
+        withConnectionStub.restore();
```
dariakp, line 247: feels like the stub restore should happen later in the cleanup, or at least after we catch the thrownError in L249

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -157,129 +167,109 @@ function cloneMap(map) {
...
           );
-        } else {
-          phaseDone();
-        }
-      },
-      err => {
-        expect(err).to.not.exist;
-        done();
+        };
+        expect(thrownError).to.satisfy(isApplicationError);
```
dariakp, line 258: can we add a custom message here for the expect fail case?

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -361,25 +356,29 @@ function assertOutcomeExpectations(topology, events, outcome) {
...
-    expect(description).to.include.keys(translatedKey);
+    expect(description).to.include.keys(key);
 
     if (outcomeValue == null) {
-      expect(description[translatedKey]).to.not.exist;
+      expect(description[key]).to.be.undefined;
     } else {
-      expect(description).property(translatedKey).to.eql(outcomeValue, `(key="${translatedKey}")`);
+      if (typeof outcomeValue === 'object') {
+        expect(description).to.have.property(key).that.deep.equals(outcomeValue);
```
dariakp, line 378: you can use `deep.property(key, outcomeValue)` here for deep comparison

dariakp (changes requested):
some questions and suggestions here

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: I am able to delete it and nothing changes I think because as long as the test completes the topology is garbage collected so there's no need to clean up listeners.

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: Changed!

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: Done!

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: Moved it below awaiting the error, good call

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -232,7 +234,6 @@ async function executeSDAMTest(testData) {
         }
       }
 
-      topology.removeAllListeners('error');
```
dariakp: I actually don't think it's correct to remove this - any test that has more than one phase would get multiple handlers assigned without that error listener, so your array would get double the data, but you couldn't catch it because we only assert on the event types and the message, not on the count of the registered events. So maybe we need to add an assertion on the length of `errorEvents`

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -253,7 +256,10 @@ async function executeSDAMTest(testData) {
             error instanceof MongoServerError
           );
         };
-        expect(thrownError).to.satisfy(isApplicationError);
+        expect(
+          thrownError,
+          'expected the error thrown to be one of MongoNetworkError, MongoNetworkTimeoutError or MongoServerError. Referred to in the spec as an "Application Error"'
```
dariakp, line 261:
```suggestion
          'expected the error thrown to be one of MongoNetworkError, MongoNetworkTimeoutError or MongoServerError (referred to in the spec as an "Application Error")'
```

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -374,15 +374,6 @@ function assertOutcomeExpectations(topology, events, outcome) {
...
-    if (outcomeValue == null) {
-      expect(description[key]).to.be.undefined;
-    } else {
-      if (typeof outcomeValue === 'object') {
-        expect(description).to.have.property(key).that.deep.equals(outcomeValue);
-      } else {
-        expect(description).to.have.property(key, outcomeValue);
-      }
-    }
+    expect(description).to.have.deep.property(key, outcomeValue == null ? undefined : outcomeValue);
```
dariakp, line 377: Just to double check, if the outcomeValue is null, we still expect the property to exist on the description, just with a value of "undefined"?

dariakp (changes requested):

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: I checked the length and moved the listener outside the loop to make sure it's not added more than once

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: agreed, pulled this into the other fixes.

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: I've updated the code here, the remaining keys being checked where all always defined except for logicalSessionTimeoutMinutes, which is always defined but is set explicitly to undefined.

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
dariakp: Two concerns:
1) The check you added doesn't guarantee that we don't get more events than we should, so it's guarding against no events being registered but not guarding against events getting doubled up, for example;

2) before, we were only attaching the handler and collecting events per phase, which makes sense if we're trying to isolate what is happening each time; *now* we have one listener and one giant pile of errorEvents the entirety of which gets checked for every phase regardless of when the particular event was caught, and there is no reset, so every phase adds more events into the pile and then checks all the existing events and the new events - that seems wrong

dariakp: I think it's fine to leave the error handler attachment at the top, but the collected errors definitely need to be cleared at the end of each phase, and I would still like to find a way to assert on the number of those expected error events for each phase

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -0,0 +1,513 @@
...
+  const knownTopologyDescriptionOutcomeKeys: ReadonlyArray<keyof TopologyDescriptionOutcome> = [
+    'topologyType',
+    'setName',
+    'servers',
+    'logicalSessionTimeoutMinutes',
+    'maxSetVersion',
+    'maxElectionId',
+    'compatible'
+  ] as const;
+  expect(outcome).to.have.any.keys(knownTopologyDescriptionOutcomeKeys);
```
dariakp, line 156: if you're trying to assert that the outcome keys are a subset of the known keys, you want to do:
```js
for (const key of Object.keys(outcome)) {
  expect(knownTopologyDescriptionOutcomeKeys).to.include(key);
}
```

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -0,0 +1,513 @@
...
+  }
+
+  if (event.newDescription != null && event.newDescription.servers != null) {
+    result.newDescription.servers = cloneMap(event.newDescription.servers);
+  }
+
+  return result;
+}
+
+const SDAM_EVENTS = [
```
dariakp, line 300: is this just the union of `HEARTBEAT_EVENTS` and `TOPOLOGY_EVENTS` excluding ERROR, TIMEOUT, and CLOSE? would it be worth defining these in terms of existing constants instead of manually entered strings?

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -0,0 +1,513 @@
...
+      delete expectedServer.pool;
+    }
+
+    const normalizedExpectedServer = normalizeServerDescription(expectedServer);
+    const omittedFields = findOmittedFields(normalizedExpectedServer);
+    const actualServer = actualServers.get(serverName);
+    expect(actualServer).to.matchMongoSpec(normalizedExpectedServer);
+
+    if (omittedFields.length !== 0) {
+      expect(actualServer).to.not.have.all.keys(omittedFields);
```
dariakp, line 476: this should be `any` not `all`, because using an "all" in the negation would pass if some omittedFields were still present. I strongly recommend referring to the `keys` [chai docs](https://www.chaijs.com/api/bdd/#method_keys) whenever you use `any` or `all` because it's very easy to get wrong

dariakp (changes requested):

test/spec/server-discovery-and-monitoring/rs/secondary_ignore_ok_0.json
```diff
@@ -1,5 +1,5 @@
 {
-  "description": "New primary",
+  "description": "Secondary ignored when ok is zero",
```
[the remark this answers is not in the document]
nbbeeken: It's the same commit that the setVersion/electionId landed in, is there something we should do to further connect it?

https://github.com/mongodb/specifications/commit/5bd06a81d97850a1365d73f08bbb7dea562e72b9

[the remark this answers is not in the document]
dariakp: it just might be confusing that this is a partial spec sync; like when we look at the main PR, we'd likely compare files, and then someone might have a question about why these ones are missing; I think it's ok to include them in this PR, but you should update *this* PR's description to say that it includes a partial sync of that commit, and then update the *other* PR description to say that this specific test was synced in this linked PR, so that other reviewers won't have to go looking for them

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: Yea this assertion was here before but I missed it, looks like we have some bugs if this assertion is necessary. 
For example "Non-stale network error" in `test/spec/server-discovery-and-monitoring/errors/non-stale-network-error.json` expects topologyVersion to be null, because the server is marked unknown, I'm adding this to NODE-4159 I think it directly relates.

dariakp: yay for catching errors by using correct assertions, lol
maybe instead of leaving this assertion in the weird state, we could do a `for .. of` on the `omittedFields` and expect `.to.not.have.property` with exceptions for the known incompatibilities?

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
nbbeeken: done!

test/spec/server-discovery-and-monitoring/rs/secondary_ignore_ok_0.json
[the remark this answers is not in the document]
nbbeeken: done!

baileympearson (dismissed):
LGTM!

dariakp (approved):

test/unit/assorted/server_discovery_and_monitoring.spec.test.ts
dariakp: Leaving this open for context for other reviewers, can be resolved if everyone approves

dariakp (approved):

durran (approved):

PR #3183, opened 2022-03-29
fix(NODE-4129): constrain `watch` type parameter to extend `ChangeStream` type parameter

eps1lon:
#### What is changing?
Constrain `watch` type parameter to same constraint as `ChangeStream`

Not

Starting with https://github.com/microsoft/TypeScript/pull/48366 we'll no longer be able to assign an unconstraint type parameter to a constraint one.

Constraining the type is required to pass https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59560

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -1422,7 +1422,7 @@ export class Collection<TSchema extends Document = Document> {
    * @param pipeline - An array of {@link https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/|aggregation pipeline stages} through which to pass change stream documents. This allows for filtering (using $match) and manipulating the change stream documents.
    * @param options - Optional settings for the command
    */
-  watch<TLocal = TSchema>(
+  watch<TLocal extends Document = TSchema>(
     pipeline: Document[] = [],
     options: ChangeStreamOptions = {}
   ): ChangeStream<TLocal> {
--- a/src/db.ts
+++ b/src/db.ts
@@ -722,7 +722,7 @@ export class Db {
    * @param pipeline - An array of {@link https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/|aggregation pipeline stages} through which to pass change stream documents. This allows for filtering (using $match) and manipulating the change stream documents.
    * @param options - Optional settings for the command
    */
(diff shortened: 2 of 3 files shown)


nbbeeken:
Hi @eps1lon thanks for your help! Does only this change clear up the errors you are seeing? When I try compiling the driver with `typescript@4.7.0-dev.20220330` I see an additional 33 errors mostly related to the same thing, just in other APIs. 

Oh, or perhaps this is the only API usage you have so there are no other errors on your end. Can you confirm?

eps1lon:
I was only seeing 3 errors: https://github.com/DefinitelyTyped/DefinitelyTyped/runs/5743860548?check_suite_focus=true#step:6:170

From what I can tell it's the same line that produces all 3 of those. But there may be more errors related to this issue.

nbbeeken:
Ah yea, our `mongodb.d.ts` file only has our public API and as I look closer I think the errors are all coming from internal types. 

This change looks good to me but were you able to see a passing run with this change? Just confirming because I want to make sure this PR will unblock you and we can follow up with the errors I'm seeing.

dariakp (dismissed):

eps1lon:
> This change looks good to me but were you able to see a passing run with this change?

Hard to judge since it's unclear what command produces the types. Just guessin from the naming that this fixes it.

nbbeeken:
I ran the tests locally, it looks like we need updates to the watch functions on Collection and MongoClient as well. I've updated those. 

Just in case you missed it I'm still seeing an error from another package:
```txt
 ../styled-components/index.d.ts(119,68): error TS2344: Type 'P' does not satisfy the constraint '{ theme?: T | undefined; }'.
```

eps1lon:
> Just in case you missed it I'm still seeing an error from another package:

I can deal with that later. Thanks for pushing this further. The additional changes look good to me!

baileympearson (approved):

nbbeeken (approved):

eps1lon:
@nbbeeken Thanks for the quick assistance! Do know roughly when this'll be released?

nbbeeken:
@eps1lon Sorry about the delay, the fix is available now in v4.5.0!

PR #3190, opened 2022-04-06
test: failing aws test should not check for region

baileympearson:
#### What is changing?

The aws clsfe test that used to check that an error was thrown if the region was invalid now only checks that a generic error is thrown and doesn't search for the region in the error message.

This is a drivers-wide bug.  It seems like this will be the fix (see https://github.com/mongodb/specifications/pull/1171) and we can clean up the todo once the Node ticket is opened.


nbbeeken (approved):
LGTM

PR #3191, opened 2022-04-06
feat(NODE-4059): ChangeStreamDocument not fully typed to specification

nbbeeken:
#### What is changing?

ChangeStreamDocument had some errors and friction in type info. I've restructured it to be a descriminated union of the possible change stream events which lends itself well to TS.

This does introduce compilation changes since properties that used to be optional now do not exist depending on the change stream event type. We could consider mitigating this by bringing back the properties but typing them as something that indicates they shouldn't have been used with a particular event (for example delete has no fullDocument field)

Type hinting on change stream documents helps sort through the variety of events that can occur.


src/change_stream.ts
```diff
@@ -84,16 +85,16 @@ export interface ResumeOptions {
 }
 
 /**
- * Represents the logical starting point for a new or resuming {@link https://docs.mongodb.com/manual/changeStreams/#std-label-change-stream-resume| Change Stream} on the server.
+ * Represents the logical starting point for a new or resuming on the server.
```
baileympearson, line 88:
```suggestion
 * Represents the logical starting point for a new ChangeStream or resuming a ChangeStream on the server.
```

src/change_stream.ts
```diff
@@ -107,94 +108,220 @@ export interface PipeOptions {
...
+   * @see https://docs.mongodb.com/manual/reference/command/aggregate
+   */
   batchSize?: number;
 }
 
 /** @public */
-export interface ChangeStreamDocument<TSchema extends Document = Document> {
+export interface ChangeStreamFullNameSpace {
+  /** The namespace */
+  ns: { db: string; coll: string };
```
baileympearson, line 141: I notice further down in the `rename` interface, you reuse `{ db: string; coll: string }`.  Ultimately, `{ db: string; coll: string }` is the namespace.  Could we create the type as
```typescript
interface ChangeStreamFullNamespace { db: string; coll: string }
```
so we can reuse it where needed?  Then we can add `ns: ChangeStreamFullNamespace` to the `ChangeStreamDocumentCommon` object.

src/change_stream.ts
```diff
@@ -107,94 +108,220 @@ export interface PipeOptions {
...
   /**
-   * Only present for ops of type ‘insert’, ‘update’, ‘replace’, and
-   * ‘delete’.
-   *
-   * For unsharded collections this contains a single field, _id, with the
-   * value of the _id of the document updated.  For sharded collections,
-   * this will contain all the components of the shard key in order,
-   * followed by the _id if the _id isn’t part of the shard key.
+   * The transaction number.
+   * Only present if the operation is part of a multi-document transaction.
```
baileympearson, line 172: Can we note here that the type of transaction number depends on the `promoteLongs` flag?

src/change_stream.ts
```diff
@@ -107,94 +108,220 @@ export interface PipeOptions {
...
    */
-  updateDescription?: UpdateDescription<TSchema>;
+  lsid?: ServerSessionId;
+}
 
+/**
+ * @public
+ * @see https://www.mongodb.com/docs/manual/reference/change-events/#insert-event
+ */
+export interface InsertChangeStreamDocument<TSchema extends Document = Document>
```
baileympearson, line 187: nit: I think `ChangeStream<Command>Result` might make more sense than `<Command>ChangeStreamDocument` as a general format for these interfaces but I don't feel too strongly.  Thoughts?

test/tools/unified-spec-runner/entities.ts
```diff
@@ -395,10 +403,20 @@ export class EntitiesMap<E = Entity> extends Map<string, E> {
...
       } else if ('collection' in entity) {
         const db = map.getEntity('db', entity.collection.database);
-        const collection = db.collection(
-          entity.collection.collectionName,
-          patchCollectionOptions(entity.collection.collectionOptions)
-        );
+        const collection = await db
+          .createCollection(
+            entity.collection.collectionName,
+            patchCollectionOptions(entity.collection.collectionOptions)
```
baileympearson, line 409: `patchCollectionOptions` is currently a no-op that essentially just casts the type of the options.  Is there a reason we need this utility?  I think we could just cast it inline.

test/tools/unified-spec-runner/entities.ts
```diff
@@ -395,10 +403,20 @@ export class EntitiesMap<E = Entity> extends Map<string, E> {
         map.set(entity.database.id, db);
       } else if ('collection' in entity) {
         const db = map.getEntity('db', entity.collection.database);
-        const collection = db.collection(
-          entity.collection.collectionName,
-          patchCollectionOptions(entity.collection.collectionOptions)
-        );
+        const collection = await db
```
baileympearson, line 406: question: is this logic added to handle the scenario where a user has two entities with the same collection name in a single test suite?  If so, would it make more sense to flip the logic - list collections while querying for the collection, and then create it if it doesn't exist?

test/integration/enumerate_databases.test.ts
```diff
@@ -233,6 +233,6 @@ const testSuite = new UnifiedTestSuiteBuilder('listDatabases with comment option
   )
   .toJSON();
 
-describe('listDatabases w/ comment option', () => {
+describe('listDatabases with comment option', () => {
```
baileympearson, line 236: non-blocking: Feel free to use the new `toMocha` method you added on `enumerate_databases` and `enumerate_indexes` if you'd like, but not necessary for the review!

baileympearson (changes requested):
Some questions and suggestions

src/change_stream.ts
nbbeeken: I'm wondering if it's more friction that it's worth to include this in the type annotation and instead document it as a possiblity based on the option.

test/tools/unified-spec-runner/entities.ts
nbbeeken: Until this PR we haven't used the unified runner to run a drop db / rename collection. They both require the collection to exist in advance of the operations stage of the test. I could instead adjust the test to perform an insert to create the collection, but this seems like a reasonable adjustment to the runner.

test/tools/unified-spec-runner/entities.ts
nbbeeken: No reason, it's a no-op for now, because we're lucky to have the JSON match our API but we should actually validate that in that helper, I can make a ticket for that TODO

src/change_stream.ts
nbbeeken: done!

src/change_stream.ts
nbbeeken: Changed this, we can't add it to common because it's missing from a number of the events

src/change_stream.ts
nbbeeken: Renamed them all, I'm still using Document at the end because I don't want to change the name of the `ChangeStreamDocument` type and we should keep these consistent, but reordering the indicator makes sense.

test/integration/enumerate_databases.test.ts
nbbeeken: Done!

src/change_stream.ts
baileympearson: So include this as part of the documentation for the `promoteLongs` flag?  I'm okay with that

test/tools/unified-spec-runner/entities.ts
baileympearson: Ah I see, that makes sense.  Sounds good

test/tools/unified-spec-runner/entities.ts
baileympearson: Oh that makes sense.  I was mainly wondering if the logic would be easier to follow if we flipped it around to look for collections first.  But I don't feel super strongly about this, feel free to ignore this thought

```typescript
let collection = await db.listCollections({ collectionName: entitiy.collection.collectionName })
    .map(({ name }) => db.collection(name))
    .toArray()[0]
if (!collection) {
  collection = db.createCollection( .... )
}
```

PaulEndri:
Hi, me again, hope it's okay. I'm actually really excited for this particular change as I've been using changestreams for a while, though looking at it, I do have a question: I think this change would break typescript validation if someone is trying to access to `.fullDocument` without first explicitly checking by type. Pretty sure the break is intended because this is technically more correct, but could it be done so that somehow the `.watch` can automatically limit the returned changeStream types? Maybe through an additional generic? I don't know if it's possible to dynamically pull based on the parameters, but I doubt it. It's just a weird developer experience if I specify to only listen to say chances and replacements and I have to additionally, in the execution logic, additionally specify it

nbbeeken:
Hi @PaulEndri always appreciate the feedback, something I hadn't originally considered when starting this work was the impact that the pipeline has on the change stream document as well, since currently the generic only controls the schema there isn't a way to account for a possible change like `$addFields` in the types. I've put this on the PR for the back burner for the moment while we consider how we could better support that possibility. It may come down to offering a new API that can capture the potential variability of the type. If we had something that better accounted for the pipeline then something like:
```ts
client.watchFor([{ $match: { operationType: 'insert' } }])
```
might be able to filter the actual change stream document type returned. Any ideas are welcome additions to the conversation 🙂

nbbeeken:
I'm proposing we add a second optional generic argument that can override the type of the whole change stream document. Unfortunately you can't skip arguments so the schema will need to be passed in first, but the second argument can be extended from our existing types making adding/omitting properties somewhat ergonomic. 

The impact on users should still be the same as the change proposed so far where certain properties need narrowing to access. Those who want a solution that doesn't require runtime narrowing could use this second argument to strictly type their changes or pass in `any` removing all strictness.

```ts
type InsertChangeWithComment = ChangeStreamInsertDocument<Schema> & { comment: string };

const pipelineChangeStream = collection.watch<Schema, InsertChangeWithComment>([
  { $addFields: { comment: 'big changes' } }, 
  { $match: { operationType: 'insert' } },
]);

pipelineChangeStream.addListener('change', change => {
  expectType<string>(change.comment);
  // No need to narrow in code because the generics did that for us!
  expectType<Schema>(change.fullDocument);
});
```

nbbeeken:
Waiting on #3215 to slim this PR down

baileympearson (dismissed):
LGTM

dariakp (approved):

baileympearson (approved):

PR #3192, opened 2022-04-06
test(NODE-4152): convert cmap test runner to ts and decouple from test context

dariakp:
### Description
NODE-4152
#### What is changing?
Refer to the CMAP runner spec definition [here](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/tests/README.rst).

- The cmap test runner has been converted to typescript and decoupled from the actual unit test definitions
- Most of the heavy lifting is done by the new `ThreadContext` class, which wraps the previously free floating variables pertaining to test pool setup and monitoring and keeps track of separate "threads". _Note:_ For our purposes, a "thread" is a concurrent execution context, basically a promise chain.
- The operation definitions were left unchanged from the original implementation, except that they are now bound to the thread context (the test operations were already correct based on the spec)
- The test assertions have been tightened to fully match the spec

N/A

The upcoming maxConnecting work introduces new cmap integration tests that use the same base format and assertion logic as the current unit tests. The runner needs to be decoupled from the unit test context in order to be reusable in integration tests.


test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+
+export class ThreadContext {
+  pool: ConnectionPool;
+  threads: Map<any, Thread>;
+  connections: Map<string, Connection>;
+  orphans: Set<Connection>;
+  poolEvents = [];
+  poolEventsEventEmitter = new EventEmitter();
+  hostAddress;
+  supportedOperations;
```
nbbeeken, line 227:
```suggestion
  supportedOperations: ReturnType<typeof getTestOpDefinitions>;
```

It can start here but then lead to typing the indexing of `_knownCommands` in thread, I know we have a runtime check but the more we try to strictly type the better early warning one gets when they start modifying this code for the next runner update. (Within reason, we've already improved greatly I think the type improvements I'm calling out are not blockers.)

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+
+    const threadKey = op.name === 'checkOut' ? op.thread || MAIN_THREAD_KEY : MAIN_THREAD_KEY;
+    const thread = threadContext.getThread(threadKey);
+
+    thread.queue(op);
+  }
+
+  await mainThread.finish().catch(e => {
+    actualError = e;
+  });
```
nbbeeken, line 314:
```suggestion
const actualError = await mainThread.finish().catch(error => error);
```
We could still consider our inline catch pattern and update the else block to assert the value is just not a decendent of error? Or finish could return a struct of `{result, error}` I'm just thinking it's easier to get type information when the code doesn't asign to scope variable, but this comment is conditional on other type improvements so depends on where we land.

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+    }
+  }
+}
+
+/**
+ * Implements the spec test match function, see:
+ * [CMAP Spec Test README](https://github.com/mongodb/specifications/tree/master/source/connection-monitoring-and-pooling/tests#spec-test-match-function)
+ */
+const compareInputToSpec = (input, expected) => {
+  if (expected === 42) {
```
nbbeeken, line 131: The spec says `If expected is "42" or 42:` so I think we should be checking the string case as well, even if there are none currently.

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+    expect(input).to.be.ok; // not null or undefined
+    return;
+  }
+
+  if (Array.isArray(expected)) {
```
nbbeeken, line 136: > Assert that actual is of the same JSON type as expected

Not sure this applies to us 🤔 should we be doing something like `expect(JSON.parse(JSON.stringify(actual)).to.equal.(expected)`? I'm thinking of the edge cases where NaN turns into null or something

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+type cmapPoolOptions = {
+  maxPoolSize?: number;
+  minPoolSize?: number;
+  maxIdleTimeMS?: number;
+  waitQueueTimeoutMS?: number;
+};
+
+type cmapEvent = {
+  type: string;
+  address?: 42;
```
nbbeeken, line 26: Per the other 42 comment, these have to be `42 | '42'` Should we pull it out to a named type to indicate it's magical purpose?

```ts
type TheMeaningOfLife = 42 | '42';
```

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+  'style',
+  'description',
+  'poolOptions',
+  'operations',
+  'error',
+  'events',
+  'ignore'
+];
+export type cmapTest = {
+  name?: string; // filename path added by the spec loader
```
nbbeeken, line 44: If it's always added is it optional?

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+      throw new Error(`Attempted to release non-existient connection ${op.connection}`);
+    }
+
+    return threadContext.pool.checkIn(connection);
+  },
+  clear: function () {
+    return threadContext.pool.clear();
+  },
+  close: function () {
+    return util.promisify(ConnectionPool.prototype.close).call(threadContext.pool);
```
nbbeeken, line 180: Is it worth awaiting this in here for the stack trace reasons?

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+  },
+  wait: async function (options) {
+    const ms = options.ms;
+    return sleep(ms);
+  },
+  start: function (options) {
+    const target = options.target;
+    const thread = threadContext.getThread(target);
+    thread.start();
```
nbbeeken, line 189: This is so strange, start is equivalent to calling the resolve defined inside a promise's executor function passed to its constructor. So this isn't async? and shouldn't be awaited? Just trying to follow the logic of what this does... 

Edit: Oh so I think I see it, basically this holds up the start promise, and since our run operations are chained on after this, calling start will unblock them in a sense. Is that right?

test/unit/assorted/connection_monitoring_and_pooling.spec.test.ts
```diff
@@ -0,0 +1,37 @@
+import { isHello } from '../../../src/utils';
+import { loadSpecTests } from '../../spec';
+import { cmapTest, runCmapTest, ThreadContext } from '../../tools/cmap_spec_runner';
+import * as mock from '../../tools/mongodb-mock/index';
+
+describe('Connection Monitoring and Pooling Spec Tests', function () {
+  let hostAddress, threadContext;
```
nbbeeken, line 7: Type annotations?

test/unit/assorted/connection_monitoring_and_pooling.spec.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  after(() => mock.cleanup());
+  before(async () => {
+    const server = await mock.createServer();
```
nbbeeken, line 10: Not something to do here but do we actually have a ticket to move off this or do we all just share the sentiment? I feel like for simple rounttripping of hellos there's something with sinon that we could do. But we can defer.

nbbeeken (changes requested):
Mostly nice to haves, some requirements with matches logic

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
+import { expect } from 'chai';
+import { EventEmitter } from 'events';
+import * as util from 'util';
+
+import { Connection } from '../../src';
+import { ConnectionPool } from '../../src/cmap/connection_pool';
+import { sleep } from './utils';
+
+type cmapOperation =
```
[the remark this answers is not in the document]
dariakp: sure thing, there was a lot of refactoring here, didn't pay attention to capitalization lol

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  start: () => void;
+
+  queue(op: cmapOperation) {
+    if (this._killed || this._error) {
+      return;
+    }
+
+    this._promise = this._promise.then(() => this._runOperation(op)).catch(e => (this._error = e));
```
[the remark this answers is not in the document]
dariakp: I left this logic untouched from the original implementation; I can update it to remove the unnecessary assignment, but I don't think the mongo error bit is necessary since we do check the error in the assertions later on

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+ */
+const compareInputToSpec = (input, expected) => {
+  if (expected === 42) {
+    expect(input).to.be.ok; // not null or undefined
+    return;
+  }
+
+  if (Array.isArray(expected)) {
+    expect(input).to.be.an('array');
+    expected.forEach((expectedValue, index) => {
```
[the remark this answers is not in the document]
dariakp: for of is a bit awkward with indices, and doing a `.entries` on an array feels icky, but maybe that's just me

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+      compareInputToSpec(input[expectedPropName], expectedValue);
+    }
+    return;
+  }
+
+  expect(input).to.equal(expected);
+};
+
+const getTestOpDefinitions = (threadContext: ThreadContext) => ({
+  checkOut: async function (op) {
```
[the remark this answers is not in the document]
dariakp: there isn't a nice way to do this without breaking up cmapOperation into a union of distinct operation types which felt like overkill on an already large PR; I'd rather defer this to when we make further refinements to absorb the integration tests

test/tools/cmap_spec_runner.ts
dariakp: I'll see what I can do

test/tools/cmap_spec_runner.ts
dariakp: I don't have a preference, again, I left the actual logic of the test execution unchanged as much as possible to limit the scope of stuff that needs to be reviewed

test/tools/cmap_spec_runner.ts
dariakp: that was intentional actually - the thought was we keep it as is and update it whenever we sync such a test, because it's going to fail at that point, and it reduces the mental load of trying to check everything the function does to just the things that are relevant

test/tools/cmap_spec_runner.ts
dariakp: no, the final assertion check already takes care of it; for an array, we do check explicitly, for everything else, it's going to iterate through it down to the last key which has to be a primitive type, where a reference check is made, you can't have two references equal each other if they aren't of the same type; if you have an input that is NaN for the "expected" then that last assertion will always fail; if you have an input that is a NaN, again, it won't pass the check

test/tools/cmap_spec_runner.ts
dariakp: I think the link to the CMAP spec readme heading is sufficient, and I have the same response here as on the other comment - let's have the bare minimum logic that gets the test working for the existing definitions. You may notice that the thread is supposed to be able to be specified on any command, but I excluded it from types for any where it actually wouldn't be defined. I want the runner to be representative of actual expected input that we have actually seen. Everything else will error if new things are added, and can be addressed then.

test/tools/cmap_spec_runner.ts
dariakp: it's only added if you load the tests via the spec loader, it's conceivable that you may parse the definitions directly and this should still work, since it's not actually part of the json test definition

test/tools/cmap_spec_runner.ts
dariakp: sure

test/tools/cmap_spec_runner.ts
dariakp: yup; this logic hasn't changed from the original implementation

test/unit/assorted/connection_monitoring_and_pooling.spec.test.ts
dariakp: sure

test/unit/assorted/connection_monitoring_and_pooling.spec.test.ts
dariakp: out of scope, I don't know about a ticket, there's higher priority things to do - it works for now

test/tools/cmap_spec_runner.ts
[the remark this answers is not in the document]
dariakp: @nbbeeken I actually double checked this, and, per the [official docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises):
> Here's the magic: the then() function returns a new promise, different from the original

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,336 @@
...
+
+type cmapEvent = {
+  type: string;
+  address?: 42;
+  connectionId?: number;
+  options?: 42 | cmapPoolOptions;
+  reason: string;
+};
+
+const knownTestKeys = [
```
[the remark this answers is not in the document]
dariakp: done

test/tools/cmap_spec_runner.ts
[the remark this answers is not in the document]
dariakp: done

test/tools/cmap_spec_runner.ts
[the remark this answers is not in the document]
dariakp: refactored to use entries

test/tools/cmap_spec_runner.ts
dariakp: made just this change

test/tools/cmap_spec_runner.ts
dariakp: left this part unchanged for now, let me know if you have a strong preference here

test/tools/cmap_spec_runner.ts
dariakp: done

test/unit/assorted/connection_monitoring_and_pooling.spec.test.ts
dariakp: added

nbbeeken (dismissed):

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,338 @@
+import { expect } from 'chai';
+import { EventEmitter } from 'events';
+import * as util from 'util';
+
+import { Connection, HostAddress } from '../../src';
+import { ConnectionPool } from '../../src/cmap/connection_pool';
+import { sleep } from './utils';
+
+type CmapOperation =
```
baileympearson, line 9: just a question - do we have a practice of when we use interfaces and when we use types?

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,338 @@
...
+  poolEvents = [];
+  poolEventsEventEmitter = new EventEmitter();
+  hostAddress: HostAddress;
+  supportedOperations: ReturnType<typeof getTestOpDefinitions>;
+
+  constructor(hostAddress) {
+    this.threads = new Map();
+    this.connections = new Map();
+    this.orphans = new Set();
+    this.poolEvents = [];
```
durran, line 235: nit: I think these 2 lines can be removed as we are defaulting them to exactly the same objects above on line 226 & 227.

durran (changes requested):

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,338 @@
...
+
+type CmapEvent = {
+  type: string;
+  address?: 42;
+  connectionId?: number;
+  options?: 42 | CmapPoolOptions;
+  reason: string;
+};
+
+const knownTestKeys: Array<keyof CmapTest> = [
```
baileympearson, line 32:
```suggestion
const KNOWN_TEST_KEYS: Array<keyof CmapTest> = [
```

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,338 @@
...
+  // concurrent execution context
+  constructor(operations) {
+    this._promise = new Promise(resolve => {
+      this.start = () => resolve();
+    });
+
+    this._knownCommands = operations;
+  }
+
+  start: () => void;
```
baileympearson, line 97: nit: can we move this above the constructor so property declarations are grouped together?

test/tools/cmap_spec_runner.ts
```diff
@@ -0,0 +1,338 @@
...
+  } else {
+    expect(actualError).to.not.exist;
+  }
+
+  const actualEvents = threadContext.poolEvents.filter(
+    ev => !ignoreEvents.includes(getEventType(ev))
+  );
+
+  expect(actualEvents).to.have.lengthOf(expectedEvents.length);
+  for (const expected of expectedEvents) {
```
baileympearson, line 332:
```suggestion
for (const [expected, actual] of zip(expectedEvents, actualEvents)) {
    const { type: eventType, ...eventPropsToCheck } = expected;
    expect(actual.constructor.name).to.equal(`${eventType}Event`);
    compareInputToSpec(actual, eventPropsToCheck);
}
```

baileympearson (changes requested):

test/tools/cmap_spec_runner.ts
dariakp: yeah let me clean this up, most of these don't need to be set in the constructor

test/tools/cmap_spec_runner.ts
dariakp: I usually think of types as more static (unchanging) than interfaces, but [functionally](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) they're more or less equivalent in most cases and in this one specifically. @nbbeeken might have a stronger opinion here

test/tools/cmap_spec_runner.ts
dariakp: right now this is the same as in the uri spec runner but I agree that in this file it's not consistent with the events list constant (since I didn't rename that from the original implementation) - do we have an opinion for all caps vs regular camelcase for general test constants? @nbbeeken @durran

test/tools/cmap_spec_runner.ts
dariakp: sure

test/tools/cmap_spec_runner.ts
dariakp: what's the benefit beyond personal preference?

baileympearson: It's just preference - I don't like mutating objects when I can help it.  Either way is okay

test/tools/cmap_spec_runner.ts
baileympearson: Would test constants be different than regular constants?  Our convention with constants seems to be all caps and snake case

test/tools/cmap_spec_runner.ts
nbbeeken: I agree that here it's equivalent, and we don't need to ever `extends` these so type does the job.

From the linked docs though I do like this feature of interfaces: 
- Interface names will [always appear in their original form](https://www.typescriptlang.org/play?#code/PTAEGEHsFsAcEsA2BTATqNrLusgzngIYDm+oA7koqIYuYQJ56gCueyoAUCKAC4AWHAHaFcoSADMaQ0PCG80EwgGNkALk6c5C1EtWgAsqOi1QAb06groEbjWg8vVHOKcAvpokshy3vEgyyMr8kEbQJogAFND2YREAlOaW1soBeJAoAHSIkMTRmbbI8e6aPMiZxJmgACqCGKhY6ABGyDnkFFQ0dIzMbBwCwqIccabcYLyQoKjIEmh8kwN8DLAc5PzwwbLMyAAeK77IACYaQSEjUWY2Q-YAjABMAMwALA+gbsVjNXW8yxySoAADaAA0CCaZbPh1XYqXgOIY0ZgmcK0AA0nyaLFhhGY8F4AHJmEJILCWsgZId4NNfIgGFdcIcUTVfgBlZTOWC8T7kAJ42G4eT+GS42QyRaYbCgXAEEguTzeXyCjDBSAAQSE8Ai0Xsl0K9kcziExDeiQs1lAqSE6SyOTy0AKQ2KHk4p1V6s1OuuoHuzwArMagA) in error messages, but only when they are used by name.

When we do type checking for tests or just in general this is a good thing to keep in mind for UX. But no need to make a change here.

test/tools/cmap_spec_runner.ts
durran: I'd say all caps is how I would generally define constants.

nbbeeken: Using all caps is a good hint that something is meant to be a set of static of well known values. If we go with caps I think we should call it `TEST_KEYS`, but I don't feel strongly. If it's not part of an API then it's changeable at will 🤷

test/tools/cmap_spec_runner.ts
durran: Thanks. I'm a LGTM after that in case it gets late here in the evening.

test/tools/cmap_spec_runner.ts
nbbeeken: I don't think I would consider this a preference as long as there is reason to *not* mutate a value. A reason I can think of is that when you are debugging, you cannot go back and reference what was in the array on an earlier iteration because it is gone from the list. But I do think this is just QoL, but QoL is important too 🙂  (but not a blocker)

test/tools/cmap_spec_runner.ts
dariakp: updated

test/tools/cmap_spec_runner.ts
dariakp: split the difference to `CMAP_TEST_KEYS`

test/tools/cmap_spec_runner.ts
dariakp: done

durran (approved):

test/tools/cmap_spec_runner.ts
dariakp: In this case, actualEvents exists specifically for this check, so I don't think the extra dependency import gains us anything

test/tools/cmap_spec_runner.ts
baileympearson: Named types also appear in error messages when used by name too 😃 ([playground](https://www.typescriptlang.org/play?#code/PTAEGEHsFsAcEsA2BTATqNrLusgzngIYDm+oA7koqIYuYQJ56gCueyoAUCKAC4AWHAHaFcoSADMaQ0PCG80EwgGNkALk6deDWBwCyo6LVABeUAG9Ooa6BG41oPL1RzinAL6aJLIct7xIGWRlfkgDaCNEAApoB3DIgEoLKxtlQLxIFAA6REhiGKy7ZASPTR5kLOIs0AAVQQxULHQAI2Rc8goqGjpGZjYOAWFRfUNjbjBeSFBUZAk0PinBvh0Ocn54ENlmZAAPXT9kABMNYND42ijzW2GHAEYAJgBmABZH0HcS8dr67V1xKQABtAAaBBDMtnx6rsVLxHMMaMwjBFaAAaL7NFiwwjMeC8ADkzCEkFhrWQMkO8BmfkQDGuuEOKNqKwAysoXLBeF9yIE8bDcPIAjJcbIZEtMNhQLgCCRXF4fH5BRgQpAAIJCeCRGIOK5FBxOFxCYjvJKWGygNJCDLZXL5aCFYYlTycU6q9WanU3UAPF4AVmNQA))

Agreed that there's no need to make a change here so I'll resolve this conversation

nbbeeken (approved):

baileympearson (approved):

PR #3199, opened 2022-04-12
fix(NODE-3928): don't throw error in Response constructor

durran:
### Description

Don't raise an error in the `Response` constructor.

#### What is changing?

Moves all logic that reads from the message body into the `parse` method to allow for any raised exception to be properly handled in a callback.

No.

NODE-3928

### DIFF
--- a/test/unit/cmap/commands.test.js
+++ b/test/unit/cmap/commands.test.js
@@ -35,9 +35,8 @@ describe('commands', function () {
         it('does not throw an exception', function () {
           let error;
           try {
-            const response = new Response(message, header, body);
-          }
-          catch(err) {
+            new Response(message, header, body);
+          } catch (err) {
             error = err;
           }
           expect(error).to.be.undefined;
--- a/test/unit/cmap/commands.test.js
+++ b/test/unit/cmap/commands.test.js
@@ -1,0 +1,24 @@
+const { expect } = require('chai');
+const { Response } = require('../../../src/cmap/commands');
+
+describe('commands', function () {
+  describe('Response', function () {
+    describe('#constructor', function () {
+      context('when the message body is invalid', function () {
+        const message = Buffer.from([]);
+        const header = {
+          length: 0,
+          requestId: 0,
+          responseTo: 0,
+          opCode: 0
+        };
+        const body = Buffer.from([]);
+
+        it('does not throw an exception', function () {
+          const response = new Response(message, header, body);
+          expect(response.numberReturned).to.be.undefined;
+        });
+      });
+    });
+  });
+});
--- a/test/unit/cmap/commands.test.js
+++ b/test/unit/cmap/commands.test.js
@@ -3,6 +3,24 @@ const { Response } = require('../../../src/cmap/commands');
 
 describe('commands', function () {
   describe('Response', function () {
+    describe('#parse', function () {
+      context('when the message body is invalid', function () {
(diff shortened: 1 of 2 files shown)


test/unit/cmap/commands.test.js
```diff
@@ -0,0 +1,24 @@
...
+        const header = {
+          length: 0,
+          requestId: 0,
+          responseTo: 0,
+          opCode: 0
+        };
+        const body = Buffer.from([]);
+
+        it('does not throw an exception', function () {
+          const response = new Response(message, header, body);
```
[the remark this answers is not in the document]
durran: I've added all the additional cases now. Note that the test that the `parse` method throws when the values are invalid does not check the error message as I suspected the windows error message would differ from the -nix error messages.

baileympearson (dismissed):
Thanks for adding the tests.  Looks good!

test/unit/cmap/commands.test.js
```diff
@@ -0,0 +1,92 @@
...
+
+        it('does not throw an exception', function () {
+          let error;
+          try {
+            new Response(message, header, body);
+          } catch (err) {
+            error = err;
+          }
+          expect(error).to.be.undefined;
+        });
```
nbbeeken, line 43: Let's add a test case that reproduced the original bug:
```js
const b = Buffer.alloc(5 * 4) // 5 int32s
b.writeInt32LE(-1, 16) // numReturned location
/// constructor doesn't throw, parse does with RangeError
```

test/unit/cmap/commands.test.js
```diff
@@ -0,0 +1,92 @@
+const { expect } = require('chai');
+const { Response } = require('../../../src/cmap/commands');
+
+describe('commands', function () {
+  describe('Response', function () {
+    describe('#parse', function () {
+      context('when the message body is invalid', function () {
```
nbbeeken, line 7: Let's expand the title here, what about the body is invalid? beacuse it's empty correct? We should add a case with invalid data as well, which I think is only the numReturned field (See comment with code).

test/unit/cmap/commands.test.js
```diff
@@ -0,0 +1,92 @@
...
+          length: 0,
+          requestId: 0,
+          responseTo: 0,
+          opCode: 0
+        };
+        const body = Buffer.from([]);
+
+        it('throws an exception', function () {
+          const response = new Response(message, header, body);
+          expect(() => response.parse()).to.throw();
```
nbbeeken, line 19: This throws because you can't read an int32 from no bytes, let's add to the throw condition to check for that (like message or error type)

nbbeeken (changes requested):

test/unit/cmap/commands.test.js
durran: Test case added and checks error type and message.

test/unit/cmap/commands.test.js
durran: I've broke this down into 2 subcontexts now that say 1) the body is empty and 2) nuReturned is invalid.

test/unit/cmap/commands.test.js
durran: Now checks both for `RangeError` and the message.

nbbeeken (approved):
LGTM!

baileympearson (approved):

PR #3200, opened 2022-04-12
chore: fix sinon leak in tests

nbbeeken:
#### What is changing?

We accidentally left a sinon spy outstanding which accumulates memory for all the usages of the function. It started to cause an issue on my connect expirementation PR because failures seems to exagerrate the problem, with this change the memory usage of a test run remains constant.

Quick fix to help others that might hit this.

### DIFF
--- a/test/integration/node-specific/bson-options/utf8_validation.test.ts
+++ b/test/integration/node-specific/bson-options/utf8_validation.test.ts
@@ -1,9 +1,7 @@
 import { expect } from 'chai';
-import { spy } from 'sinon';
+import * as sinon from 'sinon';
 
 import * as BSON from '../../../../src/bson';
-
-const deserializeSpy = spy(BSON, 'deserialize');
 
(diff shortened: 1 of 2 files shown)


test/integration/node-specific/bson-options/utf8_validation.test.ts
```diff
@@ -16,10 +14,18 @@ const EXPECTED_VALIDATION_ENABLED_ARGUMENT = {
 };
 
 describe('class BinMsg', () => {
+  let deserializeSpy: sinon.SinonSpy;
+
   beforeEach(() => {
+    deserializeSpy = sinon.spy(BSON, 'deserialize');
     deserializeSpy.resetHistory();
```
baileympearson, line 21: nit: if we're creating a new spy every time, do we need to reset it on creation?

nbbeeken: fixed!

durran (approved):
Nice catch!

baileympearson (approved):

PR #3201, opened 2022-04-14
refactor(NODE-2992): clean up server selection logic runner

baileympearson:
#### What is changing?

Implements the server selection logic tests are described here: https://github.com/mongodb/specifications/tree/master/source/server-selection/tests#server-selection-logic-tests

Nope! 

Our server selection test runner was incorrect.  That, and Typescript test runners = happy developers 


test/unit/assorted/server_selection.spec.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import {
+  collectServerSelectionLogicTests,
+  runServerSelectionLogicTest
+} from './server_selection_logic_spec_utils';
+
+describe('Server Selection Logic (spec)', function () {
+  beforeEach(function () {
+    if (this.currentTest.title.match(/Possible/)) {
+      (this.currentTest as any).skipReason = 'Nodejs driver does not support PossiblePrimary';
```
nbbeeken, line 11: I'd rather the red underline remain rather than casting this. When we do start type checking our tests correctly we'll catch issues if this feature was accidentally used incorrectly.

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,151 @@
+import { Document, EJSON } from 'bson';
```
nbbeeken, line 1: "logic_spec_utils" implies to me these won't be used in the forthcoming integration tests? Could we keep all this logic inside the `server_selection.spec.test.ts` test file. I think if the logic isn't common to anything else keeping it all in one file is worth keeping things associated. But if this is reusable, then maybe the name should be more generic? My end goal is just making discoverability as clear as possible.

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,151 @@
...
+import { TopologyDescription } from '../../../src/sdam/topology_description';
+import { serverDescriptionFromDefinition } from './server_selection_spec_helper';
+
+interface ServerSelectionLogicTestServer {
+  address: string;
+  avg_rtt_ms: number;
+  type: ServerType;
+  tags?: TagSet;
+}
+interface Test {
```
nbbeeken, line 23:
```suggestion
interface ServerSelectionTest {
```

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,151 @@
...
+        .filter(f => extname(f) === '.json')
+        .map(f => {
+          const fileContents = readFileSync(join(specDir, testType, subType, f), {
+            encoding: 'utf-8'
+          });
+          const test = EJSON.parse(fileContents, { relaxed: true }) as unknown as Document;
+          test.name = basename(f, '.json');
+          test.type = testType;
+          test.subType = subType;
+          return test;
```
nbbeeken, line 143: Similar to what I did [here](https://github.com/mongodb/node-mongodb-native/blob/main/test/unit/assorted/server_discovery_and_monitoring.spec.test.ts#L149-L171) for the SDAM runner can we assert what we know to be true about the test shape at this stage, it gives us a warning for future syncs if there's something wrong with the format. **Or** if there's a new key that we don't have documented in our types, which means we probably don't test for it

nbbeeken (changes requested):

test/unit/assorted/server_selection_logic_spec_utils.ts
baileympearson: Some of these functions might be reused in the forthcoming work and when that happens, I'll move them.  But each of the three test runners for server selection have a different test format, so a lot of the code here will not be reused.

test/unit/assorted/server_selection_logic_spec_utils.ts
baileympearson: Renamed!  Let me know if you're okay with the name.

test/unit/assorted/server_selection.spec.test.ts
baileympearson: Done!

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,154 @@
...
+  let selector;
+  if (testDefinition.operation === 'write') {
+    selector = ServerSelectors.writableServerSelector();
+  } else if (testDefinition.operation === 'read' || testDefinition.read_preference) {
+    try {
+      const readPreference = readPreferenceFromDefinition(testDefinition.read_preference);
+      selector = ServerSelectors.readPreferenceServerSelector(readPreference);
+    } catch (e) {
+      expect(e, ejson`Invalid readPreference: ${testDefinition.read_preference}`).not.to.exist;
+    }
```
nbbeeken, line 106: Did you want to catch this error and check it outside the block? I think e will always exist here.

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
```
nbbeeken, line 107: Can we add an else condition here, if it should be unreachable then we can `expect.fail`

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  const result = selector(topologyDescription, serversInTopology);
+
+  expect(result.length).to.equal(expectedServers.size);
```
nbbeeken, line 111: A selector should always return at least 1 server right? can we assert that in addition to the size match?

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,154 @@
...
+    );
+    const result = {};
+    for (const subType of testsOfType) {
+      result[subType] = readdirSync(join(specDir, testType, subType))
+        .filter(f => extname(f) === '.json')
+        .map(f => {
+          const fileContents = readFileSync(join(specDir, testType, subType, f), {
+            encoding: 'utf-8'
+          });
+          const test = EJSON.parse(fileContents, { relaxed: true }) as unknown as Document;
```
nbbeeken, line 142: 🤔  We should be careful with relaxed true, I think we don't have to change it here, but something we should consider auditing if we're potentially transforming information we shouldn't be

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,154 @@
...
+
+function readPreferenceFromDefinition(definition) {
+  const mode = definition.mode
+    ? definition.mode.charAt(0).toLowerCase() + definition.mode.slice(1)
+    : 'primary';
+
+  const options: ReadPreferenceOptions = {};
+  if (typeof definition.maxStalenessSeconds !== 'undefined')
+    options.maxStalenessSeconds = definition.maxStalenessSeconds;
+  const tags = definition.tag_sets || [];
```
nbbeeken, line 51:
```suggestion
  const tags = definition.tag_sets ?? [];
```

test/unit/assorted/server_selection_logic_spec_utils.ts
```diff
@@ -0,0 +1,154 @@
...
+import { basename, extname, join } from 'path';
+
+import {
+  ReadPreference,
+  ReadPreferenceMode,
+  ReadPreferenceOptions
+} from '../../../src/read_preference';
+import { ServerType, TopologyType } from '../../../src/sdam/common';
+import { ServerDescription, TagSet } from '../../../src/sdam/server_description';
+import * as ServerSelectors from '../../../src/sdam/server_selection';
```
nbbeeken, line 13: Can we import only the selectors we're testing? it makes it easy to grok the coverage

test/unit/assorted/server_selection_logic_spec_utils.ts
nbbeeken: I see how the tests cover this themselves, so no need for an input validator.

nbbeeken (changes requested):

test/unit/assorted/server_selection_logic_spec_utils.ts
baileympearson: I was using this as a way to automatically throw, because it would exist.  I think I'll go with an `expect.fail` if that works for you?

test/unit/assorted/server_selection_logic_spec_utils.ts
baileympearson: Nope!  We return an array. And there are tests that assert there will be no matching servers - see https://github.com/mongodb/specifications/blob/master/source/server-selection/tests/server_selection/Unknown/read/SecondaryPreferred.yml for example

test/unit/assorted/server_selection_logic_spec_utils.ts
baileympearson: I tried removing `relaxed: true` and the tests still pass.  I'm okay making this change now if you are!

test/unit/assorted/server_selection_logic_spec_utils.ts
baileympearson: Done!

test/unit/assorted/server_selection_logic_spec_utils.ts
nbbeeken: relaxed true is the default so you would need to test with `relaxed: false`

test/unit/assorted/server_selection_logic_spec_utils.ts
nbbeeken: Oh yea that would make it clear that it's a never an expected case, should we catch at all? I suppose the message is more helpful than what comes out of the API? but then that raises the question, should the API throw a better message? For now expect.fail will serve our needs

test/unit/assorted/server_selection_logic_spec_utils.ts
nbbeeken: Okay awesome, ty for checking.

test/unit/assorted/server_selection_logic_spec_utils.ts
nbbeeken: From slack: this adds too much complexity, no value

nbbeeken (approved):
lgtm!

PR #3202, opened 2022-04-15
test(NODE-4152): make connect auth check configurable

nbbeeken:
#### What is changing?

We can now disable the initial ping with a flag. The unified runner and spec runner now disable the initial ping, and have had their "initial ping" handler code removed. 

Necessary drive by fixes:
- SCRAM iteration tests declared to the Topology directly which is something we want to do less of in tests anyway but this resulted in a timeout before fixing the defaulting of the flag. 
- EVG typescript tests were not marked as "test"s so they would appear purple instead of red when they failed
- TS fixes inside UnifiedMongoClient class, should be a type error free file now

For use only in testing, we can disable the fact that connect will send a ping if there are credentials, this will make auth errors surface during the first operation a test performs instead of during entity creation.

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -1239,5 +1239,8 @@ export const DEFAULT_OPTIONS = new CaseInsensitiveMap(
     .map(([k, d]) => [k, d.default])
 );
 
-/** Set of permitted feature flags @internal */
+/**
+ * Set of permitted feature flags
+ * @internal
+ */
 export const FEATURE_FLAGS = new Set([Symbol.for('@@mdb.skipPingOnConnect')]);
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -273,12 +273,7 @@ export function parseOptions(
   // Feature flags
   for (const flag of Object.getOwnPropertySymbols(options)) {
     if (FEATURE_FLAGS.has(flag)) {
-      Object.defineProperty(mongoOptions, flag, {
-        value: options[flag],
-        enumerable: false,
-        writable: true,
-        configurable: true
-      });
+      mongoOptions[flag] = options[flag];
     }
   }
 
@@ -1245,4 +1240,4 @@ export const DEFAULT_OPTIONS = new CaseInsensitiveMap(
 );
 
 /** Set of permitted feature flags @internal */
-export const FEATURE_FLAGS = new Set([Symbol.for('@@mdb.skipInitialPing')]);
+export const FEATURE_FLAGS = new Set([Symbol.for('@@mdb.skipPingOnConnect')]);
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -269,6 +269,18 @@ export function parseOptions(
   const { hosts, isSRV } = url;
 
   const mongoOptions = Object.create(null);
+
+  // Feature flag capturing
+  for (const [flag, defaultFeatureSetting] of FEATURE_FLAGS) {
+    // Use 'in' to support null values for flags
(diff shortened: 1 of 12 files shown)


test/integration/node-specific/feature_flags.test.ts
```diff
@@ -0,0 +1,37 @@
+import { expect } from 'chai';
+
+describe.only('Feature Flags', () => {
```
baileympearson, line 3: oh no!  a rogue `.only`

test/unit/connection_string.test.ts
```diff
@@ -410,4 +410,46 @@ describe('Connection String', function () {
       expect(options).to.have.nested.property('credentials.source', 'thisShouldBeAuthSource');
     });
   });
+
+  describe('feature flags', () => {
+    it('should map known symbols onto options', () => {
```
baileympearson, line 415:
```suggestion
    it('should should ignore unknown symbols', () => {
```

test/unit/connection_string.test.ts
```diff
@@ -410,4 +410,46 @@ describe('Connection String', function () {
...
+      const randomFlag = Symbol();
+      const client = new MongoClient('mongodb://iLoveJavaScript', { [randomFlag]: 23n });
+      expect(client.s.options).to.not.have.property(randomFlag);
+    });
+
+    it('should map all default symbols and values onto options', () => {
+      const client = new MongoClient('mongodb://iLoveJavaScript');
+      expect(FEATURE_FLAGS.size).to.be.greaterThanOrEqual(1);
```
baileympearson, line 423: Are they expectations about the size of feature flags necessary?  FEATURE_FLAGS is a const set

baileympearson (changes requested):

src/connection_string.ts
```diff
@@ -1230,3 +1238,6 @@ export const DEFAULT_OPTIONS = new CaseInsensitiveMap(
     .filter(([, descriptor]) => descriptor.default != null)
     .map(([k, d]) => [k, d.default])
 );
+
+/** Set of permitted feature flags @internal */
```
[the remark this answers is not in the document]
nbbeeken: fixed!

src/mongo_client.ts
```diff
@@ -254,6 +254,9 @@ export interface MongoClientOptions extends BSONSerializeOptions, SupportedNodeC
   srvPoller?: SrvPoller;
   /** @internal */
   connectionType?: typeof Connection;
+
+  /** @internal */
+  [featureFlag: symbol]: any;
```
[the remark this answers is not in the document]
nbbeeken: This is also used for the options argument in the constructor, so that's why the change was needed, the internal will keep it out of public docs, maybe when/if we do a public feature flag we would do something but currently I feel like its overkill to put in that logic.

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -0,0 +1,45 @@
+import { expect } from 'chai';
+
+describe('Feature Flags', () => {
+  beforeEach(function () {
```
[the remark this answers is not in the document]
nbbeeken: Done!

test/unit/connection_string.test.ts
nbbeeken: fixed!

test/unit/connection_string.test.ts
nbbeeken: This test isn't even doing anything of worth anymore now that we removed the defaulting so I removed it

baileympearson (approved):

PR #3203, opened 2022-04-17
fix(NODE-4108): improve return type for `withTransaction()`

vkarpov15:
### Description

The `withTransaction()` function signature implies that `withTransaction()` returns the same value that the user-provided callback returns. But that isn't the case, it always returns something like this:

```
{
  ok: 1,
  '$clusterTime': {
    clusterTime: new Timestamp({ t: 1650213729, i: 1 }),
    signature: {
      hash: new Binary(Buffer.from("0000000000000000000000000000000000000000", "hex"), 0),
      keyId: 0
    }
  },
  operationTime: new Timestamp({ t: 1650213729, i: 1 })
}
```

Reported in https://github.com/Automattic/mongoose/issues/9919

#### What is changing?

Function signature for `withTransaction()`

### DIFF
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -467,10 +467,10 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    * @param fn - A lambda to run within a transaction
    * @param options - Optional settings for the transaction
    */
-  withTransaction<T = void>(
-    fn: WithTransactionCallback<T>,
+  withTransaction(
+    fn: WithTransactionCallback<void>,
     options?: TransactionOptions
-  ): ReturnType<typeof fn> {
+  ): Promise<void> {
     const startTime = now();
     return attemptTransaction(this, startTime, fn, options);
   }


src/sessions.ts
```diff
@@ -467,10 +467,10 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    * @param fn - A lambda to run within a transaction
    * @param options - Optional settings for the transaction
    */
-  withTransaction<T = void>(
-    fn: WithTransactionCallback<T>,
+  withTransaction(
+    fn: WithTransactionCallback<void>,
     options?: TransactionOptions
-  ): ReturnType<typeof fn> {
+  ): Promise<void> {
```
nbbeeken, line 473: Even though the return type is meant to be an internal command result we should reflect what's actually returned here and we can fix this in the next major version. This function will return the command result if the transaction commits successfully, it will throw if it aborts and it will return undefined if it's aborted manually within the user provided function, so it's important that the result at least demonstrate that it can be truthy or not under the right circumstance. 
So we should put the type in here that matches the actual result from the function, and add tests to prevent a regression.

nbbeeken (changes requested):
It just so happens a ticket to fix this work was next up on our scheduled queue, let me know if it would be easier for you if I took on making the changes mentioned:

nbbeeken:
https://github.com/mongodb/node-mongodb-native/pull/3236

PR #3205, opened 2022-04-18
chore(NODE-3723): add node 16 to ci

nbbeeken:
#### What is changing?

Adds Node 16 to the matrix, removes unused test tasks. Node 18 will be added in a follow up

New node new me


nbbeeken:
In the most recent run I forgot to check off 16 it should be running now. it has passed in the previous commits already though. We currently use 12 by default, we should keep it that way I think

baileympearson (approved):

durran (approved):

PR #3206, opened 2022-04-19
feat(NODE-4136): revise FLE shared library typings for spec changes

addaleax:
### Description

Typings changes for https://github.com/mongodb/libmongocrypt/pull/306

### DIFF
--- a/src/deps.ts
+++ b/src/deps.ts
@@ -277,8 +277,12 @@ export interface AutoEncryptionOptions {
    * Other validation rules in the JSON schema will not be enforced by the driver and will result in an error.
    */
   schemaMap?: Document;
+  /** @experimental */
+  encryptedFieldsMap?: Document;
   /** Allows the user to bypass auto encryption, maintaining implicit decryption */
   bypassAutoEncryption?: boolean;
+  /** @experimental */
+  bypassQueryAnalysis?: boolean;
   options?: {
     /** An optional hook to catch logging messages from the underlying encryption engine */
     logger?: (level: AutoEncryptionLoggerLevel, message: string) => void;
--- a/src/deps.ts
+++ b/src/deps.ts
@@ -313,13 +313,15 @@ export interface AutoEncryptionOptions {
      * If this option is not provided and `csfleRequired` is not specified,
      * the AutoEncrypter will attempt to spawn and/or use mongocryptd according
      * to the mongocryptd-specific `extraOptions` options.
+     *
+     * Specifying a path prevents mongocryptd from being used as a fallback.
      */
     csflePath?: string;
     /**
      * If specified, never use mongocryptd and instead fail when the CSFLE shared library
      * could not be loaded.
      *
-     * This is implied when `csflePath` is specified.
+     * This is always true when `csflePath` is specified.
      */
     csfleRequired?: boolean;
     /**
--- a/src/deps.ts
+++ b/src/deps.ts
@@ -296,13 +296,36 @@ export interface AutoEncryptionOptions {
     /** Command line arguments to use when auto-spawning a mongocryptd */
     mongocryptdSpawnArgs?: string[];
     /**
-     * Full path to a CSFLE shared library to be used (instead of mongocryptd)
-     * @experimental
+     * Full path to a CSFLE shared library to be used (instead of mongocryptd).
+     *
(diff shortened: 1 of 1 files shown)


src/deps.ts
```diff
@@ -296,13 +296,36 @@ export interface AutoEncryptionOptions {
...
+     * This is implied when `csflePath` is specified.
+     */
+    csfleRequired?: boolean;
     /**
      * Search paths for a CSFLE shared library to be used (instead of mongocryptd)
-     * @experimental
+     * Only for driver testing!
+     * @internal
      */
     csfleSearchPaths?: string[];
```
addaleax, line 330: Happy to drop this from the typings entirely if you prefer.

src/deps.ts
```diff
@@ -296,13 +296,36 @@ export interface AutoEncryptionOptions {
...
+     * It can be an absolute or relative path. If the path is relative and
+     * its first component is `$ORIGIN`, it will be replaced by the directory
+     * containing the mongodb-client-encryption native addon file. Otherwise,
+     * the path will be interpreted relative to the current working directory.
+     *
+     * Currently, loading different CSFLE shared library files from different
+     * MongoClients in the same process is not supported.
+     *
+     * If this option is provided and no CSFLE shared library could be loaded
+     * from the specified location, creating the MongoClient will fail.
```
nbbeeken, line 311: I believe connecting a MongoClient will fail since that's when we call the `autoEncrypter` init functions. I think it's worth making the distinction in case folks want to check for an error

src/deps.ts
```diff
@@ -296,13 +296,36 @@ export interface AutoEncryptionOptions {
...
+     * If this option is not provided and `csfleRequired` is not specified,
+     * the AutoEncrypter will attempt to spawn and/or use mongocryptd according
+     * to the mongocryptd-specific `extraOptions` options.
      */
     csflePath?: string;
+    /**
+     * If specified, never use mongocryptd and instead fail when the CSFLE shared library
+     * could not be loaded.
+     *
+     * This is implied when `csflePath` is specified.
```
nbbeeken, line 322:
```suggestion
     * This is always true when `csflePath` is specified. Specifying a path prevents mongocryptd from being used as a fallback.
```
Does this mean I cannot provide a path and still expect mongocryptd to work as a fallback? That makes sense to me since the users of the lib are folks who want to move off of or can't use cryptd, but I'm thinking we could make it clear. 
My suggested text is  maybe in the wrong spot, maybe it would be good to add after `loading different CSFLE shared library files from different MongoClients in the same process is not supported.` Thoughts?

src/deps.ts
nbbeeken: Keeping it and marking it as internal means we could drop it any time, for now it's useful for testing so let's keep it :)

nbbeeken (changes requested):

src/deps.ts
addaleax: Creating the libmongocrypt handle is part of creating `AutoEncrypter`, which in turn is parts of options parsing. This is the stack trace:

```
TypeError: A CSFLE override path was specified [/tmp/blah.so], but we failed to open a dynamic library at that location
    at new AutoEncrypter (/home/addaleax/src/libmongocrypt/bindings/node/lib/autoEncrypter.js:135:26)
    at new Encrypter (/home/addaleax/src/libmongocrypt/bindings/node/node_modules/mongodb/lib/encrypter.js:48:30)
    at parseOptions (/home/addaleax/src/libmongocrypt/bindings/node/node_modules/mongodb/lib/connection_string.js:330:34)
    at new MongoClient (/home/addaleax/src/libmongocrypt/bindings/node/node_modules/mongodb/lib/mongo_client.js:62:63)
```

src/deps.ts
addaleax: Sure – I’ve split the sentence and added the second part to `csflePath` rather than this option, I think that makes sense :+1:

src/deps.ts
nbbeeken: Oh perfect, good to know, I think the earlier the failure the better anyway 🚀

nbbeeken (dismissed):

nbbeeken (approved):
LGTM ty!

durran (approved):

PR #3215, opened 2022-04-25
refactor(NODE-4059): move free functions into ChangeStream class

nbbeeken:
#### What is changing?

In the upcoming change stream work a new type parameter made it difficult to plumb through all the new generic constriants to our floating functions. As all these functions do operate on a parameterized change stream it makes sense to make them private methods that have access to the type cosntraints/variables. 

The result of the executeOperation call is now typed as well, in addition to a small change to isResumable, all unknown errors are not resumeable.

More sensible generic sharing to the helper methods. None of the `private _X` methods should accept a generic of their own they will rely on the class scoped generics now. 

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -1,8 +1,9 @@
 import Denque = require('denque');
 import type { Readable } from 'stream';
 
-import type { Document, Timestamp } from './bson';
+import type { Document, Long, Timestamp } from './bson';
 import { Collection } from './collection';
+import { CHANGE, CLOSE, END, ERROR, INIT, MORE, RESPONSE, RESUME_TOKEN_CHANGED } from './constants';
 import {
   AbstractCursor,
   AbstractCursorEvents,
@@ -15,11 +16,10 @@ import {
   isResumableError,
   MongoAPIError,
   MongoChangeStreamError,
-  MongoError,
   MongoRuntimeError
 } from './error';
 import { MongoClient } from './mongo_client';
-import { InferIdType, Nullable, TypedEventEmitter } from './mongo_types';
+import { InferIdType, Nullable, TODO_NODE_3286, TypedEventEmitter } from './mongo_types';
 import { AggregateOperation, AggregateOptions } from './operations/aggregate';
 import type { CollationOptions, OperationParent } from './operations/command';
 import { executeOperation, ExecutionResult } from './operations/execute_operation';
@@ -29,6 +29,7 @@ import type { ClientSession } from './sessions';
 import {
   calculateDurationInMs,
   Callback,
+  filterOptions,
   getTopology,
   maxWireVersion,
   maybePromise,
@@ -67,6 +68,16 @@ const CHANGE_DOMAIN_TYPES = {
   CLUSTER: Symbol('Cluster')
 };
 
+interface TopologyWaitOptions {
+  start?: number;
+  timeout?: number;
+  readPreference?: ReadPreference;
+}
+
+const SELECTION_TIMEOUT = 30000;
+
+const CHANGE_STREAM_EVENTS = [RESUME_TOKEN_CHANGED, END, CLOSE];
+
 const NO_RESUME_TOKEN_ERROR =
   'A change stream document has been received that lacks a resume token (_id).';
 const NO_CURSOR_ERROR = 'ChangeStream has no cursor';
@@ -101,6 +112,18 @@ export type OperationTime = Timestamp;
 export interface PipeOptions {
   end?: boolean;
 }
+
+/** @internal */
+export type ChangeStreamAggregateRawResult<TChange> = {
+  $clusterTime: { clusterTime: Timestamp };
+  cursor: {
+    postBatchResumeToken: unknown;
+    ns: string;
+    id: number | Long;
+  } & ({ firstBatch: TChange[] } | { nextBatch: TChange[] });
+  ok: 1;
+  operationTime: Timestamp;
+};
 
 /**
  * Options that can be passed to a ChangeStream. Note that startAfter, resumeAfter, and startAtOperationTime are all mutually exclusive, and the server will error if more than one is specified.
@@ -233,24 +256,24 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
   [kMode]: false | 'iterator' | 'emitter';
 
   /** @event */
-  static readonly RESPONSE = 'response' as const;
+  static readonly RESPONSE = RESPONSE;
   /** @event */
-  static readonly MORE = 'more' as const;
+  static readonly MORE = MORE;
   /** @event */
-  static readonly INIT = 'init' as const;
+  static readonly INIT = INIT;
   /** @event */
-  static readonly CLOSE = 'close' as const;
+  static readonly CLOSE = CLOSE;
   /**
    * Fired for each new matching change in the specified namespace. Attaching a `change`
    * event listener to a Change Stream will switch the stream into flowing mode. Data will
    * then be passed as soon as it is available.
    * @event
    */
-  static readonly CHANGE = 'change' as const;
+  static readonly CHANGE = CHANGE;
   /** @event */
-  static readonly END = 'end' as const;
+  static readonly END = END;
   /** @event */
-  static readonly ERROR = 'error' as const;
+  static readonly ERROR = ERROR;
   /**
    * Emitted each time the change stream stores a new resume token.
    * @event
@@ -294,7 +317,7 @@ export class ChangeStream<TSchema extends Document = Document> extends TypedEven
     this[kResumeQueue] = new Denque();
 
     // Create contained Change Stream cursor
-    this.cursor = createChangeStreamCursor(this, options);
+    this.cursor = this._createChangeStreamCursor(options);
 
     this[kClosed] = false;
     this[kMode] = false;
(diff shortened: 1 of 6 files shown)


baileympearson (changes requested):

baileympearson (dismissed):

baileympearson (approved):

dariakp (changes requested):

src/error.ts
```diff
@@ -894,7 +894,11 @@ export function isNetworkTimeoutError(err: MongoError): err is MongoNetworkError
 //
 // An error on an aggregate command is not a resumable error. Only errors on a getMore command may be considered resumable errors.
 
-export function isResumableError(error?: MongoError, wireVersion?: number): boolean {
+export function isResumableError(error?: Error, wireVersion?: number): boolean {
+  if (!(error instanceof MongoError)) {
```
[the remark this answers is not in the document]
nbbeeken: Cleaned up the following logic and added more coverage

src/index.ts
```diff
@@ -169,6 +169,7 @@ export type { OrderedBulkOperation } from './bulk/ordered';
 export type { UnorderedBulkOperation } from './bulk/unordered';
 export type {
   ChangeStream,
+  ChangeStreamAggregateRawResult,
```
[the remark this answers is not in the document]
nbbeeken: It's not public it has to be exported to make it accessible from the top level import even though it's marked internal. I tried not exporting it but api-extractor still wants it exported probably has to do with how it's used as a generic on an imported function rather than entirely internal to the file like the TopologyOptions interface.

src/change_stream.ts
```diff
@@ -102,6 +113,18 @@ export interface PipeOptions {
   end?: boolean;
 }
 
+/** @internal */
+export type ChangeStreamAggregateRawResult<TChange> = {
```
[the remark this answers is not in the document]
nbbeeken: No we don't need to make this accessible to users, it's to provide a type for executeOperation so we can check more of the usage in change_stream.ts

src/change_stream.ts
```diff
@@ -102,6 +113,18 @@ export interface PipeOptions {
...
+  $clusterTime: { clusterTime: Timestamp };
+  cursor: {
+    postBatchResumeToken: unknown;
```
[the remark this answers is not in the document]
nbbeeken: Nope! good catch, fixed

test/unit/error.test.ts
```diff
@@ -546,6 +548,56 @@ describe('MongoErrors', () => {
...
     });
 
+    it('should return false for Error with code', () => {
+      const errorWithCode = new (class extends Error {
+        code = MONGODB_ERROR_CODES.CursorNotFound;
+      })();
+      expect(isResumableError(errorWithCode)).to.be.false;
+    });
+
+    it('should return false nullish arguments', () => {
```
dariakp, line 558:
```suggestion
    it('should return false for nullish error argument regardless of wire version', () => {
```
and the assertion in line 561 should pass in and 8 and a 9 instead of null for the second argument as well as test undefined for the error arg

test/unit/error.test.ts
```diff
@@ -546,6 +548,56 @@ describe('MongoErrors', () => {
       expect(isResumableError(new MongoNetworkError('ah!'))).to.be.true;
     });
 
+    it('should return false for Error with code', () => {
```
dariakp, line 551: this should be something like "should return false for an error that is not a MongoError regardless of code property or wireversion" and there should be four test errors: a plain error with and without wire version argument and a code error with and without wire version (where the with wire version case uses `9` as the version, but can also use `8` as an extra case if desired)

test/unit/error.test.ts
```diff
@@ -546,6 +548,56 @@ describe('MongoErrors', () => {
...
+      expect(isResumableError(errorWithCode)).to.be.false;
+    });
+
+    it('should return false nullish arguments', () => {
+      expect(isResumableError()).to.be.false;
+      expect(isResumableError(null)).to.be.false;
+      expect(isResumableError(null, null)).to.be.false;
+    });
+
+    it('should return true for labelless MongoError with CursorNotFound code', () => {
```
dariakp, line 564:
```suggestion
    it('should return true for labelless MongoError with CursorNotFound code regardless of wire version', () => {
```

test/unit/error.test.ts
```diff
@@ -546,6 +548,56 @@ describe('MongoErrors', () => {
...
+      const mongoError = new MongoError('ah!');
+      mongoError.code = MONGODB_ERROR_CODES.CursorNotFound;
+      expect(isResumableError(mongoError)).to.be.true;
+      expect(isResumableError(mongoError, 9)).to.be.true;
+      expect(isResumableError(mongoError, 8)).to.be.true;
+    });
+
+    it('should return true for labeled MongoError above wireVersion 9', () => {
```
dariakp, line 572: if you don't use a separate it block for labeled errors that don't have a wire version above 9, then you should be explicit about the test case wording as well as group the false assertions together separately from the true assertions

```suggestion
    it('should return true for labeled MongoError only if the wireVersion is at least 9', () => {
```

baileympearson (approved):

test/unit/error.test.ts
```diff
@@ -546,6 +548,56 @@ describe('MongoErrors', () => {
...
+      expect(isResumableError(mongoError, 8)).to.be.false;
+    });
+
+    it('should return false for resumable codes if wireVersion is above 9', () => {
+      const mongoError = new MongoError('ah!');
+      mongoError.code = MONGODB_ERROR_CODES.ShutdownInProgress; // Shutdown in progress is resumable
+      expect(isResumableError(mongoError, 9)).to.be.false; // 4.4+ uses label only except for CursorNotFound
+    });
+
+    it('should return true for resumable codes if wireVersion is below 9', () => {
```
dariakp, line 586:
```suggestion
    it('should return true for resumable codes if wireVersion is below 9 or unspecified', () => {
```

test/unit/error.test.ts
```diff
@@ -546,6 +548,56 @@ describe('MongoErrors', () => {
...
+
+    it('should return true for labeled MongoError above wireVersion 9', () => {
+      const mongoError = new MongoError('ah!');
+      mongoError.addErrorLabel(MongoErrorLabel.ResumableChangeStreamError);
+      expect(isResumableError(mongoError)).to.be.false;
+      expect(isResumableError(mongoError, 9)).to.be.true;
+      expect(isResumableError(mongoError, 8)).to.be.false;
+    });
+
+    it('should return false for resumable codes if wireVersion is above 9', () => {
```
dariakp, line 580: use "above" for `>`, and "at least" for `>=`
```suggestion
    it('should return false for resumable codes if wireVersion is at least 9', () => {
```

test/unit/error.test.ts
```diff
@@ -546,6 +548,56 @@ describe('MongoErrors', () => {
...
+    });
+
+    it('should return true for resumable codes if wireVersion is below 9', () => {
+      const mongoError = new MongoError('ah!');
+      mongoError.code = MONGODB_ERROR_CODES.ShutdownInProgress; // Shutdown in progress is resumable
+      expect(isResumableError(mongoError)).to.be.true;
+      expect(isResumableError(mongoError, 8)).to.be.true;
+    });
+
+    it('should return false for non numeric code', () => {
```
dariakp, line 593:
```suggestion
    it('should return false for non numeric code regardless of wire version', () => {
```

dariakp (changes requested):
test just need a little cleanup, otherwise lgtm

test/unit/error.test.ts
nbbeeken: Added to this, I made the rules about how the function handles a plain error more strict, throwing on code or hasErrorLabel usage.

test/unit/error.test.ts
```diff
@@ -534,4 +535,17 @@ describe('MongoErrors', () => {
       }
     });
   });
+
+  describe('isResumableError()', () => {
+    it('should return false for errors that are not MongoError', () => {
```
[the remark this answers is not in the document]
dariakp: This test can probably be deleted now, but I just noticed that we should probably add the "regardless of wire version" wording and expectations to the `MongoNetworkError` below, too

test/unit/error.test.ts
```diff
@@ -534,4 +537,83 @@ describe('MongoErrors', () => {
...
+      // Expectations that prove this syntax provides what is desired for the test
+      expect(errorWithCode).to.be.instanceOf(Error);
+      expect(errorWithCode).to.not.be.instanceOf(MongoError);
+      // Testing that even coded and labeled non-MongoErrors will not get resumed
+      expect(isResumableError(errorWithCode)).to.be.false;
+      expect(isResumableError(errorWithCode, 8)).to.be.false;
+      expect(isResumableError(errorWithCode, 9)).to.be.false;
+    });
+
+    it('should return false for nullish error argument regardless of wire version', () => {
```
dariakp, line 574: my last comment seems to have gotten lost on this one, but the test in L577 should actually check the wire version, so use the 8 and 9, and also should test for undefined

test/unit/error.test.ts
```diff
@@ -534,4 +537,83 @@ describe('MongoErrors', () => {
...
+
+    it('should return true for labelless MongoError with CursorNotFound code regardless of wire version', () => {
+      const mongoError = new MongoError('ah!');
+      mongoError.code = MONGODB_ERROR_CODES.CursorNotFound;
+      expect(isResumableError(mongoError)).to.be.true;
+      expect(isResumableError(mongoError, 9)).to.be.true;
+      expect(isResumableError(mongoError, 8)).to.be.true;
+    });
+
+    it('should return true for labeled MongoError only if the wireVersion is at least 9', () => {
```
dariakp, line 588: can we group the false assertions separately from the true ones here since it's kind of two test cases in one?

dariakp (changes requested):

dariakp (approved):

baileympearson (approved):

PR #3218, opened 2022-04-26
feat(NODE-4202): add FLE 2 behavior for create/drop collection

addaleax:
### Description

See NODE-4202 / https://github.com/mongodb/specifications/commit/0142955943d91a783fc0314fb41a14e392bd15b6

#### What is changing?

Collection creation and dropping are extended to account for FLE2 properties according to spec.

### DIFF
--- a/src/encrypter.ts
+++ b/src/encrypter.ts
@@ -72,7 +72,10 @@ export class Encrypter {
     if (internalClient == null) {
       const clonedOptions: MongoClientOptions = {};
 
-      for (const key of Object.keys(options)) {
+      for (const key of [
+        ...Object.getOwnPropertyNames(options),
+        ...Object.getOwnPropertySymbols(options)
+      ] as string[]) {
         if (['autoEncryption', 'minPoolSize', 'servers', 'caseTranslate', 'dbName'].includes(key))
           continue;
         Reflect.set(clonedOptions, key, Reflect.get(options, key));
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -118,6 +118,10 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
           const createOp = new CreateCollectionOperation(db, collectionName);
           await createOp.executeWithoutEncryptedFieldsCheck(server, session);
         }
+
+        if (!options.encryptedFields) {
+          this.options = { ...this.options, encryptedFields };
+        }
       }
 
       const coll = await this.executeWithoutEncryptedFieldsCheck(server, session);
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -110,9 +110,9 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
 
       if (encryptedFields) {
         // Create auxilliary collections for FLE2 support.
-        const escCollection = encryptedFields.escCollection || `enxcol_.${name}.esc`;
-        const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
-        const ecocCollection = encryptedFields.ecocCollection || `enxcol_.${name}.ecoc`;
+        const escCollection = encryptedFields.escCollection ?? `enxcol_.${name}.esc`;
+        const eccCollection = encryptedFields.eccCollection ?? `enxcol_.${name}.ecc`;
+        const ecocCollection = encryptedFields.ecocCollection ?? `enxcol_.${name}.ecoc`;
 
         for (const collectionName of [escCollection, eccCollection, ecocCollection]) {
           const createOp = new CreateCollectionOperation(db, collectionName);
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -6,6 +6,7 @@ import type { Server } from '../sdam/server';
 import type { ClientSession } from '../sessions';
 import type { Callback } from '../utils';
 import { CommandOperation, CommandOperationOptions } from './command';
+import { CreateIndexOperation } from './indexes';
 import { Aspect, defineAspects } from './operation';
 
 const ILLEGAL_COMMAND_FIELDS = new Set([
@@ -75,6 +76,8 @@ export interface CreateCollectionOptions extends CommandOperationOptions {
   timeseries?: TimeSeriesCollectionOptions;
   /** The number of seconds after which a document in a timeseries collection expires. */
   expireAfterSeconds?: number;
+  /** @experimental */
+  encryptedFields?: Document;
 }
 
 /** @internal */
@@ -92,6 +95,74 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
   }
 
   override execute(
+    server: Server,
+    session: ClientSession | undefined,
+    callback: Callback<Collection>
+  ): void {
+    const db = this.db;
+    const name = this.name;
+    const options = this.options;
+
+    const encryptedFields =
+      options.encryptedFields ??
+      db.s.client.options.autoEncryption?.encryptedFieldsMap?.[`${db.databaseName}.${name}`];
+    if (!encryptedFields) {
+      return this.executeWithoutEncryptedFieldCheck(server, session, callback);
+    }
+
+    const escCollection = encryptedFields.escCollection || `enxcol_.${name}.esc`;
+    const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
+    const ecocCollection = encryptedFields.ecocCollection || `enxcol_.${name}.ecoc`;
+    new CreateCollectionOperation(db, escCollection).executeWithoutEncryptedFieldCheck(
+      server,
+      session,
+      err => {
+        if (err) {
+          return callback(err);
+        }
+
+        new CreateCollectionOperation(db, eccCollection).executeWithoutEncryptedFieldCheck(
+          server,
+          session,
+          err => {
+            if (err) {
+              return callback(err);
+            }
+
+            new CreateCollectionOperation(db, ecocCollection).executeWithoutEncryptedFieldCheck(
+              server,
+              session,
+              err => {
+                if (err) {
+                  return callback(err);
+                }
+
+                this.executeWithoutEncryptedFieldCheck(server, session, (err, coll) => {
+                  if (err) {
+                    return callback(err);
+                  }
+
+                  new CreateIndexOperation(db, name, { __safeContent__: 1 }, {}).execute(
+                    server,
+                    session,
+                    err => {
+                      if (err) {
+                        return callback(err);
+                      }
+
+                      return callback(undefined, coll);
+                    }
+                  );
+                });
+              }
+            );
+          }
+        );
+      }
+    );
+  }
+
+  private executeWithoutEncryptedFieldCheck(
     server: Server,
     session: ClientSession | undefined,
     callback: Callback<Collection>
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -124,7 +124,12 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
 
       if (encryptedFields) {
         // Create the required index for FLE2 support.
-        const createIndexOp = new CreateIndexOperation(db, name, { __safeContent__: 1 }, {});
+        const createIndexOp = new CreateIndexOperation(
+          db,
+          name,
+          { __safeContent__: 1 },
+          { unique: true }
+        );
         await new Promise<void>((resolve, reject) => {
           createIndexOp.execute(server, session, err => (err ? reject(err) : resolve()));
         });
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -99,99 +99,75 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
     session: ClientSession | undefined,
     callback: Callback<Collection>
   ): void {
-    const db = this.db;
-    const name = this.name;
-    const options = this.options;
+    (async () => {
+      const db = this.db;
+      const name = this.name;
+      const options = this.options;
 
-    const encryptedFields =
-      options.encryptedFields ??
-      db.s.client.options.autoEncryption?.encryptedFieldsMap?.[`${db.databaseName}.${name}`];
-    if (!encryptedFields) {
-      return this.executeWithoutEncryptedFieldCheck(server, session, callback);
-    }
+      const encryptedFields: Document | undefined =
+        options.encryptedFields ??
+        db.s.client.options.autoEncryption?.encryptedFieldsMap?.[`${db.databaseName}.${name}`];
 
-    const escCollection = encryptedFields.escCollection || `enxcol_.${name}.esc`;
-    const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
-    const ecocCollection = encryptedFields.ecocCollection || `enxcol_.${name}.ecoc`;
-    new CreateCollectionOperation(db, escCollection).executeWithoutEncryptedFieldCheck(
-      server,
-      session,
-      err => {
-        if (err) {
-          return callback(err);
+      if (encryptedFields) {
+        // Create auxilliary collections for FLE2 support.
+        const escCollection = encryptedFields.escCollection || `enxcol_.${name}.esc`;
+        const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
+        const ecocCollection = encryptedFields.ecocCollection || `enxcol_.${name}.ecoc`;
+
+        for (const collectionName of [escCollection, eccCollection, ecocCollection]) {
+          const createOp = new CreateCollectionOperation(db, collectionName);
+          await createOp.executeWithoutEncryptedFieldsCheck(server, session);
         }
-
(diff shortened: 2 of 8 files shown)


src/operations/create_collection.ts
```diff
@@ -100,6 +103,74 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
...
+                      return callback(undefined, coll);
+                    }
+                  );
+                });
+              }
+            );
+          }
+        );
+      }
+    );
```
addaleax, line 162: Is it okay/preferred to perform these operations as part of another `CommandOperation` object, or would it make more sense to move this into the `db.createCollection()` wrapper that calls `executeOperation()`?

The latter a bit unappealing to me because it would mean taking care of the promise/callback handling manually, but as far as I understand, there are concrete differences between the two approaches in that `executeOperation()` performs server selection, so there is probably a “correct” choice. Please tell me if I made the wrong one :slightly_smiling_face:

test/tools/spec-runner/index.js
```diff
@@ -636,16 +649,22 @@ const kOperations = new Map([
...
     'dropCollection',
     (operation, db, context /*, options */) => {
       const collectionName = operation.arguments.collection;
+      const encryptedFields = operation.arguments.encryptedFields;
       const session = maybeSession(operation, context);
-      return db.dropCollection(collectionName, { session });
+      return db.dropCollection(collectionName, { session, encryptedFields }).catch(err => {
+        if (err.code !== MONGODB_ERROR_CODES.NamespaceNotFound) {
+          throw err;
+        }
```
addaleax, line 666: The added spec tests are pretty clear about the fact that dropping a non-existent collection would not be expected to make the test fail. Let me know if this isn’t a good way to implement that behavior.

test/tools/spec-runner/index.js
```diff
@@ -666,6 +685,74 @@ const kOperations = new Map([
...
+          .listIndexes({ session })
+          .toArray()
+          .then(results => results.map(({ name }) => name))
+          .then(indexes => expect(indexes).to.not.include(indexName))
+      ).catch(err => {
+        if (err.code !== MONGODB_ERROR_CODES.NamespaceNotFound) {
+          throw err;
+        }
+      });
+    }
```
addaleax, line 755: I saw that these are part of the unified test runner – is this okay to add here like this? Maybe porting the CSFLE tests over isn’t that hard, but it felt like that would be out of scope here.

test/tools/spec-runner/index.js
```diff
@@ -774,6 +861,9 @@ function testOperation(operation, obj, context, options) {
       const cursor = obj[operationName].apply(obj, args);
       opPromise = cursor.toArray();
     } else {
+      if (!obj[operationName]) {
+        throw new Error(`Unknown operation "${operationName}"`);
+      }
```
addaleax, line 866: This was too useful while making these changes to not leave it in the PR, even though I assume that this is legacy code

durran:
@addaleax Notes from our chat. Let's dry up the repeating calls to create collection and drop collection so we don't have the same code running for each fle collection. For the additional pings in the test runner we can add an option to the test runner's `MongoClient`: `[Symbol.for('@@mdb.skipPingOnConnect')]: true` that will skip the ping.

addaleax:
> Let's dry up the repeating calls to create collection and drop collection so we don't have the same code running for each fle collection.

@durran After thinking about using something like promisify/callbackify or using a custom async.series() equivalent, I’ve kind of come to the conclusion that it would really be easiest to just write async/await code here. I feel like that should be an acceptable way of doing things here.

> For the additional pings in the test runner we can add an option to the test runner's `MongoClient`: `[Symbol.for('@@mdb.skipPingOnConnect')]: true` that will skip the ping.

This option is actually already set also for the legacy test runner:

https://github.com/mongodb/node-mongodb-native/blob/441fc63e8b97a8e2d06019759e8994d5f2f85b89/test/tools/spec-runner/index.js#L349

I’ll try to figure out why this is happening (unless you happen to know)

src/encrypter.ts
```diff
@@ -72,7 +72,10 @@ export class Encrypter {
     if (internalClient == null) {
       const clonedOptions: MongoClientOptions = {};
 
-      for (const key of Object.keys(options)) {
+      for (const key of [
+        ...Object.getOwnPropertyNames(options),
+        ...Object.getOwnPropertySymbols(options)
+      ] as string[]) {
```
addaleax, line 78: This is for passing through `[Symbol.for('@@mdb.skipPingOnConnect')]: true` to the AutoEncrypter internal client 🙂

src/operations/create_collection.ts
durran: Per our discussion we thought it better to do this in the operation object so server selection would ensure that all the operations execute on the same server instead of server selection getting run for each individual command.

src/operations/create_collection.ts
```diff
@@ -96,31 +99,75 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
...
+        options.encryptedFields ??
+        db.s.client.options.autoEncryption?.encryptedFieldsMap?.[`${db.databaseName}.${name}`];
+
+      if (encryptedFields) {
+        // Create auxilliary collections for FLE2 support.
+        const escCollection = encryptedFields.escCollection || `enxcol_.${name}.esc`;
+        const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
+        const ecocCollection = encryptedFields.ecocCollection || `enxcol_.${name}.ecoc`;
+
+        for (const collectionName of [escCollection, eccCollection, ecocCollection]) {
```
durran, line 117: Nice refactor - I think this implementation is far cleaner and easier to read than when it was callbacks.

test/tools/spec-runner/index.js
```diff
@@ -597,6 +598,18 @@ function maybeSession(operation, context) {
   );
 }
 
+function withoutCommandMonitoring(client, fn) {
```
durran, line 601: Now that the option to ignore the ping is passed through, do we need to do this anymore?

src/operations/drop.ts
```diff
@@ -24,10 +31,81 @@ export class DropCollectionOperation extends CommandOperation<boolean> {
...
+      let encryptedFields: Document | undefined =
+        options.encryptedFields ?? encryptedFieldsMap?.[`${db.databaseName}.${name}`];
+
+      if (!encryptedFields && encryptedFieldsMap) {
+        // If the MongoClient was configued with an encryptedFieldsMap,
+        // and no encryptedFields config was available in it or explicitly
+        // passed as an argument, the spec tells us to look one up using
+        // listCollections().
+        const listCollectionsResult = await db
+          .listCollections({ name }, { nameOnly: false })
```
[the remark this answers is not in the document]
addaleax: It is – I wanted to be explicit here, since we are specifically using `listCollections` to look up a field other than name/type. Happy to adjust this, of course :slightly_smiling_face:

test/tools/spec-runner/index.js
addaleax: Yeah, this is for the `listCollections`/`listIndexes` calls that are part of the `assert(Index|Collection)(Not)?Exists`. I admittedly don’t quite know if and how the unified test runner handles this.

src/operations/drop.ts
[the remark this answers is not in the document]
durran: I'm not too bothered by it being explicitly called out here, just a question.

test/tools/spec-runner/index.js
durran: The unified runner handles it by doing the assertions on the utility client and not the client under test - so those events aren't monitored.

test/tools/spec-runner/index.js
durran: I think checking the code is fine - it's been there before 3.6.

durran (approved):

test/tools/spec-runner/index.js
addaleax: Yeah, that makes sense – the runner here also has a `utilClient`, but it closes it again before the test is run, so I figure that’s not really a viable option here?

durran: Yeah I think it's fine to do it as is.

src/operations/create_collection.ts
```diff
@@ -96,31 +99,75 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
     session: ClientSession | undefined,
     callback: Callback<Collection>
   ): void {
-    const db = this.db;
-    const name = this.name;
-    const options = this.options;
+    (async () => {
```
nbbeeken, line 102: Unfortunately we can't use async/await just yet in the driver since we support providing third party promise libs (I know..), I'm sorry I think we have to revert to the callback style. 

@durran am I missing something that makes this API exempt? I looked into FLE and it has its own promise-if-no-callback logic that will use the global promise but I don't think the driver will hit that since it will always pass in a callback to the library

src/operations/create_collection.ts
```diff
@@ -96,31 +99,75 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
...
-        return callback(err);
+      const encryptedFields: Document | undefined =
+        options.encryptedFields ??
+        db.s.client.options.autoEncryption?.encryptedFieldsMap?.[`${db.databaseName}.${name}`];
+
+      if (encryptedFields) {
+        // Create auxilliary collections for FLE2 support.
+        const escCollection = encryptedFields.escCollection || `enxcol_.${name}.esc`;
+        const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
+        const ecocCollection = encryptedFields.ecocCollection || `enxcol_.${name}.ecoc`;
```
nbbeeken, line 115:
```suggestion
        const escCollection = encryptedFields.escCollection ?? `enxcol_.${name}.esc`;
        const eccCollection = encryptedFields.eccCollection ?? `enxcol_.${name}.ecc`;
        const ecocCollection = encryptedFields.ecocCollection ?? `enxcol_.${name}.ecoc`;
```

src/operations/drop.ts
```diff
@@ -24,10 +31,81 @@ export class DropCollectionOperation extends CommandOperation<boolean> {
...
+      }
+
+      let result;
+      let errorForMainOperation;
+      try {
+        result = await this.executeWithoutEncryptedFieldsCheck(server, session);
+      } catch (err) {
+        if (
+          !encryptedFields ||
+          (err as MongoServerError).code !== MONGODB_ERROR_CODES.NamespaceNotFound
```
nbbeeken, line 61:
```suggestion
          (err instanceof MongoServerError && err.code !== MONGODB_ERROR_CODES.NamespaceNotFound)
```

src/operations/drop.ts
```diff
@@ -24,10 +31,81 @@ export class DropCollectionOperation extends CommandOperation<boolean> {
...
+        const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
+        const ecocCollection = encryptedFields.ecocCollection || `enxcol_.${name}.ecoc`;
+
+        for (const collectionName of [escCollection, eccCollection, ecocCollection]) {
+          // Drop auxilliary collections, ignoring potential NamespaceNotFound errors.
+          const dropOp = new DropCollectionOperation(db, collectionName);
+          try {
+            await dropOp.executeWithoutEncryptedFieldsCheck(server, session);
+          } catch (err) {
+            if ((err as MongoServerError).code !== MONGODB_ERROR_CODES.NamespaceNotFound) {
```
nbbeeken, line 82:
```suggestion
            if ((err instanceof MongoServerError && err.code !== MONGODB_ERROR_CODES.NamespaceNotFound)) {
```

nbbeeken (changes requested):

src/operations/create_collection.ts
addaleax: @nbbeeken But aren’t third-party promise libs only relevant for the public API? This async function is not externally visible, and even if a client provides a different promise library, they can’t expect that other libraries (or even Node.js itself) don’t use async/await at least *internally*, right?

nbbeeken: This is a really fair point 🤔  I also thought we could consider that new APIs don't have to abide by the deprecated option. We actually already have a case where snappy 7+ will use native promises, and we callbackify the API.  I'm onboard then. This would be the first time we ship async await in the driver, so lemme get the rest of the team's attention. 2017 here we come :) 

cc: @dariakp @baileympearson

nbbeeken (approved):
LGTM

src/operations/create_collection.ts
```diff
@@ -96,31 +99,75 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
...
+          const createOp = new CreateCollectionOperation(db, collectionName);
+          await createOp.executeWithoutEncryptedFieldsCheck(server, session);
+        }
+      }
+
+      const coll = await this.executeWithoutEncryptedFieldsCheck(server, session);
+
+      if (encryptedFields) {
+        // Create the required index for FLE2 support.
+        const createIndexOp = new CreateIndexOperation(db, name, { __safeContent__: 1 }, {});
```
addaleax, line 127: Heads up: This may need to become a unique index to avoid easy-to-make user errors based on the slack conversation in https://mongodb.slack.com/archives/C0316KB5VM4/p1651503928461139?thread_ts=1651261523.846079&cid=C0316KB5VM4

nbbeeken: Should we do this now or wait until it is spec'ed. I think it's pretty much assured we need to set unique. Is there a ticket to block on?

nbbeeken (changes requested):

src/operations/create_collection.ts
```diff
@@ -96,31 +99,80 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
...
+          const createOp = new CreateCollectionOperation(db, collectionName);
+          await createOp.executeWithoutEncryptedFieldsCheck(server, session);
+        }
+      }
+
+      const coll = await this.executeWithoutEncryptedFieldsCheck(server, session);
+
+      if (encryptedFields) {
+        // Create the required index for FLE2 support.
+        const createIndexOp = new CreateIndexOperation(
```
durran, line 127: I added the unique option here.

src/operations/create_collection.ts
addaleax: @nbbeeken @durran … and we decided not to do this since it would not work on sharded setups :slightly_smiling_face: So let’s stick to the spec here for now. We’ll probably have an ongoing conversation about this since I do think it’s a footgun, but this probably isn’t the solution.

addaleax: I thought so too! But yeah, makes sense that this doesn’t work :)

src/operations/drop.ts
```diff
@@ -24,10 +31,84 @@ export class DropCollectionOperation extends CommandOperation<boolean> {
...
+        const escCollection = encryptedFields.escCollection || `enxcol_.${name}.esc`;
+        const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
+        const ecocCollection = encryptedFields.ecocCollection || `enxcol_.${name}.ecoc`;
+
+        for (const collectionName of [escCollection, eccCollection, ecocCollection]) {
+          // Drop auxilliary collections, ignoring potential NamespaceNotFound errors.
+          const dropOp = new DropCollectionOperation(db, collectionName);
+          try {
+            await dropOp.executeWithoutEncryptedFieldsCheck(server, session);
+          } catch (err) {
```
dariakp, line 81: are we intentionally swallowing any errors that might not be instances of MongoServerError?

dariakp (commented):
just one question here

dariakp (commented):
just one question here

src/operations/drop.ts
addaleax: ah yeah no, thanks for noticing – messed this up during code review in 75c02b70ee573232737dbc727c998c628ad1cc0a, is fixed now :slightly_smiling_face:

src/operations/create_collection.ts
```diff
@@ -118,6 +118,10 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
           const createOp = new CreateCollectionOperation(db, collectionName);
           await createOp.executeWithoutEncryptedFieldsCheck(server, session);
         }
+
+        if (!options.encryptedFields) {
+          this.options = { ...this.options, encryptedFields };
+        }
```
addaleax, line 124: Sorry, added this very late but I think this is a definite omission in the spec tests – opened https://github.com/mongodb/specifications/pull/1205 so this doesn’t happen to anybody else

addaleax: And now the spec tests are failing because the `promoteValues` fix from https://github.com/mongodb/libmongocrypt/pull/324 is missing – I think that’s okay for me, unless it’s important that these tests here are skipped until that PR is merged & released

baileympearson (approved):
Per Neal's direction, I only looked at the changes in create_collection.ts and the async await code.  👍  for a good idea here, LGTM

nbbeeken (approved):

src/operations/create_collection.ts
dariakp: I think that's ok

dariakp (approved):

PR #3219, opened 2022-04-26
feat(NODE-2992): consider server load during server selection

baileympearson:
#### What is changing?

This PR updates server selection to consider the number of operations running against a particular server.

- a new property was added to the server class to keep track of the current number of operations running against a particular server
- server selection was updated randomly select 2 servers from the latency window, and select the server with the smaller operation count
- A new prose test was added to ensure that if a server is slowing down, requests are routed to the healthy server
- tests were added to the server class to ensure that the count is decremented appropriately for each operation
- a new spec test runner was added to run the latency window spec tests, and the latency window spec tests were synced.

No.

This is a part of the [avoiding connection storms](https://jira.mongodb.org/browse/DRIVERS-781) epic.  Keeping track of the current server load will allow server selection to choose servers with fewer request being sent to them and help prevent connection storms.

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -15,7 +15,7 @@ declare global {
       apiVersion?: '1';
       clientSideEncryption?: boolean;
       serverless?: 'forbid' | 'allow' | 'require';
-      auth: 'enabled' | 'disabled';
+      auth?: 'enabled' | 'disabled';
     };
 
     sessions?: {
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -388,7 +388,7 @@ export abstract class AbstractCursor<
     });
   }
 
-  close(): void;
+  close(): Promise<void>;
   close(callback: Callback): void;
   /**
    * @deprecated options argument is deprecated
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -333,6 +333,10 @@ export class Server extends TypedEventEmitter<ServerEvents> {
     //       attempt to check out a connection from the pool.  This ensures that operations that
     //       are waiting for a connection are included in the operation count.  Load balanced
     //       mode will only ever have a single server, so the operation count doesn't matter.
+    //       Incrementing the operation count above the logic to handle load balanced mode would
+    //       require special logic to decrement it again, or would double increment (the load
+    //       balanced code makes a recursive call).  Instead, we increment the count after this
+    //       check.
     if (this.loadBalanced && session && conn == null && isPinnableCommand(cmd, session)) {
       this.s.pool.checkOut((err, checkedOut) => {
         if (err || checkedOut == null) {
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -93,6 +93,8 @@ export interface ServerPrivate {
   pool: ConnectionPool;
   /** MongoDB server API version */
   serverApi?: ServerApi;
+  /** A count of the operations currently running against the server. */
+  operationCount: number;
 }
 
 /** @public */
@@ -147,7 +149,8 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       logger: new Logger('Server'),
       state: STATE_CLOSED,
       topology,
-      pool: new ConnectionPool(poolOptions)
+      pool: new ConnectionPool(poolOptions),
+      operationCount: 0
     };
 
     for (const event of [...CMAP_EVENTS, ...APM_EVENTS]) {
@@ -335,14 +338,16 @@ export class Server extends TypedEventEmitter<ServerEvents> {
         session.pin(checkedOut);
         this.command(ns, cmd, finalOptions, callback);
       });
-
       return;
     }
+
+    this.s.operationCount += 1;
 
     this.s.pool.withConnection(
       conn,
       (err, conn, cb) => {
         if (err || !conn) {
+          this.s.operationCount -= 1;
           markServerUnknown(this, err);
           return cb(err);
         }
@@ -351,7 +356,10 @@ export class Server extends TypedEventEmitter<ServerEvents> {
           ns,
           cmd,
           finalOptions,
-          makeOperationHandler(this, conn, cmd, finalOptions, cb)
+          makeOperationHandler(this, conn, cmd, finalOptions, (error, response) => {
+            this.s.operationCount -= 1;
+            cb(error, response);
+          })
         );
       },
       callback
@@ -373,15 +381,26 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       return;
     }
 
+    this.s.operationCount += 1;
+
     this.s.pool.withConnection(
       options.session?.pinnedConnection,
       (err, conn, cb) => {
         if (err || !conn) {
+          this.s.operationCount -= 1;
           markServerUnknown(this, err);
           return cb(err);
         }
 
-        conn.getMore(ns, cursorId, options, makeOperationHandler(this, conn, {}, options, cb));
+        conn.getMore(
+          ns,
+          cursorId,
+          options,
+          makeOperationHandler(this, conn, {}, options, (error, response) => {
+            this.s.operationCount -= 1;
+            cb(error, response);
+          })
+        );
       },
       callback
     );
@@ -405,10 +424,12 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       return;
     }
 
+    this.s.operationCount += 1;
     this.s.pool.withConnection(
       options.session?.pinnedConnection,
       (err, conn, cb) => {
         if (err || !conn) {
+          this.s.operationCount -= 1;
           markServerUnknown(this, err);
           return cb(err);
         }
@@ -417,7 +438,10 @@ export class Server extends TypedEventEmitter<ServerEvents> {
           ns,
           cursorIds,
           options,
-          makeOperationHandler(this, conn, {}, undefined, cb)
+          makeOperationHandler(this, conn, {}, undefined, (error, response) => {
+            this.s.operationCount -= 1;
+            cb(error, response);
+          })
         );
       },
       callback
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -328,6 +328,11 @@ export class Server extends TypedEventEmitter<ServerEvents> {
     // NOTE: This is a hack! We can't retrieve the connections used for executing an operation
     //       (and prevent them from being checked back in) at the point of operation execution.
     //       This should be considered as part of the work for NODE-2882
+    // NOTE:
+    //       When incrementing operation count, it's important that we increment it before we
+    //       attempt to check out a connection from the pool.  This ensures that operations that
+    //       are waiting for a connection are included in the operation count.  Load balanced
+    //       mode will only ever have a single server, so the operation count doesn't matter.
     if (this.loadBalanced && session && conn == null && isPinnableCommand(cmd, session)) {
       this.s.pool.checkOut((err, checkedOut) => {
         if (err || checkedOut == null) {
--- a/src/sdam/topology.ts
+++ b/src/sdam/topology.ts
@@ -843,24 +843,6 @@ function topologyTypeFromOptions(options?: TopologyOptions) {
   return TopologyType.Unknown;
 }
 
-function randomSelection(array: ServerDescription[]): {
-  description1: ServerDescription;
-  description2: ServerDescription;
-} {
-  const servers = shuffle(array);
-  if (servers.length === 1) {
-    return {
-      description1: servers[0],
-      description2: servers[0]
-    };
-  }
-
-  return {
-    description1: servers[0],
-    description2: servers[1]
-  };
-}
-
 /**
  * Creates new server instances and attempts to connect them
  *
@@ -982,12 +964,19 @@ function processWaitQueue(topology: Topology) {
       continue;
     }
 
-    const { description1, description2 } = randomSelection(selectedDescriptions);
-    const server1 = topology.s.servers.get(description1.address);
-    const server2 = topology.s.servers.get(description2.address);
+    let selectedServer;
+    if (selectedDescriptions.length === 1) {
+      selectedServer = topology.s.servers.get(selectedDescriptions[0].address);
+    } else {
+      const descriptions = shuffle(selectedDescriptions, 2);
+      const server1 = topology.s.servers.get(descriptions[0].address);
+      const server2 = topology.s.servers.get(descriptions[1].address);
 
-    const selectedServer =
-      server1 && server2 && server1.s.operationCount < server2.s.operationCount ? server1 : server2;
+      selectedServer =
+        server1 && server2 && server1.s.operationCount < server2.s.operationCount
+          ? server1
+          : server2;
+    }
 
     const transaction = waitQueueMember.transaction;
     if (isSharded && transaction && transaction.isActive && selectedServer) {
--- a/src/sdam/topology.ts
+++ b/src/sdam/topology.ts
@@ -978,6 +978,15 @@ function processWaitQueue(topology: Topology) {
           : server2;
     }
 
+    if (!selectedServer) {
+      waitQueueMember.callback(
+        new MongoServerSelectionError(
+          'server selection returned a server description but the server was not found in the topology',
+          topology.description
+        )
+      );
+      return;
+    }
     const transaction = waitQueueMember.transaction;
     if (isSharded && transaction && transaction.isActive && selectedServer) {
       transaction.pinServer(selectedServer);
--- a/src/sdam/topology.ts
+++ b/src/sdam/topology.ts
@@ -843,8 +843,22 @@ function topologyTypeFromOptions(options?: TopologyOptions) {
   return TopologyType.Unknown;
 }
 
-function randomSelection(array: ServerDescription[]): ServerDescription {
-  return array[Math.floor(Math.random() * array.length)];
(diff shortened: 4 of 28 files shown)


global.d.ts
```diff
@@ -15,7 +15,7 @@ declare global {
       apiVersion?: '1';
       clientSideEncryption?: boolean;
       serverless?: 'forbid' | 'allow' | 'require';
-      auth: 'enabled' | 'disabled';
+      auth?: 'enabled' | 'disabled';
```
baileympearson, line 18: Also included in some of Neal's PRs.  If those PRs merge first, I'll rebase main to pull it in.

This fixes type errors caused by our custom overloads to Mocha's `it`.

src/cursor/abstract_cursor.ts
```diff
@@ -388,7 +388,7 @@ export abstract class AbstractCursor<
     });
   }
 
-  close(): void;
+  close(): Promise<void>;
```
baileympearson, line 391: I noticed this in some of my tests when calling close.  If we want to make this fix separately, I can break it out into a different PR.

src/sdam/server.ts
```diff
@@ -335,14 +338,16 @@ export class Server extends TypedEventEmitter<ServerEvents> {
         session.pin(checkedOut);
```
baileympearson, line 338: This is as far up as I can comment in this method.  Just above this comment (lines 328-340), we have logic to handle load balanced mode separately.  We don't need to consider the operation count for load balanced mode, because there is only ever one server to select when in load balanced mode.  the load balanced logic above does call `server.command` recursively, so in that recursive call, we will increment and decrement the operation count.  That's okay - it won't affect server selection.

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+
+import { Collection, MongoClient } from '../../../src';
+import { ConnectionPool } from '../../../src/cmap/connection_pool';
+
+const TEST_METADATA: MongoDBMetadataUI = {
```
baileympearson, line 7: Note that I didn't add specific tests to ensure that we actually are incrementing the operation count.  The only way I could think to test this would be

1. set a failpoint with a really long delay on the server
2. don't await the command when executing it
3. assert that the count is incremented
4. await the comment now
5. assert the count is zero again

This felt really brittle.  Definitely open to suggestions on better ways to test this.

Ultimately, I decided against writing tests for this scenario because the incrementing / decrementing behavior is tested in the new prose test.

baileympearson: I'm also adding test cases for the server.command method in load balanced mode but I don't think that needs to stop a primary review 😄

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+  requires: {
+    topology: 'single',
+    mongodb: '>=4.0.0'
+  }
+};
+
+describe('Server Selection', function () {
+  let client: MongoClient;
+  let collection: Collection;
```
nbbeeken, line 16:
```suggestion
  let collection: Collection<{ name: string }>;
```

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    await client.close();
+    client = undefined;
+  });
+
+  context('operationCount', function () {
+    context('operationCount is adjusted properly on successful operation', function () {
+      it('is zero after a successful command', TEST_METADATA, async function () {
+        const server = Array.from(client.topology.s.servers.values())[0];
+        expect(server.s.operationCount).to.equal(0);
+        const commandSpy = sinon.spy(server, 'command');
```
nbbeeken, line 36: After your spy has gotten the intel it needs lets make sure to restore it 🕵️‍♀️

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+        await cursor.close();
+
+        expect(killCursorsSpy.called).to.be.true;
+        expect(server.s.operationCount).to.equal(0);
+      });
+    });
+
+    context('operationCount is adjusted properly when operations fail', function () {
+      afterEach(async function () {
+        await client.db('admin').command({
```
nbbeeken, line 81: Kind of a preference but can you pull out the failpoints to something named for the intention, like this one is to disable. And use the FailPoint type annotation so future readers can hover for more info

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+          configureFailPoint: 'failCommand',
+          mode: 'off',
+          data: {
+            failCommands: ['insert', 'getMore', 'killCursors'] // TODO : fill this out,
```
nbbeeken, line 85: marking TODO

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+          }
+        });
+      });
+
+      it('is zero after a command fails', TEST_METADATA, async function () {
+        await client.db('admin').command({
+          configureFailPoint: 'failCommand',
+          mode: 'alwaysOn',
+          data: {
+            failCommands: ['insert'], // TODO : fill this out,
```
nbbeeken, line 95: marking TODO

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+      it('is zero after a getMore fails', TEST_METADATA, async function () {
+        const cursor = collection.find({}, { batchSize: 1 });
+        await cursor.next();
+
+        await client.db('admin').command({
+          configureFailPoint: 'failCommand',
+          mode: 'alwaysOn',
+          data: {
+            failCommands: ['getMore'], // TODO : fill this out,
```
nbbeeken, line 125: marking TODO

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+      it('is zero after a killCursors fails', TEST_METADATA, async function () {
+        const cursor = collection.find({}, { batchSize: 1 });
+        await cursor.next(); // initialize the cursor
+
+        await client.db('admin').command({
+          configureFailPoint: 'failCommand',
+          mode: 'alwaysOn',
+          data: {
+            failCommands: ['killCursors'], // TODO : fill this out,
```
nbbeeken, line 151: marking TODO

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+            errorCode: 80
+          }
+        });
+
+        const server = Array.from(client.topology.s.servers.values())[0];
+        expect(server.s.operationCount).to.equal(0);
+        const killCursorsSpy = sinon.spy(server, 'killCursors');
+
+        await cursor.close().catch(err => err);
+        // expect(error).to.exist;
```
nbbeeken, line 161: Did you mean to capture the error here?

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+            const server = Array.from(client.topology.s.servers.values())[0];
+            expect(server.s.operationCount).to.equal(0);
+
+            sinon.stub(ConnectionPool.prototype, 'checkOut').callsFake(function (cb) {
+              cb(new Error('unable to checkout connection'), undefined);
+            });
+            const killCursorsSpy = sinon.spy(server, 'killCursors');
+
+            await cursor.close().catch(err => err);
+            // expect(error).to.exist;
```
nbbeeken, line 239: Is there supposed to be an error here?

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+            expect(killCursorsSpy.called).to.be.true;
+            expect(server.s.operationCount).to.equal(0);
+          }
+        );
```
nbbeeken, line 244: I think a good additon would be a test that sets up a retryable error and if I'm not mistaken we might have to spy but the operationCount should go up/down/up/down right?

nbbeeken (changes requested):

src/sdam/topology.ts
```diff
@@ -968,8 +982,13 @@ function processWaitQueue(topology: Topology) {
       continue;
     }
 
-    const selectedServerDescription = randomSelection(selectedDescriptions);
-    const selectedServer = topology.s.servers.get(selectedServerDescription.address);
+    const { description1, description2 } = randomSelection(selectedDescriptions);
```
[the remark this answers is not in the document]
baileympearson: Yeah agreed!  It's not doing much here anyways.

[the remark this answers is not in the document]
baileympearson: I'll make the changes you suggested for the randomSelection when I inline the code too.

test/unit/assorted/server_selection_latency_window_utils.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  for (let i = 0; i < test.iterations; ++i) {
+    const server: Server = await promisify(topology.selectServer.bind(topology))(
+      ReadPreference.NEAREST,
+      {}
+    );
+    selectedServers.push(server);
+  }
+
+  expect(selectedServers).to.have.lengthOf(test.iterations);
```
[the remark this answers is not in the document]
baileympearson: It can't.  I'll remove it!

test/unit/assorted/server_selection_latency_window_utils.ts
```diff
@@ -0,0 +1,134 @@
...
+      ).to.be.false;
+    }
+  }
+}
+
+function calculateObservedFrequencies(
+  observedServers: ReadonlyArray<Server>,
+  iterations: number
+): FrequencyMap {
+  const actualResults: FrequencyMap = {};
```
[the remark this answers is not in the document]
baileympearson: Sure.  Done!

test/unit/assorted/server_selection_latency_window_utils.ts
```diff
@@ -0,0 +1,134 @@
...
+type Outcome = ServerSelectionLatencyWindowTest['outcome'];
+type FrequencyMap = Outcome['expected_frequencies'];
+
+export function loadLatencyWindowTests(directory: string) {
+  const files = readdirSync(directory).filter(fileName => fileName.includes('.json'));
+
+  const tests: ServerSelectionLatencyWindowTest[] = [];
+
+  for (const fileName of files) {
+    const path = join(directory, '/', fileName);
```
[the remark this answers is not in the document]
baileympearson: fixed!

test/unit/assorted/server_selection.spec.test.ts
```diff
@@ -29,3 +35,25 @@ describe('Server Selection Logic (spec)', function () {
     });
   }
 });
+
+describe('Server Selection Latency Window Tests (spec)', function () {
+  const selectionSpecDir = join(__dirname, '../../spec/server-selection/in_window');
```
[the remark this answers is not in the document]
baileympearson: Done!

test/unit/assorted/server_selection.spec.test.ts
```diff
@@ -29,3 +35,25 @@ describe('Server Selection Logic (spec)', function () {
...
+  const tests = loadLatencyWindowTests(selectionSpecDir);
+  let serverConnect;
```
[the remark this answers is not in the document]
baileympearson: Done!

test/integration/server-selection/server_selection.test.ts
baileympearson: Unless you feel strongly about this, I'm going to leave them inline.  I don't think that pulling them into variables increases readability substantially.  I will type them with the `FailPoint` interface though.

test/integration/server-selection/server_selection.test.ts
baileympearson: Discussed over slack and we won't add this case.

test/integration/server-selection/server_selection.test.ts
baileympearson: Eating my own words and making the change lol

src/sdam/topology.ts
```diff
@@ -968,8 +964,20 @@ function processWaitQueue(topology: Topology) {
       continue;
     }
 
-    const selectedServerDescription = randomSelection(selectedDescriptions);
-    const selectedServer = topology.s.servers.get(selectedServerDescription.address);
+    let selectedServer;
+    if (selectedDescriptions.length === 1) {
+      selectedServer = topology.s.servers.get(selectedDescriptions[0].address);
+    } else {
```
nbbeeken, line 970: Is it worth adding a special case for 2 servers here? 2 servers can be selected and compared, 3 or more have to be shuffled, right?

nbbeeken (dismissed):
LGTM

test/integration/server-selection/server_selection_operation_count.prose.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    await promisify(setTimeout)(250);
+  }
+
+  client.removeListener('connectionCreated', onConnectionCreated);
+
+  if (connectionCount !== POOL_SIZE * 2) {
+    throw new Error('Connection pool did not fill up');
+  }
+}
+const TEST_METADATA: MongoDBMetadataUI = { requires: { mongodb: '>=4.2.9', topology: 'sharded' } };
```
dariakp, line 44: Can we label this `Step 1: ...`?

test/unit/assorted/server_selection_latency_window_utils.ts
```diff
@@ -0,0 +1,150 @@
...
+    'Encountered an empty set of frequencies to assert on.  Is there something wrong with the test or the runner?'
+  ).to.have.length.greaterThan(0);
+  for (const [address, frequency] of Object.entries(expected_frequencies)) {
+    if (frequency === 0) {
+      expect(observed_frequencies).not.to.haveOwnProperty(address);
+    } else {
+      expect(observed_frequencies).to.haveOwnProperty(address).to.exist;
+      const actual_frequency = observed_frequencies[address];
+      const is_too_low = actual_frequency < frequency - tolerance;
+      expect(
```
dariakp, line 68: for the sake of readability here, since we aren't actually distinguishing too high and too low in the assertions, could we instead do ```if (isTooHigh || isTooLow) { expect.fail(`...tolerance error message...`) }```

test/unit/assorted/server_selection_latency_window_utils.ts
```diff
@@ -0,0 +1,150 @@
...
+    const contents = readFileSync(path, { encoding: 'utf-8' });
+    tests.push(EJSON.parse(contents) as ServerSelectionLatencyWindowTest);
+  }
+
+  return tests;
+}
+
+function compareResultsToExpected(
+  { tolerance, expected_frequencies }: Outcome,
+  observed_frequencies: FrequencyMap
```
dariakp, line 54: can we consistently use camelCase for variables that we define vs props we read from the input file? it's rather disorienting to read the mix and match. 
1. For example, the variable in this function is defined as `observed_frequencies`, but the function itself is called with `observedFrequencies`. 
2. I see we already convert the `expected_frequencies` to `expectedFrequencies` below in L56, but then reference it directly again in L61 instead of using the variable we defined.
3. actual_frequency, is_too_high, and is_too_low aren't read from the input, so there doesn't really seem to be a reason to use underscores there

dariakp (changes requested):

test/integration/server-selection/server_selection_operation_count.prose.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+describe('Server Selection Operation Count Prose', function () {
```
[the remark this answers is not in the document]
baileympearson: I'll update the test name (topical comment, given our discussion of testing conventions this morning)! 

As far as the second part goes - is the ask to rename the file `server_selection.prose.operation_count.test.ts`?  The test is already broken out into its own prose test file.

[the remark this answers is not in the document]
dariakp: Sorry for the confusion, I meant, is it even worth having it in its own file right now vs in a general `server_selection.prose.test.ts` - basically, I was saying that it might be fine to just keep it as `server_selection.prose.test.ts` for now and not try to give it its own special name. But, I'm not strongly opposed to leaving it in a dedicated file as long as we follow the .prose.<specific test>.test convention.

src/sdam/topology.ts
```diff
@@ -963,13 +959,25 @@ function processWaitQueue(topology: Topology) {
...
+      // don't shuffle the array if there are only two elements
+      const descriptions =
+        selectedDescriptions.length === 2 ? selectedDescriptions : shuffle(selectedDescriptions, 2);
+      const server1 = topology.s.servers.get(descriptions[0].address);
+      const server2 = topology.s.servers.get(descriptions[1].address);
+
+      selectedServer =
+        server1 && server2 && server1.s.operationCount < server2.s.operationCount
```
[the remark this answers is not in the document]
baileympearson: There shouldn't be - all the code from 935 to here is synchronous, so the nothing should update the servers in the topology.

I'll make this change.

nbbeeken (dismissed):

nbbeeken (dismissed):

nbbeeken (approved):

dariakp (approved):
LGTM, thanks!

PR #3221, opened 2022-04-27
fix(NODE-3565): Error for insertMany with partially empty array is unhelpful

kampofo:
### Description
[NODE-3565](https://jira.mongodb.org/browse/NODE-3565)
#### What is changing? 
- Adding a conditional to check if invalid operation is passed to raw, then throwing a more descriptive error.
- Adding int tests to verify the new error is being thrown at the appropriate time

No

The previous error being thrown was not descriptive.

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
 import {
   BSONSerializeOptions,
   Document,
@@ -1133,6 +1134,9 @@ export abstract class BulkOperationBase {
 
   /** Specifies a raw operation to perform in the bulk write. */
   raw(op: AnyBulkWriteOperation): this {
+    if (!op) {
+      throw new MongoInvalidArgumentError('Cannot perform bulkwrite operation on undefined operation');
+    }
     if ('insertOne' in op) {
       const forceServerObjectId = shouldForceServerObjectId(this);
       if (op.insertOne && op.insertOne.document == null) {
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -1,4 +1,3 @@
-/* eslint-disable prettier/prettier */
 import {
   BSONSerializeOptions,
   Document,
@@ -1135,7 +1134,7 @@ export abstract class BulkOperationBase {
   /** Specifies a raw operation to perform in the bulk write. */
   raw(op: AnyBulkWriteOperation): this {
     if (!op) {
-      throw new MongoInvalidArgumentError('Cannot perform bulkwrite operation on undefined operation');
+      throw new MongoInvalidArgumentError('Cannot call .raw() with an undefined operation');
     }
     if ('insertOne' in op) {
       const forceServerObjectId = shouldForceServerObjectId(this);
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -1133,8 +1133,8 @@ export abstract class BulkOperationBase {
 
   /** Specifies a raw operation to perform in the bulk write. */
   raw(op: AnyBulkWriteOperation): this {
-    if (!op) {
-      throw new MongoInvalidArgumentError('Cannot call .raw() with an undefined operation');
+    if (op == null || typeof op !== 'object') {
+      throw new MongoInvalidArgumentError('Operation must be an object with an operation key');
     }
     if ('insertOne' in op) {
       const forceServerObjectId = shouldForceServerObjectId(this);
--- a/src/operations/insert.ts
+++ b/src/operations/insert.ts
@@ -139,7 +139,7 @@ export class InsertManyOperation extends AbstractOperation<InsertManyResult> {
       if (err || res == null) {
         if (err && err.message === 'Operation must be an object with an operation key') {
           err = new MongoInvalidArgumentError(
-            'Argument "docs" is a sparse array containing element(s) that are null or undefined'
+            'Collection.insertMany() cannot be called with an array that has null/undefined values'
           );
         }
         return callback(err);
--- a/src/operations/insert.ts
+++ b/src/operations/insert.ts
@@ -136,7 +136,14 @@ export class InsertManyOperation extends AbstractOperation<InsertManyResult> {
     );
 
     bulkWriteOperation.execute(server, session, (err, res) => {
-      if (err || res == null) return callback(err);
+      if (err || res == null) {
+        if (err && err.message === 'Operation must be an object with an operation key') {
+          err = new MongoInvalidArgumentError(
+            'Argument "docs" is a sparse array containing element(s) that are null or undefined'
+          );
+        }
+        return callback(err);
+      }
       callback(undefined, {
         acknowledged: writeConcern?.w !== 0 ?? true,
         insertedCount: res.insertedCount,
--- a/test/integration/crud/bulk.test.js
+++ b/test/integration/crud/bulk.test.js
@@ -27,10 +27,12 @@ describe('Bulk', function () {
   before(function () {
     return setupDatabase(this.configuration);
   });
-  describe('BulkOperationBase', () => {
+  // eslint-disable-next-line no-restricted-properties
+  describe.only('BulkOperationBase', () => {
     describe('#raw', function () {
+      let client;
       beforeEach(async function () {
-        const client = this.configuration.newClient();
+        client = this.configuration.newClient();
         await client.connect();
       });
       afterEach(async function () {
@@ -59,17 +61,19 @@ describe('Bulk', function () {
     });
   });
 
-  describe('Db.collection', function () {
+  // eslint-disable-next-line no-restricted-properties
+  describe.only('Db.collection', function () {
     describe('#insertMany', function () {
+      let client;
       beforeEach(async function () {
-        const client = this.configuration.newClient();
+        client = this.configuration.newClient();
         await client.connect();
       });
       afterEach(async function () {
         await client.close();
       });
       context('when passed an invalid or sparse list', function () {
-        it('insertmany should throw a MongoInvalidArgument error when called with a valid operation', async function () {
+        it('insertMany should throw a MongoInvalidArgument error when called with a valid operation', async function () {
           try {
             const docs = [];
             docs[1] = {}; // works for docs[0] = {}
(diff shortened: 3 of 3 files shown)


test/integration/crud/bulk.test.js
```diff
@@ -22,6 +27,34 @@ describe('Bulk', function () {
   before(function () {
     return setupDatabase(this.configuration);
   });
+  describe('class BulkOperationBase', () => {
```
baileympearson, line 30: small comment - based on the doc I sent I think we should use just the classname and then #raw for the method name

test/integration/crud/bulk.test.js
```diff
@@ -7,7 +7,12 @@ const {
   ignoreNsNotFound,
```
baileympearson, line 7: I think we also need a regression test here.  We've added two unit tests covering the behavior we modified but we should add a test that proves that the original bug is no longer reproducible (this prevents a "regression" in the future, where we accidentally introduce the bug back into the driver).

This should be relatively straightforward.  We need to test that calling insertMany with a sparse array throws the useful error now.  Fortunately for us, Anna gave us a reproduction (the script included in the jira ticket) so we basically just need to turn the script into a test that asserts the correct error is thrown.

kampofo: ok so something similar to the tests i had written initially?

baileympearson: I don't remember exactly how your tests looked initially, but if you deleted that work because of something I said I'm sorry 😬

test/integration/crud/bulk.test.js
```diff
@@ -22,6 +27,34 @@ describe('Bulk', function () {
   before(function () {
     return setupDatabase(this.configuration);
   });
+  describe('class BulkOperationBase', () => {
+    context('.raw()', () => {
+      it('should throw a MongoInvalidArgument error when called with an undefined operation', async function () {
+        const client = this.configuration.newClient();
```
baileympearson, line 33: Last comment - I think we should initialize and close the client in a beforeEach hook and an afterEach hook respectively.  This test uses a shared variable and a before/after Each hook to initialize it and close it for each test as an example - https://github.com/mongodb/node-mongodb-native/blob/a880eccade5491a0687af3f5c8c14d427e844995/test/integration/server-selection/server_selection_operation_count.prose.test.ts#L89

test/integration/crud/bulk.test.js
kampofo: no its completely fine it wasn't anything major!

baileympearson (approved):

src/bulk/common.ts
```diff
@@ -1133,6 +1133,9 @@ export abstract class BulkOperationBase {
 
   /** Specifies a raw operation to perform in the bulk write. */
   raw(op: AnyBulkWriteOperation): this {
+    if (!op) {
+      throw new MongoInvalidArgumentError('Cannot call .raw() with an undefined operation');
```
nbbeeken, line 1137:
```suggestion
    if (op != null || typeof op != 'object') {
      throw new MongoInvalidArgumentError('Operation must be an object with an operation key');
```
I think this error would catch a few more error cases where someone passes in something truthy but the incorrect type altogether. Also, we try to phrase error messages in terms of what must be done to correct the code.

test/integration/crud/bulk.test.js
```diff
@@ -22,6 +27,76 @@ describe('Bulk', function () {
...
+        await client.close();
+      });
+      context('when called with an undefined operation', function () {
+        it('should throw a MongoInvalidArgument error ', async function () {
+          try {
+            client.db('test').collection('test').initializeUnorderedBulkOp().raw(undefined);
+            expect.fail('Failure to throw error');
+          } catch (error) {
+            expect(error).to.be.instanceOf(MongoInvalidArgumentError);
+          }
```
nbbeeken, line 47:
```suggestion
          const bulkOp = client.db('test').collection('test').initializeUnorderedBulkOp()
          let error = await bulkOp.raw(undefined).catch(error => error);
          expect(error).to.be.instanceOf(MongoInvalidArgumentError);
          error = await bulkOp.raw(true).catch(error => error);
          expect(error).to.be.instanceOf(MongoInvalidArgumentError);
          error = await bulkOp.raw(3).catch(error => error);
          expect(error).to.be.instanceOf(MongoInvalidArgumentError);
```

This syntax is just a preference but it makes it a bit cleaner to add additional coverage here, I'm throwing in a number and boolean to hit that non-object case I suggested

nbbeeken (changes requested):

src/bulk/common.ts
kampofo: upon changing the conditional to what you suggested, started getting a bunch of errors. Ive attached a screen shot of the error im getting to this message.
Screen Shot 2022-05-02 at 4 02 36 PM

nbbeeken: My mistake! I meant `op == null || ...`

kampofo: oooooohhh!

nbbeeken (dismissed):

test/integration/crud/bulk.test.js
```diff
@@ -22,6 +27,74 @@ describe('Bulk', function () {
...
+      let client;
+      beforeEach(async function () {
+        client = this.configuration.newClient();
+        await client.connect();
+      });
+      afterEach(async function () {
+        await client.close();
+      });
+      context('when passed an invalid or sparse list', function () {
+        it('insertMany should throw a MongoInvalidArgument error when called with a valid operation', async function () {
```
dariakp, line 72: I think this should say "invalid', but, in particular, we want to be specific that we are testing a sparse array in this example. I'm not 100% sure how helpful the actual thrown error would be to the user in this case, since they would get "Operation must be an object with an operation key" - which doesn't necessarily tell them what is wrong with the top level argument they passed into insertMany.

dariakp (changes requested):

test/integration/crud/bulk.test.js
```diff
@@ -22,6 +27,86 @@ describe('Bulk', function () {
...
+      context('when called with an undefined operation', function () {
+        it('should throw a MongoInvalidArgument error ', async function () {
+          const bulkOp = client.db('test').collection('test').initializeUnorderedBulkOp();
+          expect(() => bulkOp.raw(undefined)).to.throw(MongoInvalidArgumentError);
+          expect(() => bulkOp.raw(true)).to.throw(MongoInvalidArgumentError);
+          expect(() => bulkOp.raw(3)).to.throw(MongoInvalidArgumentError);
+        });
+
+        it('should throw an error with the specifc message: "Operation must be an object with an operation key"', async function () {
+          const bulkOp = client.db('test').collection('test').initializeUnorderedBulkOp();
```
baileympearson, line 49: We can clean this test up if we use `.to.match`:

```suggestion
          expect(() => bulkOp.raw(undefined))
            .to.throw(MongoInvalidArgumentError)
            .to.match(/Operation must be an object with an operation key/);
```

test/integration/crud/bulk.test.js
```diff
@@ -22,6 +27,86 @@ describe('Bulk', function () {
...
+      });
+      context('when passed an invalid docs argument', function () {
+        it('insertMany should throw a MongoInvalidArgument error when called with a invalid operation', async function () {
+          try {
+            const docs = [];
+            docs[1] = { color: 'red' };
+            await client.db('test').collection('test').insertMany(docs);
+            expect.fail('Expected insertMany to throw error, failed to throw error');
+          } catch (error) {
+            expect(error).to.be.instanceOf(MongoInvalidArgumentError);
```
baileympearson, line 91: I think we should also check the error message here.  Since the reported bug relates to the error message thrown, we should confirm here that we're now throwing the better error message 😄

src/operations/insert.ts
```diff
@@ -136,7 +136,14 @@ export class InsertManyOperation extends AbstractOperation<InsertManyResult> {
     );
 
     bulkWriteOperation.execute(server, session, (err, res) => {
-      if (err || res == null) return callback(err);
+      if (err || res == null) {
+        if (err && err.message === 'Operation must be an object with an operation key') {
+          err = new MongoInvalidArgumentError(
+            'Argument "docs" is a sparse array containing element(s) that are null or undefined'
```
baileympearson, line 142:
```suggestion
            'Collection.insertMany() cannot be called with an array that has null/undefined values'
```

test/integration/crud/bulk.test.js
```diff
@@ -22,6 +27,86 @@ describe('Bulk', function () {
...
+      let client;
+      beforeEach(async function () {
+        client = this.configuration.newClient();
+        await client.connect();
+      });
+      afterEach(async function () {
+        await client.close();
+      });
+      context('when passed an invalid docs argument', function () {
+        it('insertMany should throw a MongoInvalidArgument error when called with a invalid operation', async function () {
```
baileympearson, line 84:
```suggestion
        it('insertMany should throw a MongoInvalidArgument error when called with an invalid operation', async function () {
```

baileympearson (changes requested):

test/integration/crud/bulk.test.js
kampofo: did u mean to put the string in the to.match in quotes?

test/integration/crud/bulk.test.js
kampofo: done

kampofo: done

src/operations/insert.ts
kampofo: done

test/integration/crud/bulk.test.js
kampofo: done

test/integration/crud/bulk.test.js
kampofo: done

baileympearson (approved):
One non-blocking comment, otherwise LGTM

test/integration/crud/bulk.test.js
```diff
@@ -22,6 +27,84 @@ describe('Bulk', function () {
...
+      let client;
+      beforeEach(async function () {
+        client = this.configuration.newClient();
+        await client.connect();
+      });
+      afterEach(async function () {
+        await client.close();
+      });
+      context('when passed an invalid docs argument', function () {
+        it('insertMany should throw a MongoInvalidArgument error when called with an invalid operation', async function () {
```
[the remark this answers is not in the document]
kampofo: done

nbbeeken (approved):

baileympearson (approved):

PR #3223, opened 2022-04-28
fix(NODE-3833): return early on end if gridfs upload stream is already ended

dariakp:
### Description
NODE-3833

#### What is changing?
Fixed the `end` override on upload gridfs stream to return early on repeat calls.

We don't want to write duplicate chunks to the steam after it has already ended.

Among other things, this fix ensures we are not affected by the following `pipeline` bug:
https://github.com/nodejs/node/issues/42866

- [N/A] New TODOs have a related JIRA ticket (though there is a relevant follow up ticket: NODE-4219)


nbbeeken (approved):
LGTM

baileympearson (approved):

PR #3228, opened 2022-05-03
remove map function from FILE_PATH_OPTIONS

grjan7:
Using lowercased items for FILE_PATH_OPTIONS would eliminate the need for mapping and lower casing that would save 100+ loops it has to go through with .map( ) and .toLowerCase( ) functions.

#### What is changing?

Removing redundant steps. 

No.


durran:
@grjan7 Thank you for the contribution which is greatly appreciated, however we won't be merging this in. We prefer to have the string keys shown in the code in their original format that they are passed to the driver so it's easy to reference what the option names are. The `map()` function is also only looping 5 strings so no performance problem there as well as the fact it's done when the module is required and not at runtime.

grjan7:
Thanks for your appreciation and explanation. I understand your preference for string format retention and its necessity in this context. Your tradeoff for readability is valued. However, I want to mention that map( ) undergoes 5 loops, and toLowerCase( ) undergoes 48 loops (combined all steps involved in converting string to lowercase counts to 100+ steps). You can consider this point if the code requires optimization.

PR #3229, opened 2022-05-03
feat(NODE-4189): Support clustered collections

lerouxb:
MongoDB 5.3 added support for clustered collections. Clustered collections are collections with a clustered index.

I based the changes in this PR off how timeseries collections work.

Scope: https://docs.google.com/document/d/1r_1BBIlfELnKBz5qAzzLDUB-BJfGrQKTS2TNHXu4DRg/edit#heading=h.b1os3ai9s8t3

### DIFF
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -47,8 +47,8 @@ export interface TimeSeriesCollectionOptions extends Document {
  */
 export interface ClusteredCollectionOptions extends Document {
   name?: string;
-  key?: Document;
-  unique?: boolean;
+  key: Document;
+  unique: boolean;
   v?: string;
 }
 
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -50,7 +50,6 @@ export interface ClusteredCollectionOptions extends Document {
   name?: string;
   key: Document;
   unique: boolean;
-  v?: string;
 }
 
 /** @public */
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -84,7 +84,7 @@ export interface CreateCollectionOptions extends CommandOperationOptions {
   pkFactory?: PkFactory;
   /** A document specifying configuration options for timeseries collections. */
   timeseries?: TimeSeriesCollectionOptions;
-  /** A document specifying configuration options for clustered collections. */
+  /** A document specifying configuration options for clustered collections. For MongoDB 5.3 and above. */
   clusteredIndex?: ClusteredCollectionOptions;
   /** The number of seconds after which a document in a timeseries or clustered collection expires. */
   expireAfterSeconds?: number;
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -41,6 +41,17 @@ export interface TimeSeriesCollectionOptions extends Document {
   granularity?: 'seconds' | 'minutes' | 'hours' | string;
 }
 
+/** @public
+ * Configuration options for clustered collections
+ * @see https://www.mongodb.com/docs/v5.3/core/clustered-collections/
+ */
+export interface ClusteredCollectionOptions extends Document {
+  name?: string;
+  key?: Document;
+  unique?: boolean;
+  v?: string;
+}
+
 /** @public */
 export interface CreateCollectionOptions extends CommandOperationOptions {
   /** Returns an error if the collection does not exist */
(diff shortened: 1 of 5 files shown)


src/operations/create_collection.ts
```diff
@@ -41,6 +41,17 @@ export interface TimeSeriesCollectionOptions extends Document {
   granularity?: 'seconds' | 'minutes' | 'hours' | string;
 }
 
+/** @public
+ * Configuration options for clustered collections
+ * @see https://www.mongodb.com/docs/v5.3/core/clustered-collections/
```
lerouxb, line 46: I suppose I need a TODO and ticket here.

src/operations/create_collection.ts
```diff
@@ -41,6 +41,17 @@ export interface TimeSeriesCollectionOptions extends Document {
...
+ */
+export interface ClusteredCollectionOptions extends Document {
+  name?: string;
+  key?: Document;
+  unique?: boolean;
```
lerouxb, line 51: Right now unique is required and has to be `unique: true`, but it seems like there are plans for this in the future. If you leave it out or set it to anything else the server will complain.

src/operations/create_collection.ts
```diff
@@ -41,6 +41,17 @@ export interface TimeSeriesCollectionOptions extends Document {
...
+  v?: string;
```
lerouxb, line 52: This is optional and if it is specified the only correct value is currently `v: '2'`.

src/operations/create_collection.ts
```diff
@@ -41,6 +41,17 @@ export interface TimeSeriesCollectionOptions extends Document {
   granularity?: 'seconds' | 'minutes' | 'hours' | string;
 }
 
+/** @public
+ * Configuration options for clustered collections
+ * @see https://www.mongodb.com/docs/v5.3/core/clustered-collections/
+ */
+export interface ClusteredCollectionOptions extends Document {
+  name?: string;
```
lerouxb, line 49: Name is optional and if left out the index will be named by the server automatically.

src/operations/create_collection.ts
```diff
@@ -41,6 +41,17 @@ export interface TimeSeriesCollectionOptions extends Document {
...
+  key: Document;
```
lerouxb, line 50: According to the scope, key is required. But right now the only valid valid is `{ _id: 1 }`

src/operations/create_collection.ts
```diff
@@ -41,6 +41,17 @@ export interface TimeSeriesCollectionOptions extends Document {
...
+  unique: boolean;
```
lerouxb, line 51: According to the scope, unique is required. But right now it has to be set to `true`.

lerouxb:
I copied what's done for timeseries and didn't see that come up in any tests, so found no tests to implement. Maybe I just missed them. Please let me know if I should add any and where.

src/operations/create_collection.ts
```diff
@@ -73,7 +84,9 @@ export interface CreateCollectionOptions extends CommandOperationOptions {
   pkFactory?: PkFactory;
   /** A document specifying configuration options for timeseries collections. */
   timeseries?: TimeSeriesCollectionOptions;
-  /** The number of seconds after which a document in a timeseries collection expires. */
+  /** A document specifying configuration options for clustered collections. */
+  clusteredIndex?: ClusteredCollectionOptions;
```
lerouxb, line 88: @baileympearson See creating a clustered collection involves specifying the clustered index options at collection creation time. It isn't possible to create a clustered index normally. Similar to timeseries (as the feature seems to have been extracted from there) except for with timeseries the indexes are hidden and not returned with listIndexes.

For clustered indexes (in clustered collections) the index shows up in listIndexes.

src/operations/indexes.ts
```diff
@@ -95,6 +95,8 @@ export interface IndexDescription
   collation?: CollationOptions;
   name?: string;
   key: Document;
+  /* Specifies that this index is clustered. This is not a valid option to provide to 'createIndexes', but can appear in the options returned for an index via 'listIndexes'. */
+  clustered?: boolean;
```
lerouxb, line 99: @baileympearson I'm trying to be careful here and not include the new boolean in CreateIndexesOptions (because you cannot create a clustered index normally) but only in IndexDescription.

lerouxb:
Don't know what this error is:
```
[2022/05/04 10:42:33.890]   1 failing
[2022/05/04 10:42:33.890]   1) Server Operation Count Tests
[2022/05/04 10:42:33.890]        operationCount is adjusted properly on successful operation
[2022/05/04 10:42:33.890]          is zero after a successful getMore:
[2022/05/04 10:42:33.890]      MongoServerError: cursor id 949597150296799923 is already in use
```

src/operations/create_collection.ts
baileympearson: This should be an `int` - see https://docs.google.com/document/d/1hRnqe1pq7ErcguJ5Bl8si1kHGKlZg0h25qsa3roASuo/edit#bookmark=id.oh101ver26rd

baileympearson: From that slack thread, it sounds like `v` should never be passed in by users.  So I think we should omit it from the public interface so that our interface matches the documentation for the `clusteredIndex` option [here](https://www.mongodb.com/docs/v5.3/reference/command/create/#std-label-create.clusteredIndex)

baileympearson (changes requested):

durran (approved):

PR #3230, opened 2022-05-03
feat(NODE-4185): Allow opting out of disk use on cursor builder

lerouxb:
See also:

* https://jira.mongodb.org/browse/MONGOSH-1183
* https://jira.mongodb.org/browse/PM-2742

### DIFF
--- a/src/cursor/find_cursor.ts
+++ b/src/cursor/find_cursor.ts
@@ -415,15 +415,16 @@ export class FindCursor<TSchema = any> extends AbstractCursor<TSchema> {
   allowDiskUse(allow = true): this {
     assertUninitialized(this);
 
+    if (!this[kBuiltOptions].sort) {
+      throw new MongoInvalidArgumentError('Option "allowDiskUse" requires a sort specification');
+    }
+
     // As of 6.0 the default is true. This allows users to get back to the old behaviour.
     if (!allow) {
       this[kBuiltOptions].allowDiskUse = false;
       return this;
     }
 
-    if (!this[kBuiltOptions].sort) {
-      throw new MongoInvalidArgumentError('Option "allowDiskUse" requires a sort specification');
-    }
     this[kBuiltOptions].allowDiskUse = true;
     return this;
   }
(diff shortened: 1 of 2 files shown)


addaleax (dismissed):

baileympearson:
Hey @lerouxb and @addaleax - the Node team discussed this work yesterday during our triage.  The Node driver currently doesn't pass along a value for `allowDiskUse` if no value is specified from the user.  This means that drivers changes aren't necessary for this work, because we won't be overriding the new server default with the old default.  The only other work associated with this change sync the new spec tests, which due to a miscommunication and time zone differences, @kampofo completed this morning (see https://github.com/mongodb/node-mongodb-native/pull/3231).  Sorry for the duplicated effort here but we'll close this PR and use Kobby's since the work is completed there.

addaleax:
Talked with @baileympearson on Slack, it seems like https://github.com/mongodb/node-mongodb-native/pull/3231 makes this appear like something that’s already done, but in reality the tests there don’t test the builder API, only the options object API, so we do still want this :slightly_smiling_face:

baileympearson:
Yup.  We still want to make this change, but we should add tests for the new options.  I think the tests should go in `find_cursor_methods.js`.

As far as testing this option, we can use command monitoring on the MongoClient to ensure that the `find` commands contain the allowDiskUse option when they're sent to the database with the correct value.

src/cursor/find_cursor.ts
```diff
@@ -412,8 +412,15 @@ export class FindCursor<TSchema = any> extends AbstractCursor<TSchema> {
    * @remarks
    * {@link https://docs.mongodb.com/manual/reference/command/find/#find-cmd-allowdiskuse | find command allowDiskUse documentation}
    */
-  allowDiskUse(): this {
+  allowDiskUse(allow = true): this {
     assertUninitialized(this);
+
+    // As of 6.0 the default is true. This allows users to get back to the old behaviour.
```
baileympearson, line 418: We should move this check to below the check for `sort`.  

Can we also add tests confirming that a MongoInvalidArgumentError is thrown if the user attempts to set allowDiskUse without a sort on their query?

baileympearson (changes requested):

baileympearson (approved):

durran (approved):

PR #3232, opened 2022-05-04
feat(NODE-4192): make MongoClient.connect optional

nbbeeken:
#### What is changing?

MongoClient.connect and client.connect are now optional! You can immediately proceed to running operations and the client will connect itself. Connect at first operation was chosen because it let's users still catch every command event that occurs as opposed to spinning up monitoring and "connect"-ing logic at construction time.

Yes! We should update our examples to omit calls to connect. We can also call out that there is no plan to remove connect, there are valid use cases where an initial method call to "test the waters" of your current environment are desirable.

Other drivers do not have a connect method so this aligns with us with our cohorts. Unified testing depends on the first operation being the one to create the first connection in a pool, kicking off monitoring events etc. Beyond that, it's convenient when working in a repl or scripting 😃 

### DIFF
--- a/test/integration/node-specific/mongo_client.test.ts
+++ b/test/integration/node-specific/mongo_client.test.ts
@@ -558,7 +558,7 @@ describe('class MongoClient', function () {
       expect(result).to.have.property('ok', 1);
     });
 
-    it('prevents automatic reconnect on a closed previously connected client', async () => {
+    it('prevents automatic reconnect on a closed previously explicitly connected client', async () => {
       await client.connect();
       expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', INIT_HAS_BEEN_CLOSED);
       await client.close();
@@ -567,7 +567,7 @@ describe('class MongoClient', function () {
       expect(error).to.be.instanceOf(MongoNotConnectedError);
     });
 
-    it('allows explicit reconnect on a previously closed but reconnected client', async () => {
+    it('allows explicit reconnect on a closed previously explicitly connected client', async () => {
       await client.connect();
       expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', INIT_HAS_BEEN_CLOSED);
       await client.close();
@@ -578,7 +578,7 @@ describe('class MongoClient', function () {
       expect(result).to.have.property('ok', 1);
     });
 
-    it('prevents auto reconnect on closed non-connected client', async () => {
+    it('prevents auto reconnect on closed previously implicitly connected client', async () => {
       expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', INIT_HAS_BEEN_CLOSED);
       const result = await db.command({ ping: 1 }).catch(error => error); // auto connect
       expect(result).to.not.be.instanceOf(MongoNotConnectedError);
@@ -589,7 +589,7 @@ describe('class MongoClient', function () {
       expect(error).to.be.instanceOf(MongoNotConnectedError);
     });
 
-    it('allows explicit reconnect on closed non-connected client', async () => {
+    it('allows explicit reconnect on closed previously implicitly connected client', async () => {
       expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', INIT_HAS_BEEN_CLOSED);
       const result = await db.command({ ping: 1 }).catch(error => error); // auto connect
       expect(result).to.not.be.instanceOf(MongoNotConnectedError);
@@ -602,7 +602,7 @@ describe('class MongoClient', function () {
       expect(result2).to.have.property('ok', 1);
     });
 
-    it('prevents auto reconnect on closed explicitly connected client', async () => {
+    it('prevents auto reconnect on closed explicitly connected client after an operation', async () => {
       expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', INIT_HAS_BEEN_CLOSED);
       await client.connect();
       const result = await db.command({ ping: 1 }).catch(error => error);
@@ -614,7 +614,7 @@ describe('class MongoClient', function () {
       expect(error).to.be.instanceOf(MongoNotConnectedError);
     });
 
-    it('allows explicit reconnect on closed previously connected client', async () => {
+    it('allows explicit reconnect on closed explicitly connected client after an operation', async () => {
       expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', INIT_HAS_BEEN_CLOSED);
       await client.connect();
       const result = await db.command({ ping: 1 }).catch(error => error);


test/integration/node-specific/mongo_client.test.ts
```diff
@@ -381,4 +373,150 @@ describe('MongoClient integration', function () {
...
+
+      expect(result).to.be.instanceOf(MongoServerSelectionError);
+      expect(client).to.be.instanceOf(MongoClient);
+      expect(client).to.have.property('topology').that.is.instanceOf(Topology);
+
+      await client.close();
+    }
+  );
+
+  it(
```
baileympearson, line 497: Should we test that 1) a user can explicitly reconnect and 2) a user can perform an operation after the client is closed, causing the client to implicitly reconnect?

src/change_stream.ts
```diff
@@ -600,8 +600,24 @@ export class ChangeStream<
...
+    const client: MongoClient | null =
+      this.type === CHANGE_DOMAIN_TYPES.CLUSTER
+        ? (this.parent as MongoClient)
+        : this.type === CHANGE_DOMAIN_TYPES.DATABASE
+        ? (this.parent as Db).s.client
+        : this.type === CHANGE_DOMAIN_TYPES.COLLECTION
+        ? (this.parent as Collection).s.db.s.client
+        : null;
+
+    if (client == null) {
```
baileympearson, line 612: question(s): even though this error should never be thrown, in theory _createChangeStreamCursor now throws.  
1. Should we add test cases for this scenario?
2. Should we update the usages of this method to ensure that we're catching the error and handling it appropriately in callback code?  Specifically, I'm looking at `_processError` in change_stream.ts.

src/sdam/topology.ts
```diff
@@ -203,6 +203,8 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
   /** @internal */
   _type?: string;
 
+  client!: MongoClient;
```
baileympearson, line 206: Users shouldn't access the client from the topology.  Should this property be marked as internal?

src/sdam/topology.ts
```diff
@@ -626,7 +628,7 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
 
   /** Start a logical session */
   startSession(options: ClientSessionOptions, clientOptions?: MongoOptions): ClientSession {
-    const session = new ClientSession(this, this.s.sessionPool, options, clientOptions);
+    const session = new ClientSession(this.client, this.s.sessionPool, options, clientOptions);
```
baileympearson, line 631: It appears this method is the only reason we need to store a reference to the client on the topology class.  Additionally, the client has a reference to the topology already.

How would you feel about removing this method and moving the logic into `client.startSession`?  That also prevents the circular client -> topology -> client reference.  

This method isn't marked as internal but Topology is so I think we're okay to make that change?

baileympearson (changes requested):

src/change_stream.ts
nbbeeken: We could add a test, the thing is this is unreachable because the constructor will throw if one of these three symbols isn't used. This is one of those fatal errors where something quite unexpected has occurred. The test would be making a change stream and then reaching in an changing the type? I can do that, would that make sense?

src/sdam/topology.ts
nbbeeken: The class is internal so there's no need, the surrounding internal documentation is unnecessary and we could remove it

src/sdam/topology.ts
nbbeeken: I attempted this, it's quite complex but we could do it in a follow up. I'll update when I've made a ticket

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -106,24 +106,14 @@ describe('MongoClient integration', function () {
...
-        topology: ['single']
-      }
-    },
-
-    test: function (done) {
-      const configuration = this.configuration;
-      const client = configuration.newClient('mongodb://unknownhost:36363/ddddd', {
-        serverSelectionTimeoutMS: 10
-      });
+  it('should fail to connect due to unknown host in connection string', async function () {
```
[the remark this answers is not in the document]
nbbeeken: done!

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -381,4 +373,150 @@ describe('MongoClient integration', function () {
...
+        const commandToBeStarted = once(client, 'commandStarted');
+
+        await client.connect();
+
+        const [pingOnConnect] = await commandToBeStarted;
+
+        expect(pingOnConnect).to.have.property('commandName', 'ping');
+        expect(client).to.have.property('topology').that.is.instanceOf(Topology);
+
+        await client.close();
```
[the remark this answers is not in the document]
nbbeeken: done!

src/sdam/topology.ts
baileympearson: Yeah makes sense.  I was just thinking about consistency but I don't feel too strongly either way

src/sdam/topology.ts
baileympearson: from slack - startSession on the client starts an explicit session every time whereas this can start both an explicit session or an implicit session.  So any internal consumers of Topology.startSession would have to work around the client starting the explicit session.

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -0,0 +1,570 @@
...
+    it(
+      'create topology and send ping when auth is enabled',
+      { requires: { auth: 'enabled' } },
+      async function () {
+        const commandToBeStarted = once(client, 'commandStarted');
+        await client.connect();
+        const [pingOnConnect] = await commandToBeStarted;
+        expect(pingOnConnect).to.have.property('commandName', 'ping');
+        expect(client).to.have.property('topology').that.is.instanceOf(Topology);
+        await client.close();
```
baileympearson, line 403: did you mean to leave the `client.close` here?

nbbeeken: What client.close? 😎

src/change_stream.ts
baileympearson: Yeah, that's a fair point.  I'll leave this thread open for additional reviews for context.

src/operations/execute_operation.ts
```diff
@@ -66,45 +62,46 @@ export interface ExecutionResult {
...
-    try {
-      // TODO(NODE-4151): Use skipPingOnConnect and call connect here to make client.connect optional
-      topology = getTopology(topologyProvider);
-    } catch (error) {
-      return callback(error);
+    const topology = client.topology;
+
+    if (topology == null) {
+      if (client.s.hasBeenClosed) {
+        return callback(new MongoNotConnectedError('client was closed'));
```
baileympearson, line 88: Can we add a more descriptive error message?  Something like:

"attempted to execute operation with closed client"

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -0,0 +1,569 @@
...
+        monitorCommands: true
+      });
+    });
+
+    afterEach(async function () {
+      await client.close();
+    });
+
+    it(
+      'create topology and send ping when auth is enabled',
```
baileympearson, line 395: This comment applies to a number of tests in this file.  But can we fix the wording?  Instead of `it('create topology....'` it should be `it('creates the topology...`

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -0,0 +1,569 @@
...
+        const [findCommandStarted] = await findCommandToBeStarted;
+
+        expect(pingOnConnect).to.have.property('commandName', 'ping');
+        expect(findCommandStarted).to.have.property('commandName', 'find');
+        expect(client).to.have.property('topology').that.is.instanceOf(Topology);
+      }
+    );
+
+    it(
+      'automatically connect upon first operation (find)',
```
baileympearson, line 425: Any tests that are testing the implicit connect should probably be in a separate describe block - this block is labeled `explicit #connect`

baileympearson (changes requested):

src/mongo_client.ts
```diff
@@ -446,6 +452,14 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
     forceOrCallback?: boolean | Callback<void>,
     callback?: Callback<void>
   ): Promise<void> | void {
+    // There's no way to set hasBeenClosed back to false
+    Object.defineProperty(this.s, 'hasBeenClosed', {
```
baileympearson, line 456: There was discussion in driver devs about whether or not we should allow clients to automatically reconnect if the client is closed ([thread](https://mongodb.slack.com/archives/C72LB5RPV/p1652213273714369)).  Based on the feedback, we decided no.

src/operations/execute_operation.ts
nbbeeken: In the spirit of phrasing things as "must be" I went with this:
`Client must be connected before running operations`

test/integration/node-specific/mongo_client.test.ts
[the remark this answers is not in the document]
nbbeeken: Added!

nbbeeken: I fixed most of these up, the close ones didn't really work

test/integration/node-specific/mongo_client.test.ts
nbbeeken: lmk what you think of the new organization

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -0,0 +1,569 @@
...
+          .findOne()
+          .catch(error => error);
+
+        expect(result).to.be.instanceOf(MongoServerSelectionError);
+        expect(client).to.be.instanceOf(MongoClient);
+        expect(client).to.have.property('topology').that.is.instanceOf(Topology);
+      }
+    );
+
+    it(
```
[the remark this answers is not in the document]
nbbeeken: I added an assertion to the connect after close

baileympearson (dismissed):

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -0,0 +1,605 @@
...
+      'client.close will not permit operations to auto reconnect',
```
dariakp, line 515: can we word this as an action though? e.g. "does not auto reconnect after client.close"

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -0,0 +1,605 @@
...
+      { requires: { auth: 'enabled' } },
+      async function () {
+        await client.db('test').collection('test').findOne();
+
+        expect(client).to.be.instanceOf(MongoClient);
```
dariakp, line 520: are these expectations relevant to the test case? if we're worried that implicit connect isn't correctly establishing the client, we should just have an explicit test for that and not here; same regarding topology -> undefined; if we want to explain the behavior of "close" then we should have a test that says something like "#close sets client topology to undefined"

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -0,0 +1,605 @@
...
+
+        expect(client).to.have.property('topology', undefined);
+        expect(result).to.be.instanceOf(MongoNotConnectedError);
+
+        await client.close();
+      }
+    );
+
+    it(
+      'client.close will not permit operations to auto reconnect permanently',
```
dariakp, line 541: there's too much going on in this test; for starters, this is inside an "implicit" reconnect, but it also has an explicit connect in there... and the explicit connect is unclear in terms of what it's supposed to accomplish - do we want to be able to run operations after explicitly calling connect or not? The behavior, I think, is unexpected - what does it mean to connect "successfully" but still not be able to run operations? If we aren't able to run operations after explicit connect, then we should throw an error in the connect itself, not some time later. I'd like to see test cases which clarify behavior of the "close" method in the following cases

Running no operations and no connect to start:
1) create client, immediately call close, try to run an operation
2) create client, immediately call close, explicitly call connect, try to run an operation
Running no operations but with connect to start:
3) create client, explicitly connect, immediately call close, try to run an operation
4) create client, explicitly connect, immediately call close, explicitly call connect, try to run an operation
Running an operation to implicitly connect to start:
5) create client, run an implicitly connecting operation, call close, try to run an operation
6) create client, run an implicitly connecting operation, call close, explicitly call connect, try to run an operation
Running an operation with an explicit connection to start:
7) create client, explicitly connect, run an operation, call close, try to run an operation again
8) create client, explicitly connect, run an operation, call close, explicitly call connect, try to run an operation again

In the above, the odd test cases would be testing the implicit reconnect, and the even cases would be testing the explicit reconnect

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -0,0 +1,605 @@
...
+        expect(secondResult).to.be.instanceOf(MongoNotConnectedError);
+
+        expect(client.s).to.have.ownPropertyDescriptor(
+          'hasBeenClosed',
+          postCloseHasBeenClosedDescriptor
+        );
+      }
+    );
+  });
+});
```
dariakp, line 605: Just for visibility, let's make sure we don't close the actual jira ticket until we do this part of the AC:

> Review tests and update any that explicitly call MongoClient connect.
> File a separate PR for this.

dariakp (changes requested):

test/integration/node-specific/mongo_client.test.ts
nbbeeken: Yep, @baileympearson is on top of this, and I'll be posting the test PR after we merge this avoiding rebase annoyances.

src/operations/execute_operation.ts
```diff
@@ -66,45 +62,48 @@ export interface ExecutionResult {
...
-      return callback(error);
+    const topology = client.topology;
+
+    if (topology == null) {
+      if (client.s.hasBeenClosed) {
+        return callback(
+          new MongoNotConnectedError('Client must be connected before running operations')
+        );
+      }
+      client.s.options[Symbol.for('@@mdb.skipPingOnConnect')] = true;
```
durran, line 92: Just a question - is this feature flag still needed for those who still explicitly calling `client.connect()`?

durran (changes requested):

src/operations/execute_operation.ts
nbbeeken: No one needs to set this, it was just for us to use in tests to make them consistent for other drivers. I'm then using it here to make sure the auto connect doesn't run two operations.

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -510,96 +510,122 @@ describe('class MongoClient', function () {
...
+      expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', RD_ONLY_HAS_BEEN_CLOSED);
+      const error = await db.command({ ping: 1 }).catch(error => error);
+      expect(error).to.be.instanceOf(MongoNotConnectedError);
+    });
 
-        expect(client.s).to.have.ownPropertyDescriptor(
-          'hasBeenClosed',
-          postCloseHasBeenClosedDescriptor
-        );
+    it('allows explicit reconnect on a previously closed but reconnected client', async () => {
```
dariakp, line 570:
```suggestion
    it('allows explicit reconnect on a closed previously explicitly connected client', async () => {
```

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -510,96 +510,122 @@ describe('class MongoClient', function () {
...
+      await client.close();
+      expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', RD_ONLY_HAS_BEEN_CLOSED);
+      await client.connect();
+      const result = await db.command({ ping: 1 }).catch(error => error);
+      expect(result).to.not.be.instanceOf(MongoNotConnectedError);
+      expect(result).to.have.property('ok', 1);
+    });
 
-        expect(pingOnConnect).to.have.property('commandName', 'ping');
+    it('prevents automatic reconnect on a closed previously connected client', async () => {
```
dariakp, line 561:
```suggestion
    it('prevents automatic reconnect on a closed previously explicitly connected client', async () => {
```

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -510,96 +510,122 @@ describe('class MongoClient', function () {
...
+      await client.close();
+      expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', RD_ONLY_HAS_BEEN_CLOSED);
+      await client.connect();
+      const result = await db.command({ ping: 1 }).catch(error => error);
+      expect(result).to.not.be.instanceOf(MongoNotConnectedError);
+      expect(result).to.have.property('ok', 1);
+    });
 
-        await client.close();
+    it('prevents auto reconnect on closed non-connected client', async () => {
```
dariakp, line 581:
```suggestion
    it('prevents auto reconnect on closed previously implicitly connected client', async () => {
```

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -510,96 +510,122 @@ describe('class MongoClient', function () {
...
+      const error = await db.command({ ping: 1 }).catch(error => error);
+      expect(error).to.be.instanceOf(MongoNotConnectedError);
+    });
 
-        const secondResult = await client
-          .db('test')
-          .collection('test')
-          .findOne()
-          .catch(error => error);
+    it('allows explicit reconnect on closed non-connected client', async () => {
```
dariakp, line 592:
```suggestion
    it('allows explicit reconnect on closed previously implicitly connected client', async () => {
```

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -510,96 +510,122 @@ describe('class MongoClient', function () {
...
+      await client.connect();
+      expect(client.s).to.have.ownPropertyDescriptor('hasBeenClosed', RD_ONLY_HAS_BEEN_CLOSED);
+      const result2 = await db.command({ ping: 1 }).catch(error => error);
+      expect(result2).to.not.be.instanceOf(MongoNotConnectedError);
+      expect(result2).to.have.property('ok', 1);
+    });
 
-        expect(client).to.have.property('topology', undefined);
-        expect(secondResult).to.be.instanceOf(MongoNotConnectedError);
+    it('prevents auto reconnect on closed explicitly connected client', async () => {
```
dariakp, line 605:
```suggestion
    it('prevents auto reconnect on closed explicitly connected client after an operation', async () => {
```

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -510,96 +510,122 @@ describe('class MongoClient', function () {
...
+      expect(error).to.be.instanceOf(MongoNotConnectedError);
+    });
 
-        expect(client.s).to.have.ownPropertyDescriptor(
-          'hasBeenClosed',
-          postCloseHasBeenClosedDescriptor
-        );
-      }
-    );
+    it('allows explicit reconnect on closed previously connected client', async () => {
```
dariakp, line 617:
```suggestion
    it('allows explicit reconnect on closed explicitly connected client after an operation', async () => {
```

dariakp (changes requested):
Just suggestion for clarifying test case names

src/admin.ts
```diff
@@ -83,7 +83,7 @@ export class Admin {
     options = Object.assign({ dbName: 'admin' }, options);
 
     return executeOperation(
-      this.s.db,
+      this.s.db.s.client,
```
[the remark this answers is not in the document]
nbbeeken: Chatted offline, we'll likely clean this up later for now, TS will protect against us making changes here and we can apply some sweeping change later using that hinting

dariakp (approved):

alecgibson:
Is there any documentation on how to use the client "correctly" now? This change has broken some of our unit tests where we try to tear down and recreate a `MongoClient` between them.

nbbeeken:
Hey @alecgibson the way to use the client correctly remains the same as before, the automatic connect is a new way that is also correct but has some known limitations that we're working on getting fixed very soon in the next patch! The watch method will no longer throw that you must be connected but unfortunately the change stream is unable to create a session needed to initialize the events. So to work around this I would continue using connect as was needed before and this will be covered in the fix in #3286

PR #3233, opened 2022-05-04
test(NODE-3248): sync new unified change stream tests & enhance runner

baileympearson:
#### What is changing?

- New valid-pass change stream spec tests are synced
- Change stream unified tests were synced from this spec commit - https://github.com/mongodb/specifications/pull/1156
- `rename` operation was added to the runner
- new option, `ignoreExtraEvents` was added to the runner
- dropDatabase helper now suppresses errors (https://mongodb.slack.com/archives/C72LB5RPV/p1651863870640589 for context)

Note - this PR does not include changes from NODE-3938.

Nope.


test/tools/unified-spec-runner/operations.ts
```diff
@@ -245,7 +245,13 @@ operations.set('deleteOne', async ({ entities, operation }) => {
 operations.set('dropCollection', async ({ entities, operation }) => {
   const db = entities.getEntity('db', operation.object);
   const { collection, ...opts } = operation.arguments;
-  return await db.dropCollection(collection, opts);
+  try {
+    await db.dropCollection(collection, opts);
```
dariakp, line 249: 1) I think this should still `return` the result, instead of just calling the operation, to keep consistent  with the previous behavior, in case there is something that is expecting a particular boolean response
2) I am not convinced that try/catch *here* is a good approach - this is a general definition for the drop operation, so swallowing the error would mean we can't test for cases where we definitely should *not* be getting this error from `dropCollection`: can you explain again why we need this and can we find a more surgical place to insert it? perhaps with a ticket to remove at a later point?

test/tools/unified-spec-runner/runner.ts
```diff
@@ -156,10 +156,15 @@ export async function runUnifiedTest(
     const topologyType = ctx.configuration.topologyType;
     if (topologyType === TopologyType.Sharded || topologyType === TopologyType.LoadBalanced) {
       for (const [, collection] of entities.mapOf('collection')) {
-        await utilClient.db(ns(collection.namespace).db).command({
-          distinct: collection.collectionName,
-          key: '_id'
-        });
+        try {
+          // TODO(NODE-4238): create / cleanup entities for each test suite
```
dariakp, line 160: Can you provide more context for why this is necessary?  The ticket you have says "Ensuring entities are created on the server prior to starting the tests will remove the need to catch that error." - it would be helpful to elaborate on this - which tests are failing without it and why?

test/tools/unified-spec-runner/match.ts
```diff
@@ -349,6 +353,42 @@ export function matchesEvents(
   }
 }
 
+export function matchesEvents(
+  { events: expected, ignoreExtraEvents }: ExpectedEventsForClient,
+  actual: CommandEvent[] | CmapEvent[],
+  entities: EntitiesMap
+): void {
+  ignoreExtraEvents = ignoreExtraEvents ?? false;
+
```
dariakp, line 362: I would suggest adding a couple extra variables here along the lines of
```js
const misMatchedCount = ignoreExtraEvents? actual.length < expected.length : actual.length !== expected.length
if (mismatchedCount) { /** ... repeated event count error logic ... */ }
const actualEventsToCompare = ignoreExtraEvents ? actual.slice... : actual
compare(expected, actualEventsToCompare, entities)
```
just to make it easier to focus on the difference in handling between the `ignoreExtraEvents = true` vs `false` case

dariakp (changes requested):

test/tools/unified-spec-runner/runner.ts
baileympearson: Tests are failing due to database not found errors on some server versions.  It appears that the distinct command does not check for database existence on the server, throwing the error - https://jira.mongodb.org/browse/SERVER-60533.

Since other drivers explicitly create the entities on the server, they don't see this issue.

baileympearson: Here's an evergreen patch with the try/catch removed that shows the errors - https://spruce.mongodb.com/version/627bc1f2d1fe074fc039b85b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

It's interesting that

dariakp (changes requested):

test/integration/change-streams/change_streams.spec.test.ts
```diff
@@ -7,8 +7,30 @@ import { loadSpecTests } from '../../spec';
 import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
 import { delay, setupDatabase } from '../shared';
 
+const skippedResumabilityTests = [
```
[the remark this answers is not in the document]
baileympearson: Done!

baileympearson:
@dariakp I can't respond to your comments for some reason 🤔 But I addressed the two remaining comments you had.

dariakp (approved):

nbbeeken (approved):
lgtm

PR #3234, opened 2022-05-05
fix(NODE-4232): stream() also returns generic AsyncIterable

qwelias:
#### What is changing?
AbstractCursor#stream also returns `AsyncIterator<TSchema>`

update autogenerated docs maybe

need to carry the type when doing async iteration over stream() result

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -385,7 +385,7 @@ export class ChangeStream<
   /** @internal */
   [kResumeQueue]: Denque<Callback<ChangeStreamCursor<TSchema, TChange>>>;
   /** @internal */
-  [kCursorStream]?: Readable;
+  [kCursorStream]?: Readable & AsyncIterable<TChange>;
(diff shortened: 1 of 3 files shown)


durran (changes requested):
Hi @qwelias, would you also mind adding the type and a test to `ChangeStreamCursor` as well for consistency? (https://github.com/mongodb/node-mongodb-native/blob/main/src/change_stream.ts#L545). Thanks!

durran (changes requested):
Also I believe we need a test to actually check that the iterator is valid in the for-await-of use case in the cursor type test:

```js
const cursor1 = collection.find();
const x = cursor1.stream();
for await (const doc of x) {
  expectType<{ age: number }>(doc);
}
```

nbbeeken (approved):
LGTM thanks!

baileympearson (approved):

PR #3236, opened 2022-05-06
fix(NODE-4108): improve return type for withTransaction()

nbbeeken:
### Description

The `withTransaction()` function signature implies that `withTransaction()` returns the same value that the user-provided callback returns but that is incorrect.

Reported in https://github.com/Automattic/mongoose/issues/9919

#### What is changing?

Function signature for `withTransaction()` maintains the T argument but goes unused. Added tests for asserting the expected outcomes.

Fixes the type bug, but also adds documentation to the three possible outcomes of the function.

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -36,6 +36,14 @@ declare global {
   }
 
   namespace Mocha {
+    interface SuiteFunction {
+      (title: string, metadata: MongoDBMetadataUI, fn: (this: Suite) => void): Mocha.Suite;
+    }
+
+    interface PendingSuiteFunction {
+      (title: string, metadata: MongoDBMetadataUI, fn: (this: Suite) => void): Mocha.Suite;
+    }
+
     interface TestFunction {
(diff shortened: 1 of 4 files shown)


global.d.ts
```diff
@@ -36,6 +36,14 @@ declare global {
   }
 
   namespace Mocha {
+    interface SuiteFunction {
```
durran, line 39: Not a request for changes just wanting to understand the rationale for including these 2 definitions here.

durran (approved):

global.d.ts
nbbeeken: Our metadata UI also allows specifying filters at the describe level, only in the 3 argument form of `title, metadata, function` So this just prevents type issues when using a metadata object in `describe` or `context`. I just don't think we've used metadata at the suite level much before or we haven't needed it in TS test yet. But the transactions tests will always need to filter up to 4.2 maybe 4.4 so it makes sense to specify it higher up.

PR #3243, opened 2022-05-11
fix(NODE-4139): allow monitor to yield control

durran:
### Description

Changes the monitor to process incoming hellos when the streaming protocol is enabled in a timeout, so that it can yield the process to other potential commands.

#### What is changing?

Adds a `setTimeout` in the callback after the hello command is run. This has been shown on aws lambda to reduce processing time after a function is idle for hours from over 20 seconds to ~400ms.

None

NODE-4140

### DIFF
--- a/src/sdam/monitor.ts
+++ b/src/sdam/monitor.ts
@@ -279,11 +279,13 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
       // if we are using the streaming protocol then we immediately issue another `started`
       // event, otherwise the "check" is complete and return to the main monitor loop
       if (isAwaitable && hello.topologyVersion) {
-        monitor.emit(
-          Server.SERVER_HEARTBEAT_STARTED,
-          new ServerHeartbeatStartedEvent(monitor.address)
-        );
-        start = now();
+        setTimeout(function () {
+          monitor.emit(
+            Server.SERVER_HEARTBEAT_STARTED,
+            new ServerHeartbeatStartedEvent(monitor.address)
+          );
+          start = now();
+        }, 0);
       } else {
         monitor[kRTTPinger]?.close();
         monitor[kRTTPinger] = undefined;


addaleax (approved):

src/sdam/monitor.ts
```diff
@@ -279,11 +279,13 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
       // if we are using the streaming protocol then we immediately issue another `started`
       // event, otherwise the "check" is complete and return to the main monitor loop
       if (isAwaitable && hello.topologyVersion) {
-        monitor.emit(
-          Server.SERVER_HEARTBEAT_STARTED,
-          new ServerHeartbeatStartedEvent(monitor.address)
-        );
-        start = now();
+        setTimeout(function () {
```
dariakp, line 282: can we add a test for the control yield behavior or any test that captures the intended behavior here? the change itself seems straightforward enough; since nothing in our code is listening for this, I don't think this would have any sort of adverse impact on existing monitoring and monitoring-adjacent logic, but I want to make sure, if this is indeed solving the heartbeat issue, that we have some sort of documentation for posterity that states why this is important to have in a timeout; at a minimum I'd like **a comment explaining why** we want to yield control specifically *here* prior to returning, but preferably a regression (unit) test that demonstrates the important flow

if a test is possible but may take a while, I don't mind it being in a follow up PR if we're trying to expedite this PR specifically

dariakp (changes requested):

src/sdam/monitor.ts
mbroadst, line 282: I'm concerned this isn't the right place for the fix. I think the intuition here is that the receive buffer is full of incoming heartbeat responses when the lambda instance is "unfrozen", and so you end up with a thundering herd effect - trying to process too many of these messages at the same time results in event loop starvation. 

I was initially confused that slightly deferring the emission of a monitoring event would fix this issue since it's not used internally at all. Also, we emit a different event a few lines above on L274 so why wouldn't we also apply the fix there? (Note: I there is actually a bug with the code as presented here, in that a started event can now appear _after_ a succeeded event). 

Here's what I think is going on:
Streamable hello streams responses to clients using the `moreToCome` flag. To ease the implementation at the call site, I implemented this by [requeuing a provided callback each time `moreToCome` is true](https://github.com/mongodb/node-mongodb-native/blob/d0f622b86904e6177fb29372611a75a36f7003d0/src/cmap/connection.ts#L402-L404) (I also did a really bad job of documenting this, so I would not be surprised if this behavior is unexpected to you!). If we indeed have a thundering herd of incoming `hello` messages, then I think it's totally possible that the callback is being called in a hot loop and starving the event loop. The fix here I think is to wrap the [call to the `callback`](https://github.com/mongodb/node-mongodb-native/blob/d0f622b86904e6177fb29372611a75a36f7003d0/src/cmap/connection.ts#L451) in a setTimeout/process.nextTick, so that we are able to do additional work between streamed calls.

dariakp: Thanks for the explanation! I certainly wasn't aware of the moreToCome flag for the streaming hello protocol. Is the streaming hello the only server context where this is used?

Looking at this implementation more, I'm actually not sure how it's accomplishing the performance improvement, since the only thing being delayed is the emission of the SERVER_HEARTBEAT_STARTED events. If what you are describing is accurate and I understand it correctly (i.e., this callback is basically being called in a loop with no async breaks), I would expect all of those SERVER_HEARTBEAT_SUCCEEDED events to get fired in order, before any SERVER_HEARTBEAT_STARTED events, since they are now in a settimeout, which, presumably, would get handled at the end of that loop; and, moreover, I would expect the `duration` associated with each of the SUCCEEDED events to also increase, since the `start` would no longer be reset. @durran is this what we observe happening from the logs? The only extra processing associated with the event emission would be logging - so is that what we are trying to delay here @durran ? Am I reading this correctly?

I think, in that case, Matt's suggestion does make sense - we want to async-ify the calling of the handler for each event. I am still not fully following, however, how the "synthetic" event processing works. 1) The comment above there says that the responseTo changes on each response - does it get set to the previous message's requestId?  2) If we do put the callback in a timeout, won't it mean that we are still going to end up with a stack of callbacks that all get executed back to back on the next pass through the event loop? Essentially, we'd only be buying time between reading the "stampede" of hello messages and actually executing the set of the handlers for the hello responses - is that what we want?

dariakp: Following that thought - if we already can delay the callback execution for `moreToCome`, could we detect if we have an unprocessed pile of `moreToCome` callbacks and only queue the latest? If we have queued a setTimeout, we could clear it if we have a more recent one incoming. Would that be too dangerous?

dariakp: Alternatively, we could track time between invocations on the monitoring callback itself and return if it has already been recently invoked (similar to how we debounce the wake calls for non-streaming)

mbroadst: >  Is the streaming hello the only server context where this is used?
No, `moreToCome` is also supported on exhaust cursors, but drivers don't implement support for those today. You can learn more about this flag [here](https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#moretocome).

> Looking at this implementation more, I'm actually not sure how it's accomplishing the performance improvement...
That's right, you're describing the bug I mentioned in my comment. It should be fairly easy to prove visually that this is happening (if you're printing out messages on the events). 

> I am still not fully following, however, how the "synthetic" event processing works.
Each time the driver sends a request, it enqueues a callback in its response queue waiting for a response with a `responseTo` which correlates with the request's `requestId`. When the response comes in we dequeue it and call the callback. When the response indicates it has `moreToCome`, we immediately re-enqueue the callback, but with a "correlation id" of the requestId of the response (this is an implementation detail of the server, and most likely because the server implements streaming support by running a polling cursors server-side, so there are many request/responses but just happening on the server). 

> If we do put the callback in a timeout, won't it mean that we are still going to end up with a stack of callbacks that all get executed back to back on the next pass through the event loop?
Yes, that's correct, but I also do think that's what you want. The stampede of messages isn't inherently something the driver can't handle, the problem is that we are starving the event loop and preventing the runtime from processing other messages, which is an "async sin". I suspect the reason the changes in this PR convey a performance boost is that:
- the test application is printing out a string on `serverHeartbeatEvent`
- every `moreToCome` request that comes in results in calling at least one `console.log`, so now you have event loop starvation while calling I/O

> Following that thought - if we already can delay the callback execution for moreToCome, could we detect if we have an unprocessed pile of moreToCome callbacks and only queue the latest..
I would be nervous to try do something like this. If this is a situation where the data is queued in kernel read buffers, then you can't make this decision in your driver code since the driver needs to read the request from the TCP stream first. You don't know a) when you are about to receive a "herd", and b) when it stops. 

> we could track time between invocations on the monitoring callback itself and return if it has already been recently invoked
This is similarly dangerous. If you don't know when you'll receive the last one you might debounce the callback and never "receive" the last `hello`. 


I think the easiest thing to do here is to wrap the call to the callback with `setTimeout`/`process.nextTick` and let the node runtime do what it does best.
```
const result = operationDescription.fullResult ? message : message.documents[0];
if ('moreToCome' in message && message.moreToCome) {
    setTimeout(() => callback(undefined, result), 0);
    return;
}

callback(undefined, result);
```

jyemin: @mbroadst would this approach have the effect of delaying processing of the hello response in the "normal" case, where the client is keeping up with these response, and ideally would be handling them before other activity on the event loop?

mbroadst: I wouldn't think of the "delay" added here as being meaningful in terms of monitoring correctness. Generally, this callback is called every ~10s (whatever is negotiated with the server for streaming hello's interval), so while when its called it usually is processed "immediately", I don't think that conveys any guarantees around ordering of operations - monitoring is always a best-effort engagement. The issue here is that when things have gone awry, the immediateness of this code prevents any other code from running.

Also, for the engineers who are going to conjure a fix for this problem, I suggest perusing node's [documentation on the event loop ](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick). I think we'll want to think carefully about whether `setTimeout`/`setImmediate`/`process.nextTick` is used in this case.

jyemin: Streaming hello was never about correctness though, it was all about decreasing the latency of the driver reacting to a change in server state.  So in the normal case, a response comes every 10s.  But in the abnormal case, when the server is shutting down or during an election, the response comes sooner than that and we want the driver to respond by changing SDAM state as quickly as possible, because it's likely that any other message sent to the server will go to the wrong server until that state is updated, and will have to be retried.

mbroadst: Maybe we're working with different definitions of correctness, but I meant it in terms of "being able to select the right server for an operation" (which I think jibes with what you're saying?). The delay we are discussing introducing here is very small (I don't have raw numbers, but I think we can assume its sub-millisecond). I would be very surprised if the server streamed multiple hello responses in that type of timeframe, considering elections occur on the order of tens to hundreds of milliseconds.

baileympearson: I think Matt's correct here in that the delays caused by an async timeout of some kind around the callback wouldn't have a practical implication on the correctness of server selection.

With respect to Matt's points about event loop starvation and I/O - @durran have we tried timing the function as a whole without any i/o operations after a buildup of heartbeats occur? 

I tested a near identical situation locally a while back too (see my reproduction code here - https://github.com/mongodb/node-mongodb-native/compare/lambda-local-reproduction).  This was to track how long processing each heartbeat took when they piled up using sigstop.  This reproduction has the same problems as lambda - a huge backlog of heartbeats from when the driver was sigstopped and i/o in the loop (one notable difference is that I'm writing to a filestream - I don't know enough about Node's internal implementation of writeFileStreams and console.log to know if writefileStream is more performant here).  

But either way, if the i/o in the loop is the issue, this is something we should be able to reproduce locally (maybe we can try explicitly with durran's sample code and sigstop again).

dariakp: @mbroadst Thanks for the added detail on the streaming hello, that is very helpful.

Regarding the stack of callbacks - yes, we are mitigating the problem by breaking the processing up into two parts, but my concern is that second part that was creating a backup is still going to cause a similar backup in the next go around of the loop. Because before it was:
`process all messages and all callbacks for hello responses [new messages blocked for 10 seconds]` -> `process new messages` and in the suggested solution it is `process all messages for hello responses [new messages blocked for 1 second]` -> `process new messages` -> `process all callbacks for hello responses [new messages blocked for 10-1 = 9 seconds]`. Regardless of which async method is used to queue the delayed callback execution, the next pass of the event loop will be blocked in the phase where we pushed the callbacks for the amount of time it takes to execute all of them, and that does not feel like it addresses a key issue here - the unnecessary processing of the hello's, since, ultimately, the information from the most recent hello will override any of the preceding ones and won't make a difference to anything else executing in the code since these callbacks will get executed synchronously. I understand that we don't want to leave messages hanging on the socket, so we do need to read them, my question is, once read (and, unless I am misunderstanding something, by the time we are on the last line of `connection.onMessage`, we have already taken care of that), can we discard the callback instead of queuing it. If we can detect that we are reading a streaming hello response (at the moment, it could be via the moreToCome indicator), then doing something like:
```js
if (isStreamingHello(parsedMessage)) {
  // only cancels a callback that has been scheduled in the same pass of the loop since the callback
  // will get executed at the top of the next pass of the event loop
  clearTimeout(this[kHelloTimeout]); 
  this[kHelloTimeout] = setTimeout(() => callback(...), 0);
}
```
would mean that during each single pass of the event loop, if there are multiple hello messages queued up, only the most recently sent hello gets registered by the monitor. Is that actually unsafe?

baileympearson: @dariakp Would that block the event loop on the second go?  My understanding of the event loop was that all synchronous events take precedence over asynchronous events and that the async queue is non-blocking, Once the synchronous queue is empty, we'd start handling the hellos, but once another synchronous event arrives it will take precedence.  Is that not correct?

dariakp: @baileympearson not exactly

> when the event loop enters a given phase, it will perform any operations specific to that phase, then execute callbacks in that phase's queue until the queue has been exhausted or the maximum number of callbacks has executed.

> Once the poll queue is empty the event loop will check for timers whose time thresholds have been reached. If one or more timers are ready, the event loop will wrap back to the timers phase to execute those timers' callbacks.

https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/

durran: `setTimeout()` I believe is the correct way to go in this case as `process.nextTick()` could potentially have the same issue as we are currently seeing as it would not exit the poll phase until all the i/o is completed. What @mbroadst did point out is correct in the ordering of the heartbeats and I need to move this around to make sure that each started heartbeat gets the appropriate subsequent succeeded/failed heartbeat immediately after it. That's currently not the case in the logs:

Screenshot 2022-05-12 at 16 34 36

mbroadst: @dariakp it might work, but I think at this point it might make sense to develop some microbenchmarks to ascertain the effect of different solutions in this relatively hot code path. A few thoughts:
- I think you are making an assumption that all messages are processed at exactly the same time. If that's what happens, you might be right that deferring processing each of them to the next tick might just be delaying the inevitable. I was assuming that batches of messages would be processed together, and so we would free up the event loop by allowing it to process a batch-at-a-time. Maybe we can test this theory by putting a stream in front of a large amount of in-memory BSON bytes (simulating the full kernel read buffer) and observing the behavior?
- Streaming with `moreToCome` is originally intended to allow for streaming results of queries, streaming hello just takes advantage of the protocol bits for a slightly different purpose. We do have an assumption that monitoring events are monotonically increasing, so your idea to debounce those events works, so long as we limit the debouncing just to monitoring events (I think today that would mean implementing this in the callback, not in `connection.ts` where you call the callback). I think we should also consider a solution that works for the general streaming case where you _do_ want to process all batches.

Thanks for confirming our suspicions @durran. Were you also able to confirm that moving the `setTimeout` to the callback in `connect.ts` resolves the issue?

durran:
Per our discussion this solution is gone away in favour of a message stream optimisation in lambda environments.

src/sdam/monitor.ts
durran: @mbroadst Just to update what we ended up discussion in a meeting. Adding `setTimeout` to the callback did allow control to yield and allow actual commands to process and not be blocked by all the heartbeats. However, all the heartbeats did still need to be processed, which was still taking significant time when the lambda functions were at their default resource allocation (128mb ram/2 cores). What we decided to do to help with that issue, was to change the `MessageStream` to know if it was for a monitoring connection, and when reading incoming messages from the buffer would only ever actually process the last message on the buffer. (https://github.com/mongodb/node-mongodb-native/commit/4b9ad777fb78cc37b71dece0e647593950efccb8). This has alleviated the time to process issue on the small functions. (Note when I tested on lambda functions with more resources the problem never initially happened. For example giving the function 1gb ram, I could never see this issue)

mbroadst: @durran ok, thanks for the update!

PR #3244, opened 2022-05-11
feat(NODE-4079): estimated document count uses count

durran:
### Description

Updates `estimatedDocumentCount` to use the `count` command.

#### What is changing?

Changes the operation to use a `count` command instead of an `aggregate`.

None

NODE-4079


src/operations/estimated_document_count.ts
```diff
@@ -36,9 +36,8 @@ export class EstimatedDocumentCountOperation extends CommandOperation<number> {
     if (maxWireVersion(server) < 12) {
       return this.executeLegacy(server, session, callback);
     }
```
baileympearson, line 38: Can we 

1. get rid of all wire version checks?  We're using the `count` command all the time now.
2. Get rid of `executeLegacy`?

baileympearson (changes requested):
The spec changes say we need to document the behavior (in addition to the release notes).  Should we modify the comment on `Collection.estimatedDocumentCount` to include the relevant information?

see https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount

baileympearson:
Can we also create a jira ticket and mark it for the next minor (4.7.0) with these changes, so we remember to include the relevant information in the release notes?  See https://mongodb.slack.com/archives/GGWBN4ZNK/p1652725071464819

durran:
> Can we also create a jira ticket and mark it for the next minor (4.7.0) with these changes, so we remember to include the relevant information in the release notes? See https://mongodb.slack.com/archives/GGWBN4ZNK/p1652725071464819

https://jira.mongodb.org/browse/NODE-4253

src/operations/estimated_document_count.ts
durran: Nice catch - I've updated per the suggestions.

durran:
> The spec changes say we need to document the behavior (in addition to the release notes). Should we modify the comment on `Collection.estimatedDocumentCount` to include the relevant information?
> 
> see https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount

I've updated the TSDoc to note this and link to the MongoDB docs.

baileympearson:
@durran 

> > The spec changes say we need to document the behavior (in addition to the release notes). Should we modify the comment on `Collection.estimatedDocumentCount` to include the relevant information?
> > see [https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst?rgh-link-date=2022-05-16T19%3A00%3A32Z#estimateddocumentcount)
> 
> I've updated the TSDoc to note this and link to the MongoDB docs.

There's also this blurb from the spec:

> Drivers MUST document that, due to an oversight in versions 5.0.0-5.0.8 of MongoDB, the count command, which estimatedDocumentCount uses in its implementation, was not included in v1 of the Stable API, and so users of the Stable API with estimatedDocumentCount are recommended to upgrade their server version to 5.0.9+ or set apiStrict: false to avoid encountering errors.

Should we mention this in our comment as well?

src/operations/estimated_document_count.ts
```diff
@@ -66,12 +40,12 @@ export class EstimatedDocumentCountOperation extends CommandOperation<number> {
...
 
     super.executeCommand(server, session, cmd, (err, response) => {
-      if (err) {
+      if (err && (err as MongoServerError).code !== 26) {
```
baileympearson, line 43: Should `(err as MongoServerError).code !== 26` exist anymore?  This logic was used with the `$collStats`-based version of `estimatedDocumentCount`.  Since we're no longer using the `$collStats`-based version, I don't think we should keep this logic.

durran: Removed.

durran:
> @durran
> 
> > > The spec changes say we need to document the behavior (in addition to the release notes). Should we modify the comment on `Collection.estimatedDocumentCount` to include the relevant information?
> > > see [https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst?rgh-link-date=2022-05-16T19%3A00%3A32Z#estimateddocumentcount)
> > 
> > 
> > I've updated the TSDoc to note this and link to the MongoDB docs.
> 
> There's also this blurb from the spec:
> 
> > Drivers MUST document that, due to an oversight in versions 5.0.0-5.0.8 of MongoDB, the count command, which estimatedDocumentCount uses in its implementation, was not included in v1 of the Stable API, and so users of the Stable API with estimatedDocumentCount are recommended to upgrade their server version to 5.0.9+ or set apiStrict: false to avoid encountering errors.
> 
> Should we mention this in our comment as well?

Put in TSDoc as well now.

baileympearson (dismissed):

test/integration/collection-management/collection.test.js
```diff
@@ -535,81 +535,81 @@ describe('Collection', function () {
     });
   });
 
-  describe('(countDocuments)', function () {
```
durran, line 538: These tests pass locally but fail on CI so I just rewrote them. They also weren't testing anything in some cases since any type of count would return `0` if no documents were inserted.

PR #3248, opened 2022-05-16
refactor(NODE-2993): align internal cmap implementation with spec

dariakp:
### Description
NODE-2993 part 1; also fixes a bug where the `totalConnectionCount` would double count the available connections tracked by `kConnections`

#### What is changing?
- Replaced `kPermit` logic: added a new property to track pending connections and updated the existing checkout connection tracking to be more accurate
- Refactored CMAP spec runner: before/after logic now happens in the runner file so that it can be reused with our own "spec"-like tests
- The CMAP test runner can now optionally inject pool stats into pool events; we use this to check that the connection counts for pending, available, and checked out are updated correctly at each step (NOTE: at the point of the `ConnectionCreatedEvent`, the connection is still pending and continues to contribute to the pending count until it is either checked out or added to the available connection pool) 

No

Make it easier to implement the maxConnecting spec.

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -39,7 +39,7 @@ const kLogger = Symbol('logger');
 /** @internal */
 const kConnections = Symbol('connections');
 /** @internal */
-const kPermits = Symbol('permits');
+const kPending = Symbol('pending');
 /** @internal */
 const kMinPoolSizeTimer = Symbol('minPoolSizeTimer');
 /** @internal */
@@ -112,11 +112,10 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
   [kLogger]: Logger;
   /** @internal */
   [kConnections]: Denque<Connection>;
-  /**
-   * An integer expressing how many total connections are permitted
-   * @internal
-   */
-  [kPermits]: number;
+  /** @internal */
+  [kPending]: number;
+  /** @internal */
+  [kCheckedOut]: number;
   /** @internal */
   [kMinPoolSizeTimer]?: NodeJS.Timeout;
   /**
@@ -136,8 +135,6 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
   [kWaitQueue]: Denque<WaitQueueMember>;
   /** @internal */
   [kMetrics]: ConnectionPoolMetrics;
-  /** @internal */
-  [kCheckedOut]: number;
   /** @internal */
   [kProcessingWaitQueue]: boolean;
 
@@ -216,7 +213,8 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
 
     this[kLogger] = new Logger('ConnectionPool');
     this[kConnections] = new Denque();
-    this[kPermits] = this.options.maxPoolSize;
+    this[kPending] = 0;
+    this[kCheckedOut] = 0;
     this[kMinPoolSizeTimer] = undefined;
     this[kGeneration] = 0;
     this[kServiceGenerations] = new Map();
@@ -225,7 +223,6 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     this[kCancellationToken].setMaxListeners(Infinity);
     this[kWaitQueue] = new Denque();
     this[kMetrics] = new ConnectionPoolMetrics();
-    this[kCheckedOut] = 0;
     this[kProcessingWaitQueue] = false;
 
     process.nextTick(() => {
@@ -244,14 +241,24 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     return this[kGeneration];
   }
 
-  /** An integer expressing how many total connections (active + in use) the pool currently has */
+  /** An integer expressing how many total connections (available + pending + in use) the pool currently has */
   get totalConnectionCount(): number {
-    return this[kConnections].length + (this.options.maxPoolSize - this[kPermits]);
+    return (
+      this.availableConnectionCount + this.pendingConnectionCount + this.currentCheckedOutCount
+    );
   }
 
   /** An integer expressing how many connections are currently available in the pool. */
   get availableConnectionCount(): number {
     return this[kConnections].length;
+  }
+
+  get pendingConnectionCount(): number {
+    return this[kPending];
+  }
+
+  get currentCheckedOutCount(): number {
+    return this[kCheckedOut];
   }
 
   get waitQueueSize(): number {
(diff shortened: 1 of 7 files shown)


test/tools/cmap_spec_runner.ts
```diff
@@ -135,31 +137,35 @@ class Thread {
  * Implements the spec test match function, see:
  * [CMAP Spec Test README](https://github.com/mongodb/specifications/tree/master/source/connection-monitoring-and-pooling/tests#spec-test-match-function)
  */
-const compareInputToSpec = (input, expected) => {
+const compareInputToSpec = (input, expected, message) => {
   // the spec uses 42 and "42" as special keywords to express that the value does not matter
   // however, "42" does not appear in the spec tests, so only the numeric value is checked here
   if (expected === 42) {
-    expect(input).to.be.ok; // not null or undefined
+    expect(input, message).to.be.ok; // not null or undefined
```
nbbeeken, line 144:
```suggestion
expect(input, message).to.not.be.null;
expect(input, message).to.not.be.undefined;
```

Should we narrow this to gaurd against the other falsey values in JS that we **woudn't** want accepted here?

test/tools/cmap_spec_runner.ts
```diff
@@ -241,16 +247,22 @@ export class ThreadContext {
...
-  constructor(hostAddress: HostAddress, poolOptions: Partial<ConnectionPoolOptions> = {}) {
+  constructor(
+    hostAddress: HostAddress,
+    poolOptions: Partial<ConnectionPoolOptions> = {},
+    contextOptions: { injectPoolStats: boolean }
+  ) {
     this.#poolOptions = poolOptions;
     this.#hostAddress = hostAddress;
     this.#supportedOperations = getTestOpDefinitions(this);
+    this.#injectPoolStats = contextOptions.injectPoolStats;
```
nbbeeken, line 265:
```suggestion
this.#injectPoolStats ??= contextOptions.injectPoolStats;
```

The class default will always be overriden, and the spec test leaves this option undefined. Nothing breaks becasue of faleyness but might as well make sure this is set correctly.

test/tools/cmap_spec_runner.ts
```diff
@@ -372,6 +390,102 @@ export async function runCmapTest(test: CmapTest, threadContext: ThreadContext)
...
+        let utilClient: MongoClient;
+
+        const skipDescription = options?.testsToSkip?.find(
+          ({ description }) => description === test.description
+        );
+        if (skipDescription) {
+          const matchesLoadBalanceSkip =
+            skipDescription.skipIfCondition === 'loadBalanced' && this.configuration.isLoadBalanced;
+          if (matchesLoadBalanceSkip) {
+            (this.currentTest as Mocha.Test & { skipReason?: string }).skipReason =
```
nbbeeken, line 421: If you open the global.d.ts file you should be able to get the correct type info, not ideal but I'd rather not add casting. Or we can mark it with `TODO(NODE-4247)`

test/tools/cmap_spec_runner.ts
```diff
@@ -372,6 +390,102 @@ export async function runCmapTest(test: CmapTest, threadContext: ThreadContext)
...
+        const allRequirements = test.runOn || [];
+
+        const someRequirementMet =
+          !allRequirements.length ||
+          (await isAnyRequirementSatisfied(this.currentTest.ctx, allRequirements, utilClient));
+
+        if (!someRequirementMet) {
+          await utilClient.close();
+          this.skip();
+          // NOTE: the rest of the code below won't execute after the skip is invoked
```
nbbeeken, line 445: I know TS warns you but another way we can help indicate this to readers is to `return this.skip()` but no need to change.

nbbeeken (changes requested):
small notes, looks good!

src/cmap/connection_pool.ts
```diff
@@ -339,7 +341,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       this[kConnections].unshift(connection);
     }
 
-    this[kCheckedOut] = this[kCheckedOut] - 1;
+    this[kCheckedOut]--;
```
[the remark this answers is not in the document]
dariakp: I only changed it to make the style consistent on the properties that I touched here (i.e., connection counting related); I don't think we need to touch the LB generation code

test/tools/cmap_spec_runner.ts
dariakp: We should have caught this earlier, but better late than never :) - thought for some reason that it was just null/undefined. We do *not* want to throw on falsy values here unless they're explicitly null or undefined (per the [spec match function description](https://github.com/mongodb/specifications/tree/master/source/connection-monitoring-and-pooling/tests#spec-test-match-function)). I'll update to your suggestion, but not via github because I think the spacing will be off.

test/tools/cmap_spec_runner.ts
dariakp: it's actually because of the runner - look in line 456, but no harm with the extra guard

test/tools/cmap_spec_runner.ts
dariakp: are you asking me to add `currentTest` to that list with our `Mocha.Runnable` type? because if not, it would still be a cast here

test/tools/cmap_spec_runner.ts
dariakp: I've considered it (when I first wrote this in the other PR), but I went against it because the return type for these functions is supposed to be `void`, so returning `this.skip()` would contradict the type, and placing a return after the skip highlights it as unreachable, hence the comment instead

test/tools/cmap_spec_runner.ts
nbbeeken: `currentTest` is typed as `Mocha.Test` which extends `Mocha.Runnable` which is where we define the `skipReason`, so it should be available on this variable. Is it giving you type error anyway?

test/tools/cmap_spec_runner.ts
dariakp: Actually, this particular guard would be not what we want, since it would prevent overriding the default... the nullish coalescing compares the left hand side, not the right hand side; if you'd like, I can set it to `!!contextOptions.injectPoolStats` here and remove the defaulting on the class

test/tools/cmap_spec_runner.ts
dariakp: I wouldn't have updated this line if `this.currentTest.skipReason =` didn't yield a red squiggly

dariakp: (Not sure where you're seeing the extends Runnable definitions here, but they're not getting picked up)

nbbeeken (approved):
LGTM

baileympearson (approved):

PR #3251, opened 2022-05-18
refactor(NODE-4246): update FLE2 collection management

addaleax:

No new docs

See https://github.com/mongodb/specifications/pull/1215/files

### DIFF
--- a/src/operations/drop.ts
+++ b/src/operations/drop.ts
@@ -51,24 +51,6 @@ export class DropCollectionOperation extends CommandOperation<boolean> {
         encryptedFields = listCollectionsResult?.[0]?.options?.encryptedFields;
       }
 
-      let result;
-      let errorForMainOperation;
-      try {
-        result = await this.executeWithoutEncryptedFieldsCheck(server, session);
-      } catch (err) {
-        if (
-          !encryptedFields ||
-          !(err instanceof MongoServerError) ||
-          err.code !== MONGODB_ERROR_CODES.NamespaceNotFound
-        ) {
-          throw err;
-        }
-        // Save a possible NamespaceNotFound error for later
-        // in the encryptedFields case, so that the auxilliary
-        // collections will still be dropped.
-        errorForMainOperation = err;
-      }
-
       if (encryptedFields) {
         const escCollection = encryptedFields.escCollection || `enxcol_.${name}.esc`;
         const eccCollection = encryptedFields.eccCollection || `enxcol_.${name}.ecc`;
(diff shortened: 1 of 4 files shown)


durran:
Just wanting to get the csfle tests passing in evergreen first.

durran (dismissed):
@addaleax can you rebase now so the tests pass then LGTM. Thanks!

addaleax:
@durran Sure, done!

src/operations/drop.ts
```diff
@@ -88,13 +70,9 @@ export class DropCollectionOperation extends CommandOperation<boolean> {
...
           }
         }
-
-        if (errorForMainOperation) {
-          throw errorForMainOperation;
-        }
       }
 
-      return result;
+      return await this.executeWithoutEncryptedFieldsCheck(server, session);
```
dariakp, line 75: Just wanted to clarify - when the spec says that ``namespace not found`` errors must be ignored in the list of operations, the last of which is to drop the <collectionName> collection - is that main collection included in the ignore errors requirement? If so, that's going to be an issue here, though we *are* intending to update the behavior eventually to suppress namespace not found errors always: https://jira.mongodb.org/browse/NODE-4243

dariakp:
@durran @addaleax Just a note that #3262 pulls in changes to the same files (I haven't compared to check if they are the same changes), so whichever PR goes in first, the other will need to be rebased

src/operations/drop.ts
addaleax: I don't think it would be a good choice of behavior if `.dropCollection()` were to return/not return an error depending on whether the collection has an `encryptedFields` config. The point of the special FLE2 behavior of the drop/create collection methods is that the auxilliary collections are managed properly, not that FLE2 collections require special treatment in general.

It's great that NODE-4243 is being planned, it's definitely what I would expect and seems to match (at least some) other drivers as well, and as far as I concerned it's also a good answer to this question. :)

dariakp: I agree that it would be suboptimal UX to have diverging behavior, I was asking more to see whether we *need* to pull in NODE-4243 to actually be spec compliant here

addaleax: @dariakp For context, here’s the original discussion that led to the addition of that note in the spec: https://github.com/mongodb/specifications/pull/1215#discussion_r874581207 (note that Kevin did check the previous behavior of the Node.js driver code as well). The important part here is that dropping the auxiliary collections should not prevent dropping of the data collection.

I feel confident saying that NODE-4243 is not necessary for FLE2 spec compliancy specifically.

durran:
@dariakp I'm happy to just merge this one and close the other so no rebase needs to happen. Are you fine with that?

dariakp:
@durran Yeah, that works for me

PR #3255, opened 2022-05-19
feat(NODE-2993): implement maxConnecting

dariakp:
### Description
NODE-2993

#### What is changing?
- Connection pool can now attempt to check out up to maxConnecting number of connections in parallel (previously it would only establish the requested connections one at a time)
- ensureMinPoolSize will now only attempt to create one connection at a time so as not to block checkout requests from obtaining connections as quickly as possible (since ensureMinPoolSize always adds the connections to the available pool first, a checkout request would otherwise have to wait to retrieve it from the pool instead of obtaining it immediately; this nonblocking behavior is relied upon by the "threads blocked by maxConnecting check out minPoolSize connections" spec test)
- The ConnectionCreatedEvent is now fired *before* actually establishing the connection, as required by the spec (the maxConnecting tests rely on this to differentiate between pending and established connections)
- There was a bug in the cmap runner where the operations weren't actually queued using the main thread but rather in parallel across threads, this PR includes a fix for that
- Note: the new waitQueueTimeoutMS test is skipped because it is not applicable given our waitQueueTimeoutMS implementation (see the corresponding test yaml file for the relevant note)

No

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -537,6 +537,12 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
   };
 
   pool[kPending]++;
+  // This is our version of a "virtual" no-I/O connection as the spec requires
+  pool.emit(
+    ConnectionPool.CONNECTION_CREATED,
+    new ConnectionCreatedEvent(pool, { id: connectOptions.id } as Connection)
+  );
+
   connect(connectOptions, (err, connection) => {
     if (err || !connection) {
       pool[kPending]--;
@@ -559,8 +565,6 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
     for (const event of [...APM_EVENTS, Connection.CLUSTER_TIME_RECEIVED]) {
       connection.on(event, (e: any) => pool.emit(event, e));
     }
-
-    pool.emit(ConnectionPool.CONNECTION_CREATED, new ConnectionCreatedEvent(pool, connection));
 
     if (pool.loadBalanced) {
       connection.on(Connection.PINNED, pinType => pool[kMetrics].markPinned(pinType));
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -508,9 +508,9 @@ function ensureMinPoolSize(pool: ConnectionPool) {
     // connection permits because that potentially delays the availability of
     // the connection to a checkout request
     createConnection(pool, (err, connection) => {
+      pool[kPending]--;
       if (!err && connection) {
         pool[kConnections].push(connection);
-        pool[kPending]--;
         process.nextTick(processWaitQueue, pool);
       }
       pool[kMinPoolSizeTimer] = setTimeout(() => ensureMinPoolSize(pool), 10);
@@ -533,7 +533,7 @@ function connectionIsIdle(pool: ConnectionPool, connection: Connection) {
   return !!(pool.options.maxIdleTimeMS && connection.idleTime > pool.options.maxIdleTimeMS);
 }
 
-function createConnection(pool: ConnectionPool, callback?: Callback<Connection>) {
+function createConnection(pool: ConnectionPool, callback: Callback<Connection>) {
   const connectOptions: ConnectionOptions = {
     ...pool.options,
     id: pool[kConnectionCounter].next().value,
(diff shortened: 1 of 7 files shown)


src/cmap/connection_pool.ts
```diff
@@ -107,7 +107,7 @@ export type ConnectionPoolEvents = {
  */
 export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
   closed: boolean;
-  options: Readonly<ConnectionPoolOptions>;
+  options: Readonly<ConnectionPoolOptions & { maxConnecting: number }>;
```
dariakp, line 110: This is temporary and will move properly into `ConnectionPoolOptions` with the next ticket that makes maxConnecting configurable

src/cmap/connection_pool.ts
```diff
@@ -530,6 +537,12 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
   };
 
   pool[kPending]++;
+  // This is our version of a "virtual" no-I/O connection as the spec requires
```
dariakp, line 540: It would be more optimal to actually create a connection object here, so I'm open to other ideas, but all of the "better" approaches I can think of require an extensive refactor of the underlying connect method, that is several levels removed from the branching logic that calls different connection class constructors depending on the circumstances

src/cmap/connection_pool.ts
```diff
@@ -530,6 +537,12 @@ function createConnection(pool: ConnectionPool, callback?: Callback<Connection>)
...
+  pool.emit(
+    ConnectionPool.CONNECTION_CREATED,
+    new ConnectionCreatedEvent(pool, { id: connectOptions.id } as Connection)
```
nbbeeken, line 543: From tech chat: We can change the constructor to accept `{id: number|string}` and it will still accept a connection. It's internal so we're fine to do this.

src/cmap/connection_pool.ts
```diff
@@ -642,44 +653,49 @@ function processWaitQueue(pool: ConnectionPool) {
     }
   }
 
-  const maxPoolSize = pool.options.maxPoolSize;
-  if (pool.waitQueueSize && (maxPoolSize <= 0 || pool.totalConnectionCount < maxPoolSize)) {
+  const { maxPoolSize, maxConnecting } = pool.options;
+  while (
+    pool.waitQueueSize &&
```
nbbeeken, line 658:
```suggestion
    pool.waitQueueSize !== 0 &&
```

src/cmap/connection_pool.ts
```diff
@@ -642,44 +653,49 @@ function processWaitQueue(pool: ConnectionPool) {
...
+    pool.pendingConnectionCount < maxConnecting &&
+    pool.availableConnectionCount <= 0 &&
+    (maxPoolSize <= 0 || pool.totalConnectionCount < maxPoolSize)
```
nbbeeken, line 661: When does maxPoolSize go negative? should this be === 0?

src/cmap/connection_pool.ts
```diff
@@ -642,44 +653,49 @@ function processWaitQueue(pool: ConnectionPool) {
...
+            new ConnectionCheckOutFailedEvent(pool, err)
+          );
+        } else if (connection) {
+          pool[kCheckedOut]++;
+          pool[kPending]--;
+          pool.emit(
+            ConnectionPool.CONNECTION_CHECKED_OUT,
+            new ConnectionCheckedOutEvent(pool, connection)
+          );
+        }
```
nbbeeken, line 688: This is one of those code structures caused by TS. err and connection are mutually exclusive, we should have written something more like:
```js
if (err || !connection) {
  // err logic
  return
}
```
maybe that's why the early return looked erroneous?

src/cmap/connection_pool.ts
dariakp: this is just a holdover of the original logic (see L646 of the original code)

src/cmap/connection_pool.ts
dariakp: no, the early return (originally L656) was erroneous because if we encounter a timed out queue member we wouldn't try to process anymore of the waitqueue; ostensibly, if a queue member timed out, any remaining queue members are also timed out, but that doesn't mean we should just leave them sitting in the queue until some other event re-triggers its processing.

src/cmap/connection_pool.ts
dariakp: done

src/cmap/connection_pool.ts
dariakp: updated

nbbeeken (dismissed):

PR #3256, opened 2022-05-19
feat(NODE-4139): streaming protocol message changes

durran:
### Description

Only emits the last message from the message stream's buffer when using the streaming protocol.

#### What is changing?

The monitor will now tell the connection that it's using the streaming protocol, which in turn will
set the flag on the connection's message stream. When the stream is in this mode, it will only ever
emit the message event with the very last message in the buffer pool. So if there's more data in the
buffer, the stream essentially skips its emit.

None

NODE-4139

### DIFF
--- a/src/cmap/message_stream.ts
+++ b/src/cmap/message_stream.ts
@@ -180,7 +180,7 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>) 
       }
     }
     return false;
-  }
+  };
 
   let ResponseType = messageHeader.opCode === OP_MSG ? BinMsg : Response;
   if (messageHeader.opCode !== OP_COMPRESSED) {
--- a/src/cmap/message_stream.ts
+++ b/src/cmap/message_stream.ts
@@ -54,7 +54,7 @@ export class MessageStream extends Duplex {
   /** @internal */
   [kBuffer]: BufferPool;
   /** @internal */
-  isStreamingProtocol = false;
+  isMonitoringConnection = false;
 
   constructor(options: MessageStreamOptions = {}) {
     super(options);
@@ -168,15 +168,14 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>) 
     opCode: message.readInt32LE(12)
   };
 
-  const streamingProtocolHasAnotherHello = () => {
-    if (stream.isStreamingProtocol) {
+  const monitorHasAnotherHello = () => {
+    if (stream.isMonitoringConnection) {
       // Can we read the next message size?
       if (buffer.length >= 4) {
         const sizeOfMessage = buffer.peek(4).readInt32LE();
         if (sizeOfMessage < buffer.length) {
           return true;
         }
-        return false;
       }
     }
     return false;
@@ -186,10 +185,10 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>) 
   if (messageHeader.opCode !== OP_COMPRESSED) {
     const messageBody = message.slice(MESSAGE_HEADER_SIZE);
 
-    // If we are a monitoring message stream using the streaming protocol and
+    // If we are a monitoring connection message stream and
     // there is more in the buffer that can be read, skip processing since we
     // want the last hello command response that is in the buffer.
-    if (streamingProtocolHasAnotherHello()) {
+    if (monitorHasAnotherHello()) {
       processIncomingData(stream, callback);
     } else {
       stream.emit('message', new ResponseType(message, messageHeader, messageBody));
@@ -226,10 +225,10 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>) 
       return;
     }
 
-    // If we are a monitoring message stream using the streaming protocol and
+    // If we are a monitoring connection message stream and
     // there is more in the buffer that can be read, skip processing since we
     // want the last hello command response that is in the buffer.
-    if (streamingProtocolHasAnotherHello()) {
+    if (monitorHasAnotherHello()) {
       processIncomingData(stream, callback);
     } else {
       stream.emit('message', new ResponseType(message, messageHeader, messageBody));
--- a/src/cmap/message_stream.ts
+++ b/src/cmap/message_stream.ts
@@ -173,7 +173,7 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>) 
       // Can we read the next message size?
       if (buffer.length >= 4) {
         const sizeOfMessage = buffer.peek(4).readInt32LE();
-        if (sizeOfMessage < buffer.length) {
+        if (sizeOfMessage <= buffer.length) {
           return true;
         }
       }
--- a/src/cmap/message_stream.ts
+++ b/src/cmap/message_stream.ts
@@ -53,11 +53,17 @@ export class MessageStream extends Duplex {
   maxBsonMessageSize: number;
   /** @internal */
   [kBuffer]: BufferPool;
+  /** @internal */
+  isStreamingProtocol = false;
 
   constructor(options: MessageStreamOptions = {}) {
     super(options);
     this.maxBsonMessageSize = options.maxBsonMessageSize || kDefaultMaxBsonMessageSize;
     this[kBuffer] = new BufferPool();
+  }
+
+  get buffer(): BufferPool {
+    return this[kBuffer];
   }
 
   override _write(chunk: Buffer, _: unknown, callback: Callback<Buffer>): void {
@@ -165,12 +171,20 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>) 
   let ResponseType = messageHeader.opCode === OP_MSG ? BinMsg : Response;
   if (messageHeader.opCode !== OP_COMPRESSED) {
     const messageBody = message.slice(MESSAGE_HEADER_SIZE);
-    stream.emit('message', new ResponseType(message, messageHeader, messageBody));
 
-    if (buffer.length >= 4) {
+    // If we are a monitoring message stream using the streaming protocol and
+    // there is more in the buffer that can be read, skip processing since we
+    // want the last hello command response that is in the buffer.
+    if (stream.isStreamingProtocol && buffer.length >= 4) {
       processIncomingData(stream, callback);
     } else {
-      callback();
+      stream.emit('message', new ResponseType(message, messageHeader, messageBody));
+
+      if (buffer.length >= 4) {
+        processIncomingData(stream, callback);
+      } else {
+        callback();
+      }
     }
 
     return;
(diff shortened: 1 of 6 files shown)


test/unit/cmap/message_stream.test.js
```diff
@@ -19,6 +19,34 @@ function bufferToStream(buffer) {
 }
 
 describe('Message Stream', function () {
+  context('when the stream uses the streaming protocol', function () {
```
dariakp, line 22: could be good to add an analogous test to show that all messages are read outside the streaming protocol

test/unit/cmap/message_stream.test.js
```diff
@@ -19,6 +19,34 @@ function bufferToStream(buffer) {
...
+    it('only reads the last message in the buffer', function (done) {
+      const inputStream = bufferToStream(data);
+      const messageStream = new MessageStream();
+      messageStream.isStreamingProtocol = true;
+
+      messageStream.once('message', msg => {
+        msg.parse();
+        expect(msg)
+          .to.have.property('documents')
+          .that.deep.equals([{ [LEGACY_HELLO_COMMAND]: 1 }]);
```
dariakp, line 40: might it be better to use a buffer that uses a regular hello command instead of legacy since the legacy one is what we're aiming to eventually remove?

test/unit/cmap/message_stream.test.js
```diff
@@ -19,6 +19,34 @@ function bufferToStream(buffer) {
...
+        '370000000100000001000000010000000000000000000000000000000000000001000000130000001069736d6173746572000100000000',
+      'hex'
+    );
+
+    it('only reads the last message in the buffer', function (done) {
+      const inputStream = bufferToStream(data);
+      const messageStream = new MessageStream();
+      messageStream.isStreamingProtocol = true;
+
+      messageStream.once('message', msg => {
```
dariakp, line 36: instead of `done`, I think we can use async await here to make the test and the assertions more straightforward via the `once` from node events module, i.e. `await once(messageStream, 'message')`

dariakp (changes requested):
just a couple of things on the tests; the implementation looks good to me

jyemin (changes requested):

test/unit/cmap/message_stream.test.js
durran: I've refactored all the tests now to make it more clear what's happening, including a utility function to generate OP_MSG buffers from a document. The document I use now is a simple response from a normal hello to test.

test/unit/cmap/message_stream.test.js
durran: Updated to async/await

src/cmap/message_stream.ts
```diff
@@ -165,12 +171,20 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>)
...
+    // there is more in the buffer that can be read, skip processing since we
+    // want the last hello command response that is in the buffer.
+    if (stream.isStreamingProtocol && buffer.length >= 4) {
       processIncomingData(stream, callback);
     } else {
-      callback();
+      stream.emit('message', new ResponseType(message, messageHeader, messageBody));
+
+      if (buffer.length >= 4) {
+        processIncomingData(stream, callback);
```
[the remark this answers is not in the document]
durran: That is correct.

src/cmap/message_stream.ts
```diff
@@ -165,12 +171,20 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>)
   let ResponseType = messageHeader.opCode === OP_MSG ? BinMsg : Response;
   if (messageHeader.opCode !== OP_COMPRESSED) {
     const messageBody = message.slice(MESSAGE_HEADER_SIZE);
-    stream.emit('message', new ResponseType(message, messageHeader, messageBody));
 
-    if (buffer.length >= 4) {
+    // If we are a monitoring message stream using the streaming protocol and
+    // there is more in the buffer that can be read, skip processing since we
+    // want the last hello command response that is in the buffer.
+    if (stream.isStreamingProtocol && buffer.length >= 4) {
```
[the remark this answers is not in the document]
durran: I did it like this so that if there is at least 4 bytes left in the buffer, that will be the 4 bytes that can be read for the size of the next message. If those 4 bytes exist, we know there is another whether it is entirely on the buffer or not and the code will recurse until it gets it. This way we avoid potentially reading more messages when there are partials and attempt to truly get the very last message that was sent.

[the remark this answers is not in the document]
jyemin: We don't want to get the the last message that was _sent_ (that is unknowable, actually).  We want to get the last message that was _received_.   So the check should be, semantically, whether the buffer contains another full message.  Only then should we skip the current message. This check seems straightforward to implement.  The logic 
 is basically already implemented earlier in the method.  It just needs to be extracted into a function that, perhaps, could be re-used.

durran:
Screenshot 2022-05-20 at 15 39 47

src/cmap/message_stream.ts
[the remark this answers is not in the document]
durran: So basically extract a method like `readLastMessage()` that checks the length of the next message in the buffer, checks if more than that number of bytes are present, then recursively checks the same again until a partial is there, and take the one before it?

jyemin (changes requested):
Looking to those logs, I wonder if we have another problem (which we could choose to ignore): could this change result in unmatched heartbeat started/succeeded events?  Or could that already happen even in the current implementation?

src/cmap/message_stream.ts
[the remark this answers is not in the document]
jyemin: That would work (though it doesn't have to be implemented recursively).

src/cmap/message_stream.ts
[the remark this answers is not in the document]
dariakp: Do we currently create a new connection every time we ping the server in the polling monitor case? If not, could we get into this situation if we have a chance to send multiple requests before processing the response?

Otherwise, this would only be applicable if we implement exhaust cursor support at some point. Perhaps we should call back with an error in this case so that we can detect if something unexpected is happening.

test/unit/cmap/message_stream.test.js
```diff
@@ -18,117 +20,101 @@ function bufferToStream(buffer) {
...
         messageStream.on('message', msg => {
           messageCount++;
-          if (error) {
-            done(new Error(`expected error: ${error}`));
-            return;
+          msg.parse();
+          expect(msg).to.have.property('documents').that.deep.equals([response]);
+          // Test will not complete until 3 messages processed.
+          if (messageCount === 3) {
+            done();
```
dariakp, line 75: you can actually use `on` as an async iterator (and you can probably use `once` for the error listeners) to make all these tests async/await: https://nodejs.org/api/events.html#eventsonemitter-eventname-options

dariakp (changes requested):

src/cmap/message_stream.ts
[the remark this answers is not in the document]
durran: No we use the same connection in the polling scenario but we won't get into this scenario as the streaming protocol flag wont be set.

src/cmap/message_stream.ts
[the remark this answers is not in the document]
durran: Updated the check to peek the size and added extra test for the case where a partial message is present.

test/unit/cmap/message_stream.test.js
durran: Updated to make all tests use async/await

src/cmap/message_stream.ts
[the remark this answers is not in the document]
jyemin: This thread can be resolved from my point of view.  Even though this is dead code, it's the way it's worked for a long time and doesn't need to be changed as part of this PR.

@dariakp when your follow-up question has been been addressed please resolve the thread.

src/cmap/message_stream.ts
```diff
@@ -162,15 +168,37 @@ function processIncomingData(stream: MessageStream, callback: Callback<Buffer>)
...
 
+  const streamingProtocolHasAnotherHello = () => {
+    if (stream.isStreamingProtocol) {
+      // Can we read the next message size?
+      if (buffer.length >= 4) {
+        const sizeOfMessage = buffer.peek(4).readInt32LE();
+        if (sizeOfMessage < buffer.length) {
+          return true;
+        }
+        return false;
```
jyemin, line 179: Nit.  You can remove this and just fall through.

src/sdam/monitor.ts
```diff
@@ -279,6 +279,10 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
       // if we are using the streaming protocol then we immediately issue another `started`
       // event, otherwise the "check" is complete and return to the main monitor loop
       if (isAwaitable && hello.topologyVersion) {
```
jyemin, line 281: I don't understand this conditional.  `isAwaitable` is initialized above with `monitor[kServer].description.topologyVersion != null`.  I think that is the description from the handshake.  But then we `and` that here with the new `hello.topologyVersion`.  For clarity I think that either `isAwaiting` should be removed from this conditional, or else we should just check `isAwaiting` and move the conditional up to line 235.

I'm not sure which is correct, or if both are correct, but there don't seem to be any tests that prove that  `connection.isStreamingProtocol` is ever set to `true`, so I have some cause for concern.

Perhaps it would be safer to just always set `connection.isStreamingProtocol` to `true` for all monitor connections.  It seems like that would be harmless (we could change the property is `isMonitorConnection`).

durran: I've updated per the suggestion to set the flag on all monitoring connections and added a unit test to the monitor to ensure its connection has the flag set. Property renamed to `isMonitoringConnection`.

jyemin (approved):
LGTM

test/unit/cmap/message_stream.test.js
```diff
@@ -18,117 +20,116 @@ function bufferToStream(buffer) {
...
+    });
 
-        let messageCount = 0;
-        messageStream.on('message', msg => {
-          messageCount++;
-          if (error) {
-            done(new Error(`expected error: ${error}`));
-            return;
-          }
+    it('only reads the last message in the buffer', async function () {
```
dariakp, line 39: Could we differentiate the content of the last message in some way so that we know that we in fact read the last message and not, say, the first?

dariakp (changes requested):
just one suggestion to better isolate the condition under test

src/cmap/message_stream.ts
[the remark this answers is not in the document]
dariakp: followed up async: we only send out the next polling request after receiving a response to the previous one, so no concerns there, resolving

test/unit/cmap/message_stream.test.js
durran: Your suggestion caught a bug. So thanks for that. :) https://github.com/mongodb/node-mongodb-native/pull/3256/commits/71fd458585c90c098e9690394ac572400b9c0c28

dariakp (approved):

PR #3261, opened 2022-05-23
feat(NODE-3750): make maxConnecting configurable

dariakp:
### Description
NODE-3750, changes implement the following spec commit: [ce242d](https://github.com/mongodb/specifications/commit/ce242ddbb5377c509057e4e07b9d56c011bb0229)

#### What is changing?
* Added maxConnecting as a pool option that is externally configurable as a client option
    * per the spec, maxConnecting <= 0 should warn, so, in line with our warning = error convention, maxConnecting <=0 will error (note that there are two different errors: one from uint validation, and one to handle 0 specifically)
* Synced new cmap test: pool-checkout-custom-maxConnecting-is-enforced
* Updated uri options connection-pool-options test file with new tests and implemented maxConnection option parsing in the test runner
* Updated existing node driver "all options" tests to include maxConnecting - there isn't any complicated option parsing happening, so it didn't seem to require an additional stand-alone test beyond what is covered by the URI spec tests themselves

Yes, in addition to the API docs, we should also update the connection docs page for the node driver since it lists other pool options there.

Spec


durran (changes requested):
Can we sync the yml test as well for uri options? (connection-pool-options.yml)

dariakp:
@durran good catch, I didn't even realize I missed them

src/connection_string.ts
```diff
@@ -856,6 +856,16 @@ export const OPTIONS = {
       return new Logger('MongoClient', { loggerLevel: value as LoggerLevel });
     }
   },
+  maxConnecting: {
+    default: 2,
+    transform({ name, values: [value] }): number {
+      const maxStalenessSeconds = getUint(name, value);
```
durran, line 862: Could we maybe rename this variable since it's not a time value but a count value.

durran (changes requested):

src/connection_string.ts
dariakp: Yes, definitely - I don't even know how that happened 😅

baileympearson (approved):

PR #3267, opened 2022-05-25
test(NODE-4274): fix `match` logic in unified spec runner

baileympearson:
#### What is changing?

This PR fixes a few bugs in our matching logic in the unified spec runner.

No.

Correct assertions in our spec tests.


test/tools/unified-spec-runner/match.ts
```diff
@@ -328,22 +408,31 @@ function compareEvents(
...
     const expectedEvent = expected[index];
+    const rootPrefix = `events[${index}]`;
 
     if (expectedEvent.commandStartedEvent) {
-      expect(actualEvent).to.be.instanceOf(CommandStartedEvent);
-      resultCheck(actualEvent, expectedEvent.commandStartedEvent, entities, [
-        `events[${index}].commandStartedEvent`
-      ]);
+      const path = `${rootPrefix}.commandStartedEvent`;
+      if (!(actualEvent instanceof CommandStartedEvent)) {
```
baileympearson, line 415: the `instanceof` + `expect.fail` allows typescript to narrow the type, so that the call to `compareCommand<type>Events` knows it has an object of the correct type.

test/tools/unified-spec-runner/match.ts
```diff
@@ -225,9 +249,7 @@ export function specialCheck(
     // $$unsetOrMatches
     if (actual === null || actual === undefined) return;
     else {
```
baileympearson, line 251: Depth represents how many levels deep in an object we are.

For `$$unsetOrMatches`, we don't want to update the depth because the `$$unsetOrMatches` operator isn't part of the document we're evaluating - it's more of a "meta" key telling us how to evaluate the object.

dariakp (changes requested):
I haven't had a chance to review everything yet, but I noticed the following

test/tools/unified-spec-runner/match.ts
baileympearson: Comment no longer relevant now that I've refactored to use `isRootDocument` instead of `depth`.

test/tools/unified-spec-runner/match.ts
```diff
@@ -139,78 +139,96 @@ export function resultCheck(
...
+
   if (typeof expected === 'object' && expected) {
     // Expected is an object
     // either its a special operator or just an object to check equality against
 
     if (isSpecialOperator(expected)) {
       // Special operation check is a base condition
       // specialCheck may recurse depending upon the check ($$unsetOrMatches)
-      specialCheck(actual, expected, entities, path, depth);
+      specialCheck(actual, expected, entities, path, false);
```
dariakp, line 169: should this be `isRootDocument` instead of `false`? There's a bunch of failing tests, and this is the logic that has changed since the last commit

dariakp (changes requested):

dariakp (approved):

nbbeeken (approved):

PR #3272, opened 2022-05-25
test: improve change stream tests

nbbeeken:
#### What is changing?

As precursor to startSession without connection work I had to fix up change stream tests

Consolidates the initialize logic for iterable mode and event emitter mode. Clarifies clean up routines, makes heavy use of promises, bundling up many async tasks into one entry / one exit procedures.

Need to file TODOs

### DIFF
--- a/test/integration/change-streams/change_stream.test.ts
+++ b/test/integration/change-streams/change_stream.test.ts
@@ -1,8 +1,7 @@
 import { strict as assert } from 'assert';
 import { expect } from 'chai';
-import * as crypto from 'crypto';
 import { on, once } from 'events';
-import { PassThrough, Transform } from 'stream';
+import { PassThrough } from 'stream';
 import { promisify } from 'util';
 
 import {
@@ -488,7 +487,7 @@ describe('Change Streams', { sessions: { skipLeakTests: true } }, function () {
   });
 
   it('should invalidate change stream on collection rename using event listeners', {
-    metadata: { requires: { topology: 'replicaset' } },
+    metadata: { requires: { topology: 'replicaset', mongodb: '>=4.2' } },
     async test() {
       const willBeChange = once(changeStream, 'change');
       await once(changeStream.cursor, 'init');
@@ -516,7 +515,7 @@ describe('Change Streams', { sessions: { skipLeakTests: true } }, function () {
   });
 
   it('should invalidate change stream on database drop using iterator form', {
-    metadata: { requires: { topology: 'replicaset' } },
+    metadata: { requires: { topology: 'replicaset', mongodb: '>=4.2' } },
     async test() {
       const db = client.db('droppableDb');
       const collection = db.collection('invalidateCallback');
@@ -762,47 +761,6 @@ describe('Change Streams', { sessions: { skipLeakTests: true } }, function () {
       expect(parsedEvent).to.have.nested.property('fullDocument.a', 1);
     }
   });
-
-  it.skip('should support piping of Change Streams through multiple pipes', {
-    metadata: { requires: { topology: 'replicaset' } },
-    async test() {
-      const cipher = crypto.createCipher('aes192', 'a password');
-      const decipher = crypto.createDecipher('aes192', 'a password');
-
-      // Make a stream transforming to JSON and piping to the file
-      const stream = changeStream.stream();
-      const basicStream = stream.pipe(
-        new Transform({
-          transform: (data, encoding, callback) => callback(null, JSON.stringify(data)),
-          objectMode: true
-        })
-      );
-      const pipedStream = basicStream.pipe(cipher).pipe(decipher);
-      pipedStream.setEncoding('utf8');
-
-      const dataIterator = on(pipedStream, 'data');
-      const willBeEnd = once(pipedStream, 'end');
-
-      await collection.insertMany([{ a: 1407 }]);
-      let dataEmitted = '';
-      let parsedData;
-      for await (const data of dataIterator) {
-        dataEmitted += data;
-        try {
-          parsedData = JSON.parse(dataEmitted);
-          break;
-        } catch {
-          continue;
-        }
-      }
-
-      stream.emit('end');
-      await willBeEnd;
-
-      expect(parsedData).to.have.nested.property('operationType', 'insert');
-      expect(parsedData).to.have.nested.property('fullDocument.a', 1407);
-    }
-  }).skipReason = 'I am unable to get the data event to emit';
 
   it('should maintain change stream options on resume', {
     metadata: { requires: { topology: 'replicaset' } },
@@ -1103,7 +1061,7 @@ describe('Change Streams', { sessions: { skipLeakTests: true } }, function () {
   });
 
   describe('Change Stream Resume Error Tests', function () {
-    it('should continue emitting change events after a resumable error', {
+    it.skip('should continue emitting change events after a resumable error', {
       metadata: { requires: { topology: 'replicaset' } },
       async test() {
         const changes = on(changeStream, 'change');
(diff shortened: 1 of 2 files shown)


test/integration/change-streams/change_stream.test.ts
```diff
@@ -696,392 +460,244 @@ describe('Change Streams', function () {
...
-        });
-
-        // Trigger the first database event
-        waitForStarted(changeStream, () => {
-          this.defer(database.collection('invalidateListeners').insertMany([{ a: 1 }]));
-        });
-      });
-    }
-  });
+    metadata: { requires: { topology: 'replicaset', mongodb: '>3.6' } },
```
baileympearson, line 491:
```suggestion
    metadata: { requires: { topology: 'replicaset' } },
```

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1434,240 +986,180 @@ describe('Change Streams', function () {
...
-            expect(docs[0]).to.containSubset({
-              operationType: 'insert',
-              fullDocument: { a: 42 }
-            });
-            expect(docs[1]).to.containSubset({
-              operationType: 'insert',
-              fullDocument: { b: 24 }
-            });
-            done();
+      metadata: { requires: { topology: 'replicaset', mongodb: '>3.6' } },
```
baileympearson, line 1107:
```suggestion
      metadata: { requires: { topology: 'replicaset' } },
```

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1434,240 +986,180 @@ describe('Change Streams', function () {
...
-                    fullDocument: { b: 24 }
-                  });
-                  done();
-                });
-              });
-              collection.insertOne({ b: 24 });
-            });
-          });
-        });
+      metadata: { requires: { topology: 'replicaset', mongodb: '>3.6' } },
```
baileympearson, line 1132:
```suggestion
      metadata: { requires: { topology: 'replicaset' } },
```

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1679,19 +1171,19 @@ describe('Change Streams', function () {
...
+        //     expect(result).to.exist;
+        //     triggerResumableError(changeStream, 250, () => {
+        //       collection.insertOne({ x: 2 }, (err, result) => {
+        //         expect(err).to.not.exist;
+        //         expect(result).to.exist;
+        //       });
+        //     });
+        //   });
+        // });
+      }
```
baileympearson, line 1186: Did you mean to delete this?

test/integration/change-streams/change_stream.test.ts
nbbeeken: `>3.6` holds significance here and the other cases, this test doesn't pass on 3.6 it did not seem worth digging into why for now. I can file TODOs but the tests themselves haven't been evaluated for merit so I feel like it's an acceptable gap. Thoughts?

test/integration/change-streams/change_stream.test.ts
nbbeeken: No, the test is skipped and connected to a ticket, but I didn't want to take the time to fix it since it is skipped, I have to comment this out because it's using functions that no longer exist.

baileympearson: Ah I see now.  Makes sense!

baileympearson: Going to leave this open in case Daria wonders too.

test/integration/change-streams/change_stream.test.ts
baileympearson: Agreed, I think that's fine.  Thanks for explaining

baileympearson (approved):

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1127,108 +743,70 @@ describe('Change Streams', function () {
...
 
-        waitForStarted(changeStream, () => {
-          this.defer(collection.insertMany([{ a: 1407 }]));
-        });
-      });
+      expect(parsedData).to.have.nested.property('operationType', 'insert');
+      expect(parsedData).to.have.nested.property('fullDocument.a', 1407);
     }
-  });
+  }).skipReason = 'I am unable to get the data event to emit';
```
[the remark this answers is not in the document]
dariakp: I don't understand what the test is actually trying to test - we have a test above that ensuring that piping works. Although, it's using the iteration mode init - what's the purpose of init iterator vs not? Should we have a piping test without the iterator?

test/integration/change-streams/change_stream.test.ts
dariakp: continuing to emit change events after a resumable error seems important; 

and I disagree with the general approach of changing the requirements to `>= 3.6` elsewhere - either the test is worth having, in which case it needs to work on all our supported servers, including 3.6, or it's not worth having, in which case we should be getting rid of it entirely; doing `>3.6` just makes it confusing and immediately raises the question of why the test isn't applicable on 3.6.

there are a total of 6 tests that have been updated in this way - it should not add that much scope to read through them and figure out whether the cases as stated are worth testing; and for the stated cases that are worth testing, to determine whether the approach in the test is correct or should be adjusted, and then either make the changes or file follow ups as needed

dariakp (changes requested):

test/integration/change-streams/change_stream.test.ts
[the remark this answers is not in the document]
nbbeeken: This is testing the node stream mode, which isn't specific to change streams it's on the abstract cursor so we have coverage elsewhere that doesn't use deprecated crypto methods.

durran:
If we are refactoring the tests in this PR should we also make them adhere to our new test structure guidelines?

test/integration/change-streams/change_stream.test.ts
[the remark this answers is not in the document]
dariakp: If we are confident that we cover this elsewhere, we don't need to keep the test

test/integration/change-streams/change_stream.test.ts
nbbeeken: 3.6 doesn't emit invalidate events which some of these tests rely on, so I've marked those as 4.2+
The resumability ones timeout on 3.6, investigating, stay tuned

dariakp: but why did the original tests work? or were they skipped? or just not asserting correctly?

test/integration/change-streams/change_stream.test.ts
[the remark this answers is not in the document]
nbbeeken: Yea, I've removed it

test/integration/change-streams/change_stream.test.ts
nbbeeken: The resume tests are strange, they don't simulate the network error coming from the network calls, I had Bailey look over my shoulder for this one and he's pointing out that the tests `emit('error'` which circumvents all the cursor cleanup logic that would be done by abstract cursor. So I've skipped them with a todo of the fixing resumability ticket we're getting to soon.

And for the invalidate one's sorry I flipped it, rename is the event that was introduced in 4.0.1, the test before didn't look for rename it just ran the event listener until it saw invalidate and then would call done, so I have added a new assertion since with promises I have to iterate passed the rename event, I can revert this but I think we on the border of testing server behavior here anyway.

nbbeeken:
> If we are refactoring the tests in this PR should we also make them adhere to our new test structure guidelines?
I'd rather not do that now as this is blocking higher priority work, we'll be iterating on the change stream tests with our scheduled feature work and bug fixes so right now I just want to address the goal of getting our tests in a state where the clean up logic is traceable.

baileympearson (approved):

PR #3275, opened 2022-05-30
fix(NODE-4281): ensure that the driver always uses Node.js timers

addaleax:
#### What is changing?

Ensure that the driver always uses the Node.js timers API, rather
than the global one, in case they diverge. This affects Compass,
where the `setTimeout(...).unref()` usage currently results in
uncaught exceptions because `setTimeout()` returns a number in
browsers.

This change adds `import ... from 'timers';` where necessary and
adds a linter rule to prevent regressions. If this is not
an acceptable solution, we can go back to the drawing board,
but this seems like a good solution that doesn’t add too much
overhead when writing driver code.

No

See ticket – `setTimeout().unref()` doesn’t work in Compass renderers

### DIFF
--- a/test/unit/timer_sandbox.js
+++ b/test/unit/timer_sandbox.js
@@ -1,0 +1,22 @@
+'use strict';
+const sinon = require('sinon');
+
+/**
+ * sinon.useFakeTimers() only affects global methods, this function
+ * creates a sinon sandbox that ensures that require('timers')
+ * also uses the mocked variants.
+ *
+ * @returns {sinon.SinonSandbox}
+ */
+exports.createTimerSandbox = () => {
+  const timerSandbox = sinon.createSandbox();
+  const timers = require('timers');
+  for (const method in timers) {
+    if (method in global) {
+      timerSandbox.replace(timers, method, (...args) => {
+        return global[method](...args);
+      });
+    }
+  }
+  return timerSandbox;
+};
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -38,6 +38,21 @@
         "property": "only"
       }
     ],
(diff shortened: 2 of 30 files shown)


durran:
Just for clarification, this is specifically for when the driver is used in the Electron renderer process, where the global functions are the browser functions? And we still have Node available so simply requiring 'timers' gets us the Node specific ones in that environment?

addaleax:
@durran Yes, exactly :+1:

durran (dismissed):

test/unit/timer_sandbox.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+const sinon = require('sinon');
+
```
baileympearson, line 3: I'm okay with this approach but could another approach be only linting for global setTimeout/Immediate/Intervals in the `src` directory?  As long as Compass isn't running our tests I think that should be okay.

I don't have a strong preference, but the benefit of only linting the `src` directory is one fewer custom test configurations to maintain.  @addaleax

nbbeeken (approved):

test/unit/timer_sandbox.js
addaleax: @baileympearson I’d also be okay with applying the linting only there, I guess what would mean a new `.eslintrc` file inside `src`? That should be easy enough.

That being said, this particular file would still be necessary, because it’s about how sinon mocks the timers functions (which are then picked up by files in the `src/` directory)

baileympearson: Oh, right.  That makes sense.

If we need this file either way, I'd rather lint everything for consistency 😃

baileympearson (dismissed):

PR #3276, opened 2022-05-31
test(NODE-4270): convert command monitoring tests to unified format

durran:
### Description

Converts all command monitoring spec tests to the unified format.

#### What is changing?

Adds the new unified tests, removes the legacy tests and runner.

No

NODE-4270


test/spec/command-monitoring/unified/server-connection-id.json
```diff
@@ -0,0 +1,101 @@
+{
+  "description": "server-connection-id",
```
dariakp, line 2: This test should be done as part of NODE-3449 / DRIVERS-1141, I don't think we should sync it here unless we are also ready to update our logic to append `serverConnectionId` when it is available

test/spec/command-monitoring/unified/pre-42-server-connection-id.json
```diff
@@ -0,0 +1,101 @@
+{
```
dariakp, line 1: same comment here regarding NODE-3449 / DRIVERS-1141

test/spec/command-monitoring/unified/find.json
```diff
@@ -0,0 +1,484 @@
...
+          "name": "find",
+          "object": "collection",
+          "arguments": {
+            "filter": {
+              "_id": {
+                "$gt": 1
+              }
+            },
+            "sort": {
+              "_id": 1
```
dariakp, line 126: Can we sync this file  (and the latest yaml) with the latest commit here: https://github.com/mongodb/specifications/commit/bf372a162c304b3f57dccbda00346241ad0cbf17
it's fresh off the press for DRIVERS-1096, and it's "optional" if the tests are already passing, but I think we may as well keep our tests in sync with latest

dariakp (changes requested):

test/spec/command-monitoring/unified/find.json
durran: Updated.

test/spec/command-monitoring/unified/server-connection-id.json
durran: I've removed here.

test/spec/command-monitoring/unified/pre-42-server-connection-id.json
durran: Removed.

test/integration/command-monitoring/command_monitoring.spec.test.ts
```diff
@@ -0,0 +1,16 @@
...
+import { loadSpecTests } from '../../spec';
+import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
+
+// This is skipped because our command monitoring happens at the connection
+// level and is using the server reply for the single insert that the bulk
+// performed since it was only one document in the test. The test expectation
+// is that we are using the bulk write result which was returned to the user
+// as the reply in the command succeeded event instead of our raw reply from
+// the server. There's nothing we can change here.
+const SKIP = ['A successful unordered bulk write with an unacknowledged write concern'];
```
dariakp, line 12: do we have coverage for this scenario in some other way elsewhere? or can we substitute our own replacement version of the test by just copy/pasting the json and modifying the expectation?

dariakp (changes requested):

test/integration/command-monitoring/command_monitoring.spec.test.ts
durran: Well this scenario isn't actually valid for us. The command monitoring spec does not specify where in the driver's object hierarchy it needs to be implemented, nor does it specify if the reply in the command succeeded event is the raw response from the server or the result that the user would see. Drivers were free to implement this as they choose and this test. while valid for Swift and others, isn't valid for Node or Ruby for example. Since this spec test is for command monitoring and not crud, it didn't seem to be covered by another test here - the crud unified tests have assertions for the unacknowledged scenarios.

dariakp: ok that seems fine to me

nbbeeken (approved):

PR #3282, opened 2022-06-02
test: resumability for change stream unified tests 

baileympearson:
#### What is changing?

This PR fixes resumabillity for change stream unified tests.  It also deletes the legacy change stream tests, now that all the unified tests are running.

No.


baileympearson:
evergreen run

test/tools/unified-spec-runner/operations.ts
```diff
@@ -322,7 +322,7 @@ operations.set('iterateUntilDocumentOrError', async ({ entities, operation }) =>
     return await cursor.next();
   }
```
nbbeeken, line 323:
```ts
  const changeStreamOrCursor =
    entities.getEntity('stream', operation.object, false) ??
    entities.getEntity('cursor', operation.object, false);

  if (changeStreamOrCursor == null) {
    // iterateUntilDocumentOrError is used for changes streams and regular cursors.
    // we have no other way to distinguish which scenario we are testing when we run an
    // iterateUntilDocumentOrError operation, so we first try to get the changeStream and
    // if that fails, we know we need to get a cursor
    throw new Error('ah!');
  }

  return await changeStreamOrCursor.next();
```

nbbeeken (dismissed):
LGTM we talked about an improvement that would be nice here, optional

nbbeeken (dismissed):

nbbeeken (approved):

PR #3283, opened 2022-06-02
chore: add generated docs files to release commit

baileympearson:
#### What is changing?

the `-a` flag is added to `standard-version` in our release script, so that generated documentation will be included in the release commit.

note that this doesn't automatically generate the documentation.  but without this change, any generated files will be ignored by `standard-version` when the release commit is created.


package.json
```diff
@@ -118,7 +118,7 @@
     "fix:eslint": "npm run check:eslint -- --fix",
     "prepare": "node etc/prepare.js",
     "preview:docs": "ts-node etc/docs/preview.ts",
-    "release": "standard-version -i HISTORY.md",
+    "release": "standard-version a -i HISTORY.md",
```
nbbeeken, line 121: Needs `-a`
and we chatted a bit offline: we might want a script here that asserts the repo has no other outstanding staged files as this flag could accidentally commit source changes. I would say add a `git reset` here but that could be destructive so something that would just exit failure if any staged changes are outside the docs folder would cover us.

nbbeeken (changes requested):

dariakp (changes requested):

package.json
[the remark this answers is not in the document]
nbbeeken: :jinx:

nbbeeken (dismissed):

out.txt
```diff
@@ -0,0 +1 @@
+hello
```
dariakp, line 1: if this is a test file, it seems that the script doesn't quite work as expected

dariakp (changes requested):

out.txt
baileympearson: Yeah, the changes I made this afternoon weren't quite ready for another review it seems 😅

baileympearson:
Here's a (long) video showing the new functionality 

https://drive.google.com/file/d/1Vq88D--WWUpGLm_O8koP2jSTO4HDYyyn/view?usp=sharing

nbbeeken (approved):

dariakp (approved):

PR #3289, opened 2022-06-13
fix(NODE-4125): change stream resumability

baileympearson:
This PR address resumability issues in the iterator-based API for change streams.

As much as I tried to separate the refactor from the code changes, I ended up making a few refactors as a part of this PR.

1. 44635ba30e9dc93559ba1bfe98fcc65ff062e79a makes the cursor property on the change streams always defined (non-optional).  This was primarily to make tests that use the cursor easier to work with.
2. 881dcf2c460a82018b3476262b989c8f24809fb7 breaks apart the error handling for iterator and event emitter modes.  Although this does result in some duplicated code, it did simply each method and allowed the promisifying of iterator-mode resume without modifying event emitter mode resuming.
3. I made each iterator method (tryNext, next and hasNext) internally use async IIFEs.  I did this for two reasons.  First, it made the resume logic a bit simpler to read and write.  Second, once we get to the async/await work, we can simply remove the `maybePromise` wrapper and the IIFE, and we should have the async functionality we need already written for us.
### Relevant Commits

8820ea16e31874a22bfe1ddafc9423d379327a29 is the most important commit in this PR, as it actually fixes the resumability for iterator mode.

94a48533255022d2e02a44552c551d80887f5cad makes a drive-by fix to remove an internal function in favor of server selection in the resume process.

### Note

As it turns out, the stream based change stream API has never been resumable.  I can file a follow up ticket to make our stream resumable, if we want to, but it seemed out of scope for this bug fix work.

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -641,7 +641,7 @@ export class ChangeStream<
             const hasNext = await this.cursor.hasNext();
             return callback(undefined, hasNext);
           } catch (error) {
-            this._closeWithError(error, callback);
+            this.close(() => callback(error));
           }
         }
       })();
@@ -665,7 +665,7 @@ export class ChangeStream<
             const change = await this.cursor.next();
             this._processNewChange(change ?? null, callback);
           } catch (error) {
-            this._closeWithError(error, callback);
+            this.close(() => callback(error));
           }
         }
       })();
@@ -691,7 +691,7 @@ export class ChangeStream<
             const change = await this.cursor.tryNext();
             callback(undefined, change ?? null);
           } catch (error) {
-            this._closeWithError(error, callback);
+            this.close(() => callback(error));
           }
         }
       })();
@@ -802,13 +802,12 @@ export class ChangeStream<
     return changeStreamCursor;
   }
 
-  /** @internal */
-  private _closeWithError(error: AnyError, callback?: Callback): void {
-    if (!callback) {
+  private _closeEmitterModeWithError(error: AnyError): void {
       this.emit(ChangeStream.ERROR, error);
-    }
 
-    this.close(() => callback && callback(error));
+    this.close(() => {
+      // nothing to do
+    });
   }
 
   /** @internal */
@@ -816,7 +815,7 @@ export class ChangeStream<
     this._setIsEmitter();
     const stream = this[kCursorStream] ?? cursor.stream();
     this[kCursorStream] = stream;
-    stream.on('data', change => this._processNewChange(change));
+    stream.on('data', change => this._processNewChangeEmitterMode(change));
     stream.on('error', error => this._processErrorStreamMode(error));
   }
 
@@ -872,11 +871,11 @@ export class ChangeStream<
 
       const topology = getTopology(this.parent);
       topology.selectServer(this.cursor.readPreference, {}, err => {
-        if (err) return this._closeWithError(err);
+        if (err) return this._closeEmitterModeWithError(err);
         this.cursor = this._createChangeStreamCursor(this.cursor.resumeOptions);
       });
     } else {
-      this._closeWithError(error);
+      this._closeEmitterModeWithError(error);
     }
   }
 
@@ -896,13 +895,13 @@ export class ChangeStream<
       const topology = getTopology(this.parent);
       topology.selectServer(this.cursor.readPreference, {}, err => {
         // if the topology can't reconnect, close the stream
-        if (err) return this._closeWithError(err, callback);
+        if (err) return this.close(() => callback(error));
 
         this.cursor = this._createChangeStreamCursor(this.cursor.resumeOptions);
         callback();
       });
     } else {
-      this._closeWithError(error, callback);
+      this.close(() => callback(error));
     }
   }
 }
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -700,7 +700,7 @@ export class ChangeStream<
 
   /** Is the cursor closed */
   get closed(): boolean {
-    return this[kClosed] || (this.cursor.closed ?? false);
+    return this[kClosed] || this.cursor.closed;
   }
 
   /** Close the Change Stream */
@@ -722,9 +722,13 @@ export class ChangeStream<
    * NOTE: When using a Stream to process change stream events, the stream will
    * NOT automatically resume in the case a resumable error is encountered.
    *
-   * @throws MongoDriverError if this.cursor is undefined
+   * @throws MongoChangeStreamError if the underlying cursor or the change stream is closed
    */
   stream(options?: CursorStreamOptions): Readable & AsyncIterable<TChange> {
+    if (this.closed) {
+      throw new MongoChangeStreamError(CHANGESTREAM_CLOSED_ERROR);
+    }
+
     this.streamOptions = options;
     return this.cursor.stream(options);
   }
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -657,13 +657,13 @@ export class ChangeStream<
       (async () => {
         try {
           const change = await this.cursor.next();
-          this._processNewChange(change ?? null, callback);
+          this.handleChangeCallback(change ?? null, callback);
         } catch (error) {
           const errorOnResume = await this._processErrorAsync(error).catch(err => err);
           if (errorOnResume) return callback(errorOnResume);
           try {
             const change = await this.cursor.next();
-            this._processNewChange(change ?? null, callback);
+            this.handleChangeCallback(change ?? null, callback);
           } catch (error) {
             this.close(() => callback(error));
           }
@@ -696,6 +696,15 @@ export class ChangeStream<
         }
       })();
     });
+  }
+
+  handleChangeCallback(change: TChange | null, callback: Callback): void {
+    try {
+      const processedChange = this._processChange(change);
+      callback(undefined, processedChange);
+    } catch (error) {
+      callback(error);
+    }
   }
 
   /** Is the cursor closed */
@@ -805,7 +814,7 @@ export class ChangeStream<
   }
 
   private _closeEmitterModeWithError(error: AnyError): void {
-      this.emit(ChangeStream.ERROR, error);
+    this.emit(ChangeStream.ERROR, error);
 
     this.close(() => {
       // nothing to do
@@ -832,22 +841,20 @@ export class ChangeStream<
     this[kCursorStream] = undefined;
   }
 
-  /** @internal */
-  private _processNewChange(change: TChange | null, callback?: Callback<TChange>) {
+  private _processChange(change: TChange | null): TChange {
     if (this[kClosed]) {
       // TODO(NODE-3485): Replace with MongoChangeStreamClosedError
-      if (callback) callback(new MongoAPIError(CHANGESTREAM_CLOSED_ERROR));
-      return;
+      throw new MongoAPIError(CHANGESTREAM_CLOSED_ERROR);
     }
 
     // a null change means the cursor has been notified, implicitly closing the change stream
     if (change == null) {
       // TODO(NODE-3485): Replace with MongoChangeStreamClosedError
-      return this._closeWithError(new MongoRuntimeError(CHANGESTREAM_CLOSED_ERROR), callback);
+      throw new MongoRuntimeError(CHANGESTREAM_CLOSED_ERROR);
     }
 
     if (change && !change._id) {
-      return this._closeWithError(new MongoChangeStreamError(NO_RESUME_TOKEN_ERROR), callback);
+      throw new MongoChangeStreamError(NO_RESUME_TOKEN_ERROR);
     }
 
     // cache the resume token
@@ -857,9 +864,16 @@ export class ChangeStream<
     // between resumeToken and startAtOperationTime if we need to reconnect the cursor
     this.options.startAtOperationTime = undefined;
 
-    // Return the change
-    if (!callback) return this.emit(ChangeStream.CHANGE, change);
-    return callback(undefined, change);
+    return change;
+  }
+
+  private _processNewChangeEmitterMode(change: TChange | null): void {
+    try {
+      const processedChange = this._processChange(change);
+      this.emit(ChangeStream.CHANGE, processedChange);
+    } catch (error) {
+      this.emit(ChangeStream.ERROR, error);
+    }
   }
 
   /** @internal */
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -1,4 +1,3 @@
-import Denque = require('denque');
 import type { Readable } from 'stream';
 import { promisify } from 'util';
 
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -48,7 +48,6 @@ const CHANGE_STREAM_EVENTS = [RESUME_TOKEN_CHANGED, END, CLOSE];
 
 const NO_RESUME_TOKEN_ERROR =
   'A change stream document has been received that lacks a resume token (_id).';
-const NO_CURSOR_ERROR = 'ChangeStream has no cursor';
 const CHANGESTREAM_CLOSED_ERROR = 'ChangeStream is closed';
 
 /**
@@ -632,7 +631,6 @@ export class ChangeStream<
     this._setIsIterator();
     return maybePromise(callback, callback => {
       (async () => {
-        if (!this.cursor) return callback(new MongoChangeStreamError(NO_CURSOR_ERROR));
         try {
           const hasNext = await this.cursor.hasNext();
           return callback(undefined, hasNext);
@@ -657,7 +655,6 @@ export class ChangeStream<
     this._setIsIterator();
     return maybePromise(callback, callback => {
       (async () => {
-        if (!this.cursor) return callback(new MongoChangeStreamError(NO_CURSOR_ERROR));
         try {
           const change = await this.cursor.next();
           this._processNewChange(change ?? null, callback);
@@ -684,7 +681,6 @@ export class ChangeStream<
     this._setIsIterator();
     return maybePromise(callback, callback => {
       (async () => {
-        if (!this.cursor) return callback(new MongoChangeStreamError(NO_CURSOR_ERROR));
         try {
           const change = await this.cursor.tryNext();
           callback(undefined, change ?? null);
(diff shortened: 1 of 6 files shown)


src/change_stream.ts
```diff
@@ -655,11 +630,23 @@ export class ChangeStream<
...
   hasNext(callback?: Callback): Promise<boolean> | void {
     this._setIsIterator();
-    return maybePromise(callback, cb => {
-      this._getCursor((err, cursor) => {
-        if (err || !cursor) return cb(err); // failed to resume, raise an error
-        cursor.hasNext(cb);
-      });
+    return maybePromise(callback, callback => {
+      (async () => {
+        if (!this.cursor) return callback(new MongoChangeStreamError(NO_CURSOR_ERROR));
```
durran, line 635: I believe this line can be removed along with the `NO_CURSOR_ERROR` constant. I've removed this locally on this branch and all the change stream integration tests still pass.

src/change_stream.ts
```diff
@@ -668,18 +655,23 @@ export class ChangeStream<
...
-      this._getCursor((err, cursor) => {
-        if (err || !cursor) return cb(err); // failed to resume, raise an error
-        cursor.next((error, change) => {
-          if (error) {
-            this[kResumeQueue].push(() => this.next(cb));
-            this._processError(error, cb);
-            return;
+    return maybePromise(callback, callback => {
+      (async () => {
+        if (!this.cursor) return callback(new MongoChangeStreamError(NO_CURSOR_ERROR));
```
durran, line 660: I believe this line can be removed.

src/change_stream.ts
```diff
@@ -690,11 +682,23 @@ export class ChangeStream<
...
   tryNext(callback?: Callback<Document | null>): Promise<Document | null> | void {
     this._setIsIterator();
-    return maybePromise(callback, cb => {
-      this._getCursor((err, cursor) => {
-        if (err || !cursor) return cb(err); // failed to resume, raise an error
-        return cursor.tryNext(cb);
-      });
+    return maybePromise(callback, callback => {
+      (async () => {
+        if (!this.cursor) return callback(new MongoChangeStreamError(NO_CURSOR_ERROR));
```
durran, line 687: I believe this line can be removed.

durran (changes requested):
I removed all the if `this.cursor` checks on this PR and made `this.cursor?` lines just `this.cursor` and all change stream integration tests passed. Just want to make sure that's expected and that extra code could be removed (since we always have a cursor it seems). Would you like me to just push my changes to the branch or am I missing something where the cursor would not be there?

durran:
My patch run: https://spruce.mongodb.com/version/62ab44d632f417132f2a9a8a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

durran (approved):
LGTM - Thanks!

src/change_stream.ts
```diff
@@ -655,11 +629,22 @@ export class ChangeStream<
   hasNext(callback: Callback<boolean>): void;
   hasNext(callback?: Callback): Promise<boolean> | void {
     this._setIsIterator();
-    return maybePromise(callback, cb => {
-      this._getCursor((err, cursor) => {
-        if (err || !cursor) return cb(err); // failed to resume, raise an error
-        cursor.hasNext(cb);
-      });
+    return maybePromise(callback, callback => {
```
dariakp, line 632: Are we concerned at all about shadowing the `callback` variable here? and in the subsequent methods? This syntax makes it harder to understand what's actually going on

src/change_stream.ts
```diff
@@ -655,11 +629,22 @@ export class ChangeStream<
...
+      (async () => {
+        try {
+          const hasNext = await this.cursor.hasNext();
+          return callback(undefined, hasNext);
+        } catch (error) {
+          const errorOnResume = await this._processErrorAsync(error).catch(err => err);
```
dariakp, line 638: what is "errorOnResume" supposed to mean here? The name of `_processErrorAsync` doesn't really give a hint as to what it really does, but seems like what it would return is any error that is not resumable, so maybe `unresumableError` is a more apt name here (and in the other methods)?

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1563,3 +1493,584 @@ describe('Change Streams', function () {
...
+    { error: 'CursorNotFound', code: 43 }
+  ];
+
+  const is4_2Server = (serverVersion: string) =>
+    gte(serverVersion, '4.2.0') && lt(serverVersion, '4.3.0');
+
+  beforeEach(async function () {
+    client = this.configuration.newClient({ monitorCommands: true });
+    client.on('commandStarted', filterForCommands(['aggregate'], aggregateEvents));
+    await client.connect();
```
nbbeeken, line 1535: No need for connect anymore :)

src/change_stream.ts
```diff
@@ -655,11 +629,22 @@ export class ChangeStream<
...
+          if (errorOnResume) return callback(errorOnResume);
+          try {
+            const hasNext = await this.cursor.hasNext();
```
dariakp, line 641: am I reading it correctly here that in the case of a resumable error, we attempt to resume exactly once? if so, is that explicit spec behavior? (same question on the other methods here)

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1563,3 +1493,584 @@ describe('Change Streams', function () {
...
+                ? 'failCommand'
+                : 'failGetMoreAfterCursorCheckout',
+              mode: { times: 1 },
+              data: {
+                failCommands: ['getMore'],
+                errorCode: code
+              }
+            });
+
+            await collection.insertOne({ name: 'bailey' });
```
dariakp, line 1571: is this a sneaky way to immortalize yourself in the codebase? 😆

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1563,3 +1493,584 @@ describe('Change Streams', function () {
...
+  });
+
+  it(
+    'caches the server version after the initial aggregate call',
+    { requires: { topology: '!single' } },
+    async function () {
+      changeStream = collection.watch([], changeStreamResumeOptions);
+      await initIteratorMode(changeStream);
+
+      expect(changeStream.cursor.maxWireVersion).not.to.be.undefined;
```
dariakp, line 2055: should this first check to make sure that the maxWireVersion is initially undefined? and that the subsequent version is a number?

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1563,3 +1493,584 @@ describe('Change Streams', function () {
...
+
+  it(
+    'caches the server version after each getMore call',
+    { requires: { topology: '!single' } },
+    async function () {
+      changeStream = collection.watch([], changeStreamResumeOptions);
+      await initIteratorMode(changeStream);
+
+      const maxWireVersion = changeStream.cursor.maxWireVersion;
+      changeStream.cursor.maxWireVersion = 20;
```
dariakp, line 2067: are you trying to test that the version is overwritten after the update? if so, would it be better to set it to something that's super low instead? since in the future, the wire versions will go up? also, the test case name should maybe say that it `updates the cached server version`

I do wonder though why the test setup is different in this case vs the aggregate case; feels like maybe both of them should make more than one call in order to satisfy the "each" part of the test case description

dariakp (changes requested):

src/change_stream.ts
```diff
@@ -655,11 +629,22 @@ export class ChangeStream<
...
   hasNext(callback?: Callback): Promise<boolean> | void {
     this._setIsIterator();
-    return maybePromise(callback, cb => {
-      this._getCursor((err, cursor) => {
-        if (err || !cursor) return cb(err); // failed to resume, raise an error
-        cursor.hasNext(cb);
-      });
+    return maybePromise(callback, callback => {
+      (async () => {
+        try {
```
nbbeeken, line 634: might need `.then`? reaching out on slack

src/change_stream.ts
baileympearson: Yes, this is spec behavior - https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#why-are-changestreams-required-to-retry-once-on-a-resumable-error

The idea is that some errors might be transient, but might also indicate an actual problem.  So we attempt to resume once in case it's not a problem, but if it fails a second time we propagate the error to the user.

src/change_stream.ts
nbbeeken: I'm in favor of this, it's a good idea that we should always shadow "callback" inside maybePromise. The outer `callback` is nullish where as the inner one is always defined (either as a promise resolver or the user passed in callback). By shadowing the name we prevent code inside this scope from accidentally reaching up and calling a nullish variable. TS will also catch that but I think its confusing in the reverse, when reading "maybePromise" code I shouldn't have to see two callback variables in the scope, only one of them is ever the correct one to use so might as well use the same name again.

baileympearson: This was intended - @nbbeeken feel free to chime in if I misinterpreted our conversation the other day.

The second argument to `maybePromise` is a function that `maybePromise` invokes.  `maybePromise` passes the callback argument in for us (which could be a Promise.resolve or something similar).  By shadowing the outer callback, we ensure that we will never acccidentally call the outer callback in the inner function.  For example:

```typescript
function foo(callback: Callback<any>) {
    callback('ooops, called the wrong callback');
  });
}
```

test/integration/change-streams/change_stream.test.ts
baileympearson: caught red handed! 😱 

I usually do this for testing and must not have noticed when I pushed.  Should I use a number instead of someones name?

test/integration/change-streams/change_stream.test.ts
baileympearson: There's only ever one aggregate call (it just initializes the change stream), so the aggregate test should only initialize the cursor once.  As far as I know, initializing an initialized cursor is undefined behavior.

Good point about the `each` point - I'll update the test to call getMore again and confirm that the value has been set after each call.

src/change_stream.ts
baileympearson: If there's any sort of error during the resume process, or if it's an unresumable error.  I'll see if I can find a better name that's more descriptive for both the method and the variable.

src/change_stream.ts
dariakp: I don't entirely buy that justification because these sorts of things should be protected against via regression testing, not code obfuscation; i.e., if it's a legitimate concern, then add a test to ensure that the outer callback isn't called. There's no way for someone just looking at this code to know whether the shadowing is intentional or not and whether we do or don't ever want to call that outer function.

test/integration/change-streams/change_stream.test.ts
dariakp: I mean I don't really care, I was just amused, up to you if you want to tie your reputation to this test 😄

src/change_stream.ts
dariakp: As a relevant aside: clever code, even when well-intentioned, tends to be difficult to maintain in the long term, since it almost inevitably forces other to play the "guess the intention" game

dariakp: I would actually prefer to add the https://eslint.org/docs/rules/no-shadow lint rule to avoid what is generally an antipattern due to readability issues. I think in this case, it's an instance of misdirected defensive programming. If it's not a case we're truly worried about, and it's not worth testing, then it's not worth coding for. If it's a case we are worried about, then we *should* test for it. People reading the code shouldn't have to guess why you decided to shadow a variable - whether it was purposeful or an accident, and what the purpose may have been.

dariakp: A summary from slack discussion with Neal: 
I personally favor good variable names and a no-shadow lint rule, but I don't think it needs to be in the scope of the change streams work; I do want a comment if we keep the shadowing because that's the very minimum you need for clarity of intent here, so, Bailey - you are welcome to either add distinct but descriptive callback names here, or add a comment to each instance of the shadowing with an explanation of why the shadowing is intentional

test/integration/change-streams/change_stream.test.ts
baileympearson: I'll leave it.  fwiw I usually rotate between all our names, so perhaps in the future you'll see a commit from me that has `{name: 'daria' }` or `{name: 'neal'}` or `{ name: 'durran' }`

src/change_stream.ts
baileympearson: I created a jira ticket, left a TODO and added a comment explaining the shadowing.

test/integration/change-streams/change_stream.test.ts
baileympearson: Removed!

dariakp (commented):
the changes addressing my comments look good to me, I just have a suggestion on one of the tests

src/change_stream.ts
```diff
@@ -860,127 +861,74 @@ export class ChangeStream<
...
-    // if initial error wasn't resumable, raise an error and close the change stream
-    return this._closeWithError(error, callback);
   }
 
   /** @internal */
-  private _getCursor(callback: Callback<ChangeStreamCursor<TSchema, TChange>>) {
+  private _processErrorIteratorMode = promisify(this._processErrorIteratorModeCallback);
+
+  /** @internal */
+  private _processErrorIteratorModeCallback(error: AnyError, callback: Callback) {
```
nbbeeken, line 913: Can we just make this async?

src/change_stream.ts
```diff
@@ -668,18 +659,30 @@ export class ChangeStream<
...
+        } catch (error) {
+          try {
+            await this._processErrorIteratorMode(error);
+            const change = await this.cursor.next();
+            const processedChange = this._processChange(change ?? null);
+            return processedChange;
+          } catch (error) {
+            await this.close().catch(err => err);
+            throw error;
           }
```
nbbeeken, line 680:
```suggestion
          } finally {
            await this.close();
          }
```
I'm not seeing where .close will throw or reject, so I'm not sure we should attach a silencer, thoughts? also catching then throwing seems like a use case for finally

nbbeeken (changes requested):

src/change_stream.ts
nbbeeken: Ah I see, we don't want to have to do that everytime, but we should leave a comment saying thats why we did this, so we can removing it when serverSelection is awaitable

baileympearson: We could also expose a promisified server selection function on the topology that we could make use of now throughout the driver.  Then we wouldn't have to promisify it each time.

And fwiw, I actually don't think it's bad to have the promisifying here.  It's not a hot code path - it's the error handling function.  So, this would either get called in one-off transient failures before the change stream resumes or twice in a row in the case of an resumable error that can't resume, and then the change stream would be closed.

thoughts?

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1563,3 +1493,603 @@ describe('Change Streams', function () {
...
+            .to.have.property('options')
+            .that.containSubset(changeStreamResumeOptions);
+
+          await collection.insertOne({ name: 'bailey' });
+
+          await changeStream.next();
+
+          expect(changeStream.cursor)
+            .to.have.property('options')
+            .that.containSubset(changeStreamResumeOptions);
```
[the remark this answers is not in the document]
baileympearson: No?  containsSubset checks that the argument is a subset of the original.  So we assert that every change stream option is contained in the cursor options, demonstrating that they were copied over correctly.

src/change_stream.ts
```diff
@@ -860,127 +861,74 @@ export class ChangeStream<
...
+    if (isResumableError(error, this.cursor.maxWireVersion)) {
+      this.cursor.close();
 
-    // no cursor, queue callback until topology reconnects
-    this[kResumeQueue].push(callback);
-  }
+      const topology = getTopology(this.parent);
+      topology.selectServer(this.cursor.readPreference, {}, err => {
+        // if the topology can't reconnect, close the stream
+        if (err) return this.close(() => callback(error));
```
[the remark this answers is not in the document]
baileympearson: done

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1563,3 +1493,603 @@ describe('Change Streams', function () {
...
+
+  it(
+    'caches the server version after the initial aggregate call',
+    { requires: { topology: '!single' } },
+    async function () {
+      changeStream = collection.watch([], changeStreamResumeOptions);
+      expect(changeStream.cursor.maxWireVersion).to.be.undefined;
+      await initIteratorMode(changeStream);
+
+      expect(typeof changeStream.cursor.maxWireVersion).to.equal('number');
```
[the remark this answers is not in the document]
baileympearson: done

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1563,3 +1493,603 @@ describe('Change Streams', function () {
...
+      }
+
+      it(
+        'maintains change stream options on resume',
+        { requires: { topology: '!single', mongodb: '>=4.2' } },
+        async function () {
+          changeStream = collection.watch([], changeStreamResumeOptions);
+          await initIteratorMode(changeStream);
+
+          await client.db('admin').command(<FailPoint>{
```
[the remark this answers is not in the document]
baileympearson: fixed

test/integration/change-streams/change_stream.test.ts
```diff
@@ -2048,28 +2047,48 @@ describe('ChangeStream resumability', function () {
   it(
     'caches the server version after the initial aggregate call',
     { requires: { topology: '!single' } },
+    async function () {
+      changeStream = collection.watch([], changeStreamResumeOptions);
+      expect(changeStream.cursor.maxWireVersion).to.be.undefined;
+      await initIteratorMode(changeStream);
+
+      expect(typeof changeStream.cursor.maxWireVersion).to.equal('number');
```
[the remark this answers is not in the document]
baileympearson: fixed!

nbbeeken (approved):

nbbeeken (approved):

test/integration/change-streams/change_stream.test.ts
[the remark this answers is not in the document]
dariakp: looks like the other instance of it was updated, but not this specific one

dariakp (approved):

nbbeeken (approved):

PR #3297, opened 2022-06-23
test(NODE-4227): add explicit encryption prose tests

durran:
### Description

Adds explicit encryption prose tests

#### What is changing?

Adds 5 new prose tests for #12 in the client side encryption spec. https://github.com/mongodb/specifications/blob/7f37c9325b423223edf54d795afed444e6989b98/source/client-side-encryption/tests/README.rst#explicit-encryption

I've also included the spec prose test language in code comments in the test here for easier following.

No

NODE-4227

### DIFF
--- a/test/integration/client-side-encryption/client_side_encryption.prose.test.js
+++ b/test/integration/client-side-encryption/client_side_encryption.prose.test.js
@@ -1434,226 +1434,6 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
     });
   });
 
-  context('14. Decryption Events', metadata, function () {
-    let setupClient;
-    let clientEncryption;
-    let keyId;
-    let cipherText;
-    let malformedCiphertext;
-    let encryptedClient;
-    let aggregateSucceeded;
-    let aggregateFailed;
-
-    beforeEach(async function () {
-      const mongodbClientEncryption = this.configuration.mongodbClientEncryption;
-      // Create a MongoClient named ``setupClient``.
-      setupClient = this.configuration.newClient();
-      // Drop and create the collection ``db.decryption_events``.
-      const db = setupClient.db('db');
-      await dropCollection(db, 'decryption_events');
-      await db.createCollection('decryption_events');
-      // Create a ClientEncryption object named ``clientEncryption`` with these options:
-      //   ClientEncryptionOpts {
-      //     keyVaultClient: <setupClient>,
-      //     keyVaultNamespace: "keyvault.datakeys",
-      //     kmsProviders: { "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }
-      //   }
-      clientEncryption = new mongodbClientEncryption.ClientEncryption(setupClient, {
-        keyVaultNamespace: 'keyvault.datakeys',
-        kmsProviders: getKmsProviders(LOCAL_KEY),
-        bson: BSON
-      });
-      // Create a data key with the "local" KMS provider.
-      // Storing the result in a variable named ``keyID``.
-      keyId = await clientEncryption.createDataKey('local');
-      // Use ``clientEncryption`` to encrypt the string "hello" with the following ``EncryptOpts``:
-      //   EncryptOpts {
-      //     keyId: <keyID>,
-      //     algorithm: "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
-      //   }
-      // Store the result in a variable named ``ciphertext``.
-      cipherText = await clientEncryption.encrypt('hello', {
-        keyId: keyId,
-        algorithm: 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic'
-      });
-      // Copy ``ciphertext`` into a variable named ``malformedCiphertext``.
-      // Change the last byte to 0. This will produce an invalid HMAC tag.
-      const buffer = Buffer.from(cipherText.buffer);
-      buffer.writeInt8(0, buffer.length - 1);
-      malformedCiphertext = new Binary(buffer, 6);
-      // Create a MongoClient named ``encryptedClient`` with these ``AutoEncryptionOpts``:
-      //   AutoEncryptionOpts {
-      //     keyVaultNamespace: "keyvault.datakeys";
-      //     kmsProviders: { "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }
-      //   }
-      // Configure ``encryptedClient`` with "retryReads=false".
-      encryptedClient = this.configuration.newClient(
-        {},
-        {
-          retryReads: false,
-          monitorCommands: true,
-          autoEncryption: {
-            keyVaultNamespace: 'keyvault.datakeys',
-            kmsProviders: getKmsProviders(LOCAL_KEY)
-          }
-        }
-      );
-      // Register a listener for CommandSucceeded events on ``encryptedClient``.
-      encryptedClient.on('commandSucceeded', event => {
-        if (event.commandName === 'aggregate') {
-          aggregateSucceeded = event;
-        }
-      });
-      // The listener must store the most recent CommandFailedEvent error for the "aggregate" command.
-      encryptedClient.on('commandFailed', event => {
-        if (event.commandName === 'aggregate') {
-          aggregateFailed = event;
-        }
-      });
-    });
-
-    afterEach(async function () {
-      aggregateSucceeded = undefined;
-      aggregateFailed = undefined;
-      await setupClient.close();
-      await encryptedClient.close();
-    });
-
-    context('Case 1: Command Error', metadata, function () {
-      beforeEach(async function () {
-        // Use ``setupClient`` to configure the following failpoint:
-        //    {
-        //         "configureFailPoint": "failCommand",
-        //         "mode": {
-        //             "times": 1
-        //         },
-        //         "data": {
-        //             "errorCode": 123,
-        //             "failCommands": [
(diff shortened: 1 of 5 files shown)


test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1426,6 +1434,291 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+    let key1Id;
+    let setupClient;
+    let keyVaultClient;
+    let clientEncryption;
+    let encryptedClient;
+
+    beforeEach(async function () {
+      const mongodbClientEncryption = this.configuration.mongodbClientEncryption;
+      // Load the file encryptedFields.json as encryptedFields.
+      encryptedFields = EJSON.parse(
```
durran, line 1450: Data for the `encryptedFields` option and key document are now provided in 2 new files in the etc/ directory in the csfle spec directory.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1426,6 +1434,291 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+          }
+        });
+
+        it('returns all documents with contention', async function () {
+          // Use encryptedClient to run a "find" operation on the db.explicit_encryption
+          // collection with the filter { "encryptedIndexed": <findPayload2> }.
+          // Assert 10 documents are returned. Assert each returned document contains the
+          // field { "encryptedIndexed": "encrypted indexed value" }.
+          const collection = encryptedClient.db('db').collection('explicit_encryption');
+          const result = await collection.find({ encryptedIndexed: findPayload2 }).toArray();
```
durran, line 1631: Just pointing out I find this absolutely crazy. A value is directly encrypted on line 1603 with a contention factor equal to the amount of documents in the collection that would match the original value of the field. But each field's encrypted value would not match the encrypted value in theory as all the bin data would be different for each document on insert. But it still matches all of them. Mind blown.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1426,6 +1434,291 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+        // collection with the filter { "_id": 1 }.
+        // Assert one document is returned containing the field
+        // { "encryptedUnindexed": "encrypted unindexed value" }.
+        const collection = encryptedClient.db('db').collection('explicit_encryption');
+        const result = await collection.findOne({ _id: 1 });
+        expect(result.encryptedIndexed).to.equal('encrypted unindexed value');
+      });
+    });
+
+    context('Case 4: can roundtrip encrypted indexed', eeMetadata, function () {
```
durran, line 1673: These last 2 are just encrypt/decrypt roundtrips. So pretty straightforward.

durran:
The failing tests need another mongodb-client-encryption release.

test/integration/shared.js
```diff
@@ -38,8 +38,8 @@ function delay(timeout) {
   });
 }
 
-function dropCollection(dbObj, collectionName) {
-  return dbObj.dropCollection(collectionName).catch(ignoreNsNotFound);
+function dropCollection(dbObj, collectionName, options = {}) {
+  return dbObj.dropCollection(collectionName, options).catch(ignoreNsNotFound);
```
durran, line 42: Just adding the ability to pass options to the `dropCollection` helper. If we pass `encryptedFields` here an extra 3 collections will get dropped. (The esc/ecc/ecoc collections)

durran:
> Do you want to drop these tests from the libmongocrypt repo? https://github.com/mongodb/libmongocrypt/blob/93b82421dfe9ad472ef12b26a47aa94e631f9be1/bindings/node/test/clientEncryption.test.js#L584-L797
> 
> Fwiw, I really like that the Node.js driver deviates from other drivers and instead maintains a lot more functionality in the libmongocrypt repo (i.e. the full `AutoEncrypter` and `ClientEncryption` classes, rather than just raw bindings).

I think it's fine to keep them there - can't have too many tests. :)

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -40,6 +40,14 @@ const metadata = {
   }
 };
 
+const eeMetadata = {
+  requires: {
+    clientSideEncryption: true,
+    mongodb: '>=6.0.0',
+    topology: ['replicaset', 'sharded']
```
dariakp, line 47: what about load balanced? is that considered more like a standalone?

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1426,6 +1434,291 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+      });
+
+      it('returns the decrypted value', async function () {
+        // Use encryptedClient to run a "find" operation on the db.explicit_encryption
+        // collection with the filter { "encryptedIndexed": <findPayload> }.
+        // Assert one document is returned containing the field
+        // { "encryptedIndexed": "encrypted indexed value" }.
+        const collection = encryptedClient.db('db').collection('explicit_encryption');
+        const result = await collection.findOne({ encryptedIndexed: findPayload });
+        expect(result.encryptedIndexed).to.equal('encrypted indexed value');
```
dariakp, line 1550: this is an entirely optional suggestion, but since we are testing for equality, you could use the `expect(result).to.have.property('encryptedIndexed', 'encrypted indexed value')` for an assertion with a slightly better message in case of failure throughout this file

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1426,6 +1434,291 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+        // Store the result in insertPayload.
+        insertPayload = await clientEncryption.encrypt('encrypted unindexed value', {
+          keyId: key1Id,
+          algorithm: 'Unindexed'
+        });
+        // Use encryptedClient to insert the document { "_id": 1, "encryptedUnindexed": <insertPayload> }
+        // into db.explicit_encryption.
+        await encryptedClient.db('db').collection('explicit_encryption').insertOne({
+          _id: 1,
+          encryptedIndexed: insertPayload
```
dariakp, line 1658: this should be `encryptedUnindexed`

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1426,6 +1434,291 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+      });
+
+      it('returns unindexed documents', async function () {
+        // Use encryptedClient to run a "find" operation on the db.explicit_encryption
+        // collection with the filter { "_id": 1 }.
+        // Assert one document is returned containing the field
+        // { "encryptedUnindexed": "encrypted unindexed value" }.
+        const collection = encryptedClient.db('db').collection('explicit_encryption');
+        const result = await collection.findOne({ _id: 1 });
+        expect(result.encryptedIndexed).to.equal('encrypted unindexed value');
```
dariakp, line 1669: needs to assert on `encryptedUnindexed` (also see my note above about using `property` assertions)

dariakp (changes requested):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
durran: We don't run the other fle tests against load balanced clusters so I was keeping in line with that. The spec in general pre-dates the load balancer spec and does not mention that topology at all, and enabling it got some strange setup errors that I felt were outside the scope of the ticket. I can create a ticket to look into running all the fle tests with load balancers as well.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
durran: Updated.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
durran: Updated

test/integration/client-side-encryption/client_side_encryption.prose.test.js
durran: Updated.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
dariakp: I think we should check with the spec owners regarding this, and maybe file a drivers ticket to add lb to the testing matrix if it makes sense to do so, then the node ticket will automatically come down to us

dariakp (approved):

nbbeeken (approved):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
addaleax: Just for context, the reason for requiring these topologies is that QE requires transactions, which in turn requires a replset topology. I guess technically one could run a load balancer in front of a single standalone node, but I assume that won’t happen in practice or in tests?

durran: Just as a follow up, other drivers are not running prose tests against a load balancer. However, the LB spec says it should run all tests in unified format so we'll need to make sure the new FLE unified tests are running against a load balanced topology.

PR #3298, opened 2022-06-23
feat(NODE-4301): remove shared lib experimental tags

durran:
### Description

Removes the experimental tags for the shared csfle library.

#### What is changing?

The `extraOptions.cryptSharedLibPath` and `extraOptions.cryptSharedLibRequired` provided as auto encryption options are GA and no longer experimental.

No changes to mongodb-client-encryption.

Update release notes on next release to note this.

NODE-4301

### DIFF
--- a/src/deps.ts
+++ b/src/deps.ts
@@ -344,7 +344,7 @@ export interface AutoEncryptionOptions {
      *
      * Specifying a path prevents mongocryptd from being used as a fallback.
      *
-     * @experimental Requires the MongoDB Crypt shared library, available in MongoDB 6.0 or higher.
+     * Requires the MongoDB Crypt shared library, available in MongoDB 6.0 or higher.
      */
     cryptSharedLibPath?: string;
     /**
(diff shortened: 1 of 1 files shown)


addaleax (approved):
LGTM, but I’m curious, since we had a long discussion about this: My understanding was that I was asked to add these in https://github.com/mongodb/node-mongodb-native/pull/3280 specifically *because 6.0 was not GA yet*. It still isn’t –  what changed?

dariakp:
> LGTM, but I’m curious, since we had a long discussion about this: My understanding was that I was asked to add these in #3280 specifically _because 6.0 was not GA yet_. It still isn’t – what changed?

I am also curious about this. We should probably have a team discussion, but I think the idea was that we wouldn't remove the experimental tags until we were 100% sure the api would not possibly change for the 6.0 server release; did we obtain any communication that confirmed that this API is certified stable ahead of the GA?

nbbeeken (approved):

PR #3304, opened 2022-06-27
chore(NODE-4321): add typescript to eslint rules

nbbeeken:
#### What is changing?

Adds `@typescript-eslint/recommended-requiring-type-checking` rules which include enforcing promise returning functions are awaited.

Skips the following:
- @typescript-eslint/no-unsafe-member-access
- @typescript-eslint/no-unsafe-argument
- @typescript-eslint/no-unsafe-assignment
- @typescript-eslint/no-unsafe-return
- @typescript-eslint/no-unsafe-call
- @typescript-eslint/restrict-plus-operands
- @typescript-eslint/restrict-template-expressions

I'll file follow up tickets for us to address these.

There are outstanding TODOs that need discussion.

We can use async await internally as long as we pass the result through our callback / maybePromise handling logic. We want this lint rule to help us catch any mistakes where we might forget to await a promise returning funciton.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -182,6 +182,29 @@
       }
     },
     {
+      // Settings for typescript test files
+      "files": [
+        "src/**/*.ts"
+      ],
+      "parser": "@typescript-eslint/parser",
+      "parserOptions": {
+        "project": ["./tsconfig.json"]
+      },
+      "extends": [
+        "plugin:@typescript-eslint/recommended-requiring-type-checking"
+      ],
+      "rules": {
+        "@typescript-eslint/no-unsafe-member-access": "off",
+        "@typescript-eslint/no-unsafe-argument": "off",
+        "@typescript-eslint/no-unsafe-assignment": "off",
+        "@typescript-eslint/no-unsafe-return": "off",
+        "@typescript-eslint/no-unsafe-call": "off",
+
+        "@typescript-eslint/restrict-plus-operands": "off",
+        "@typescript-eslint/restrict-template-expressions": "off"
+      }
+    },
+    {
       // Settings for typescript type test files
       "files": [
         "*.test-d.ts"
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -894,10 +894,7 @@ export class ChangeStream<
 
     if (isResumableError(changeStreamError, this.cursor.maxWireVersion)) {
       this._endStream();
-      this.cursor.close().then(
-        () => null,
-        () => null
-      ); // Ignoring the result of close is intentional
+      this.cursor.close().catch(() => null);
 
       const topology = getTopology(this.parent);
       topology.selectServer(this.cursor.readPreference, {}, serverSelectionError => {
@@ -927,10 +924,7 @@ export class ChangeStream<
     }
 
     if (isResumableError(changeStreamError, this.cursor.maxWireVersion)) {
-      this.cursor.close().then(
-        () => null,
-        () => null
-      ); // Ignoring the result of close is intentional
+      this.cursor.close().catch(() => null);
 
       const topology = getTopology(this.parent);
       topology.selectServer(this.cursor.readPreference, {}, serverSelectionError => {
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -897,7 +897,7 @@ export class ChangeStream<
       this.cursor.close().then(
         () => null,
         () => null
-      ); // TODO(NODE-XXX): Is this correct?
+      ); // Ignoring the result of close is intentional
 
       const topology = getTopology(this.parent);
       topology.selectServer(this.cursor.readPreference, {}, serverSelectionError => {
@@ -930,7 +930,7 @@ export class ChangeStream<
       this.cursor.close().then(
         () => null,
         () => null
-      ); // TODO(NODE-XXX): Is this correct?;
+      ); // Ignoring the result of close is intentional
 
       const topology = getTopology(this.parent);
       topology.selectServer(this.cursor.readPreference, {}, serverSelectionError => {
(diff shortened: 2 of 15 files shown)


src/deps.ts
```diff
@@ -141,6 +131,7 @@ interface AWS4 {
    * @param credentials - AWS credential details, sessionToken should be omitted entirely if its false-y
    */
   sign(
+    this: void,
```
baileympearson, line 134: Removing this throws an error in mongodb_aws.ts, but I actually think that binding the function in mongodb_aws is the correct solution, not adding this to the type.

```typescript
   const sign = aws4.sign.bind(aws4);
```

thoughts?

.eslintrc.json
```diff
@@ -181,6 +181,29 @@
         "typescript-eslint/ban-ts-comment": "off"
       }
     },
+    {
+      // Settings for typescript test files
+      "files": [
+        "src/**/*.ts"
```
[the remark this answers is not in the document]
baileympearson: Thanks for filing the follow up ticket, I think it's important to ensure that our tests are correct too

baileympearson (approved):

src/change_stream.ts
```diff
@@ -923,7 +927,10 @@ export class ChangeStream<
     }
 
     if (isResumableError(changeStreamError, this.cursor.maxWireVersion)) {
-      this.cursor.close();
+      this.cursor.close().then(
```
dariakp, line 930: can we `.catch` or go with the rule disable approach here? I'm all for being explicit, but the `.then` tends to suggest that there might be a result returned from this function that one *could* care about, and that's not true for functions that don't return anything

dariakp (changes requested):

src/cmap/wire_protocol/compression.ts
```diff
@@ -102,9 +103,10 @@ export function decompress(
       if (Snappy[PKG_VERSION].major <= 6) {
         Snappy.uncompress(compressedData, { asBuffer: true }, callback);
       } else {
-        Snappy.uncompress(compressedData, { asBuffer: true })
-          .then(buffer => callback(undefined, buffer))
-          .catch(error => callback(error));
+        Snappy.uncompress(compressedData, { asBuffer: true }).then(
+          buffer => callback(undefined, buffer),
+          error => callback(error)
```
dariakp, line 108: just double checking that this is what we want; the previous handler would have caught an error thrown inside the `.then` (i.e., the .then call to the callback); the new behavior would only handle the error returned from the original `uncompress`

src/change_stream.ts
```diff
@@ -923,7 +924,7 @@ export class ChangeStream<
     }
 
     if (isResumableError(changeStreamError, this.cursor.maxWireVersion)) {
-      this.cursor.close();
+      this.cursor.close().catch(() => null);
```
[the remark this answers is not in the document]
nbbeeken: Appreciate the suggestion! IIUC wouldn't leaving the promise without a rejection handler risk exiting the process if it did error? it's spec-ed behavior to ignore errors from close (@baileympearson do you have the reference?)

src/change_stream.ts
nbbeeken: went with .catch(() => null) (still returning a value but null is the absence of something 🤔)

src/cmap/wire_protocol/compression.ts
nbbeeken: Yea, this is a common mistake when callbackifying a promise API. if the callback throws (which it shouldn't but nonetheless it could) then you call the callback with its own error, this avoids that.

dariakp (approved):

src/change_stream.ts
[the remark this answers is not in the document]
baileympearson: I don't know if it spec-ed.  We had a discussion about this on the PR already but I'll post this again with more detail. 

Relevant ticket: [jira.mongodb.org/browse/NODE-4217](https://jira.mongodb.org/browse/NODE-4217)
Related PR: https://github.com/mongodb/node-mongodb-native/pull/3226

I encountered this issue for cursor.close and filed the aforementioned ticket and left the TODOs that you see in the above PR.  When we triaged/refined it, we decided that the current behavior was correct and closed the ticket as won't do (at least, I _thought_ the jira status should be won't do but I guess we kinda ruined that flow by adding AC to the ticket related to removing TODOs.  Maybe that's a process improvement we could make for the future).  You can see Marin's AC on the ticket is to remove the TODOs I added, which I did in the PR.  We don't have any real record of this but I do remember Jeff mentioning that this would be the intended behavior.  My understanding was that the justification was two-fold:

1. What exactly would the user do in the case that cursor.close failed? Try again?  If killCursors fails, there isn't much they can do because it probably means something larger is going wrong.
2. Cursors timeout on the server anyways, so if the cursor fails to close eventually it will be GC'd.

I think we could revisit this though because I think an argument could be made for cursor.close() propagating errors.

-------

Now this particular instance is a separate issue, irrespective whether or not cursor.close() swallows errors.  This call to cursor.close() is inside the error handler for ChangeStreams.  This code is run if there is an error with the change stream.  The desired (but not explicitly spec'd behavior that I know of) would be

1. Error occurs in the CS (resumable or not)
2. We attempt to resume.  This involves cleaning up the previous cursor and restarting the stream
3. Resuming, or failing that, propagating the error to the user

What we don't want is the error from cursor.close() either crashing the user's process (unhandled process rejection) or overridding the original change stream error.  The fact that our CS uses cursors internally and that they're closed during the resume process is an implementation detail and not something that a user cares about.  So in this scenario, irrespective of whether or not cursor.close() swallows errors, we should swallow the error.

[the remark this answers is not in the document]
addaleax: > IIUC wouldn't leaving the promise without a rejection handler risk exiting the process if it did error?

Right, it does make sense to leave a `.catch(() => {})` in places where you intentionally ignore errors :+1:

PR #3311, opened 2022-07-01
fix(NODE-4173): WIP send maxAwaitTimeMS to the server

nbbeeken:
#### What is changing?

The AbstractCursor wil now save maxAwaitTimeMS or maxTimeMS to it's options as `maxAwaitTimeMS` and the getMore command construction logic sends maxTimeMS to the server (as expected).

maxAwaitTimeMS was being ignored and not sent to the server. This corrects that.


src/cursor/abstract_cursor.ts
```diff
@@ -182,8 +183,8 @@ export abstract class AbstractCursor<
       this[kOptions].comment = options.comment;
     }
 
-    if (typeof options.maxTimeMS === 'number') {
-      this[kOptions].maxTimeMS = options.maxTimeMS;
+    if (typeof options.maxTimeMS === 'number' || typeof options.maxAwaitTimeMS === 'number') {
+      this[kOptions].maxAwaitTimeMS = options.maxAwaitTimeMS ?? options.maxTimeMS;
```
addaleax, line 187: From: https://www.mongodb.com/docs/v6.0/reference/method/cursor.maxAwaitTimeMS/

> This method, [maxAwaitTimeMS()](https://www.mongodb.com/docs/v6.0/reference/method/cursor.maxAwaitTimeMS/#mongodb-method-cursor.maxAwaitTimeMS), sets a limit on how long a [tailable cursor](https://www.mongodb.com/docs/v6.0/core/tailable-cursors/#std-label-tailable-cursors-landing-page) waits for the next response. [maxTimeMS()](https://www.mongodb.com/docs/v6.0/reference/method/cursor.maxTimeMS/#mongodb-method-cursor.maxTimeMS) sets a limit on total processing time.

I assume this is wrong then? (I assume it is. The legacy shell also sets `getMoreCmd.maxTimeMS = this._maxAwaitTimeMS;`.)

nbbeeken: Yea it's a strange case:
From the spec: "`maxAwaitTimeMS` - this option is an alias for maxTimeMS, used on getMore commands." So this is correct only for getMores so this is the wrong place to do this overriding. I'm going to write up some kickoff notes on the JIRA and I think the approach is that cursors should construct their GetMoreOperations and that would give them more control over this option.

nbbeeken:
https://github.com/mongodb/node-mongodb-native/pull/3319
My git fu somehow auto closed this 🤔  started a new PR

PR #3315, opened 2022-07-07
feat(NODE-3517): improve index spec handling and type definitions

aditi-khare-mongoDB:
### Description
```parseIndexOptions()``` was a helper function that allowed users to multiple different input formattings for CreateIndex and normalizes them to one format.

The function was buggy, in the wrong file, and had some unused outputs, so I renamed it to ```getFieldHash``` and put it inside the ```makeIndexSpec``` function. 

Yes, 

1. Bug Fix: need to notify users that the tuple input formatting did not work correctly for IndexDirections other than 1
2. Map type is now an acceptable input formatting, and it is the only one that that will preserve numerical key ordering.

**Bug:**
the previous ```parseIndexOptions``` was buggy when a [string, IndexDirection] tuple was part of the input. 

Desired Output:
```{string: IndexDirection}```

Current Output:
(if IndexDirection is numerical) ```{string: 1}```
OR
(if IndexDirection is not numerical): ```{{string: 1} {IndexDirection: 1}```

**New Feature:**
Can now input a Map with a single key and value, or an array of length = 1 Maps as input.
** This is important as now we can preserve numerical key ordering **

Input Examples:
ex: ```new Map<string, IndexDirection>([['sample_index', -1]])```

ex2: ```[
        new Map<string, IndexDirection>([['sample_index1', 1]]),
        new Map<string, IndexDirection>([['sample_index2', -1]]),
        new Map<string, IndexDirection>([['sample_index3', '2d']])
      ]```

Addresses:
* https://jira.mongodb.org/browse/NODE-4446
* assertIndexNotExists operation in unified runner fix

PR Summary:
* Map as a valid input type in TS definition
* Uses Map under the hood to ensure key order is preserved, esp with FLE
* Tuples passed at the top level to `createIndex` were incorrectly parsed as string input
  * `createIndex(['myKey', 1])` would create `{ 'myKey': 1, '1': 1 }`. 
  * Now it's correctly detected if the second arg is one of the known index directions. 
  * We recommend using a Map to avoid all the edge cases here.
* Type strictness on this nesting of array (one or more)
* Type strictness for createIndexes aligned with createIndex
  * No longer accepts just Document, checks that the values are a known IndexDirection

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -594,11 +594,27 @@ export class CryptoConnection extends Connection {
       return;
     }
 
+    const sort: Map<string, number> | null = cmd.find || cmd.findAndModify ? cmd.sort : null;
+    const indexKeys: Map<string, number>[] | null = cmd.createIndexes
+      ? cmd.indexes.map((index: { key: Map<string, number> }) => index.key)
+      : null;
+
     autoEncrypter.encrypt(ns.toString(), cmd, options, (err, encrypted) => {
       if (err || encrypted == null) {
         callback(err, null);
         return;
       }
+
+      // Repair JS Map loss
+      if ((cmd.find || cmd.findAndModify) && sort != null) {
+        encrypted.sort = sort;
+      }
+      if (cmd.createIndexes && indexKeys != null) {
+        for (const [offset, index] of indexKeys.entries()) {
+          encrypted.indexes[offset].key = index;
+        }
+      }
+
       super.command(ns, encrypted, options, (err, response) => {
         if (err || response == null) {
           callback(err, response);
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -594,6 +594,12 @@ export class CryptoConnection extends Connection {
       return;
     }
 
+    // Save sort or indexKeys based on the command being run
+    // the encrypt API serializes our JS objects to BSON to pass to the native code layer
+    // and then deserializes the encrypted result, the protocol level components
+    // of the command (ex. sort) are then converted to JS objects potentially losing
+    // import key order information. These fields are never encrypted so we can save the values
+    // from before the encryption and replace them after encryption has been performed
     const sort: Map<string, number> | null = cmd.find || cmd.findAndModify ? cmd.sort : null;
     const indexKeys: Map<string, number>[] | null = cmd.createIndexes
       ? cmd.indexes.map((index: { key: Map<string, number> }) => index.key)
@@ -605,11 +611,11 @@ export class CryptoConnection extends Connection {
         return;
       }
 
-      // Repair JS Map loss
-      if ((cmd.find || cmd.findAndModify) && sort != null) {
+      // Replace the saved values
+      if (sort != null && (cmd.find || cmd.findAndModify)) {
         encrypted.sort = sort;
       }
-      if (cmd.createIndexes && indexKeys != null) {
+      if (indexKeys != null && cmd.createIndexes) {
         for (const [offset, index] of indexKeys.entries()) {
           encrypted.indexes[offset].key = index;
         }
--- a/src/operations/indexes.ts
+++ b/src/operations/indexes.ts
@@ -56,7 +56,6 @@ function isIndexDirection(x: any): x is IndexDirection {
     typeof x === 'number' || x === '2d' || x === '2dsphere' || x === 'text' || x === 'geoHaystack'
   );
 }
-
 /** @public */
 export type IndexSpecification = OneOrMore<
   | string
@@ -260,8 +259,6 @@ export class CreateIndexesOperation<
     this.indexes = normalizedIndexes;
   }
 
-  /* TODO: create name in the parent class constructor */
-  /* create a type assertion to stop typescript errors */
   override execute(
     server: Server,
     session: ClientSession | undefined,
--- a/src/operations/indexes.ts
+++ b/src/operations/indexes.ts
@@ -54,11 +54,7 @@ export type IndexDirection = -1 | 1 | '2d' | '2dsphere' | 'text' | 'geoHaystack'
 
 /** @public */
 export type IndexSpecification = OneOrMore<
-  | string
-  | [string, IndexDirection]
-  | { [key: string]: IndexDirection }
-  | [string, IndexDirection][]
-  | { [key: string]: IndexDirection }[]
+  string | [string, IndexDirection] | { [key: string]: IndexDirection }
 >;
 
 /** @public */
--- a/src/operations/indexes.ts
+++ b/src/operations/indexes.ts
@@ -1,3 +1,5 @@
+import { json } from 'stream/consumers';
+
 import type { Document } from '../bson';
 import type { Collection } from '../collection';
 import type { Db } from '../db';
@@ -6,7 +8,7 @@ import type { OneOrMore } from '../mongo_types';
 import { ReadPreference } from '../read_preference';
 import type { Server } from '../sdam/server';
 import type { ClientSession } from '../sessions';
-import { Callback, maxWireVersion, MongoDBNamespace, parseIndexOptions } from '../utils';
+import { Callback, isObject, maxWireVersion, MongoDBNamespace } from '../utils';
 import {
   CollationOptions,
   CommandOperation,
@@ -51,6 +53,11 @@ const VALID_INDEX_OPTIONS = new Set([
 
 /** @public */
 export type IndexDirection = -1 | 1 | '2d' | '2dsphere' | 'text' | 'geoHaystack' | number;
+function isIndexDirection(x: any): x is IndexDirection {
+  return (
+    typeof x === 'number' || x === '2d' || x === '2dsphere' || x === 'text' || x === 'geoHaystack'
+  );
+}
 
 /** @public */
 export type IndexSpecification = OneOrMore<
@@ -126,14 +133,42 @@ export interface CreateIndexesOptions extends CommandOperationOptions {
   hidden?: boolean;
 }
 
-function makeIndexSpec(indexSpec: IndexSpecification, options: any): IndexDescription {
-  const indexParameters = parseIndexOptions(indexSpec);
+export function makeIndexSpec(indexSpec: IndexSpecification, options: any): IndexDescription {
+  function getFieldHash(indexSpec: IndexSpecification) {
+    let fieldHash: Map<string, IndexDirection> = new Map();
+
+    let indexSpecArr: IndexSpecification[];
+
+    // wrap in array if needed
+    if (!Array.isArray(indexSpec) || (indexSpec.length === 2 && isIndexDirection(indexSpec[1]))) {
+      indexSpecArr = [indexSpec];
+    } else {
+      indexSpecArr = indexSpec;
+    }
+
+    // iterate through array and handle different types
+    indexSpecArr.forEach((f: any) => {
+      if ('string' === typeof f) {
+        fieldHash.set(f, 1);
+      } else if (Array.isArray(f)) {
+        fieldHash.set(f[0], f[1]);
+      } else if (isObject(f)) {
+        for (const [key, value] of Object.entries(f)) {
+          fieldHash.set(key, value);
+        }
+      }
+    });
+
+    return fieldHash;
+  }
+
+  const fieldHash = getFieldHash(indexSpec);
 
   // Generate the index name
-  const name = typeof options.name === 'string' ? options.name : indexParameters.name;
+  const name = typeof options.name === 'string' ? options.name : null;
 
   // Set up the index
-  const finalIndexSpec: Document = { name, key: indexParameters.fieldHash };
+  const finalIndexSpec: Document = { name, key: fieldHash };
 
   // merge valid index options into the index spec
   for (const optionName in options) {
@@ -272,8 +307,7 @@ export class CreateIndexOperation extends CreateIndexesOperation<string> {
     //   coll.createIndex({ a: 1 });
     //   coll.createIndex([['a', 1]]);
     // createIndexes is always called with an array of index spec objects
-
-    super(parent, collectionName, [makeIndexSpec(indexSpec, options)], options);
+    super(parent, collectionName, [makeIndexSpec(indexSpec, options)]);
   }
   override execute(
     server: Server,
--- a/src/operations/indexes.ts
+++ b/src/operations/indexes.ts
@@ -1,5 +1,3 @@
-import { json } from 'stream/consumers';
-
 import type { Document } from '../bson';
 import type { Collection } from '../collection';
 import type { Db } from '../db';
--- a/src/operations/indexes.ts
+++ b/src/operations/indexes.ts
@@ -59,7 +59,10 @@ function isIndexDirection(x: any): x is IndexDirection {
 
 /** @public */
 export type IndexSpecification = OneOrMore<
-  string | [string, IndexDirection] | { [key: string]: IndexDirection }
+  | string
+  | [string, IndexDirection]
+  | { [key: string]: IndexDirection }
+  | Map<string, IndexDirection>
 >;
 
 /** @public */
@@ -87,7 +90,7 @@ export interface IndexDescription
   > {
   collation?: CollationOptions;
   name?: string;
-  key: Document;
+  key: Document | Map<string, IndexDirection>;
 }
 
 /** @public */
@@ -131,31 +134,39 @@ export interface CreateIndexesOptions extends CommandOperationOptions {
   hidden?: boolean;
 }
 
+function isSingleIndexTuple(t: unknown): t is [string, IndexDirection] {
+  return Array.isArray(t) && t.length === 2 && isIndexDirection(t[1]);
+}
+
 export function makeIndexSpec(indexSpec: IndexSpecification, options: any): IndexDescription {
   function getFieldHash(indexSpec: IndexSpecification) {
-    let fieldHash: Map<string, IndexDirection> = new Map();
+    const fieldHash: Map<string, IndexDirection> = new Map();
 
     let indexSpecArr: IndexSpecification[];
 
-    // wrap in array if needed
-    if (!Array.isArray(indexSpec) || (indexSpec.length === 2 && isIndexDirection(indexSpec[1]))) {
+    // Wrap indexSpec in array if needed
+    if (!Array.isArray(indexSpec) || isSingleIndexTuple(indexSpec)) {
       indexSpecArr = [indexSpec];
     } else {
       indexSpecArr = indexSpec;
     }
 
-    // iterate through array and handle different types
-    indexSpecArr.forEach((f: any) => {
-      if ('string' === typeof f) {
-        fieldHash.set(f, 1);
-      } else if (Array.isArray(f)) {
-        fieldHash.set(f[0], f[1]);
-      } else if (isObject(f)) {
-        for (const [key, value] of Object.entries(f)) {
+    // Iterate through array and handle different types
+    for (const spec of indexSpecArr) {
+      if ('string' === typeof spec) {
+        fieldHash.set(spec, 1);
+      } else if (Array.isArray(spec)) {
+        fieldHash.set(spec[0], spec[1]);
+      } else if (spec instanceof Map) {
+        for (const [key, value] of spec) {
+          fieldHash.set(key, value);
+        }
+      } else if (isObject(spec)) {
+        for (const [key, value] of Object.entries(spec)) {
           fieldHash.set(key, value);
         }
       }
-    });
+    }
 
     return fieldHash;
   }
@@ -225,9 +236,32 @@ export class CreateIndexesOperation<
     this.options = options ?? {};
     this.collectionName = collectionName;
 
-    this.indexes = indexes;
+    // Ensure we generate the correct name if the parameter is not set
+    const normalizedIndexes = [];
+    for (const userIndex of indexes) {
+      const key =
+        userIndex.key instanceof Map ? userIndex.key : new Map(Object.entries(userIndex.key));
+      const index: Omit<IndexDescription, 'key'> & { key: Map<string, IndexDirection> } = {
+        ...userIndex,
+        key
+      };
+      if (index.name == null) {
+        const keys = [];
+
+        for (const [name, direction] of index.key) {
+          keys.push(`${name}_${direction}`);
+        }
+
+        // Set the name
+        index.name = keys.join('_');
+      }
+      normalizedIndexes.push(index);
+    }
+    this.indexes = normalizedIndexes;
   }
 
+  /* TODO: create name in the parent class constructor */
+  /* create a type assertion to stop typescript errors */
   override execute(
     server: Server,
     session: ClientSession | undefined,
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

**Comment by nbbeeken** [comment]
Requesting Review from: @aditi-khare-mongoDB (GH permission issues)

File: test/integration/client-side-encryption/driver.test.ts [hunk 1]
```diff
@@ -176,33 +183,20 @@ describe('Client Side Encryption Functional', function () {
...
+      { bsonRegExp: true },
+      { ignoreUndefined: true }
     ];
 
-    testCases.forEach(bsonOptions => {
+    for (const bsonOptions of testCases) {
       const name = `should respect bson options ${JSON.stringify(bsonOptions)}`;
 
-      it(name, metadata, function () {
+      it(name, metadata, async function () {
```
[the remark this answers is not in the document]
nbbeeken: As a courtesy to our future selves we sometimes will do drive by modernizations of code if it is being touched by the current work. This test failed momentarily and I noticed we could fix it to stop using containsSubset and could use async await.

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -105,10 +115,12 @@ describe('Client Side Encryption Functional', function () {
...
 
   describe('BSON Options', function () {
+    let client: MongoClient;
+
     beforeEach(function () {
-      this.client = this.configuration.newClient();
+      client = this.configuration.newClient();
 
-      const noop = () => {};
+      const noop = () => null;
```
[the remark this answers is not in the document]
nbbeeken: Our linter complains about empty functions, rather than disable the rule, we've used `return null` in a few places, it does have different meaning, we should keep in mind that this means the function will return something different from before. But it has served us to offer meaning that the result value is "omitted".

test/unit/operations/indexes.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      { a: 1 },
+      // @ts-expect-error: Testing bad options get filtered
+      { randomOptionThatWillNeverBeAdded: true, storageEngine: { iLoveJavascript: 1 } }
+    );
+    expect(indexOutput.indexes).to.have.lengthOf(1);
+    expect(indexOutput.indexes[0]).to.have.property('key').that.is.instanceOf(Map);
+    expect(indexOutput.indexes[0]).to.have.property('name', 'a_1');
+    expect(indexOutput.indexes[0])
+      .to.have.property('storageEngine')
+      .that.deep.equals({ iLoveJavascript: 1 });
```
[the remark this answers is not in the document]
nbbeeken: I too love javascript, thanks!

test/integration/client-side-encryption/driver.test.ts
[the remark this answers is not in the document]
aditi-khare-mongoDB: That sounds good, and is good to know for future implementations!

src/cmap/connection.ts
```diff
@@ -594,11 +594,27 @@ export class CryptoConnection extends Connection {
...
+      ? cmd.indexes.map((index: { key: Map<string, number> }) => index.key)
+      : null;
+
     autoEncrypter.encrypt(ns.toString(), cmd, options, (err, encrypted) => {
       if (err || encrypted == null) {
         callback(err, null);
         return;
       }
+
+      // Repair JS Map loss
```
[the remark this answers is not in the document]
nbbeeken: Yes, I expanded on the comment. I put a note on our tech adgenda for this, the ticket would be rather involved so we should collect some ideas about different approaches first.

test/tools/spec-runner/index.js
```diff
@@ -485,6 +485,18 @@ function validateExpectations(commandEvents, spec, savedSessionData) {
       actualCommand.sort = { [expectedKey]: actualCommand.sort.get(expectedKey) };
     }
 
+    if (expectedCommand.createIndexes) {
+      for (const [i, dbIndex] of actualCommand.indexes.entries()) {
+        expect(Object.keys(expectedCommand.indexes[i].key)).to.have.lengthOf(1);
+        expect(dbIndex.key).to.be.instanceOf(Map);
+        expect(dbIndex.key.size).to.equal(1);
```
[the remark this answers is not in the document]
nbbeeken: Done!

test/integration/client-side-encryption/driver.test.ts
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good!

src/operations/indexes.ts
```diff
@@ -51,14 +51,17 @@ const VALID_INDEX_OPTIONS = new Set([
...
+  );
+}
 /** @public */
 export type IndexSpecification = OneOrMore<
   | string
   | [string, IndexDirection]
   | { [key: string]: IndexDirection }
-  | [string, IndexDirection][]
-  | { [key: string]: IndexDirection }[]
+  | Map<string, IndexDirection>
```
nbbeeken, line 64: @dariakp @baileympearson @durran Perhaps we omit the Map as a possible value from here and in `IndexDescription`. This just adds an inconsistency we haven't done elsewhere in our APIs I'm thinking we might want to consider a better approach if we want our APIs to effectively communicate what it supported. For example, if `Document` was actually separated `InputDocument` and `OutputDocument` the idea would be that everywhere where the driver accepts some kind of BSON object it could actually accept:
```ts
type InputDocument = Record<string, any> | Map<string, any>
```
This small example obviously has big limitations but perhaps there is some better driver wide solution to the fact that Maps can be accepted anywhere BSON documents are taken. Adding this Map here now could prevent us from nicely solving this in the future, (ex. the Document type doesn't have assertions on the value type but the Map does). Thoughts?

src/cmap/connection.ts
```diff
@@ -594,11 +594,33 @@ export class CryptoConnection extends Connection {
       return;
     }
 
+    // Save sort or indexKeys based on the command being run
```
baileympearson, line 597: This doesn't feel like the correct solution.  If I understand the issue correctly, when libmongocrypt sends back an encrypted command as a serialized bson object, we deserialize it into a JS object to send to the server.  Why wouldn't we fix the problem by deserializing into a map instead of an object in mongodb-client-encryption instead of working around it in the driver?

src/operations/indexes.ts
```diff
@@ -196,7 +219,25 @@ export class CreateIndexesOperation<
     this.options = options ?? {};
     this.collectionName = collectionName;
 
-    this.indexes = indexes;
+    // Ensure we generate the correct name if the parameter is not set
```
baileympearson, line 222: Can we move this code into a function?  We discussed this a bit in the office, but I still think for readability it should be broken out.  The fact that a comment is necessary to explain what this block of code is doing strongly indicates that this should be broken out into a function.

test/unit/operations/indexes.test.ts
```diff
@@ -0,0 +1,143 @@
+import { expect } from 'chai';
+
+import {
+  CreateIndexesOptions,
+  CreateIndexOperation,
+  IndexDirection
+} from '../../../src/operations/indexes';
+import { ns } from '../../../src/utils';
+
+describe('makeIndexSpec()', () => {
```
baileympearson, line 10:
```suggestion
describe('CreateIndexOperation', () => {
  context('#constructor()', () => {
   ...
```

baileympearson (changes requested):

test/unit/operations/indexes.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  const testCases = [
+    {
+      description: 'single string',
```
baileympearson, line 13: Can we add tests for the index name as well?

src/operations/indexes.ts
```diff
@@ -51,14 +51,17 @@ const VALID_INDEX_OPTIONS = new Set([
 
 /** @public */
 export type IndexDirection = -1 | 1 | '2d' | '2dsphere' | 'text' | 'geoHaystack' | number;
-
+function isIndexDirection(x: any): x is IndexDirection {
```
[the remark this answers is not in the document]
nbbeeken: sure done!

src/operations/indexes.ts
```diff
@@ -51,14 +51,17 @@ const VALID_INDEX_OPTIONS = new Set([
...
+  return (
+    typeof x === 'number' || x === '2d' || x === '2dsphere' || x === 'text' || x === 'geoHaystack'
+  );
+}
 /** @public */
 export type IndexSpecification = OneOrMore<
   | string
   | [string, IndexDirection]
   | { [key: string]: IndexDirection }
```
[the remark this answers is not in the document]
nbbeeken: Of the three we're addressing here should we pull them each out and try and make this PR only about the refactor (i.e. why this bug but not the others)? (I think the tuple one might be difficult but we can see).

src/operations/indexes.ts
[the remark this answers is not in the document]
nbbeeken: Yep, good idea, added!

src/cmap/connection.ts
nbbeeken: Would the code not look the same just in `mongodb-client-encryption`? The layer at which BSON is deserialized is still inside the BSON library so performing the logic seen here would be nearly identical, just within the `encrypt` method. Since it's something we would have to test in this repo anyway, (at least integration test) feels like the right spot is here.

src/operations/indexes.ts
nbbeeken: Sure done!

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -215,10 +209,86 @@ describe('Client Side Encryption Functional', function () {
...
+      });
+    }
+  });
+
+  describe('key order aware command properties', () => {
+    let client: MongoClient;
+    let collection: Collection;
+
+    beforeEach(async function () {
+      if (this.configuration.clientSideEncryption == null) {
```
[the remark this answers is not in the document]
nbbeeken: It is to avoid the dependency missing error that will be thrown by the newClient call that takes an autoEncryption option.

test/tools/unified-spec-runner/runner.ts
```diff
@@ -256,8 +256,8 @@ export function runUnifiedSuite(
 ): void {
   for (const unifiedSuite of specTests) {
     context(String(unifiedSuite.description), function () {
-      for (const test of unifiedSuite.tests) {
-        it(String(test.description), async function () {
+      for (const [index, test] of unifiedSuite.tests.entries()) {
+        it(String(test.description === '' ? `Test ${index}` : test.description), async function () {
```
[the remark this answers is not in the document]
nbbeeken: The FLE tests brought in empty test names, I can revert this, but the logs look odd with the little skip arrow symbol we have pointing to an empty line. I figure enumerating the tests in order they are parsed is a good way to find them later

test/unit/operations/indexes.test.ts
nbbeeken: fixed!

test/unit/operations/indexes.test.ts
nbbeeken: Is that not what the `should set name correctly...` it block does?

src/operations/indexes.ts
```diff
@@ -196,7 +219,25 @@ export class CreateIndexesOperation<
...
+        // Ensure the key is a Map to preserve index key ordering
+        key: userIndex.key instanceof Map ? userIndex.key : new Map(Object.entries(userIndex.key))
+      };
+      if (index.name == null) {
+        // Ensure every index is named
+        const keys = [];
+        for (const [name, direction] of index.key) {
+          keys.push(`${name}_${direction}`);
+        }
+        index.name = keys.join('_');
```
dariakp, line 236: I'm pretty sure the loop above is equivalent to:
`index.name = Array.from(index.key).flat().join('_');`

test/tools/unified-spec-runner/operations.ts
```diff
@@ -76,14 +76,19 @@ operations.set('assertIndexNotExists', async ({ operation, client }) => {
   const collection = client
     .db(operation.arguments.databaseName)
     .collection(operation.arguments.collectionName);
+
+  const listIndexCursor = collection.listIndexes();
+  let indexes;
   try {
-    expect(await collection.indexExists(operation.arguments.indexName)).to.be.true;
+    indexes = await listIndexCursor.toArray();
```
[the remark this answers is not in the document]
nbbeeken: It asserted the reverse of what the operation expects (expected indexExists to return true for the assertIndexNotExists operation.) So it needed to be flipped but also we made the code the same as the code for assertExists, listIndexes is spec-ed so seems like we should use that in the spec test runner.

src/operations/indexes.ts
```diff
@@ -196,7 +219,25 @@ export class CreateIndexesOperation<
...
+      }
+      namedIndexes.push(index);
+    }
+    this.indexes = namedIndexes;
```
dariakp, line 240: Since we aren't changing the length or rearranging the indexes, I feel like this might be more straightforwardly expressed as
`this.indexes = indexes.map( userIndex => { ... return index; } )`

dariakp (changes requested):

src/operations/indexes.ts
nbbeeken: done!

src/operations/indexes.ts
nbbeeken: done, it's still pulled out to a helper method but perhaps the code is streamlined enough to be in the constructor again. Thoughts @baileympearson?

src/operations/indexes.ts
```diff
@@ -195,8 +228,7 @@ export class CreateIndexesOperation<
 
     this.options = options ?? {};
     this.collectionName = collectionName;
-
-    this.indexes = indexes;
+    this.indexes = fillInIndexNames(indexes);
```
dariakp, line 231: Since we're actually normalizing the index key types here, should we update the `indexes` class property type to reflect that key will always be a map?

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -215,10 +209,86 @@ describe('Client Side Encryption Functional', function () {
...
+
+    afterEach(async () => {
+      if (client) await client.close();
+    });
+
+    describe('find', () => {
+      it('should maintain ordered sort', metadata, async function () {
+        const events: CommandStartedEvent[] = [];
+        client.on('commandStarted', ev => events.push(ev));
+        const sort: [string, SortDirection][] = [
```
dariakp, line 250: should we freeze this object to make sure it's not being mutated (same in the other test cases in this file)? otherwise the event equality check is less meaningful

test/tools/spec-runner/index.js
```diff
@@ -485,6 +485,22 @@ function validateExpectations(commandEvents, spec, savedSessionData) {
...
+      // to indicate whether order matters for any given key and set general
+      // expectations accordingly
+      for (const [i, dbIndex] of actualCommand.indexes.entries()) {
+        expect(Object.keys(expectedCommand.indexes[i].key)).to.have.lengthOf(1);
+        expect(dbIndex.key).to.be.instanceOf(Map);
+        expect(dbIndex.key.size).to.equal(1);
+      }
+      actualCommand.indexes = actualCommand.indexes.map(dbIndex => ({
+        ...dbIndex,
+        key: Object.fromEntries(dbIndex.key.entries())
```
dariakp, line 500: I don't think you need to call `key.entries()` here since you're checking that key is a map instance above, so you should be able to just do `fromEntries(dbIndex.key)`

test/tools/unified-spec-runner/match.ts
```diff
@@ -166,6 +166,16 @@ export function resultCheck(
       expect(actual[key]).to.have.all.keys(expectedSortKey);
       const objFromActual = { [expectedSortKey]: actual[key].get(expectedSortKey) };
       resultCheck(objFromActual, value, entities, path, checkExtraKeys);
+    } else if (key === 'createIndexes') {
+      for (const index of actual.indexes) {
+        expect(index).to.have.property('key').that.is.instanceOf(Map);
+        expect(
```
dariakp, line 172: I think we should be checking the size of the expected key and not just the actual one (similar to the above), since it's a limitation of the test framework (obviously the current code will still fail if the expected has more keys than the actual, but this way we're making it clear that it's a test input expectation incompatibility and not necessarily something wrong with the *result* of the test)

test/types/community/createIndex.test-d.ts
```diff
@@ -11,3 +11,16 @@ const indexName = collection.createIndex({}, options);
...
 expectType<Promise<string>>(indexName);
 expectType<Document | undefined>(options.partialFilterExpression);
+
+// One
+collection.createIndex('someKey');
+collection.createIndex(['someKey', 1]);
+collection.createIndex(new Map([['someKey', 1]]));
+collection.createIndex({ a: 1, b: -1 });
+collection.createIndex({ a: '2dsphere', b: -1 });
+// OrMore
```
dariakp, line 21: shouldn't we have a test with more than one array element? maybe mix & match types?

test/unit/operations/indexes.test.ts
dariakp: I think Bailey is looking for a test case where the name *is* explicitly provided

dariakp (changes requested):

test/integration/client-side-encryption/driver.test.ts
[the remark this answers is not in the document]
baileympearson: So if clientSideEncryption is nullish, we return.  Is that supposed to skip these tests?  If so, we should explicitly skip them and attach a skip reason.

Also, if that's the intention, should we follow the precedent we just set and use `this.configuration.clientEncryption.enabled`?

src/cmap/connection.ts
baileympearson: I was imagining we'd build a feature into bson to serialize directly into maps instead of objects (if it doesn't exist), and then configure bson to use that for everything in FLE.  With eventual on-demand-bson we could combine this approach with that, and selectively deserialize directly into maps but handle it in bson with a configuration.

src/operations/indexes.ts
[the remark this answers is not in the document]
baileympearson: what "three" are you referencing here? three bugs?

src/operations/indexes.ts
baileympearson: I'm fine either way

test/integration/client-side-encryption/driver.test.ts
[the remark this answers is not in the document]
nbbeeken: It's skipped by the metadata attached to each test

baileympearson (changes requested):
requesting changes for visibility

test/types/community/createIndex.test-d.ts
nbbeeken: Added!

test/tools/unified-spec-runner/match.ts
nbbeeken: Yea good call

test/tools/spec-runner/index.js
nbbeeken: done!

test/integration/client-side-encryption/driver.test.ts
nbbeeken: Done!

src/operations/indexes.ts
nbbeeken: Done!

test/unit/operations/indexes.test.ts
nbbeeken: added!

src/operations/indexes.ts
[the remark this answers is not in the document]
nbbeeken: - Map in TS
- Key order FLE
- Tuple parsing issue
- Type strictness on this nesting of array (one or more)
- Type strictness for createIndexes aligned with createIndex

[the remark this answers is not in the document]
dariakp: I think whether we pull these out into separate PRs/fixes or not, we *should* make sure that the PR description accurately reflects whatever fixes are included and that we are *super* clear in our release notes about the changes in behavior. It's also worth noting that the types in the documentation won't get updated on a patch, unless we manually regenerate the 4.8, which could be confusing. I think because of the scope of the changes here, this set of improvements that's coming in as a bug fix is better marked as a `feat:`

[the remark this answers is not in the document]
dariakp: Oh and let's triple check to make sure that every single one of those things has full test coverage

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -0,0 +1,297 @@
...
+        };
+      }
+
+      let encryption: ClientEncryption;
+      let dataDb: Db;
+      let keyVaultDb: Db;
+
+      const mongodbClientEncryption = this.configuration.mongodbClientEncryption;
+      const kmsProviders = this.configuration.kmsProviders(crypto.randomBytes(96));
+      return client
```
dariakp, line 141: just wondering why we aren't using async/await here instead

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -0,0 +1,297 @@
...
+        expect(findEvent).to.have.property('commandName', 'find');
+        expect(findEvent).to.have.nested.property('command.sort').deep.equal(new Map(sort));
+      });
+    });
+
+    describe('findAndModify', () => {
+      it('should maintain ordered sort', metadata, async function () {
+        const events: CommandStartedEvent[] = [];
+        client.on('commandStarted', ev => events.push(ev));
+        const sort: [string, SortDirection][] = [
```
dariakp, line 267: pointing out this case since you only addressed the freeze in the `find` function test case

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -0,0 +1,297 @@
...
+          .to.have.nested.property('command.sort')
+          .deep.equal(new Map(sort));
+      });
+    });
+
+    describe('createIndexes', () => {
+      it('should maintain ordered index keys', metadata, async function () {
+        const events: CommandStartedEvent[] = [];
+        client.on('commandStarted', ev => events.push(ev));
+        const indexDescription: [string, IndexDirection][] = [
```
dariakp, line 284: pointing out this case since you only addressed the freeze in the `find` function test case

test/integration/client-side-encryption/driver.test.ts
dariakp: There's 2 more cases

test/integration/client-side-encryption/driver.test.ts
nbbeeken: fixed

test/integration/client-side-encryption/driver.test.ts
nbbeeken: fixed

test/integration/client-side-encryption/driver.test.ts
nbbeeken: This was a drive by TS convert, but don't mind if I do!

src/operations/indexes.ts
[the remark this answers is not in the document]
nbbeeken: Collected a summary in the description and between the type/fle/indexes.test.ts additions I see coverage for each point.

baileympearson (approved):

dariakp (approved):

PR #3319, opened 2022-07-13
fix(NODE-4413): set maxTimeMS on getMores when maxAwaitTimeMS is specified

nbbeeken:
#### What is changing?

- Capture maxAwaitTimeMS option on cursor options
- Add documentation to options
- Move error handling to getMore.execute (I can put this back, it seems like the right place in terms of handling issues)
- GetMore constructor must always take in options
- Added tests to maxTimeMS.test.ts and converted to TS
- Added tests to change streams.

Fix maxAwaitTimeMS not being respected by change stream getMores (all getMores were impacted actually)


test/integration/crud/maxTimeMS.test.ts
```diff
@@ -0,0 +1,268 @@
...
+    let cursor: FindCursor<{ _id: number }>;
+    let events: CommandStartedEvent[];
+
+    beforeEach(async function () {
+      client = this.configuration.newClient({ monitorCommands: true });
+      await client
+        .db()
+        .dropCollection('cappedAt3')
+        .catch(() => null);
+      await sleep(100); // Need to make sure it has had time to drop entirely
```
baileympearson, line 144: Why?  Does dropCollection return eagerly?

test/integration/crud/maxTimeMS.test.ts
```diff
@@ -0,0 +1,268 @@
...
+    afterEach(async function () {
+      events = [];
+      await cursor?.close();
+      await client?.close();
+    });
+
+    const tailableValues = [true, false, undefined];
+    const awaitDataValues = [true, false, undefined];
+    const maxTimeMSValues = [100, 0, undefined];
+    const maxAwaitTimeMSValues = [100, 0, undefined];
```
baileympearson, line 165: How were these values chosen?  Is it a >0, 0 and undefined value?  

Should we also be testing with negative values?

src/cursor/abstract_cursor.ts
```diff
@@ -617,21 +645,8 @@ export abstract class AbstractCursor<
...
-    }
-
-    if (server == null) {
-      callback(new MongoRuntimeError('Unable to iterate cursor without selected server'));
-      return;
-    }
-
-    const getMoreOperation = new GetMoreOperation(cursorNs, cursorId, server, {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const getMoreOperation = new GetMoreOperation(this[kNamespace], this[kId]!, this[kServer]!, {
```
baileympearson, line 649: Do we have tests that test for errors being thrown when execute is called on a zero-ed cursor or if the server is undefined?  If not, can we add them to demonstrate this behavior still exists?

baileympearson (changes requested):

test/integration/crud/maxTimeMS.test.ts
nbbeeken: I shouldn't need this, in sharded cluster topologies reusing a namespace that was just dropped isn't reccomended IIRC and I thought I was running into that. dropped.

test/integration/crud/maxTimeMS.test.ts
nbbeeken: I don't think we need negative tests here, MongoDB will error. I tested zero because its a falsey case, and 100 is just a postive value

src/cursor/abstract_cursor.ts
nbbeeken: Added! :)

test/unit/operations/get_more.test.ts
```diff
@@ -192,6 +192,54 @@ describe('GetMoreOperation', function () {
         }
       });
     });
+
+    context('error cases', () => {
```
dariakp, line 196: what about if server is undefined, or was that code always unreachable?

test/integration/crud/maxTimeMS.test.ts
```diff
@@ -0,0 +1,266 @@
...
+  MongoServerError
+} from '../../../src';
+import { assert as test, setupDatabase } from '../shared';
+
+describe('MaxTimeMS', function () {
+  before(function () {
+    return setupDatabase(this.configuration);
+  });
+
+  it('should correctly respect the maxTimeMS property on count', function (done) {
```
dariakp, line 19: can we do async/await in these tests? moving them to ts already effectively makes it a brand new file to review and it's only a handful of tests so shouldn't add much scope to refactor but will make it easier to review the actual tests

dariakp (changes requested):

test/unit/operations/get_more.test.ts
nbbeeken: Should always be unreachable, server selection would have to fail, and we would have something wrong in our error handling of server selection if we proceeded to execute with a nullish server.

test/integration/crud/maxTimeMS.test.ts
nbbeeken: Done!

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1372,23 +1372,98 @@ describe('Change Streams', function () {
...
+        .expectEvents({
+          client: 'client0',
+          ignoreExtraEvents: true, // Sharded clusters have extra getMores
+          events: [
+            { commandStartedEvent: { commandName: 'aggregate', command: { maxTimeMS: 5000 } } },
+            { commandStartedEvent: { commandName: 'insert' } },
+            {
+              commandStartedEvent: {
+                commandName: 'getMore',
+                command: { maxTimeMS: { $$exists: false } }
```
dariakp, line 1442: it's not clear from the test case name why it should not exist on the aggregate command but not the getMore command here

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1372,23 +1372,98 @@ describe('Change Streams', function () {
...
+        })
+        .operation({
+          object: 'changeStreamOnClient',
+          name: 'iterateUntilDocumentOrError',
+          ignoreResultAndError: true
+        })
+        .expectEvents({
+          client: 'client0',
+          events: [
+            { commandStartedEvent: { commandName: 'aggregate' } },
```
dariakp, line 1407: why do we assert absence of the prop on the getMore in the 2nd case, but not on the initial aggregate here? what's the actual expected behavior from these options in the two cases?

test/integration/change-streams/change_stream.test.ts
nbbeeken: Added a case to demonstrate what should happen when both options are provided, as well as updated test names.

test/integration/change-streams/change_stream.test.ts
nbbeeken: Added an (not) existence check, and clarified the titles

baileympearson (approved):

dariakp (approved):

PR #3330, opened 2022-07-24
Remove @types/mongodb as it's included in the library

dolanmiu:
There is no need to install the `@types/mongodb` separately as it's already included in the main package

The `@types` is removed from `DefinitelyTyped` already:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/notNeededPackages.json#L2956-L2958

And also NPM says:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,12 +65,6 @@ yarn add mongodb
 
 This will download the MongoDB driver and add a dependency entry in your `package.json` file.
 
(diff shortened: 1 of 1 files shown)


durran (approved):

durran:
Thanks for the PR @dolanmiu but we are actually recommending to install `@types/node` in our docs, not `@types/mongodb` which has been removed from definitely typed.

PR #3333, opened 2022-07-26
test(NODE-4463): download shared lib in evergreen

durran:
### Description

Downloads the crypt shared lib for the appropriate version and sets its path in the environment.

#### What is changing?

Adds a new evergreen task that will download the crypt shared lib before running the csfle custom tests. This does not include the changes to the test runner to use it - that is the next subtask.

https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_next_ubuntu1804_custom_dependency_tests_run_custom_csfle_tests_patch_8ecbabcc4dec6c77db4e5ce5ef4a2ca5629c4608_62e12ede2a60ed018c89d3bc_22_07_27_12_26_07/0?type=T#L1437

None.

NODE-4463/NODE-4089

### DIFF
--- a/.evergreen/config.in.yml
+++ b/.evergreen/config.in.yml
@@ -818,9 +818,16 @@ functions:
           . ./prepare_client_encryption.sh
           rm -f ./prepare_client_encryption.sh
 
+          MONGODB_VERSION=${VERSION}
+          if [ -z "$MONGODB_VERSION" ]; then
+            # default to latest to match behavior of run-orchestration.sh.
+            MONGODB_VERSION=latest
+          fi
+
           . $DRIVERS_TOOLS/.evergreen/download-mongodb.sh
           get_distro
           # get_distro defines $DISTRO.
+          echo "distro='$DISTRO' version='$MONGODB_VERSION'".
           get_mongodb_download_url_for "$DISTRO" "$MONGODB_VERSION"
           # get_mongodb_download_url_for defines $MONGO_CRYPT_SHARED_DOWNLOAD_URL and $EXTRACT.
           if [ -z "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" ]; then
--- a/.evergreen/config.in.yml
+++ b/.evergreen/config.in.yml
@@ -798,12 +798,12 @@ functions:
         script: |
           ${PREPARE_SHELL}
           cat <<EOT > prepare_client_encryption.sh
-          export CLIENT_ENCRYPTION=${CLIENT_ENCRYPTION}
+          export CLIENT_ENCRYPTION='${CLIENT_ENCRYPTION}'
           export CSFLE_KMS_PROVIDERS='${CSFLE_KMS_PROVIDERS}'
-          export AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}"
-          export AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}"
-          export CSFLE_GIT_REF="${CSFLE_GIT_REF}"
-          export CDRIVER_GIT_REF="${CDRIVER_GIT_REF}"
+          export AWS_ACCESS_KEY_ID='${AWS_ACCESS_KEY_ID}'
+          export AWS_SECRET_ACCESS_KEY='${AWS_SECRET_ACCESS_KEY}'
+          export CSFLE_GIT_REF='${CSFLE_GIT_REF}'
+          export CDRIVER_GIT_REF='${CDRIVER_GIT_REF}'
           EOT
     - command: shell.exec
       type: test
--- a/.evergreen/config.in.yml
+++ b/.evergreen/config.in.yml
@@ -818,40 +818,8 @@ functions:
           . ./prepare_client_encryption.sh
           rm -f ./prepare_client_encryption.sh
 
-          MONGODB_VERSION=${VERSION}
-          if [ -z "$MONGODB_VERSION" ]; then
-            # default to latest to match behavior of run-orchestration.sh.
-            MONGODB_VERSION=latest
-          fi
-
-          . $DRIVERS_TOOLS/.evergreen/download-mongodb.sh
-          get_distro
-          # get_distro defines $DISTRO.
-          echo "distro='$DISTRO' version='$MONGODB_VERSION'".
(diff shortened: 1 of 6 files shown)


.evergreen/config.in.yml
```diff
@@ -789,6 +789,73 @@ functions:
...
+          EOT
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: "src"
+        timeout_secs: 60
+        script: |
+          ${PREPARE_SHELL}
+
+          # Disable xtrace (just in case it was accidentally set).
```
baileympearson, line 816: Instead of inlining a shell script here - could we break this into a separate script?  There's no way to manually run individual evergreen tasks locally, but we can run bash scripts.  That could be beneficial in the future for setting environments up locally the same as they're configured in CI.

baileympearson (changes requested):

.evergreen/config.in.yml
durran: I've moved out into its own script now.

baileympearson (dismissed):
These changes look good to me.  One additional change I'd like is to update the runner so that it prints out whether or not the shard library is enabled before tests run, as a part of this block:

Screen Shot 2022-07-28 at 9 55 00 AM

But since the next subtask is concerned with updating the runner, I think it's okay to make this change as a part of NODE-4483.

durran:
> These changes look good to me. One additional change I'd like is to update the runner so that it prints out whether or not the shard library is enabled before tests run, as a part of this block:
> 
> Screen Shot 2022-07-28 at 9 55 00 AM
> 
> But since the next subtask is concerned with updating the runner, I think it's okay to make this change as a part of [NODE-4483](https://jira.mongodb.org/browse/NODE-4483).

Yeah I'll add it to that PR.

.evergreen/config.in.yml
```diff
@@ -789,6 +789,41 @@ functions:
...
+        working_dir: "src"
+        script: |
+          ${PREPARE_SHELL}
+          cat <<EOT > prepare_client_encryption.sh
+          export CLIENT_ENCRYPTION=${CLIENT_ENCRYPTION}
+          export CSFLE_KMS_PROVIDERS='${CSFLE_KMS_PROVIDERS}'
+          export AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}"
+          export AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}"
+          export CSFLE_GIT_REF="${CSFLE_GIT_REF}"
+          export CDRIVER_GIT_REF="${CDRIVER_GIT_REF}"
```
addaleax, line 806: nit, feel free to ignore: If the differences in quoting here are intentional, I would leave a comment about that, otherwise I'd use single quotes everywhere

test/readme.md
```diff
@@ -408,6 +408,8 @@ The following steps will walk you through how to run the tests for CSFLE.
 
    The output of the tests will include sections like "Client Side Encryption Corpus," "Client Side Encryption Functional," "Client Side Encryption Prose Tests," and "Client Side Encryption."
 
+  To run the functional tests using the crypt shared lirary instead of mongocryptd, download the appropriate version of the crypt shared library for the enterprise server version [here](https://www.mongodb.com/download-center/enterprise/releases) and then set the location of it in the environment variable `CRYPT_SHARED_LIB_PATH`.
```
addaleax, line 411: typo:

```suggestion
  To run the functional tests using the crypt shared library instead of mongocryptd, download the appropriate version of the crypt shared library for the enterprise server version [here](https://www.mongodb.com/download-center/enterprise/releases) and then set the location of it in the environment variable `CRYPT_SHARED_LIB_PATH`.
```

addaleax (approved):

.evergreen/config.in.yml
```diff
@@ -789,6 +789,41 @@ functions:
...
+    - command: shell.exec
+      type: test
+      params:
+        silent: true
+        working_dir: "src"
+        script: |
+          ${PREPARE_SHELL}
+          cat <<EOT > prepare_client_encryption.sh
+          export CLIENT_ENCRYPTION=${CLIENT_ENCRYPTION}
+          export CSFLE_KMS_PROVIDERS='${CSFLE_KMS_PROVIDERS}'
```
dariakp, line 802: I want to echo Anna's comment here about consistent quote use; on the one hand, it's kind of a nit, but on the other hand, we *should* fix the typo in the readme, and as long as we're pushing changes to this PR, we might as well make the files internally consistent

dariakp (changes requested):

.evergreen/config.in.yml
durran: I've made this all single quoted now.

.evergreen/config.in.yml
durran: It's all single quoted now.

durran:
Rebased and updated shared lib tests to point to pinned commit and master as well.

dariakp (approved):

.evergreen/generate_evergreen_tasks.js
```diff
@@ -537,7 +537,7 @@ const oneOffFuncAsTasks = oneOffFuncs.map(oneOffFunc => ({
 }));
 
 oneOffFuncAsTasks.push({
-  name: 'run-custom-csfle-tests-pinned-commit',
+  name: 'run-custom-csfle-tests-pinned-commit-mongocryptd',
```
baileympearson, line 540:
```suggestion
  name: 'run-custom-csfle-tests-mongocryptd-pinned-commit',
```

non-blocking - this is just for consistency with the shared library tests.

.evergreen/generate_evergreen_tasks.js
```diff
@@ -564,7 +564,7 @@ oneOffFuncAsTasks.push({
 });
 
 oneOffFuncAsTasks.push({
-  name: 'run-custom-csfle-tests-master',
+  name: 'run-custom-csfle-tests-master-mongocryptd',
```
baileympearson, line 567:
```suggestion
  name: 'run-custom-csfle-tests-mongocryptd-master',
```

non-blocking - this is just for consistency with the shared library tests.

baileympearson (approved):
two non-blocking suggestions

baileympearson (approved):

PR #3334, opened 2022-07-26
test(NODE-4395): add csfle master && pinned commit tasks to CI

baileympearson:
#### What is changing?

A new task as been added to the custom dependency tests to test against libmongocrypt master.  The existing task was also renamed to indicate that it tests against a pinned commit, instead of the latest master.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1621,7 +1621,7 @@ tasks:
       - func: run bson-ext test
         vars:
           NODE_LTS_NAME: erbium
-  - name: run-custom-csfle-tests
+  - name: run-custom-csfle-tests-pinned-commit
     tags:
       - run-custom-dependency-tests
     commands:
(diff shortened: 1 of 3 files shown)


durran (changes requested):
I feel like this: https://github.com/mongodb/node-mongodb-native/blob/main/.evergreen/run-custom-csfle-tests.sh#L30 should maybe default back to master like before, so if we miss the variable in the future somewhere with new potential tasks we don't accidentally point at an old commit if we've bumped elsewhere. But I'm not too opinionated about it and am fine if it stays as is.

durran (approved):

dariakp (approved):

dariakp (changes requested):
Quick question - I noticed both of the tasks that ran in the CI for this PR seem to have downloaded the same commit, is the variable currently set somewhere else, too, that's overriding master?

dariakp (approved):

PR #3335, opened 2022-07-26
test(NODE-4331): add prose test 13

baileympearson:
#### What is changing?

CSFLE Prose test 13 is implemented, finishing up the key management API work (with the exception of releasing mongodb-client-encryption).

This PR address the following ticket:

- https://jira.mongodb.org/browse/NODE-4331

No.

### DIFF
--- a/test/integration/client-side-encryption/client_side_encryption.prose.test.js
+++ b/test/integration/client-side-encryption/client_side_encryption.prose.test.js
@@ -1843,9 +1843,7 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
 
         // 4. Use client_encryption to add a keyAltName "def" to the key created in Step 1 and assert the operation fails due to a duplicate key server error (error code 11000).
         try {
-          await clientEncryption.createDataKey('local', {
-            keyAltNames: ['def']
-          });
+          await clientEncryption.addKeyAltName(_id, 'def');
           expect.fail(
             'Error in step 4) expected clientEncryption.createDataKey to throw duplicate key error but it did not'
           );
--- a/test/integration/client-side-encryption/client_side_encryption.prose.test.js
+++ b/test/integration/client-side-encryption/client_side_encryption.prose.test.js
@@ -1775,28 +1775,28 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
         });
 
         // 2. Repeat Step 1 and assert the operation fails due to a duplicate key server error (error code 11000).
-        const maybeErrorStep2 = await clientEncryption
+        const resultStep2 = await clientEncryption
           .createDataKey('local', {
             keyAltNames: ['abc']
           })
           .catch(e => e);
         expect(
-          maybeErrorStep2,
+          resultStep2,
           'Error in step 2) expected clientEncryption.createDataKey to throw duplicate key error but it did not'
         ).to.be.instanceof(MongoServerError);
-        expect(maybeErrorStep2).have.property('code', 11000);
+        expect(resultStep2).have.property('code', 11000);
 
         // 3. Use client_encryption to create a new local data key with a keyAltName "def" and assert the operation fails due to a duplicate key server error (error code 11000).
-        const maybeErrorStep3 = await clientEncryption
+        const resultStep3 = await clientEncryption
           .createDataKey('local', {
             keyAltNames: ['def']
           })
           .catch(e => e);
         expect(
-          maybeErrorStep3,
+          resultStep3,
           'Error in step 3) expected clientEncryption.createDataKey to throw duplicate key error but it did not'
         ).to.be.instanceof(MongoServerError);
-        expect(maybeErrorStep3).have.property('code', 11000);
+        expect(resultStep3).have.property('code', 11000);
       });
     });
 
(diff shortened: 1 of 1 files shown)


test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1723,6 +1723,150 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+        } catch (err) {
+          expect.fail(
+            'Error in step 3) expected clientEncryption.createDataKey not to fail, but received',
+            err
+          );
+        }
+
+        // 4. Use client_encryption to add a keyAltName "def" to the key created in Step 1 and assert the operation fails due to a duplicate key server error (error code 11000).
+        try {
+          await clientEncryption.createDataKey('local', {
```
addaleax, line 1846: The description says to add the keyAltName to the key created earlier, not create an existing key, so I think this should be `.addKeyAltName()`?

baileympearson: Yes, good catch!

addaleax (dismissed):
:shipit:

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1723,6 +1723,148 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+            err
+          );
+        }
+
+        // 4. Use client_encryption to add a keyAltName "def" to the key created in Step 1 and assert the operation fails due to a duplicate key server error (error code 11000).
+        try {
+          await clientEncryption.addKeyAltName(_id, 'def');
+          expect.fail(
+            'Error in step 4) expected clientEncryption.createDataKey to throw duplicate key error but it did not'
+          );
```
nbbeeken, line 1849: Talked offline about doing something with assertions inside try potentially messing with the assertion inside catch.

nbbeeken (changes requested):

nbbeeken (dismissed):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1723,6 +1723,148 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+            err
+          );
+        }
+
+        // 2. Use client_encryption to add a keyAltName "abc" to the key created in Step 1 and assert the operation does not fail.
+        try {
+          await clientEncryption.addKeyAltName(_id, 'abc');
+        } catch (err) {
+          expect.fail(
+            'Error in step 2) expected clientEncryption.createDataKey not to fail, but received',
```
dariakp, line 1828: this needs to be `addAltKeyName` not `createDataKey` in this and all the custom error messages below

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1723,6 +1723,148 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+            err
+          );
+        }
+
+        // 3. Repeat Step 2, assert the operation does not fail, and assert the returned key document contains the keyAltName "abc" added in Step 2.
+        try {
+          const { keyAltNames } = await clientEncryption.addKeyAltName(_id, 'abc');
+          expect(keyAltNames).to.contain('abc');
+        } catch (err) {
+          expect.fail(
```
dariakp, line 1838: I think this transforms the potential error in L1836 in an undesirable way, i.e., instead of getting the expectation failure for keyAltNames, we're instead getting an error that says the method call failed

elsewhere we've adopted a `await method.catch(err => err)` pattern, which could be the solution here and elsewhere in the test code, but for an immediate fix that keeps the pattern you are using the same, you should postpone the destructuring that you do in L1835, and check the property immediately after the try/catch block

dariakp (changes requested):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1723,6 +1723,123 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
     });
   });
 
+  context.only('13. Unique Index on keyAltNames', function () {
```
addaleax, line 1726:
```suggestion
  context('13. Unique Index on keyAltNames', function () {
```

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1723,6 +1723,123 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+          .addKeyAltName(_id, 'abc')
+          .catch(e => e);
+        expect(
+          maybeErrorStep3,
+          `Error in step 3) expected clientEncryption.addKeyAltName not to fail, but received ${maybeErrorStep3}`
+        ).not.to.be.instanceof(MongoServerError);
+        expect(maybeErrorStep3).to.have.property('keyAltNames').to.include('abc');
+
+        // 4. Use client_encryption to add a keyAltName "def" to the key created in Step 1 and assert the operation fails due to a duplicate key server error (error code 11000).
+
```
addaleax, line 1822:
```suggestion
```

addaleax (dismissed):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
baileympearson: 🤦

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1723,6 +1723,120 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+    context('Case 1', metadata, function () {
+      it('createDataKey() handles duplicate key errors on the keyvault collection', async function () {
+        // 1. Use client_encryption to create a new local data key with a keyAltName "abc" and assert the operation does not fail.
+        await clientEncryption.createDataKey('local', {
+          keyAltNames: ['abc']
+        });
+
+        // 2. Repeat Step 1 and assert the operation fails due to a duplicate key server error (error code 11000).
+        const maybeErrorStep2 = await clientEncryption
```
dariakp, line 1778: what do you think of using `resultStep2` instead of `maybeError` here and throughout this set of tests? I am just thinking that naming variables what we expect to receive in the test would make it a bit easier to follow the logic, since it's not that it may or may not be an error and either is acceptable for this test, it's that it *should* be an error result; and then further down, we *expect* the result to *not* be an error.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1723,6 +1723,120 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+    context('Case 2', metadata, function () {
+      it('addKeyAltName() handles duplicate key errors on the keyvault collection', async function () {
+        // 1. Use client_encryption to create a new local data key and assert the operation does not fail.
+        const _id = await clientEncryption.createDataKey('local');
+
+        // 2. Use client_encryption to add a keyAltName "abc" to the key created in Step 1 and assert the operation does not fail.
+        await clientEncryption.addKeyAltName(_id, 'abc');
+
+        // 3. Repeat Step 2, assert the operation does not fail, and assert the returned key document contains the keyAltName "abc" added in Step 2.
+        const maybeErrorStep3 = await clientEncryption.addKeyAltName(_id, 'abc').catch(e => e);
```
dariakp, line 1812: I think in the cases where we expect success (here and in step 5), we don't need the catch and assign, since if it's an error, it'll get thrown and the test will fail anyway (like we have in the _id step above); if we're worried that an error is returned but not thrown from the method, we can still do the expect not error check on the result

dariakp (changes requested):

dariakp (approved):

PR #3337, opened 2022-07-27
fix(NODE-4467): Add back support for `oplogReplay` option as deprecated

sampaiodiego:
### Description
The option has been deprecated for MongoDB >=4.4 but should still be available to MongoDB <=4.2

#### What is changing?
Oplog tailing for MongoDB server <=4.2 can still use this flag in order to have a better performance.

No

The initial query to tail the oplog on huge MongoDB 4.2 servers was taking too long to complete.

### DIFF
--- a/src/operations/find.ts
+++ b/src/operations/find.ts
@@ -63,6 +63,11 @@ export interface FindOptions<TSchema extends Document = Document> extends Comman
   showRecordId?: boolean;
   /** Map of parameter names and values that can be accessed using $$var (requires MongoDB 5.0). */
   let?: Document;
+  /**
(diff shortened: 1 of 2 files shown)


durran (changes requested):
Thanks @sampaiodiego for the PR! Could you add a unit test similar to the other operation constructor tests (example: https://github.com/mongodb/node-mongodb-native/blob/main/test/unit/operations/aggregate.test.js) so have a test to avoid another regression?

sampaiodiego:
sure thing @durran .. I've added a few tests, let me know if that's good enough.. thx

durran (approved):

nbbeeken (approved):

PR #3338, opened 2022-07-28
test(NODE-4483): set crypt shared lib path

durran:
### Description

Runs the csfle tests both against mongocryptd and the shared lib.

#### What is changing?

Sets the CRYPT_SHARED_LIB_PATH env variable which the spec runner sees and appropriately sets the extra options in auto encryption - causing the shared library to be used.

None

NODE-4483


.evergreen/config.in.yml
```diff
@@ -819,7 +819,9 @@ functions:
           rm -f ./prepare_client_encryption.sh
 
           VERSION=${VERSION} DRIVERS_TOOLS=${DRIVERS_TOOLS} \
-            bash ${PROJECT_DIRECTORY}/.evergreen/prepare-crypt-shared-lib.sh
+            . ${PROJECT_DIRECTORY}/.evergreen/prepare-crypt-shared-lib.sh
+
+          echo "setting CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
```
durran, line 824: This echo is here  and the var is definitely set: https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_next_ubuntu1804_custom_dependency_tests_run_custom_csfle_shared_lib_tests_pinned_commit_patch_c70c69edb7b51aeac143aa2eb73aaaf465c41572_62f12690d1fe07159c0476cb_22_08_08_15_06_57/0?type=T#L260

.evergreen/config.in.yml
```diff
@@ -819,7 +819,9 @@ functions:
...
 
           MONGODB_URI="${MONGODB_URI}" CRYPT_SHARED_LIB_PATH="${CRYPT_SHARED_LIB_PATH}" \
```
durran, line 826: It is passed explicitly to the next script.

.evergreen/run-custom-csfle-tests.sh
```diff
@@ -10,6 +10,11 @@ if [ -z ${CSFLE_KMS_PROVIDERS+omitted} ]; then echo "CSFLE_KMS_PROVIDERS is unse
 export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
 export AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
 export CSFLE_KMS_PROVIDERS=${CSFLE_KMS_PROVIDERS}
+echo "csfle CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
```
durran, line 13: It's now gone: https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_next_ubuntu1804_custom_dependency_tests_run_custom_csfle_shared_lib_tests_pinned_commit_patch_c70c69edb7b51aeac143aa2eb73aaaf465c41572_62f12690d1fe07159c0476cb_22_08_08_15_06_57/0?type=T#L270

test/tools/runner/hooks/configuration.js
```diff
@@ -149,7 +149,8 @@ const testConfigBeforeHook = async function () {
     ldap: MONGODB_URI.includes('authMechanism=PLAIN'),
     ocsp: process.env.OCSP_TLS_SHOULD_SUCCEED != null && process.env.CA_FILE != null,
     socks5: MONGODB_URI.includes('proxyHost='),
-    compressor: process.env.COMPRESSOR
+    compressor: process.env.COMPRESSOR,
+    cryptSharedLibPath: process.env.CRYPT_SHARED_LIB_PATH
```
durran, line 153: Set to empty string: https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_next_ubuntu1804_custom_dependency_tests_run_custom_csfle_shared_lib_tests_pinned_commit_patch_c70c69edb7b51aeac143aa2eb73aaaf465c41572_62f13ed0562343194bc2de30_22_08_08_16_50_25/0?type=T#L801

.evergreen/config.in.yml
addaleax: @durran I think this is because evergreen interprets `${CRYPT_SHARED_LIB_PATH}` as an evergreen expansion here, before bash even sees it. Using `$CRYPT_SHARED_LIB_PATH` without curly braces might work?

addaleax: Actually, even just dropping the variable from this line here should be a good fix, since `prepare-crypt-shared-lib.sh` exports it into the environment already

addaleax (approved):

.evergreen/config.in.yml
```diff
@@ -810,19 +810,25 @@ functions:
       params:
```
baileympearson, line 810: since you didn't touch these lines it's non-blocking but should we update this function to use `subprocess.exec` instead of `shell.exec`?

baileympearson:
@durran 

Can we update the testing readme with instructions on how to setup and run our test suite with the shared library?

---

The changes in this PR only enable the shared library for legacy csfle tests.  Should we also enable the shared library for the csfle prose tests / manual integration tests and the new unified key management tests (and any future unified csfle tests)?

durran:
> @durran
> 
> Can we update the testing readme with instructions on how to setup and run our test suite with the shared library?
> 
> The changes in this PR only enable the shared library for legacy csfle tests. Should we also enable the shared library for the csfle prose tests / manual integration tests and the new unified key management tests (and any future unified csfle tests)?

Was in the last PR. It's just above https://github.com/mongodb/node-mongodb-native/blob/main/test/readme.md#kmip-fle-support-tests

addaleax (approved):

durran:
@baileympearson I've added the shared lib to the unified runner now as well when present, however I think the related drivers ticket (DRIVERS-2355) and it's associated commit (https://github.com/mongodb/specifications/commit/9a0a385dc897319ecc10dd3dd19bc9d8e4c10de7) are mentioning to run against the shared library in the spec tests section before prose tests are talked about, and none of the prose tests themselves were updated. So I think I'm more inclined not to add them there as then the options we pass won't match the specs verbatim any more.

.evergreen/config.in.yml
durran: I may try to do that in the next PR since since this is working now don't want to tinker with it. :)

baileympearson (approved):
LGTM, since [DRIVERS-2355](https://jira.mongodb.org/browse/DRIVERS-2355) will include updates to our prose tests to run on the shared library where applicable.

dariakp (approved):

PR #3339, opened 2022-07-28
fix(NODE-4273): pass 'comment' option through to distinct command

baileympearson:
#### What is changing?

The `comment` option is now attached to the `distinct` command.

No.


src/operations/distinct.ts
```diff
@@ -61,6 +60,12 @@ export class DistinctOperation extends CommandOperation<any[]> {
       cmd.maxTimeMS = options.maxTimeMS;
     }
 
+    // we check for undefined specifically here to allow falsy values
+    // eslint-disable-next-line no-restricted-syntax
+    if (typeof options.comment !== 'undefined') {
+      cmd.comment = options.comment;
+    }
```
nbbeeken, line 67: Do we wanna add falsey coverage?

nbbeeken (dismissed):

nbbeeken:
Changes LGTM, CI should be fixed with a rebase we think.

nbbeeken (dismissed):

test/integration/node-specific/comment_with_falsy_values.test.ts
```diff
@@ -1,8 +1,8 @@
 import { Long } from '../../../src';
 import { TestBuilder, UnifiedTestSuiteBuilder } from '../../tools/utils';
 
-const falsyValues = [0, false, '', Long.ZERO, null, NaN] as const;
-const falsyToString = (value: typeof falsyValues[number]) => {
+export const falsyValues = [0, false, '', Long.ZERO, null, NaN] as const;
```
dariakp, line 4: instead of exporting from a test file, should we just add this to our test utils? it's not a good idea to entangle the test files like this, but also, see my comment on the other file regarding whether we definitely need that to be seprate

test/integration/node-specific/distinct.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    collection = await client.db('comment-falsy-values').createCollection('collection');
+    commands = [];
+  });
+
+  afterEach(async function () {
+    await collection.drop();
+    await client.close();
+  });
+
+  context('comment with falsy values', function () {
```
dariakp, line 24: why can't these be added to the list of other falsy value comment tests?

dariakp (changes requested):

test/integration/node-specific/distinct.test.ts
baileympearson: Unlike all the other commands tested in the other file, the structure of the `distinct` command is different (the api is `distinct(string, filter?, options?)` while the others are all `<method>(options?)`).  I chose to write a standalone test file instead of modifying the existing test runner because it seemed like the more straightforward approach.  I am willing to reconsider this if you feel strongly though

test/integration/node-specific/comment_with_falsy_values.test.ts
baileympearson: Sure, I'll move this to the utils if we decide to leave the standalone distinct test file (see my response to your comment)

test/integration/node-specific/distinct.test.ts
dariakp: For simplicity, maybe we can just move the test to the same test file as the others just outside the loop? That way they are grouped together, but we can decide later to combine them if there are more use cases

test/integration/node-specific/comment_with_falsy_values.test.ts
```diff
@@ -173,4 +175,39 @@ describe('Comment with falsy values', () => {
...
+        await collection.distinct('some-key', {}, { comment: falsyValue }).catch(() => null);
+
+        expect(commands).to.have.lengthOf(1);
+        const distinctCommand = commands.find(command => command.commandName === 'distinct');
+        expect(distinctCommand).to.exist;
+
+        // chai does not narrow types, so TS doesn't know the distinct command exists at this point.
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        expect(distinctCommand!.command).to.haveOwnProperty('comment');
+      });
```
dariakp, line 210: Should we also check that the value is set to the passed in value like we do with the other tests above?

dariakp (changes requested):

test/integration/node-specific/comment_with_falsy_values.test.ts
baileympearson: Yeah, I can add a check for the value as well

dariakp (approved):

nbbeeken (approved):

PR #3340, opened 2022-07-28
feat(NODE-4336): deprecate old write concern options and add missing writeConcern to MongoClientOptions

nbbeeken:
#### What is changing?

- Added writeConcern to MongoClientOptions
- Added deprecation flags to the top level versions of the options
- Added a type test

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -1188,7 +1188,7 @@ export const OPTIONS = {
 
       throw new MongoParseError(`Invalid WriteConcern cannot parse: ${JSON.stringify(value)}`);
     }
-  } as OptionDescriptor,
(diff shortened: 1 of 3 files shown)


dariakp (commented):
The changes LGTM, but I think we should use feat, because we are updating the deprecation markers in the docs; I think the title should be something like `feat(NODE-4336): deprecate old write concern options and add missing writeConcern to MongoClientOptions`

nbbeeken:
Alright yea, fair enough better to call it out then it perhaps get overlooked in a sub bullet

dariakp (approved):
Approving pending CI

durran (approved):

PR #3342, opened 2022-08-01
fix(NODE-4496): counter values incorrecly compared when instance of Long

nbbeeken:
#### What is changing?

- Correct the issue with the long check and comparison using the wrong value.
- Clarify variable ordering with better naming
- Bring in spec docs for comparison
- Mark other places of Long BSON related issues
- Add unit tests passing both numbers and Longs into comparison function

Fix comparison bug


src/sdam/server_description.ts
```diff
@@ -105,6 +105,7 @@ export class ServerDescription {
     if (options?.topologyVersion) {
       this.topologyVersion = options.topologyVersion;
     } else if (hello?.topologyVersion) {
+      // TODO(NODE-2674): Preserve int64 sent from MongoDB
```
dariakp, line 108: while you're in this file, could you also fix the null access issue in `equals`? we access `other.topologyVersion` in L185, but in L194 we check that `other != null`

dariakp (changes requested):

dariakp (changes requested):
need to not have lint/unit test failures here

src/sdam/server_description.ts
```diff
@@ -179,15 +180,15 @@ export class ServerDescription {
    * Determines if another `ServerDescription` is equal to this one per the rules defined
    * in the {@link https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#serverdescription|SDAM spec}
    */
-  equals(other: ServerDescription): boolean {
+  equals(other?: ServerDescription | null): boolean {
     const topologyVersionsEqual =
-      this.topologyVersion === other.topologyVersion ||
-      compareTopologyVersion(this.topologyVersion, other.topologyVersion) === 0;
+      this.topologyVersion === other?.topologyVersion ||
```
dariakp, line 185: this comparison feels like it violates this portion in the `compareTopologyVersion` logic:

> If the response topologyVersion is unset or the ServerDescription's topologyVersion is null, the client MUST assume the response is more recent.

why do we check strict equality here instead of just using the outcome of `compareTopologyVersion` from the outset? are we explicitly trying to catch `null === null`?

src/sdam/server_description.ts
```diff
@@ -179,15 +180,15 @@ export class ServerDescription {
...
+      compareTopologyVersion(this.topologyVersion, other?.topologyVersion) === 0;
 
-    const electionIdsEqual: boolean =
-      this.electionId && other.electionId
-        ? other.electionId && this.electionId.equals(other.electionId)
-        : this.electionId === other.electionId;
+    const electionIdsEqual =
+      this.electionId && other?.electionId != null
+        ? other.electionId && compareObjectId(this.electionId, other.electionId) === 0
```
dariakp, line 190: the check in L189 implies that `other.electionId != null`, so the first condition of this `&&` is redundant

test/unit/sdam/server_description.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    const processIdZero = new ObjectId('00'.repeat(12));
+    const processIdOne = new ObjectId('00'.repeat(11) + '01');
+    const compareTopologyVersionTests: {
+      title: string;
+      currentTv?: TopologyVersion;
+      newTv?: TopologyVersion;
+      out: 0 | -1 | 1;
+    }[] = [
+      {
+        title: 'should return that currentTv and newTv are equal',
```
dariakp, line 68: can you fix the descriptions for all of these to actually state what condition is being tested?
e.g., in this case, `should return that the versions are equal  when process ids are equal and both counter values are Long.ZERO`, in the next case `should return that the versions are equal  when process ids are equal and both counter values are non-zero numbers` (and by the way, we should add plain number `0` as a test case)
I think this could also benefit  from splitting up the cases into several context blocks `when process ids are equal`, and `when process ids are not equal`

dariakp (changes requested):

src/sdam/server_description.ts
nbbeeken: Thanks for calling this out, when I made the change to respect that part of the spec we get test failures so something is amiss here. I think that the equality case and the comparison test have different rules about nullishness. I'll file a ticket.

NODE-4510

src/sdam/server_description.ts
nbbeeken: This is part of the same bug mentioned above I'll note it in the ticket. I'll clean this up still though

test/unit/sdam/server_description.test.ts
nbbeeken: Organized the tests, and improved names :)

test/unit/sdam/server_description.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    const processIdZero = new ObjectId('00'.repeat(12));
+    const processIdOne = new ObjectId('00'.repeat(11) + '01');
+    type CompareTopologyVersionTest = {
+      title: string;
+      currentTv?: TopologyVersion;
+      newTv?: TopologyVersion;
+      out: 0 | -1 | 1;
+    };
+
+    const compareTopologyVersionEqualTests: CompareTopologyVersionTest[] = [
```
dariakp, line 68: Would it be worth also having non-zero Longs in the counter, for completeness?

test/unit/sdam/server_description.test.ts
```diff
@@ -0,0 +1,156 @@
...
+          'when process ids are equal and counter values are equal but new has a different type',
+        // @ts-expect-error: Testing that the function handles numbers
+        currentTv: { processId: processIdZero, counter: 2 },
+        newTv: { processId: processIdZero, counter: Long.fromNumber(2) },
+        out: 0
+      }
+    ];
+    const compareTopologyVersionLessThanTests: CompareTopologyVersionTest[] = [
+      {
+        title: 'when both versions are nullish',
```
dariakp, line 110: would it be worth having two more tests that have different kinds of nullish for completeness? (null with undefined, and undefined with null)

dariakp (changes requested):

test/unit/sdam/server_description.test.ts
nbbeeken: Yep, added!

test/unit/sdam/server_description.test.ts
nbbeeken: Sure!

durran (approved):

PR #3344, opened 2022-08-02
test(NODE-4490): update expected error in fle2-InsertFind-Unindexed 

baileympearson:
#### What is changing?

Syncs a test fix for a new error message to fix a failing csfle test.

No.

Tests aren't supposed to fail.

### DIFF
--- a/test/spec/client-side-encryption/tests/legacy/fle2-InsertFind-Unindexed.yml
+++ b/test/spec/client-side-encryption/tests/legacy/fle2-InsertFind-Unindexed.yml
@@ -80,4 +80,4 @@ tests:
         arguments:
           filter: { encryptedUnindexed: "value123" }
         result:
-          errorContains: "Cannot query"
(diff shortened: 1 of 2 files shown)


test/spec/client-side-encryption/tests/legacy/fle2-InsertFind-Unindexed.yml
```diff
@@ -80,4 +80,4 @@ tests:
         arguments:
           filter: { encryptedUnindexed: "value123" }
         result:
-          errorContains: "Cannot query"
+          errorContains: "encrypt"
```
dariakp, line 83: the yml file has comments for this, we should pull them in just to keep the line numbers in sync

dariakp (changes requested):

nbbeeken (approved):

dariakp (approved):

PR #3345, opened 2022-08-02
test(NODE-4489): add CSFLE prose test 16

baileympearson:
#### What is changing?

Implements csfle prose test 16 - https://jira.mongodb.org/browse/NODE-4489.  This is a regression test for rewrapManyDataKey.

No.

### DIFF
--- a/test/integration/client-side-encryption/client_side_encryption.prose.test.js
+++ b/test/integration/client-side-encryption/client_side_encryption.prose.test.js
@@ -2081,9 +2081,9 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
       }
     }
 
-    beforeEach(async function () {
-      client1 = await this.configuration.newClient().connect();
-      client2 = await this.configuration.newClient().connect();
+    beforeEach(function () {
+      client1 = this.configuration.newClient();
+      client2 = this.configuration.newClient();
     });
 
     afterEach(async function () {
@@ -2113,7 +2113,8 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
                   tlsCAFile: process.env.KMIP_TLS_CA_FILE,
                   tlsCertificateKeyFile: process.env.KMIP_TLS_CERT_FILE
                 }
-              }
+              },
+              bson: BSON
             }
           );
 
@@ -2139,7 +2140,8 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
                   tlsCAFile: process.env.KMIP_TLS_CA_FILE,
                   tlsCertificateKeyFile: process.env.KMIP_TLS_CERT_FILE
                 }
-              }
+              },
+              bson: BSON
             }
           );
 
--- a/test/integration/client-side-encryption/client_side_encryption.prose.test.js
+++ b/test/integration/client-side-encryption/client_side_encryption.prose.test.js
@@ -2109,7 +2109,13 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
             client1,
             {
               keyVaultNamespace: 'keyvault.datakeys',
-              kmsProviders: getKmsProviders()
+              kmsProviders: getKmsProviders(),
+              tlsOptions: {
+                kmip: {
+                  tlsCAFile: process.env.KMIP_TLS_CA_FILE,
+                  tlsCertificateKeyFile: process.env.KMIP_TLS_CERT_FILE
+                }
+              }
             }
           );
 
@@ -2129,7 +2135,13 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
             client1,
             {
               keyVaultNamespace: 'keyvault.datakeys',
-              kmsProviders: getKmsProviders()
+              kmsProviders: getKmsProviders(),
+              tlsOptions: {
+                kmip: {
+                  tlsCAFile: process.env.KMIP_TLS_CA_FILE,
+                  tlsCertificateKeyFile: process.env.KMIP_TLS_CERT_FILE
+                }
+              }
             }
           );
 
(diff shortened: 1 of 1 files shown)


test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -2046,4 +2046,124 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+      client1 = await this.configuration.newClient().connect();
+      client2 = await this.configuration.newClient().connect();
+    });
+
+    afterEach(async function () {
+      await client1.close();
+      await client2.close();
+    });
+
+    for (const { srcProvider, dstProvider } of generateTestCombinations()) {
```
addaleax, line 2094: This is a great pattern that I’ll absolutely keep in mind – it’s much nicer than nesting for loops around the test implementation :slightly_smiling_face:

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -2046,4 +2046,124 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+
+          // Step 4. Call ``clientEncryption1.encrypt`` with the value "test"
+          const cipherText = await clientEncryption1.encrypt('test', {
+            keyId,
+            algorithm: 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic'
+          });
+
+          // Step 5. Create a ``ClientEncryption`` object named ``clientEncryption2``
+          const clientEncryption2 = new this.configuration.mongodbClientEncryption.ClientEncryption(
+            client1,
```
addaleax, line 2133: I think this should technically be

```suggestion
            client2,
```

That being said … I’m not sure how much this really matters. In terms of adhering to the spec test, strictly speaking the test would need to create a new pair of clients for *every combination* of providers. If that’s something we don’t want to do, because it’s slow or really for whatever reason, then that reason probably can be extended to just using a single client for all ClientEncryption objects as well.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -2046,4 +2046,124 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+
+          expect(rewrapManyDataKeyResult).to.have.property('bulkWriteResult');
+          expect(rewrapManyDataKeyResult.bulkWriteResult).to.have.property('nModified', 1);
+
+          // 7. Call ``clientEncryption1.decrypt`` with the ``ciphertext``. Assert the return value is "test".
+          const decryptResult1 = await clientEncryption1.decrypt(cipherText);
+          expect(decryptResult1).to.equal('test');
+
+          // 8. Call ``clientEncryption2.decrypt`` with the ``ciphertext``. Assert the return value is "test".
+          const decryptResult2 = await clientEncryption1.decrypt(cipherText);
```
addaleax, line 2163:
```suggestion
          const decryptResult2 = await clientEncryption2.decrypt(cipherText);
```

test/integration/client-side-encryption/client_side_encryption.prose.test.js
baileympearson: Good catch! 

We actually do already create two clients for each pair of providers - it's done in the beforeEach hook for the test.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
baileympearson: 🙂

test/integration/client-side-encryption/client_side_encryption.prose.test.js
baileympearson: Great catch!

test/integration/client-side-encryption/client_side_encryption.prose.test.js
baileympearson: Generally in our tests we create a new client for each test.  I don't know if that slows down our tests substantially but I'd guess that it doesn't matter much.  My guess is that since our tests use a real server, the bottleneck is really network i/o with the server and the server writing to the disk, not our actual test code.  It's worth thinking about though if our tests start taking a long time to run.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -2046,4 +2046,124 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+        for (const dstProvider of providers) {
+          yield { srcProvider, dstProvider };
+        }
+      }
+    }
+
+    beforeEach(async function () {
+      client1 = await this.configuration.newClient().connect();
+      client2 = await this.configuration.newClient().connect();
+    });
```
nbbeeken, line 2087: No need to connect anymore woohoo!

baileympearson: If this is a blocking comment I'll fix it but otherwise I'll leave it as-is

nbbeeken: Yea please remove it TIA

baileympearson: mongodb-client-encryption doesn't pin a version of the driver - I can't think of why this may be a problem but do you see any concern with testing a feature only supported in 4.7.0+ here?

baileympearson: Do we ever test mongodb-client-encryption with versions of the driver other than latest actually?  I don't think so.  

I'm not sure which features of the driver we use in libmongocrypt when using the client but if it's anything non-standard (not stable API) could it be valuable to add multiple driver versions to our libmongocrypt CI?

nbbeeken (dismissed):

addaleax (approved):

durran (approved):

nbbeeken (approved):

PR #3347, opened 2022-08-05
test(NODE-4323): add a unit test task with bson-ext installed

nbbeeken:
#### What is changing?

bson-ext now has a unit test task

Adds more utf8 validation behavior coverage

### DIFF
--- a/.evergreen/config.in.yml
+++ b/.evergreen/config.in.yml
@@ -837,15 +837,19 @@ functions:
           - "${PROJECT_DIRECTORY}/.evergreen/run-snappy-version-test.sh"
 
   "run bson-ext test":
-    - command: shell.exec
+    - command: subprocess.exec
       type: test
       params:
         working_dir: "src"
         timeout_secs: 60
-        script: |
-          ${PREPARE_SHELL}
+        env:
+          PROJECT_DIRECTORY: ${PROJECT_DIRECTORY}
+          MONGODB_URI: ${MONGODB_URI}
+          TEST_NPM_SCRIPT: ${TEST_NPM_SCRIPT}
+        binary: bash
+        args:
+          - '${PROJECT_DIRECTORY}/.evergreen/run-bson-ext-test.sh'
 
-          MONGODB_URI="${MONGODB_URI}" bash ${PROJECT_DIRECTORY}/.evergreen/run-bson-ext-test.sh
 
   "upload test results":
     # Upload the xunit-format test results.
--- a/.evergreen/generate_evergreen_tasks.js
+++ b/.evergreen/generate_evergreen_tasks.js
@@ -507,15 +507,26 @@ BUILD_VARIANTS.push({
 const oneOffFuncs = [
   { func: 'run custom snappy tests' },
   {
+    name: 'run-bson-ext-integration',
     func: 'run bson-ext test',
     vars: {
-      NODE_LTS_NAME: LOWEST_LTS
+      NODE_LTS_NAME: LOWEST_LTS,
+      TEST_NPM_SCRIPT: 'check:test'
+    }
+  },
+  {
+    name: 'run-bson-ext-unit',
+    func: 'run bson-ext test',
+    vars: {
+      NODE_LTS_NAME: LOWEST_LTS,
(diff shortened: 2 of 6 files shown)


.evergreen/config.in.yml
```diff
@@ -837,15 +837,19 @@ functions:
           - "${PROJECT_DIRECTORY}/.evergreen/run-snappy-version-test.sh"
 
   "run bson-ext test":
-    - command: shell.exec
+    - command: subprocess.exec
```
durran, line 840: Thanks for changing this.

durran (dismissed):

test/unit/commands.test.ts
```diff
@@ -45,9 +51,12 @@ describe('BinMsg BSON utf8 validation', () => {
     // this is a sanity check to make sure nothing unexpected is happening in the deserialize method itself
 
     const options = { validation: { utf8: { writeErrors: false } as const } };
-    expect(BSON.deserialize(invalidUtf8ErrorMsgDeserializeInput, options)).to.deep.equals(
-      invalidUtf8InWriteErrorsJSON
-    );
+    const deserializerCall = () => BSON.deserialize(invalidUtf8ErrorMsgDeserializeInput, options);
```
baileympearson, line 54: This comment applies to all tests in this file.

I think we should break apart the each test into two separate tests - one when bson ext is installed and one when it is not.  Right now, in the case that bson ext is installed, we're not actually testing what the test description says.  A test structure like the following would be easier to understand and would accurately describe the behavior of the test

```typescript
context('when bson ext is installed', function() {
  it('....', {})
})

context('when bson ext is not installed', function() {
  it('bson correctly deserializes data with replacement characters for invalid utf8 in writeErrors object', () => {
    ... 
  });
})

```

baileympearson: This might require a larger restructure of the file but I think the better test organization is worth it

.evergreen/generate_evergreen_tasks.js
```diff
@@ -507,15 +507,27 @@ BUILD_VARIANTS.push({
...
-      NODE_LTS_NAME: LOWEST_LTS
+      NODE_LTS_NAME: LOWEST_LTS,
+      TEST_NPM_SCRIPT: 'check:unit'
     }
   }
 ];
 
 const oneOffFuncAsTasks = oneOffFuncs.map(oneOffFunc => ({
-  name: `${oneOffFunc.func.split(' ').join('-')}`,
+  name: `${
```
baileympearson, line 528: rather than patching the logic here, what if we just gave the other instances of oneOffFuncs a `name` property?  Having a "name" makes more sense to me anyways and is one less thing to maintain.

baileympearson (changes requested):

test/unit/commands.test.ts
nbbeeken: Made two blocks, skipped stuff, and renamed tests

.evergreen/generate_evergreen_tasks.js
nbbeeken: sure, done

test/unit/commands.test.ts
```diff
@@ -41,104 +42,229 @@ const nKeyWithInvalidUtf8 =
 const msgBodyNKeyWithInvalidUtf8 = Buffer.from(nKeyWithInvalidUtf8, 'hex');
 
 describe('BinMsg BSON utf8 validation', () => {
-  it('bson correctly deserializes data with replacement characters for invalid utf8 in writeErrors object', () => {
-    // this is a sanity check to make sure nothing unexpected is happening in the deserialize method itself
+  describe('when bson-ext is installed', () => {
+    before(function () {
+      if (!isBSONExtImported()) {
+        this.skip();
```
baileympearson, line 48: Can we attach a skip reason?  This comment also applies to the js-bson section of the tests.

baileympearson (approved):

test/unit/commands.test.ts
nbbeeken: sure!

durran (approved):

baileympearson (approved):

dariakp (approved):

PR #3348, opened 2022-08-05
fix(NODE-4159,NODE-4512): remove servers with incorrect setName from topology and fix unix socket parsing

nbbeeken:
#### What is changing?

- Remove servers
- Null out fields on server description instead of omitted keys
- Fixes unix socket issue: https://jira.mongodb.org/browse/NODE-4512

Aligns with spec test, easier assertion against unit specs

### DIFF
--- a/src/sdam/server_description.ts
+++ b/src/sdam/server_description.ts
@@ -1,5 +1,5 @@
 import { Document, Long, ObjectId } from '../bson';
-import { MongoInvalidArgumentError, MongoRuntimeError, MongoServerError } from '../error';
+import { MongoRuntimeError, MongoServerError } from '../error';
 import { arrayStrictEqual, compareObjectId, errorStrictEqual, HostAddress, now } from '../utils';
 import type { ClusterTime } from './common';
 import { ServerType } from './common';
--- a/src/sdam/server_description.ts
+++ b/src/sdam/server_description.ts
@@ -1,5 +1,5 @@
 import { Document, Long, ObjectId } from '../bson';
-import type { MongoServerError } from '../error';
+import { MongoInvalidArgumentError, MongoServerError } from '../error';
 import { arrayStrictEqual, compareObjectId, errorStrictEqual, HostAddress, now } from '../utils';
 import type { ClusterTime } from './common';
 import { ServerType } from './common';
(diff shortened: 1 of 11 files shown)


dariakp (changes requested):

test/unit/operations/find.test.ts
```diff
@@ -40,7 +40,11 @@ describe('FindOperation', function () {
   describe('#execute', function () {
     context('command construction', () => {
       const namespace = ns('db.collection');
-      const server = new Server(new Topology([], {} as any), new ServerDescription(''), {} as any);
+      const server = new Server(
+        new Topology([], {} as any),
+        new ServerDescription('a:1'),
```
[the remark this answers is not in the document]
dariakp: that was kind of my thought, if empty string results in a malformed address, it would be better to guard against it with an actual error, otherwise it's not actually clear that we need to pass in something more legitimate looking and why it might be necessary, so it could trip up future developers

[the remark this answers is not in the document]
nbbeeken: Yea, I agree, added an if stmt, I blocked both nullish and empty string inputs, any remaining validation is up to HostAddress

src/sdam/server_description.ts
```diff
@@ -82,6 +82,12 @@ export class ServerDescription {
     hello?: Document,
     options: ServerDescriptionOptions = {}
   ) {
+    if (address == null || address === '') {
+      throw new MongoInvalidArgumentError(
```
dariakp, line 86: should this be a `MongoRuntimeError` instead if it's not meant to be directly instantiated by users? I'm actually confused why the class is marked `public` but the comment above it says "internal type" and then this constructor is marked `@internal`... what are the rules?

nbbeeken: The class's **type** is public because it can be encountered through SDAM event emitters, but it is not meant to be constructed by anyone else so we mark the constructor internal (similar patterns are in our events, Db, and Collection classes). So you're correct this should be a runtime error, our driver would have to accidentally have missed some critical validation to encounter this.

```ts
var mdb = require('mongodb')
mdb.ServerDescription; // undefined
```

dariakp (approved):

src/sdam/topology_description.ts
```diff
@@ -64,22 +64,10 @@ export class TopologyDescription {
...
-    if (maxElectionId) {
-      this.maxElectionId = maxElectionId;
-    }
-
-    if (commonWireVersion) {
-      this.commonWireVersion = commonWireVersion;
-    }
+    this.setName = setName ?? null;
+    this.maxElectionId = maxElectionId ?? null;
+    this.maxSetVersion = maxSetVersion ?? null;
```
durran, line 69: Defaulted in the constructor so could be simplified.

```suggestion
    this.setName = setName;
    this.maxElectionId = maxElectionId;
    this.maxSetVersion = maxSetVersion;
```

durran (changes requested):

durran (approved):

baileympearson (approved):

PR #3354, opened 2022-08-11
test(NODE-4422): use shared lib in prose tests

durran:
### Description

Uses the shared lib in the csfle prose tests when the environment supports it.

#### What is changing?

Modifies all csfle prose, corpus, and driver tests to set `extraOptions.sharedCryptLibPath` when CRYPT_SHARED_LIB_PATH is in the environment. Updates the evg config to test csfle against 5.0, rapid, and latest, where CRYPT_SHARED_LIB_PATH will get set currently only with latest and 5.0 and rapid will run against mongocryptd.

None

### DIFF
--- a/.evergreen/config.in.yml
+++ b/.evergreen/config.in.yml
@@ -767,37 +767,6 @@ functions:
         script: |
           ${PREPARE_SHELL}
           cat <<EOT > prepare_client_encryption.sh
-          export CLIENT_ENCRYPTION=${CLIENT_ENCRYPTION}
-          export CSFLE_KMS_PROVIDERS='${CSFLE_KMS_PROVIDERS}'
-          export AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}"
-          export AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}"
-          export CSFLE_GIT_REF="${CSFLE_GIT_REF}"
-          export CDRIVER_GIT_REF="${CDRIVER_GIT_REF}"
-          EOT
-    - command: shell.exec
-      type: test
-      params:
-        working_dir: "src"
-        timeout_secs: 60
-        script: |
-          ${PREPARE_SHELL}
-
-          # Disable xtrace (just in case it was accidentally set).
-          set +x
-          . ./prepare_client_encryption.sh
-          rm -f ./prepare_client_encryption.sh
-
-          MONGODB_URI="${MONGODB_URI}" bash ${PROJECT_DIRECTORY}/.evergreen/run-custom-csfle-tests.sh
-
-  "run custom csfle shared lib tests":
-    - command: shell.exec
-      type: test
-      params:
-        silent: true
-        working_dir: "src"
-        script: |
-          ${PREPARE_SHELL}
-          cat <<EOT > prepare_client_encryption.sh
           export CLIENT_ENCRYPTION='${CLIENT_ENCRYPTION}'
           export CSFLE_KMS_PROVIDERS='${CSFLE_KMS_PROVIDERS}'
           export AWS_ACCESS_KEY_ID='${AWS_ACCESS_KEY_ID}'
@@ -824,9 +793,8 @@ functions:
 
           source ${PROJECT_DIRECTORY}/.evergreen/prepare-crypt-shared-lib.sh
           export MONGODB_URI="${MONGODB_URI}"
-          export CRYPT_SHARED_LIB_PATH
 
-          echo "setting CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
+          echo "CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
 
           bash ${PROJECT_DIRECTORY}/.evergreen/run-custom-csfle-tests.sh
 
--- a/.evergreen/config.in.yml
+++ b/.evergreen/config.in.yml
@@ -154,6 +154,7 @@ functions:
       params:
         working_dir: "src"
         timeout_secs: 300
+        shell: bash
         script: |
           ${PREPARE_SHELL}
 
@@ -163,6 +164,13 @@ functions:
             . ./prepare_client_encryption.sh
             rm -f ./prepare_client_encryption.sh
           fi
+
+          export VERSION=${VERSION}
+          export DRIVERS_TOOLS=${DRIVERS_TOOLS}
+
+          source ${PROJECT_DIRECTORY}/.evergreen/prepare-crypt-shared-lib.sh
+
+          echo "CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
 
           MONGODB_URI="${MONGODB_URI}" \
           AUTH=${AUTH} SSL=${SSL} TEST_CSFLE=true \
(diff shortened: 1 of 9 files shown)


.evergreen/generate_evergreen_tasks.js
```diff
@@ -549,114 +549,35 @@ const oneOffFuncAsTasks = oneOffFuncs.map(oneOffFunc => ({
...
-    { func: 'bootstrap kms servers' },
-    {
-      func: 'run custom csfle shared lib tests',
-      vars: {
-        VERSION: 'latest',
-        CSFLE_GIT_REF: 'master'
-      }
-    }
-  ]
+[ '5.0', 'rapid', 'latest' ].forEach(version => {
```
durran, line 552: These 3 cover the 5.x/6.x cases plus whatever current rapid is.

test/integration/client-side-encryption/client_side_encryption.prose.corpus.test.js
```diff
@@ -214,10 +215,12 @@ describe('Client Side Encryption Prose Corpus Test', function () {
               tlsCertificateKeyFile: process.env.KMIP_TLS_CERT_FILE
             }
           };
+          const extraOptions = getEncryptExtraOptions();
           const autoEncryption = {
             keyVaultNamespace,
             kmsProviders,
-            tlsOptions
+            tlsOptions,
+            extraOptions
```
durran, line 223: All these tests now include `extraOptions.sharedCryptLibPath` when `CRYPT_SHARED_LIB_PATH` is in the env.

test/unit/operations/get_more.test.ts
```diff
@@ -198,7 +198,11 @@ describe('GetMoreOperation', function () {
     });
 
     context('error cases', () => {
-      const server = new Server(new Topology([], {} as any), new ServerDescription(''), {} as any);
+      const server = new Server(
```
durran, line 201: Unrelated to this ticket but this test was failing on main.

.evergreen/config.in.yml
```diff
@@ -824,9 +793,8 @@ functions:
 
           source ${PROJECT_DIRECTORY}/.evergreen/prepare-crypt-shared-lib.sh
           export MONGODB_URI="${MONGODB_URI}"
-          export CRYPT_SHARED_LIB_PATH
 
-          echo "setting CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
+          echo "CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
```
durran, line 797: It's not actually setting it here, just echoing the value.

.evergreen/prepare-crypt-shared-lib.sh
```diff
@@ -12,7 +12,6 @@ get_mongodb_download_url_for "$DISTRO" "$MONGODB_VERSION"
 # get_mongodb_download_url_for defines $MONGO_CRYPT_SHARED_DOWNLOAD_URL and $EXTRACT.
 if [ -z "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" ]; then
   echo "There is no crypt_shared library for distro='$DISTRO' and version='$MONGODB_VERSION'".
-  exit 1
```
durran, line 15: Not exiting here allows us to run the same script in evergreen for all csfle tests, so it continues running when server is less than 6.0 and the env var doesn't get set.

jimmywarting:
why don't you have issues enabled?

anyhow, wanted to report an issue with async iterators (cursors) being quite slow...

i found a way to make it faster using `cursor.forEach` and still `yield` all the values
i think the problem is b/c it's using a bunch of nextTick or promises or what not...

This is the slow version:
```js
const cursor = await table.aggregate(...)
for await (const doc of cursor) {
  // ... slow
}
// But i found that forEach was much faster...
table.aggregate(...).forEach()
```

logging the code of `cursor[Symbol.asyncIterator].toString()` showed me that it look something like this:
```js
function asyncIterator() {
  const cursor = this;

  return {
    next: function() {
      return Promise.resolve()
        .then(() => cursor.next())
        .then(value => {
          if (!value) {
            return cursor.close().then(() => ({ value, done: true }));
          }
          return { value, done: false };
        });
    }
  };
}
```
I bet all the slow part/bits are all the asynchronous promises that `.then()` makes that makes it slower? i don't know...

so i found a way to make it faster:
```js
// Turn a object that only has a forEach method into a async generator
async function* fasterAsyncGenerator(cursor) {
    const result = [];
    let done = false;
    let resolve = () => {};
    // When data arrives in chunks, then forEach will be called synchronous 
    // {batchSize} times and then wait for new chunks of data to arrive.
    cursor.forEach(doc => {
        result.push(doc);
        resolve();
    }, () => (done = true));

    while (!done || result.length) {
        // this while loop will be called (total of docs) / chunkSize
        await new Promise(rs => resolve = rs);
        yield * result.splice(0); // the length of result is chunkSize most often
    }
}
```
and using the code above as such:
```js
const cursor = await table.aggregate(...)
for await (const doc of fasterAsyncGenerator(cursor)) {
  // 
}
```
made the code faster. it went from 1.168s down to 680.719ms. (currently using mongodb: v3.6.3)

baileympearson:
Hi @jimmywarting - we use Jira to track bug issues and feature requests, not Github issues.  We have instructions for filing a bug report on our readme [here](https://github.com/mongodb/node-mongodb-native#bugs--feature-requests).  Going through our Jira ensures that we will see it during our weekly triage meetings.  Would you mind filing a bug report on Jira reporting this issue?  We would greatly appreciate it if you didn't comment on unrelated PRs with bug reports 🙂.  

This isn't a full answer, but your solution builds an array containing all results of cursor into memory.  One benefit of the async iterator approach or the forEach approach is that no more memory is consumed than is necessary.  Have you tried using `toArray()` instead?  That may give you the performance you're looking for:

```typescript
const resultArray = await table.aggregate(...).toArray()
for (const doc of resultArray) {
  // 
}
```

.evergreen/generate_evergreen_tasks.js
```diff
@@ -549,114 +549,35 @@ const oneOffFuncAsTasks = oneOffFuncs.map(oneOffFunc => ({
...
+  [ '41afd44ca04d246998969c53de4e0f22802b0c8a', 'master' ].forEach(ref => {
+    oneOffFuncAsTasks.push({
+      name: `run-custom-csfle-tests-${version}-${ref == 'master' ? ref : 'pinned-commit'}`,
```
baileympearson, line 555:
```suggestion
      name: `run-custom-csfle-tests-${version}-${ref === 'master' ? ref : 'pinned-commit'}`,
```

baileympearson (changes requested):
We also run the csfle tests as a part of the general integration test task for each server/topology (it's run as a part of run-test.sh) for all servers >= 4.2.  We also need to ensure we're using the appropriate library for these tasks as well.

Additionally, we install mongodb-client-encryption for the bson-ext tests.  Can we check if the csfle tests run in the bson-ext task, and if they do can we update these as well?

durran:
> We also run the csfle tests as a part of the general integration test task for each server/topology (it's run as a part of run-test.sh) for all servers >= 4.2. We also need to ensure we're using the appropriate library for these tasks as well.
> 
> Additionally, we install mongodb-client-encryption for the bson-ext tests. Can we check if the csfle tests run in the bson-ext task, and if they do can we update these as well?

It's now updated to run on everything now.

baileympearson (approved):
@durran One last comment - do you want to rebase?  It looks like the get_more unit test is still showing up in the diff.  Otherwise, looks good to me.  Thanks 😄

baileympearson (approved):

PR #3357, opened 2022-08-15
feat(NODE-4548): export ChangeStream class from top-level

nbbeeken:
#### What is changing?

NODE-4548
We need to make ChangeStream accessible

(Also cleaned up inconsistent naming in deprecated APIs)

Backwards compat.


dariakp (changes requested):
Can you explain why the raw user options need to be exposed?

baileympearson (changes requested):
Should this be a `feat` like Daria mentioned, and can we update the ticket description as well to reflect that you no longer are exporting the user options?

baileympearson (dismissed):

dariakp (approved):
Confirmed the use in the legacy client, here: https://github.com/mongodb-js/nodejs-mongodb-legacy/pull/1 so this should be ready to go

src/collection.ts
```diff
@@ -1610,13 +1610,13 @@ export class Collection<TSchema extends Document = Document> {
    * Updates documents.
    *
    * @deprecated use updateOne, updateMany or bulkWrite
-   * @param selector - The selector for the update operation.
+   * @param filter - The filter for the update operation.
```
durran, line 1613: Thanks for changing these to our spec language.

durran (approved):

PR #3360, opened 2022-08-15
feat(NODE-4540): add deprecation notice of callback support

nbbeeken:
#### What is changing?

Callback usage will now emit a deprecation warning once and instructions about how to migrate. 

Preview:
```txt
(node:55630) [MONGODB DRIVER] Warning: 
Callback support is deprecated and will be removed in the next major version.
If this impacts your usage of the driver there are few migration paths to consider:
  - You can start to using async/await syntax, which works with our driver today.
  - You can use .then/.catch at the end of any of our promise returning APIs and call you callback from within those.
  - You can install and import 'mongodb-legacy' a package that wraps all the async driver APIs and continues to provide the optional callback support.
You can read more about this here: LINK

(Use `node --trace-warnings ...` to show where the warning was created)
```

Maintainable API, Sensible Error handling, consistent return values, less overloads, better type info.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -457,6 +457,14 @@ export function maybePromise<T>(
         resolve(res);
       };
     });
+  } else {
+    const cbMsg = `\nCallback support is deprecated and will be removed in the next major version.
+If this impacts your usage of the driver there are few migration paths to consider:
+  - You can start to using async/await syntax, which works with our driver today.
+  - You can use .then/.catch at the end of any of our promise returning APIs and call you callback from within those.
+  - You can install and import 'mongodb-legacy' a package that wraps all the async driver APIs and continues to provide the optional callback support.
+You can read more about this here: LINK\n`;
+    emitWarningOnce(cbMsg);
   }
 
   wrapper((err, res) => {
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -457,6 +457,14 @@ export function maybePromise<T>(
         resolve(res);
       };
     });
+  } else {
+    const cbMsg = `\nCallback support is deprecated and will be removed in the next major version.
+If this impacts your usage of the driver there are few migration paths to consider:
```
dariakp, line 462:
```suggestion
If this impacts your usage of the driver there are a few migration paths to consider:
```

src/utils.ts
```diff
@@ -457,6 +457,14 @@ export function maybePromise<T>(
...
+  - You can start to using async/await syntax, which works with our driver today.
```
dariakp, line 463:
```suggestion
  - You can start using async/await syntax, which works with our driver today.
```

src/utils.ts
```diff
@@ -457,6 +457,14 @@ export function maybePromise<T>(
...
+  - You can use .then/.catch at the end of any of our promise returning APIs and call you callback from within those.
+  - You can install and import 'mongodb-legacy' a package that wraps all the async driver APIs and continues to provide the optional callback support.
```
dariakp, line 465:
```suggestion
  - You can install and import 'mongodb-legacy': a package that wraps all the currently existing async driver APIs and continues to provide the optional callback support.
```

src/utils.ts
```diff
@@ -457,6 +457,14 @@ export function maybePromise<T>(
...
+You can read more about this here: LINK\n`;
```
dariakp, line 466: Leaving comment to make sure we remember to update this

dariakp (changes requested):

src/utils.ts
```diff
@@ -1240,22 +1248,30 @@ export const DEFAULT_PK_FACTORY = {
  * ```
  */
 export const MONGODB_WARNING_CODE = 'MONGODB DRIVER' as const;
+export const MONGODB_DEPRECATIONS: {
+  emittedWarnings: Set<string>;
+  emitWarning: (message: string) => void;
+} = Object.seal(
+  Object.assign(Object.create(null), {
+    emittedWarnings: new Set<string>(),
+    emitWarning: (message: string) => process.emitWarning(message, { code: MONGODB_WARNING_CODE })
```
addaleax, line 1257: Actually, one thing that is nice about the `Logger` API is that it is configurable on a per-MongoClient basis, not mutable global state. This PR isn't really changing that, but if there's an obvious way to make that happen, that would probably be quite a bit more useful for mongosh.

nbbeeken:
Closing this in favor of a documentation only approach, will revisit runtime warning

PR #3361, opened 2022-08-15
test(NODE-4333): add unit test to check dependencies

durran:
### Description

Adds test to ensure dependencies don't accidentally change.

#### What is changing?

- Adds a unit test to test dependencies
- Adds a script to run the single unit test
- Adds an evergreen target to execute the script as a part of the lint task.

None

NODE-4333

### DIFF
--- a/test/unit/dependency.test.ts
+++ b/test/unit/dependency.test.ts
@@ -1,0 +1,13 @@
+import { expect } from 'chai';
+
+import { dependencies } from '../../package.json';
+
+const EXPECTED_DEPENDENCIES = ['bson', 'denque', 'mongodb-connection-string-url', 'socks'];
+
+describe('package.json', function () {
+  describe('dependencies', function () {
+    it('only contains the expected dependencies', function () {
+      expect(Object.getOwnPropertyNames(dependencies)).to.deep.equal(EXPECTED_DEPENDENCIES);
+    });
+  });
+});
--- a/.evergreen/check-dependencies.sh
+++ b/.evergreen/check-dependencies.sh
@@ -1,0 +1,2 @@
+#! /usr/bin/env bash
+npm run check:dependencies
--- a/.evergreen/check-dependencies.sh
+++ b/.evergreen/check-dependencies.sh
@@ -1,2 +1,0 @@
-#! /usr/bin/env bash
-npm run check:dependencies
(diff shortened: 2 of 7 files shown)


test/unit/dependency.test.ts
```diff
@@ -0,0 +1,13 @@
+import { expect } from 'chai';
+
+import { dependencies } from '../../package.json';
+
+const EXPECTED_DEPENDENCIES = ['bson', 'denque', 'mongodb-connection-string-url', 'socks'];
```
baileympearson, line 5: A couple of questions

- Do we want to check the versions too?  That might help us prevent accidental version bumps but it would also mean we could no longer rely on automated tools (like Snyk) to bump our dependencies for us.  I can go either way so I'm curious to hear your thoughts.
- The ticket mentions dev dependencies.  Did you choose not to check our dev dependencies as well?  I personally don't think it's necessary, I was just curious.

baileympearson (dismissed):
@durran One last question.  The ticket also mentions potentially implementing this as a github action, instead of an evergreen task.  The benefit is that it would be more visible on PRs that the check failed, as opposed to the more general "CI failed for some reason" that we will see if we leave it as an evergreen task.

It's an open question on the ticket so either work, but I'm curious if you choose evergreen for any particular reason.

test/unit/dependency.test.ts
durran: 1. I didn't check versions specifically for the Snyk reason.
2. I also didn't think doing dev dependencies was necessary.

durran:
> @durran One last question. The ticket also mentions potentially implementing this as a github action, instead of an evergreen task. The benefit is that it would be more visible on PRs that the check failed, as opposed to the more general "CI failed for some reason" that we will see if we leave it as an evergreen task.
> 
> It's an open question on the ticket so either work, but I'm curious if you choose evergreen for any particular reason.

@baileympearson Github action now in place.

baileympearson (changes requested):
Can we also remove the evergreen changes now that we're running it in github actions?

durran:
> Can we also remove the evergreen changes now that we're running it in github actions?

Done.

baileympearson (approved):

dariakp (approved):
Very nice, thank you!

PR #3362, opened 2022-08-16
fix(NODE-4516): dont require a connected client

durran:
### Description

Allows mongodb-client-encryption to be run without a connected `MongoClient`

#### What is changing?

- Removes the internal `bson` auto encryption option and no longer passes it to the auto encrypter when instantiating it.
- Removes the internal `bson` options from the topology.

https://github.com/mongodb/libmongocrypt/pull/442

None

NODE-4516


dariakp (changes requested):
Per the slack conversation, let's revisit the approach to make sure we don't break users of the older versions of bson and also consider bson-ext

PR #3363, opened 2022-08-16
fix(NODE-4429): select server sync for endSessions during close

durran:
### Description

This prevents the `endSessions` command from going through server selection when no server would be selected after calling `client.close()`. This could happen in situations where the server has gone down but server selection keeps forcing monitor checks for up to `serverSelectionTimeoutMS` when an immediate return is desired.

#### What is changing?

- `MongoClient`'s `close()` method now does a quick server selection to get the list of servers that would be returned with a primary preferred read preference. If nothing is returned then calling `endSessions` is skipped.

None

NODE-4429

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -522,7 +522,7 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
         })
         .then(() => {
           if (this.topology == null) {
-            return callback();
+            return;
           }
           // clear out references to old topology
           const topology = this.topology;
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -21,6 +21,7 @@ import { PromiseProvider } from './promise_provider';
 import type { ReadConcern, ReadConcernLevel, ReadConcernLike } from './read_concern';
 import { ReadPreference, ReadPreferenceMode } from './read_preference';
 import type { TagSet } from './sdam/server_description';
+import { readPreferenceServerSelector } from './sdam/server_selection';
 import type { SrvPoller } from './sdam/srv_polling';
 import type { Topology, TopologyEvents } from './sdam/topology';
 import { ClientSession, ClientSessionOptions, ServerSessionPool } from './sessions';
@@ -500,6 +501,16 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
 
       Promise.all(activeSessionEnds)
         .then(() => {
+          // If we would attempt to select a server and get nothing back we short circuit
+          // to avoid the server selection timeout.
+          const selector = readPreferenceServerSelector(ReadPreference.primaryPreferred);
+          const topologyDescription = this.topology.description;
+          const serverDescriptions = Array.from(topologyDescription.servers.values());
+          const servers = selector(topologyDescription, serverDescriptions);
+          if (servers.length === 0) {
+            return callback();
+          }
+
           const endSessions = Array.from(this.s.sessionPool.sessions, ({ id }) => id);
           if (endSessions.length === 0) return;
           return this.db('admin')
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -501,6 +501,9 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
 
       Promise.all(activeSessionEnds)
         .then(() => {
+          if (this.topology == null) {
+            return;
+          }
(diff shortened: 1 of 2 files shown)


test/integration/node-specific/mongo_client.test.ts
```diff
@@ -649,5 +650,25 @@ describe('class MongoClient', function () {
       expect(endEvents).to.have.lengthOf(1);
       expect(endEvents[0]).to.have.property('reply', undefined); // noReponse: true
     });
+
+    context('when server selection would return no servers', () => {
+      const serverDescription = new ServerDescription('a:1');
+
+      it('short circuits and does not end sessions', async () => {
```
durran, line 657: This test will hang and wait for server selection without the fix in, thus reproducing the path.

nbbeeken: Just to demonstrate that `endSession` does not do anything but bookkeeping, let's assert that the `client.s.sessionPool` still have one session in it both before and after close.

src/mongo_client.ts
```diff
@@ -500,6 +501,16 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
         .then(() => {
+          // If we would attempt to select a server and get nothing back we short circuit
+          // to avoid the server selection timeout.
+          const selector = readPreferenceServerSelector(ReadPreference.primaryPreferred);
+          const topologyDescription = this.topology.description;
+          const serverDescriptions = Array.from(topologyDescription.servers.values());
+          const servers = selector(topologyDescription, serverDescriptions);
+          if (servers.length === 0) {
+            return callback();
+          }
```
nbbeeken, line 512: I think we want to return here, could be wrong, but the callback should only be referenced in the last "then" of the chain. 

So actually a few lines down where we do: 
```ts
if (this.topology == null) {
  return callback();
}
```
Is a teeny bug

nbbeeken (changes requested):

src/mongo_client.ts
durran: I've updated to assert there is only 1 session in the pool and changed the check to just return.

test/integration/node-specific/mongo_client.test.ts
durran: Added.

nbbeeken (approved):
Small TS error I think but LGTM otherwise

src/mongo_client.ts
```diff
@@ -500,6 +501,16 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
 
       Promise.all(activeSessionEnds)
         .then(() => {
+          // If we would attempt to select a server and get nothing back we short circuit
+          // to avoid the server selection timeout.
+          const selector = readPreferenceServerSelector(ReadPreference.primaryPreferred);
+          const topologyDescription = this.topology.description;
```
dariakp, line 507: Can `this.topology` be undefined here? If so, what would be the correct way to handle that?

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -649,5 +650,26 @@ describe('class MongoClient', function () {
       expect(endEvents).to.have.lengthOf(1);
       expect(endEvents[0]).to.have.property('reply', undefined); // noReponse: true
     });
+
+    context('when server selection would return no servers', () => {
```
dariakp, line 654: Given the other comment about possibly undefined topology, it would be good to also cover that edge case, too, if we can

dariakp (changes requested):

src/mongo_client.ts
durran: I've add the check after to address this.

test/integration/node-specific/mongo_client.test.ts
durran: The test case to cover this would be quite involved to get the driver in an internal state where the topology was actually null at that point in the code. The only situation where it could happen would be simultaneous async calls to `close()` where one had gotten to the point of closing the topology before the other one got to the point of ending sessions. I started to try and write a test and it just got way out of hand trying to mimic that behaviour in it.

dariakp: ok, given it's an edge case I think it's not a big deal, and we're just following the pattern in the .then chain

leaving the comment open for visibility

dariakp (approved):

baileympearson (approved):

nbbeeken (approved):

PR #3378, opened 2022-08-19
test(NODE-4479): adjust number of iterations of flakey test

baileympearson:
#### What is changing?

The server selection operation count latency window test often fails because the observed percentage of requests sent to each mongos is outside the threshold.  Experimenting locally does demonstrate that the percentage approaches 50% to each mongos as the total number of operations increases (which makes sense - as the total time the test runs increases, any ephemeral variations will be averaged away). 

This PR bumps the total number of iterations to a number that when run locally leaves the latency window.  This is purely empirical, and it is possible that the number of iterations is not high enough to reliably run in CI but I believe this should resolve the flakiness of the tests.  I discussed these changes with Patrick and other drivers still see these tests flaking as well.  I propose merging this PR and watching for the flakey test, and if a period of time goes by and we don't see any flakiness, we can propose modifying the spec.

No.

Cleaning up our flakey CI.

### DIFF
--- a/test/integration/server-selection/server_selection.prose.operation_count.test.ts
+++ b/test/integration/server-selection/server_selection.prose.operation_count.test.ts
@@ -150,7 +150,9 @@ describe('operationCount-based Selection Within Latency Window - Prose Test', fu
     const numberTaskGroups = 10;
     const numberOfTasks = 1000;
     const totalNumberOfTasks = numberTaskGroups * numberOfTasks;
-
+    
+    // This test has proved flakey, not just for Node.  The number of iterations for the test has been increased,
+    // to prevent the test from failing.
     // Step 8: Start 10 concurrent threads / tasks that each run 100 findOne operations with empty filters using that client.
     await Promise.all(
       Array.from({ length: numberTaskGroups }, () => runTaskGroup(collection, numberOfTasks))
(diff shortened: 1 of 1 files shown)


durran:
Still happened in 1 run: https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_next_ubuntu_18.04_erbium_test_5.0_sharded_cluster_patch_a3ee7e2b0d4480c8ce6dd58e6ad17e1d62b06774_62ffe54ee3c331205ae78097_22_08_19_19_32_31/0?type=T#L5929

durran (approved):
Looks like 1000 did the trick.

test/integration/server-selection/server_selection.prose.operation_count.test.ts
```diff
@@ -147,13 +147,19 @@ describe('operationCount-based Selection Within Latency Window - Prose Test', fu
   it('equally distributes operations with both hosts are fine', TEST_METADATA, async function () {
     const collection = client.db('test-db').collection('collection0');
 
+    const numberTaskGroups = 10;
+    const numberOfTasks = 1000;
+    const totalNumberOfTasks = numberTaskGroups * numberOfTasks;
+
     // Step 8: Start 10 concurrent threads / tasks that each run 100 findOne operations with empty filters using that client.
-    await Promise.all(Array.from({ length: 10 }, () => runTaskGroup(collection, 100)));
+    await Promise.all(
```
dariakp, line 155: Just a suggestion to add a comment explaining why we're not actually doing Step 8 as written

baileympearson:
flakey test logged here -https://jira.mongodb.org/browse/NODE-4581

PR #3385, opened 2022-08-26
ci(NODE-4589): add benchmark testing task to CI

baileympearson:
#### What is changing?

This PR configures CI to run our driver performance benchmarks.  It adjusts the output of the script to be in a format that can be sent to Evergreen.

It also does some misc code cleanup.

- Some functions have been refactored to use async/await (but not benchmarking functions)
- Some functions have been given doc comments w/ jsdoc types.  These were added while I was developing to help me trace the code, and they seemed worth leaving.

No.


test/benchmarks/driverBench/index.js
```diff
@@ -1,11 +1,12 @@
 'use strict';
 
 const MongoBench = require('../mongoBench');
+const { writeFile } = require('fs/promises');
 
 const Runner = MongoBench.Runner;
 const commonHelpers = require('./common');
 
-const { LEGACY_HELLO_COMMAND } = require('../../../src/constants');
+const LEGACY_HELLO_COMMAND = 'ismaster';
```
durran, line 9: Was there a reason for this? This seems to go against our "remove oppressive language" tickets a while back and I think it's better that the old command is just defined in 1 place.

durran:
Nice refactor to async functions, cleans up the code nicely. Perf numbers showing up for others looking:

Screenshot 2022-08-29 at 21 41 22

durran (changes requested):

test/benchmarks/driverBench/index.js
baileympearson: The import fails - we can't import from the `.ts` file in our `.js`.  We also don't export the constants, so it's not available exported from the compiled JS.  I considered

- using ts-node instead of node when running the benchmarks (happy to go back to this one)
- converting the benchmarks to TS

but decided to just redefine the constant.  

I'm open to other approaches though

durran: I actually just realised we can just remove it completely and use `hello` in this case directly since we're running against rapid.

baileympearson: we actually missed this in the initial oppressive language epic but the spec has been changed to always use `hello` for that benchmark.  so removing `LEGACY_HELLO_COMMAND` is the right way forward

nbbeeken (changes requested):

test/benchmarks/driverBench/index.js
```diff
@@ -357,21 +357,31 @@ benchmarkRunner
...
+        },
+        metrics: [{ name: 'megabytes_per_second', value: result }]
+      };
+    });
   })
   .then(data => {
-    data.bsonType = BSON.serialize.toString().includes('native code') ? 'bson-ext' : 'js-bson';
-    console.log(data);
+    const results = JSON.stringify(data, undefined, 2);
+    console.error(inspect(results, { depth: Infinity }));
```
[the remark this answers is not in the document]
nbbeeken: Going to be addressed in the next PR.

test/benchmarks/driverBench/index.js
```diff
@@ -333,6 +331,8 @@ const benchmarkRunner = new Runner()
 benchmarkRunner
   .run()
   .then(microBench => {
+    // TODO - test against different BSON versions in CI
```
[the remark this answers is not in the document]
nbbeeken: Will be filed shortly @baileympearson if you wanna comment it here when you have it that'd be grand

nbbeeken (approved):

PR #3388, opened 2022-08-29
feat(NODE-4547): mark all callback APIs as deprecated

nbbeeken:
#### What is changing?

* tsdoc indicated deprecations for callbacks

* deprecation will strike through code in editors supporting type info

- N/A Changes are covered by tests
- N/A New TODOs have a related JIRA ticket

### DIFF
--- a/src/admin.ts
+++ b/src/admin.ts
@@ -29,26 +29,15 @@ export interface AdminPrivate {
  * @public
  *
  * @example
- * ```js
- * const MongoClient = require('mongodb').MongoClient;
- * const test = require('assert');
- * // Connection url
- * const url = 'mongodb://localhost:27017';
- * // Database Name
- * const dbName = 'test';
+ * ```ts
+ * import { MongoClient } from 'mongodb';
  *
- * // Connect using MongoClient
- * MongoClient.connect(url, function(err, client) {
- *   // Use the admin database for the operation
- *   const adminDb = client.db(dbName).admin();
- *
- *   // List all the available databases
- *   adminDb.listDatabases(function(err, dbs) {
- *     expect(err).to.not.exist;
- *     test.ok(dbs.databases.length > 0);
- *     client.close();
- *   });
- * });
+ * const client = new MongoClient('mongodb://localhost:27017');
+ * const admin = client.db().admin();
+ * const dbInfo = await admin.listDatabases();
+ * for (const db of dbInfo.databases) {
+ *   console.dir(db.name);
+ * }
  * ```
  */
 export class Admin {
--- a/src/admin.ts
+++ b/src/admin.ts
@@ -224,8 +224,10 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   removeUser(username: string): Promise<boolean>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   removeUser(username: string, callback: Callback<boolean>): void;
   removeUser(username: string, options: RemoveUserOptions): Promise<boolean>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   removeUser(username: string, options: RemoveUserOptions, callback: Callback<boolean>): void;
   removeUser(
     username: string,
--- a/src/admin.ts
+++ b/src/admin.ts
@@ -36,7 +36,7 @@ export interface AdminPrivate {
  * const admin = client.db().admin();
  * const dbInfo = await admin.listDatabases();
  * for (const db of dbInfo.databases) {
- *   console.dir(db.name);
+ *   console.log(db.name);
  * }
  * ```
  */
--- a/src/admin.ts
+++ b/src/admin.ts
@@ -71,8 +71,10 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   command(command: Document): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   command(command: Document, callback: Callback<Document>): void;
   command(command: Document, options: RunCommandOptions): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   command(command: Document, options: RunCommandOptions, callback: Callback<Document>): void;
   command(
     command: Document,
@@ -96,8 +98,10 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   buildInfo(): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   buildInfo(callback: Callback<Document>): void;
   buildInfo(options: CommandOperationOptions): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   buildInfo(options: CommandOperationOptions, callback: Callback<Document>): void;
   buildInfo(
     options?: CommandOperationOptions | Callback<Document>,
@@ -115,8 +119,10 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   serverInfo(): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   serverInfo(callback: Callback<Document>): void;
   serverInfo(options: CommandOperationOptions): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   serverInfo(options: CommandOperationOptions, callback: Callback<Document>): void;
   serverInfo(
     options?: CommandOperationOptions | Callback<Document>,
@@ -134,8 +140,10 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   serverStatus(): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   serverStatus(callback: Callback<Document>): void;
   serverStatus(options: CommandOperationOptions): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   serverStatus(options: CommandOperationOptions, callback: Callback<Document>): void;
   serverStatus(
     options?: CommandOperationOptions | Callback<Document>,
@@ -153,8 +161,10 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   ping(): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   ping(callback: Callback<Document>): void;
   ping(options: CommandOperationOptions): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   ping(options: CommandOperationOptions, callback: Callback<Document>): void;
   ping(
     options?: CommandOperationOptions | Callback<Document>,
@@ -174,12 +184,16 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   addUser(username: string): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   addUser(username: string, callback: Callback<Document>): void;
   addUser(username: string, password: string): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   addUser(username: string, password: string, callback: Callback<Document>): void;
   addUser(username: string, options: AddUserOptions): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   addUser(username: string, options: AddUserOptions, callback: Callback<Document>): void;
   addUser(username: string, password: string, options: AddUserOptions): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   addUser(
     username: string,
     password: string,
@@ -247,8 +261,10 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   validateCollection(collectionName: string): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   validateCollection(collectionName: string, callback: Callback<Document>): void;
   validateCollection(collectionName: string, options: ValidateCollectionOptions): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   validateCollection(
     collectionName: string,
     options: ValidateCollectionOptions,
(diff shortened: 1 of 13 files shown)


src/collection.ts
```diff
@@ -136,30 +136,28 @@ export interface CollectionPrivate {
...
+ *   name: string;
+ *   kind: 'dog' | 'cat' | 'fish';
+ * }
+ *
+ * const client = new MongoClient('mongodb://localhost:27017');
+ * const pets = client.db().collection<Pet>('pets');
+ *
+ * const petCursor = pets.find();
+ *
+ * for (const pet of petCursor) {
```
baileympearson, line 158:
```suggestion
 * for await (const pet of petCursor) {
```

src/collection.ts
```diff
@@ -1098,15 +1141,19 @@ export class Collection<TSchema extends Document = Document> {
    * @see https://docs.mongodb.com/manual/reference/operator/query/centerSphere/#op._S_centerSphere
    */
   countDocuments(): Promise<number>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   countDocuments(callback: Callback<number>): void;
   countDocuments(filter: Filter<TSchema>): Promise<number>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   countDocuments(callback: Callback<number>): void;
```
baileympearson, line 1148: can we remove this duplicate overload?

src/collection.ts
```diff
@@ -1173,10 +1223,13 @@ export class Collection<TSchema extends Document = Document> {
 
   // Embedded documents overload
   distinct(key: string): Promise<any[]>;
```
baileympearson, line 1225: why did these overloads not make it into api.js?

src/collection.ts
```diff
@@ -704,9 +727,11 @@ export class Collection<TSchema extends Document = Document> {
 
   // allow an override of the schema.
```
baileympearson, line 728: also curious why these overloads didn't make it into api.js

src/collection.ts
```diff
@@ -1582,7 +1647,7 @@ export class Collection<TSchema extends Document = Document> {
    * one will be added to each of the documents missing it by the driver, mutating the document. This behavior
    * can be overridden by setting the **forceServerObjectId** flag.
    *
-   * @deprecated Use insertOne, insertMany or bulkWrite instead.
+   * @deprecated Use insertOne, insertMany or bulkWrite instead. Callbacks are deprecated and will be removed in the next major version
```
baileympearson, line 1650: This comment applies to any previously deprecated functions that take callbacks in this PR.

How do want to handle these cases?  I couldn't find an existing ticket, so as far as I can tell we have no plans to remove these functions in 5.0.  My preference would be to remove these functions, and if so, I think we should 1. file a jira ticket for their removal and make sure it's tagged for 5.0 and 2. remove the callback specific deprecation.

If we don't decide to go this route - I think we should properly overload all these functions and deprecate the callback overloads specifically, mentioning their removal in v5.  

Thoughts?

src/collection.ts
```diff
@@ -352,20 +354,22 @@ export class Collection<TSchema extends Document = Document> {
...
    *
-   *  { deleteOne: { filter: {c:1} } }
+   *   { deleteOne: { filter: { c: 1 } } },
    *
-   *  { deleteMany: { filter: {c:1} } }
+   *   { deleteMany: { filter: { c: 1 } } },
    *
-   *  { replaceOne: { filter: {c:3}, replacement: {c:4}, upsert:true} }
+   *   { replaceOne: { filter: { c: 3 }, replacement: { c: 4 }, upsert: true } },
+   * ]
```
baileympearson, line 372:
```suggestion
   * Legal operation types are
   * - `insertOne`
   * - `updateOne`
   * - `updateMany`
   * - `deleteOne`
   * - `deleteMany`
   * - `replaceOne`
```

src/db.ts
```diff
@@ -110,18 +110,21 @@ export interface DbOptions extends BSONSerializeOptions, WriteConcernOptions, Lo
...
+ * }
+ *
+ * const client = new MongoClient('mongodb://localhost:27017');
+ * const db = client.db();
+ *
+ * // Create a collection that validates our union
+ * await db.createCollection<Pet>('pets', {
+ *   validator: { $expr: { $in: ['$kind', ['dog', 'cat', 'fish']] } }
+ * })
  * ```
```
baileympearson, line 128: What is the purpose of this example?

baileympearson: Should option specific examples be attached to the individual options, not the class doc comment?

src/mongo_client.ts
```diff
@@ -318,36 +318,15 @@ const kOptions = Symbol('options');
...
- * const dbName = 'test';
- * // Connect using MongoClient
- * const mongoClient = new MongoClient(url);
- * mongoClient.connect(function(err, client) {
- *   const db = client.db(dbName);
- *   client.close();
- * });
- * ```
+ * ```ts
+ * import { MongoClient } from 'mongodb';
```
baileympearson, line 322: Same question as the doc comment on db.  What is the purpose of the comment, and why are we only demonstrating command monitoring but otherwise not mentioning it at all?

baileympearson (changes requested):
I went through the PR and for every method listed in api.js in the legacy repo, I made sure it was deprecated in the driver.  I then deleted it from my local copy of api.js.  Afterwards, I was left with this:

```javascript
/* eslint-disable prettier/prettier */


const { byStrings, sorted } = require('./utils');

module.exports = Object.create(null);
Object.defineProperty(module.exports, '__esModule', { value: true });

const commonCursorApis = [
  { className: 'AbstractCursor', method: 'close', returnType: 'Promise<void>' },
  { className: 'AbstractCursor', method: 'forEach', returnType: 'Promise<void>', possibleCallbackPositions: [1] },
  { className: 'AbstractCursor', method: 'hasNext', returnType: 'Promise<boolean>' },
  { className: 'AbstractCursor', method: 'next', returnType: 'Promise<TSchema | null>' },
  { className: 'AbstractCursor', method: 'toArray', returnType: 'Promise<TSchema[]>' },
  { className: 'AbstractCursor', method: 'tryNext', returnType: 'Promise<TSchema | null>' },
]
module.exports.commonCursorApis = commonCursorApis;

const cursorClasses = [
  'FindCursor',
  'AggregationCursor',
  'ListIndexesCursor',
  'ListCollectionsCursor'
]
module.exports.cursorClasses = cursorClasses;

const api = [
  // Super class of cursors, we do not directly override these but override them in the inherited classes
  ...commonCursorApis.flatMap(({ method, returnType, possibleCallbackPositions }) => cursorClasses.map(cursorClass => ({ className: cursorClass, method, returnType, possibleCallbackPositions }))),

  { className: 'Admin', method: 'removeUser', returnType: 'Promise<boolean>' },

  { className: 'AggregationCursor', method: 'explain', returnType: 'Promise<Document>' },

  { className: 'ChangeStream', method: 'close', returnType: 'Promise<void>' },

  { className: 'FindCursor', method: 'count', returnType: 'Promise<number>' },
  { className: 'FindCursor', method: 'explain', returnType: 'Promise<Document>' },

  // Manually test the static version of connect
  // This is listed here as a reference for completeness, but it is tested manually
  // it is checked to exist in index.test.js
  // its functionally tested in maybe_callback.test.js
  // { className: 'MongoClient', method: 'static connect', returnType: 'Promise<this>' },
];

module.exports.api = api;
module.exports.classNames = new Set(api.map(({className}) => className))
module.exports.classNameToMethodList = new Map(api.map((api, _, array) =>
  [api.className, sorted(Array.from(new Set(Array.from(array.filter(v => v.className === api.className), method => method))), (a, b) => byStrings(a.method, b.method))]
));

```

It looks like we're missing the cursor apis and a few misc functions.

src/collection.ts
nbbeeken: Which ones? they are present in the types: https://github.com/mongodb-js/nodejs-mongodb-legacy/blob/main/mongodb-legacy.d.ts#L404 (unless I'm missing something)

src/collection.ts
nbbeeken: fixed!

src/collection.ts
nbbeeken: can you clarify which? 
https://github.com/mongodb-js/nodejs-mongodb-legacy/blob/main/mongodb-legacy.d.ts#L619 here?

src/db.ts
nbbeeken: I was going off of what was here before + some extra, the idea being to give an example of what this thing is for, it's the first thing you see when you land on the API page for `Db`. I just added something that gives the example some heft (an async operation that actually reaches out to mongodb)

src/mongo_client.ts
nbbeeken: to give an example of what this thing is for, it's the first thing you see when you land on the API page for `MongoClient`, we have extensive docs on how to use command monitoring on the manual site, this is just to kick off what area of the API you're looking at.

src/collection.ts
nbbeeken: Isn't there value in having examples that can be copy pasta-ed?

src/admin.ts
```diff
@@ -29,26 +29,15 @@ export interface AdminPrivate {
...
- *   // List all the available databases
- *   adminDb.listDatabases(function(err, dbs) {
- *     expect(err).to.not.exist;
- *     test.ok(dbs.databases.length > 0);
- *     client.close();
- *   });
- * });
+ * const client = new MongoClient('mongodb://localhost:27017');
+ * const admin = client.db().admin();
+ * const dbInfo = await admin.listDatabases();
```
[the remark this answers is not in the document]
nbbeeken: I don't think it is that important, the examples at the top of the classes are a sampling of API, they are meant to read more than used. If they are used and you don't have top-level await, our manual docs demonstrate the `async run() {}; run()` pattern plenty, and the syntax error you would get will lead to an explaination of how to get top-level await if one desires it.

src/collection.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-4614

src/collection.ts
baileympearson: Sure, but the doesn't really match the examples - `Legal operation types are` implies a list of the valid operation types whereas we have a list of examples.  I am fine with both or rewording the title

src/cursor/abstract_cursor.ts
```diff
@@ -344,7 +345,9 @@ export abstract class AbstractCursor<
 
   /** Get the next available document from the cursor, returns null if no more documents are available. */
   next(): Promise<TSchema | null>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   next(callback: Callback<TSchema | null>): void;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
```
baileympearson, line 350:
```suggestion
```

src/cursor/find_cursor.ts
```diff
@@ -155,6 +155,7 @@ export class FindCursor<TSchema = any> extends AbstractCursor<TSchema> {
 
   /** Execute the explain for the cursor */
   explain(): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   explain(callback: Callback): void;
   explain(verbosity?: ExplainVerbosityLike): Promise<Document>;
   explain(
```
baileympearson, line 161: We're missing a deprecation on the overload where explain has verbosity and a callback

src/collection.ts
baileympearson: Can we remove the callback deprecation warning here as well?

src/collection.ts
baileympearson: It's not a type change if it's a duplicate overload.  it's literally a duplicate of line 1145

src/collection.ts
nbbeeken: We should mention both but maybe I could rephrase? since currently migrating off of `.insert` is 
```diff
- c.insert({ doc: 1 }, callback)
+ c.insertMany([{ doc: 1 }], callback)
```
Now that callbacks will be removed in addition it is more involved than just wrapping the first argument in brackets

src/cursor/find_cursor.ts
nbbeeken: This is the function's defintion, there is no overload that accepts `[verbosity, callback]`

src/cursor/aggregation_cursor.ts
```diff
@@ -78,6 +78,7 @@ export class AggregationCursor<TSchema = any> extends AbstractCursor<TSchema> {
 
   /** Execute the explain for the cursor */
   explain(): Promise<Document>;
+  /** @deprecated Callbacks are deprecated and will be removed in the next major version */
   explain(callback: Callback): void;
   explain(verbosity: ExplainVerbosityLike): Promise<Document>;
   explain(
```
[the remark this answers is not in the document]
nbbeeken: This is the function's defintion, there is no overload that accepts `[verbosity, callback]`

src/cursor/abstract_cursor.ts
nbbeeken: Let's save type changes for another PR/ticket

src/collection.ts
nbbeeken: An overload has is emitted to the definitions, removing it changes how the defintions resolve the correct call format. I agree we should remove it, I just want this PR to involve only docs changes, the plan is to generate the docs pages when we're done checking that the @deprecation tag is in all the right places

src/collection.ts
baileympearson: I still think we should only mention the whole deprecation.  `insert` (and similar functions) will be entirely removed in v5.  I think the user flow of 

- user sees insert is deprecated and switches to insertOne 
- user sees insertOne with callbacks is deprecated and switches to promises 

is fine because at each stage, they see an specific deprecation warning and know how to resolve it.  a single warning that combines both (i.e., `Use insertOne, insertMany or bulkWrite instead. Callbacks are deprecated and will be removed in the next major version`) is more confusing.  Maybe we should ask the team / Matt

src/cursor/find_cursor.ts
baileympearson: see above

src/collection.ts
baileympearson: This won't affect function call resolution at all - if that overload is matched, it will match the first instance of that overload.  If not, it will never match the second.

Can you explain the plan about generating the docs more?  Will that be included in this PR or in follow up work?

nbbeeken: In this PR I'll pull in the generated docs changes so we can merge it all together, but GH diff is difficult with many changes so I'm saving that for when the review is complete

Yea that's correct about the behavior at least to my understanding

src/collection.ts
nbbeeken: @dariakp @durran

src/cursor/abstract_cursor.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-4623

src/cursor/aggregation_cursor.ts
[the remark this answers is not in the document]
nbbeeken: https://jira.mongodb.org/browse/NODE-4623

src/collection.ts
dariakp: from the conversation in slack, callback deprecation and removal is distinct and more certain than the whole api deprecation and removal, we should have separate notes so that the corresponding tasks can be completed separately

baileympearson (approved):

PR #3390, opened 2022-08-30
fix(NODE-4557): randomize servers when there are only 2 eligible servers 

baileympearson:
#### What is changing?

When we encounter exactly 2 eligible servers in the server selection process, we now shuffle the array of two servers.   This prevents a scenario where, when we receive a sequence of operations and each operation completes before the next (or approximately this scenario), we would select the same server for each request.

No.

### DIFF
--- a/src/sdam/topology.ts
+++ b/src/sdam/topology.ts
@@ -903,6 +903,20 @@ function processWaitQueue(topology: Topology) {
       continue;
     } else if (selectedDescriptions.length === 1) {
       selectedServer = topology.s.servers.get(selectedDescriptions[0].address);
+    } else if (selectedDescriptions.length === 2) {
+      // If there are only two servers, we avoid shuffling the array of
+      // server descriptions
+      const server1 = topology.s.servers.get(selectedDescriptions[0].address);
+      const server2 = topology.s.servers.get(selectedDescriptions[1].address);
+
+      if (server1?.s.operationCount === server2?.s.operationCount) {
+        selectedServer = Math.floor(Math.random() * 2) === 0 ? server1 : server2;
+      } else {
+        selectedServer =
+          server1 && server2 && server1.s.operationCount < server2.s.operationCount
+            ? server1
+            : server2;
+      }
     } else {
       const descriptions = shuffle(selectedDescriptions, 2);
       const server1 = topology.s.servers.get(descriptions[0].address);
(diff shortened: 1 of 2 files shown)


nbbeeken (approved):
LGTM

src/sdam/topology.ts
```diff
@@ -903,10 +903,22 @@ function processWaitQueue(topology: Topology) {
...
     } else if (selectedDescriptions.length === 1) {
       selectedServer = topology.s.servers.get(selectedDescriptions[0].address);
+    } else if (selectedDescriptions.length === 2) {
+      // If there are only two servers, we avoid shuffling the array of
+      // server descriptions
+      const server1 = topology.s.servers.get(selectedDescriptions[0].address);
+      const server2 = topology.s.servers.get(selectedDescriptions[1].address);
+
+      if (server1?.s.operationCount === server2?.s.operationCount) {
+        selectedServer = Math.floor(Math.random() * 2) === 0 ? server1 : server2;
```
durran, line 913: I am assuming this is for performance to avoid using the `shuffle` util method?

baileympearson: Yup!

nbbeeken (dismissed):

durran (approved):

src/sdam/topology.ts
```diff
@@ -903,10 +903,21 @@ function processWaitQueue(topology: Topology) {
       continue;
     } else if (selectedDescriptions.length === 1) {
       selectedServer = topology.s.servers.get(selectedDescriptions[0].address);
+    } else if (selectedDescriptions.length === 2) {
```
jyemin, line 906: Why not fall through to the `else` here instead of special-casing two servers?   If it's just to get a slight performance boost, it doesn't seem worth it unless it's measurable and significant.

nbbeeken: Micro benchmarks is all I got so we can go with the assumption that the performance will be outclassed by other steps in the process of executing an operation and just use shuffle here. We discussed optimizing the 2 item case within shuffle, which we can do separately, and that'll be best of both worlds

nbbeeken:
```txt
shuffle(["a", "b"], 2)                x 36,828,427 ops/sec ±0.25% (98 runs sampled)
Math.random() > 0.5 ? item1 : item2   x 155,421,810 ops/sec ±0.10% (98 runs sampled)
Math.floor(Math.random() * 2) === 0   x 135,506,163 ops/sec ±0.12% (101 runs sampled)

Fastest is Math.random() > 0.5 ? item1 : item2
```

jyemin: Since an application isn't likely to break a few thousand operations per second, this is unlikely to make a difference in practice.

nbbeeken (approved):

jyemin (approved):

PR #3396, opened 2022-09-02
feat(NODE-4607): add exports needed by legacy client

nbbeeken:
#### What is changing?

Added:

- ClientSession
- GridFSBucketWriteStream
- GridFSBucketReadStream (not needed, but symmetry)
- OrderedBulkOperation
- UnorderedBulkOperation

This is not a feature for users, but we need these exports at the top level to make the legacy wrapper package be able to import without using nested paths.

Open Q:
I want to add a little unit test that checks our exports, is checking just keys fine? and is there any obvious test organization changes you see?


test/unit/index.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  ['TopologyDescriptionChangedEvent',    'reason'],
+  ['TopologyOpeningEvent',               'reason'],
+  ['SrvPollingEvent',                    'reason'],
+
+  ['GridFSBucketReadStream',             'reason'],
+  ['GridFSBucketWriteStream',            'reason'],
+  ['OrderedBulkOperation',               'reason'],
+  ['UnorderedBulkOperation',             'reason'],
+
+  // TS-NODE Adds these keys but they are undefined... weird
```
dariakp, line 129: do we want to do any further investigation into this? having a comment saying that this is weird is a bit of an anti-pattern for the repo - it's not going to be anymore clear to the next person coming across this

test/unit/index.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  ['Collection',                         'reason'],
+  ['Db',                                 'reason'],
+  ['FindCursor',                         'reason'],
+  ['GridFSBucket',                       'reason'],
+  ['ListCollectionsCursor',              'reason'],
+  ['ListIndexesCursor',                  'reason'],
+  ['Logger',                             'reason'],
+  ['MongoClient',                        'reason'],
+  // global promise setter
+  ['Promise',                            'global promise setter'],
```
dariakp, line 76: was the intention to use the string values for some sort of more detailed testing? I see we're only using ts-node, can we clarify the intention behind the "reason"? if we aren't going to use the other string, I'd suggest breaking up the exports into separately named sets or plain arrays and then testing the union or individual members as desired; currently the map approach seems unnecessarily complicated

test/unit/index.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    const difference = setDifference(currentExports, expectedExports.keys());
+    expect(
+      difference,
+      `Found extra exports [${Array.from(difference).join(
+        ', '
+      )}], if these are expected, just add them to the list in this file`
+    ).to.have.lengthOf(0);
+  });
+
+  it('meta test: ts-node adds keys to the module that point to undefined', () => {
```
dariakp, line 152: assuming we figure out why this is happening and that it is correct, we should still keep with the convention for `it` phrasing, i.e., change this to something like `should export keys added by ts-node as undefined`

test/unit/index.test.ts
```diff
@@ -0,0 +1,161 @@
...
+      sorted(expectedExports.keys(), byStrings)
+    );
+  });
+
+  it('should export only keys in our exports list', () => {
+    const currentExports = Object.keys(mongodb);
+    const difference = setDifference(currentExports, expectedExports.keys());
+    expect(
+      difference,
+      `Found extra exports [${Array.from(difference).join(
```
dariakp, line 146: I think this doesn't differentiate between extra exports vs missing exports, do we want to make this more precise?

also would this be easier if we use arrays instead of sets? take a sorted array and test for deep equality, that should give a better error message automatically; what's the actual benefit of using a set here?

dariakp (changes requested):

test/unit/index.test.ts
nbbeeken: Yea I have no idea what makes these two props special, they don't exist when I import from lib, only when I run the driver in ts-node do they appear. I think its fine to live with, I can clarify the comment. The assertions that these remain equal to undefined should cover us.

I don't want to import from lib because it will block the whole mocha run if we push a small accidental TS issue (annoying when someone is iterating a lot)

test/unit/index.test.ts
nbbeeken: Yea I can change this to an array, I set up the map preparing for some discussion on how to organize the tests, but I think its fine how we currently have it, it's a pretty simple test of key existence

test/unit/index.test.ts
nbbeeken: yep, fixed

test/unit/index.test.ts
nbbeeken: Yea the strict equality of two sorted arrays covers us, if either side has items added or removed it should show up

test/unit/index.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+describe('mongodb entrypoint', () => {
+  it('should export all and only the expected keys in expected_exports', () => {
+    expect(sorted(Object.keys(mongodb), byStrings)).to.deep.equal(
+      sorted(EXPECTED_EXPORTS, byStrings)
+    );
+  });
+
+  it('should export keys added by ts-node as undefined', () => {
+    expect(TS_NODE_EXPORTS).to.have.length.greaterThan(0); // Are there no longer fake ts-node exports?
```
dariakp, line 136: I don't think we need the comment unless it's going to suggest an action (e.g., delete the declared const array because the test is now an no-op)

dariakp (changes requested):
just one small thing

dariakp (approved):

baileympearson (approved):

PR #3397, opened 2022-09-04
fix(NODE-3986): unskip MONGODB-AWS test

biniona-mongodb:
### Description

Update driver to throw exception when the following conditions are met:

- Using `MONGODB-AWS` auth mechanism
- A `username` is provided
- A `password` is not provided 

#### What is changing?

Adding check to see if preceding conditions are met when constructing `mongoOptions` object.

Unskip [test](https://github.com/mongodb/specifications/blob/master/source/auth/tests/connection-string.json#L453-L456) that checks this behavior.

No

[NODE-3986](https://jira.mongodb.org/browse/NODE-3986)

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -8,7 +8,12 @@ import { MongoCredentials } from './cmap/auth/mongo_credentials';
 import { AUTH_MECHS_AUTH_SRC_EXTERNAL, AuthMechanism } from './cmap/auth/providers';
 import { Compressor, CompressorName } from './cmap/wire_protocol/compression';
 import { Encrypter } from './encrypter';
-import { MongoAPIError, MongoInvalidArgumentError, MongoParseError } from './error';
+import {
+  MongoAPIError,
+  MongoInvalidArgumentError,
+  MongoMissingCredentialsError,
+  MongoParseError
+} from './error';
 import { Logger, LoggerLevel } from './logger';
 import {
   DriverInfo,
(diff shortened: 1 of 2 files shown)


src/connection_string.ts
```diff
@@ -407,6 +407,12 @@ export function parseOptions(
       });
     }
 
+    if (isAws && mongoOptions.credentials.username && !mongoOptions.credentials.password) {
+      throw new MongoParseError(
+        `${mongoOptions.credentials.mechanism} must receive a password when a username is specified`
```
durran, line 412:
```suggestion
        `When using ${mongoOptions.credentials.mechanism} password must be set when a username is specified`
```

durran (changes requested):
Suggesting small wording change.

src/connection_string.ts
biniona-mongodb: Implemented suggested change 🙂

src/connection_string.ts
```diff
@@ -407,6 +407,12 @@ export function parseOptions(
       });
     }
 
+    if (isAws && mongoOptions.credentials.username && !mongoOptions.credentials.password) {
+      throw new MongoParseError(
```
dariakp, line 411: "ParseError" is not really the correct error type to use here, since it's not that we were unable to parse the options, it's that one of the expected options wasn't provided. This is an instance of a `MongoAPIError`, perhaps the more specific `MongoMissingCredentialsError`.

dariakp (changes requested):

src/connection_string.ts
biniona-mongodb: Thanks for the context, was not aware of NODE-4015. Using `MongoMissingCredentialsError` as suggested, agree the description of this error seems to fit this scenario well: 

>  An error generated when the user fails to provide authentication credentials before attempting
    to connect to a mongo server instance.

dariakp (approved):

durran (approved):

PR #3399, opened 2022-09-07
refactor(NODE-4617): use promise apis in benchmarks

baileympearson:
#### What is changing?

Our benchmarks now test the promise-based API instead of the callback based api.

No.

We should benchmark the API we plan to support.

### DIFF
--- a/test/benchmarks/mongoBench/runner.js
+++ b/test/benchmarks/mongoBench/runner.js
@@ -9,33 +9,10 @@ function percentileIndex(percentile, total) {
   return Math.max(Math.floor((total * percentile) / 100 - 1), 0);
 }
 
-function timeDoneTask(task, ctx) {
-  return new Promise((resolve, reject) => {
-    let called = false;
-    const start = performance.now();
-    task.call(ctx, err => {
-      const end = performance.now(start);
-      if (called) return;
-      if (err) return reject(err);
-      return resolve((end - start) / 1000);
-    });
-  });
-}
-
 async function timeTask(task, ctx) {
-  // Some tasks are async, so they take callbacks.
-  if (task.length) {
-    return timeDoneTask(task, ctx);
-  }
-
   const start = performance.now();
-  const ret = task.call(ctx);
-  let end = performance.now();
-
-  if (ret && ret.then) {
-    await ret;
-    end = performance.now();
-  }
+  await task.call(ctx);
+  const end = performance.now();
 
   return (end - start) / 1000;
 }
--- a/test/benchmarks/mongoBench/runner.js
+++ b/test/benchmarks/mongoBench/runner.js
@@ -165,22 +165,14 @@ class Runner {
     let time = performance.now() - start;
     let count = 1;
 
-    if (benchmark._taskType === 'sync') {
+    const taskTimer = benchmark._taskType === 'sync' ? timeSyncTask : timeAsyncTask;
+
       while (time < maxExecutionTime && (time < minExecutionTime || count < minExecutionCount)) {
         await benchmark.beforeTask.call(ctx);
-        const executionTime = timeSyncTask(benchmark.task, ctx);
+      const executionTime = await taskTimer(benchmark.task, ctx);
         rawData.push(executionTime);
         count++;
         time = performance.now();
-      }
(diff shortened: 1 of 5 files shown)


test/benchmarks/mongoBench/runner.js
```diff
@@ -9,33 +9,10 @@ function percentileIndex(percentile, total) {
...
   const start = performance.now();
-  const ret = task.call(ctx);
-  let end = performance.now();
-
-  if (ret && ret.then) {
-    await ret;
-    end = performance.now();
-  }
+  await task.call(ctx);
+  const end = performance.now();
```
baileympearson, line 15: I would like to discuss this.  The bson benchmarks are synchronous.  I removed the logic to conditionally await the return of `task.call` because you can await synchronous expressions, but this does add one additional microtask before ending the timing to synchronous tasks.  I think this is okay - I think that the additional time taken before ending the timing is negligible compared to the task execution.

But I'm open to adding back the logic to only conditionally await the tasks.

test/benchmarks/mongoBench/suites/singleBench.js
```diff
@@ -89,7 +58,12 @@ function makeSingleBench(suite) {
         .beforeTask(dropCollection)
         .beforeTask(createCollection)
         .beforeTask(initCollection)
-        .task(makeTestInsertOne(10000))
+        .task(async function () {
+          for (let i = 0; i < 10000; ++i) {
+            const doc = Object.assign({}, this.doc);
```
nbbeeken, line 63: Can we move this line to a `beforeTask`?

test/benchmarks/mongoBench/runner.js
nbbeeken: I think it's worth fixing, nodejs has `utils/type` module that exports a `isAsyncFunction` type checker we can use. Which means we must pass `async functions` to the `.task(` method to get them awaited, I think in the sync case I'd save the result of task.call(x) and if it is a promise throw, just to fail if someone introduces a manual promise retuning task

test/benchmarks/mongoBench/suites/singleBench.js
```diff
@@ -107,7 +81,12 @@ function makeSingleBench(suite) {
         .beforeTask(dropCollection)
         .beforeTask(createCollection)
         .beforeTask(initCollection)
-        .task(makeTestInsertOne(10))
+        .task(async function () {
+          for (let i = 0; i < 10; ++i) {
+            const doc = Object.assign({}, this.doc);
```
nbbeeken, line 86: Can we move this line to a `beforeTask`?

nbbeeken (changes requested):

test/benchmarks/mongoBench/runner.js
nbbeeken: From slack: Going with an .isAsync or .isSync flag of some kind

test/benchmarks/mongoBench/runner.js
```diff
@@ -180,12 +165,22 @@ class Runner {
...
     let count = 1;
 
-    while (time < maxExecutionTime && (time < minExecutionTime || count < minExecutionCount)) {
-      await benchmark.beforeTask.call(ctx);
-      const executionTime = await timeTask(benchmark.task, ctx);
-      rawData.push(executionTime);
-      count++;
-      time = performance.now();
+    if (benchmark._taskType === 'sync') {
+      while (time < maxExecutionTime && (time < minExecutionTime || count < minExecutionCount)) {
```
dariakp, line 169: is there a downside to doing the if/else earlier, something like `const timeTask = becnhmark._taskType === 'sync' ? timeSyncTask : timeAsyncTask` and keeping the while loop the same since `await` works equally well in front of sync and async things?

dariakp (changes requested):

test/benchmarks/mongoBench/runner.js
baileympearson: Nope. that works too

baileympearson: Done!

dariakp (approved):

PR #3400, opened 2022-09-08
feat(NODE-4234): get aws kms providers from env

durran:
### Description

Pulls AWS KMS providers from the environment if the user provided aws kms providers are empty.

#### What is changing?

- Allows definition of a user provided callback to get the KMS providers that will be used if present. This was already available in mongodb-client-encryption and is now just passed through from the driver. (NODE-3934)
- Defines an internal callback in the driver to also pull AWS KMS providers from the environment in the same way that MONGODB-AWS does this, but ignoring the URI values. This acts as a fallback if the user provided callback does not return credentials or does not exist.
- Exports the driver defined callback function so users can also use it when using the `ClientEncryption` object directly.
- Modifies the unified test and legacy spec test runners to remove the AWS KMS providers in the tests when `REFRESH_AWS_CREDENTIALS` is present in the environment in order to spec test the refresh functionality against AWS environments
- Refactors FLE and AWS test tasks in Evergreen in order to mix and match AWS environments with other tasks.

Yes. The `onKmsProviderRefresh` needs to be documented.

NODE-4234


durran:
Patch for AWS changes: https://spruce.mongodb.com/version/631f12e70305b92ba80e795e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #3406, opened 2022-09-12
refactor(NODE-4631): change_stream, gridfs to use maybeCallback

nbbeeken:
#### What is changing?

- Adds maybeCallback helper with tests
- Changes PromiseProvider to communicate an unset state
- Refactors:
  - ChangeStream
  - ClientSession
  - GridFS

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -31,6 +31,7 @@ import {
   Callback,
   getTopology,
   hasAtomicOperators,
+  maybeCallback,
   MongoDBNamespace,
   resolveOptions
 } from '../utils';
@@ -1270,39 +1271,44 @@ export abstract class BulkOperationBase {
     options?: BulkWriteOptions | Callback<BulkWriteResult>,
     callback?: Callback<BulkWriteResult>
   ): Promise<BulkWriteResult> | void {
-    if (typeof options === 'function') (callback = options), (options = {});
-    options = options ?? {};
+    callback =
+      typeof callback === 'function'
+        ? callback
+        : typeof options === 'function'
+        ? options
+        : undefined;
 
-    if (this.s.executed) {
-      return handleEarlyError(new MongoBatchReExecutionError(), callback);
-    }
+    return maybeCallback(async () => {
+      options = options != null && typeof options !== 'function' ? options : {};
 
-    const writeConcern = WriteConcern.fromOptions(options);
-    if (writeConcern) {
-      this.s.writeConcern = writeConcern;
-    }
+      if (this.s.executed) {
+        throw new MongoBatchReExecutionError();
+      }
 
-    // If we have current batch
-    if (this.isOrdered) {
-      if (this.s.currentBatch) this.s.batches.push(this.s.currentBatch);
-    } else {
-      if (this.s.currentInsertBatch) this.s.batches.push(this.s.currentInsertBatch);
-      if (this.s.currentUpdateBatch) this.s.batches.push(this.s.currentUpdateBatch);
-      if (this.s.currentRemoveBatch) this.s.batches.push(this.s.currentRemoveBatch);
-    }
-    // If we have no operations in the bulk raise an error
-    if (this.s.batches.length === 0) {
-      const emptyBatchError = new MongoInvalidArgumentError(
-        'Invalid BulkOperation, Batch cannot be empty'
-      );
-      return handleEarlyError(emptyBatchError, callback);
-    }
+      const writeConcern = WriteConcern.fromOptions(options);
+      if (writeConcern) {
+        this.s.writeConcern = writeConcern;
+      }
 
-    this.s.executed = true;
-    const finalOptions = { ...this.s.options, ...options };
-    const operation = new BulkWriteShimOperation(this, finalOptions);
+      // If we have current batch
+      if (this.isOrdered) {
+        if (this.s.currentBatch) this.s.batches.push(this.s.currentBatch);
+      } else {
+        if (this.s.currentInsertBatch) this.s.batches.push(this.s.currentInsertBatch);
+        if (this.s.currentUpdateBatch) this.s.batches.push(this.s.currentUpdateBatch);
+        if (this.s.currentRemoveBatch) this.s.batches.push(this.s.currentRemoveBatch);
+      }
+      // If we have no operations in the bulk raise an error
+      if (this.s.batches.length === 0) {
+        throw new MongoInvalidArgumentError('Invalid BulkOperation, Batch cannot be empty');
+      }
 
-    return executeOperation(this.s.collection.s.db.s.client, operation, callback);
+      this.s.executed = true;
+      const finalOptions = { ...this.s.options, ...options };
+      const operation = new BulkWriteShimOperation(this, finalOptions);
+
+      return await executeOperation(this.s.collection.s.db.s.client, operation);
+    }, callback);
   }
 
   /**
@@ -1350,20 +1356,6 @@ Object.defineProperty(BulkOperationBase.prototype, 'length', {
     return this.s.currentIndex;
   }
 });
-
-/** helper function to assist with promiseOrCallback behavior */
-function handleEarlyError(
-  err?: AnyError,
-  callback?: Callback<BulkWriteResult>
-): Promise<BulkWriteResult> | void {
-  if (typeof callback === 'function') {
-    callback(err);
-    return;
-  }
-
-  const PromiseConstructor = PromiseProvider.get() ?? Promise;
-  return PromiseConstructor.reject(err);
-}
 
 function shouldForceServerObjectId(bulkOperation: BulkOperationBase): boolean {
   if (typeof bulkOperation.s.options.forceServerObjectId === 'boolean') {
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -22,7 +22,6 @@ import { executeOperation } from '../operations/execute_operation';
 import { InsertOperation } from '../operations/insert';
 import { AbstractOperation, Hint } from '../operations/operation';
 import { makeUpdateStatement, UpdateOperation, UpdateStatement } from '../operations/update';
-import { PromiseProvider } from '../promise_provider';
 import type { Server } from '../sdam/server';
 import type { Topology } from '../sdam/topology';
 import type { ClientSession } from '../sessions';
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -1276,38 +1276,41 @@ export abstract class BulkOperationBase {
         : typeof options === 'function'
         ? options
         : undefined;
+    options = options != null && typeof options !== 'function' ? options : {};
 
-    return maybeCallback(async () => {
-      options = options != null && typeof options !== 'function' ? options : {};
-
-      if (this.s.executed) {
+    if (this.s.executed) {
+      // eslint-disable-next-line @typescript-eslint/require-await
+      return maybeCallback(async () => {
         throw new MongoBatchReExecutionError();
-      }
+      }, callback);
+    }
 
-      const writeConcern = WriteConcern.fromOptions(options);
-      if (writeConcern) {
-        this.s.writeConcern = writeConcern;
-      }
+    const writeConcern = WriteConcern.fromOptions(options);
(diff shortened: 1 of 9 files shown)


src/utils.ts
```diff
@@ -436,6 +436,35 @@ export function* makeCounter(seed = 0): Generator<number> {
...
+  callback?: Callback<T>
+): Promise<T> | void {
+  const PromiseConstructor = PromiseProvider.get();
+
+  const promise = promiseFn();
+  if (callback == null && PromiseConstructor == null) {
+    return promise;
+  }
+
+  if (PromiseConstructor != null) {
```
baileympearson: We also need to check here that callback is not defined - otherwise, if a user defines a custom promise library, they can never use callbacks

src/promise_provider.ts
```diff
@@ -4,11 +4,11 @@ import { MongoInvalidArgumentError } from './error';
 const kPromise = Symbol('promise');
 
 interface PromiseStore {
-  [kPromise]?: PromiseConstructor;
+  [kPromise]: PromiseConstructor | null;
```
baileympearson, line 7: haven't reviewed this file yet - pending team discussion about breaking changes here

src/mongo_client.ts
```diff
@@ -593,8 +593,12 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
       }
     } catch (error) {
-      if (callback) return callback(error);
-      else return PromiseProvider.get().reject(error);
+      if (callback) {
+        return callback(error);
+      }
+
+      const PromiseConstructor = PromiseProvider.get() ?? Promise;
+      return PromiseConstructor.reject(error);
```
baileympearson, line 601: why not refactor the whole method to use maybeCallback? 

it would just look something like

```typescript
  const client = new MongoClient(url, options)
  return await client.connect();
}, callback);

```

src/bulk/common.ts
```diff
@@ -1346,13 +1346,13 @@ function handleEarlyError(
...
-  const Promise = PromiseProvider.get();
   if (typeof callback === 'function') {
     callback(err);
     return;
   }
 
-  return Promise.reject(err);
+  const PromiseConstructor = PromiseProvider.get() ?? Promise;
+  return PromiseConstructor.reject(err);
 }
```
baileympearson, line 1356: There are only two instances of handleErrorEarly in the entire repo, both in BulkOperationBase.execute().  How would you feel about just refactoring bulkOperationBase to use maybeCallback and getting rid of this helper entirely?

If we don't want to do that - why not just use maybeCallback?  This helper can be written like follows
```
  return maybeCallback(() => {
    throw err;
  }, callback);
```

src/mongo_client.ts
```diff
@@ -645,9 +649,9 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
     }
 
     const session = this.startSession(options);
```
baileympearson, line 651: again - can we use maybeCallback here?

baileympearson (changes requested):
General comment - I think we should use `maybeCallback` in cases we don't use maybePromise now, but use the promise provider.  We should encapsulate all knowledge of the promise provider into maybeCallback.

src/bulk/common.ts
nbbeeken: Sure refactor sounds good

src/gridfs/index.ts
```diff
@@ -144,28 +144,17 @@ export class GridFSBucket extends TypedEventEmitter<GridFSBucketEvents> {
...
-
-          return callback();
-        });
-      });
-    });
+    return maybeCallback(async () => {
+      const { deletedCount } = await this.s._filesCollection.deleteOne({ _id: id });
+      await this.s._chunksCollection.deleteMany({ files_id: id });
+
+      // Delete orphaned chunks before returning FileNotFound
```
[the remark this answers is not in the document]
nbbeeken: Yep, fixed

test/integration/node-specific/cursor_async_iterator.test.js
```diff
@@ -109,7 +109,7 @@ describe('Cursor Async Iterator Tests', function () {
 
     afterEach(() => {
       promiseSpy.restore();
-      PromiseProvider.set(Promise);
+      PromiseProvider.set(null);
```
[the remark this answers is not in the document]
nbbeeken: This is the afterEach hook, the beforeEach sets a custom promise

test/integration/node-specific/custom_promise_library.test.js
```diff
@@ -9,8 +9,22 @@ class CustomPromise extends Promise {}
...
 describe('Optional PromiseLibrary', function () {
+  beforeEach(() => {
+    PromiseProvider.set(null);
+  });
+
   afterEach(() => {
+    PromiseProvider.set(null);
+  });
+
+  it('should initially be set to null', () => {
```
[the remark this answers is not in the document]
nbbeeken: It's hard thing to test, upon loading the node process when our driver is first imported it should be set to null, but I'm not sure where I can put a test to make sure it's run "early", so it's intended to test the driver but it does always pass because of the beforeEach

test/unit/mongo_client.test.js
```diff
@@ -110,7 +110,6 @@ describe('MongoOptions', function () {
         return 'very unique';
       }
     },
-    promiseLibrary: global.Promise,
```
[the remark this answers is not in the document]
nbbeeken: See other comment, global side effect

src/utils.ts
nbbeeken: Fixed!

src/bulk/common.ts
```diff
@@ -1260,39 +1261,44 @@ export abstract class BulkOperationBase {
...
-    }
+    callback =
+      typeof callback === 'function'
+        ? callback
+        : typeof options === 'function'
+        ? options
+        : undefined;
+
+    return maybeCallback(async () => {
+      options = options != null && typeof options !== 'function' ? options : {};
```
baileympearson, line 1272: can this be moved out of `maybeCallback`?

test/integration/node-specific/custom_promise_library.test.js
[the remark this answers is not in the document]
baileympearson: Just move it into a place where there isn't a before each, maybe?

baileympearson (changes requested):
most comments addressed/resolved, some questions and suggestions remaining

src/mongo_client.ts
```diff
@@ -580,22 +581,18 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
-    }
+    callback =
+      typeof callback === 'function'
+        ? callback
+        : typeof options === 'function'
+        ? options
+        : undefined;
+
+    return maybeCallback(async () => {
+      options = typeof options !== 'function' ? options : undefined;
```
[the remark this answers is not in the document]
nbbeeken: the change to the type of options doesn't preserved into the callback cus it "could" change (it can't, but TS doesn't have a way of knowing that)

test/types/community/client.test-d.ts
```diff
@@ -35,7 +35,6 @@ const options: MongoClientOptions = {
   promoteBuffers: false,
   authMechanism: 'SCRAM-SHA-1',
   forceServerObjectId: false,
-  promiseLibrary: Promise,
```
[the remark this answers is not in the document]
nbbeeken: Oh sorry yea this one doesn't effect anything, the other one is the runtime usage. We have the explicit type test to check that promiseLibrary is deprecated, I think we can remove this from here now which is just a selection of MongoOptions. (one less search and replace)

test/unit/mongo_client.test.js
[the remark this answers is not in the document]
nbbeeken: Since now we expect the provider to have null by default, this is setting the global promise store, so the tests that assert the same promise from the API is ferried along fail

src/mongo_client.ts
nbbeeken: Sure, we needed an extra overload to support this, and I used null to make the usages distinct

src/bulk/common.ts
nbbeeken: Not without casting away a TS error, so this is preferable, see other reply

src/mongo_client.ts
```diff
@@ -642,16 +642,14 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
-      })
-      .finally(() => {
+
+    return maybeCallback(async () => {
+      try {
+        await withSessionCallback(session);
+      } finally {
         session.endSession().catch(() => null);
-      });
+      }
```
baileympearson, line 650: optional - we can remove the `.catch(() => null` in favor of try with an empty catch, like we discussed yesterday at team office

test/types/community/client.test-d.ts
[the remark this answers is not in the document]
baileympearson: leaving this thread open for context but I consider this resolved

baileympearson (changes requested):

src/mongo_client.ts
nbbeeken: Done!

baileympearson (changes requested):

src/bulk/common.ts
```diff
@@ -1270,11 +1270,19 @@ export abstract class BulkOperationBase {
...
+      typeof callback === 'function'
+        ? callback
+        : typeof options === 'function'
+        ? options
+        : undefined;
+    options = options != null && typeof options !== 'function' ? options : {};
 
     if (this.s.executed) {
-      return handleEarlyError(new MongoBatchReExecutionError(), callback);
+      // eslint-disable-next-line @typescript-eslint/require-await
```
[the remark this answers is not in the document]
nbbeeken: Wrapping the execOp in maybeCallback (naively, so that's likely why) messes with session allocation because it makes the maybePromise logic run. We can add a note to the refactor exec operation refactor ticket to comeback here, or I can revert the changes here and we can do it altogether.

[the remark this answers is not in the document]
baileympearson: We'll tackle this once execute_operation has been refactored

test/unit/utils.test.ts
```diff
@@ -642,4 +645,133 @@ describe('driver utils', function () {
       });
     });
   });
+
+  describe('maybeCallback()', () => {
+    it('should accept to two arguments', () => {
```
dariakp, line 650: "should accept two arguments"? or "up to two"?

test/unit/utils.test.ts
```diff
@@ -642,4 +645,133 @@ describe('driver utils', function () {
...
+        const superPromiseFailure = Promise.reject(new Error('ah!'));
+        const result = maybeCallback(() => superPromiseFailure);
+        expect(result).to.not.equal(superPromiseFailure);
+        expect(result).to.be.instanceOf(BluebirdPromise);
+        // @ts-expect-error: There is no overload to change the return type not be nullish,
+        // and we do not want to add one in fear of making it too easy to neglect adding the callback argument
+        expect(await result.catch(e => e)).to.have.property('message', 'ah!');
+      });
+
+      it('should return void event if a custom promise is set and a callback is provided', async () => {
```
dariakp, line 765:
```suggestion
      it('should return void even if a custom promise is set and a callback is provided', async () => {
```

dariakp (changes requested):

test/unit/utils.test.ts
nbbeeken: Sure I can correct this wording, however we should never call it without 2 arguments though, its just something we can't verify since there's no difference between a variable = undefined vs omitting the argument. We added a TS overload to allow null to kind of hint at this but we just have to rely on testing / double checking the code to make sure we pass through the callback variable if there is one.

nbbeeken: fixed!

dariakp (approved):

nbbeeken:
@baileympearson this is unblocked with the recent release. I've rebased so we can double check we're up to date on changes while we held off on this. (just an FYI no rush at all!)

baileympearson (approved):

PR #3407, opened 2022-09-12
fix(NODE-3144): pool clear event ordering and retryability tests

dariakp:
### Description
NODE-3144

#### What is changing?
- Fixed the ordering of PoolCleared event to happen before the corresponding CheckOutFailed events from waitqueue eviction
- Implemented retryable reads and writes prose tests
  - As part of this, I reorganized the existing retryable write prose test so that it would uniformly reset the failPoint with the new added test, but otherwise left it alone

No

Spec compliance

### DIFF
--- a/test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
+++ b/test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
@@ -1,0 +1,146 @@
+import { expect } from 'chai';
+
+import { Collection, MongoClient } from '../../../src';
+
+describe('Retryable Reads Spec Prose', () => {
+  let client: MongoClient, failPointName;
+
+  afterEach(async () => {
+    try {
+      if (failPointName) {
+        await client.db('admin').command({ configureFailPoint: failPointName, mode: 'off' });
+      }
+    } finally {
+      failPointName = undefined;
+      await client?.close();
+    }
+  });
+
+  describe('PoolClearedError Retryability Test', () => {
+    // This test will be used to ensure drivers properly retry after encountering PoolClearedErrors.
+    // It MUST be implemented by any driver that implements the CMAP specification.
+    // This test requires MongoDB 4.2.9+ for blockConnection support in the failpoint.
+
+    let observedEvents: Array<{ name: string; event: Record<string, any> }>;
+    let testCollection: Collection;
+    beforeEach(async function () {
+      // 1. Create a client with maxPoolSize=1 and retryReads=true.
+      // If testing against a sharded deployment, be sure to connect to only a single mongos. <-- TODO: what does that look like?
+      client = this.configuration.newClient({
+        maxPoolSize: 1,
+        retryReads: true,
+        monitorCommands: true
+      });
+      await client.connect();
+
+      testCollection = client.db('retryable-reads-prose').collection('pool-clear-retry');
+      await testCollection.drop().catch(() => null);
+      await testCollection.insertMany([{ test: 1 }, { test: 2 }]);
+
+      // 2. Enable the following failpoint:
+      // NOTE: "9. Disable the failpoint" is done in afterEach
+      failPointName = 'failCommand';
+      const failPoint = await client.db('admin').command({
+        configureFailPoint: failPointName,
+        mode: { times: 1 },
+        data: {
+          failCommands: ['find'],
+          errorCode: 91,
+          blockConnection: true,
+          blockTimeMS: 1000
+        }
+      });
+
+      expect(failPoint).to.have.property('ok', 1);
+
+      observedEvents = [];
+      for (const observedEvent of [
+        'connectionCheckOutStarted',
+        'connectionCheckedOut',
+        'connectionCheckOutFailed',
+        'connectionPoolCleared',
+        'commandStarted'
+      ]) {
+        client.on(observedEvent, ev => {
+          observedEvents.push({ name: observedEvent, event: ev });
+        });
+      }
+    });
+
+    it('should emit events in the expected sequence', {
+      metadata: { requires: { mongodb: '>=4.2.9' } },
+      test: async function () {
+        // 3. Start two threads and attempt to perform a findOne simultaneously on both.
+        const results = await Promise.all([
+          testCollection.findOne({ test: 1 }),
+          testCollection.findOne({ test: 2 })
+        ]);
+
+        client.removeAllListeners();
+        // 4. Verify that both findOne attempts succeed.
+        expect(results[0]).to.have.property('test', 1);
+        expect(results[1]).to.have.property('test', 2);
+
+        // 5. Via CMAP monitoring, assert that the first check out succeeds.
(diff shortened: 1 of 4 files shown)


test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
```diff
@@ -0,0 +1,146 @@
...
+          testCollection.findOne({ test: 1 }),
+          testCollection.findOne({ test: 2 })
+        ]);
+
+        client.removeAllListeners();
+        // 4. Verify that both findOne attempts succeed.
+        expect(results[0]).to.have.property('test', 1);
+        expect(results[1]).to.have.property('test', 2);
+
+        // 5. Via CMAP monitoring, assert that the first check out succeeds.
```
durran, line 84: For this sequence I found if the observed events were split into 2 arrays for cmap events and command events, I could reduce all the code for asserting the events down to this:

```js
// 5. Via CMAP monitoring, assert that the first check out succeeds.
expect(cmapEvents.shift().name).to.equal('connectionCheckOutStarted', 'expected 1) checkout 1 to start');
expect(cmapEvents.shift().name).to.equal('connectionCheckOutStarted', 'expected 2) checkout 2 to start');
expect(cmapEvents.shift().name).to.equal('connectionCheckedOut', 'expected 3) first checkout to succeed');
// 6. Via CMAP monitoring, assert that a PoolClearedEvent is then emitted.
expect(cmapEvents.shift().name).to.equal('connectionPoolCleared', 'expected 4) pool to clear');
// 7. Via CMAP monitoring, assert that the second check out then fails due
// to a connection error.
const nextEvent = cmapEvents.shift();
expect(nextEvent.name).to.equal('connectionCheckOutFailed', 'expected 5) checkout 2 to fail');
expect(nextEvent.event).to.have.property('reason', 'connectionError');

// 8. Via Command Monitoring, assert that exactly three find CommandStartedEvents were observed in total.
expect(commandEvents).to.have.lengthOf(3, 'expected 3 find command started events');
```

It's obviously less code but curious if there was a specific reason for doing the find by index method?

test/integration/retryable-writes/retryable_writes.spec.prose.test.ts
```diff
@@ -1,64 +1,200 @@
...
+        ]);
+
+        client.removeAllListeners();
+        // iv. Verify that both insertOne attempts succeed.
+        const result = await testCollection.find().toArray();
+        expect(result).to.have.lengthOf(2);
+        const mappedAndSortedResult = result.map(item => item.test).sort();
+        expect(mappedAndSortedResult).to.deep.equal([1, 2]);
+
+        // v. Via CMAP monitoring, assert that the first check out succeeds.
```
durran, line 138: Same comment for here as for retryable reads.

durran (changes requested):

test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    let testCollection: Collection;
+    beforeEach(async function () {
+      // 1. Create a client with maxPoolSize=1 and retryReads=true.
+      // If testing against a sharded deployment, be sure to connect to only a single mongos. <-- TODO: what does that look like?
+      client = this.configuration.newClient({
+        maxPoolSize: 1,
+        retryReads: true,
+        monitorCommands: true
+      });
+      await client.connect();
```
durran, line 34: We could remove this since the drop call will do the connection. But also fine leaving it.

test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
dariakp: My original reasoning for using indices was because I did not want to make assumptions about any other monitoring/pool events that may or may not happen in between the events of interest, but I suppose since we explicitly opt into listening for explicit events, asserting on the exact sequence of the tracked events is relatively low risk for us even if the future implementation of cmap or monitoring changes to emit additional events. What are your thoughts?

test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
dariakp: oh good point, I forgot about auto connect

test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
```diff
@@ -0,0 +1,136 @@
...
+          failCommands: ['find'],
+          errorCode: 91,
+          blockConnection: true,
+          blockTimeMS: 1000
+        }
+      });
+
+      expect(failPoint).to.have.property('ok', 1);
+
+      cmapEvents = [];
```
durran, line 56:
```suggestion
      cmapEvents = [];
      commandStartedEvents = [];
```

test/integration/retryable-writes/retryable_writes.spec.prose.test.ts
```diff
@@ -1,64 +1,189 @@
...
+          errorCode: 91,
+          blockConnection: true,
+          blockTimeMS: 1000,
+          errorLabels: ['RetryableWriteError']
+        }
+      });
+
+      expect(failPoint).to.have.property('ok', 1);
+
+      cmapEvents = [];
```
durran, line 108:
```suggestion
      cmapEvents = [];
      commandStartedEvents = [];
```

durran (approved):

test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    // This test will be used to ensure drivers properly retry after encountering PoolClearedErrors.
+    // It MUST be implemented by any driver that implements the CMAP specification.
+    // This test requires MongoDB 4.2.9+ for blockConnection support in the failpoint.
+
+    let cmapEvents: Array<{ name: string; event: Record<string, any> }>;
+    let commandStartedEvents: Array<Record<string, any>>;
+    let testCollection: Collection;
+    beforeEach(async function () {
+      // 1. Create a client with maxPoolSize=1 and retryReads=true.
+      // If testing against a sharded deployment, be sure to connect to only a single mongos. <-- TODO: what does that look like?
```
nbbeeken, line 30: Optional, but you can use the uri helper passed to newClient:
```ts
client = this.configuration.newClient(this.configuration.uri({
        maxPoolSize: 1,
        retryReads: true,
        monitorCommands: true,
        useMultipleMongoses: false // If testing against a sharded deployment, be sure to connect to only a single mongos.
}))
```

if you want to lock in the single mongos behavior rather than depending on the default

nbbeeken (dismissed):

test/integration/retryable-reads/retryable_reads.spec.prose.test.ts
baileympearson:
```suggestion
      // If testing against a sharded deployment, be sure to connect to only a single mongos.
```

baileympearson (approved):

PR #3408, opened 2022-09-13
feat(NODE-4634): add support for bulk FindOperators.hint()

addaleax:
#### What is changing?

- `FindOperators.update()` and `.updateOne()` use updated TS definitions to allow pipeline-style updates (which are supported by the server and by the actual implementation itself)
- `FindOperators.hint()` is added as an analogue to the `.collation()`, `.arrayHints()` and `.upsert()` methods
- Tests for both of the above are added, modeled on existing tests

I would assume so.

mongosh cannot implement `.hint()` in its own bulk write builder API without one of:

- Modifying driver internals, or
- Implementing the entire bulk builder API itself on top of `collection.bulkWrite()`, or
- This change

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -869,7 +869,7 @@ export class FindOperators {
     return this;
   }
 
-  /** Specifies hint for UpdateOne or UpdateMany bulk operations. */
+  /** Specifies hint for the bulk operation. */
   hint(hint: Hint): this {
     if (!this.bulkOperation.s.currentOp) {
       this.bulkOperation.s.currentOp = {};
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -784,7 +784,7 @@ export class FindOperators {
   }
 
   /** Add a multiple update operation to the bulk operation */
-  update(updateDocument: Document): BulkOperationBase {
+  update(updateDocument: Document | Document[]): BulkOperationBase {
     const currentOp = buildCurrentOp(this.bulkOperation);
     return this.bulkOperation.addToOperationsList(
       BatchType.UPDATE,
@@ -796,7 +796,7 @@ export class FindOperators {
   }
 
   /** Add a single update operation to the bulk operation */
-  updateOne(updateDocument: Document): BulkOperationBase {
+  updateOne(updateDocument: Document | Document[]): BulkOperationBase {
     if (!hasAtomicOperators(updateDocument)) {
       throw new MongoInvalidArgumentError('Update document requires atomic operators');
     }
@@ -866,6 +866,16 @@ export class FindOperators {
     }
 
     this.bulkOperation.s.currentOp.arrayFilters = arrayFilters;
+    return this;
+  }
+
+  /** Specifies hint for UpdateOne or UpdateMany bulk operations. */
+  hint(hint: Hint): this {
+    if (!this.bulkOperation.s.currentOp) {
+      this.bulkOperation.s.currentOp = {};
+    }
+
+    this.bulkOperation.s.currentOp.hint = hint;
     return this;
   }
 }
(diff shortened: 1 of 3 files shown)


test/integration/crud/bulk.test.ts
```diff
@@ -1688,6 +1688,55 @@ describe('Bulk', function () {
...
+        });
+        expect(events[1]).property('commandName').to.equal('delete');
+        const deleteCommand = events[1].command;
+        expect(deleteCommand).property('deletes').to.be.an('array').with.length(2);
+        deleteCommand.deletes.forEach(statement => {
+          expect(statement).property('hint').to.eql({ b: 1 });
+        });
+      } finally {
+        await client.close();
+      }
```
nbbeeken, line 1736: There's an afterEach hook that handles closing the client for each of these tests

test/integration/crud/bulk.test.ts
```diff
@@ -1737,6 +1786,28 @@ describe('Bulk', function () {
...
+      bulk.find({ a: 2 }).update([{ $project: { a: { $add: ['$a', 100] } } }]);
+
+      await bulk.execute();
+
+      try {
+        const contents = await coll.find().project({ _id: 0 }).toArray();
+        expect(contents).to.deep.equal([{ a: 11 }, { a: 102 }]);
+      } finally {
+        await client.close();
+      }
```
nbbeeken, line 1807: afterEach handles this

test/integration/crud/bulk.test.ts
```diff
@@ -1688,6 +1688,55 @@ describe('Bulk', function () {
     }
   });
 
+  it('should apply hint via FindOperators', {
+    metadata: { requires: { mongodb: '>= 4.2' } },
```
nbbeeken, line 1692: I see failures on 4.2 for this, should it just be "greater than"?

src/bulk/common.ts
```diff
@@ -868,6 +868,16 @@ export class FindOperators {
     this.bulkOperation.s.currentOp.arrayFilters = arrayFilters;
     return this;
   }
+
+  /** Specifies hint for UpdateOne or UpdateMany bulk operations. */
```
nbbeeken, line 872: Hint is used on deletes and replaces too (IIUC), maybe let's just be generic here and say it specifies the hint for the bulk operation

test/integration/crud/bulk.test.ts
addaleax: Done, and ditto for the other added test.

Nearly every test in this file explicitly closes the client, though. Happy to do a follow-up PR where this is aligned across the file.

test/integration/crud/bulk.test.ts
addaleax: I was going by the version number mentioned on https://www.mongodb.com/docs/manual/reference/method/Bulk.find.hint/, but didn’t realize that at least for deletes at least 4.3.4 is required. I’m good with just bumping the entire test to `>= 4.4`, but also happy to split it if that’s better (there are also spec tests for `bulkWrite()` that cover the different server version ranges).

src/bulk/common.ts
addaleax: Sure, done!

test/integration/crud/bulk.test.ts
nbbeeken: This file used to explicitly connect at the top of each test so that's why there's a close to mirror it, we cleaned up the connects but not closes, no harm in closing more than once but we don't need to in new tests. 

That'd be nice :) but probably not worth the time

test/integration/crud/bulk.test.ts
nbbeeken: I'm happy with 4.4+ coverage

nbbeeken (approved):

baileympearson (approved):

PR #3409, opened 2022-09-13
fix(NODE-3541): improve error message

LinusU:
#### What is changing?

The error message when trying to to an `update` op with both a `hint` and a write concern with `w: 0`.

I don't believe so.

The previous error message led the user to believe that the server was too old, even when that was in fact not the case. The code now only gives the old message when the server is old, and gives a new message if that is not the case.

We encountered this error today and it was hard to track down the actual source.

- [x] PR title follows the correct format: `<type>(NODE-xxxx)<!>: <description>` (**not sure if this is correct since it's only a step in the right direction, doesn't close the entire bug**)
- [x] New TODOs have a related JIRA ticket (**not new but points to an existing ticket**)


src/operations/update.ts
```diff
@@ -125,7 +125,16 @@ export class UpdateOperation extends CommandOperation<Document> {
     const unacknowledgedWrite = this.writeConcern && this.writeConcern.w === 0;
     if (unacknowledgedWrite || maxWireVersion(server) < 5) {
```
baileympearson, line 126:
```suggestion
    if (unacknowledgedWrite) {
```

src/operations/update.ts
```diff
@@ -125,7 +125,16 @@ export class UpdateOperation extends CommandOperation<Document> {
...
+        if (maxWireVersion(server) < 5) {
+          callback(new MongoCompatibilityError(`Servers < 3.4 do not support hint on update`));
+        } else {
+          // TODO: https://jira.mongodb.org/browse/NODE-3541
+          callback(
+            new MongoCompatibilityError(
+              `This Node.js driver do not support hint together with unacknowledged writes`
+            )
+          );
+        }
```
baileympearson, line 137:
```suggestion
          // TODO(NODE-3541): support unacknowledged writes in CRUD operations
          callback(
            new MongoCompatibilityError(
              `This Node.js driver does not support hint together with unacknowledged writes`
            )
          );
```

baileympearson (changes requested):
Hey @LinusU - thanks for submitting this change.  I think this is an easy win.  I left two suggestions.  Wire version 5 corresponds to server 3.4, which is below our minimum supported server version, so we can remove the check altogether 😄 

Also - thanks for the TODO comment!  I suggested a reformat of the TODO so that it aligns with our convention (`TODO(<jira ticket number): <description`).

If you aren't able to make these changes, let me know.  I'm happy to take over the work, but if you'd like to make the changes, go for it 🙂

LinusU:
@baileympearson thank you for reviewing, with the information you provided I managed to implement the entire feature instead of just updating the error message!

👉 https://github.com/mongodb/node-mongodb-native/pull/3415

baileympearson (changes requested):
can we fix the error message in delete as well?

baileympearson:
Hey @LinusU - are you able to make the changes I requested?  If not, I'm happy to take over the PR from here if that's okay with you.

baileympearson (approved):

dariakp (approved):

durran (approved):

LinusU:
@baileympearson sorry for not getting back to you earlier, and thank you so much for landing this PR! ❤️

PR #3410, opened 2022-09-13
fix(NODE-4621): ipv6 address handling in HostAddress

nbbeeken:
### Description

NODE-4621

#### What is changing?

Always include brackets when getting the whole hostAddress string.

- Ensure `Connection` instances have the correct formatted address for ipv6
- HostAddress.toString no longer accepts a flag for disabling IPv6 brackets (they should never have been created without brackets)
- Wrap the error thrown from `new URL` it was very unclear only printing: `TypeError: Invalid URL` prints the input hostname now and attaches the original error as a cause.
- Fixed up a cursor test while debugging it, I can revert, it just happened to demonstrate an issue with bracketing
- fixed type of `this.configuration.options.hostAddresses` to not be nullish
- Added the --ipv6 flag to our mtools script
- cmap spec runner improvement, print the original error stack and name, otherwise its hard to trace
- fixed uri_spec_runner to handle new error wrapping logic, I can remove the TypeError handling now, thoughts?
- Added unit tests to connection_string to demonstrate what does and does not work with IPv6 addresses
- **Notably:** Added IPv6 integration tests that only run on **Windows** and **ReplicaSet** configs.
  - This is because Ubuntu 18.04 (our current default OS) does **not** have a hostname lookup for `localhost` to `::1` by default, Windows, does, and on macos we don't run replica_sets 

When a port exists (which always does for us) you must include brackets for the whole host address

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1142,44 +1142,55 @@ export class BufferPool {
 
 /** @public */
 export class HostAddress {
-  host;
-  port;
+  host: string | undefined;
+  port: number | undefined;
   // Driver only works with unix socket path to connect
   // SDAM operates only on tcp addresses
-  socketPath;
-  isIPv6;
+  socketPath: string | undefined;
+  isIPv6 = false;
 
   constructor(hostString: string) {
     const escapedHost = hostString.split(' ').join('%20'); // escape spaces, for socket path hosts
-    const { hostname, port } = new URL(`mongodb://${escapedHost}`);
 
     if (escapedHost.endsWith('.sock')) {
       // heuristically determine if we're working with a domain socket
       this.socketPath = decodeURIComponent(escapedHost);
-    } else if (typeof hostname === 'string') {
-      this.isIPv6 = false;
+      delete this.port;
+      delete this.host;
+      return;
+    }
 
-      let normalized = decodeURIComponent(hostname).toLowerCase();
-      if (normalized.startsWith('[') && normalized.endsWith(']')) {
-        this.isIPv6 = true;
-        normalized = normalized.substring(1, hostname.length - 1);
-      }
+    const urlString = `iLoveJS://${escapedHost}`;
+    let url;
+    try {
+      url = new URL(urlString);
+    } catch (urlError) {
+      const runtimeError = new MongoRuntimeError(`Unable to parse ${escapedHost} with URL`);
+      runtimeError.cause = urlError;
+      throw runtimeError;
+    }
 
-      this.host = normalized.toLowerCase();
+    const hostname = url.hostname;
+    const port = url.port;
 
-      if (typeof port === 'number') {
-        this.port = port;
-      } else if (typeof port === 'string' && port !== '') {
-        this.port = Number.parseInt(port, 10);
-      } else {
-        this.port = 27017;
-      }
+    let normalized = decodeURIComponent(hostname).toLowerCase();
+    if (normalized.startsWith('[') && normalized.endsWith(']')) {
+      this.isIPv6 = true;
+      normalized = normalized.substring(1, hostname.length - 1);
+    }
 
-      if (this.port === 0) {
-        throw new MongoParseError('Invalid port (zero) with hostname');
-      }
+    this.host = normalized.toLowerCase();
+
+    if (typeof port === 'number') {
+      this.port = port;
+    } else if (typeof port === 'string' && port !== '') {
+      this.port = Number.parseInt(port, 10);
     } else {
-      throw new MongoInvalidArgumentError('Either socketPath or host must be defined.');
+      this.port = 27017;
+    }
+
+    if (this.port === 0) {
+      throw new MongoParseError('Invalid port (zero) with hostname');
     }
     Object.freeze(this);
   }
@@ -1189,15 +1200,12 @@ export class HostAddress {
   }
 
   inspect(): string {
-    return `new HostAddress('${this.toString(true)}')`;
+    return `new HostAddress('${this.toString()}')`;
   }
 
-  /**
-   * @param ipv6Brackets - optionally request ipv6 bracket notation required for connection strings
-   */
-  toString(ipv6Brackets = false): string {
+  toString(): string {
     if (typeof this.host === 'string') {
-      if (this.isIPv6 && ipv6Brackets) {
+      if (this.isIPv6) {
         return `[${this.host}]:${this.port}`;
       }
       return `${this.host}:${this.port}`;
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1142,11 +1142,9 @@ export class BufferPool {
 
 /** @public */
 export class HostAddress {
-  host: string | undefined;
-  port: number | undefined;
-  // Driver only works with unix socket path to connect
-  // SDAM operates only on tcp addresses
-  socketPath: string | undefined;
+  host: string | undefined = undefined;
+  port: number | undefined = undefined;
+  socketPath: string | undefined = undefined;
   isIPv6 = false;
 
   constructor(hostString: string) {
@@ -1155,8 +1153,6 @@ export class HostAddress {
     if (escapedHost.endsWith('.sock')) {
       // heuristically determine if we're working with a domain socket
       this.socketPath = decodeURIComponent(escapedHost);
-      delete this.port;
-      delete this.host;
       return;
     }
 
--- a/test/integration/node-specific/ipv6.test.ts
+++ b/test/integration/node-specific/ipv6.test.ts
@@ -41,7 +41,8 @@ describe('IPv6 Addresses', () => {
     const ipv6LocalhostAddresses = this.configuration.options.hostAddresses.map(({ port }) => ({
       host: '::1',
       port,
-      isIPv6: true
+      isIPv6: true,
+      socketPath: undefined
     }));
     expect(client.options.hosts).to.deep.equal(ipv6LocalhostAddresses);
   });
--- a/test/integration/node-specific/ipv6.test.ts
+++ b/test/integration/node-specific/ipv6.test.ts
@@ -12,11 +12,11 @@ describe('IPv6 Addresses', () => {
 
   beforeEach(async function () {
     if (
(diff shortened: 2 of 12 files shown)


addaleax (approved):

src/utils.ts
```diff
@@ -1142,44 +1142,55 @@ export class BufferPool {
...
     const escapedHost = hostString.split(' ').join('%20'); // escape spaces, for socket path hosts
-    const { hostname, port } = new URL(`mongodb://${escapedHost}`);
 
     if (escapedHost.endsWith('.sock')) {
       // heuristically determine if we're working with a domain socket
       this.socketPath = decodeURIComponent(escapedHost);
-    } else if (typeof hostname === 'string') {
-      this.isIPv6 = false;
+      delete this.port;
+      delete this.host;
```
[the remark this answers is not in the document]
nbbeeken: I should've called this out, I'm being overly defensive here, newer `"target"` settings of typescript will make these properties exist because it'll emit JS that support class property syntax, and we have assertions that say these properties should not exist. This is not necessary now though, and/or we could also relax the tests.

test/integration/node-specific/ipv6.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      host: '::1',
+      port,
+      isIPv6: true
+    }));
+    expect(client.options.hosts).to.deep.equal(ipv6LocalhostAddresses);
+  });
+
+  it('should successfully connect using IPv6 loopback addresses', async function () {
+    const localhostHosts = this.configuration.options.hostAddresses.map(
+      ({ port }) => `localhost:${port}` // ::1 will be swapped out for localhost
```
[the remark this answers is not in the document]
nbbeeken: Yes that's correct, and in the case of our CI set up we turn on the IPv6 flag and we bind to `127.0.0.1,::1` ([that's setup here](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/orchestration/configs/replica_sets/auth.json#L9)). So it seems the server will use the localhost hostname in this scenario. 

However, I used you reproduction from the other ticket relating to replicaset connection issues and it appears that if you only pass `::1` as the bind_ip setting than the server will not use `localhost`. So the following is from a script that performs and insertOne & find

```js
var connectionString = "mongodb://bob:pwd@[::1]:27017,[::1]:27018,[::1]:27019/?directConnection=false&replicaSet=rs"
```

```js
{ hello: { ok: 1, me: '[::1]:27017', hosts: [ '[::1]:27017', '[::1]:27018', '[::1]:27019' ], primary: '[::1]:27017', secondary: false, setName: 'rs', isWritablePrimary: true } }
{ hello: { ok: 1, me: '[::1]:27018', hosts: [ '[::1]:27017', '[::1]:27018', '[::1]:27019' ], primary: '[::1]:27017', secondary: true, setName: 'rs', isWritablePrimary: undefined } }
{ hello: { ok: 1, me: '[::1]:27019', hosts: [ '[::1]:27017', '[::1]:27018', '[::1]:27019' ], primary: '[::1]:27017', secondary: true, setName: 'rs', isWritablePrimary: undefined } }
{ hello: { ok: 1, me: '[::1]:27017', hosts: [ '[::1]:27017', '[::1]:27018', '[::1]:27019' ], primary: '[::1]:27017', secondary: false, setName: 'rs', isWritablePrimary: true } }
```

src/utils.ts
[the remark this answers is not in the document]
addaleax: Interesting! I knew that this was happening for initialized class properties, but not uninitialized ones.

In any case, I'll resolve this, but it's interesting because `delete obj.property` has traditionally been (and I think still is) a performance cliff in V8 because it changes the shape of the object after the fact.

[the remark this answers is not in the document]
nbbeeken: https://www.typescriptlang.org/tsconfig#useDefineForClassFields
So it will start with ES2022 unless you flip this config. Perf cliff sounds like something worth fixing esp since this constructor is called quite frequently, I will adjust 😄

test/tools/uri_spec_runner.ts
```diff
@@ -71,6 +71,8 @@ export function executeUriValidationTest(
     } catch (err) {
       if (err instanceof TypeError) {
         expect(err).to.have.property('code').equal('ERR_INVALID_URL');
+      } else if (err instanceof MongoRuntimeError) {
```
dariakp, line 74: if this is meant to be a replacement for the `TypeError` check, then we should remove it; are there still instances where we'd get a TypeError instead?

test/tools/cluster_setup.sh
```diff
@@ -13,15 +13,15 @@ SHARDED_DIR=${SHARDED_DIR:-$DATA_DIR/sharded_cluster}
 
 if [[ $1 == "replica_set" ]]; then
     mkdir -p $REPLICASET_DIR # user / password
-    mlaunch init --dir $REPLICASET_DIR --auth --username "bob" --password "pwd123" --replicaset --nodes 3 --arbiter --name rs --port 31000 --enableMajorityReadConcern --setParameter enableTestCommands=1
+    mlaunch init --dir $REPLICASET_DIR --ipv6 --auth --username "bob" --password "pwd123" --replicaset --nodes 3 --arbiter --name rs --port 31000 --enableMajorityReadConcern --setParameter enableTestCommands=1
```
dariakp, line 16: is this enough to make this script work well on node 18 now?

test/unit/connection_string.test.ts
```diff
@@ -573,4 +574,51 @@ describe('Connection String', function () {
...
+    it('should allow multiple bracketed portless localhost IPv6 addresses', () => {
+      const client = new MongoClient('mongodb://[::1],[::1],[::1]/test');
+      expect(client.options.hosts).to.deep.equal([
+        { host: '::1', port: 27017, isIPv6: true, socketPath: undefined },
+        { host: '::1', port: 27017, isIPv6: true, socketPath: undefined },
+        { host: '::1', port: 27017, isIPv6: true, socketPath: undefined }
+      ]);
+    });
+
+    it('should allow multiple bracketed portless localhost IPv6 addresses', () => {
```
dariakp, line 604: this test case title is the same as the test case above

test/unit/connection_string.test.ts
```diff
@@ -573,4 +574,51 @@ describe('Connection String', function () {
...
+        'mongodb://[ABCD:f::abcd:abcd:abcd:abcd],[ABCD:f::abcd:abcd:abcd:abcd],[ABCD:f::abcd:abcd:abcd:abcd]/test'
+      );
+      expect(client.options.hosts).to.deep.equal([
+        { host: 'abcd:f::abcd:abcd:abcd:abcd', port: 27017, isIPv6: true, socketPath: undefined },
+        { host: 'abcd:f::abcd:abcd:abcd:abcd', port: 27017, isIPv6: true, socketPath: undefined },
+        { host: 'abcd:f::abcd:abcd:abcd:abcd', port: 27017, isIPv6: true, socketPath: undefined }
+      ]);
+    });
+
+    it('should allow multiple bracketed port-full IPv6 addresses', () => {
```
dariakp, line 615: -> `should allow multiple bracketed IPv6 addresses with specified ports`

test/integration/node-specific/ipv6.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  beforeEach(async function () {
+    if (
+      process.platform !== 'win32' ||
+      this.configuration.topologyType !== TopologyType.ReplicaSetWithPrimary
+    ) {
+      if (this.currentTest) {
+        // Ubuntu 18 does not support localhost AAAA lookups (IPv6)
+        // Windows (VS2019) has the AAAA lookup
+        // We do not run a replica set on macos
```
dariakp, line 21: if we don't run replica sets on macos, then this would already be skipped on macos by the topologyType check, no? I think it might make more sense to exclude the platform that does not support macos AAAA lookups than to selectively only include windows because that's the one that has the replica set and the support. Or am I missing something?

test/integration/node-specific/ipv6.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  it('should successfully connect using IPv6 loopback addresses', async function () {
+    const localhostHosts = this.configuration.options.hostAddresses.map(
+      ({ port }) => `localhost:${port}` // ::1 will be swapped out for localhost
+    );
+    await client.db().command({ ping: 1 });
+    // After running the first command we should receive the hosts back as reported by the mongod in a hello response
+    // mongodb will report the bound host address, in this case "localhost"
+    expect(client.topology).to.exist;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
dariakp, line 58: is it time to add the no-non-null-assertion to the test ts config? (not in this PR, necessarily, but it's going to be an issue for all tests where we assert existence via `expect`)

dariakp (changes requested):

test/integration/node-specific/ipv6.test.ts
nbbeeken: Nope that's correct, this would work on MacOS the following tests just depend on being in a replica set. Changed!

test/integration/node-specific/ipv6.test.ts
nbbeeken: Yea we can do that, if we add it the linter should remove all the execeptions we have added (this included) so I can put that up seperately

test/tools/cluster_setup.sh
nbbeeken: Yep, in node 18 "localhost" will translate to "::1" which the mongod's will now bind to with this flag

test/tools/uri_spec_runner.ts
nbbeeken: I don't think so, fixed!

test/unit/connection_string.test.ts
nbbeeken: updated: `remote`

dariakp (approved):

PR #3411, opened 2022-09-14
refactor(NODE-4637): clean up async interval

baileympearson:
#### What is changing?

This PR does a bit of cleanup on the async interval using in SDAM monitoring.  Specifically it

- Moves the async interval into the monitor.ts file from the utils
- Move the associated unit tests 
- Converts monitor.test.js to Typescript.  This was necessary because the existing interval tests were in Typescript, so I could either convert them to JS or convert monitor.test.js to TS.  I chose the latter.
- converts the `makeInterruptibleAsyncInterval` factory function into a class
- renames some variables to more closely align with their actual purpose and renames the interval `SDAMMonitorInterval` (name can be workshopped).

To aid in review, moved the interval to monitor.ts (and the associated tests) in the first commit, and then tried to be granular with the changes from then on.  My intention was to put all the code reorganization changes in to the first commit, so the actual code changes that I made can be easily viewed in a diff.

No.

General code cleanup and better code organization.


src/sdam/topology.ts
```diff
@@ -386,7 +386,9 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
   }
```
baileympearson, line 387: The missing overloads in this file were revealed when I converted the monitor tests to TS.  I'm happy to revert these changes and cast in the test file instead.

nbbeeken: It's internal I think it's fine.

nbbeeken (approved):
LGTM, barring some bikeshedding on slack

nbbeeken (approved):

PR #3413, opened 2022-09-14
refactor(NODE-4638): prevent parallel calls to `fn` in the async interval

baileympearson:
#### What is changing?

The monitor has been refactored to more closely align the spec.  Additionally, this PR also prevents the monitor from re-executing the `fn` while `fn` is currently executing.

The unit tests have been reworked to match the new behavior.  The passing unit tests have the following output (included to help demonstrate the behavior of the monitor):

Screen Shot 2022-09-29 at 1 16 30 PM

No.

### DIFF
--- a/src/sdam/monitor.ts
+++ b/src/sdam/monitor.ts
@@ -487,6 +487,8 @@ export class MonitorInterval {
   lastCallTime: number;
   isExpeditedCheckScheduled = false;
   stopped = false;
+  isExecutionInProgress = false;
+  hasExecutedOnce = false;
 
   heartbeatFrequencyMS: number;
   minHeartbeatFrequencyMS: number;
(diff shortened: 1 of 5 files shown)


src/sdam/monitor.ts
```diff
@@ -484,61 +478,57 @@ export interface MonitorIntervalOptions {
...
-
-    // This is possible in virtualized environments like AWS Lambda where our
-    // clock is unreliable. In these cases the timer is "running" but never
-    // actually completes, so we want to execute immediately and then attempt
-    // to reschedule.
-    if (timeUntilNextCall < 0) {
+    const currentTime = now();
+    const timeSinceLastCall = currentTime - this.lastExecutionEnded;
+
+    if (!this.hasExecutedOnce) {
```
dariakp, line 509: 1) why do we specifically need this `hasExecutedOnce` property? won't we already hit the executeAndReschedule logic at the bottom of this function if this is the very first time it's called? I don't see any early returns that would prevent it.

2) Do we have a test case for unreliable clocks where we get a timeSinceLastCall is negative? We'd end up rescheduling for an interval greater than `minHeartbeatFrequencyMS`, which isn't what we want. Or are we confident we don't want to do anything about that case?

baileympearson: @dariakp 

1. If the function is woken before the function executes but is woken less than minHeartbeatFrequencyMS after it was initialized, we want to execute immediately.  If we remove this check, we'd fall through to `timeSinceLastCall < minHeartbeatFrequencyMS` and reschedule the call instead of executing it immediately.
2. Since the logic has been flipped, I don't think this is relevant anymore.  The `unreliable clocks` doesn't refer to `now`, it refers to `setTimeout` (and node timers) being unreliable.  Previously, it was possible that the timer didn't fire and consequently, `timeUntilNextCall` would be negative.  We now keep track of the time since last call, which will never be negative.  Every time this function is woken, `currentTime` will be calculated and will always be >= the last call time.

src/sdam/monitor.ts
```diff
@@ -484,61 +478,55 @@ export interface MonitorIntervalOptions {
...
-    // This is possible in virtualized environments like AWS Lambda where our
-    // clock is unreliable. In these cases the timer is "running" but never
-    // actually completes, so we want to execute immediately and then attempt
-    // to reschedule.
-    if (timeUntilNextCall < 0) {
-      this._executeAndReschedule();
+    const currentTime = now();
+    const timeSinceLastCall = currentTime - this.lastExecutionEnded;
+
+    if (timeSinceLastCall < 0) {
```
dariakp, line 508: just needs the TODO we discussed

dariakp (approved):

nbbeeken (approved):

PR #3415, opened 2022-09-15
fix(NODE-3541): allow unacknowledged write with hint

LinusU:
### Description

This is a replacement to #3409, from the knowledge I got there I could implement this properly instead of just improving the error message!

Now it actually fixes the entire ticket 🎉 

#### What is changing?

The driver now allows unacknowledged writes when a hint is present in the following scenarios:

- In `update` calls when the server is 4.2 or newer
- In `delete` calls when the server is 4.4 or newer

Not as far as I know, because I couldn't find any documentation on this limit in the first place.

This allows us to submit a hint to an unacknowledged write. See more background in [NODE-3541](https://jira.mongodb.org/browse/NODE-3541).

- [ ] Changes are covered by tests **see below**

-----

I also need to update the tests here, but since this is my first contribution here I would love some guidance.

What I think we need to do is to update the `test/spec/crud/unified/unacknowledged-*-hint-clientError.yml` files and make them only test for the error on specific server versions, but expect success on others. If anyone could point me in the right direction here I would be grateful. Should we also rename the files to remove "clientError" from the name? Actually, the ticket mentions syncing unified CRUD tests, that might be what we should instead?

ping @baileympearson, thank you for reviewing my first PR, and giving me info to continue this!

### DIFF
--- a/src/operations/delete.ts
+++ b/src/operations/delete.ts
@@ -90,10 +90,14 @@ export class DeleteOperation extends CommandOperation<Document> {
         : undefined;
     }
 
-    const unacknowledgedWrite = this.writeConcern && this.writeConcern.w === 0;
-    if (unacknowledgedWrite || maxWireVersion(server) < 5) {
-      if (this.statements.find((o: Document) => o.hint)) {
-        callback(new MongoCompatibilityError(`Servers < 3.4 do not support hint on delete`));
+    if (maxWireVersion(server) < 9) {
+      const hintPresent = this.statements.some(o => o.hint);
+      const unacknowledgedWrite = this.writeConcern && this.writeConcern.w === 0;
+
+      if (hintPresent && unacknowledgedWrite) {
+        callback(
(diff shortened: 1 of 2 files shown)


baileympearson:
Hey @LinusU - thanks for this contribution.  There are still some open questions about this ticket and at this time, it's not ready to be implemented.  Our team will need to discuss the ticket in refinement to iron out some of the details.

We don't have the bandwidth to pick this up at the moment but we will put the ticket in our triage queue to revisit it when we get the chance and when we do implement it, we'll make sure you get credit for your contribution as well.

I think we can move forward with https://github.com/mongodb/node-mongodb-native/pull/3409 in the meantime though!

LinusU:
I understand, let me know if there is anything you want me to do here 👍

baileympearson:
Fixed in https://github.com/mongodb/node-mongodb-native/pull/4647

PR #3416, opened 2022-09-15
chore(NODE-4641): remove wire version checks for unsupported servers

LinusU:
### Description

As a follow up to #3409, this removes all wire version checks which check for a version that is no longer supported by this driver.

#### What is changing?

Code that wasn't used is being removed.

No

My motivation behind this was to clean up the code, and remove unnecessary checks.

- [x] PR title follows the correct format: `<type>(NODE-xxxx)<!>: <description>` **I think so, there is no issue number for this but saw other commits that omitted this**
- [x] Changes are covered by tests **no current tests should fail since this code wasn't used**

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -645,7 +645,7 @@ function supportsOpMsg(conn: Connection) {
     return false;
   }
 
-  return !description.__nodejs_mock_server__;
+  return maxWireVersion(conn) >= 6 && !description.__nodejs_mock_server__;
 }
 
 function streamIdentifier(stream: Stream, options: ConnectionOptions): string {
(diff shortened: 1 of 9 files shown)


src/cmap/connection.ts
```diff
@@ -645,7 +645,7 @@ function supportsOpMsg(conn: Connection) {
     return false;
   }
 
-  return maxWireVersion(conn) >= 6 && !description.__nodejs_mock_server__;
+  return !description.__nodejs_mock_server__;
```
baileympearson, line 648:
```suggestion
  return maxWireVersion(conn) >= 6 && !description.__nodejs_mock_server__;
```
We actually need to leave this check in place.  For the initial connection handshake, we need to use OP_QUERY instead of OP_MESSAGE

baileympearson (changes requested):
Can we also update the wire version checks in update and delete related to `hint`?

baileympearson:
Hey @LinusU - are you able to make the changes I requested?  If not, I'm happy to take over the PR from here if that's okay with you.

PR #3417, opened 2022-09-15
feat(NODE-1603): add async_hooks support

nathanlepori:
#### What is changing?
Adding minimal support for async_hooks to version 4.9.0. See #1846 for more details. This PR is a forward port of that.

No

We need to be able to use [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) for an internal project. However without embedder API/async_hooks support the storage would be lost after a call to MongoDB. This PR enables async context tracking across MongoDB calls.

- [x] New TODOs have a related JIRA ticket (N/A)


nathanlepori:
Ping @mbroadst @daprahamian for review and feedback.
A few remarks about this PR. 

- This currently breaks on Node < 8.0.0 because of the unsupported async_hooks API. If supporting older versions of Node is a requirement please let me know.
- I only added an AsyncResource to the cursor class(es) because in the original PR it was unused in other classes.
- This is a very minimal implementation to allow async execution tracking, if a more complete implementation is required (AsyncResource implemented on additional classes, more callbacks running in async scope) please let me know.

src/operations/execute_operation.ts
```diff
@@ -136,11 +136,15 @@ export function executeOperation<
 
     try {
       executeWithServerSelection<TResult>(topology, session, operation, (error, result) => {
+        const asyncResource = operation.options.asyncResource;
```
addaleax, line 139: Is there a reason not to create an `asyncResource` for all `AbstractOperation`s? That seems semantically correct and would broaden async_hooks support without introducing more complexity (possibly even removing, as `maybeRunInAsyncScope` becomes unnecessary).

nathanlepori: That's a valid point. I just assumed there was a reason to bind the resource's lifecycle to the cursor like in the original PR. But on second thought the old codebase didn't use a class for operations so that might be the reason behind the choice. @mbroadst can you please confirm if that is the case and if you see any issues with binding the AsyncResource to operations?
I will do some testing with @addaleax 's version and see if there are any technical issues.

nathanlepori: That's a valid point. I just assumed it was better to bind the resource's lifecycle to the cursor like in the original PR. But on second thought it might have been because there was no operation class in the old codebase. @mbroadst can you please confirm that and give your thoughts on this? I will test @addaleax 's version as well to check if there are any technical issues with it.

nbbeeken:
Hey @nathanlepori Just an fyi the maintainers of this project are @dariakp @durran and @baileympearson and myself (@addaleax lends us an extremely helpful hand now and then). We appreciate the help getting this feature together, we've got work in our queue lined up to refactor executeOperation soon to use `async`/`await` which would likely impact the approach we take here. For now, we'll hold off on async_hooks support and return to this when we've finished with our internal migrations to async/await.

nathanlepori:
No problem just ping me when your work is merged and if I still have time I can make the necessary adjustments. If the idea going forward is to migrate the internals to native promises this might not even be necessary anymore, but we'll see. Our internal project relies heavily on AsyncLocalStorage so it's in my interest to have this feature implemented.

baileympearson:
hey @nathanlepori - version 4.11.0 refactored some internals of the driver to use async-await instead of callbacks and as a result, versions 4.11+ should properly track the context across driver API calls.  Using the following example on 4.10 and 4.11:

```javascript
import { MongoClient } from 'mongodb';
import { AsyncLocalStorage } from 'node:async_hooks';

const client = new MongoClient(process.env.MONGODB_URI);
await client.connect();

const collection = client.db('foo').collection('bar');

const localStorage = new AsyncLocalStorage();

function callbacks() {
	localStorage.run({ storeId: 1 }, () => {
		collection.insertOne({ name: "bailey" }, () => {
			const store = localStorage.getStore();
			console.error('callbacks', { store });
		})
	})
}

function promises() {
	localStorage.run({ storeId: 1 }, async () => {
		await collection.insertOne({ name: 'bailey' });
		const store = localStorage.getStore();
		console.error('callbacks', { store });
	})
}

callbacks();
promises();

setTimeout(() => {
	client.close()
}, 4000)
```

output:
```
// 4.10.0
callbacks { store: undefined }
promises { store: { storeId: 1 } }

// 4.11.x+
promises { store: { storeId: 1 } }
callbacks { store: { storeId: 1 } }
```

Not all of our internals properly bind asynchronous contexts, but using async-await at the top level of our driver API should be enough to preserve the context for users.  Can you confirm whether or not these changes satisfy your use case?

nathanlepori:
Hi @baileympearson that's great to hear! Thanks for notifying me. The project which is using this library is currently on hold so I can't tell you when I would be able to test it. But from the example provided it looks like the same use case so I can assume it's fixed. Closing the PR.

PR #3421, opened 2022-09-21
refactor(NODE-4633): executeOperation to use maybeCallback

nbbeeken:
### Description 

NODE-4633

#### What is changing?

- `package-lock.json` and `package.json`
  - Pulls in the latest node types to fix lint, should disappear when #3422 merges
- `src/bulk/common.ts`
  - contains the refactor discussed in #3406 removes the two usages of maybeCallback for one unified one, made possible by the executeOperation PR.
- `src/collection.ts` and `src/operations/rename.ts` and `src/operations/stats.ts` and `src/operations/delete.ts`
  - The promisify type was incompatible with these operation classes because they don't pass their types into the super class generic. Delete was able to be fixed the others lead to numerous TS issues, this all relates to NODE-3286.
- `src/operations/command.ts`
  - The override was redundant, could lead to mistakes down the road where the return type generic doesn't make it to the super class.
- `src/operations/execute_operation.ts`
  - The whole reason we're here 😸 Leaving `executeOperation` nearly empty to just call a new TBD named async function was intentional to avoid causing renames across tons of files, but we can do a follow up if we want to do that. 
  - ~Perhaps we don't need to run serverSelection twice?~ (session support check)
    - No we do, we need to use a difference selector than the user / operation may require.
  - Added a helper to get TS to think topology always exists (it will since we call connect)
  - retryOperation is a helper function now, this isn't a big change it's just not defined inline, so we have to pass it the variables it needs instead of it binding to the current scope.
    - retyability is undergoing a change with CSOT so I think more refactor + feature work will improve this further
- `src/operations/operation.ts`
  - Creates the promisified wrapper for .execute()
- `src/sdam/server.ts`
  - `noResponse` was causing endSessions to ALWAYS throw, which we squash, but it prevented executeOperation from retrying. Fixed.
- `src/sdam/topology.ts`
  - Makes a promisified wrapper for serverSelection, I would happy to make this a proper async function in a follow up. It's not super complex but will have a significant impact on tests so best to do it on its own.
- `test/integration/crud/bulk.test.ts`
  - ⚠️ Forgot to await an `insertMany` here async/await makes this **always** fail now. fixed!
- `test/integration/crud/maxTimeMS.test.ts`
  - ⚠️ Forgot to await an `insertMany` here async/await makes this **always** fail now. fixed!
  - Added printing stack traces to assertions.
- `test/integration/server-selection/operation_count.test.ts`
  - the operationCount used to be reached synchronously, we need to add a sleep here to allow the code to reach the increment
- `test/integration/sessions/sessions.spec.prose.test.ts` and `test/integration/sessions/sessions.test.ts`
  - Our minimum session usage for all the parallel operations we run in these tests is now 2. With async await we will never put the first created session back before the next operation attempts to pull one out of the pool. Unlike multithreaded drivers that may have some level of nondeterminism, connection checkIn -> release is deterministically after the next operation reaches the checkOut -> acquire step. As we further refactor to async await I don't expect this to go any higher than 2 but it is possible it could decrease again to 1 if the order of async steps allows us to return the session before the next operation tries to acquire one. 

async await is easier to reason about, error wise, ordering wise, and forces consistent execution.

### DIFF
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -78,18 +78,21 @@ export function executeOperation<
   return maybeCallback(() => executeOperationAsync(client, operation), callback);
 }
 
-async function assertTopology(client: MongoClient) {
+async function maybeConnect(client: MongoClient): Promise<Topology> {
   const { topology } = client;
   if (topology == null) {
     if (client.s.hasBeenClosed) {
       throw new MongoNotConnectedError('Client must be connected before running operations');
     }
     client.s.options[Symbol.for('@@mdb.skipPingOnConnect')] = true;
-    await client.connect();
-    delete client.s.options[Symbol.for('@@mdb.skipPingOnConnect')];
+    try {
+      await client.connect();
+    } finally {
+      delete client.s.options[Symbol.for('@@mdb.skipPingOnConnect')];
+    }
   }
   if (client.topology == null) {
-    throw new MongoNotConnectedError('Client must be connected before running operations');
+    throw new MongoRuntimeError('client.connect did not create a topology but also did not throw');
   }
   return client.topology;
 }
@@ -103,7 +106,7 @@ async function executeOperationAsync<
     throw new MongoRuntimeError('This method requires a valid operation instance');
   }
 
-  const topology = await assertTopology(client);
+  const topology = await maybeConnect(client);
 
   if (topology.shouldCheckForSessionSupport()) {
     await topology.selectServerAsync(ReadPreference.primaryPreferred, {});
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -191,28 +191,24 @@ async function executeOperationAsync<
   const hasReadAspect = operation.hasAspect(Aspect.READ_OPERATION);
   const hasWriteAspect = operation.hasAspect(Aspect.WRITE_OPERATION);
 
-  if ((hasReadAspect && willRetryRead) || (hasWriteAspect && willRetryWrite)) {
-    if (hasWriteAspect && willRetryWrite) {
-      operation.options.willRetryWrite = true;
-      session.incrementTransactionNumber();
-    }
+  if (hasWriteAspect && willRetryWrite) {
+    operation.options.willRetryWrite = true;
+    session.incrementTransactionNumber();
   }
 
   try {
-    try {
-      return await operation.executeAsync(server, session);
-    } catch (operationError) {
-      if ((hasReadAspect && willRetryRead) || (hasWriteAspect && willRetryWrite)) {
-        if (operationError instanceof MongoError) {
-          return await retryOperation(operation, operationError, {
-            session,
-            topology,
-            selector
-          });
-        }
+    return await operation.executeAsync(server, session);
+  } catch (operationError) {
+    if ((hasReadAspect && willRetryRead) || (hasWriteAspect && willRetryWrite)) {
+      if (operationError instanceof MongoError) {
+        return await retryOperation(operation, operationError, {
+          session,
+          topology,
+          selector
+        });
       }
-      throw operationError;
     }
+    throw operationError;
   } finally {
     if (session?.owner != null && session.owner === owner) {
       await session.endSession().catch(() => null);
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -75,10 +75,10 @@ export function executeOperation<
   T extends AbstractOperation<TResult>,
   TResult = ResultTypeFromOperation<T>
 >(client: MongoClient, operation: T, callback?: Callback<TResult>): Promise<TResult> | void {
-  return maybeCallback(() => εχεςμτεθρεπατιοη(client, operation), callback);
+  return maybeCallback(() => executeOperationAsync(client, operation), callback);
 }
 
-const assertTopology = async (client: MongoClient) => {
+async function assertTopology(client: MongoClient) {
   const { topology } = client;
   if (topology == null) {
     if (client.s.hasBeenClosed) {
@@ -92,9 +92,9 @@ const assertTopology = async (client: MongoClient) => {
     throw new MongoNotConnectedError('Client must be connected before running operations');
   }
   return client.topology;
-};
+}
 
-async function εχεςμτεθρεπατιοη<
+async function executeOperationAsync<
   T extends AbstractOperation<TResult>,
   TResult = ResultTypeFromOperation<T>
 >(client: MongoClient, operation: T): Promise<TResult> {
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -78,9 +78,17 @@ export function executeOperation<
   return maybeCallback(() => executeOperationAsync(client, operation), callback);
 }
 
-async function maybeConnect(client: MongoClient): Promise<Topology> {
-  const { topology } = client;
-  if (topology == null) {
+async function executeOperationAsync<
+  T extends AbstractOperation<TResult>,
+  TResult = ResultTypeFromOperation<T>
+>(client: MongoClient, operation: T): Promise<TResult> {
+  if (!(operation instanceof AbstractOperation)) {
+    // TODO(NODE-3483): Extend MongoRuntimeError
+    throw new MongoRuntimeError('This method requires a valid operation instance');
+  }
+
+  if (client.topology == null) {
+    // Auto connect on operation
     if (client.s.hasBeenClosed) {
       throw new MongoNotConnectedError('Client must be connected before running operations');
     }
@@ -91,22 +99,11 @@ async function maybeConnect(client: MongoClient): Promise<Topology> {
       delete client.s.options[Symbol.for('@@mdb.skipPingOnConnect')];
     }
   }
-  if (client.topology == null) {
+
+  const { topology } = client;
+  if (topology == null) {
     throw new MongoRuntimeError('client.connect did not create a topology but also did not throw');
   }
-  return client.topology;
-}
-
-async function executeOperationAsync<
-  T extends AbstractOperation<TResult>,
-  TResult = ResultTypeFromOperation<T>
->(client: MongoClient, operation: T): Promise<TResult> {
-  if (!(operation instanceof AbstractOperation)) {
-    // TODO(NODE-3483): Extend MongoRuntimeError
-    throw new MongoRuntimeError('This method requires a valid operation instance');
-  }
-
-  const topology = await maybeConnect(client);
 
   if (topology.shouldCheckForSessionSupport()) {
     await topology.selectServerAsync(ReadPreference.primaryPreferred, {});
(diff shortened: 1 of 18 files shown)


src/operations/execute_operation.ts
```diff
@@ -75,97 +75,73 @@ export function executeOperation<
   T extends AbstractOperation<TResult>,
   TResult = ResultTypeFromOperation<T>
 >(client: MongoClient, operation: T, callback?: Callback<TResult>): Promise<TResult> | void {
+  return maybeCallback(() => εχεςμτεθρεπατιοη(client, operation), callback);
+}
+
+const assertTopology = async (client: MongoClient) => {
```
baileympearson, line 81: Let's be consistent with arrow functions or `function` functions in the same file.
```suggestion
async function assertTopology(client: MongoClient) {
```

src/operations/execute_operation.ts
```diff
@@ -75,97 +75,73 @@ export function executeOperation<
...
-) {
   const readPreference = operation.readPreference ?? ReadPreference.primary;
   const inTransaction = !!session?.inTransaction();
 
-  if (inTransaction && !readPreference.equals(ReadPreference.primary)) {
-    return callback(
-      new MongoTransactionError(
-        `Read preference in a transaction must be primary, not: ${readPreference.mode}`
-      )
+  if (!readPreference.equals(ReadPreference.primary) && inTransaction) {
```
baileympearson, line 142:
```suggestion
  if (inTransaction && !readPreference.equals(ReadPreference.primary)) {
```

src/operations/execute_operation.ts
```diff
@@ -188,111 +164,113 @@ function executeWithServerSelection<TResult>(
...
 
-    if (isWriteOperation && originalError.code === MMAPv1_RETRY_WRITES_ERROR_CODE) {
-      return callback(
-        new MongoServerError({
-          message: MMAPv1_RETRY_WRITES_ERROR_MESSAGE,
-          errmsg: MMAPv1_RETRY_WRITES_ERROR_MESSAGE,
-          originalError
-        })
-      );
+  if (session == null || !operation.hasAspect(Aspect.RETRYABLE)) {
```
baileympearson, line 169: question: under what circumstances would we have a null session?

src/operations/execute_operation.ts
```diff
@@ -188,111 +164,113 @@ function executeWithServerSelection<TResult>(
...
+    // No session or non-retryable operation, simple early exit
+    try {
+      return await operation.executeAsync(server, session);
```
baileympearson, line 172: optional for clarity
```suggestion
      return await operation.executeAsync(server, null);
```

src/operations/execute_operation.ts
baileympearson, line 169: I wouldn't be opposed to breaking this if statement into two - one to handle the null session case and another to handle non-retryable operations.  It would be roughly the same code in both except the null session block wouldn't have the end sessions logic.  

It's just preference though so this is optional

src/operations/execute_operation.ts
```diff
@@ -188,111 +164,113 @@ function executeWithServerSelection<TResult>(
...
+        }
       }
+      throw operationError;
+    }
+  } finally {
+    if (session?.owner != null && session.owner === owner) {
+      await session.endSession().catch(() => null);
+    }
+  }
+}
```
baileympearson, line 218: in the original execute_operation, we only execute the operation when either `(hasReadAspect && willRetryRead) || (hasWriteAspect && willRetryWrite))`, otherwise we fall through the the call to execute that doesn't retry.  We need to preserve this behavior and move this try-catch into the above if-statement.  That also means we may want to flip the logic back so that we have the catch-all call to execute without retry at the end of the method.

baileympearson (changes requested):

src/operations/execute_operation.ts
nbbeeken: if topology.hasSessionSupport() is false, which is the case if any one of the servers in the topology report a null `logicalSessionTimeoutMinutes`

src/operations/execute_operation.ts
nbbeeken: The session can still exist here so this would break non-retryable operations

src/operations/stats.ts
```diff
@@ -33,11 +33,7 @@ export class CollStatsOperation extends CommandOperation<Document> {
     this.collectionName = collection.collectionName;
   }
 
-  override execute(
-    server: Server,
-    session: ClientSession | undefined,
-    callback: Callback<CollStats>
-  ): void {
+  override execute(server: Server, session: ClientSession | undefined, callback: Callback): void {
```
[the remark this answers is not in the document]
nbbeeken: undid this

src/operations/rename.ts
```diff
@@ -38,11 +38,7 @@ export class RenameOperation extends RunAdminCommandOperation {
     this.newName = newName;
   }
 
-  override execute(
-    server: Server,
-    session: ClientSession | undefined,
-    callback: Callback<Collection>
-  ): void {
+  override execute(server: Server, session: ClientSession | undefined, callback: Callback): void {
```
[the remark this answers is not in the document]
nbbeeken: The promisify of execute on all the operation classes was not getting the return type automatically correctly (callback type makes the result optional) so anyway I fixed these return types elsewhere so I've undone these changes.

src/operations/execute_operation.ts
```diff
@@ -75,97 +75,73 @@ export function executeOperation<
...
+  if (topology == null) {
+    if (client.s.hasBeenClosed) {
+      throw new MongoNotConnectedError('Client must be connected before running operations');
+    }
+    client.s.options[Symbol.for('@@mdb.skipPingOnConnect')] = true;
+    await client.connect();
+    delete client.s.options[Symbol.for('@@mdb.skipPingOnConnect')];
+  }
+  if (client.topology == null) {
+    throw new MongoNotConnectedError('Client must be connected before running operations');
```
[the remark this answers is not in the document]
nbbeeken: Changed to a runtime error, it should be unreachable code it's just our current typescript set up that requires this.

src/operations/execute_operation.ts
```diff
@@ -75,97 +75,73 @@ export function executeOperation<
...
+
+const assertTopology = async (client: MongoClient) => {
+  const { topology } = client;
+  if (topology == null) {
+    if (client.s.hasBeenClosed) {
+      throw new MongoNotConnectedError('Client must be connected before running operations');
+    }
+    client.s.options[Symbol.for('@@mdb.skipPingOnConnect')] = true;
+    await client.connect();
+    delete client.s.options[Symbol.for('@@mdb.skipPingOnConnect')];
```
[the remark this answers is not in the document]
nbbeeken: done!

src/operations/execute_operation.ts
```diff
@@ -75,97 +75,73 @@ export function executeOperation<
   T extends AbstractOperation<TResult>,
   TResult = ResultTypeFromOperation<T>
 >(client: MongoClient, operation: T, callback?: Callback<TResult>): Promise<TResult> | void {
+  return maybeCallback(() => εχεςμτεθρεπατιοη(client, operation), callback);
+}
+
+const assertTopology = async (client: MongoClient) => {
+  const { topology } = client;
```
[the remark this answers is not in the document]
nbbeeken: I renamed and improved the control flow here, keeping the assertion in here though seems like an acceptable approach because it's really unreachable to have a nullish topology after connect.

src/operations/execute_operation.ts
nbbeeken: done

test/integration/server-selection/operation_count.test.ts
```diff
@@ -124,6 +124,9 @@ describe('Server Operation Count Tests', function () {
         collection.insertOne({ count: 1 })
       );
 
+      // operation count will not be incremented synchronously after execOp async/await refactor
```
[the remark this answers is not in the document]
nbbeeken: done

src/operations/execute_operation.ts
nbbeeken: from slack: there is gating in the catch block, no change in behavior

src/operations/execute_operation.ts
nbbeeken: done

src/operations/execute_operation.ts
nbbeeken: done

src/operations/execute_operation.ts
```diff
@@ -188,111 +167,114 @@ function executeWithServerSelection<TResult>(
...
+    session.incrementTransactionNumber();
+  }
 
-      operation.execute(server, session, callback);
-    });
+  try {
+    return await operation.executeAsync(server, session);
+  } catch (operationError) {
+    if ((hasReadAspect && willRetryRead) || (hasWriteAspect && willRetryWrite)) {
+      if (operationError instanceof MongoError) {
```
baileympearson, line 208:
```suggestion
    const isRetryableOperation = (hasReadAspect && willRetryRead) || (hasWriteAspect && willRetryWrite);
    if (isRetryableOperation && operationError instanceof MongoError) {
```

baileympearson: optional cleanup - something like this or just combining the if statements

src/operations/execute_operation.ts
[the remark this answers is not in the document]
baileympearson: I think we should move away from the naming convention `maybe<something>` because it's not particularly descriptive and after v5, we won't have need for `maybeCallback` anymore.

maybe even just `autoConnect` would be better but we could also just move this code back inline - it's only used in once place and isn't terribly complicated so I see no problem with having it inline

baileympearson (changes requested):

src/operations/execute_operation.ts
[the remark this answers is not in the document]
nbbeeken: Inlined

baileympearson (approved):

test/integration/sessions/sessions.spec.prose.test.ts
```diff
@@ -46,6 +46,6 @@ describe('ServerSession', () => {
     expect(events).to.have.lengthOf(operations.length);
 
     // This is a guarantee in node, unless you are performing a transaction (which is not being done in this test)
-    expect(new Set(events.map(ev => ev.command.lsid.id.toString('hex')))).to.have.lengthOf(1);
+    expect(new Set(events.map(ev => ev.command.lsid.id.toString('hex')))).to.have.lengthOf(2);
```
dariakp, line 49: why did this change?

nbbeeken: > Our minimum session usage for all the parallel operations we run in these tests is now 2. With async await we will never put the first created session back before the next operation attempts to pull one out of the pool. Unlike multithreaded drivers that may have some level of nondeterminism, connection checkIn -> release is deterministically after the next operation reaches the checkOut -> acquire step. As we further refactor to async await I don't expect this to go any higher than 2 but it is possible it could decrease again to 1 if the order of async steps allows us to return the session before the next operation tries to acquire one.

dariakp (approved):

PR #3425, opened 2022-09-28
fix(NODE-4186): accept ReadPreferenceLike in TransactionOptions type

noahsilas:
### Description

This commit expands the TransactionOptions type definition to allow values like `'primary'`.

#### What is changing?

Before, the type definition only allowed passing an instance of the `ReadPreference` class. This differs from documented and common usage. We expand the type definition from `ReadPreference` to `ReadPreferenceLike`.

Updating this type annotation will reduce confusion for Typescipt users of the library.

No, this makes the types match the existing documentation more closely.  For instance, in https://www.mongodb.com/docs/drivers/node/current/fundamentals/transactions/#transaction-settings, this API is documented as accepting a string: 

> const transactionOptions = {
>   readPreference: 'primary',
>   readConcern: { level: 'local' },
>   writeConcern: { w: 'majority' },
>   maxCommitTimeMS: 1000
> };
> session.startTransaction(transactionOptions);

The documented code snippet above should not cause a Typescript error.

We can also see in https://github.com/mongodb/node-mongodb-native/blob/48fa588362/src/transactions.ts#L108 that the transaction constructor will pass its options through to `ReadPreference.fromOptions`, which will handle coercing a string into a ReadPreference instance, so the documentation appears to be correct while the type definition is lacking.

### DIFF
--- a/src/transactions.ts
+++ b/src/transactions.ts
@@ -2,6 +2,7 @@ import type { Document } from './bson';
 import { MongoRuntimeError, MongoTransactionError } from './error';
 import type { CommandOperationOptions } from './operations/command';
 import { ReadConcern, ReadConcernLike } from './read_concern';
+import type { ReadPreferenceLike } from './read_preference';
 import { ReadPreference } from './read_preference';
 import type { Server } from './sdam/server';
 import { WriteConcern } from './write_concern';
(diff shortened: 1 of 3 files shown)


nbbeeken:
@noahsilas Thanks for the help here! I've added a test to just double check that this keeps working at runtime (as it already has been)

nbbeeken (approved):

baileympearson (approved):

noahsilas:
Thanks for the support on the testing @nbbeeken ! Appreciate it a lot :smile:

PR #3426, opened 2022-09-28
feat(NODE-4650): handle handshake errors with SDAM

dariakp:
### Description
NODE-4650/NODE-3135

#### What is changing?
- Updating SDAM `handleError` logic to handle auth handshake errors [according to the spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#error-handling-pseudocode)
- Unskipping the corresponding SDAM auth tests
- New unit tests just for the new portions of the handleError logic
- Injecting a new error `connectionGeneration` property when `createConnection` receives an error so that it can be checked for staleness even in the absence of a connection (which makes this a `feat:`)

No

Spec compliance

### DIFF
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -29,6 +29,7 @@ import {
   MongoInvalidArgumentError,
   MongoNetworkError,
   MongoNetworkTimeoutError,
+  MongoRuntimeError,
   MongoServerClosedError,
   MongoServerError,
   MongoUnexpectedServerResponseError,
@@ -348,8 +349,12 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       (err, conn, cb) => {
         if (err || !conn) {
           this.s.operationCount -= 1;
+          if (!err) {
+            return cb(new MongoRuntimeError('Failed to create connection without error'));
+          }
           if (!(err instanceof PoolClearedError)) {
-            markServerUnknown(this, err);
+            this.handleError(err);
+            // markServerUnknown(this, err);
           } else {
             err.addErrorLabel(MongoErrorLabel.RetryableWriteError);
           }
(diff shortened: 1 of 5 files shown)


src/sdam/server.ts
```diff
@@ -348,8 +349,11 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
           this.s.operationCount -= 1;
+          if (!err) {
+            return cb(new MongoRuntimeError('Failed to create connection without error'));
+          }
           if (!(err instanceof PoolClearedError)) {
-            markServerUnknown(this, err);
+            this.handleError(err);
           } else {
             err.addErrorLabel(MongoErrorLabel.RetryableWriteError);
           }
```
nbbeeken, line 359: minor: flip the if stmt, we've before preferred to keep instanceof inside an if truthy cus it's easier to read, but just an 🧅

dariakp: this isn't new logic, we already talked about the if statement on the PR that introduced it; iirc your opinion was ultimately that we should be comfortable reading inverted instanceof checks ;)

nbbeeken (dismissed):

src/sdam/server.ts
```diff
@@ -378,17 +382,24 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
-
-        if (!this.loadBalanced) {
-          error.addErrorLabel(MongoErrorLabel.ResetPool);
-          markServerUnknown(this, error);
-        } else if (connection) {
-          this.s.pool.clear(connection.serviceId);
-        }
+
+    // ignore stale errors
+    if (error.connectionGeneration && error.connectionGeneration < this.s.pool.generation) {
```
baileympearson, line 387:
```suggestion
    const isStaleGenerationError = error.connectionGeneration && error.connectionGeneration < this.s.pool.generation;
    if (isStaleGenerationError) {
```

comments can easily get out of date, but variable names cannot 🙂

nbbeeken (approved):

test/unit/sdam/server.test.ts
```diff
@@ -0,0 +1,162 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
baileympearson, line 1: I didn't see anything in the unskipped spec tests, but should we add tests for stale generation errors to ensure that we do not handle errors from a stale generation?

baileympearson (changes requested):

test/unit/sdam/server.test.ts
dariakp: There is a set of unit tests covering it

baileympearson (approved):

nbbeeken (approved):

PR #3428, opened 2022-09-28
refactor(NODE-4632): async await in MongoClient, ClientSession, and AbstractCursor

nbbeeken:
### Description

NODE-4632

#### What is changing?

- `src/change_stream.ts` and `src/collection.ts` and `src/cursor/find_cursor.ts` and `src/db.ts` and `src/mongo_types.ts` and `src/operations/drop.ts`
  - spelling mistakes
- `src/connection_string.ts` and `test/unit/connection_string.test.ts` and `test/unit/mongo_client.test.js`
  - refactor `resolveSRVRecord` to be async, wrap dns APIs in promisify
  - `validateLoadBalancedOptions` now throws since the error does not need to be passed to a callback
- `src/cursor/abstract_cursor.ts`
  - Implement the async iterator using native syntax when a custom promise is not set. I think this is important to do now ensuring is maintains the expected behavior so when we drop custom promise support we don't miss something in the other API removals/changes that would be part of the major.
  - Utilize the async iterator for toArray and forEach
  - promifisify next - we may want to make this really async in a follow up.
  - ditto cleanupCursor
- `src/encrypter.ts`
  - async connectInternalClient, this method is internal, does not need to support maybeCallback
- `src/gridfs/download.ts` and `test/integration/gridfs/gridfs_stream.test.js`
  - remove `nextTick` usage, the callback for close was being called sync, now that we always do things async it changed the order of events. removing nextTick brings us back to that behavior
- `src/gridfs/upload.ts`
  - no longer relevant comment
- `src/mongo_client.ts`
  - maybeCallback connect - moved the floating function logic into the class, make TS happier and simplifies repition.
    - removed `no mongos proxies found` condition, this is unreachable, we removed this case in 3.x
  - maybeCallback close - this was mostly using promises already. topology and encrypter close is still wrapped though.
- `src/operations/connect.ts`
  - removed, see above
- `src/sessions.ts`
  - endSession, commit/abortTransaction async-ified.
  - promisified endTransaction, this one is a bit involved to convert since the finally vs handle the error is carefully part of correct transaction logic. should be done separately.
- `src/utils.ts`
  - Deleted maybePromise 🎉 🥳 🍾 🎂 
  - changed the name of a variable so that there is no reference to maybePromise, definitelyPromise from here on out
- `test/integration/collation/collations.test.js`
  - this test was failing for a bit, fixed the capital
- `test/integration/crud/aggregation.test.ts`
  - added a beforeEachClient, and fixed the test the proved we were throwing synchronously ⚠️ This is no longer possible!
- `test/integration/crud/explain.test.js`
  - Explain demonstrates an exhaust cursor in one next call, so I had to debug a bug here, test got refactored in the process.
- `test/integration/crud/misc_cursors.test.js`
  - toArray and forEach do not throw on exhausted cursors, but .next and async iteration will, test needed debugging got refactored.
- `test/integration/node-specific/cursor_stream.test.js`
  - this test was hanging, the end event never came, I refactored but made one change to introduce a call to resume which is what I think was blocking the event. Before our sync call to callbacks would get us the event to emit even if the stream is paused, so this was just incorrect stream usage IIUC.
- `test/integration/server-selection/readpreference.test.js`
  - These tests were using sinon to check exact args passes to server selection, the args are slightly different but same logic is being checked.


src/connection_string.ts
```diff
@@ -70,97 +70,91 @@ function matchesParentDomain(srvAddress: string, parentDomain: string): boolean
  * @param uri - The connection string to parse
  * @param options - Optional user provided connection string options
  */
-export function resolveSRVRecord(options: MongoOptions, callback: Callback<HostAddress[]>): void {
+export async function resolveSRVRecord(options: MongoOptions): Promise<HostAddress[]> {
+  // We need to do this here to make sinon tests work :(
```
baileympearson, line 74: this can't be done outside of `resolveSRVRecord`?

src/connection_string.ts
```diff
@@ -526,20 +518,20 @@ function validateLoadBalancedOptions(
   hosts: HostAddress[] | string[],
   mongoOptions: MongoOptions,
   isSrv: boolean
-): MongoParseError | undefined {
+): void {
```
baileympearson, line 521: Can we annotate this function as throwing / add a comment?  The error information used to be codified in the return type, but now it's lost.

src/cursor/abstract_cursor.ts
```diff
@@ -284,11 +286,40 @@ export abstract class AbstractCursor<
   }
 
   [Symbol.asyncIterator](): AsyncIterator<TSchema, void> {
+    async function* nativeAsyncIterator(this: AbstractCursor<TSchema>) {
```
baileympearson, line 289: Is this necessary?  I understand the desire to use generators wherever possible, but wouldn't it be simpler to just leave the implementation as it was (perhaps adjusting for closed cursors)?

src/mongo_client.ts
```diff
@@ -437,12 +436,57 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
     }
 
-    return maybePromise(callback, cb => {
-      connect(this, this[kOptions], err => {
-        if (err) return cb(err);
-        cb(undefined, this);
-      });
-    });
+    return maybeCallback(async () => {
+      // If a connection already been established, we can terminate early
```
baileympearson, line 440:
```suggestion
```

src/mongo_client.ts
```diff
@@ -437,12 +436,57 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+      }
+
+      const options = this[kOptions];
+
+      if (typeof options.srvHost === 'string') {
+        const hosts = await resolveSRVRecord(options);
+
+        for (const [index, host] of hosts.entries()) {
+          options.hosts[index] = host;
+        }
```
baileympearson, line 452:
```suggestion
        options.hosts = await resolveSRVRecord(options);
```
any reason we can't do it this way?

src/mongo_client.ts
```diff
@@ -437,12 +436,57 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+      }
+
+      // Create the topology
```
baileympearson, line 455:
```suggestion
```

src/mongo_client.ts
```diff
@@ -437,12 +436,57 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+        try {
+          await promisify(callback => topology.connect(options, callback))();
+        } catch (error) {
+          topology.close({ force: true });
+          throw error;
+        }
+      };
+
+      if (this.autoEncrypter) {
+        // initialize CSFLE if requested
```
baileympearson, line 478:
```suggestion
```

src/mongo_client.ts
```diff
@@ -437,12 +436,57 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+        const initAutoEncrypter = promisify(callback => this.autoEncrypter?.init(callback));
+        await initAutoEncrypter();
+        await topologyConnect();
+        await options.encrypter.connectInternalClient();
+      } else {
+        // otherwise connect normally
```
baileympearson, line 484:
```suggestion
```

src/mongo_client.ts
```diff
@@ -475,66 +519,53 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
-          const servers = selector(topologyDescription, serverDescriptions);
-          if (servers.length === 0) {
-            return;
-          }
-
-          const endSessions = Array.from(this.s.sessionPool.sessions, ({ id }) => id);
-          if (endSessions.length === 0) return;
-          return this.db('admin')
+      await Promise.all(activeSessionEnds);
+
```
baileympearson, line 532: Are we missing a check for a null topology here?  There used to be one

test/integration/crud/misc_cursors.test.js
```diff
@@ -31,50 +31,29 @@ describe('Cursor', function () {
...
-            cursor.toArray(err => {
-              expect(err).to.not.exist;
+    expect(cursor.closed).to.be.true;
 
-              // Should fail if called again (cursor should be closed)
-              cursor.toArray(err => {
-                expect(err).to.not.exist;
+    // should not fail instead return empty array
+    const secondToArray = await cursor.toArray().catch(error => error);
+    expect(secondToArray).to.be.an('array');
```
baileympearson, line 48:
```suggestion
    expect(secondToArray).to.be.an('array');
    expect(secondToArray).to.have.lengthOf(0);
```

baileympearson: we should also assert that the array is actually empty

test/integration/crud/misc_cursors.test.js
```diff
@@ -31,50 +31,29 @@ describe('Cursor', function () {
...
+    expect(firstToArray).to.be.an('array');
 
-            cursor.toArray(err => {
-              expect(err).to.not.exist;
+    expect(cursor.closed).to.be.true;
 
-              // Should fail if called again (cursor should be closed)
-              cursor.toArray(err => {
-                expect(err).to.not.exist;
+    // should not fail instead return empty array
```
baileympearson, line 46:
```suggestion
```

test/integration/crud/misc_cursors.test.js
```diff
@@ -31,50 +31,29 @@ describe('Cursor', function () {
...
-                    done();
-                  }
-                );
-              });
-            });
-          });
-        });
-      });
-    }
+    // should not fail instead returns nothing
```
baileympearson, line 50:
```suggestion
```

baileympearson (changes requested):
Looks really good overall, just some questions and minor comments

src/mongo_client.ts
nbbeeken: options is frozen, so it would crash

src/mongo_client.ts
nbbeeken: We should be able to move the one above to below here, the only way it would become null is if there was parallel closes I guess?

src/connection_string.ts
nbbeeken: switching to promises and fixed the sinon usages

src/connection_string.ts
nbbeeken: done!

src/cursor/abstract_cursor.ts
nbbeeken: > Implement the async iterator using native syntax when a custom promise is not set. I think this is important to do now ensuring is maintains the expected behavior so when we drop custom promise support we don't miss something in the other API removals/changes that would be part of the major.

I think it's worth doing now, but I think we can better reuse the code, take another look, I use the same iterator, just wrap the promises if needed

src/mongo_client.ts
nbbeeken: done

src/mongo_client.ts
nbbeeken: done

src/mongo_client.ts
nbbeeken: done

test/integration/crud/misc_cursors.test.js
nbbeeken: good point, added

test/integration/crud/misc_cursors.test.js
nbbeeken: done

src/change_stream.ts
```diff
@@ -918,36 +918,26 @@ export class ChangeStream<
...
-    if (isResumableError(changeStreamError, this.cursor.maxWireVersion)) {
-      this.cursor.close().catch(() => null);
-
-      const topology = getTopology(this.parent);
-      topology.selectServer(this.cursor.readPreference, {}, serverSelectionError => {
-        // if the topology can't reconnect, close the stream
-        if (serverSelectionError) return this.close(() => callback(changeStreamError));
+    if (!isResumableError(changeStreamError, this.cursor.maxWireVersion)) {
+      return this.close();
+    }
```
baileympearson, line 930:
```suggestion
    if (!isResumableError(changeStreamError, this.cursor.maxWireVersion)) {
      try {
        await this.close();
      } catch (err) {
        // do nothing
      } finally {
        throw changeStreamError
      }
    }
```

src/change_stream.ts
```diff
@@ -918,36 +918,26 @@ export class ChangeStream<
...
-      this.close(() => callback(changeStreamError));
+    await this.cursor.close().catch(() => null);
+    const topology = getTopology(this.parent);
+    try {
+      await topology.selectServerAsync(this.cursor.readPreference, {});
+      this.cursor = this._createChangeStreamCursor(this.cursor.resumeOptions);
+    } catch (serverSelectionError) {
+      // if the topology can't reconnect, close the stream
+      await this.close();
+      throw serverSelectionError;
```
baileympearson, line 940:
```suggestion
      throw changeStreamError;
```

baileympearson (changes requested):

src/change_stream.ts
nbbeeken: done

src/change_stream.ts
nbbeeken: done

src/cursor/abstract_cursor.ts
baileympearson: So the idea is, once we're in v5, so rewrite it as a standalone async generator?  

```typescript
[Symbol.asyncIterator] = async function*() { .... } 
```
?

src/cursor/abstract_cursor.ts
```diff
@@ -284,11 +286,34 @@ export abstract class AbstractCursor<
...
+      if (this.closed) {
+        return;
+      }
+
+      while (true) {
+        const document = await this.next();
+
+        if (document == null) {
```
baileympearson, line 297: I was looking at refactoring `next` on Friday, and I think it's slightly more complicated than this.  This forbids any values of `null` from being emitted from the cursor, so something like 

```typescript
for await (const document of cursor.map(() => null) {

}
```

will not actually iterate the cursor.

I think we need a sort of Result-like abstraction to return from `next` (I'm looking at it now actually) because `next` needs to differentiate between the cursor being closed (currently `next` returns null in this scenario) and handling a `null` document.  If `next` instead returned something like `{ value: T | null } | null`, we could rely on `null` indicating that the cursor was exhausted.

test/unit/sdam/srv_polling.test.ts
```diff
@@ -87,15 +88,32 @@ describe('Mongos SRV Polling', function () {
...
+      it('should do something if dns API breaks', async function () {
+        const poller = new SrvPoller({ srvHost: SRV_HOST, loggerLevel: 'error' });
+
+        // @ts-expect-error: Testing what happens if node breaks DNS API
+        sinon.stub(dns.promises, 'resolveSrv').resolves(null);
+
+        const loggerError = sinon.stub(poller.logger, 'error').returns();
+
+        poller._poll();
+        await sleep(1);
```
baileympearson, line 110: The `await` here is necessary because `_poll` is void, but we need to wait on the promises, right?

The SrvPoller is internal.  What if we returned a promise from `_poll` and just call `await poller._poll()` here instead of adding the `sleep`?  I think generally that should be the approach we prefer to using promises in a function and sleeping to resolve them in a test.

test/integration/crud/misc_cursors.test.js
```diff
@@ -31,50 +31,28 @@ describe('Cursor', function () {
     await client.close();
   });
 
-  it('cursorShouldBeAbleToResetOnToArrayRunningQueryAgain', {
-    // Add a tag that our runner can trigger on
-    // in this case we are setting that node needs to be higher than 0.10.X to run
-    metadata: {
-      requires: { topology: ['single', 'replicaset', 'sharded'] }
-    },
+  it('should not fail on toArray and forEach calls after cursor is closed', async function () {
```
baileympearson, line 34:
```suggestion
  it('should not throw an error when toArray and forEach are called after cursor is closed', async function () {
```

test/unit/assorted/imports.test.ts
```diff
@@ -15,7 +15,7 @@ function* walk(root) {
   }
 }
 
-describe('importing mongodb driver', () => {
+describe.skip('importing mongodb driver', () => {
```
baileympearson, line 18:
```suggestion
describe('importing mongodb driver', () => {
```

baileympearson: how did this get past the lint? 🤔

baileympearson (changes requested):

src/cursor/abstract_cursor.ts
nbbeeken: Pretty much more like: 
```ts
class {
async *[Symbol.asyncIterator]() {
  // contents of the iterator we've declared inline for v4
}
}
```
so the function will be directly on the class instead of this custom promise wrapping. The idea is to not do the done/value stuff ourselves, it's built in to the language if you just use the syntax.

test/unit/assorted/imports.test.ts
nbbeeken: What lint? there's nothing wrong with skipping tests

test/unit/sdam/srv_polling.test.ts
nbbeeken: fixed!

baileympearson (approved):

PR #3429, opened 2022-09-29
feat(NODE-3255): add minPoolSizeCheckIntervalMS option to connection pool

dariakp:
### Description
NODE-3255

#### What is changing?
- Added `minPoolSizeCheckIntervalMS` option to the connection pool which controls how frequently the `ensureMinPoolSize` runs
  - Added a unit test for the option and the default of 100ms
- The corresponding `backgroundThreadIntervalMS` option in the CMAP test runner is now correctly parsed and passed to the test pools

No

Better spec testing

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -90,6 +90,8 @@ export interface ConnectionPoolOptions extends Omit<ConnectionOptions, 'id' | 'g
   waitQueueTimeoutMS: number;
   /** If we are in load balancer mode. */
   loadBalanced: boolean;
+  /** @experimental */
+  minPoolSizeCheckIntervalMS: number;
 }
 
 /** @internal */
@@ -234,6 +236,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       maxConnecting: options.maxConnecting ?? 2,
       maxIdleTimeMS: options.maxIdleTimeMS ?? 0,
       waitQueueTimeoutMS: options.waitQueueTimeoutMS ?? 0,
+      minPoolSizeCheckIntervalMS: options.minPoolSizeCheckIntervalMS ?? 100,
       autoEncrypter: options.autoEncrypter,
       metadata: options.metadata
     });
@@ -683,12 +686,18 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
         }
         if (this[kPoolState] === PoolState.ready) {
           clearTimeout(this[kMinPoolSizeTimer]);
-          this[kMinPoolSizeTimer] = setTimeout(() => this.ensureMinPoolSize(), 10);
+          this[kMinPoolSizeTimer] = setTimeout(
+            () => this.ensureMinPoolSize(),
+            this.options.minPoolSizeCheckIntervalMS
+          );
         }
       });
     } else {
       clearTimeout(this[kMinPoolSizeTimer]);
-      this[kMinPoolSizeTimer] = setTimeout(() => this.ensureMinPoolSize(), 100);
+      this[kMinPoolSizeTimer] = setTimeout(
+        () => this.ensureMinPoolSize(),
+        this.options.minPoolSizeCheckIntervalMS
+      );
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/cmap/connection_pool.ts
```diff
@@ -234,6 +236,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       maxConnecting: options.maxConnecting ?? 2,
       maxIdleTimeMS: options.maxIdleTimeMS ?? 0,
       waitQueueTimeoutMS: options.waitQueueTimeoutMS ?? 0,
+      minPoolSizeCheckIntervalMS: options.minPoolSizeCheckIntervalMS ?? 100,
```
dariakp, line 239: 100 is arbitrary; this is left entirely to the drivers' discretion in the spec, but we were using `10` and `100` before, depending on the circumstance, and `100` seemed frequent enough

src/cmap/connection_pool.ts
```diff
@@ -90,6 +90,8 @@ export interface ConnectionPoolOptions extends Omit<ConnectionOptions, 'id' | 'g
   waitQueueTimeoutMS: number;
   /** If we are in load balancer mode. */
   loadBalanced: boolean;
+  /** @experimental */
+  minPoolSizeCheckIntervalMS: number;
```
dariakp, line 94: I did not want to reference a background thread in the name, so `minPoolSizeCheckIntervalMS` is what I came up with, open to other ideas though

src/cmap/connection_pool.ts
```diff
@@ -90,6 +90,8 @@ export interface ConnectionPoolOptions extends Omit<ConnectionOptions, 'id' | 'g
   waitQueueTimeoutMS: number;
   /** If we are in load balancer mode. */
   loadBalanced: boolean;
+  /** @experimental */
```
dariakp, line 93: do we have a preference for marking this experimental or internal? when we release the next version, we may want to omit this item from the feature list or call it out as internal in the release notes, since users can't really set it directly anwyay

src/cmap/connection_pool.ts
nbbeeken: Yea 100 sounds good

src/cmap/connection_pool.ts
nbbeeken: The difference will be that internal will get removed from the public types, so I think that would be the better annotation if this isn't publicly configurable

src/cmap/connection_pool.ts
nbbeeken: Name is clear to me, and it is internal if it bothers us later we could change it

nbbeeken (approved):

test/integration/server-selection/server_selection.prose.operation_count.test.ts
```diff
@@ -27,19 +26,12 @@ async function runTaskGroup(collection: Collection, count: 10 | 100 | 1000) {
...
-  for (let i = 0; connectionCount < POOL_SIZE * 2 && i < 10; ++i) {
-    await promisify(setTimeout)(250);
-  }
-
-  client.removeListener('connectionCreated', onConnectionCreated);
-
-  if (connectionCount !== POOL_SIZE * 2) {
-    throw new Error('Connection pool did not fill up');
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for await (const _event of on(client, 'connectionCreated')) {
```
baileympearson, line 30: I like the refactor but I originally added the error to make it more clear what happened if the pool ever failed to fill up and the test timed out.

Maybe we could race this function against a timer if a better error message is something we care about?

baileympearson (changes requested):
Also commented on slack - `minPoolsizeCheckFrequencyMS` instead to match our existing options `heartbeatFrequencyMS` and `minHeartbeatFrequencyMS`?

test/integration/server-selection/server_selection.prose.operation_count.test.ts
dariakp: I can update the naming according to what we discussed, but I don't think the error here is important. If it times out before the beforeEach is satisfied, it times out; a beforeEach isn't the right place to assert on how long something should take, and, imho, this isn't really different from awaiting any other precondition in a beforeEach (e.g., connecting to a server)

nbbeeken (approved):

PR #3433, opened 2022-10-04
feat(NODE-3875): support recursive schema types

nbbeeken:
#### What is changing?

Adds a type check to limit recursion to 8 levels of depth. Mutually recursive types are now accepted and stop getting type checking / hinting after a certain depth.

Mutually recursive types can be useful for modeling data

### DIFF
--- a/src/mongo_types.ts
+++ b/src/mongo_types.ts
@@ -498,6 +498,13 @@ export type PropertyType<Type, Property extends string> = string extends Propert
  * @public
  * returns tuple of strings (keys to be joined on '.') that represent every path into a schema
  * https://docs.mongodb.com/manual/tutorial/query-embedded-documents/
+ *
+ * @remarks
+ * Through testing we determined that a depth of 8 is safe for the typescript compiler
+ * and provides reasonable compilation times. This number is otherwise not special and
+ * should be changed if issues are found with this level of checking. Beyond this
+ * depth any helpers that make use of NestedPaths should devolve to not asserting any
+ * type safety on the input.
  */
 export type NestedPaths<Type, Depth extends number[]> = Depth['length'] extends 8
   ? []
--- a/src/mongo_types.ts
+++ b/src/mongo_types.ts
@@ -272,7 +272,7 @@ export type MatchKeysAndValues<TSchema> = Readonly<
     [Property in `${NestedPathsOfType<TSchema, Record<string, any>[]>}.$${
       | `[${string}]`
       | ''}.${string}`]?: any; // Could be further narrowed
-  }
+  } & Document
 >;
 
 /** @public */
--- a/src/mongo_types.ts
+++ b/src/mongo_types.ts
@@ -68,7 +68,7 @@ export type WithoutId<TSchema> = Omit<TSchema, '_id'>;
 export type Filter<TSchema> =
   | Partial<TSchema>
   | ({
-      [Property in Join<NestedPaths<WithId<TSchema>>, '.'>]?: Condition<
+      [Property in Join<NestedPaths<WithId<TSchema>, []>, '.'>]?: Condition<
         PropertyType<WithId<TSchema>, Property>
       >;
     } & RootFilterOperators<WithId<TSchema>>);
@@ -263,7 +263,7 @@ export type OnlyFieldsOfType<TSchema, FieldType = any, AssignableType = FieldTyp
 /** @public */
 export type MatchKeysAndValues<TSchema> = Readonly<
   {
-    [Property in Join<NestedPaths<TSchema>, '.'>]?: PropertyType<TSchema, Property>;
+    [Property in Join<NestedPaths<TSchema, []>, '.'>]?: PropertyType<TSchema, Property>;
   } & {
     [Property in `${NestedPathsOfType<TSchema, any[]>}.$${`[${string}]` | ''}`]?: ArrayElement<
       PropertyType<TSchema, Property extends `${infer Key}.$${string}` ? Key : never>
@@ -499,19 +499,21 @@ export type PropertyType<Type, Property extends string> = string extends Propert
  * returns tuple of strings (keys to be joined on '.') that represent every path into a schema
  * https://docs.mongodb.com/manual/tutorial/query-embedded-documents/
  */
-export type NestedPaths<Type> = Type extends
-  | string
-  | number
-  | boolean
-  | Date
-  | RegExp
-  | Buffer
-  | Uint8Array
-  | ((...args: any[]) => any)
-  | { _bsontype: string }
+export type NestedPaths<Type, Depth extends number[]> = Depth['length'] extends 10
+  ? []
+  : Type extends
+      | string
+      | number
+      | boolean
+      | Date
+      | RegExp
+      | Buffer
+      | Uint8Array
+      | ((...args: any[]) => any)
+      | { _bsontype: string }
   ? []
   : Type extends ReadonlyArray<infer ArrayType>
-  ? [] | [number, ...NestedPaths<ArrayType>]
+  ? [] | [number, ...NestedPaths<ArrayType, [...Depth, 1]>]
   : Type extends Map<string, any>
   ? [string]
   : Type extends object
@@ -529,9 +531,9 @@ export type NestedPaths<Type> = Type extends
           ArrayType extends Type
           ? [Key] // we have a recursive array union
           : // child is an array, but it's not a recursive array
-            [Key, ...NestedPaths<Type[Key]>]
+            [Key, ...NestedPaths<Type[Key], [...Depth, 1]>]
         : // child is not structured the same as the parent
-          [Key, ...NestedPaths<Type[Key]>] | [Key];
+          [Key, ...NestedPaths<Type[Key], [...Depth, 1]>] | [Key];
     }[Extract<keyof Type, string>]
   : [];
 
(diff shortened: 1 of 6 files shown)


test/types/community/collection/updateX.test-d.ts
```diff
@@ -221,7 +221,10 @@ expectError<UpdateFilter<TestModel>>({
   $set: { 'subInterfaceField.nestedObject': { a: 1, b: '2' } }
 });
 expectError(buildUpdateFilter({ $set: { 'subInterfaceField.field2': 2 } }));
-expectError(buildUpdateFilter({ $set: { 'unknown.field': null } }));
+
+// NODE-3875 introduced intersection with Document to the MatchKeysAndValues so this no longer errors
+// expectError(buildUpdateFilter({ $set: { 'unknown.field': null } }));
```
baileympearson, line 226: Instead of commenting out the test, should we instead `expectAssignable`?

test/types/community/collection/updateX.test-d.ts
```diff
@@ -241,7 +244,10 @@ expectAssignable<UpdateFilter<TestModel>>({ $setOnInsert: { stringField: 'a' } }
 expectError(buildUpdateFilter({ $setOnInsert: { stringField: 123 } }));
 expectAssignable<UpdateFilter<TestModel>>({ $setOnInsert: { 'subInterfaceField.field1': '2' } });
 expectError(buildUpdateFilter({ $setOnInsert: { 'subInterfaceField.field2': 2 } }));
-expectError(buildUpdateFilter({ $setOnInsert: { 'unknown.field': null } }));
+
+// NODE-3875 introduced intersection with Document to the MatchKeysAndValues so this no longer errors
+// expectError(buildUpdateFilter({ $setOnInsert: { 'unknown.field': null } }));
```
baileympearson, line 249: same comment as above here

baileympearson (changes requested):

test/types/basic_schema.test-d.ts
```diff
@@ -20,7 +20,7 @@ expectType<Collection<ACounterWithId>>(new Collection<ACounterWithId>(db, ''));
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Simple Schema that does not define an _id
 // With _id
-type InsertOneArgOf<S> = Parameters<Collection<S>['insertOne']>[0];
+type InsertOneArgOf<S extends Document> = Parameters<Collection<S>['insertOne']>[0];
```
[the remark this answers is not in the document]
nbbeeken: My local version of tsd was updated which pulls in a new TS version to test against and this is a requirement since TS 4.8 figured I could just preemptively fix this.

test/types/community/collection/recursive-types.test-d.ts
```diff
@@ -254,15 +246,19 @@ expectAssignable<Filter<A>>({
   'b.c.d.a.b.c.d.a.b.name': 'a'
 });
 
-expectNotAssignable<Filter<A>>({
-  'b.c.d.a.b.c.d.a.b.name': 2
+// Beyond the depth supported, there is no type checking
+expectAssignable<Filter<A>>({
```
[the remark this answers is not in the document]
nbbeeken: Yes, fixed!

test/types/community/collection/updateX.test-d.ts
nbbeeken: Yea, agreed, fixed!

baileympearson (approved):

src/mongo_types.ts
```diff
@@ -499,19 +499,21 @@ export type PropertyType<Type, Property extends string> = string extends Propert
...
-  | string
-  | number
-  | boolean
-  | Date
-  | RegExp
-  | Buffer
-  | Uint8Array
-  | ((...args: any[]) => any)
-  | { _bsontype: string }
+export type NestedPaths<Type, Depth extends number[]> = Depth['length'] extends 10
```
dariakp, line 502: 1) Let's make sure we document why we're using whatever depth number we go with both here and in our public facing recursive type limitation docs

2) We may be able to pull in additional test examples from NODE-4594, or at least check to see how well this solution addresses the performance concerns raised there

dariakp (changes requested):

test/types/community/slow-model.test-d.ts
```diff
@@ -0,0 +1,412 @@
...
+  private slowCol: Collection<SlowModel>;
+
+  constructor(db: Db) {
+    this.circularCol = db.collection('circular');
+    this.slowCol = db.collection('slow');
+  }
+
+  public method = async (): Promise<void> => {
+    /*
+     * This call shows the base issue with recursive types
```
dariakp, line 399: we should add a comment at the top crediting the user with the test code

dariakp (changes requested):

test/types/community/slow-model.test-d.ts
nbbeeken: Good call! Added

test/types/community/collection/recursive-types.test-d.ts
```diff
@@ -1,23 +1,70 @@
...
-  b: {}
+expectAssignable<Filter<Author>>({
+  favoritePublication: {
+    title: 'book title',
+    author: {
+      name: 'author name'
+    }
+  }
+});
+expectNotType<UpdateFilter<Author>>({
```
dariakp, line 26: what's the purpose of this specific test?

test/types/community/collection/recursive-types.test-d.ts
```diff
@@ -1,23 +1,70 @@
...
+      author: {
+        name: 23
+      }
+    }
+  }
+});
+
+// Extremely deep type checking for recursive schemas
+expectNotAssignable<Filter<Author>>({
+  'favoritePublication.author.favoritePublication.author.favoritePublication.author.favoritePublication.title': 23
```
dariakp, line 40: can we note what the expected depth is that we're checking? because right now the test lengths seem arbitrary, like if the idea is to show every depth up to 10, then maybe we should number/present them in increasing order; otherwise going from 8 to 10 to 7 makes it hard to tell what the intention really is and makes it easy to introduce mistakes

for example, I am pretty sure L43 is supposed to check 8 and not 10, because at 10 it passes just because it's beyond the limit

and then in the one that checks 7 (L46), we should include a comparable passing example to demonstrate that it's indeed the type of the property that is the differentiating factor

test/types/community/collection/recursive-types.test-d.ts
```diff
@@ -1,23 +1,70 @@
...
+});
+expectAssignable<Filter<Author>>({
+  'favoritePublication.author.favoritePublication.author.favoritePublication.author.favoritePublication.author.favoritePublication.title':
+    'good soup'
+});
+expectNotAssignable<Filter<Author>>({
+  'favoritePublication.author.favoritePublication.author.favoritePublication.author.name': 23
+});
+
+// Beyond the depth of 10, `extends Document` permits anything (number for name is permitted)
```
dariakp, line 50: I think this should be "beyond 8", and the example modified accordingly

test/types/community/collection/recursive-types.test-d.ts
```diff
@@ -1,23 +1,70 @@
...
+expectAssignable<Filter<Author>>({
+  'favoritePublication.author.favoritePublication.author.favoritePublication.author.favoritePublication.author.favoritePublication.author.name': 23
+});
+
+// Update filter has similar depth limit
+expectAssignable<UpdateFilter<Author>>({
+  $set: {
+    'favoritePublication.author.favoritePublication.author.favoritePublication.author.favoritePublication.author.name':
```
dariakp, line 58: wouldn't 9 be not type checked? also, if update is "similar", then we should use the same test cases (-1 length in each) and be clear about how they are similar, i.e., it's one less because there is an extra level of inference

test/types/community/slow-model.test-d.ts
```diff
@@ -0,0 +1,419 @@
+/*
+ * The following type testing was adapted from @ermik
```
dariakp, line 2: should we create a release 4.11.0 ticket and call it out there so that we make sure we give credit in our release notes as well?

dariakp (changes requested):

test/types/community/collection/recursive-types.test-d.ts
nbbeeken: Added a comment, its just double checking that adding the `& Document` case doesn't lose checking when you aren't even using dot notation

test/types/community/collection/recursive-types.test-d.ts
nbbeeken: I added a test for every level up to 8 for both Filter and UpdateFilter, should make it clear what works and what doesn't. Thoughts?

test/types/community/collection/recursive-types.test-d.ts
nbbeeken: fixed with new tests

test/types/community/collection/recursive-types.test-d.ts
nbbeeken: I changed the prop to bestBook because I was honestly losing count of the depth due to long strings, but now we have a test for each level on both filter types.

test/types/community/slow-model.test-d.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-4718

test/types/community/collection/recursive-types.test-d.ts
dariakp: much easier to read!

test/types/community/collection/recursive-types.test-d.ts
```diff
@@ -0,0 +1,315 @@
...
+expectAssignable<Filter<Author>>({
+  bestBook: {
+    title: 'book title',
+    author: {
+      name: 'author name'
+    }
+  }
+});
+
+// Check that devolving to Document after a certain recursive depth does not effect checking
```
dariakp, line 27:
```suggestion
// Check that devolving to Document after a certain recursive depth does not affect checking
```

baileympearson (approved):

PR #3437, opened 2022-10-06
fix(NODE-4475): make interrupted message more specific

durran:
### Description

Fixes change stream retryability on 4.4 servers and lower when performing a failover.

#### What is changing?

This is the "quick and dirty" fix that just avoids pushing a `null` to the readable cursor stream before calling `destroy` on the stream when the error contains the message "interrupted at shutdown".

Existing spec tests cover the more specific error - see: https://github.com/mongodb/node-mongodb-native/blob/main/test/spec/retryable-writes/legacy/insertOne-serverErrors.yml#L468

Note: see https://github.com/mongodb/node-mongodb-native/pull/3438 for what a start of a refactor away from checking for this error message and creating separate streams for the change stream cursor and the others would look like. However it seems the "operation was interrupted" message affects change streams as well so I'll leave it as draft for now.

None

NODE-4475

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -895,14 +895,22 @@ class ReadableCursorStream extends Readable {
         }
 
         // NOTE: This is also perhaps questionable. The rationale here is that these errors tend
-        //       to be "operation interrupted", where a cursor has been closed but there is an
+        //       to be "operation was interrupted", where a cursor has been closed but there is an
         //       active getMore in-flight. This used to check if the cursor was killed but once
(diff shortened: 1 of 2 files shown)


baileympearson (changes requested):
From tech chat - let's try to add tests that

- confirm a change stream resumes if it encounters a resumable error with `interrupted` in the error message
- confirm that a change stream does not resume if it encounters an error with `operation was interrupted` as the message

Also - @durran want to file a follow up ticket to investigate removing the two special cases here?

dariakp (approved):

PR #3439, opened 2022-10-07
chore(NODE-4700): update dependencies

nbbeeken:
#### What is changing?

See package.json

- Pulling in the latest removes a bug in the linter on the leak_checker
- Aligns our tsd tests with the latest typescript that we compile with
- Removes the warning where eslint plugins complained about too new a TS version
- Pulls in two patches to our prod deps, socks5 and the connection string parser (safe bumps, good to pull in)


test/unit/assorted/imports.test.ts
```diff
@@ -15,7 +15,7 @@ function* walk(root) {
   }
 }
 
-describe.skip('importing mongodb driver', () => {
```
nbbeeken, line 18: @baileympearson I coulda swore I reversed this skip in the previous PR I mistakenly pushed the skip to. I skip it locally because its really slow when I want to rerun the whole suite of unit tests. Anyway, its still passing we should just flip this back.

baileympearson (approved):

PR #3440, opened 2022-10-07
refactor(NODE-4689): track checked out connections

baileympearson:
https://jira.mongodb.org/browse/NODE-4094

#### What is changing?

The connection pool now tracks checked out connections instead of only keeping track of the number of checked out connections.

No.

https://jira.mongodb.org/browse/NODE-4094

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -324,9 +324,11 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
   }
 
   /**
-   * This is exposed ONLY for use in mongosh, so they can
-   * kill connections if a user quits the shell with operations
-   * in progress.
+   * This is exposed ONLY for use in mongosh, to enable
+   * killing all connections if a user quits the shell with
+   * operations in progress.
+   *
+   * This property may be removed as a part of NODE-3263.
    */
   get checkedOutConnections() {
     return this[kCheckedOut];
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -324,6 +324,15 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
   }
 
   /**
+   * This is exposed ONLY for use on mongosh, so they can
+   * kill connections if a user quits the shell with operations
+   * in progress.
+   */
+  get checkedOutConnections() {
+    return this[kCheckedOut];
+  }
+
+  /**
    * Get the metrics information for the pool when a wait queue timeout occurs.
    */
   private waitQueueErrorMetrics(): string {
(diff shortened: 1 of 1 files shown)


src/cmap/connection_pool.ts
```diff
@@ -252,7 +252,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     this[kLogger] = new Logger('ConnectionPool');
     this[kConnections] = new Denque();
     this[kPending] = 0;
-    this[kCheckedOut] = 0;
+    this[kCheckedOut] = new Set();
```
jyemin, line 255: Since this is of type `Set`, I'm just curious how equality is defined for connections.  Is it just reference equality?

baileympearson: The `Set` will use referential equality for the connections, yes.

src/cmap/connection_pool.ts
```diff
@@ -323,6 +323,15 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     return this[kServer].description.error;
   }
 
+  /**
+   * This is exposed ONLY for use in mongosh, so they can
+   * kill connections if a user quits the shell with operations
+   * in progress.
+   */
+  get checkedOutConnections() {
```
dariakp, line 331: should we mark this `@internal` as we do with the non-getter properties on the pool? actually, looking at all the stuff currently marked internal, I'm not sure what our reasoning was for explicitly marking things vs not, but if we're going to mark anything on the pool as not intended for customer use, this property should definitely be included; @nbbeeken - any recollection on the context for the things currently marked internal?

src/cmap/connection_pool.ts
```diff
@@ -323,6 +323,15 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     return this[kServer].description.error;
   }
 
+  /**
+   * This is exposed ONLY for use in mongosh, so they can
```
dariakp, line 327: this is kind of nitpicky since this is currently removed from our generated API docs, but since these code comments are still in a way part of our public facing documentation, can we rephrase the sentence to use passive voice (e.g., "so they can kill connections" -> "to enable killing all connections") and also reference NODE-3263 (perhaps in a separate TODO comment)? Unless we are 100% confident that we aren't going to do NODE-3263 in any form, this getter might go away, or might be something that isn't used in the shell but is used by us instead to perform similar functionality on, say, `close(force = true)`

dariakp (changes requested):
Just a couple of points for discussion

src/cmap/connection_pool.ts
baileympearson: The `ConnectionPool` as a whole is marked internal, so the class as a whole is non-public and marking properties as internal isn't necessary.

I can mark this property as internal and I don't feel strongly but I'm not sure which is preferable - marking it internal because we definitely don't want someone to use it or not marking it internal for consistency with the other getters.  Thoughts?

nbbeeken (approved):

PR #3443, opened 2022-10-10
refactor(NODE-4690): update cmap and test runners to support interruptInUseConnections

baileympearson:
#### What is changing?

This PR does the following:

- syncs new unified and cmap unit tests from [this](https://github.com/mongodb/specifications/commit/23bd4db8fcfc1dbd5d5c123f936134913720e85b) spec change
- updates the cmap runner to support passing `interruptInUseConnections` to `pool.clear()`
- updates the unified runner to support `interruptInUseConnections`
- Updates the PoolClearedEvent to support `interruptInUseConnections` (necessary for the tests)

No.

### DIFF
--- a/src/cmap/connection_pool_events.ts
+++ b/src/cmap/connection_pool_events.ts
@@ -192,13 +192,10 @@ export class ConnectionPoolClearedEvent extends ConnectionPoolMonitoringEvent {
   /** @internal */
   constructor(
     pool: ConnectionPool,
-    {
-      serviceId,
-      interruptInUseConnections
-    }: { serviceId?: ObjectId; interruptInUseConnections?: boolean } = {}
+    options: { serviceId?: ObjectId; interruptInUseConnections?: boolean } = {}
   ) {
     super(pool);
-    this.serviceId = serviceId;
-    this.interruptInUseConnections = interruptInUseConnections;
+    this.serviceId = options.serviceId;
+    this.interruptInUseConnections = options.interruptInUseConnections;
   }
 }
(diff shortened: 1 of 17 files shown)


src/cmap/connection_pool_events.ts
```diff
@@ -187,9 +187,18 @@ export class ConnectionPoolClearedEvent extends ConnectionPoolMonitoringEvent {
...
+  interruptInUseConnections?: boolean;
+
   /** @internal */
-  constructor(pool: ConnectionPool, serviceId?: ObjectId) {
+  constructor(
+    pool: ConnectionPool,
+    {
+      serviceId,
+      interruptInUseConnections
+    }: { serviceId?: ObjectId; interruptInUseConnections?: boolean } = {}
```
baileympearson, line 198: This is a new pattern for command event constructors but I think here it's okay - without it, we'd we be forced to add in optional argument handling logic in the constructor.  I'd rather use an object.

baileympearson, line 198: I'm open to refactoring this if people want.  This is the only command monitoring event that now takes an object instead of standalone parameters.  I chose this approach even though it deviates from our convention because this approach allows us not to worry about optional argument handling.

durran: We do have another CMAP event that does pass an object: https://github.com/mongodb/node-mongodb-native/blob/d989f44b0b6089ad7e9389997e34dc005392c69e/src/cmap/connection_pool_events.ts#L74 - so my preference would be to at least keep them in line which ever way we choose since one is a named object and the other is not.

durran (approved):

nbbeeken (approved):
CI looks concerning but unrelated to this work LGTM

PR #3444, opened 2022-10-11
fix: Support template and number keys in schemas

GrantGryczan:

### DIFF
--- a/src/mongo_types.ts
+++ b/src/mongo_types.ts
@@ -269,7 +269,7 @@ export type MatchKeysAndValues<TSchema> = Readonly<
       PropertyType<TSchema, Property extends `${infer Key}.$${string}` ? Key : never>
     >;
   } & {
-    [Property in `${NestedPathsOfType<TSchema, Record<string, any>[]>}.$${
+    [Property in `${NestedPathsOfType<TSchema, Record<string | number, any>[]>}.$${
       | `[${string}]`
       | ''}.${string}`]?: any; // Could be further narrowed
   }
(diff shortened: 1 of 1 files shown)


GrantGryczan:
Question: Must I open an issue on Jira in order to fix it in a PR?

src/mongo_types.ts
```diff
@@ -516,23 +520,23 @@ export type NestedPaths<Type> = Type extends
   ? [string]
   : Type extends object
   ? {
-      [Key in Extract<keyof Type, string>]: Type[Key] extends Type // type of value extends the parent
-        ? [Key]
-        : // for a recursive union type, the child will never extend the parent type.
-        // but the parent will still extend the child
-        Type extends Type[Key]
```
GrantGryczan, line 523: This prevented template string keys from working because `` { a: { `example-${number}`: 1 } } extends { `example-${number}`: 1 } ``. This fact is exclusive to records with template string keys because template strings allow for infinite possible key values, which informs TypeScript that it should allow any keys on the object.

src/mongo_types.ts
```diff
@@ -516,23 +520,23 @@ export type NestedPaths<Type> = Type extends
...
-        ? [Key]
```
GrantGryczan, line 524: This prevented template string keys from working because `` { a: Record<`example-${number}`, any> } extends Record<`example-${number}`, any> ``, which means this would output `['a']` instead of `['a', `example-${number}`]`, despite there being no recursion there.

This fact is exclusive to records with template string keys because template strings allow for infinite possible key values, which apparently informs TypeScript that it should allow any keys on the object.

That's why I had to change some of the recursive typing support here.

baileympearson:
Hey @GrantGryczan, thanks for the contribution.  At this time, we're not accepting changes to our types.  After https://github.com/mongodb/node-mongodb-native/pull/3433 merges, we plan on leaving TS as-is in the 4.x versions of the driver, and are considering the level of Typescript support we offer in the next major version.  We'll keep this PR in mind for the future, in case we consider reintroducing support for dot notation in the future.

GrantGryczan:
Thanks for letting me know @baileympearson, I appreciate your consideration. Though I'm very concerned because this bug is giving me countless type errors in my project, as I heavily use `number` and template string keys in my `Record` types. Will those type errors go away after the aforementioned merge? Or if not, should I try to fix this again after the merge?

GrantGryczan:
A couple days and still no response, so I'll ask again: what should I do in order for this package to be usable for me? Should I submit a new PR now that #3433 has merged, or just keep using my unmaintainable fork of this repo?

PR #3446, opened 2022-10-13
feat(NODE-4721): add aws-sdk as optional dependency

durran:
### Description

Uses the AWS-SDK for finding credentials when present in dependencies, otherwise uses the internal implementation.

#### What is changing?

- Adds @aws-sdk/credential-providers as an optional dependency.
- Uses the SDK when installed to find credentials from the environment when missing.

Patch build with AWS tests for with and without the optional dependencies: 

https://spruce.mongodb.com/version/634d68f92fbabe2d0a6b964f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Can make a note to users that they can install the SDK to have it be used.

NODE-4721

### DIFF
--- a/.evergreen/generate_evergreen_tasks.js
+++ b/.evergreen/generate_evergreen_tasks.js
@@ -308,7 +308,7 @@ const AWS_AUTH_TASKS = [];
 
 for (const VERSION of AWS_AUTH_VERSIONS) {
   const name = ex => `aws-${VERSION}-auth-test-${ex.split(' ').join('-')}`;
-  const aws_funcs = [
+  const awsFuncs = [
     { func: 'run aws auth test with regular aws credentials' },
     { func: 'run aws auth test with assume role credentials' },
     { func: 'run aws auth test with aws EC2 credentials' },
@@ -317,7 +317,7 @@ for (const VERSION of AWS_AUTH_VERSIONS) {
     { func: 'run aws ECS auth test' }
   ];
 
-  const aws_tasks = aws_funcs.map(fn => ({
+  const awsTasks = awsFuncs.map(fn => ({
     name: name(fn.func),
     commands: [
       { func: 'install dependencies' },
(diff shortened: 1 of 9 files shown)


.evergreen/generate_evergreen_tasks.js
```diff
@@ -336,8 +336,34 @@ for (const VERSION of AWS_AUTH_VERSIONS) {
...
 
-  TASKS.push(...aws_tasks);
-  AWS_AUTH_TASKS.push(...aws_tasks.map(t => t.name));
+  const awsNoOptionalTasks = awsFuncs.map(fn => ({
+    name: `${name(fn.func)}-no-optional`,
+    commands: [
+      {
+        func: 'install dependencies',
+        vars: {
+          NPM_OPTIONS: '--no-optional'
```
durran, line 345: We run the aws tests twice each now, once with the sdk installed and once without.

.evergreen/generate_evergreen_tasks.js
```diff
@@ -308,7 +308,7 @@ const AWS_AUTH_TASKS = [];
 
 for (const VERSION of AWS_AUTH_VERSIONS) {
   const name = ex => `aws-${VERSION}-auth-test-${ex.split(' ').join('-')}`;
-  const aws_funcs = [
+  const awsFuncs = [
```
durran, line 311: Just changing name to JS conventions.

nbbeeken (approved):
LGTM, I think we have really good coverage here, thanks for the CI adjustments. I think we will want to put a note about this in the Manual docs, so I did mark the ticket as docs changes needed.

avaly:
The install size of the `mongodb` package doubled in size in the last version: https://packagephobia.com/result?p=mongodb

Mostly because of this new dependency being added.

Can we please remove it from this package and instruct users to add this dependency in their project, if needed?

PR #3453, opened 2022-10-20
fix(NODE-4735): fix change stream consecutive resumabilty

baileympearson:
#### What is changing?

Change streams should loop until a document is produced, so that in the case of a successful resume, a subsequent resumable error is treated as its own resumable error.

No.

Correct change stream resume behavior.

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -648,7 +648,7 @@ export class ChangeStream<
   hasNext(callback?: Callback): Promise<boolean> | void {
     this._setIsIterator();
     return maybeCallback(async () => {
-      for (;;) {
+      while(true) {
         try {
           const hasNext = await this.cursor.hasNext();
           return hasNext;
@@ -675,7 +675,7 @@ export class ChangeStream<
   next(callback?: Callback<TChange>): Promise<TChange> | void {
     this._setIsIterator();
     return maybeCallback(async () => {
-      for (;;) {
+      while(true) {
         try {
           const change = await this.cursor.next();
           const processedChange = this._processChange(change ?? null);
@@ -705,7 +705,7 @@ export class ChangeStream<
   tryNext(callback?: Callback<Document | null>): Promise<Document | null> | void {
     this._setIsIterator();
     return maybeCallback(async () => {
-      for (;;) {
+      while(true) {
         try {
           const change = await this.cursor.tryNext();
           return change ?? null;
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -648,21 +648,21 @@ export class ChangeStream<
   hasNext(callback?: Callback): Promise<boolean> | void {
     this._setIsIterator();
     return maybeCallback(async () => {
-      try {
-        const hasNext = await this.cursor.hasNext();
-        return hasNext;
-      } catch (error) {
+      for (;;) {
         try {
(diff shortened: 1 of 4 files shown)


src/change_stream.ts
```diff
@@ -648,21 +648,21 @@ export class ChangeStream<
   hasNext(callback?: Callback): Promise<boolean> | void {
     this._setIsIterator();
     return maybeCallback(async () => {
-      try {
-        const hasNext = await this.cursor.hasNext();
-        return hasNext;
-      } catch (error) {
+      for (;;) {
```
baileympearson, line 651:
```suggestion
      while(true) {
```

src/change_stream.ts
```diff
@@ -675,23 +675,22 @@ export class ChangeStream<
   next(callback?: Callback<TChange>): Promise<TChange> | void {
     this._setIsIterator();
     return maybeCallback(async () => {
-      try {
-        const change = await this.cursor.next();
-        const processedChange = this._processChange(change ?? null);
-        return processedChange;
-      } catch (error) {
+      for (;;) {
```
baileympearson, line 678:
```suggestion
      while(true) {
```

src/change_stream.ts
```diff
@@ -706,21 +705,21 @@ export class ChangeStream<
   tryNext(callback?: Callback<Document | null>): Promise<Document | null> | void {
     this._setIsIterator();
     return maybeCallback(async () => {
-      try {
-        const change = await this.cursor.tryNext();
-        return change ?? null;
-      } catch (error) {
+      for (;;) {
```
baileympearson, line 708:
```suggestion
      while(true) {
```

src/change_stream.ts
dariakp, line 651: This is less obvious than a while true loop, so unless there's a particularly good reason to use this, I'd prefer the while loop (yes even with the lint warning, maybe even with a comment describing why this is definitely intentional)

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1767,7 +1718,44 @@ describe('ChangeStream resumability', function () {
             expect(aggregateEvents).to.have.lengthOf(2);
           }
         );
+
+        it(
+          `supports consecutive resumes on error code ${code} ${error}`,
```
dariakp, line 1723: Just a note that these test additions will likely result in merge conflicts for https://github.com/mongodb/node-mongodb-native/pull/3454 and we should be extra careful in that PR, once it's rebased, that all these tests are pulled in

src/change_stream.ts
```diff
@@ -648,21 +648,25 @@ export class ChangeStream<
   hasNext(callback?: Callback): Promise<boolean> | void {
     this._setIsIterator();
     return maybeCallback(async () => {
-      try {
-        const hasNext = await this.cursor.hasNext();
-        return hasNext;
-      } catch (error) {
+      // Change streams must resume indefinitely as long while each resume event succeeds.
```
dariakp, line 651: Not sure the best way to fix "as long while", but it doesn't work semantically at the moment

dariakp (changes requested):

dariakp (approved):

nbbeeken (approved):
LGTM

PR #3454, opened 2022-10-24
feat(NODE-4683): make ChangeStream an async iterable

andymina:
#### What is changing?

Adds the `asyncIterator` symbol to the ChangeStream class.

Yes.

Currently, Change streams support manual promise-based iteration using `next`, `hasNext` and `tryNext`.  Making the ChangeStream an async iterable allows for idiomatic use in for-await loops.

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -737,7 +737,8 @@ export class ChangeStream<
     }, callback);
   }
 
-  async *[Symbol.asyncIterator](): AsyncIterator<TChange, void> {
+  // TODO(andymina): ask about never as third template parameter
+  async *[Symbol.asyncIterator](): AsyncGenerator<TChange, void, never> {
     if (this.closed) {
       return;
     }
@@ -747,7 +748,11 @@ export class ChangeStream<
         yield await this.next();
       }
     } finally {
-      await this.close();
+      try {
+        await this.close();
+      } catch (error) {
+        // we're not concerned with errors from close()
+      }
     }
   }
 
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -737,13 +737,15 @@ export class ChangeStream<
     }, callback);
   }
 
-  async *[Symbol.asyncIterator](): AsyncGenerator<TChange, void, never> {
+  async *[Symbol.asyncIterator](): AsyncGenerator<TChange, void, void> {
     if (this.closed) {
       return;
     }
 
     try {
-      while (await this.hasNext()) {
+      // Change streams run indefinitely as long as errors are resumable
+      // So the only loop breaking condition is if `next()` throws
+      while (true) {
         yield await this.next();
       }
     } finally {
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -737,7 +737,6 @@ export class ChangeStream<
     }, callback);
   }
 
-  // TODO(andymina): ask about never as third template parameter
   async *[Symbol.asyncIterator](): AsyncGenerator<TChange, void, never> {
     if (this.closed) {
       return;
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -742,11 +742,7 @@ export class ChangeStream<
       return;
     }
 
-    while (true) {
-      if (!(await this.hasNext())) {
-        break;
-      }
-
+    while (await this.hasNext()) {
       yield await this.next();
     }
   }
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -751,7 +751,7 @@ export class ChangeStream<
     } finally {
       try {
         await this.close();
-      } catch (error) {
+      } catch {
         // we're not concerned with errors from close()
       }
     }
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -742,8 +742,12 @@ export class ChangeStream<
       return;
     }
 
-    while (await this.hasNext()) {
-      yield await this.next();
+    try {
+      while (await this.hasNext()) {
+        yield await this.next();
+      }
+    } finally {
+      await this.close();
     }
   }
 
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -737,26 +737,18 @@ export class ChangeStream<
     }, callback);
   }
 
-  [Symbol.asyncIterator](): AsyncIterator<TChange, void> {
-    async function* nativeAsyncIterator(this: ChangeStream<TSchema, TChange>) {
-      if (this.closed) {
-        return;
-      }
-
-      while (true) {
-        if (!(await this.hasNext())) {
-          break;
-        }
-
-        yield await this.next();
-      }
+  async *[Symbol.asyncIterator](): AsyncIterator<TChange, void> {
+    if (this.closed) {
+      return;
     }
 
-    const iterator = nativeAsyncIterator.call(this);
+    while (true) {
+      if (!(await this.hasNext())) {
+        break;
+      }
 
-    return {
-      next: () => iterator.next()
-    };
+      yield await this.next();
+    }
   }
 
   /** Is the cursor closed */
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -737,6 +737,28 @@ export class ChangeStream<
     }, callback);
   }
 
+  [Symbol.asyncIterator](): AsyncIterator<TChange, void> {
+    async function* nativeAsyncIterator(this: ChangeStream<TSchema, TChange>) {
+      if (this.closed) {
+        return;
+      }
+
+      while (true) {
+        if (!(await this.hasNext())) {
+          break;
+        }
+
+        yield await this.next();
+      }
+    }
+
+    const iterator = nativeAsyncIterator.call(this);
+
+    return {
+      next: () => iterator.next()
+    };
+  }
+
   /** Is the cursor closed */
   get closed(): boolean {
     return this[kClosed] || this.cursor.closed;
--- a/test/integration/change-streams/change_stream.test.ts
+++ b/test/integration/change-streams/change_stream.test.ts
@@ -10,7 +10,6 @@ import { promisify } from 'util';
 import {
   AbstractCursor,
   ChangeStream,
-  ChangeStreamDocument,
   ChangeStreamOptions,
   Collection,
   CommandStartedEvent,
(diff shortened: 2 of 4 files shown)


src/change_stream.ts
```diff
@@ -726,6 +726,16 @@ export class ChangeStream<
     }, callback);
   }
 
+  async *[Symbol.asyncIterator](): AsyncIterator<TChange, void> {
+    if (this.closed) {
+      return;
+    }
+
+    while (await this.hasNext()) {
```
baileympearson, line 734: sorry I didn't consider this earlier, but I think we should add a `try-finally` block here.  We don't need the `catch`, because we want to propagate the error, but in the `finally`, we should close the change stream.  This would cleanup the change stream automatically if a user `break`s out of the for-await loop.

This introduces some extra test cases as well

- when there is an error thrown, we should confirm that the change stream is closed
- when if `return` is called on the async iterator, then the change stream is closed

src/change_stream.ts
```diff
@@ -726,6 +726,16 @@ export class ChangeStream<
     }, callback);
   }
 
+  async *[Symbol.asyncIterator](): AsyncIterator<TChange, void> {
```
baileympearson, line 729: Can we add a test that demonstrates that when we attempt to iterate a closed change stream, nothing happens?

test/integration/change-streams/change_stream.test.ts
```diff
@@ -2138,6 +2138,185 @@ describe('ChangeStream resumability', function () {
         );
       });
     });
+
```
baileympearson, line 2141: Two other important tests:

1. we should demonstrate that we cannot use Symbol.asyncIterator in conjunction with the event emitter based iteration. So the following should throw an error (and I believe it already does):

```typescript
const cs = collection.watch([]);
cs.on('change', change => { .... });

for await (const change of cs) { // this should throw

}
```

2. we should demonstrate that we can use Symbol.asyncIterator in conjunction with the raw iterator based api:
```typescript
const cs = collection.watch([]);

await cs.next();  // iterate a bit using the iterator API

for await (const change of cs) { // this should not throw

}
```

test/integration/change-streams/change_stream.test.ts
```diff
@@ -2138,6 +2138,185 @@ describe('ChangeStream resumability', function () {
...
+            await collection.insertOne({ city: 'New York City' });
+
+            let total_changes = 0;
+            for await (const change of changeStream) {
+              total_changes++;
+              if (total_changes === 1) {
+                changeStream.close();
+              }
+            }
```
baileympearson, line 2171: for-await in conjunction with change streams is beautiful 🔥

test/integration/change-streams/change_stream.test.ts
```diff
@@ -2138,6 +2138,185 @@ describe('ChangeStream resumability', function () {
...
+              data: {
+                failCommands: ['getMore'],
+                errorCode: code,
+                errmsg: message
+              }
+            } as FailPoint);
+
+            await collection.insertOne({ city: 'New York City' });
+
+            let total_changes = 0;
```
baileympearson, line 2165: I know I recommended this pattern so sorry for changing my mind 😬.  I think we should manually iterate the iterator:

```typescript
const changeStreamIterator = changeStream[Symbol.asyncIterator]()
await changeStreamIterator.next();
```

This also would clean up the code a bit, since we wouldn't need to manually keep track of the count 🙂

src/change_stream.ts
andymina: Good catch, makes total sense.

test/integration/change-streams/change_stream.test.ts
andymina: All good, this is a lot cleaner and easier to maintain anyways.

test/integration/change-streams/change_stream.test.ts
```diff
@@ -948,31 +952,168 @@ describe('Change Streams', function () {
...
+          for (const doc of docs) {
+            const change = await changeStreamIterator.next();
+            const { fullDocument } = change.value;
+            expect(fullDocument.city).to.equal(doc.city);
+          }
 
-        const thirdTry = await changeStream.tryNext();
-        expect(thirdTry).to.be.null;
-      }
+          changeStream.close();
```
baileympearson, line 1002: Question and comment.  

Question: is this necessary?  I think the afterEach hook at the top of the file closes the change stream after each test.
Comment: make sure you're awaiting async operations!  ChangeStream.close() is async, so if you do need to close it here, make sure you await it.

test/integration/change-streams/change_stream.test.ts
```diff
@@ -948,31 +952,168 @@ describe('Change Streams', function () {
...
+          changeStream = collection.watch([]);
+          await initIteratorMode(changeStream);
+
+          const docs = [{ city: 'New York City' }, { city: 'Seattle' }, { city: 'Boston' }];
+          await collection.insertMany(docs);
+
+          const changeStreamAsyncIteratorHelper = async function (changeStream: ChangeStream) {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            for await (const change of changeStream) {
+              return;
```
baileympearson, line 1019: I like the async function helper you're using to test `return`!  Clever.  

If you'd like, an alternative approach could be to simply call `return` on the object.

```typescript
const iterator = changeStream[Symbol.asyncIterator]();
await iterator.return();
```

This is optional though, whichever you'd prefer.

test/integration/change-streams/change_stream.test.ts
```diff
@@ -2138,6 +2276,144 @@ describe('ChangeStream resumability', function () {
...
+                errmsg: message
+              }
+            } as FailPoint);
+
+            await collection.insertOne({ city: 'New York City' });
+            await changeStreamIterator.next();
+
+            expect(aggregateEvents).to.have.lengthOf(2);
+
+            changeStream.close();
```
baileympearson, line 2307: same comment about closing the change stream as above.  this applies to all tests you've added actually

baileympearson (changes requested):
Looks good!  As discussed in person, we're going to add a test to demonstrate that when ChangeStream.close() throws inside the `finally` block, we swallow the error.

test/integration/change-streams/change_stream.test.ts
andymina: didn't know this option existed 😅 , its a lot cleaner so i went with it.

test/integration/change-streams/change_stream.test.ts
```diff
@@ -948,31 +953,178 @@ describe('Change Streams', function () {
...
-        const firstTry = await changeStream.tryNext();
-        expect(firstTry).to.be.null;
+    describe('#asyncIterator', function () {
+      it(
+        'can iterate through changes',
+        { requires: { topology: '!single', mongodb: '>=4.2' } },
+        async function () {
+          changeStream = collection.watch([]);
+          await initIteratorMode(changeStream);
+          const changeStreamIterator = changeStream[Symbol.asyncIterator]();
```
dariakp, line 992: This seems like an odd way of testing the iterator; the goal in the ticket says "Making the ChangeStream an async iterable would allow idiomatic use in for-await loops." Instead of manually calling the next function of the iterator, we should be testing this the way it's intended to be used by our users,
```
for (const change of await changeStream) {
   const { fullDocument } = change.value;
   const expectedDoc = docs.shift();
   expect(fullDocument.city).to.equal(expectedDoc.city);
   if (docs.length === 0) {
     break;
   }
}
expect(docs).to.have.length(0, 'expected to find all docs before exiting loop');
// ^ just a sanity check to make sure we didn't early-exit without error
```

and this goes for all iterator test cases in this file

dariakp (changes requested):

test/integration/change-streams/change_stream.test.ts
baileympearson: @dariakp Andy wrote the tests this way based on input from me. Andy originally had the tests like you're suggesting but I suggested this approach.

The async iterator method returns a generator.  A for-await loop is just syntactic sugar over manual iteration. I think it's preferable to describe the expected behavior of our generator in the tests by testing each method. The for-await loop will work automatically. 

This also cleans up tests too. Some of the error tests had additional, impossible conditions that we needed to consider when using a for-await loop. For example, consider the test that confirms that when an error is thrown internally, the error is thrown. Originally, we used a for-await loop but the test had an empty loop body (because the first iteration would throw an error and we would never enter the loop), so the options were to either lint the warning away or add an expect.fail.

This approach is 
- More granular
- easier to work with in many cases and requires no breaking from loops
- Completely equivalent to testing using for-await loops

dariakp: What you are describing makes a case for unit testing the behavior of `next()` in the new iterator, and I am certainly not going to object to unit tests being added for this purpose (or these tests being moved or copied to the unit test directory with some adjustments). However, we are talking about integration tests, not unit tests, and we do need to have integration tests that cover *actual use cases* of the feature being implemented. Iterating for-of, and, in particular, breaking out of for-of loops will have very specific behaviors depending on how the iterator is implemented. If we decide to manually create the async iterator or iterable in the future, our integration tests should still be valid. Same if Node.js decides to change exactly how the for-of uses the next() function, or introduces new iteration protocols that we want to use instead of this one. The bottom line is that the feature isn't whether `Symbol.asyncIterator` is defined, it's that the for await of loop works.

References:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of#iterating_over_sync_iterables_and_generators
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols
- NODE-4598

dariakp: Also, just wanted to add that `expect.fail` actually would be the appropriate thing for an integration test in the error case, because, again, we are testing what our users would expect to see in such a case, and it informs how we document suggested error handling to our users

nbbeeken:
Just a reminder to rebase this on main, the linked PR Daria mentioned just above ^ is now merged and has relevant tests that should be pulled in here.

src/change_stream.ts
```diff
@@ -726,6 +726,24 @@ export class ChangeStream<
...
 
+  async *[Symbol.asyncIterator](): AsyncGenerator<TChange, void, never> {
+    if (this.closed) {
+      return;
+    }
+
+    try {
+      while (await this.hasNext()) {
+        yield await this.next();
+      }
```
nbbeeken, line 737:
```suggestion
      // Change streams run indefinitely as long as errors are resumable
      // So the only loop breaking condition is if `next()` throws
      // eslint-disable-next-line no-constant-condition
      while (true) {
        yield await this.next();
      }
```

src/change_stream.ts
```diff
@@ -726,6 +726,24 @@ export class ChangeStream<
     }, callback);
   }
 
+  async *[Symbol.asyncIterator](): AsyncGenerator<TChange, void, never> {
```
nbbeeken, line 729:
```suggestion
  async *[Symbol.asyncIterator](): AsyncGenerator<TChange, void, void> {
```

I thought typescript warned against us using `never` here, can we add a type test to `test/types/change_stream.test-d.ts` that just uses the for-await syntax

nbbeeken (changes requested):

test/integration/change-streams/change_stream.test.ts
baileympearson: The feature isn't only about for-await loops.  Async generators can be used in other ways:

- https://github.com/ReactiveX/IxJS, a popular functional programming library can now be used with change streams
- change streams can now be used as streams natively with Readable.from

The ticket you linked is one example (for-await) but the actual feature is the async generator we now support.

as far as the tests on this PR go, if you'd like tests using a for-await loop we can add them but I think there's value in testing each part of the async generator individually.

baileympearson (changes requested):

test/integration/change-streams/change_stream.test.ts
dariakp: From slack:
- We'll use for..await for the main integration test cases (full iteration without errors and with resumable errors; manual break from loop and then creating a new loop; and encountering non-resumable errors)
- There is value in explicitly testing the `asyncIterable`, so we can keep tests that provide additional coverage to that end, with anything that calls for mocking to unit tests

src/change_stream.ts
```diff
@@ -737,6 +737,26 @@ export class ChangeStream<
...
+    try {
+      // Change streams run indefinitely as long as errors are resumable
+      // So the only loop breaking condition is if `next()` throws
+      while (true) {
+        yield await this.next();
+      }
+    } finally {
+      try {
+        await this.close();
+      } catch (error) {
```
nbbeeken, line 754:
```suggestion
      } catch {
```

nbbeeken (changes requested):

src/change_stream.ts
baileympearson: I'm surprised the linter didn't complain about this

test/integration/change-streams/change_stream.test.ts
andymina: Implemented!

dariakp (approved):
Thanks for the revisions, looking good!

There are a few places where the assertions could be improved, but I think this is overall good enough to merge (I don't think it's worth the extra CI cycle to make the adjustment).

For example, going forward, using something like
`expect(changeStream.cursor).property('closed', true);`
or even
`expect(changeStream).nested.property('cursor.closed', true);`
instead of
`expect(changeStream.cursor.closed).to.be.true;`
would give better error printouts in the logs

dariakp (changes requested):
Just need to fix the one unit test failure

nbbeeken (approved):

nbbeeken (approved):

dariakp (approved):

PR #3457, opened 2022-10-27
WIP interrupt in-flight operations on heartbeat failure 

baileympearson:


src/cmap/connection_pool.ts
```diff
@@ -572,7 +633,12 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     return !!(this.options.maxIdleTimeMS && connection.idleTime > this.options.maxIdleTimeMS);
   }
 
-  private connectionIsPerished(connection: Connection) {
+  /**
+   * Destroys a connection if the connection is perished.
+   *
+   * @returns `true` if the connection was destroyed, `false` otherwise.
+   */
+  private destroyConnectionIfPerished(connection: Connection) {
```
baileympearson, line 641: minor code cleanup here - the name was confusing since it seemed to indicate it was a pure function returning true/false

src/cmap/errors.ts
```diff
@@ -27,10 +27,11 @@ export class PoolClearedError extends MongoNetworkError {
   /** The address of the connection pool */
   address: string;
 
-  constructor(pool: ConnectionPool) {
-    super(
-      `Connection pool for ${pool.address} was cleared because another operation failed with: "${pool.serverError?.message}"`
-    );
+  constructor(pool: ConnectionPool, message?: string) {
```
baileympearson, line 30: PoolClearedError is not exported (but it's not marked internal), so it should be okay to modify the constructor.

src/cmap/connection_pool.ts
```diff
@@ -432,18 +439,72 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
   }
 
+  /**
+   * Closes all checked in perished connections in the pool with a resumable PoolClearedOnNetworkError.
+   *
+   * If interruptInUseConnections is `true`, this method attempts to kill checked out connections as well.
+   * Only connections where `connection.generation <= minGeneration` are killed.  Connections are closed with a
+   * resumable PoolClearedOnNetworkTimeoutError.
+   */
+  private pruneConnections({
```
dariakp, line 471: why does this method do anything with the connections that aren't checked out? the spec doesn't say anything about aggressively removing stale connections from the pool (in fact, it says the opposite, that they should be "lazily" closed), it's only concerned about **in use** connections and failing any operations that might be using them as quickly as possible. Our logic already ensures that a stale connection will never be handed to an operation, so the logic in lines 478-489 is unnecessary (and also incorrect, since you'd be actually applying an an in use error to connections that are not in use). I suggest renaming this method to some variation of "interruptInUseConnections" (maybe "interruptActiveConnections" to refer to the in use and pending ones), and guard L457-459 with an if check on `interruptInUseConnections`

dariakp (changes requested):

src/cmap/connection_pool.ts
```diff
@@ -432,18 +439,72 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       return;
     }
 
+    const oldGeneration = this[kGeneration];
```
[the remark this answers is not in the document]
baileympearson: Just to confirm - the ask is simply to move line 442 below the comment?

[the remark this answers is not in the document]
dariakp: yes, and to move the lines at the top of the function into the loadbalanced comment so that we could eventually separate these functions entirely; actually separating them into two separate functions does not have to happen here (but could, if you were so inclined, the pros are that it would make this separation of concerns cleaner, the cons is that it's a somewhat bigger refactor)

src/cmap/errors.ts
```diff
@@ -39,6 +40,22 @@ export class PoolClearedError extends MongoNetworkError {
   }
 }
 
+/**
+ * An error indicating that a connection pool has been cleared after the monitor for that server timed out.
+ * @category Error
+ */
+export class PoolClearedOnNetworkError extends PoolClearedError {
```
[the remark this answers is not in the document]
baileympearson: 1. specificity in errors
2. simplicity in implementation.  using the PoolClearedError would require either refactoring the PoolClearedError to have logic in the constructor to correctly set the message, or hard coding the message in each place we created a pool cleared error.  The custom error doesn't have this issue
3. the ability to add a RetryableWriteError label in the constructor

test/integration/connection-monitoring-and-pooling/connection_monitoring_and_pooling.spec.test.ts
```diff
@@ -19,27 +19,28 @@ const LB_SKIP_TESTS: SkipDescription[] = [
   skipReason: 'cannot run against a load balanced environment'
 }));
 
-const INTERRUPT_IN_USE_CONNECTIONS_TESTS: SkipDescription[] = [
-  'Connections MUST be interrupted as soon as possible (interruptInUseConnections=true)',
```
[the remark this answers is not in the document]
durran: If we are in load balanced mode and a `serviceId` is not present in the handshake then the driver must error, so we should never get in a situation further than the initial handshake where a `serviceId` is not present. So if it's not provided to the clear that indicates to me there's a bug in the test setup.

src/cmap/connection_pool.ts
```diff
@@ -391,10 +396,10 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       this[kConnections].unshift(connection);
     }
 
-    this[kCheckedOut].delete(connection);
+    const wasConnectionDeleted = this[kCheckedOut].delete(connection);
     this.emit(ConnectionPool.CONNECTION_CHECKED_IN, new ConnectionCheckedInEvent(this, connection));
 
-    if (willDestroy) {
+    if (wasConnectionDeleted && willDestroy) {
```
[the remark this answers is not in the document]
baileympearson: See above.

src/cmap/connection_pool.ts
```diff
@@ -408,7 +413,8 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
    * Pool reset is handled by incrementing the pool's generation count. Any existing connection of a
    * previous generation will eventually be pruned during subsequent checkouts.
    */
-  clear(serviceId?: ObjectId): void {
+  clear(options: { serviceId?: ObjectId; interruptInUseConnections?: boolean } = {}): void {
+    const { serviceId } = options;
```
dariakp, line 417: can we move this to be under the comment in L422 and then modify current L423 to only gate on `this.loadBalanced` and throw a runtime error if `serviceId` is not present (just like we do for `generation == null`)? In a real life scenario, this should not happen, and if it did, it would not be correct to fall through to the non-loadbalanced logic; however, in the test scenarios this should help us filter out the tests that aren't relevant for the loadbalanced mode (i.e., all unified tests that call `clear` since for the unified spec, the `clear` is intended to be exclusively the non-loadbalanced clear)

src/cmap/connection_pool.ts
```diff
@@ -431,19 +437,51 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+
+    if (interruptInUseConnections) {
+      process.nextTick(() => this.interruptInUseConnections(oldGeneration));
+    }
+
     this.processWaitQueue();
   }
 
+  /**
+   * Closes all checked in perished connections in the pool with a resumable PoolClearedOnNetworkError.
```
dariakp, line 463:
```suggestion
   * Closes all stale in-use connections in the pool with a resumable PoolClearedOnNetworkError.
```

src/cmap/connection_pool.ts
```diff
@@ -431,19 +437,51 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+   *
+   * If interruptInUseConnections is `true`, this method attempts to kill checked out connections as well.
```
dariakp, line 465: remove this sentence since it's no longer accurate

src/cmap/errors.ts
```diff
@@ -39,6 +40,22 @@ export class PoolClearedError extends MongoNetworkError {
...
+  constructor(pool: ConnectionPool) {
+    super(pool, `Connection to ${pool.address} interrupted due to server monitor timeout`);
+
+    this.addErrorLabel(MongoErrorLabel.RetryableWriteError);
```
dariakp, line 51: we should not do this here because this error already inherits its retryability logic from the `PoolClearedError`

dariakp (changes requested):

src/cmap/connection_pool.ts
```diff
@@ -431,19 +437,51 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+   * Only connections where `connection.generation <= minGeneration` are killed.  Connections are closed with a
+   * resumable PoolClearedOnNetworkTimeoutError.
+   */
+  private interruptInUseConnections(minGeneration: number) {
+    for (const connection of this[kCheckedOut]) {
+      if (connection.generation <= minGeneration) {
+        this[kCheckedOut].delete(connection);
+        connection.onError(new PoolClearedOnNetworkError(this));
+        this.emit(
```
dariakp, line 474: I think for the sake of consistency, we *should* use the `this.destroyConnection` method instead of just "faking" the event emission here; if we just do the "onError" call on the connection, it won't properly fall through to the connection's own destroy method which is what's responsible for clearing the connection's event listeners, and even though right now those event listeners are only for lb mode, it's conceivable that we might have other listeners in the future and we don't want to accidentally introduce a leak. The connection's destroy method outside the listener removal is a no-op on an already destroyed stream, so this should be safe; however, we should also update the connection's `onError` method to also set `this.destroyed = true`. Then, with this refactor, in the `checkIn` method, we should be able to check for `connection.destroyed` directly, and instead of the misleading `willDestroy` that is only a `willDestroyIfNotDeleted`, we can update all that as follows:
- rename `willDestroy -> shouldDestroy`(or `isUsable` or `shouldReturnToThePool`, but something that communicates that the point is to determine whether we can reuse the connection, just be sure to flip the `!` if flipping the meaning)
- update check in following line to `if (!shouldDestroy)`
- update L399 to not worry whether the connection has already been deleted from the checked out set
- update L402 to `if (shouldDestroy && !connection.destroyed)` (you could also call `connection.destroyed -> isDestroyed`)

dariakp (changes requested):

src/cmap/connection_pool.ts
baileympearson: done!

src/cmap/errors.ts
dariakp: Per earlier discussion, let's add the label in `PoolClearedError` and remove the manual adding of the label in `withConnection`, and also get rid of the `isRetryableWritesEnabled` and the corresponding gate in `makeOperationHandler` entirely since `executeOperation` already has the `willRetryWrite` (and read) logic

src/cmap/connection_pool.ts
dariakp: Per the earlier discussion, let's update this method to just call `checkIn` instead of manually emitting the event and removing the connection from the checked out set, and then update the `checkIn` method to be a no-op if the connection is not in that checked out set

src/cmap/connection_pool.ts
dariakp: we'll want to also skip the (correctly) failing LB tests now since those unit tests aren't applicable to LB clear

src/cmap/connection_pool.ts
```diff
@@ -391,10 +399,10 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       this[kConnections].unshift(connection);
     }
 
-    this[kCheckedOut].delete(connection);
+    const wasConnectionDeleted = this[kCheckedOut].delete(connection);
```
dariakp, line 402: we can remove this gate now that we have the guard in L390

dariakp (changes requested):
Resolved all comments that have been addressed and just adding this comment to make sure we don't miss it, in the interests of making the re-review easier

src/cmap/connection_pool.ts
baileympearson: done

dariakp (commented):
The code changes look fine now, just need to file the follow up (see open thread). However, there is a suspicious cmap test failing on test-latest-server-v1-api (rhel8 node 12), and a lot more failures than could comfortably be accounted for by known issues. @baileympearson can you take a look to make sure these are transient?

nbbeeken (approved):
optional nits, LGTM

src/cmap/connection_pool.ts
```diff
@@ -431,19 +445,45 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+  private interruptInUseConnections(minGeneration: number) {
+    for (const connection of this[kCheckedOut]) {
+      if (connection.generation <= minGeneration) {
+        this.checkIn(connection);
+        connection.onError(new PoolClearedOnNetworkError(this));
+      }
+    }
+
+    // TODO(NODE-4784): track pending connections and cancel
+    // this[kCancellationToken].emit('cancel');
```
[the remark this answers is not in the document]
baileympearson: Do you want me to remove this?  I intentionally left this in place, since the code is associated with the comment.

src/cmap/errors.ts
baileympearson: https://jira.mongodb.org/browse/NODE-4838 relevant ticket

src/cmap/connection_pool.ts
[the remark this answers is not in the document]
nbbeeken: Up to you, I thought the attached ticket is an involved design w/ a writing ticket and all so I thought the code wouldn't be relevant when that's actually ready to go. But cancelling of some kind will be done so not totally off the mark

src/cmap/connection_pool.ts
```diff
@@ -572,7 +612,12 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     return !!(this.options.maxIdleTimeMS && connection.idleTime > this.options.maxIdleTimeMS);
   }
 
-  private connectionIsPerished(connection: Connection) {
+  /**
+   * Destroys a connection if the connection is perished.
+   *
+   * @returns `true` if the connection was destroyed, `false` otherwise.
+   */
+  private destroyConnectionIfPerished(connection: Connection) {
```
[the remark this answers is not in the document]
baileympearson: done!

src/cmap/connection_pool.ts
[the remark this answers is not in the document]
baileympearson: okay, I removed the comment

nbbeeken (approved):

dariakp (approved):

PR #3458, opened 2022-11-01
feat(NODE-4733): deprecate result and getLastOp

durran:
### Description

Deprecates `BulkWriteResult`'s `result` and `getLastOp`.

#### What is changing?

Deprecates `BulkWriteResult`'s `result` and `getLastOp`.

None

NODE-4733

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -327,6 +327,7 @@ export class BulkWriteResult {
     }
   }
 
+  /* @deprecated Will be removed in 5.0 release */
   toJSON(): BulkResult {
     return this.result;
   }
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -176,6 +176,7 @@ export class Batch<T = Document> {
  * The result of a bulk write.
  */
 export class BulkWriteResult {
+  /** @deprecated Will be removed in 5.0 */
   result: BulkResult;
 
   /**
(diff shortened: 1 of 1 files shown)


nbbeeken (approved):
LGTM and thanks for the update in NODE-4034

nbbeeken (changes requested):
Sorry I looked closer at BulkResult and found two things we want to mark here as well

src/bulk/common.ts
```diff
@@ -176,6 +176,7 @@ export class Batch<T = Document> {
  * The result of a bulk write.
  */
 export class BulkWriteResult {
+  /** @deprecated Will be removed in 5.0 */
   result: BulkResult;
```
nbbeeken, line 180: We actually need to think about how this type is public and exposes things we intend to remove, opTime. We should mark BulkResult as deprecate it and mention it in the removal ticket as well, it won't need to be deleted there since it's used plenty through out internal bulk handling logic but marking it internal will delete it from the public types so it needs to be done in the major. 

The `toJSON()` method was likely added because of this weirdness with the nested result. Now that it's going to strictly define the correct props we don't need a toJSON method so lets deprecate it and mention it in the removal ticket.

nbbeeken (changes requested):

src/bulk/common.ts
durran: BulkResult now deprecated as well with a note it will be made internal. NODE-4034 updated for the 2 new deprecations.

dariakp (approved):

PR #3459, opened 2022-11-02
feat(NODE-4681): deprecate modify result

durran:
### Description

Deprecates the `ModifyResult` type.

#### What is changing?

- Deprecates the `ModifyResult` type.
- Adds commented out type tests to use later when returning the actual result document.

None

NODE-4681

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -101,7 +101,9 @@ import { WriteConcern, WriteConcernOptions } from './write_concern';
 
 /**
  * @public
- * @deprecated Will return the actual result document in 5.0
+ * @deprecated This type will be completely removed in 5.0 and findOneAndUpdate,
+ *             findOneAndDelete, and findOneAndReplace will then return the
+ *             actual result document.
  */
 export interface ModifyResult<TSchema = Document> {
   value: WithId<TSchema> | null;
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -99,7 +99,10 @@ import {
 } from './utils';
 import { WriteConcern, WriteConcernOptions } from './write_concern';
 
-/** @public */
+/**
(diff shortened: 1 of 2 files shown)


test/types/community/collection/findX.test-d.ts
```diff
@@ -348,3 +348,21 @@ expectType<WithId<{ a: number; b: string }> | null>(
 expectType<WithId<{ a: number; b: string }> | null>(
   (await coll.findOneAndDelete({ a: 3 }, { projection: { _id: 0 } })).value
 );
+
+// NODE-3568: Uncomment when ModifyResult is removed in 5.0
+// expectType<Document> | null>((await coll.findOneAndDelete({ a: 3 })));
```
dariakp, line 353: For these expectations, the actual outcome will be `WithId<TSchema> | null` where `TSchema` is `{ a: number; b: string }` because of line 329

src/collection.ts
```diff
@@ -99,7 +99,10 @@ import {
 } from './utils';
 import { WriteConcern, WriteConcernOptions } from './write_concern';
 
-/** @public */
+/**
+ * @public
+ * @deprecated Will return the actual result document in 5.0
```
dariakp, line 104: I think we want to clarify here that this type will be removed and the `findOneAnd...` methods will return the result document directly

dariakp (changes requested):

test/types/community/collection/findX.test-d.ts
durran: Updated.

src/collection.ts
durran: Comment now more detailed about the deprecation.

PR #3461, opened 2022-11-03
docs(NODE-4798): move deprecated overloads to the bottom

ImRodry:
#### What is changing?

This PR simply moves all deprecated function overloads to the bottom of their signatures so that the overloads that should be used come up first and are preferred by TS

#### Is there new documentation needed for these changes?

No, they were already documented in the PR that introduced the deprecations

When using TS's built-in types such as ReturnType or others, TS will usually grab the first overload of the function. Along with that, if you ever had an error in some of your function's arguments, the overload would often fallback to the callback one since, in the presence of an error, TS fins the first overload that matches the amount of parameters passed, causing your function's return type to be `void`. Also by doing this it will make it easier to remove all of these overloads when the time comes for that, and it will also make the diff look nicer

https://jira.mongodb.org/browse/NODE-4798

Got this error when running it but I believe that's not my fault
image

### DIFF
--- a/src/admin.ts
+++ b/src/admin.ts
@@ -60,9 +60,9 @@ export class Admin {
    * @param callback - An optional callback, a Promise will be returned if none is provided
    */
   command(command: Document): Promise<Document>;
+  command(command: Document, options: RunCommandOptions): Promise<Document>;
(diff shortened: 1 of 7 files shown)


ImRodry:
Dunno why CI is failing and I can't see why either

durran (approved):
Verified the deprecated overloads do now show up as the last suggestions in VS Code. Example: 

Screenshot 2022-11-09 at 15 56 19

PR #3463, opened 2022-11-10
fix(NODE-4783): handle orphaned operation descriptions

durran:
### Description

In the case where streaming hello responses are skipped in the driver, the last hello that is processed will get ignored by the connection's `onMessage` handler because the `responseTo` field of the hello does not match the original `requestId`

#### What is changing?

This was the least invasive change I could think of doing without reworking the monitoring connection to not depend on its internal queue. That is the basis of not just matching the request and responses but all the callback execution and error handling also is closely coupled to the queue.

This change basically knows that when an operation description is not found in the queue for the `responseTo` field and it is a monitoring connection, that hellos have been skipped and there will be one operation description in the queue for the original hello that was sent. We remove this one from the queue and the `onMessage` handler will auto-correct and put a new operation description back in the queue with the correct key.

None

NODE-4783 / SERVER-45775

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -405,7 +405,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     // making the `responseTo` change on each response
     this[kQueue].delete(message.responseTo);
     if ('moreToCome' in message && message.moreToCome) {
-      // NODE-4783: If the operation description check above does find an orphaned
+      // If the operation description check above does find an orphaned
       // description and sets the operationDescription then this line will put one
       // back in the queue with the correct requestId and will resolve not being able
       // to find the next one via the responseTo of the next streaming hello.
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -374,23 +374,24 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this.emit('message', message);
     let operationDescription = this[kQueue].get(message.responseTo);
 
+    // Protect against multiplexing.
+    if (this[kQueue].size > 1) {
+      this.onError(new MongoRuntimeError(INVALID_QUEUE_SIZE));
+      return;
+    }
+
     if (!operationDescription && this.isMonitoringConnection) {
       // This is how we recover when the initial hello's requestId is not
       // the responseTo when hello responses have been skipped:
 
-      // First check if the map is of invalid size
-      if (this[kQueue].size > 1) {
-        this.onError(new MongoRuntimeError(INVALID_QUEUE_SIZE));
-      } else {
-        // Get the first orphaned operation description.
-        const entry = this[kQueue].entries().next();
-        if (entry) {
-          const [requestId, orphaned]: [number, OperationDescription] = entry.value;
-          // If the orphaned operation description exists then set it.
-          operationDescription = orphaned;
-          // Remove the entry with the bad request id from the queue.
-          this[kQueue].delete(requestId);
-        }
+      // Get the first orphaned operation description.
+      const entry = this[kQueue].entries().next();
+      if (entry) {
+        const [requestId, orphaned]: [number, OperationDescription] = entry.value;
+        // If the orphaned operation description exists then set it.
+        operationDescription = orphaned;
+        // Remove the entry with the bad request id from the queue.
+        this[kQueue].delete(requestId);
       }
     }
 
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -374,24 +374,23 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this.emit('message', message);
     let operationDescription = this[kQueue].get(message.responseTo);
 
-    // Protect against multiplexing.
-    if (this[kQueue].size > 1) {
-      this.onError(new MongoRuntimeError(INVALID_QUEUE_SIZE));
-      return;
-    }
-
     if (!operationDescription && this.isMonitoringConnection) {
       // This is how we recover when the initial hello's requestId is not
       // the responseTo when hello responses have been skipped:
 
-      // Get the first orphaned operation description.
-      const entry = this[kQueue].entries().next();
-      if (entry) {
-        const [requestId, orphaned]: [number, OperationDescription] = entry.value;
-        // If the orphaned operation description exists then set it.
-        operationDescription = orphaned;
-        // Remove the entry with the bad request id from the queue.
-        this[kQueue].delete(requestId);
+      // First check if the map is of invalid size
+      if (this[kQueue].size > 1) {
+        this.onError(new MongoRuntimeError(INVALID_QUEUE_SIZE));
+      } else {
+        // Get the first orphaned operation description.
+        const entry = this[kQueue].entries().next();
+        if (entry) {
+          const [requestId, orphaned]: [number, OperationDescription] = entry.value;
+          // If the orphaned operation description exists then set it.
+          operationDescription = orphaned;
+          // Remove the entry with the bad request id from the queue.
+          this[kQueue].delete(requestId);
+        }
       }
     }
 
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -375,8 +375,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     let operationDescription = this[kQueue].get(message.responseTo);
 
     if (!operationDescription && this.isMonitoringConnection) {
-      // NODE-4783: How do we recover from this when the initial hello's requestId is not
-      // the responseTo when hello responses have been skipped?
+      // This is how we recover when the initial hello's requestId is not
+      // the responseTo when hello responses have been skipped:
 
       // First check if the map is of invalid size
       if (this[kQueue].size > 1) {
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -369,7 +369,23 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
 
     // always emit the message, in case we are streaming
     this.emit('message', message);
-    const operationDescription = this[kQueue].get(message.responseTo);
+    let operationDescription = this[kQueue].get(message.responseTo);
+
+    if (!operationDescription && this.isMonitoringConnection) {
+      // NODE-4783: How do we recover from this when the initial hello's requestId is not
+      // the responseTo when hello responses have been skipped?
+      //
+      // Get the first orphaned operation description.
+      const entry = this[kQueue].entries().next();
+      if (entry) {
+        const [requestId, orphaned]: [number, OperationDescription] = entry.value;
+        // If the orphaned operation description exists then set it.
+        operationDescription = orphaned;
+        // Remove the entry with the bad request id from the queue.
+        this[kQueue].delete(requestId);
+      }
+    }
+
     if (!operationDescription) {
       return;
     }
@@ -381,7 +397,10 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     // making the `responseTo` change on each response
     this[kQueue].delete(message.responseTo);
     if ('moreToCome' in message && message.moreToCome) {
-      // requeue the callback for next synthetic request
+      // NODE-4783: If the operation description check above does find an orphaned
+      // description and sets the operationDescription then this line will put one
+      // back in the queue with the correct requestId and will resolve not being able
+      // to find the next one via the responseTo of the next streaming hello.
       this[kQueue].set(message.requestId, operationDescription);
     } else if (operationDescription.socketTimeoutOverride) {
       this[kStream].setTimeout(this.socketTimeoutMS);
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -17,6 +17,7 @@ import {
   MongoMissingDependencyError,
   MongoNetworkError,
   MongoNetworkTimeoutError,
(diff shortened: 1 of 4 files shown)


test/unit/cmap/connection.test.ts
```diff
@@ -137,6 +157,89 @@ describe('new Connection()', function () {
...
+            opCode: msg.readInt32LE(12)
+          };
+          const msgBody = msg.subarray(16);
+
+          const message = new BinMsg(msg, msgHeader, msgBody);
+          connection.onMessage(message);
+        });
+
+        it('calls the operation description callback with the document', function () {
+          expect(callbackSpy).to.be.calledWith(undefined, document);
```
durran, line 204: This would fail before the changes as the early would turn would cause the callback never to be called.

test/unit/cmap/connection.test.ts
```diff
@@ -137,6 +157,89 @@ describe('new Connection()', function () {
...
+          };
+
+          // Stick an operation description in the queue.
+          queue.set(1, operationDescription);
+          // Emit a message that won't match the existing operation description.
+          const msg = generateOpMsgBuffer(document);
+          const msgHeader: MessageHeader = {
+            length: msg.readInt32LE(0),
+            requestId: msg.readInt32LE(4),
+            responseTo: msg.readInt32LE(8),
```
durran, line 194: This will default to 0 so won't match.

test/unit/cmap/connection.test.ts
baileympearson, line 204:
```suggestion
          expect(callbackSpy).to.be.calledExactlyOnceWith(undefined, document);
```

test/unit/cmap/connection.test.ts
```diff
@@ -137,6 +157,89 @@ describe('new Connection()', function () {
...
+            opCode: msg.readInt32LE(12)
+          };
+          const msgBody = msg.subarray(16);
+
+          const message = new BinMsg(msg, msgHeader, msgBody);
+          connection.onMessage(message);
+        });
+
+        it('calls the operation description callback with the document', function () {
+          expect(callbackSpy).to.be.calledWith(undefined, document);
```
baileympearson, line 237:
```suggestion
          expect(callbackSpy).to.be.calledExactlyOnceWith(undefined, document);
```

baileympearson (changes requested):
The current tests look great.  

Should we consider a regression test that tests that when multiple hellos come through the socket?  It would check the following scenarios:

1. skip all stale hellos
2. the connection class finds the appropriate callback and calls it exactly once

This would give me more confidence in the skipping stale hellos fix as a whole.

---

I think this test would be relatively straightforward.  I think we could create a mock duplex stream as a dummy socket ( similar to fakesocket ), hold a reference to it, initialize the connection class as you do in the other tests, and then push a sequence of hellos into the socket.

Thoughts? @durran

test/unit/cmap/connection.test.ts
```diff
@@ -137,6 +158,136 @@ describe('new Connection()', function () {
...
+    context('when the connection is a monitoring connection', function () {
+      let queue: Map<number, OperationDescription>;
+      let driverSocket: FakeSocket;
+      let connection: Connection;
+
+      beforeEach(function () {
+        driverSocket = sinon.spy(new FakeSocket());
+      });
+
+      context('when multiple hellos exist on the stream', function () {
```
durran, line 171: This now actually pushes a buffer with 3 hellos onto the input stream as the additional test.

test/tools/utils.ts
```diff
@@ -354,6 +354,18 @@ export class TestBuilder {
   }
 }
 
+export function bufferToStream(buffer) {
```
durran, line 357: Unrelated moving of this function from the message stream unit test to test utils.

durran:
> 

Add per: https://github.com/mongodb/node-mongodb-native/pull/3463/files#diff-88580e6012f69fd7e6bad7e3b88b16feba23c2ba5d3953f026cbc768a9081535R171

baileympearson (approved):

test/unit/cmap/connection.test.ts
```diff
@@ -137,6 +158,135 @@ describe('new Connection()', function () {
...
+          // Stick an operation description in the queue.
+          queue.set(1, operationDescription);
+
+          // Push the buffer of 3 hellos to the input stream
+          inputStream.push(buffer);
+          inputStream.push(null);
+        });
+
+        it('calls the operation description callback with the document', async function () {
+          await on(inputStream, 'message');
```
dariakp, line 203: this would resolve immediately, no? iirc, `on` returns an `AsyncIterableIterator`

you can await a `once(...)` instead, but I think what you actually want is to set up the listener before you push things into the `inputStream`, and then await `onChangeIterator.next()` here; alternatively, since the current await does resolve sync, perhaps we don't need to await anything

test/unit/cmap/connection.test.ts
```diff
@@ -137,6 +158,135 @@ describe('new Connection()', function () {
...
+
+          // Stick an operation description in the queue.
+          queue.set(1, operationDescription);
+
+          // Push the buffer of 3 hellos to the input stream
+          inputStream.push(buffer);
+          inputStream.push(null);
+        });
+
+        it('calls the operation description callback with the document', async function () {
```
dariakp, line 202: I think we should specify which document we expect the op description to get called with here, (e.g., first or last?)

test/unit/cmap/connection.test.ts
```diff
@@ -137,6 +158,135 @@ describe('new Connection()', function () {
...
+      });
+
+      context('when multiple hellos exist on the stream', function () {
+        let callbackSpy;
+        const inputStream = new Readable();
+        const document = { ok: 1 };
+
+        beforeEach(function () {
+          callbackSpy = sinon.spy();
+          const firstHello = generateOpMsgBuffer(document);
```
dariakp, line 178: I think we should differentiate the 3 hello docs so that we can check that the correct expected document is being processed, something like `generateOpMsgBuffer({ ...document, id: 1 })`

test/unit/cmap/connection.test.ts
dariakp: can we add this as a code comment instead so that if we need to modify these tests later it's more clear what the line is doing?

dariakp (changes requested):

test/unit/cmap/connection.test.ts
durran: I've updated this so the last hello is now `{ isWritablePrimary: true }` now to distinguish it.

test/unit/cmap/connection.test.ts
```diff
@@ -137,6 +158,189 @@ describe('new Connection()', function () {
...
+
+          // Push the buffer of 3 hellos to the input stream
+          inputStream.push(buffer);
+          inputStream.push(null);
+        });
+
+        it('calls the callback with the last hello document', async function () {
+          const messages = await once(connection, 'message');
+          expect(messages[0].responseTo).to.equal(0);
+          expect(callbackSpy).to.be.calledOnceWith(undefined, last);
```
durran, line 206: Expects the last document now.

src/cmap/connection.ts
```diff
@@ -369,7 +369,23 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
     this.emit('message', message);
-    const operationDescription = this[kQueue].get(message.responseTo);
+    let operationDescription = this[kQueue].get(message.responseTo);
+
+    if (!operationDescription && this.isMonitoringConnection) {
+      // NODE-4783: How do we recover from this when the initial hello's requestId is not
+      // the responseTo when hello responses have been skipped?
+      //
+      // Get the first orphaned operation description.
+      const entry = this[kQueue].entries().next();
```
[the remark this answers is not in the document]
durran: Now calls all callbacks with a `MongoRuntimeError` if the map size is greater than 1.

test/unit/cmap/connection.test.ts
durran: I needed to await something as the expectation was getting checked before the input stream piped to the message stream. I changed this to actually now add an extra assertion by awaiting `once` on the connection itself (since it does emit a message event as well on `onMessage` and asserting the first message has the expected `responseTo` value.

test/unit/cmap/connection.test.ts
durran: Now checks for the last hello document.

dariakp (dismissed):

src/cmap/connection.ts
```diff
@@ -369,7 +372,28 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
     this.emit('message', message);
-    const operationDescription = this[kQueue].get(message.responseTo);
+    let operationDescription = this[kQueue].get(message.responseTo);
+
+    if (!operationDescription && this.isMonitoringConnection) {
+      // This is how we recover when the initial hello's requestId is not
+      // the responseTo when hello responses have been skipped:
+
+      // First check if the map is of invalid size
+      if (this[kQueue].size > 1) {
```
jyemin, line 382: Would it make sense to move this check out of the conditional?  It seems like when it's not a monitoring connection the same condition should hold, since we don't multiplex connections

src/cmap/connection.ts
```diff
@@ -369,7 +372,28 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+        this.onError(new MongoRuntimeError(INVALID_QUEUE_SIZE));
+      } else {
+        // Get the first orphaned operation description.
+        const entry = this[kQueue].entries().next();
+        if (entry) {
```
jyemin, line 387: Under what conditions could this evaluate to false?  Or is it defensive programming?

src/cmap/connection.ts
```diff
@@ -369,7 +372,28 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
 
     // always emit the message, in case we are streaming
     this.emit('message', message);
-    const operationDescription = this[kQueue].get(message.responseTo);
+    let operationDescription = this[kQueue].get(message.responseTo);
+
+    if (!operationDescription && this.isMonitoringConnection) {
```
jyemin, line 377: What if it's a monitoring connection that doesn't support streaming responses?  Should we still enter this code block?  If not, is there a way to check for that condition?

src/cmap/connection.ts
durran: It's defensive as Typescript will fail to compile since the call to `next()` can be undefined. This should never happen.

jyemin: Should an error be raised instead, or is this the normal way of handling such situations?

src/cmap/connection.ts
durran: This code block should never get entered for a non-streaming monitoring connection as the operation description must always be there in that case.

jyemin: True, but you could say the same about a non-monitoring connection. Would you want just:

```
if (!operationDescription)
```

src/cmap/connection.ts
durran: If moved the check up now.

dariakp: Ostensibly, the answer here is "yes", but that would be a potential behavior change that might be better reserved for NODE-4820 and at a mininum would require additional tests that feel out of scope for the actual fix

src/cmap/connection.ts
```diff
@@ -369,7 +372,29 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
 
     // always emit the message, in case we are streaming
     this.emit('message', message);
-    const operationDescription = this[kQueue].get(message.responseTo);
+    let operationDescription = this[kQueue].get(message.responseTo);
+
+    // Protect against multiplexing.
+    if (this[kQueue].size > 1) {
```
dariakp, line 378: we'd want to add a test for a non-monitoring connection for this

src/cmap/connection.ts
durran: We can't error handle here as if there was in fact no entry then we have no callback to pass the error to as it hangs off the operation description, so we fall back to just returning below. This is obviously not ideal but we have a ticket for that (NODE-4820) to refactor the queue out entirely.

src/cmap/connection.ts
durran: Theoretically we could, but I wanted the scope of this fix just to be limited to monitoring connections and handle the entire connection refactor with NODE-4820.

src/cmap/connection.ts
durran: Added.

src/cmap/connection.ts
dariakp: What Durran said, this will become a non-issue after NODE-4820, and if we do have any unexpected behavior with missing operation descriptions on regular operations, that would be a better time to change the behavior. I'm not sure if we have an easy way of differentiating streaming monitoring connections from non-streaming (if we do, then I'm not opposed to adding an extra guard here). But keeping it for all monitoring connections seems low risk, since the worst case scenario is we hand a hello response to a hello handler that's detached from the original request.

dariakp (approved):

src/cmap/connection.ts
durran: Moved back into the monitoring connection block due to change stream resumability being affected by that change.

jyemin: Interesting.  I'd like to understand how how, but can discuss outside the PR.

jyemin (approved):
LGTM

dariakp (approved):

PR #3474, opened 2022-11-21
ci(NODE-4856): break apart ts tests into separate tasks and skip failing tasks on typescript next

baileympearson:
This PR breaks apart our typescript CI tasks into a task that checks compilation of the driver and a task that checks that our public types can be resolved.

This PR also skips any failing typescript tasks.

### DIFF
--- a/.evergreen/run-typescript.sh
+++ b/.evergreen/run-typescript.sh
@@ -5,14 +5,25 @@ source "${PROJECT_DIRECTORY}/.evergreen/init-nvm.sh"
 
 set -o xtrace
 
-function get_current_ts_version {
-    node -e "console.log(require('./package-lock.json').dependencies.typescript.version)"
+if [ -z "$TS_CHECK" ]; then echo "TS_CHECK must be set to either COMPILE_DRIVER or CHECK_TYPES"; exit 1; fi
+if [ -z "$TS_VERSION" ]; then echo "TS_VERSION must be set"; exit 1; fi
+
+if [ ! -f "mongodb.d.ts" ]; then
+    # mongodb.d.ts should always exist because of the installation in prior steps
+    # but in case it doesn't, build it
+    npm i
+fi
+
+function get_ts_version {
+    if [ "$TS_VERSION" == "current" ]; then
+        echo $(node -e "console.log(require('./package-lock.json').dependencies.typescript.version)")
+    else
+        echo $TS_VERSION
+    fi
 }
 
-CURRENT_TS_VERSION=$(get_current_ts_version)
-
 export TSC="./node_modules/typescript/bin/tsc"
-export TS_VERSION=${TS_VERSION:=$CURRENT_TS_VERSION}
+export TS_VERSION=$(get_ts_version)
 
 npm install --no-save --force typescript@"$TS_VERSION"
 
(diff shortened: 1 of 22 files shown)


.evergreen/run-typescript.sh
```diff
@@ -5,14 +5,25 @@ source "${PROJECT_DIRECTORY}/.evergreen/init-nvm.sh"
 
 set -o xtrace
 
-function get_current_ts_version {
-    node -e "console.log(require('./package-lock.json').dependencies.typescript.version)"
-}
+if [ -z "$TS_CHECK" ]; then echo "TS_CHECK must be set to either COMPILE_DRIVER or CHECK_TYPES"; exit 1; fi
```
nbbeeken, line 8: If those are the only two valid settings for this var, can we check against that being the only permitted values here.
_edit_ ah just got to the end of the script that has the fail case, that's fine, but as a reader of this line of script I'd expect it to be checked here, but I'm fine with how it is, so just leaving this as a suggestion

.evergreen/run-typescript.sh
```diff
@@ -5,14 +5,25 @@ source "${PROJECT_DIRECTORY}/.evergreen/init-nvm.sh"
...
+if [ -z "$TS_VERSION" ]; then echo "TS_VERSION must be set"; exit 1; fi
 
-CURRENT_TS_VERSION=$(get_current_ts_version)
+if [ ! -f "mongodb.d.ts" ]; then
+    # mongodb.d.ts should always exist because of the installation in prior steps
+    # but in case it doesn't, build it
+    npm i
```
nbbeeken, line 14: The comment you have there is relevant if type defs are missing in CI, just for a bit more viz, you could turn it into an `echo`

.evergreen/run-typescript.sh
```diff
@@ -5,14 +5,25 @@ source "${PROJECT_DIRECTORY}/.evergreen/init-nvm.sh"
...
+fi
+
+function get_ts_version {
```
nbbeeken, line 17:
```suggestion
function get_ts_version() {
```
just a preference, up to you

nbbeeken (changes requested):
The comment about the echo is the only real one I feeling strongly about changing, otherwise lgtm

.evergreen/run-typescript.sh
baileympearson: nope, I prefer this as well. I dislike that bash has 4 different syntaxes for declaring functions.

.evergreen/run-typescript.sh
baileympearson: Good point, I can add a check

Is it worth keeping the guard at the end of the script?  There's nothing stopping future devs from modifying `TS_CHECK` in the middle of the script, after it has been validated here but before we reach the actual code that uses it.  That's why I have the "else" block at the end.

nbbeeken: True, the inspiration just comes from how set but incorrect values can sometimes hide errors. The check at the end would've caught the issue when we go to test, a check at the top prevents any of that setup code (maybe npm install) from running before we get there, both have value and if its not too much to add why not be extra defensive. `#justCIThings`

.evergreen/run-typescript.sh
baileympearson: done

.evergreen/run-typescript.sh
baileympearson: addressed

baileympearson: I left the check at the end and added validation at the start

nbbeeken (approved):

PR #3475, opened 2022-11-22
feat(NODE-4810): define the new Logger

andymina:
#### What is changing?

Defines the new Logger class as described by NODE-3974. This PR handles the options parsing and instantiation of the logger.


src/mongo_logger.ts
```diff
@@ -0,0 +1,144 @@
...
+  maxDocumentLength: number;
+  logPath: Writable;
+
+  constructor(options: MongoLoggerOptions) {
+    // validate log path
+    if (typeof options.logPath === 'string') {
+      this.logPath =
+        options.logPath === 'stderr' || options.logPath === 'stdout'
+          ? process[options.logPath]
+          : fs.createWriteStream(options.logPath, { flags: 'a+' });
```
baileympearson, line 82: I want team discussion around this but I think we should not initialize the write stream in the constructor.

- i/o in constructors is bad practice generally
- we have no mechanism to propagate an error (which occurs asynchronously) when opening the file.  regardless of how we handle errors later on, this is an error we will want to show users (we should at the very least tell them that the logger is turned off when they expect it turned on, even if we swallow up any other logging errors) but parseOptions is not async, so we can't propagate the error there and the mongo client isn't created yet so we can't emit an error event from parseOptions.

I think "initializing" the logger could be done in a separate step on first-operation (or connect, which is really the same thing) and then error handling is easy:

- We have the mongo client at that point, so we can emit an error event if we choose that route for error handling
- We are now in an async context, so we can throw an error asynchronously if we decide to go that route.

Thoughts @durran @dariakp @andymina @nbbeeken ?

baileympearson (changes requested):

src/mongo_logger.ts
```diff
@@ -0,0 +1,144 @@
+import * as fs from 'fs';
+import type { Writable } from 'stream';
+
+import { getUint } from './connection_string';
```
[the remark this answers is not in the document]
andymina: done, thanks for spotting this!

src/mongo_logger.ts
```diff
@@ -0,0 +1,144 @@
...
+  /** Severity level for server selection component */
+  serverSelection: SeverityLevel;
+  /** Severity level for CMAP */
+  connection: SeverityLevel;
+  /** Default severity level to be if any of the above are unset */
+  defaultSeverity: SeverityLevel;
+  /** Max length of embedded EJSON docs. Setting to 0 disables truncation. Defaults to 1000. */
+  maxDocumentLength: number;
+  /** Destination for log messages. Must be 'stderr', 'stdout', a file path, or a Writable. Defaults to 'stderr'. */
+  logPath: string | Writable;
```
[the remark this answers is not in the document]
andymina: i initially had it as a logDestination but ultimately changed it for consistency between the name for it in MongoClient (`mongodbLogPath`) and the env var (`MONGODB_LOG_PATH`). i've changed the MongoClient, class prop, and MongoLoggerOptions prop to all be logDestination for consistency.

test/unit/mongo_client.test.js
```diff
@@ -847,4 +848,32 @@ describe('MongoOptions', function () {
...
+      'MONGODB_LOG_CONNECTION',
+      'MONGODB_LOG_ALL'
+    ];
+
+    for (const name of severityVars) {
+      it(`should enable logging if at least ${name} is set to a valid value`, function () {
+        process.env[name] = SeverityLevel.CRITICAL;
+        const client = new MongoClient('mongodb://localhost:27017');
+        expect(client.mongoLogger).to.be.instanceOf(MongoLogger);
+        process.env[name] = undefined;
```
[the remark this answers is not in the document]
andymina: oops, thanks for this insight

test/unit/mongo_client.test.js
```diff
@@ -847,4 +848,32 @@ describe('MongoOptions', function () {
...
+      });
+    }
+
+    for (const name of severityVars) {
+      it(`should not enable logging if ${name} is set to an invalid value`, function () {
```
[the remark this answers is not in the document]
andymina: changing the test name to make this clearer. in the example that you provided, logging would still be enabled, however it does lack a warning. adding the warning in now.

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    it('will only use MONGODB_LOG_ALL for component severities that are not set or invalid', function () {
+      expect(loggerOptions.command).to.equal(loggerOptions.defaultSeverity); // empty str
+      expect(loggerOptions.topology).to.equal(loggerOptions.defaultSeverity); // undefined
+      expect(loggerOptions.serverSelection).to.equal(loggerOptions.defaultSeverity); // invalid
+    });
+
+    it('can set severity levels per component', function () {
+      const { componentSeverities } = new MongoLogger(loggerOptions);
```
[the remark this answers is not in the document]
andymina: i initially wanted to use the class directly as that's what the actual logging operation would rely, but using `loggerOptions` does simplify things

src/mongo_logger.ts
```diff
@@ -0,0 +1,144 @@
...
+   * take precedence over both environment variables and MongoClient options.
+   *
+   * @remarks
+   * When parsing component severity levels, invalid values are treated as unset and replaced with
+   * the default severity.
+   *
+   * @param clientOptions - options set for the logger in the MongoClient options
+   * @returns a MongoLoggerOptions object to be used when instantiating a new MongoLogger
+   */
+  static resolveOptions(clientOptions?: MongoLoggerMongoClientOptions): MongoLoggerOptions {
```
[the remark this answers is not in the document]
andymina: neal and i spoke yesterday about moving away from using the global process and i think this would be a step in the right direction of making things more configurable and hopefully future-proofing this code

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      process.env.MONGODB_LOG_CONNECTION = 'CRITICAL';
+      process.env.MONGODB_LOG_ALL = 'eRrOr';
+      process.env.MONGODB_LOG_MAX_DOCUMENT_LENGTH = '100';
+      process.env.MONGODB_LOG_PATH = 'stderr';
+
+      loggerOptions = MongoLogger.resolveOptions();
+    });
+
+    it('treats severity values as case-insensitive', function () {
+      expect(loggerOptions.connection).to.equal(SeverityLevel.CRITICAL);
```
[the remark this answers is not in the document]
andymina: the addition of the env parameter to `resolveOptions` made testing a lot easier and clearer, ty!

src/mongo_logger.ts
```diff
@@ -0,0 +1,144 @@
...
+  DEBUG: 'debug',
+  TRACE: 'trace',
+  OFF: 'off'
+} as const);
+
+/** @internal */
+export type SeverityLevel = typeof SeverityLevel[keyof typeof SeverityLevel];
+
+/** @returns one of SeverityLevel or null if passsed severity is not a valid SeverityLevel */
+function toValidSeverity(severity?: string): SeverityLevel | null {
```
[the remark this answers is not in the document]
andymina: im on board; it's clearer

src/index.ts
```diff
@@ -300,6 +300,13 @@ export type {
   SupportedTLSSocketOptions,
   WithSessionCallback
 } from './mongo_client';
+export type {
+  MongoLoggableComponent,
+  MongoLogger,
+  MongoLoggerMongoClientOptions,
+  MongoLoggerOptions,
+  SeverityLevel
+} from './mongo_logger';
```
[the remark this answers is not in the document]
andymina: im exporting the necessary types here for API extractor because if logging is enabled, the `mongoLogger` prop will be of type `MongoLogger`. without this API extractor errors. is there something else i should be doing instead?

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  MongoLoggableComponent,
+  MongoLogger,
+  MongoLoggerOptions,
+  SeverityLevel
+} from '../../src/mongo_logger';
+
+describe('Logger', function () {
+  describe('options parsing', function () {
+    let loggerOptions: MongoLoggerOptions;
```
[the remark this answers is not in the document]
andymina: good point. i'm gonna hold on creating tests for that logDestination logic in case it gets refactored as a result of our conversation [here](https://github.com/mongodb/node-mongodb-native/pull/3475#discussion_r1030746593). leaving this conversation unresolved as a reminder to ultimately write these tests

src/mongo_logger.ts
nbbeeken: From slack: I don't feel strongly about the code's location and we have error handling as a follow up ticket, we could omit the createWriteStream case of the ternary altogether right now since it's unused here, and defer to the next bit of work to stick it in the right place and correctly link up / await outcomes from i/o as needed.

src/index.ts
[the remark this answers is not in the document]
baileympearson: This is related to the api-extractor issues we discussed last week at tech chat.  I'm consider this resolved but am leaving it open as context for other reviewers.

src/mongo_logger.ts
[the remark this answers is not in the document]
baileympearson: Is the hesitation about using Document that it doesn't assert on which keys are on the object (like `MongoLoggerEnvOptions` does) or that it's incorrect?  You're right that Document isn't exactly correct because it allows arbitrary types for values (i.e., `{ 'name': 1234 }` is allowed but values can only be strings), either `Record<string, string | undefined>` or `NodeJS.ProcessEnv` would be the correct types.

I'll leave the choice up to you because I think both are fine.  But one benefit of passing in the `process.env` object directly and using either `Record<string, string | undefined>` or `NodeJS.ProcessEnv` as the type for the environment is that the `resolveOptions` function would encapsulate all options resolution.  The current approach spreads the options resolution across the `resolveOptions` function and parseOptions.

baileympearson (changes requested):

test/unit/mongo_client.test.js
```diff
@@ -847,4 +848,32 @@ describe('MongoOptions', function () {
...
+    const severityVars = [
+      'MONGODB_LOG_COMMAND',
+      'MONGODB_LOG_TOPOLOGY',
+      'MONGODB_LOG_SERVER_SELECTION',
+      'MONGODB_LOG_CONNECTION',
+      'MONGODB_LOG_ALL'
+    ];
+
+    for (const envName of severityVars) {
+      it(`should enable logging if ${envName} is set to a valid value`, function () {
```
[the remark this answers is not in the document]
andymina: i think this is clearer than the original, although it's not as explicit in stating whether logging is enabled or not. i'm down to make the change though.

test/unit/mongo_client.test.js
```diff
@@ -847,4 +848,32 @@ describe('MongoOptions', function () {
...
+
+  context('MongoLogger', function () {
+    const severityVars = [
+      'MONGODB_LOG_COMMAND',
+      'MONGODB_LOG_TOPOLOGY',
+      'MONGODB_LOG_SERVER_SELECTION',
+      'MONGODB_LOG_CONNECTION',
+      'MONGODB_LOG_ALL'
+    ];
+
```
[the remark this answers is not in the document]
andymina: i'm not sure if this change is worth it. it cleans up the tests to be a bit more readable and direct about what's expected. however i dont think having the set/unset log within the test is too bad. willing to defer to the team on this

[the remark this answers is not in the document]
baileympearson: Cool, that's fine.  We don't have team practices for this, so whatever works.

Some people (myself included) believe that the actual test blocks should only contain the assertions relevant to the test.  A good example of this is Durran's recent PR - https://github.com/mongodb/node-mongodb-native/blob/e641bd44ef39d64b8b572e7a8ab6bfc71a2b4bed/test/unit/cmap/connection.test.ts#L162.  Others don't mind mixing setup into test code.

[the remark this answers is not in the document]
andymina: i changed my mind, lol. after adding more tests i appreciated the cleanliness of this approach. thanks!

baileympearson (changes requested):

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,35 @@ describe('Feature Flags', () => {
...
+      const client = new MongoClient('mongodb://localhost:27017', { [loggerFeatureFlag]: true });
+      expect(client.mongoLogger).to.be.instanceOf(MongoLogger);
+    });
+
+    it('should not instantiate a MongoLogger when set to false', () => {
+      const client = new MongoClient('mongodb://localhost:27017', { [loggerFeatureFlag]: false });
+      expect(client).property('mongoLogger', null);
+    });
+
+    it('should instantiate a MongoLogger when set to undefined', () => {
```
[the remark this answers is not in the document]
andymina: typo, fixed now!

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    const loggerOptions = MongoLogger.resolveOptions({ MONGODB_LOG_PATH: 'STDOUT' }, {});
+    const logger = new MongoLogger(loggerOptions);
+    expect(logger.logDestination).to.equal(process['stdout']);
+  });
+
+  it('sets the log destination to stderr if an invalid value is passed', function () {
+    const loggerOptions = MongoLogger.resolveOptions({ MONGODB_LOG_PATH: 'invalid' }, {});
+    const logger = new MongoLogger(loggerOptions);
+    expect(logger.logDestination).to.equal(process['stderr']);
+  });
```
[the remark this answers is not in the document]
andymina: moved to be captured under a constructor describe block like we discussed over zoom

src/mongo_logger.ts
```diff
@@ -0,0 +1,174 @@
...
+  return null;
+}
+
+/**
+ * Parses a string to be a number greater than or equal to 0 for maxDocumentLength.
+ *
+ * @param s - the value to be parsed
+ * @returns the int value parsed or 1000 if the value could not be parsed
+ */
+function parseMaxDocumentLength(s?: string): number {
```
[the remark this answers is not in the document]
andymina: 🤦🏾

test/unit/mongo_client.test.js
```diff
@@ -847,4 +848,68 @@ describe('MongoOptions', function () {
...
+              [loggerFeatureFlag]: true
+            });
+            expect(client).property('mongoLogger', null);
+          });
+        });
+      }
+    });
+
+    context('when there are no environment variables set', function () {
+      it('should not instatiate a MongoLogger if environment variables are not set', function () {
```
baileympearson, line 900:
```suggestion
      it('should not instatiate a MongoLogger', function () {
```

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,204 @@
...
+          it('defaults to stderr', function () {
+            const loggerOptions = MongoLogger.resolveOptions(
+              { MONGODB_LOG_PATH: '' },
+              { mongodbLogPath: undefined }
+            );
+            expect(loggerOptions.logDestination).to.equal('stderr');
+          });
+        });
+
+        context('when MONGODB_LOG_PATH is not an empty string', function () {
```
baileympearson, line 170:
```suggestion
        context('when MONGODB_LOG_PATH is a non-empty string', function () {
```
This applies to the test two below as well.

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,204 @@
...
+              MONGODB_LOG_MAX_DOCUMENT_LENGTH: '500'
+            },
+            {}
+          );
+          expect(loggerOptions.maxDocumentLength).to.equal(500);
+        });
+      });
+    });
+
+    context('logDestination', function () {
```
baileympearson, line 148: This looks great - I think we're missing a test for the condition where both MONGODB_LOG_PATH and mongodbLogPath are both non-empty strings, asserting that we preference the environment.

baileympearson (changes requested):
Three small comments, looks good otherwise!  I'm requesting changes for visibility but I'm going to put the PR into team review.

src/connection_string.ts
```diff
@@ -863,9 +886,12 @@ export const OPTIONS = {
   loggerLevel: {
     target: 'logger',
     transform({ values: [value] }) {
-      return new Logger('MongoClient', { loggerLevel: value as LoggerLevel });
+      return new LegacyLogger('MongoClient', { loggerLevel: value as LegacyLoggerLevel });
     }
   },
+  mongoLogger: {
```
dariakp, line 892: do we actually want to define this here? it's not actually an option, rather an api provided by the client

src/connection_string.ts
```diff
@@ -507,6 +500,36 @@ export function parseOptions(
...
+    const loggingComponents = [
+      loggerOptions.command,
+      loggerOptions.topology,
+      loggerOptions.serverSelection,
+      loggerOptions.connection,
+      loggerOptions.defaultSeverity
+    ];
+
+    if (loggingComponents.some(severity => severity !== SeverityLevel.OFF)) {
+      mongoOptions.mongoLogger = new MongoLogger(loggerOptions);
```
dariakp, line 529: why do we set the logger on the mongoOption at all? the instantiation should really be a part of the client instantiation, not options parsing; if we really want, we can add the parsed logger options to the mongo options; the point of the options is to provide a reference to the settings that went into configuring the client

test/unit/mongo_client.test.js
```diff
@@ -847,4 +848,68 @@ describe('MongoOptions', function () {
       });
     });
   });
+
+  context('MongoLogger', function () {
```
dariakp, line 852: as I mentioned in the other file, the job of the options parsing is to ensure that we are correctly parsing the options into the format expected by the logging class, so I would expect these tests to be updated to test that the options are set as expected;

I would actually go a step further and outsource the instantiation of the logger from the options to the logger class itself (i.e., let the logger class constructor return `null`), that way we have a clear division of responsibilities: the options parser consolidates options, the logger class decides what to do with them

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,35 @@ describe('Feature Flags', () => {
...
+    after(() => {
+      process.env = cachedEnv;
+    });
+
+    it('should instantiate a MongoLogger when set to true', () => {
+      const client = new MongoClient('mongodb://localhost:27017', { [loggerFeatureFlag]: true });
+      expect(client.mongoLogger).to.be.instanceOf(MongoLogger);
+    });
+
+    it('should not instantiate a MongoLogger when set to false', () => {
```
dariakp, line 67: I think we should make the property not exist when set to false or not set

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,35 @@ describe('Feature Flags', () => {
...
     }
   });
+
+  describe('@@mdb.enableMongoLogger', () => {
+    let cachedEnv;
+    const loggerFeatureFlag = Symbol.for('@@mdb.enableMongoLogger');
+
+    before(() => {
+      cachedEnv = process.env;
+      process.env['MONGODB_LOG_COMMAND'] = 'emergency';
```
dariakp, line 55: probably want to split this into two `context` blocks, one for when logging for a component is enabled, and one for when it is not enabled

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,204 @@
+import { expect } from 'chai';
+
+import { MongoLogger, SeverityLevel } from '../../src/mongo_logger';
+
+describe('class MongoLogger', function () {
+  describe('#constructor', function () {
+    it('treats the log destination value of stderr as case-insensitve', function () {
+      const loggerOptions = MongoLogger.resolveOptions({ MONGODB_LOG_PATH: 'STDERR' }, {});
```
dariakp, line 8: I think we are mix and matching the `resolveOptions` unit testing with the properties of the constructor itself; be precise - is it the options resolver that's responsible for the standardization of the option spelling or the constructor? if it's the options resolver, then this should be tested there, if it's the constructor, then you shouldn't pre-process what you are passing into it (i.e., provide the exact input to the constructor that you are testing)

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,204 @@
...
+            },
+            {}
+          );
+          expect(loggerOptions.maxDocumentLength).to.equal(500);
+        });
+      });
+    });
+
+    context('logDestination', function () {
+      context('when mongodbLogPath is undefined', function () {
```
dariakp, line 149: I think the contexts should either be split by the value of `mongodbLogPath` or by the value of `MONGODB_LOG_PATH` - mix and matching them in the top level describe is confusing; the tests here would be best served by a table, so that we can tell that we are testing all combinations of interest:
```js
const logDestinationTestCases = [
  {
     description: 'should default to stderr when mongoLogPath and MONGODB_LOG_PATH are undefined',
     MONGODB_LOG_PATH: undefined,
     mongodbLogPath: undefined,
     expectedDestination: 'stderr'
  },
  ...
];

for ... of
   it(testCase.description, () => {
     const loggerOptions = MongoLogger.resolveOptions({ MONGODB_LOG_PATH }, { mongodbLogPath });
     expect(loggerOptions).to.have.property('logDestination', expectedDestination);
   })
```

dariakp (changes requested):
I reviewed everything except the `mongo_logger` file itself because I want to make sure we have the external behavior ironed out first via relevant tests; I'll come back to the mongo_logger file once my other comments are addressed

src/mongo_logger.ts
```diff
@@ -0,0 +1,184 @@
...
+  OFF: 'off'
+} as const);
+
+/** @internal */
+export type SeverityLevel = typeof SeverityLevel[keyof typeof SeverityLevel];
+
+/**
+ * Parses a string as one of SeverityLevel
+ *
+ * @param name - the name of the variable to be parsed
```
nbbeeken, line 25: Just a bit of cleanup, `name` is not a param

test/unit/mongo_logger.test.ts
baileympearson: I agree about mixing and matching the top level contexts - but I disagree about the table being easier to read.  I suggested the nested context approach for readability both in tests and test output.  The long, oneline descriptions make the test output harder to grok because the relevant conditions are at the end of each line instead of forming a new section in the output (like they do with nested contexts).  I also don't think the table approach makes it any more clear that we're testing all relevant cases?

src/mongo_logger.ts
```diff
@@ -0,0 +1,184 @@
...
+/** @internal */
+export class MongoLogger {
+  componentSeverities: Record<MongoLoggableComponent, SeverityLevel>;
+  maxDocumentLength: number;
+  logDestination: Writable;
+
+  constructor(options: MongoLoggerOptions) {
+    // TODO(NODE-4849): add filepath and Writable support
+    this.logDestination =
+      options.logDestination.toLowerCase() === 'stdout' ? process['stdout'] : process['stderr'];
```
nbbeeken, line 128:
```suggestion
      options.logDestination.toLowerCase() === 'stdout' ? process.stdout : process.stderr;
```
Just a nit

test/unit/mongo_client.test.js
baileympearson: Can you elaborate on this?  The logger class constructor currently returns null.  With the exception of one option (that could be moved to resolveOptions), there is already a clear separation of responsibilities?

These tests are responsible for testing whether or not the actual logger gets instantiated on the MongoClient, not anything related to options parsing.

src/mongo_logger.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    return {
+      command: parseSeverityFromString(envOptions.MONGODB_LOG_COMMAND) ?? defaultSeverity,
+      topology: parseSeverityFromString(envOptions.MONGODB_LOG_TOPOLOGY) ?? defaultSeverity,
+      serverSelection:
+        parseSeverityFromString(envOptions.MONGODB_LOG_SERVER_SELECTION) ?? defaultSeverity,
+      connection: parseSeverityFromString(envOptions.MONGODB_LOG_CONNECTION) ?? defaultSeverity,
+      defaultSeverity,
+      maxDocumentLength: parseMaxDocumentLength(envOptions.MONGODB_LOG_MAX_DOCUMENT_LENGTH),
+      logDestination: parseLogDestination(envOptions.MONGODB_LOG_PATH, clientOptions.mongodbLogPath)
```
addaleax, line 181: - Is this going to be the only option configurable via `clientOptions`?
- My understanding from the logging slack channel discussion was that we’ve settled on `clientOptions` winning over `envOptions`, right?

nbbeeken (commented):
some quick cleanups I noticed

src/mongo_logger.ts
dariakp: I think the design that this PR was based on did not get updated as the result of those conversations; I think we should support all options via the client, so we probably want to decide on appropriate option names for that, but I think we can do it in a separate ticket for ease of review - @andymina can you make sure to file the follow - up?

test/unit/mongo_logger.test.ts
dariakp: it's easier to check that we got all the combinations covered by having
```
MONGODB_LOG_PATH: <value>,
mongodbLogPath: <value>,
```
next to each other than by having them buried inside `it` blocks; it's also easier to make sure that the expectation is correct by just checking the `expectedDestination` variable instead of reading the whole expect assertion sentence;
and since the setup is identical for all of these, it's much easier to make sure it stays consistent if we don't replicate it across 9+ blocks;

if the concern is just the shorter lines, those aren't mutually exclusive with this setup, just remove the description variable and generate it from the 3 variables in the test case, then you can do a nested loop and add the `mongodbLogPath` or the env var as a context block and filter the cases on that value

dariakp: if you really wanted to, you could even just have an array like this:
```js
// Format: [ MONGODB_LOG_PATH, [mongodbLogPath], expectedDestination ]
[
  [ undefined, undefined, 'stderr' ],
  [ undefined, 'invalid', 'stderr'],
  ...
]
```

test/unit/mongo_client.test.js
dariakp: this context `MongoLogger` is nested inside the `MongoOptions` context, but `mongoLogger` should not be a member of `MongoOptions` at all, we should have the **parsed** options be saved onto the `MongoOptions` object, and this file should be updated to only test that these standardized options are correctly assigned to the options object (we can even keep `mongoLogger` but make it reference the options passed to it instead of the actual instance of the logger)

the options parsing should **not** generate the logger;
the client should be updated to have:
```
// instead of this: this.mongoLogger = this[kOptions].mongoLogger ?? null;

// do this:
this.mongoLogger = new MongoLogger(<this[kOptions].mongoLoggerOptions>);

// or this:
this.mongoLogger = MongoLogger.create(<this[kOptions].mongoLoggerOptions>);
// should return an instance of the logger if some options are on, otherwise can return null
```

with that change, this file should only test that the options are correctly assigned to the client options object
and the mongo logger test should make sure that all combinations of options result in the correct return result from `new MongoLogger()` or `MongoLogger.create()`

The problem with the current implementation is that we don't preserve what options are used to configure the logger, which is suboptimal for debugging, and we have the logic of the resulting logging scattered across 3 places: options parsing, client constructor, and the logger class itself. The job of options parsing should be only to validate the options and transform them into a standardized form. The job of the client constructor should be to initialize whatever APIs it needs to have. The job of the MongoLogger class should be to determine what to do with the options, including what to do when all options are off.

The only reason this comment is in the testing section is because once we do have the options saved on the options object, we'll want to make sure that they are tested here. If we just return the result of the logger option parser, then we can also just use sinon to make sure that that's what we assign to the corresponding client options.

src/connection_string.ts
```diff
@@ -507,6 +500,36 @@ export function parseOptions(
...
+      MONGODB_LOG_CONNECTION: process.env.MONGODB_LOG_CONNECTION,
+      MONGODB_LOG_ALL: process.env.MONGODB_LOG_ALL,
+      MONGODB_LOG_MAX_DOCUMENT_LENGTH: process.env.MONGODB_LOG_MAX_DOCUMENT_LENGTH,
+      MONGODB_LOG_PATH: process.env.MONGODB_LOG_PATH
+    };
+    const loggerMongoClientOptions: MongoLoggerMongoClientOptions = {
+      mongodbLogPath: mongoOptions.mongodbLogPath
+    };
+    const loggerOptions = MongoLogger.resolveOptions(loggerEnvOptions, loggerMongoClientOptions);
+    const loggingComponents = [
```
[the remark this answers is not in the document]
andymina: thanks for the feedback on this! i opted to go with optionB and moving the default inside bc i thought it looked cleaner too :)

src/connection_string.ts
andymina: replaced by `mongoLoggerOptions`

test/unit/mongo_client.test.js
andymina: bailey beat me to it, but thanks for elaborating. really helped me clear up a lot the separation of responsibilities. unfortunately no relevant code changes because of our team conversation from slack where we agreed that `mongoLogger` should always be present :(

test/integration/node-specific/feature_flags.test.ts
andymina: capturing the team convo from slack: the mongoLogger should always be present but should no op when the feature flag isn't true and logging isn't enabled

test/integration/node-specific/feature_flags.test.ts
andymina: added!

test/unit/mongo_logger.test.ts
andymina: fixed and separated

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,114 @@ describe('Feature Flags', () => {
...
+                ? 'TODO(NODE-4813): implement the emergency severity logging method'
+                : 'TODO(NODE-4814): implement the remaining severity loggers';
+            it.skip('should not be a no-op', () => {
+              const client = new MongoClient('mongodb://localhost:27017', {
+                [loggerFeatureFlag]: true
+              });
+              const stringifiedMethod = client.mongoLogger[severity].toString();
+              const expectedStringifiedMethod = `${severity}(component, message) { }`;
+              expect(stringifiedMethod).to.not.equal(expectedStringifiedMethod);
+            }).skipReason = skipReason;
```
baileympearson, line 91: These tests should be same as the tests below and not skipped, maybe with a comment explaining that for now, the behavior is the same when the feature flag is enabled as it is when the flag is disabled.

We shouldn't write tests for future behavior and skip them; instead we should write tests for the current behavior.  When we implement these methods in the future, these tests will fail.  That's expected, and these tests can then be updated to match the new behavior.

test/unit/mongo_logger.test.ts
dariakp: we actually are interested in testing a lot of permutations, and that makes the maintainability burden of copy-pasted case instances non-trivial in multiple ways:
1) Updating the code implementation (for example, nesting the logDestination in another object, or changing the format of the input to `resolveOptions`) will require changes in a dozen different places instead of just one
2) Updating the test implementation will require changes in a dozen different places instead of just one
3) Copy pasting is the root of all evil: and it's super easy to miss an incorrectly named test or an incorrect assertion when scanning over a dozen test cases all of which look approximately the same
4) Noticing inconsistencies in implementation is virtually impossible when test cases are broken up across hundreds of lines (e.g., if you decide that the way we treat `undefined` in the logger implementation should be updated, you might not think about the same case for `""` because it's a hundred lines down)

Point 2 above is particularly relevant, because we should use `expect(loggerOptions).to.have.property('logDestination', <value>)` style of assertions throughout this file because we want to have helpful errors. This would be a one line change in a table test because there is a clear separation between the expected input-output expectations of the test and the algorithm that we want to execute to test the input-output combinations.

dariakp (changes requested):

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,270 @@
...
+            {
+              MONGODB_LOG_MAX_DOCUMENT_LENGTH: 'invalid'
+            },
+            {}
+          );
+          expect(loggerOptions.maxDocumentLength).to.equal(1000);
+        });
+      });
+
+      context('when MONGODB_LOG_MAX_DOCUMENT_LENGTH can be parsed a uint', function () {
```
dariakp, line 161:
```suggestion
      context('when MONGODB_LOG_MAX_DOCUMENT_LENGTH can be parsed as a uint', function () {
```

test/unit/mongo_client.test.js
dariakp: yes, there are changes - we should add tests to make sure that the expected `mongoLoggerOptions` are getting set on the options attached to the client

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,270 @@
+import { expect } from 'chai';
+
+import { MongoLogger, SeverityLevel } from '../../src/mongo_logger';
+
+describe('class MongoLogger', function () {
+  describe('#constructor', function () {
+    context('when an invalid value is passed', function () {
```
dariakp, line 7: I don't understand the purpose of this context, 1) the constructor takes many possible values, so one test is not sufficient, and 2) the tests under the `#constructor` need to not use `resolveOptions` at all, we need to be specific about exactly what is being passed into the constructor.

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    });
+
+    context('when component severities are not set or invalid', function () {
+      it('only uses the default severity for those components', function () {
+        const loggerOptions = MongoLogger.resolveOptions(
+          {
+            MONGODB_LOG_COMMAND: '',
+            MONGODB_LOG_TOPOLOGY: undefined,
+            MONGODB_LOG_SERVER_SELECTION: 'invalid',
+            MONGODB_LOG_CONNECTION: SeverityLevel.EMERGENCY,
```
dariakp, line 90: this is not an invalid value

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,270 @@
...
+        'severitySettings.connection',
+        SeverityLevel.DEBUG
+      );
+      expect(loggerOptions).to.have.nested.property(
+        'severitySettings.default',
+        SeverityLevel.WARNING
+      );
+    });
+
+    context('when component severities are not set or invalid', function () {
```
dariakp, line 83: you are missing a case of what happens when the default severity level (`MONGODB_LOG_ALL`) is also invalid

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,270 @@
...
+        );
+        expect(loggerOptions).to.have.nested.property(
+          'severitySettings.connection',
+          loggerOptions.severitySettings.default
+        );
+      });
+    });
+
+    context('when environment variables and client options are both set', function () {
+      it('gives precedence to environment variables', function () {
```
dariakp, line 115: we actually want the opposite here

dariakp (changes requested):

src/mongo_logger.ts
```diff
@@ -0,0 +1,197 @@
...
+    /** Severity level for connection component */
+    connection: SeverityLevel;
+    /** Default severity level to be used if any of the above are unset */
+    default: SeverityLevel;
+  };
+
+  /** Max length of embedded EJSON docs. Setting to 0 disables truncation. Defaults to 1000. */
+  maxDocumentLength: number;
+  /** Destination for log messages. Must be 'stderr' or 'stdout'. Defaults to 'stderr'. */
+  logDestination: string;
```
dariakp, line 126: we need to be able to pipe to a stream - is there a separate ticket to add it? it needs to either be added here, or we need to reference the follow up ticket indicating that the implementation is incomplete; we will not be able to test anything from the spec if we cannot take a custom stream destination

dariakp (changes requested):

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,270 @@
...
+          it('treats the value as case-insensitive', function () {
+            const loggerOptions = MongoLogger.resolveOptions(
+              { MONGODB_LOG_PATH: undefined },
+              { mongodbLogPath: 'STDERR' }
+            );
+            expect(loggerOptions).to.have.property('logDestination', 'stderr');
+          });
+        });
+
+        context('when mongodbLogPath is a non-empty string', function () {
```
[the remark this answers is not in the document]
baileympearson: Agreed about using an invalid non-empty string, will update the description and probably the test structure.  

We do test the empty string already:
```typescript
context('when mongodbLogPath is an empty string', function () {
  it('defaults to stderr', function () {
    const loggerOptions = MongoLogger.resolveOptions(
      { MONGODB_LOG_PATH: undefined },
      { mongodbLogPath: '' }
    );
    expect(loggerOptions.logDestination).to.equal('stderr');
  });
});
```

w.r.t. `null` or `0` or other weird values - this is a larger discussion.  

The TS type for `mongodbLogPath` is `string | undefined.  Other falsey values (`0`, `null`, ect) are prohibited by Typescript.  An argument could be made that since we also ship the compiled JS, we should test other values too.  But thus far, our pattern has been not to test for values outside of our Typescript and my opinion is that it is perfectly reasonable to only test edge cases within our types.  Our types define the api contract of our functions for both our js and typescript users - anyone providing an object that doesn't match our type definition is working outside the bounds of our supported API.

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+      });
+
+      expect(logger).to.have.property('componentSeverities', componentSeverities);
+      expect(logger).to.have.property('maxDocumentLength', 10);
+      expect(logger).to.have.property('logDestination', stream);
+    });
+  });
+
+  describe('static #resolveOptions()', function () {
+    describe('componentSeverities', function () {
```
baileympearson, line 56: I did not table test the component severities - there was no straightforward table-test that I could think of.  I'm happy to consider alternate approaches though.

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,87 @@ describe('Feature Flags', () => {
...
+      process.env = cachedEnv;
+    });
+
+    context('when enabled', () => {
+      context('when logging is enabled for any component', () => {
+        before(() => {
+          process.env.MONGODB_LOG_COMMAND = SeverityLevel.EMERGENCY;
+        });
+
+        it('enables logging for MONGODB_LOG_COMMAND', () => {
```
dariakp, line 67: There is nothing wrong with using a specific component as a stand in for all components, but we should make sure to clarify that the point of the test is that the specified component is enabled. Otherwise we're saying that regardless of which component is set, the COMMAND component is enabled.
```suggestion
        it('enables logging for the specified component', () => {
```

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,87 @@ describe('Feature Flags', () => {
...
+          );
+        });
+      });
+
+      context('when logging is not enabled for any component', () => {
+        before(() => {
+          process.env = {};
+        });
+
+        it('does not enable logging', () => {
```
dariakp, line 83:
```suggestion
        it('does not enable logging for any component', () => {
```

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,87 @@ describe('Feature Flags', () => {
...
+          const client = new MongoClient('mongodb://localhost:27017', {
+            [loggerFeatureFlag]: true
+          });
+          expect(client.mongoLogger.componentSeverities).to.have.property(
```
dariakp, line 87: we may as well add expectations for the rest of the available components (`for..of` over the possible components + default would work)

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,87 @@ describe('Feature Flags', () => {
...
+            'command',
+            SeverityLevel.OFF
+          );
+        });
+      });
+    });
+
+    for (const featureFlagValue of [false, undefined]) {
+      context(`when set to ${featureFlagValue}`, () => {
+        context('when logging is enabled for any component', () => {
```
dariakp, line 97:
```suggestion
        context('when logging is enabled for a component', () => {
```

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,87 @@ describe('Feature Flags', () => {
...
+          before(() => {
+            process.env['MONGODB_LOG_COMMAND'] = SeverityLevel.EMERGENCY;
+          });
+
+          it('does not enable logging', () => {
```
dariakp, line 102:
```suggestion
          it('does not enable logging for any component', () => {
```

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,87 @@ describe('Feature Flags', () => {
...
+            const client = new MongoClient('mongodb://localhost:27017', {
+              [loggerFeatureFlag]: featureFlagValue
+            });
+            expect(client.mongoLogger.componentSeverities).to.have.property(
```
dariakp, line 106: same comment here about checking all components

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,87 @@ describe('Feature Flags', () => {
...
+            );
+          });
+        });
+
+        context('when logging is not enabled for any component', () => {
+          before(() => {
+            process.env = {};
+          });
+
+          it('does not enable logging', () => {
```
dariakp, line 118:
```suggestion
          it('does not enable logging for any component', () => {
```

test/integration/node-specific/feature_flags.test.ts
```diff
@@ -42,4 +45,87 @@ describe('Feature Flags', () => {
...
+            const client = new MongoClient('mongodb://localhost:27017', {
+              [loggerFeatureFlag]: featureFlagValue
+            });
+            expect(client.mongoLogger.componentSeverities).to.have.property(
```
dariakp, line 122: same comment here about checking all components

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+          const stream = new BufferingStream();
+          const logger = new MongoLogger({
+            componentSeverities: {
+              topology: 'off'
+            } as any,
+            logDestination: stream
+          } as any);
+
+          logger[severityLevel]('topology', 'message');
+          expect(stream.buffer).to.have.lengthOf(0);
```
dariakp, line 342: quick question, do we expect the actual call to write to happen synchronously here? (I see the meta-tests above to ensure that the buffer is filled synchronously, but it may well be possible that we have some async steps between the call to the logger and the call to the stream.write) if not, could be worth adding a timeout of some sort to give the event loop some time to process (basically, when we do implement the methods, I would expect the same logic to be able to be used in the positive cases)

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+      }
+    });
+
+    context('logDestination', function () {
+      const stream = new Writable();
+      const tests: Array<{
+        env: 'stderr' | 'stdout' | undefined;
+        client: MongoLoggerMongoClientOptions['mongodbLogPath'] | undefined;
+        expectedLogDestination: MongoLoggerOptions['logDestination'];
+      }> = [
```
dariakp, line 241: we'll want to add the case-insensitive cases for logDestination and also the `<empty string>` cases (just "STDOUT" as a general stand in for case-insensitive is fine, but want to make sure both it and the empty string case is tested in both env and as a client option)

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+              { MONGODB_LOG_MAX_DOCUMENT_LENGTH: input },
+              {}
+            );
+            expect(options.maxDocumentLength).to.equal(expected);
+          });
+        });
+      }
+    });
+
+    context('logDestination', function () {
```
dariakp, line 235: we'll also want to add tests for invalid options

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+            const options = MongoLogger.resolveOptions(
+              { MONGODB_LOG_ALL: SeverityLevel.ALERT },
+              {}
+            );
+            expect(options.componentSeverities).to.have.property('default', SeverityLevel.ALERT);
+          });
+        });
+        context('when no value is provided', () => {
+          it('sets default to off', () => {
+            const options = MongoLogger.resolveOptions({ MONGODB_LOG_ALL: SeverityLevel.OFF }, {});
```
dariakp, line 85: this is actually not testing when no value is provided, it's testing when the explicit `OFF` value is provided

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+        context('when a default is provided', () => {
+          it('sets default to the provided value', () => {
+            const options = MongoLogger.resolveOptions(
+              { MONGODB_LOG_ALL: SeverityLevel.ALERT },
+              {}
+            );
+            expect(options.componentSeverities).to.have.property('default', SeverityLevel.ALERT);
+          });
+        });
+        context('when no value is provided', () => {
```
dariakp, line 83: needs a case for when an invalid value is provided

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+      });
+
+      for (const component of components) {
+        const mappedComponent = mapToInternalRepresentation(component);
+        context(`${component}`, function () {
+          context(`when set to a valid value in the environment`, function () {
+            context('when there is a default provided', function () {
+              it(`sets ${mappedComponent} to the provided value and ignores the default`, function () {
+                const options = MongoLogger.resolveOptions(
+                  { [component]: SeverityLevel.ALERT, MONGODB_LOG_ALL: SeverityLevel.OFF },
```
dariakp, line 103: the default should be something other than `OFF` for this case, and then we need a separate case for when the default is `OFF`

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+                expect(options.componentSeverities).to.have.property(
+                  mappedComponent,
+                  SeverityLevel.ALERT
+                );
+              });
+            });
+            context('when there is no default provided', function () {
+              it(`sets ${mappedComponent} to the provided value`, function () {
+                const options = MongoLogger.resolveOptions(
+                  { [component]: SeverityLevel.ALERT, MONGODB_LOG_ALL: SeverityLevel.OFF },
```
dariakp, line 115: this needs to not provide the default value, not just set it to `OFF`

dariakp (changes requested):
Almost there

test/unit/mongo_logger.test.ts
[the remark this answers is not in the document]
dariakp: I left a comment regarding this in a separate thread in my most recent set of comments for easier discussion

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,347 @@
...
+      for (const { env, client, expectedLogDestination } of tests) {
+        context(
+          `environment option=${env}, client option=${
+            client instanceof Writable ? 'a writable stream' : client
+          }`,
+          () => {
+            it(`sets the log destination to ${
+              expectedLogDestination instanceof Writable
+                ? 'the provided writable stream'
+                : expectedLogDestination
```
dariakp, line 313: just a note: the expectedLogDestination is always a writable stream now

test/unit/mongo_logger.test.ts
dariakp: it's fine as is, but I added a couple of notes on missing cases

test/unit/mongo_logger.test.ts
baileympearson: i thought about that, but decided against it because 

1. i foresee no timeouts / async work in the logging scope
2. if there is, who knows how long of a timeout is sufficient now?  we don't know the extent of the async work

but I'm happy to adjust this

test/unit/mongo_logger.test.ts
baileympearson: fixed

src/mongo_logger.ts
dariakp: I filed NODE-4878

dariakp (changes requested):
Please address the comments in the feature_flags file; otherwise, just a few clean up comments on the mongo_logger.test.ts around the severity helpers and defaults; the testing is exhaustive, and there are improvements that could be made there to condense them by analyzing which cases are redundant (i.e., we want to test each component and each severity level at least once, but testing every component with every severity is overkill; likewise, when testing defaults, just one test case where nothing is set and assertions for the default on each expected thing is enough). If we want to just get these changes in, I am not going to insist on reworking the test cases right now (because they do cover everything), but then we'll need to revisit this and rewrite them in a more maintainable way when we go to implement client option handling for these settings.

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,532 @@
...
+        for (const option of validOptions) {
+          yield [option, option];
+          yield [option.toUpperCase(), option];
+        }
+      }
+
+      const invalidOptions = ['', 'invalid-string'];
+      const validNonDefaultOptions = new Map(makeValidOptions());
+
+      context('default', () => {
```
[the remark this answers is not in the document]
baileympearson: Yup, accidentally wrapped all the contexts instead of only the first 3.

test/unit/mongo_logger.test.ts
```diff
@@ -0,0 +1,528 @@
...
+                      },
+                      {}
+                    );
+                    expect(componentSeverities[componentSeverityOption]).to.equal(expectedValue);
+                  });
+                }
+              );
+            }
+            it(`sets ${componentSeverityOption} to the value of MONGODB_LOG_ALL`, () => {
+              expect(true).to.be.true;
```
dariakp, line 216: this needs to be removed

dariakp (changes requested):

dariakp (approved):

src/mongo_logger.ts
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * Parses a string to be a number greater than or equal to 0 for maxDocumentLength.
+ *
+ * @param s - the value to be parsed
+ * @returns the int value parsed or 1000 if the value could not be parsed
+ */
+function parseMaxDocumentLength(s?: string): number {
+  try {
+    const maxDocumentLength = getUint('MONGODB_LOG_MAX_DOCUMENT_LENGTH', s);
```
nbbeeken, line 104: Since we're not actually using the validation that this function provides (the catch squashes the thrown error) could the helper be put back into connection_string.ts and we use Number.parseInt + Number.isNaN here?

src/mongo_logger.ts
```diff
@@ -0,0 +1,214 @@
...
+function resolveLogPath(
+  { MONGODB_LOG_PATH }: MongoLoggerEnvOptions,
+  {
+    mongodbLogPath
+  }: {
+    mongodbLogPath?: unknown;
+  }
+): Writable {
+  const isValidLogDestinationString = (destination: string) =>
+    ['stdout', 'stderr'].includes(destination.toLowerCase());
```
nbbeeken, line 126: This is fine, but it did catch my attention as slightly unusual structuring, specifically creating a new function for a condition check.

I'd expect `['stdout', 'stderr']` to but pulled up to a module variable and then each if stmt below to call `VALID_DESTS.includes(destination.toLowerCase())`

src/mongo_logger.ts
```diff
@@ -0,0 +1,214 @@
...
+  if (typeof mongodbLogPath === 'string' && isValidLogDestinationString(mongodbLogPath)) {
+    return mongodbLogPath.toLowerCase() === 'stderr' ? process.stderr : process.stdout;
+  }
+
+  if (typeof mongodbLogPath === 'object' && mongodbLogPath instanceof Writable) {
```
nbbeeken, line 131: It's obviously not wrong for us to check `instanceof Writable` here, but we may want to consider trusting the user has passed us something with a compatible interface that isn't an instanceof a Node.js stream. This may be easier to remove once we have a clear idea about the exact APIs we need to exist (ex. `.write()`, `.flush()` etc. w.e. is minimally necessary)

src/mongo_logger.ts
dariakp: do we know for a fact that we won't validate this in the future? (particularly, when it comes from client options)

src/mongo_logger.ts
dariakp: I had this thought, too, but my original comment is now lost somewhere here or in the depths of slack; thanks for catching that - we should leave a `TODO(NODE-xxxx): check for minimal interface instead of instanceof Writable` with the ticket that will actually implement the logging methods

src/mongo_logger.ts
baileympearson: Well, we sort of use the validation.  `getUInt` also asserts that it's non-negative.  We currently catch+default instead of throwing but I think that it still makes sense to rely on the function we already have that parses uints from strings

src/mongo_logger.ts
baileympearson: Interesting, this is the approach I always use.  

1. no reason for top-level variables if they're used in one place, so place them in the function to not pollute the module's scope
2. use descriptive names

but i don't feel strongly, I can change if you want

nbbeeken: No worries, just caught my eye as different, all good 👍

src/mongo_logger.ts
nbbeeken: We might validate the options when they come from the client options but that will happen in the options parser. It's just about using try/catch for something that's a condition check, it's not the control flow you'd expect when you need to make sure something is greater than zero

src/mongo_logger.ts
dariakp: @baileympearson Instead of the empty [NODE-4886](https://jira.mongodb.org/browse/NODE-4886), can we instead update the description and AC of [NODE-4813](https://jira.mongodb.org/browse/NODE-4813)?

src/utils.ts
```diff
@@ -1422,17 +1422,15 @@ export function compareObjectId(oid1?: ObjectId | null, oid2?: ObjectId | null):
   return oid1.id.compare(oid2.id);
 }
 
-export function getInt(name: string, value: unknown): number {
+export function parseInt(value: unknown): number | null {
```
nbbeeken, line 1425:
```suggestion
export function parseInteger(value: unknown): number | null {
```

By shadowing a global with a hoisted declaration you would change all the code in the file to use this function instead of the global. Renaming it is a quick fix but ideally a linter would catch this if not tests. (not something I'm suggesting we look into here)

Also, luckily we currently don't have any code using parseInt, I've always used `Number.parseInt` but there's no lint or rule that made me do that, just 🧅

src/connection_string.ts
```diff
@@ -33,11 +33,12 @@ import {
   DEFAULT_PK_FACTORY,
   emitWarning,
   emitWarningOnce,
-  getInt,
-  getUint,
+  getInt as getIntFromOptions,
+  getUint as getUIntFromOptions,
```
nbbeeken: A bit of lint clean up here

nbbeeken (changes requested):

nbbeeken (approved):

dariakp (approved):

avaly:
I can't find any documentation about this new Logger in the [API docs](https://mongodb.github.io/node-mongodb-native/5.0/) or the [v5.0.0 changes doc](https://github.com/mongodb/node-mongodb-native/blob/main/etc/notes/CHANGES_5.0.0.md). Where can we read more about this?

baileympearson:
Hey @avaly - the new logger is still in development, is currently unused inside the driver and is not intended to be consumed yet.  Once development is finished, we'll include information and links to relevant documentation in the release notes for whichever release it is a part of.

PR #3477, opened 2022-11-22
fix(NODE-4831): check map value is not undefined

durran:
### Description

When no operation description is in the monitoring connection's queue map and a message comes in the application would error _"TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))"._

#### What is changing?

The iterator `next()` call when nothing is in the map will return `{ value: undefined, done: true }` - so we change the existence check to look at the actual value.

No

NODE-4831

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -384,8 +384,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       } else {
         // Get the first orphaned operation description.
         const entry = this[kQueue].entries().next();
-        /* eslint no-restricted-syntax: 0 */
-        if (entry.value !== undefined) {
+        if (entry.value != null) {
           const [requestId, orphaned]: [number, OperationDescription] = entry.value;
           // If the orphaned operation description exists then set it.
           operationDescription = orphaned;
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -384,7 +384,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       } else {
         // Get the first orphaned operation description.
         const entry = this[kQueue].entries().next();
-        if (entry) {
+        /* eslint no-restricted-syntax: 0 */
+        if (entry.value !== undefined) {
           const [requestId, orphaned]: [number, OperationDescription] = entry.value;
           // If the orphaned operation description exists then set it.
           operationDescription = orphaned;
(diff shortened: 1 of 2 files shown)


src/cmap/connection.ts
```diff
@@ -384,7 +384,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       } else {
         // Get the first orphaned operation description.
         const entry = this[kQueue].entries().next();
-        if (entry) {
+        /* eslint no-restricted-syntax: 0 */
```
durran, line 387: Eslint wants us to check for falsey here but we explicitly want the check to be for `undefined`.

nbbeeken: eslint wants you to use `entry.value != null` (double, not triple eq) which should cover what we're trying to gate here, but its true this will only be undefined explicitly so I don't have a strong onion otherwise

src/cmap/connection.ts
```diff
@@ -384,7 +384,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+        if (entry.value !== undefined) {
```
dariakp, line 388: What's the downside to checking for "not null" other than that we never expect it to be actually null?

test/unit/cmap/connection.test.ts
```diff
@@ -287,6 +287,34 @@ describe('new Connection()', function () {
...
+            requestId: 2,
+            responseTo: 1,
+            opCode: msg.readInt32LE(12)
+          };
+          const msgBody = msg.subarray(16);
+
+          const message = new BinMsg(msg, msgHeader, msgBody);
+          expect(() => {
+            connection.onMessage(message);
+          }).to.not.throw(/undefined is not iterable/);
```
dariakp, line 314: I think this is one of the cases where we should expect it to not throw at all, i.e., `to.not.throw()`

dariakp (changes requested):

test/unit/cmap/connection.test.ts
durran: Updated to check that is just doesnt throw.

dariakp (approved):

src/cmap/connection.ts
durran: I never expected null, only undefined, but changing to check for not null actually works fine as well and does not need the eslint override. So I updated it.

nbbeeken (dismissed):

baileympearson (dismissed):

durran:
Rebased against main again.

baileympearson (approved):

PR #3478, opened 2022-11-28
ci(NODE-4842): add latest Node version to CI

baileympearson:
#### What is changing?

Adds node latest as a smoke test into CI.  Currently, it only adds a variant for rhel8 with all tasks.

Nope.

- [] PR title follows the correct format: `<type>(NODE-xxxx)<!>: <description>`


.evergreen/generate_evergreen_tasks.js
```diff
@@ -404,11 +404,27 @@ for (const
 
     BUILD_VARIANTS.push({ name, display_name, run_on, expansions, tasks: taskNames });
   };
+
+  const configureLatestNodeSmokeTest = os.match(/^rhel/);
+  if (configureLatestNodeSmokeTest) {
```
baileympearson, line 409: AC wasn't clear (shame on me for creating the ticket) but I chose to configure the latest node smoke test to only run rhel8.  I'm open to reworking it to run on more of we want to.

nbbeeken: Given that it's a smoke test I think its fine, but Windows would be "interesting" (so would macos but I think less so). Can we just make a note on the ticket (AC or comment) that we only tested on linux so its visible if folks look back at what the "latest" covered.

.evergreen/install-dependencies.sh
```diff
@@ -19,42 +19,42 @@ mkdir -p ${NODE_ARTIFACTS_PATH}
...
+  export NODE_VERSION=$VERSION
+
+  if [ $VERSION != 'node' ]
+  then
+    NODE_VERSION=$(curl --retry 8 --retry-delay 5  --max-time 50 --silent -o- https://nodejs.org/download/release/latest-v${VERSION}.x/SHASUMS256.txt | head -n 1 | awk '{print $2};' | cut -d- -f2)
+    export NODE_VERSION=${NODE_VERSION:1} # :1 gets rid of the leading 'v'
+  fi
+}
+
+get_node_version
```
nbbeeken, line 57: You have to pipe through the shell script argument here, `$1` does not refer to the same thing in the `*)` case

.evergreen/install-dependencies.sh
```diff
@@ -19,42 +19,42 @@ mkdir -p ${NODE_ARTIFACTS_PATH}
...
+      ;;
+  esac
+
+  export NODE_VERSION=$VERSION
+
+  if [ $VERSION != 'node' ]
+  then
+    NODE_VERSION=$(curl --retry 8 --retry-delay 5  --max-time 50 --silent -o- https://nodejs.org/download/release/latest-v${VERSION}.x/SHASUMS256.txt | head -n 1 | awk '{print $2};' | cut -d- -f2)
+    export NODE_VERSION=${NODE_VERSION:1} # :1 gets rid of the leading 'v'
+  fi
```
nbbeeken, line 54: Same refactor mentioned on BSON with respect to the URL here, if we're using latest I think we can just rely on fetching that directory from nodejs releases

.evergreen/install-dependencies.sh
```diff
@@ -19,42 +19,42 @@ mkdir -p ${NODE_ARTIFACTS_PATH}
...
-  *)
-    echo "Unsupported Node LTS version $1"
-    exit 1
-    ;;
-esac
-
-NODE_VERSION=$(curl --retry 8 --retry-delay 5  --max-time 50 --silent -o- https://nodejs.org/download/release/latest-v${VERSION}.x/SHASUMS256.txt | head -n 1 | awk '{print $2};' | cut -d- -f2)
-export NODE_VERSION=${NODE_VERSION:1} # :1 gets rid of the leading 'v'
+get_node_version() {
+  case $NODE_LTS_NAME in
```
nbbeeken, line 23: Since this still relies on global env vars what does the function syntax provide here? I don't disagree with the change, but perhaps we want to pass in `$NODE_LTS_NAME` and return `$NODE_VERSION`.

test/tools/runner/hooks/configuration.js
```diff
@@ -168,7 +168,11 @@ const beforeAllPluginImports = () => {
 };
 
 function installNode18DNSHooks() {
-  if (process.version.startsWith('v18')) {
+  if (
+    process.version.startsWith('v18') ||
+    process.version.startsWith('v19') ||
+    process.env.startsWith('v20')
```
nbbeeken, line 174:
```js
process.version.startsWith('v20')
```

nbbeeken (changes requested):

test/tools/runner/hooks/configuration.js
baileympearson: lol, oops

nbbeeken (approved):
Changes LGTM, CI pending

nbbeeken (approved):

PR #3479, opened 2022-11-29
fix(NODE-4447): disable causal consistency in implicit sessions

dariakp:
### Description
NODE-4447
#### What is changing?
- Disable causal consistency by default in implicit sessions to avoid clashing with read concern settings
- Sync spec tests
- New unit tests around the functionality

No

Fixes bug described in DRIVERS-2369 (retrying operations when read concern is `linearizable` or `available`)

### DIFF
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -169,8 +169,10 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
     this[kServerSession] = this.explicit ? this.sessionPool.acquire() : null;
     this[kTxnNumberIncrement] = 0;
 
+    const canEnableCausalConsistency = this.explicit && options.snapshot !== true;
     this.supports = {
-      causalConsistency: options.snapshot !== true && options.causalConsistency !== false
+      // if we can enable causal consistency, do so by default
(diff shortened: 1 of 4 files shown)


src/sessions.ts
```diff
@@ -169,8 +169,10 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
     this[kServerSession] = this.explicit ? this.sessionPool.acquire() : null;
     this[kTxnNumberIncrement] = 0;
 
+    const defaultCausalConsistencyValue = this.explicit && options.snapshot !== true;
     this.supports = {
```
nbbeeken, line 173: I caught myself up on the spec work and all, just double checking here there is no concern that the default is changing because the goal is to address a bug with the noted read concern settings? (I see the bullet in the description but cannot be to cautious)

nbbeeken (approved):
LGTM just a quick Q

src/sessions.ts
dariakp: the default is changing for the implicit sessions only; the default for the explicit sessions is the same as it has been (true for all except snapshot-enabled), and the unit tests are there to express the expected behavior

nbbeeken (approved):

PR #3482, opened 2022-12-06
fix(NODE-4834): ensure that MessageStream is destroyed when connections are destroyed

W-A-James:
#### What is changing?

Adding logic to ensure that a `Connection`'s `MessageStream` is always destroyed when the associated TCP socket is destroyed. Also added guard logic to ensure that `onTimeout`, `onClose`, and `onError` are nops if the `Connection` has already been destroyed.

Inline documentation was added to differentiate the `closed` and `destroyed` flags and to explain the conditions under which the `onTimeout`, `onClose`, and `onError` event handlers are called.

No

Prior to this change, a `Connection`'s `MessageStream` was not destroyed when the `Connection` was destroyed, which could lead to memory leaks. 


test/unit/cmap/connection.test.ts
```diff
@@ -433,6 +438,85 @@ describe('new Connection()', function () {
...
+      const messageStreamSymbol = getSymbolFrom(connection, 'messageStream');
+      messageStream = sinon.spy(connection[messageStreamSymbol]);
+    });
+
+    afterEach(() => {
+      timerSandbox.restore();
+      clock.restore();
+    });
+
+    it('should destroy message stream and socket', () => {
```
durran, line 478:
```suggestion
    it('destroys the message stream and socket', () => {
```

test/unit/cmap/connection.test.ts
```diff
@@ -433,6 +438,85 @@ describe('new Connection()', function () {
...
+      const messageStreamSymbol = getSymbolFrom(connection, 'messageStream');
+      messageStream = sinon.spy(connection[messageStreamSymbol]);
+    });
+
+    afterEach(() => {
+      timerSandbox.restore();
+      clock.restore();
+    });
+
+    it('should destroy message stream', () => {
```
durran, line 512:
```suggestion
    it('destroys the message stream', () => {
```

test/unit/cmap/connection.test.ts
```diff
@@ -486,4 +570,44 @@ describe('new Connection()', function () {
...
+    });
+
+    it('should end tcp socket and destroy messageStream', () => {
+      connection.destroy();
+      clock.tick(1);
+      expect(messageStream.destroy).to.have.been.calledOnce;
+      expect(driverSocket.end).to.have.been.calledOnce;
+    });
+
+    it('should not call stream.end after onClose, onTimeout, or onError', () => {
```
durran, line 603:
```suggestion
    it('does not call stream.end after onClose, onTimeout, or onError', () => {
```

src/cmap/connection.ts
```diff
@@ -170,7 +170,9 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   address: string;
   socketTimeoutMS: number;
   monitorCommands: boolean;
+  /** True when after of onClose, onError, or onTimeout was called */
   closed: boolean;
+  /** True only after destroy has been called */
```
baileympearson, line 175: these variables are states.  the documentation should focus not on what methods explicitly create that state (because the class internals change) but rather what the states are.

If we decide to update this documentation, I suggest something more along the lines of:
```typescript
/** Indicates that the connection (including the underlying socket) has been closed. */
closed: boolean
/** Indicates that the connection was explicitly destroyed. In the destroyed state, the connection is also closed. */
destroyed: boolean
```

src/cmap/connection.ts
```diff
@@ -468,6 +489,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this.removeAllListeners(Connection.PINNED);
     this.removeAllListeners(Connection.UNPINNED);
 
+    this[kMessageStream].destroy();
+
     options = Object.assign({ force: false }, options);
     if (this[kStream] == null || this.destroyed) {
       this.destroyed = true;
```
baileympearson, line 496: optional - let's clean up a bit!  `this[kStream]` is non-nullable ( see line 194 ).

We can remove the `null` check here (I'm surprised Typescript didn't complain..?), and then line 473 is unnecessary.

src/cmap/connection.ts
```diff
@@ -488,12 +511,19 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       return;
     }
 
-    this[kStream].end(() => {
+    if (this[kStream].destroyed) {
```
baileympearson, line 514: I'm curious what inspired this change.  Is this a guard against destroying a destroyed stream?  Or did a test failure necessitate this check?

src/cmap/connection.ts
```diff
@@ -468,6 +489,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this.removeAllListeners(Connection.PINNED);
     this.removeAllListeners(Connection.UNPINNED);
```
baileympearson, line 490: another cleanup thought - we can now use optional chaining on the callback

```typescript
  callback()
}
```

becomes 

```typescript
callback?.();
```

or an alternative is to ensure it's always defined at the top of the function:
```typescript
callback = callback ?? () => {};
```

src/cmap/connection.ts
```diff
@@ -309,11 +311,18 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this[kLastUseTime] = now();
```
baileympearson, line 311: we now have cleanup logic scattered across four different functions in the connection class.  Is there any reason we can't rely on `destroy({ force: true})` in all three helpers?

baileympearson (changes requested):

src/cmap/connection.ts
W-A-James: It was a bit of both. Particularly because the Node documentation specifies that after a call to `<readable|writable>.destroy`.

> subsequent calls to write() or end() will result in an ERR_STREAM_DESTROYED error

Also, our `FakeSocket` in the unit test file didn't have an `end` method so we weren't testing for this anywhere and once I added the test to check if the code path was possible, it did fail.

addaleax, line 514: Does it make sense to use [`.writableEnded`](https://nodejs.org/api/stream.html#writablewritableended) here? For writable streams there is a window of time where `.end()` has been called but `close` has not been emitted/`.destroyed` is still false.

nbbeeken: Since we always set this.destory to true, shouldn't that prevent us from reaching end after the socket has been destoryed?

src/cmap/connection.ts
nbbeeken: Yea and while we're at it let's use `return callback?.();` its a common pattern that makes it so we don't have to write `else` blocks, the callback should be the "end" of the function so returning at its invocation helps us keep that paradigm consistent.

nbbeeken (changes requested):

test/unit/cmap/connection.test.ts
```diff
@@ -34,6 +35,11 @@ class FakeSocket extends EventEmitter {
   }
   destroy() {
     // is called, has no side effects
+    this.writableEnded = true;
+  }
+  end(cb) {
+    // nextTick to simulate I/O delay
+    process.nextTick(cb);
```
nbbeeken, line 42: We should set writableEnd here too

src/cmap/connection.ts
```diff
@@ -170,7 +170,9 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   address: string;
   socketTimeoutMS: number;
   monitorCommands: boolean;
+  /** Indicates that the connection (including underlying TCP socket) has been closed. */
   closed: boolean;
+  /** Indicates that the connection has been explicitly destroyed. In the destroyed state, the connection is also closed */
   destroyed: boolean;
```
addaleax, line 176: Not a request for changes or anything, but I don’t quite understand how we’re semantically making a difference between `.closed` and `.destroyed`. I understand that `.destroyed` is the result of an explicit call to `.destroy()`, but in which case would we check for that instead of using `.closed`?

Combining this with @baileympearson’s comment immediately below, it seems to me a bit like ideally we would:
- Always call `this.destroy()` in the places where we destroy _either_ `this[kStream]` or `this[kMessageStream]` in order to clean up whatever has not been cleaned up yet
- Drop the `.destroyed` property entirely
- Maybe go even further and just make `.closed` a getter that returns `this[kStream].destroyed && this[kMessageStream].destroyed` or similar, to avoid having to track separate state?

src/cmap/connection.ts
addaleax: > let's use `return callback?.();` its a common pattern

Keep in mind that that pattern means that `callback()` can be called either synchronously or asynchronously, which is a bit of anti-pattern in Node.js, since callers would need to be prepared to handle either case

baileympearson (changes requested):

src/cmap/connection.ts
```diff
@@ -459,41 +452,25 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
-    this[kStream].end(() => {
-      this.destroyed = true;
-      if (typeof callback === 'function') {
-        callback();
-      }
-    });
+    if (!this[kStream].writableEnded) {
+      this[kStream].end(() => {
+        callback?.();
+      });
```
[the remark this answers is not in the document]
W-A-James: The close event emitted by destroy is now a nop as we set `this.closed` to true synchronously near the start of `destroy`.

As far as the callback we provide to `kStream.end`, I just pushed another change that makes it so every path through `destroy` either runs a callback asynchronously through `kStream.end` or through `process.nextTick`

src/cmap/connection.ts
```diff
@@ -459,41 +452,28 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
 
     if (options.force) {
       this[kStream].destroy();
-      this.destroyed = true;
-      if (typeof callback === 'function') {
-        callback();
-      }
-
-      return;
+      return process.nextTick(callback);
```
nbbeeken, line 469: Let's gate calling this on the existence of the callback, just so we don't need to make a new empty arrow function.

src/sdam/topology.ts
```diff
@@ -496,14 +496,14 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
-    options = options ?? {};
+    options = options ?? { force: false };
 
     if (this.s.state === STATE_CLOSED || this.s.state === STATE_CLOSING) {
       return callback?.();
     }
 
     const destroyedServers = Array.from(this.s.servers.values(), server => {
-      return promisify(destroyServer)(server, this, options as CloseOptions);
+      return promisify(destroyServer)(server, this, Object.assign({ force: false }, options));
```
nbbeeken, line 506: We can again remove the override of the reordering of arguments, we only call this from the mongoclient with either options and no callback or options and callback, so the argument fixing to support callback with no options can go away, that will make the typescript happy on this line if you write:

```ts
return promisify(destroyServer)(server, this, { force: !!options?.force });
```

src/cmap/connection.ts
```diff
@@ -149,7 +149,7 @@ export interface ConnectionOptions
 /** @public */
 export interface DestroyOptions {
   /** Force the destruction. */
-  force?: boolean;
+  force: boolean;
```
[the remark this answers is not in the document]
nbbeeken: @W-A-James Despite this and `CloseOptions` being marked `@public` they are not used on any publicly accessible APIs (connection.destroy and topology.close respectively). Normally, changing a public API wouldn't be viable in a fix PR like this but since these are not useful to users (they can't even be used as a type annotation for an input to one of our methods) I think this is an acceptable change to make here. 

Just sharing our thinking for other reviewers sake as well

nbbeeken (changes requested):

nbbeeken (dismissed):

nbbeeken (approved):

baileympearson (approved):
Two small test comments but non-blocking.

test/integration/node-specific/topology.test.js
```diff
@@ -10,12 +10,20 @@ describe('Topology', function () {
...
+        }
+        topology.close({}, err => {
+          try {
+            expect(err).to.not.exist;
+            expect(topology.isDestroyed()).to.be.true;
+            expect(states).to.eql(['connecting', 'connected', 'closing', 'closed']);
+            done();
+          } catch (error) {
+            done(error);
+          }
```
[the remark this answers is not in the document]
nbbeeken: Since toplogy.close/connect are internal we're likely to async-ify them in the driver which would cascade to fixing tests that are directly testing these internal methods (instead of via the client). That will cause this test to need re-modification later so possibly not worth changing? 
anyway, happy to update to async/await if we want that or just revert the change that was made here, this is just the classic problem of test times out instead of reporting the real error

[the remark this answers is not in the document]
W-A-James: Agreed, the change here was a stopgap to fix the control flow of this test, but this can definitely be improved.

nbbeeken (changes requested):
nearly lgtm `to.not.have.been.calledOnce` cleanups

test/unit/cmap/connection.test.ts
```diff
@@ -486,4 +593,58 @@ describe('new Connection()', function () {
...
+      connection.destroy({ force: false });
+      connection.destroy({ force: false });
+      connection.destroy({ force: false });
+      connection.destroy({ force: false });
+      clock.tick(1);
+      expect(driverSocket.destroy).to.not.have.been.called;
+      expect(driverSocket.end).to.have.been.calledOnce;
+    });
+
+    it('does not call stream.end if options.force == true', () => {
```
dariakp, line 636: While it's nice to know what `options.force == true` doesn't do, it would be even more useful to know what it does do.

Can we split up the cases outlined here into two context blocks, one with force set to true and another with force set to false and ensure that the test cases fully capture the intended behavior for each option setting?

dariakp (changes requested):

dariakp (changes requested):
@W-A-James can you also include the other two `destroy()` cases in the force true/false blocks? because currently you use `force: false` in both of them, and it's relevant to know what happens in both cases. More precisely, you could take this case:
`ends the tcp socket and destroys the messageStream` and make it a separate `ends the tcp socket` `it` block under each `true/false` block, and then take `calls stream.end exactly once when destroy is called multiple times` and move it under `false`, then make an analogous case under `true`

dariakp (approved):
Great job on the tests, thanks!

nbbeeken (approved):

PR #3483, opened 2022-12-08
ci(NODE-4872): add permissions to dependency check workflow

sashashura:
### Description

This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

#### What is changing?

Explicit permissions added

No, CI only

See the description

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -7,7 +7,7 @@ on:
     branches: [ "main" ]
 
 permissions:
-  contents: read #  to fetch code (actions/checkout)
+  contents: read
 
 jobs:
   build:
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -5,6 +5,10 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
+
+permissions:
+  contents: read #  to fetch code (actions/checkout)
+
 jobs:
   build:
     runs-on: ubuntu-latest


.github/workflows/dependencies.yml
```diff
@@ -5,6 +5,10 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
+
+permissions:
+  contents: read #  to fetch code (actions/checkout)
```
baileympearson, line 10:
```suggestion
  contents: read
```

baileympearson (changes requested):
Hey @sashashura - thanks for the contribution!  I'm going to remove the comment added in the workflow file but otherwise it looks good to me.

PR #3490, opened 2022-12-19
feat(NODE-4867)!: adopt BSON v5

nbbeeken:
#### What is changing?

Pulls in BSON v5 alpha.3, we're good to merge that version to main, we have a follow up ticket to make it 5.0.0 before releasing driver 5.0.0.

- Add a few of buffer type agnostic helpers
- Fix type issues from the new BSON version
- **Omit** useBigInt64 from our supported options, this is followup work
- ❗ Adds `BSON` the namespace to our public API, which makes `BSON.EJSON` available as well

BSON v5 is really cool


src/cmap/auth/mongodb_aws.ts
```diff
@@ -108,7 +108,7 @@ export class MongoDBAWS extends AuthProvider {
           return;
         }
 
-        if (serverNonce.compare(nonce, 0, nonce.length, 0, nonce.length) !== 0) {
+        if (!ByteUtils.equals(serverNonce.subarray(0, nonce.byteLength), nonce)) {
```
nbbeeken, line 111: This is interesting, I'm preserving what was here, but by using our local nonce length we are permitting the server to send us a nonce that's longer than the one we have in memory locally.

test/integration/crud/insert.test.js
```diff
@@ -4,7 +4,6 @@ const { format: f } = require('util');
 const { expect } = require('chai');
 
 const Script = require('vm');
-const { normalizedFunctionString } = require('bson/lib/parser/utils');
```
dariakp, line 7: what's the deal with this function and the corresponding tests?

test/integration/objectid.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    const objectId = new ObjectId();
+    expect(objectId.toString()).to.have.lengthOf(24);
+  });
+
+  it('ObjectId toJSON returns 24 character string', function () {
+    const objectId = new ObjectId();
+    expect(objectId.toJSON()).to.have.lengthOf(24);
+  });
+
+  it('shouldCorrectlyCreateOIDNotUsingObjectId', async function () {
```
dariakp, line 49: can we fix this title?

test/integration/objectid.test.ts
```diff
@@ -0,0 +1,142 @@
+import { expect } from 'chai';
+
+import { Collection, Db, MongoClient, ObjectId } from '../../src';
+import { sleep } from '../tools/utils';
+
+describe('ObjectId', function () {
```
dariakp, line 6: this entire test file is kind of a trainwreck with a mix of unit tests and setup/cleanup not handled appropriately - I understand that it's out of scope for this work, but please file a technical debt follow up to fix this specific test file

test/tools/spec-runner/matcher.js
```diff
@@ -34,6 +34,16 @@ function generateMatchAndDiffSpecialCase(key, expectedObj, actualObj, metadata)
       };
     }
 
+    if (actualObj._bsontype === 'Code' && key === 'scope') {
```
dariakp, line 37: why has that not come up before?

test/unit/bson.test.ts
```diff
@@ -1,22 +1,21 @@
...
-const BSON = require('../mongodb');
+import { BSON } from '../mongodb';
 
 describe('When importing BSON', function () {
   const types = [
     ['Long', 23],
     ['ObjectId', '123456789123456789123456'],
     ['Binary', Buffer.from('abc', 'ascii')],
-    ['Timestamp', 23],
+    ['Timestamp', 23n],
```
dariakp, line 10: is this an expected change in behavior? if it's a side effect of the bigint work, feels like something we should intentionally handle in the driver

dariakp (changes requested):

src/cmap/auth/mongodb_aws.ts
nbbeeken: I'm not sure it's possible within the confines of the auth protocol but if we send "abcd" and the server replies "abcdxyz" we won't enter this error handling because we only checked the section of the server response that was the length of what we sent to the server. But that is exactly the logic that was here before my change:

https://nodejs.org/dist/latest-v18.x/docs/api/buffer.html#bufcomparetarget-targetstart-targetend-sourcestart-sourceend

test/integration/client-side-encryption/client_side_encryption.spec.test.ts
```diff
@@ -35,14 +35,6 @@ const skippedAuthTests = [
...
   'updateOne with deterministic encryption',
   'updateMany with deterministic encryption',
-  'type=date',
-  'type=regex',
-  'type=timestamp',
-  'type=javascript',
-  'type=binData',
-  'type=int',
-  'type=objectId',
-  'type=symbol',
```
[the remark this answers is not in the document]
nbbeeken: One of them failed (`type=javascript`) because our legacy test runner was strict about properties not existing as opposed to nullish. Code now always defines the property scope, but it is set to null when there is no scope, so that's the small test adjustment with the special case for 'Code'

test/integration/crud/insert.test.js
nbbeeken: We used to "normalize" function strings in 4.x BSON (remove a space after the keyword `function`) we no longer do that so we don't need this helper, plus that import doesn't exist anymore either, we only have a single bundle file now

test/tools/spec-runner/matcher.js
nbbeeken: See comment about `type=*` tests. Code has a `scope: Document | null` property now instead of `scope?: Document`. The prop always exists, its explicitly set to null when there is no scope.

test/unit/bson.test.ts
nbbeeken: Not a side effect of bigint work, I can also change this to `{ i: number, t: number }` the Timestamp constructor got an overhaul and single number arguments are no longer permitted to avoid ambiguity

test/integration/objectid.test.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-4989
and left a todo

test/integration/objectid.test.ts
nbbeeken: camel case all the things! (fixed)

src/cmap/auth/mongodb_aws.ts
dariakp: should we file a follow up to fix this?

dariakp (approved):

src/cmap/auth/mongodb_aws.ts
nbbeeken: filed!

dariakp (approved):

PR #3495, opened 2022-12-21
refactor(NODE-4906): remove internal callback usage of public APIs

nbbeeken:
#### What is changing?

Remove internal callback usage of public methods. In order to proceed with the typescript changes to callback support it makes sense for us to correct these few places where we use callbacks that will become type errors in NODE-4522.

This is the same kind of refactor for GridFS in #3493 

Typescript compat with public API.


src/cursor/abstract_cursor.ts
```diff
@@ -841,7 +841,10 @@ function cleanupCursor(
 
     if (session) {
       if (session.owner === cursor) {
-        return session.endSession({ error }, callback);
+        session.endSession({ error }).finally(() => {
+          callback();
+        });
+        return;
```
durran, line 847: Note from chat these returns getting moved down were to satisfy the `void` return type in TS.

src/cursor/abstract_cursor.ts
```diff
@@ -855,10 +858,11 @@ function cleanupCursor(
   function completeCleanup() {
     if (session) {
       if (session.owner === cursor) {
-        return session.endSession({ error }, () => {
+        session.endSession({ error }).finally(() => {
```
durran, line 861: The `finally` here is so we ensure the callback is always called.

durran (approved):

PR #3499, opened 2022-12-21
feat(NODE-4522)!: remove callback support

nbbeeken:
#### What is changing?

> __Warning__ **ATTN** Reviewers

- There's no check / typescript warning that the async keyword has been used. I think we want that to consistently be in place, so take a close look at signatures.

Promises provide a consistent way to handle the outcome of an asynchronous operation, unifying the API here to always use the `async` syntax forces the code to always return a promise so that errors/resolutions are always handled the same way.


src/operations/validate_collection.ts
```diff
@@ -18,8 +18,9 @@ export class ValidateCollectionOperation extends CommandOperation<Document> {
   collectionName: string;
   command: Document;
 
-  constructor(admin: Admin, collectionName: string, options: ValidateCollectionOptions) {
```
nbbeeken, line 21: Removing the overloads in admin.ts validateCollection resulted in this change b/c the argument fixing logic is gone, we can do this in the helper or here, let me know which is best

src/db.ts
```diff
@@ -226,35 +225,15 @@ export class Db {
    *
    * @param name - The name of the collection to create
    * @param options - Optional settings for the command
-   * @param callback - An optional callback, a Promise will be returned if none is provided
    */
-  createCollection<TSchema extends Document = Document>(
+  async createCollection<TSchema extends Document = Document>(
```
nbbeeken, line 229: I hope leaving a comment here will make github not collapse this file, easy to overlook

src/admin.ts
```diff
@@ -57,111 +56,48 @@ export class Admin {
...
-  command(command: Document, options: RunCommandOptions, callback: Callback<Document>): void;
-  command(
-    command: Document,
-    options?: RunCommandOptions | Callback<Document>,
-    callback?: Callback<Document>
-  ): Promise<Document> | void {
-    if (typeof options === 'function') (callback = options), (options = {});
-    options = Object.assign({ dbName: 'admin' }, options);
-
+  async command(command: Document, options?: RunCommandOptions): Promise<Document> {
```
nbbeeken, line 60: diff collapse mitigation

src/bulk/common.ts
```diff
@@ -1175,56 +1174,36 @@ export abstract class BulkOperationBase {
...
-        ? options
-        : undefined;
-    return maybeCallback(async () => {
-      options = options != null && typeof options !== 'function' ? options : {};
-
-      if (this.s.executed) {
-        throw new MongoBatchReExecutionError();
-      }
+  async execute(options?: BulkWriteOptions): Promise<BulkWriteResult> {
+    options ??= {};
```
nbbeeken, line 1178: is it better to use default arg or coalesce assignment 🤔

src/admin.ts
```diff
@@ -57,111 +56,48 @@ export class Admin {
...
     return executeOperation(
       this.s.db.s.client,
-      new RunCommandOperation(this.s.db, command, options),
-      callback
+      new RunCommandOperation(this.s.db, command, { dbName: 'admin', ...options })
```
dariakp, line 63: this isn't any different than the previous behavior, but since we aren't guarding against non-object options, all our APIs will have some odd behavior if passed a string here, which is different from any api that passes options directly to the RunCommandOperation which would fail at resolveBSONOptions (based on code analysis); something that we may want to standardize/improve at some point anyway

src/admin.ts
```diff
@@ -170,49 +106,22 @@ export class Admin {
...
-      if (typeof options === 'function') (callback = options), (options = {});
-    }
-
-    options = Object.assign({ dbName: 'admin' }, options);
-
+    password?: string | AddUserOptions,
+    options?: AddUserOptions
+  ): Promise<Document> {
+    options =
+      options != null && typeof options === 'object'
```
dariakp, line 116: we removed the overload that documents this, so until we are ready to break the behavior, we should keep those overloads

src/bulk/common.ts
dariakp: default arg is cleaner imho, but it honestly doesn't matter as long as we're consistent

src/collection.ts
```diff
@@ -254,32 +253,15 @@ export class Collection<TSchema extends Document = Document> {
...
-    doc: OptionalUnlessRequiredId<TSchema>,
-    options?: InsertOneOptions | Callback<InsertOneResult<TSchema>>,
-    callback?: Callback<InsertOneResult<TSchema>>
-  ): Promise<InsertOneResult<TSchema>> | void {
-    if (typeof options === 'function') {
-      callback = options;
-      options = {};
+    options?: InsertOneOptions
+  ): Promise<InsertOneResult<TSchema>> {
+    // TODO(NODE-4751): versions of mongodb-client-encryption before v1.2.6 pass in hardcoded { w: 'majority' }
```
dariakp, line 261: NODE-4751 is done

src/cursor/find_cursor.ts
```diff
@@ -101,7 +101,8 @@ export class FindCursor<TSchema = any> extends AbstractCursor<TSchema> {
         limit && limit > 0 && numReturned + batchSize > limit ? limit - numReturned : batchSize;
 
       if (batchSize <= 0) {
-        return this.close(callback);
+        // @ts-expect-error: TODO...
```
dariakp, line 104: TODO

dariakp (changes requested):
review on first half

src/cursor/abstract_cursor.ts
```diff
@@ -903,7 +866,7 @@ class ReadableCursorStream extends Readable {
   }
 
   override _destroy(error: Error | null, callback: (error?: Error | null) => void): void {
-    this._cursor.close(err => process.nextTick(callback, err || error));
+    this._cursor.close().finally(() => callback(error));
```
dariakp, line 869: I think close can still error, no? in that case, we can still callback here with that error instead of ignoring it `.close().then(() => callback(error), err => callback(err))`, or is that not a concern?

baileympearson (changes requested):
mostly minor comments here

src/db.ts
dariakp: it still collapses it, but that's fine if we use the github checkmarks for viewed files

src/db.ts
```diff
@@ -568,47 +410,22 @@ export class Db {
...
-      }
-    } else {
-      if (typeof options === 'function') (callback = options), (options = {});
-    }
-
+    password?: string | AddUserOptions,
+    options?: AddUserOptions
+  ): Promise<Document> {
+    options =
+      options != null && typeof options === 'object'
```
dariakp, line 420: same comment here about overloads as in the admin file

src/admin.ts
nbbeeken: We can still drop the overload, and update the signature. Overloads are generally more useful when the return type will change based on a change in input, the DX of a string unioned with options object should be equivalent to what was here before.

Happy to bring it back tho, just my 2c

test/integration/crud/find_cursor_methods.test.js
```diff
@@ -204,7 +204,8 @@ describe('Find Cursor', function () {
     });
   });
 
-  context('#clone', function () {
+  // TODO(NODE-4988): Need to fix cursor.clone in mongodb-legacy
+  context.skip('#clone', function () {
```
dariakp, line 208: since it's just the two tests, could we fix just fix the tests and not have the skip?

test/integration/crud/misc_cursors.test.js
```diff
@@ -1671,7 +1671,8 @@ describe('Cursor', function () {
     }
   });
 
-  it('removes session wheen cloning a find cursor', function (done) {
+  // TODO(NODE-4988): Need to fix cursor.clone in mongodb-legacy
+  it.skip('removes session wheen cloning a find cursor', function (done) {
```
dariakp, line 1675: just two tests here

dariakp (changes requested):

src/admin.ts
nbbeeken: Agreed, and something like verify something is an object is cheap, but taking it much futher could have diminishing returns. As long as there aren't known silent failures (unless there's runtime typeof checking) generally it feels fair to lean on our documentation / TS types to communicate correctness, and failures from breaking the TS "contract" are expected behavior.

As for this spot, we could do a little ternary `...(typeof options === 'object' ? options : {})` thoughts?

src/gridfs/download.ts
```diff
@@ -182,30 +182,22 @@ export class GridFSBucketReadStream extends Readable implements NodeJS.ReadableS
    * Marks this stream as aborted (will never push another `data` event)
    * and kills the underlying cursor. Will emit the 'end' event, and then
    * the 'close' event once the cursor is successfully killed.
-   *
-   * @param callback - called when the cursor is successfully closed or an error occurred.
    */
-  abort(callback?: Callback<void>): void {
+  async abort(): Promise<void> {
```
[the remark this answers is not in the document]
dariakp: yes but let's add this to whatever deprecation task we do for the changes in this PR:
https://github.com/mongodb/node-mongodb-native/pull/3518

src/collection.ts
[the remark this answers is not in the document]
nbbeeken: fixed

nbbeeken: fixed

src/cursor/find_cursor.ts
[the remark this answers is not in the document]
nbbeeken: Oh this was silly, .finally returns a promise, the function still returns void, just have to move the `return`

nbbeeken: fixed

src/gridfs/download.ts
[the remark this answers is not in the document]
nbbeeken: I'll be sure to add a call out in the migration guide update that is yet to be written, leaving this open.

src/admin.ts
dariakp: I don't really care about this for this PR or v5 in general; validation on high traffic paths is not a good idea, just something to keep in mind, or add to the tech chat agenda for future improvement - making the failure more consistent in the case of bad options, preferably by creating a suitable place for it to happen "naturally"

src/cursor/abstract_cursor.ts
nbbeeken: close should never throw so this should be fine, inside the cursor.close method the callback is never invoked with an error

test/integration/crud/misc_cursors.test.js
nbbeeken: fixed up.. these weren't actually checking for what the title says I think 🤔

test/integration/crud/find_cursor_methods.test.js
nbbeeken: `async` -ified and fixed up 👍

src/bulk/common.ts
nbbeeken: Made this consistent in the few places it wasn't

src/admin.ts
nbbeeken: added a bullet to tech chat

nbbeeken:
our _driver_ CI is "green", however the FLE integration tests that we pull in from the other repo have callback usage in the tests: https://github.com/mongodb/libmongocrypt/pull/539

src/cursor/abstract_cursor.ts
```diff
@@ -903,7 +866,11 @@ class ReadableCursorStream extends Readable {
   }
 
   override _destroy(error: Error | null, callback: (error?: Error | null) => void): void {
-    this._cursor.close(err => process.nextTick(callback, err || error));
+    // FIXME: how is this possible?
+    const res = this._cursor.close();
+    if (res != null) {
+      res.finally(() => callback(error));
+    }
```
nbbeeken, line 873: Still debugging this ([evergreen error link](https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_next_rhel80_large_gallium_test_5.0_replica_set_patch_056c86f6dee36f805376513c8322ac1ecab3c8e6_63cc20c7d1fe07422f90da68_23_01_21_17_28_40/0?type=T#L5791)). Any ideas?

baileympearson: which evergreen run was this from?  I did an (admittedly brief) skim of the evergreen patch and didn't see it


I did see that serverless is particularly bad, but I'm not sure when that started

nbbeeken: from `rhel80_large_gallium_test_5.0_replica_set` in [this patch](https://spruce.mongodb.com/task/mongo_node_driver_next_rhel80_large_gallium_test_5.0_replica_set_patch_056c86f6dee36f805376513c8322ac1ecab3c8e6_63cc20c7d1fe07422f90da68_23_01_21_17_28_40/tests?execution=0&sortBy=STATUS&sortDir=ASC)

But it's not the patch I'm referring to, the AbstractCursor.close method is an `async function` and yet it returns `undefined` the second time it is invoked, I have to gate calling finally on a nullish check. 🤔 super strange

nbbeeken: It was `sinon` 🤦  proxies aren't obvious until you step-into the function call

src/cursor/abstract_cursor.ts
baileympearson: If we're going to rely on the assumption that .close() never rejects, we should enforce that via testing.  The problem is that we cannot.  How do you write a test that says "this method never throws"?  What conditions are you setting up in that test to make it fail if it does throw at some point in the future?  It's impossible.  

Using callbacks we could enforce it by simply making the callback `(value) => { .. } ` and not taking an error.  Using promises it is not possible.

I agree with Daria.  Because we can't encode this assumption into tests to enforce it for the future, we can't rely on it here and we should handle the error case.

nbbeeken: fixed

nbbeeken:
CI run with the commit hash of the changes in this PR: https://github.com/mongodb/libmongocrypt/pull/539

https://spruce.mongodb.com/version/63ceb9470ae6060436a4eff0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

dariakp (changes requested):

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,9 +16,56 @@ The following is a detailed collection of the changes in the major v5 release of
 
 ## Changes
```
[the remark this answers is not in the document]
nbbeeken: It's just the one that was missing a deprecation but now we're sure we're going back to mark it, so I think its fine to leave out here. It's still covered by "all asynchronous apis"

test/integration/crud/find_cursor_methods.test.js
```diff
@@ -205,48 +205,34 @@ describe('Find Cursor', function () {
   });
 
   context('#clone', function () {
-    it('should clone a find cursor', function (done) {
+    it('should clone a find cursor', async function () {
+      // TODO(NODE-4988): Need to fix cursor.clone in mongodb-legacy, callbacks do not work on cloned cursor
```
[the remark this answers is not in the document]
nbbeeken: done

test/integration/crud/misc_cursors.test.js
```diff
@@ -1671,64 +1671,28 @@ describe('Cursor', function () {
...
-  it('removes session wheen cloning a find cursor', function (done) {
-    const configuration = this.configuration;
-    client.connect((err, client) => {
-      expect(err).to.not.exist;
-
-      const db = client.db(configuration.db);
-      db.createCollection('clone_find_cursor_session', (err, collection) => {
-        expect(err).to.not.exist;
+  it('removes session when cloning an find cursor', async function () {
+    // TODO(NODE-4988): Need to fix cursor.clone in mongodb-legacy, callbacks do not work on cloned cursor
```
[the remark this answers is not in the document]
nbbeeken: done

etc/notes/CHANGES_5.0.0.md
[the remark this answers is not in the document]
dariakp: marking it isn't 100% straightforward because it didn't previously have an overload, but we can just make it match the upload one

nbbeeken:
FLE run pointing at my PR https://github.com/mongodb/libmongocrypt/pull/539
https://spruce.mongodb.com/version/63ceedf1c9ec440caadea3cb/tasks Passing 🟢

dariakp (dismissed):

PR #3500, opened 2022-12-21
feat(NODE-4684)!: remove collection insert, update, remove methods

nbbeeken:
#### What is changing?

Removes the legacy crud methods.

Modernize code base, remove duplicate code.

TODO:
- [x] Migration guide
- [x] Feedback on shimming old methods
---


nbbeeken:
I collected the [usage counts of the legacy methods in this patch](https://spruce.mongodb.com/version/63b48ec0a4cf471037033870/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). 

4.0-replica_set reports: `{ insert: 326, update: 19, remove: 9 }` so I can probably look for the removes and maybe the updates, but the inserts are many 😅

nbbeeken:
From slack: Looking into search and replace solutions, there wasn't an approach that would make a significant enough impact in the number of insert usages. We'll defer the work to clean this up to a later time

test/integration/crud/crud_api.test.ts
```diff
@@ -655,71 +655,6 @@ describe('CRUD API', function () {
     }
   });
 
-  it('should correctly execute remove methods using crud api', {
```
nbbeeken, line 658: Wasn't using .remove even and is covered by CRUD spec tests

test/integration/crud/find.test.js
```diff
@@ -1,13 +1,18 @@
...
+const { assert: test } = require('../shared');
 const { expect } = require('chai');
 const sinon = require('sinon');
 const { setTimeout } = require('timers');
 const { Code, ObjectId, Long, Binary, ReturnDocument } = require('../../../src');
 
 describe('Find', function () {
-  before(function () {
-    return setupDatabase(this.configuration);
+  let client;
```
nbbeeken, line 9: introducing mocha hooks cus there was a close leak, classic callback didn't reach the end stuff

test/integration/crud/find.test.js
```diff
@@ -926,44 +943,41 @@ describe('Find', function () {
...
-                title: 'Tobi',
-                author: 'Brian',
-                newTitle: 'Woot',
-                meta: { visitors: 0 }
-              },
-              configuration.writeConcernMax(),
-              function (err, r) {
-                // Fetch the id
-                var id = r.insertedIds[0];
+      var db = client.db(configuration.db);
```
nbbeeken, line 946: leak fix

test/integration/crud/find.test.js
```diff
@@ -1458,58 +1472,55 @@ describe('Find', function () {
...
-        transactions.push(transaction);
-        // Wrapping object
-        var wrapingObject = {
-          funds: {
-            remaining: 100.5
-          },
-
-          transactions: transactions
-        };
+      var db = client.db(configuration.db);
```
nbbeeken, line 1475: leak, etc. I'll stop commenting each one

test/integration/node-specific/auto_connect.test.ts
```diff
@@ -483,15 +483,6 @@ describe('When executing an operation for the first time', () => {
       });
     });
 
-    describe(`#insert()`, () => {
```
nbbeeken, line 486: Don't need to check auto connect on APIs that don't exist

test/integration/node-specific/operation_examples.test.ts
```diff
@@ -1802,77 +1802,6 @@ describe('Operations', function () {
     }
   });
 
-  /**
-   * Example of using keepGoing to allow batch insert using a Promise to complete even when there are illegal documents in the batch
-   *
-   * example-class Collection
-   * example-method insert
-   */
-  it('Should correctly execute insert with keepGoing option on mongod >= 1.9.1 With Promises', {
```
nbbeeken, line 1811: keepGoing removed

test/integration/objectid.test.js
```diff
@@ -1,92 +1,43 @@
...
-        expect(id.toHexString().length).to.equal(24);
-
-        // Locate the first document inserted
-        collection.findOne(id, function (err, document) {
-          expect(err).to.not.exist;
-          expect(id.toHexString()).to.equal(document._id.toHexString());
-          number_of_tests_done++;
-        });
-      });
+  it('generates new ObjectId for documents without _id property', async function () {
```
nbbeeken, line 17: It failed at first b/c of the insert usage, but then I also cleaned it up cus the setInterval logic was too interesting

test/integration/crud/remove.test.js
```diff
@@ -107,6 +103,21 @@ describe('Remove', function () {
     }
   });
 
+  it('deleteOne', async () => {
```
baileympearson, line 106: it's not clear from the description or the assertions what is under test here.  Also, why did we decide to add this test?  We have unified tests that cover all the assertions here (although implicitly)

test/integration/crud/find.test.js
baileympearson: Why did these leaks only show up now?  We didn't change any codepaths so I would expect no change in test behavior.

etc/notes/CHANGES_5.0.0.md
```diff
@@ -124,3 +124,24 @@ cursor.closed // true
...
+```ts
+// Single document insert:
+await collection.insert({ name: 'spot' });
+// Migration:
+await collection.insertMany([{ name: 'spot' }]);
+
+// Multi-document insert:
+await collection.insert([{ name: 'fido' }, { name: 'luna' }])
+// Migration:
+await collection.insertMany([{ name: 'fido' }, { name: 'luna' }])
```
baileympearson, line 141: just a thought - a table that has columns for legacy method / existing crud replacement might be an easier format for users to understand.

baileympearson (changes requested):

test/integration/collection-management/collection.test.ts
```diff
@@ -235,7 +235,7 @@ describe('Collection', function () {
           expect(err).to.not.exist;
           collection.insertOne({ c: 1 }, configuration.writeConcernMax(), err => {
             expect(err).to.not.exist;
-            collection.remove({ a: 1 }, configuration.writeConcernMax(), err => {
+            collection.deleteMany({ a: 1 }, configuration.writeConcernMax(), err => {
```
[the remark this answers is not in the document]
nbbeeken: No I don't think so, CRUD spec and generally we've got [coverage here](https://mciuploads.s3.amazonaws.com/mongo-node-driver/83bce00e28134b62f53b96a96a3d2fbaa2832a92/63b8715030661560b4b2996b/coverage/src/collection.ts.html). Removing

test/integration/crud/remove.test.js
nbbeeken: Agreed, [spec tests](https://github.com/mongodb/node-mongodb-native/blob/NODE-4684-rm-coll-insert-update-remove/test/spec/crud/unified/deleteMany-comment.json#L68) do check the "deletedCount" so no need for a dupe here.

test/tools/runner/hooks/legacy.ts
```diff
@@ -0,0 +1,34 @@
+import { Collection } from '../../../../src';
```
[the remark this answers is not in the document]
nbbeeken: done

etc/notes/CHANGES_5.0.0.md
nbbeeken: done!

baileympearson: Ah, I was thinking something along the lines of 

- replacing 129-130 with something like "Three legacy crud helpers on the collection class have been removed:"
- removing the typescript code example
- putting code examples of before / after in the table, so users clearly see a mapping between the "old code" and "new code"

test/integration/crud/find.test.js
```diff
@@ -786,83 +791,96 @@ describe('Find', function () {
...
-            // Let's modify the document in place
-            collection.findOneAndUpdate(
-              { a: 1 },
-              { $set: { b: 3 } },
-              { returnDocument: ReturnDocument.AFTER },
-              function (err, updated_doc) {
-                test.equal(1, updated_doc.value.a);
-                test.equal(3, updated_doc.value.b);
+            // Test return new document on change
+            collection.insert({ a: 1, b: 2 }, configuration.writeConcernMax(), function (err) {
```
W-A-James, line 802: Since we're trying to get rid of instances of `collection.insert`, why don't we get rid of these instances when updating old tests as part of our drive-by cleanup?

W-A-James (commented):
Apart from the question I had about replacing instances of `colllection.insert` in tests that are already being modified, lgtm.

test/integration/crud/find.test.js
nbbeeken: The change to this test was caused by unindenting the connect callback to fix a leak that showed up when it was erroring. There's a lot more cleanup that could be done here, remove callbacks, remove the old topology filtering, remove the combination of operations (dropColl, createColl, insert should all not be included in a test focused on findOneAndUpdate) and even fix the test name from being all camelcase. But there's only so much drive-by you can do before you're overloading the original task. 

Another reason I leave all of above mentioned legacy practices behind is that it earmarks the test for those of us who've been working on the code for awhile. When this test causes the next engineer some trouble, they'll know it isn't a recent addition by all these other outstanding issues with it. My drive-by goal was to address the leak issue, the next one may be to address any of the above issues when it causes trouble again.

etc/notes/CHANGES_5.0.0.md
nbbeeken: I had code samples when I first wrote the table but it isn't easy to read b/c of how wide it gets, it doesn't get the same scroll effect normal code blocks do. Let me fix up the text part and maybe move the code sample below the table

nbbeeken: Pushed the change, also the bracket are bit hard to notice, "arrayOfDocuments" is obviously just a variable name that implies something. Maybe I should use type annotations instead?

etc/notes/CHANGES_5.0.0.md
```diff
@@ -124,3 +124,33 @@ cursor.closed // true
...
+Three legacy operation helpers on the collection class have been removed:
+
+| Removed API                                    | API to migrate to                                  |
+|------------------------------------------------|----------------------------------------------------|
+| `insert(document)`                             | `insertOne(document)`                              |
+| `insert(arrayOfDocuments)`                     | `insertMany(arrayOfDocuments)`                     |
+| `update(filter)`                               | `updateMany(filter)`                               |
+| `remove(filter)`                               | `deleteMany(filter)`                               |
+
+The `insert` method could accept arrays or single documents, instead for single document inserts `insertOne` or an `insertMany` call with an array argument of one should be used.
```
baileympearson, line 139:
```suggestion
The `insert` method accepted an array of documents for multi-document inserts and a single document for single document inserts.  `insertOne` should now be used for single-document inserts and `insertMany` should be used for multi-document inserts.
```

baileympearson (approved):

PR #3502, opened 2022-12-22
feat(NODE-4598)!: close cursor on early loop break

durran:
### Description

Ensures cursor is closed when exiting the async generator.

#### What is changing?

Adds a finally clause in the async generator to close the cursor (if it's not already closed). This handles early exits from `for await of...` loops. A new integration test also is added to check this scenario.

None

NODE-4598

### DIFF
--- a/test/integration/node-specific/cursor_async_iterator.test.js
+++ b/test/integration/node-specific/cursor_async_iterator.test.js
@@ -79,5 +79,16 @@ describe('Cursor Async Iterator Tests', function () {
 
       expect(count).to.equal(1);
     });
+
+    it('cleans up cursor when breaking out of for await of loops', async function () {
+      const cursor = collection.find();
+
+      for await (const doc of cursor) {
+        expect(doc).to.exist;
+        break;
+      }
+
+      expect(cursor.closed).to.be.true;
+    });
   });
 });
--- a/test/integration/node-specific/cursor_async_iterator.test.js
+++ b/test/integration/node-specific/cursor_async_iterator.test.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { expect } = require('chai');
+const sinon = require('sinon');
 
 describe('Cursor Async Iterator Tests', function () {
   context('default promise library', function () {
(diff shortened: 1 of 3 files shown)


test/integration/node-specific/cursor_async_iterator.test.js
```diff
@@ -79,5 +79,16 @@ describe('Cursor Async Iterator Tests', function () {
...
+    it('cleans up cursor when breaking out of for await of loops', async function () {
+      const cursor = collection.find();
+
+      for await (const doc of cursor) {
+        expect(doc).to.exist;
+        break;
+      }
+
+      expect(cursor.closed).to.be.true;
+    });
```
nbbeeken, line 92: Let's add just a couple more cases / Qs:
- Won't the finally logic always be run at the end of the loop or is it only when there's break? We can add an assertion to the existing tests for the expected "closed" state either way.
- Can we add a test that attempt to keep using the cursor after the `for await` has done the clean up, ex. `expect(async () => for await (xx)).to.throw(X)`
- We have a gate on calling close only if it has not been called already (using the `closed` flag), I'm not sure it's easily reachable since we have the early return based on `closed=true`. I think it's worth capturing that behavior's significance, does close fail or repeat steps if it's called more than once?
  - We could use the iterator manually, call `cursor[Symbol.asyncIterator]()` and use `.return()` + sinon to check that .close is only called once
  - We could double check that close does not repeat steps if called more than once (this may be more trouble than it's worth in this PR, but I think `needsToEmitClosed` based on `!this[kClosed]` provides idempotency).

nbbeeken (changes requested):

test/integration/node-specific/cursor_async_iterator.test.js
durran: I've added assertions to the other tests that checks the cursor is closed on every one once they are exahusted. I've also added a test that asserts the second attempted iteration immediately returns, and also spies on the cursor to assert that `close()` is called only once. I think these cover the points you were getting at here but let me know if I missed something.

nbbeeken (approved):

PR #3508, opened 2023-01-03
feat(NODE-4509): send 1 with hello commands

durran:
### Description

Whenever the driver itself issues a `hello` or legacy hello, we now set the command value to `1` instead of `true`.

#### What is changing?
- Changes the commands to send `1` instead of `true`.
- Changes tests that also send the commands to the same.

None

NODE-4509


src/cmap/connect.ts
```diff
@@ -231,7 +231,7 @@ export function prepareHandshakeDocument(
   const { serverApi } = authContext.connection;
 
   const handshakeDoc: HandshakeDocument = {
-    [serverApi?.version ? 'hello' : LEGACY_HELLO_COMMAND]: true,
+    [serverApi?.version ? 'hello' : LEGACY_HELLO_COMMAND]: 1,
```
nbbeeken, line 234: Is there a way we can get regression coverage here? It looks like we only have a test that asserts the monitor will use `hello: 1`.

test/integration/mongodb-handshake/promise_handshake.test.js
```diff
@@ -28,7 +28,7 @@ describe('Handshake', function () {
         // Execute legacy hello command
         client
           .db(configuration.db)
-          .command({ [LEGACY_HELLO_COMMAND]: true })
+          .command({ [LEGACY_HELLO_COMMAND]: 1 })
```
nbbeeken, line 31: This test file doesn't add much value to our suite, + the strange constraints like poolsize, using legacy format strings, only on single, promise usage combined with mocha done style. Can we just delete this? 😅

src/cmap/connect.ts
durran: Added 2 unit tests for this now.

test/integration/mongodb-handshake/promise_handshake.test.js
durran: Removed.

nbbeeken (approved):

baileympearson (approved):

PR #3509, opened 2023-01-03
docs(NODE-4911): add dns connection issue notes to READ.ME

baileympearson:


README.md
```diff
@@ -133,6 +133,17 @@ Add code to connect to the server and the database **myProject**:
 > However, all async API calls support an optional callback as the final argument,
 > if a callback is provided a Promise will not be returned.
 
+> **NOTE:** Resolving DNS Connection issues
+>
+> Node18 changed the default dns resolution ordering from ipv4 first always to the ordering
```
alexbevi, line 138: Should we change this from "Node18" to "Node.js 17". It might be worth mentioning the change to [`--dns-result-order`](https://nodejs.org/docs/latest-v17.x/api/cli.html#--dns-result-orderorder) defaults?

alexbevi: I believe https://github.com/nodejs/node/pull/39987 is what brought this change in

baileympearson: Odd-numbered versions of node are development versions.  Although the change went into Node17, Node18 is the earliest LTS that supports this change.  My perspective is that it makes sense to use Node18, but I'm open to other opinions.

> It might be worth mentioning the change to [--dns-result-order](https://nodejs.org/docs/latest-v17.x/api/cli.html#--dns-result-orderorder) defaults?

Good call, I'll add this to the list of workarounds.

> I believe https://github.com/nodejs/node/pull/39987 is what brought this change in

Yeah, that's the commit.  Do you think we should mention it specifically?  

My gut says to keep it simple because although the commit does point to the exact change, that's less useful to first-time-users than the description of the issue that I provided (open to reworking it) and the listed workarounds.

dariakp: @baileympearson is on point regarding LTS versions (we should keep the docs focused on LTS versions only) and I tend to agree about not needing to reference the exact commit

README.md
```diff
@@ -133,6 +133,17 @@ Add code to connect to the server and the database **myProject**:
...
+> that dns provider returns them in.  In some environments, this can result in `localhost` resolving to
+> an ipv6 address instead of ipv4 and can result in a failure to connect to the server.
```
dariakp, line 140:
```suggestion
> Node.js 18 changed the default DNS resolution ordering from always prioritizing ipv4 to the ordering
> returned by the DNS provider.  In some environments, this can result in `localhost` resolving to
> an ipv6 address instead of ipv4 and a consequent failure to connect to the server.
```

dariakp (changes requested):
Just a few wording suggestions, wanted to keep capitalization consistent for "DNS"

README.md
alexbevi: > Odd-numbered versions of node are development versions. Although the change went into Node17, Node18 is the earliest LTS that supports this change. My perspective is that it makes sense to use Node18, but I'm open to other opinions.

That makes sense to me - no need to change the version.

alexbevi: > My gut says to keep it simple because although the commit does point to the exact change, that's less useful to first-time-users than the description of the issue that I provided (open to reworking it) and the listed workarounds.

Likely no value in adding the commit reference - I mostly put it there to help verify this is in fact where the change came in.

alexbevi (approved):
LGTM

dariakp (approved):

nbbeeken (approved):

PR #3511, opened 2023-01-04
feat(NODE-4924)!: remove mapReduce collection helper

baileympearson:
#### What is changing?

`Collection.mapReduce` is removed.

Nope.


src/utils.ts
```diff
@@ -1328,21 +1328,8 @@ export function shuffle<T>(sequence: Iterable<T>, limit = 0): Array<T> {
...
-  }
-
-  if (
-    command.mapReduce &&
-    options &&
-    options.out &&
-    (options.out.inline === 1 || options.out === 'inline')
-  ) {
-    return true;
-  }
```
addaleax, line 1343: Is it okay to remove this here? If I follow the code correctly, this method is called for all commands, including `client.db().command()`, so this change would also affect users who run `mapReduce` through that instead of the helper (and since the migration notes mention this as an alternative for cases in which users can’t use agg pipelines, that seems to be a supported use case).

baileympearson: I think this is a judgement call.  Do you have an opinion? 

When a user manually constructs the command, they can attach their own read concern.  I think it's reasonable to say that because we're dropping support for `mapReduce`, that means we drop support for attaching `readConcern`s automatically when executing a runCommand.

Alternatively, we could continue to support automatically attaching the read concern to `mapReduce` commands.  One argument in favor of this approach is that the `Db.command()` options does take a read concern, so users may assume that the read concern they supply as a part of the options object would get attached to their command.

addaleax: > the `Db.command()` options does take a read concern, so users may assume that the read concern they supply as a part of the options object would get attached to their command.

I don’t have a strong opinion, but this seems like a reasonable expectation to me, esp. if this works for other commands.

durran: I think it's reasonable to to keep it in until the driver's minimum server version has completely removed the command.

durran: Just a note as well because it seemed a bit unclear here. This is only determining if the command supports read concern - it still must be explicitly provided. This change would have just ignored the provided one in the inlined `mapReduce` case.

dariakp:
@baileympearson README changes LGTM

test/tools/spec-runner/index.js
```diff
@@ -699,13 +702,11 @@ const kOperations = new Map([
...
     'mapReduce',
+    // eslint-disable-next-line no-unused-vars
     (operation, collection, context /*, options */) => {
-      const args = operation.arguments;
-      const map = args.map;
-      const reduce = args.reduce;
-      const options = { session: maybeSession(operation, context) };
-      if (args.out) options.out = args.out;
-      return collection.mapReduce(map, reduce, options);
+      throw new Error(
```
nbbeeken, line 707: The mapReduce retryable-reads tests are somehow skipping so this error didn't make those tests fail, I'm not seeing it in the filter in `retryable_reads.spec.test.js`, any ideas?

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,30 @@ The following is a detailed collection of the changes in the major v5 release of
...
+
+The `mapReduce` helper has been removed from the `Collection` class.  The `mapReduce` operation has been
+deprecated in favor of the aggregation pipeline since MongoDB server version 5.0.  It is recommended
+to migrate code that uses `Collection.mapReduce` to use the aggregation pipeline.
+
+If the `mapReduce` command must be used, the `Db.command()` helper can be used to run the raw
+`mapReduce` command.
+
+```typescript
+const command = {
```
nbbeeken, line 29: Should we show an example of how this would have looked with the helper syntax? Like for someone new to constructing a command themselves it's not clear that the mapReduce property is the collection name.

Also since the preferred migration is to use agg, can we leave a link to this page: https://www.mongodb.com/docs/manual/reference/map-reduce-to-aggregation-pipeline/

nbbeeken (changes requested):

test/tools/spec-runner/index.js
baileympearson: Yeah, this threw me for a loop initially too.  The `filter` argument in `generateTopologyTests` returns true for tests that run, false otherwise.  The `filter` for the retryable reads spec test is actually an inclusion filter, not exclusion.

I added the doc comment to `generateTopologyTests` after getting confused by that initially.

baileympearson: I considered a minor refactor but decided not to:

```typescript
  generateTopologyTests(testSuites, testContext, spec => {
    const testsToRun = [
      ... list all test regexes here
    ];
    return testsToRun.some(regex => spec.description.match(regex));
  });

```

want me to make the change?

test/integration/client-side-encryption/client_side_encryption.spec.test.ts
```diff
@@ -61,7 +61,11 @@ const skippedAuthTests = [
 const skippedNoAuthTests = ['getMore with encryption', 'operation fails with maxWireVersion < 8'];
 
 const SKIPPED_TESTS = new Set([
-  ...(isAuthEnabled ? skippedAuthTests.concat(skippedNoAuthTests) : skippedNoAuthTests)
+  ...(isAuthEnabled ? skippedAuthTests.concat(skippedNoAuthTests) : skippedNoAuthTests),
+  ...[
+    // the node driver does not have a mapReduce helper
+    'mapReduce deterministic encryption (unsupported)'
```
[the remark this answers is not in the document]
baileympearson: Here's the spec section I based this off of: https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#auto-encrypt-and-decrypt

"an encrypted MongoClient MUST attempt to auto encrypt all commands"

etc/notes/CHANGES_5.0.0.md
baileympearson: done!

test/tools/spec-runner/index.js
nbbeeken: Your call, I'm happy that we looked into it and know why they weren't run

etc/notes/CHANGES_5.0.0.md
```diff
@@ -20,12 +20,25 @@ The following is a detailed collection of the changes in the major v5 release of
...
 If the `mapReduce` command must be used, the `Db.command()` helper can be used to run the raw
 `mapReduce` command.
 
 ```typescript
+// using the Collection.mapReduce helper in <4.x drivers
+const collection = db.collection('my-collection');
+
+await collection.mapReduce(
+  function() { emit(this.user_id, 1); },
+  function(, vals) { return 1 },
```
nbbeeken, line 34:
```suggestion
  function(k, vals) { return 1 },
```

nbbeeken (changes requested):

nbbeeken (changes requested):
CI issues, I think something caused by the merge, import related

nbbeeken (approved):

PR #3514, opened 2023-01-04
test(NODE-4919): import mongodb-legacy in tests

nbbeeken:
#### What is changing?

We use the mongodb-legacy package to continue supporting callback style tests

### DIFF
--- a/test/mongodb.ts
+++ b/test/mongodb.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-restricted-imports */
 import * as fs from 'node:fs';
 import * as path from 'node:path';
+import * as vm from 'node:vm';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 function printExports() {
@@ -21,6 +22,29 @@ function printExports() {
   for (const srcFile of driverSourceFiles) {
     console.log(`export * from '${path.relative(__dirname, srcFile)}';`);
   }
+}
+
+function importMongoDBLegacy() {
+  const mongodbLegacyEntryPoint = require.resolve('mongodb-legacy');
+  const mongodbLegacyLocation = path.dirname(mongodbLegacyEntryPoint);
+  const mongodbLegacyIndex = fs.readFileSync(mongodbLegacyEntryPoint, {
+    encoding: 'utf8'
+  });
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const localMongoDB = require('../src/index');
+  const ctx = vm.createContext({
+    module: { exports: null },
+    require: (mod: string) => {
+      if (mod === 'mongodb') {
+        return localMongoDB;
+      } else if (mod.startsWith('.')) {
+        return require(path.join(mongodbLegacyLocation, mod));
+      }
+      return require(mod);
+    }
+  });
+  vm.runInContext(mongodbLegacyIndex, ctx);
+  return ctx.module.exports;
 }
 
 export * from '../src/admin';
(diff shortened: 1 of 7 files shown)


nbbeeken:
[Patch](https://spruce.mongodb.com/version/63b5d2d42fbabe30811f8915/tasks)

test/mongodb.ts
```diff
@@ -22,6 +24,75 @@ function printExports() {
...
+ * @param exportsToOverride - An object that is an import of the MongoDB driver to be modified by this function
+ */
+function importMongoDBLegacy(exportsToOverride: Record<string, unknown>) {
+  const mongodbLegacyEntryPoint = require.resolve('mongodb-legacy');
+  const mongodbLegacyLocation = path.dirname(mongodbLegacyEntryPoint);
+  const mongodbLegacyIndex = fs.readFileSync(mongodbLegacyEntryPoint, {
+    encoding: 'utf8'
+  });
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const localMongoDB = require('../src/index');
```
dariakp, line 43: to capture the conversation from slack:
- need to add clarification/comments about how the import works and what it does so that it requires less thought to follow
- let's set up the import in a way where we can decide to turn it on or off via, say, env settings or a flag passed to a script, so that we can reuse our test suite to test new driver changes for legacy compatibility

dariakp (changes requested):

test/mongodb.ts
nbbeeken: There's a TS doc comment on the `importMongoDBLegacy` function and it now contains all the logic so there's an env var that can control it

nbbeeken:
#3522 The linter didn't detect imports from "src" correctly so we were missing some import changes. Looking into what's wrong with the linter, but regardless we'll want those imports changed as well before this PR

test/mongodb.ts
```diff
@@ -23,6 +25,75 @@ function printExports() {
   }
 }
 
+/**
+ * Using node's require resolution logic this function will locate the entrypoint for the `'mongodb-legacy'` module.
+ * Then execute the `mongodb-legacy` module in a `vm` context that replaces the global require function with a custom
```
dariakp, line 30:
```suggestion
 * Using node's require resolution logic this function will locate the entrypoint for the `'mongodb-legacy'` module,
 * then execute the `mongodb-legacy` module in a `vm` context that replaces the global require function with a custom
```

test/mongodb.ts
```diff
@@ -125,3 +196,8 @@ export * from '../src/write_concern';
 
 // Must be last for precedence
 export * from '../src/index';
+
+if (process.env.DISABLE_MONGODB_LEGACY !== 'true') {
```
dariakp, line 200: from slack, rename this variable to its logical opposite and default it to true with a TODO to remove the default in the ticket to convert callback tests

test/unit/mongodb-legacy.test.ts
```diff
@@ -0,0 +1,9 @@
+import { expect } from 'chai';
+
+import { Db } from '../mongodb';
```
dariakp, line 3: from slack, test more things, and move the file into `tools`

dariakp (changes requested):

test/unit/mongodb-legacy.test.ts
nbbeeken: done

test/mongodb.ts
nbbeeken: done

test/unit/assorted/client.test.js
```diff
@@ -38,7 +38,10 @@ describe('Client (unit)', function () {
 
     return client.connect().then(() => {
       expect(handshake).to.have.nested.property('client.driver');
-      expect(handshake).nested.property('client.driver.name').to.equal('nodejs|mongoose');
+      expect(handshake)
+        .nested.property('client.driver.name')
+        // TODO(NODE-4979): Our tests depend on mongodb-legacy
```
dariakp, line 43: I think this will be the case even after they don't depend, since it'll be a supported use case, same as mongoose really

nbbeeken: Updated the comment to be clearer, looking back on that months from now the context was missing

dariakp: I think you misunderstood my comment - even after we complete NODE-4979, we want to continue using our library to test the wrapper (in a separate build, see NODE-4980); only once the legacy wrapper is no longer supported and we don't test it in the CI (which we don't have a timeline or a ticket for at the moment, but I suppose we could create a placeholder ticket like we have for legacy hello that gets bumped from major version to major version) can we get rid of that expectation.

dariakp (approved):

test/unit/assorted/client.test.js
nbbeeken: fixed the test to handle both cases so it shouldn't fail when we flip it

dariakp (approved):

PR #3515, opened 2023-01-05
feat(NODE-4034)!: make internal bulk result private

durran:
### Description

Makes the internal `BulkResult` on the `BulkWriteResult` private.

#### What is changing?
- Marks the `BulkResult` interface as internal.
- Makes the `result` property of the `BulkWriteResult` private and non-enumerable.
- Makes all properties of the `BulkWriteResult` per the crud spec definition actual properties instead of getters.
- Removes `toJSON()`.
- DRYs up the generation of the inserted/upserted id maps.
- Eliminates `insertedIds` from the result enumerable in unified tests.

mongosh CI failures fix in this PR: https://github.com/mongodb-js/mongosh/pull/1386

None

NODE-4034/NODE-4393/NODE-4973

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -163,12 +163,33 @@ export class Batch<T = Document> {
   }
 }
 
+function generateIdMap(ids: Document[]): { [key: number]: any } {
+  const idMap: { [index: number]: any } = {};
+  for (const doc of ids ?? []) {
+    idMap[doc.index] = doc._id;
+  }
+  return idMap;
+}
 /**
  * @public
  * The result of a bulk write.
  */
 export class BulkWriteResult {
   private result: BulkResult;
+  /** Number of documents inserted. */
+  insertedCount: number;
+  /** Number of documents matched for update. */
+  matchedCount: number;
+  /** Number of documents modified. */
+  modifiedCount: number;
+  /** Number of documents deleted. */
+  deletedCount: number;
+  /** Number of documents upserted. */
+  upsertedCount: number;
+  /** Upserted document generated Id's, hash key is the index of the originating operation */
+  upsertedIds: { [key: number]: any };
+  /** Inserted document generated Id's, hash key is the index of the originating operation */
+  insertedIds: { [key: number]: any };
 
   /**
    * Create a new BulkWriteResult instance
@@ -176,46 +197,14 @@ export class BulkWriteResult {
    */
   constructor(bulkResult: BulkResult) {
     this.result = bulkResult;
+    this.insertedCount = this.result.nInserted ?? 0;
+    this.matchedCount = this.result.nMatched ?? 0;
+    this.modifiedCount = this.result.nModified ?? 0;
+    this.deletedCount = this.result.nRemoved ?? 0;
+    this.upsertedCount = this.result.upserted.length ?? 0;
+    this.upsertedIds = generateIdMap(this.result.upserted);
+    this.insertedIds = generateIdMap(this.result.insertedIds);
     Object.defineProperty(this, 'result', { value: this.result, enumerable: false });
-  }
-
-  /** Number of documents inserted. */
-  get insertedCount(): number {
-    return this.result.nInserted ?? 0;
-  }
-  /** Number of documents matched for update. */
-  get matchedCount(): number {
-    return this.result.nMatched ?? 0;
-  }
-  /** Number of documents modified. */
-  get modifiedCount(): number {
-    return this.result.nModified ?? 0;
-  }
-  /** Number of documents deleted. */
-  get deletedCount(): number {
-    return this.result.nRemoved ?? 0;
-  }
-  /** Number of documents upserted. */
-  get upsertedCount(): number {
-    return this.result.upserted.length ?? 0;
-  }
-
-  /** Upserted document generated Id's, hash key is the index of the originating operation */
-  get upsertedIds(): { [key: number]: any } {
-    const upserted: { [index: number]: any } = {};
-    for (const doc of this.result.upserted ?? []) {
-      upserted[doc.index] = doc._id;
-    }
-    return upserted;
-  }
-
-  /** Inserted document generated Id's, hash key is the index of the originating operation */
-  get insertedIds(): { [key: number]: any } {
-    const inserted: { [index: number]: any } = {};
-    for (const doc of this.result.insertedIds ?? []) {
-      inserted[doc.index] = doc._id;
-    }
-    return inserted;
   }
 
   /** Evaluates to true if the bulk operation correctly executes */
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -123,11 +123,7 @@ export type AnyBulkWriteOperation<TSchema extends Document = Document> =
   | { deleteOne: DeleteOneModel<TSchema> }
   | { deleteMany: DeleteManyModel<TSchema> };
 
-/**
- * @public
- *
- * @deprecated Will be made internal in 5.0
- */
+/** @internal */
 export interface BulkResult {
   ok: number;
   writeErrors: WriteError[];
@@ -172,8 +168,7 @@ export class Batch<T = Document> {
  * The result of a bulk write.
  */
 export class BulkWriteResult {
-  /** @deprecated Will be removed in 5.0 */
-  result: BulkResult;
+  private result: BulkResult;
 
   /**
    * Create a new BulkWriteResult instance
@@ -181,6 +176,7 @@ export class BulkWriteResult {
    */
   constructor(bulkResult: BulkResult) {
     this.result = bulkResult;
+    Object.defineProperty(this, 'result', { value: this.result, enumerable: false });
   }
 
   /** Number of documents inserted. */
@@ -314,13 +310,8 @@ export class BulkWriteResult {
     }
   }
 
-  /* @deprecated Will be removed in 5.0 release */
-  toJSON(): BulkResult {
-    return this.result;
-  }
-
   toString(): string {
-    return `BulkWriteResult(${this.toJSON()})`;
+    return `BulkWriteResult(${this.result})`;
   }
 
   isOk(): boolean {
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -530,8 +530,8 @@ function executeCommands(
     }
 
     // Merge the results together
-    const writeResult = new BulkWriteResult(bulkOperation.s.bulkResult);
     const mergeResult = mergeBatchResults(batch, bulkOperation.s.bulkResult, err, result);
+    const writeResult = new BulkWriteResult(bulkOperation.s.bulkResult);
     if (mergeResult != null) {
       return callback(undefined, writeResult);
     }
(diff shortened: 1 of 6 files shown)


src/bulk/common.ts
```diff
@@ -550,8 +530,8 @@ function executeCommands(
     }
 
     // Merge the results together
-    const writeResult = new BulkWriteResult(bulkOperation.s.bulkResult);
     const mergeResult = mergeBatchResults(batch, bulkOperation.s.bulkResult, err, result);
+    const writeResult = new BulkWriteResult(bulkOperation.s.bulkResult);
```
durran, line 534: `mergeBatchResults()` mutates the provided `bulkResult` so we move the construction of the `BulkWriteResult` until afterwards so it acts as immutable.

src/bulk/common.ts
```diff
@@ -167,59 +163,48 @@ export class Batch<T = Document> {
...
-    }
-    return inserted;
+  insertedIds: { [key: number]: any };
+
+  /**
+   * Create a new BulkWriteResult instance
+   * @internal
+   */
+  constructor(bulkResult: BulkResult) {
+    this.result = bulkResult;
```
baileympearson, line 199: Is there a reason to assign `result` here and then override it in line 207?

src/bulk/common.ts
```diff
@@ -167,59 +163,48 @@ export class Batch<T = Document> {
...
-
-  /**
-   * Create a new BulkWriteResult instance
-   * @internal
-   */
-  constructor(bulkResult: BulkResult) {
-    this.result = bulkResult;
-  }
-
+  private result: BulkResult;
```
baileympearson, line 178: All the bulk write result properties used to be getters and as a result they were readonly.  I think we should keep this behavior.

We could mark the properties as `readonly`.  This only impacts Typescript users but I think that's alright.

src/bulk/common.ts
```diff
@@ -167,59 +163,48 @@ export class Batch<T = Document> {
   }
 }
 
+function generateIdMap(ids: Document[]): { [key: number]: any } {
+  const idMap: { [index: number]: any } = {};
+  for (const doc of ids ?? []) {
```
baileympearson, line 168:
```suggestion
  for (const doc of ids) {
```
`ids` should always be defined as an array, according to Typescript

src/bulk/common.ts
```diff
@@ -167,59 +163,48 @@ export class Batch<T = Document> {
   }
 }
 
+function generateIdMap(ids: Document[]): { [key: number]: any } {
```
baileympearson, line 166: just a thought - this is only used on the bulk write result class.  for organization purposes, we could make it a private static method on the class.  

thoughts?

src/bulk/common.ts
```diff
@@ -550,8 +530,8 @@ function executeCommands(
     }
 
     // Merge the results together
-    const writeResult = new BulkWriteResult(bulkOperation.s.bulkResult);
     const mergeResult = mergeBatchResults(batch, bulkOperation.s.bulkResult, err, result);
+    const writeResult = new BulkWriteResult(bulkOperation.s.bulkResult);
     if (mergeResult != null) {
       return callback(undefined, writeResult);
     }
```
baileympearson, line 537:
```suggestion
```

optional cleanup - mergeBatchResults never returns anything, so this is dead code.

src/bulk/common.ts
baileympearson: Just clarification for further reviewers.  

Before the changes in this PR, `BulkWriteResult` used getters to provide access to each property on the raw bulk result, so constructing the `BulkWriteResult` before modifying the raw result was okay.

Now, we assign the properties directly in the constructor, so we need to create the `BulkWriteResult` after we process the raw result.

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -150,8 +150,12 @@ async function executeScenarioTest(test, ctx: RetryableWriteTestContext) {
     const result = resultOrError;
     const expected = test.outcome.result;
 
-    // TODO(NODE-4034): Make CRUD results spec compliant
-    expect(result.value ?? result).to.deep.include(expected);
+    const actual = result.value ?? result;
+    // Some of our result classes contain the optional 'acknowledged' property which is
+    // not part of the test expectations.
+    for (const property in expected) {
```
baileympearson, line 156: Should we also be checking that there are not additional properties defined on the result beyond what are expected by the test (with the exception of acknowledged)?

baileympearson (changes requested):

src/bulk/common.ts
durran: The override was to set it as non-enumerable but typescript complained if I did it all in one go here. (It couldn't figure out that `Object.defineProperty` was setting `this.result`.

test/integration/retryable-writes/retryable_writes.spec.test.ts
durran: Added test on properties on the actual bulk write result.

src/bulk/common.ts
durran: Removed.

src/bulk/common.ts
durran: Changed per suggestion.

src/bulk/common.ts
durran: Updated.

src/bulk/common.ts
durran: Changed all properties to readonly.

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -14,6 +14,17 @@ interface RetryableWriteTestContext {
...
 
+const ALLOWED_BULK_PROPERTIES = [
+  'acknowledged',
+  'insertedCount',
+  'matchedCount',
+  'modifiedCount',
+  'deletedCount',
+  'upsertedCount',
+  'upsertedIds',
+  'insertedIds'
```
baileympearson, line 25:
```suggestion
  'insertedIds',
  'upsertedId'
```

and maybe more - CI is pretty red.

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -150,8 +161,17 @@ async function executeScenarioTest(test, ctx: RetryableWriteTestContext) {
...
-    // TODO(NODE-4034): Make CRUD results spec compliant
-    expect(result.value ?? result).to.deep.include(expected);
+    const actual = result.value ?? result;
+    // Some of our result classes contain the optional 'acknowledged' property which is
+    // not part of the test expectations.
+    for (const property in expected) {
+      expect(actual).to.have.deep.property(property, expected[property]);
+    }
+
+    // Check we don't have any extra properties on the bulk write result.
```
baileympearson, line 171: Instead of whitelisting the allowed properties, we could compare the properties on `actual` to those on `expected` and assert there are no extra properties

baileympearson (changes requested):

test/integration/retryable-writes/retryable_writes.spec.test.ts
durran: Actually this I've just seen as a problem to assert as the result on `findOneAnd...` methods is not a `BulkWriteResult` but the actual matching document - so it can have any properties on it. So going to remove the check on L171.

test/integration/retryable-writes/retryable_writes.spec.test.ts
durran: This is now removed as we have cases where actual documents are returned. See next comment.

src/bulk/common.ts
```diff
@@ -172,54 +168,44 @@ export class Batch<T = Document> {
...
+   */
+  constructor(bulkResult: BulkResult) {
+    this.result = bulkResult;
+    this.insertedCount = this.result.nInserted ?? 0;
+    this.matchedCount = this.result.nMatched ?? 0;
+    this.modifiedCount = this.result.nModified ?? 0;
+    this.deletedCount = this.result.nRemoved ?? 0;
+    this.upsertedCount = this.result.upserted.length ?? 0;
+    this.upsertedIds = BulkWriteResult.generateIdMap(this.result.upserted);
+    this.insertedIds = BulkWriteResult.generateIdMap(this.result.insertedIds);
```
dariakp, line 207: Have you had a chance to check with any other drivers to see why they aren't running into the same insertedIds issues? It may be good to know if they are also implementing an internal work around, or aren't spec compliant themselves, or perhaps interpreting the spec differently.

durran: Ruby iterates over the keys in the expected result and checks the actual object for that specific property, so it does not error when the actual result has more properties than the expected. Python does the same. It seems we are probably the only driver looking for an exact match of the actual and expected object properties, which makes us less flexible on adding other properties to our result classes.

The CRUD spec is clear about required and optional fields in the result objects, but does not say if additional properties on the result classes are forbidden. It seems like Node took the spec strictly but other drivers allowed a bit more flexibility - neither is incorrect in my opinion.

baileympearson: @dariakp Are you okay with this answer?  This is the last outstanding thread on this PR.

dariakp: The "Evaluating Matches" section of the unified runner spec is very explicit about asserting that the properties of a subdocument inside an expected result property must match exactly:
https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#evaluating-matches

also 

> Test runners MUST NOT permit additional fields in nested documents. 

here: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#allowing-extra-fields-in-root-level-documents

So, per the spec, we are correct, and the spec tests must account for allowed optional properties via special test operators (basically, what you have filed in the DRIVERS-2523 is addressing a bug in the test implementation for RewrapManyDataKey; it is also an omission for the unified spec that we don't have the rewrapManyDataKey operation defined in the unified format spec and are also not referencing the spec where that operation is expected to be defined). This is an issue for insertMany tests, too, and for the "acknowledged" property in general (which perhaps no driver implements); surprised that is not coming up here (unless we already have the logic to handle it). I don't think we should treat BulkWriteResult as a root level document, because we do know what optional fields are allowed to exist on it, and we do not want it to have arbitrary fields beyond the ones defined by the spec.

dariakp: I added a comment to the drivers ticket, we don't need to keep this PR open on that account

durran: Since the test updates merged for DRIVERS-2523 I just went ahead and added them here.

baileympearson (approved):

src/bulk/common.ts
dariakp: @durran Should that ticket be updated to "teams implementing"? (That way we can tag the generated node ticket number in this PR and it'll auto-link in jira)

durran: I did update it but for some reason it didn't generate any individual language tickets. Hmm.

durran: Description updated with [NODE-4973](https://jira.mongodb.org/browse/NODE-4973) now.

durran: Also linked all 3 tickets to NODE-4034 now in JIRA.

PR #3519, opened 2023-01-11
feat(NODE-4877): Add support for useBigInt64

W-A-James:
### Description

Add support for BSON's `useBigInt64` flag to the driver

#### What is changing?

- `useBigInt64` flag is now a part of the `BSONSerializationOptions` interface
- `useBigInt64` flag is now therefore part of all implementers of this interface

No

- Integrating the BigInt64 support in the BSON library into the driver

	- [x] NODE-5072 Change pattern for setting BSON options
	- [x] NODE-5073 Remove unused `applyCommonQueryOptions` function


W-A-James:
Link to evergreen patch with `useBigInt64=true` by default to 
 see what, if anything, breaks. https://spruce.mongodb.com/version/63c7199232f4171734d3a529/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

nbbeeken:
I think the option isn't actually making it to the BSON layer, we "sanitize" the options at the wire protocol layer to make sure that arbitrary options don't reach the bson deserializer, take a look at #2840

dariakp:
This PR cannot merge until we do the 5.0 release.

nbbeeken:
Can you post a link to an evg patch where `useBigInt64` is set to true anywhere we call BSON.deserialize in the driver (should be four places in commands.ts). We want to see if the errors from that are salient to what can occur when the option is enabled 

some errors might just be assertions no longer being equal, that would be expected

W-A-James:
Evergreen patch with `useBigInt64=true` as default:

https://spruce.mongodb.com/version/63d04caea4cf47516296a359/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

nbbeeken:
Lot's of red, unfortunately enough to make it so we can't see the output, atlas connectivity failing though might be a good place to start as that "shouldn't" have been impacted at all.

```ts
// TODO(NODE-2674): Preserve int64 sent from MongoDB
```
You'll want to look for this comment, likely every spot needs an extra condition to accept bigint

W-A-James:
https://spruce.mongodb.com/version/63e40bea32f4173ea33ff9ff/tasks

Link to evergreen patch with `useBigInt64=true` by default

test/integration/crud/insert.test.js
```diff
@@ -428,25 +428,21 @@ describe('crud - insert', function () {
...
-                client.close(done);
-              });
-            }
-          );
-        });
+      test: async function () {
+        const configuration = this.configuration;
+        const client = configuration.newClient(configuration.writeConcernMax(), { maxPoolSize: 1 });
+        await client.connect();
+        this.defer(async () => {
```
durran, line 435: All the cases in these tests that defer a `client.close()` are not necessary - we can just `await client.close()` at the end of the tests.

test/integration/node-specific/bson-options/useBigInt64.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  beforeEach(function () {
+    configuration = this.configuration;
+    return setupDatabase(this.configuration);
+  });
+
+  describe('when not provided to client', async function () {
+    let client: MongoClient;
+
+    beforeEach(async function () {
+      client = await configuration.newClient(configuration.writeConcernMax());
```
durran, line 17: async/await not needed here - `newClient()` doesn't return a promise.

test/integration/node-specific/bson-options/useBigInt64.test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    it('is set to driver default (useBigInt64=false)', async function () {
+      expect(client.s.bsonOptions.useBigInt64).to.exist;
+      expect(client.s.bsonOptions.useBigInt64).to.be.false;
+    });
+  });
+
+  describe('when set at client level', function () {
+    let client: MongoClient;
+    beforeEach(async function () {
```
durran, line 32: No need for async function here.

test/integration/node-specific/bson-options/useBigInt64.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    afterEach(async function () {
+      await client.close();
+    });
+
+    it('supercedes driver level', function () {
+      expect(client.s.bsonOptions.useBigInt64).to.exist;
+      expect(client.s.bsonOptions.useBigInt64).to.be.true;
+    });
+  });
+
```
durran, line 45: Can we add a test in this file as well to check the error case `useBigInt64` set to `true` and `promoteLongs` is `false`. The patch linked shows that happening but since it's called out in the acceptance criteria we should explicitly test it here.

durran (changes requested):

test/integration/node-specific/bson-options/useBigInt64.test.ts
W-A-James: Added the new test, but the options are only checked on the call to `MongoClient.connect`. Doing the check implicitly in connect because it has to use BSON.deserialize does satisfy the AC, but feels brittle.

Do we want to explicitly do this check in the constructor for the `MongoClient`? It could be done in the `parseOptions` helper function in src/connection_string.ts where we validate the other options.

durran: Doing it in `parseOptions` sounds like a good idea.

W-A-James:
Link to new evergreen patch with `useBigInt64=true` by default https://spruce.mongodb.com/version/63efd6123627e0388b69d98e/tasks

durran (approved):

test/integration/client-side-encryption/client_side_encryption.prose.06.corpus.test.js
```diff
@@ -20,7 +20,9 @@ describe('Client Side Encryption Prose Corpus Test', function () {
 
   const corpusDir = path.resolve(__dirname, '../../spec/client-side-encryption/corpus');
   function loadCorpusData(filename) {
-    return EJSON.parse(fs.readFileSync(path.resolve(corpusDir, filename)), { relaxed: false });
+    return EJSON.parse(fs.readFileSync(path.resolve(corpusDir, filename)), {
+      relaxed: false
```
nbbeeken, line 24: mind reverting this? just cus its whitespace change, no need to modify history here (only asking since I'm asking for other changes)

nbbeeken (changes requested):

src/connection_string.ts
```diff
@@ -255,6 +255,15 @@ export function parseOptions(
...
   }
 
+  // validate BSONOptions
+  if (options.useBigInt64 && !options.promoteLongs) {
+    throw new MongoAPIError('Must request either bigint or Long for int64 deserialization');
+  }
+
+  if (options.useBigInt64 && !options.promoteValues) {
+    throw new MongoAPIError('Must request either bigint or Long for int64 deserialization');
+  }
```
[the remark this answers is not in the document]
baileympearson: unsure, the error messages are different.  but we'll never hit the second if-statement anyways.  Also I'm not sure this is the right place for these errors. Both of these options can be set per operation, so if we want to prevent users from using both options concurrently we need to put the logic in a place where every operation will encounter it.

also - it might be worthwhile to check for "existence" instead of truthiness here for `promoteValues`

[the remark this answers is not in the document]
W-A-James: No, one is comparing against `options.promoteLongs` and the other is against `options.promoteValues`

src/cmap/connection.ts
```diff
@@ -661,6 +661,7 @@ function write(
     command: !!options.command,
 
     // for BSON parsing
+    useBigInt64: typeof options.useBigInt64 === 'boolean' ? options.useBigInt64 : false,
```
baileympearson, line 664: When will this ever _not_ be set? 

`resolveBSONOptions` will ensure that this is always set to a value

src/cmap/wire_protocol/shared.ts
```diff
@@ -41,6 +41,7 @@ export function applyCommonQueryOptions(
 ): CommandOptions {
```
baileympearson, line 41: this function is unused in the driver, no need to modify it

test/unit/sessions.test.js
```diff
@@ -110,6 +116,17 @@ describe('Sessions - unit', function () {
         expect(session).property('clusterTime').to.equal(alsoValidTime);
       });
 
+      it('sets clusterTime to the one provided when the signature.keyId is a bigint', () => {
+        // NOTE: Prior to BigInt support, this would have thrown MongoInvalidArgumentError
```
baileympearson, line 120: why is this comment necessary?

baileympearson (changes requested):

src/connection_string.ts
```diff
@@ -255,6 +255,15 @@ export function parseOptions(
     mongoClient = undefined;
   }
 
+  // validate BSONOptions
+  if (options.useBigInt64 && !options.promoteLongs) {
+    throw new MongoAPIError('Must request either bigint or Long for int64 deserialization');
+  }
+
+  if (options.useBigInt64 && !options.promoteValues) {
+    throw new MongoAPIError('Must request either bigint or Long for int64 deserialization');
```
baileympearson, line 264: Instead of telling the user something imperatively, I think these error messages should describe the problem. Something like "cannot set option useBigInt64 with promoteLongs=false"

src/cmap/wire_protocol/shared.ts
W-A-James: If it's unused, should I just go ahead and remove the function entirely then?

src/connection_string.ts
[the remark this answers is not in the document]
baileympearson: follow-up - where do we test that combining options throws errors from the bson layer when the options are set for DB, collection, or operation? We only have tests for these two particular errors for the mongo client

baileympearson (changes requested):

src/cmap/wire_protocol/shared.ts
baileympearson: Not in this PR.  But go ahead and open a follow up PR to remove it (and ticket) 🙂 

I'm very particular about atomic commits in our history.

src/connection_string.ts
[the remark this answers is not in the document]
W-A-James: Good point here. In actuality though, throwing here duplicates logic from the bson library itself. If either one of these combinations of options were used in a call to BSON.deserialize, we would get a BSONError with the same message.

That makes me think removing this logic altogether is likely the best approach as having this logic here does imply we should have it in every other place you mentioned, but they are [already covered](https://github.com/mongodb/js-bson/blob/b6a15b5a5eeb0cf453f85fb5bbfeaf3add110015/src/parser/deserializer.ts#L127-L133).

@durran What are your thoughts?

nbbeeken (approved):
LGTM can't wait to use the biggest of ints!! 🎉

src/connection_string.ts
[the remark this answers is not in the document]
baileympearson: I think it's worthwhile to leave this logic here.

In this case, if both options are provided, errors will be thrown when the client is constructed (before any operations are run) if we remove this logic, no errors are surfaced for invalid client options until the first command is executed.

But we should still document with tests that we expect the bson library to throw these errors when executing a command with both of these arguments provided.

src/cmap/connection.ts
baileympearson: Sure, that works

test/integration/node-specific/bson-options/useBigInt64.test.ts
```diff
@@ -127,25 +127,119 @@ describe('useBigInt64 option', function () {
...
+            promoteLongs: false
+          });
+        }).to.throw(MongoAPIError, /Must request either bigint or Long for int64 deserialization/);
+      });
+    });
+
+    describe('when set at DB level', function () {
+      it('throws a BSONError', async function () {
+        client = configuration.newClient(configuration.writeConcernMax());
+        await client.connect();
```
baileympearson, line 153:
```suggestion
```

(optional - applies to all tests here that connect) we can omit the connecting if we want, since the driver supports auto-connection now

src/connection_string.ts
```diff
@@ -255,6 +255,15 @@ export function parseOptions(
     mongoClient = undefined;
   }
 
+  // validate BSONOptions
+  if (options.useBigInt64 && !(options.promoteLongs ?? true)) {
```
baileympearson, line 259: This works, but I think we usually check `typeof` because it's more explicit
```suggestion
  if (options.useBigInt64 && typeof options.promoteLongs === 'boolean' && options.promoteLongs) {
```

baileympearson (changes requested):

test/integration/change-streams/change_stream.test.ts
```diff
@@ -1722,6 +1724,60 @@ describe('Change Streams', function () {
       });
     });
 
+    context('useBigInt64', () => {
+      const useBigInt64FalseTest = async function (options: ChangeStreamOptions) {
```
baileympearson, line 1728:
```suggestion
      async function useBigInt64FalseTest(options: ChangeStreamOptions) {
```
or
```suggestion
      const useBigInt64FalseTest = async (options: ChangeStreamOptions) => {
```

Assigning an anonymous function to a variable is unusual - you'd typically just declare the function.
Alternatively, since we don't need to worry about binding `this`, we can just use an arrow function.

src/cursor/abstract_cursor.ts
```diff
@@ -642,6 +642,8 @@ export abstract class AbstractCursor<
           this[kId] =
             typeof response.cursor.id === 'number'
               ? Long.fromNumber(response.cursor.id)
```
baileympearson, line 644: Do we actually test this anywhere?  We don't enable `useBigInt64` in our tests, so we're not getting implicit coverage there.

The only place that I see some coverage is in the change streams tests this PR adds.

Can we add tests to confirm that cursors properly handle `useBigInt64: true`?  I think we can follow the example set in `integration/abstract_cursor.test.ts` (which uses a find cursor to test generic functionality).

test/integration/node-specific/bson-options/useBigInt64.test.ts
```diff
@@ -0,0 +1,234 @@
...
+    let client: MongoClient;
+
+    beforeEach(async function () {
+      client = configuration.newClient(configuration.writeConcernMax());
+    });
+
+    afterEach(async function () {
+      await client.close();
+    });
+
```
baileympearson, line 23: We could consider moving the `client` / `db` / `collection` variables up to the top-level describe, and then having a single `afterEach` cleanup hook.  There's a lot of duplication in cleanup

test/integration/node-specific/bson-options/useBigInt64.test.ts
```diff
@@ -0,0 +1,234 @@
...
+  beforeEach(function () {
+    configuration = this.configuration;
+    return setupDatabase(this.configuration);
+  });
+
+  describe('when not provided to client', async function () {
+    let client: MongoClient;
+
+    beforeEach(async function () {
+      client = configuration.newClient(configuration.writeConcernMax());
```
baileympearson, line 17: Just curious - why are we using `writeConcernMax`? 

Typically, we would either call `this.configuration.newClient()` directly (with no arguments) to get a standard URI, and `this.configuration.newClient({}, options)` to provide options.

test/integration/node-specific/bson-options/useBigInt64.test.ts
```diff
@@ -0,0 +1,234 @@
+import { expect } from 'chai';
```
baileympearson, line 1: this file should use snake case (`use_bigint_64.test.ts`) or something similar to follow our existing filenaming convention.

test/integration/node-specific/bson-options/useBigInt64.test.ts
```diff
@@ -0,0 +1,234 @@
...
+    });
+
+    afterEach(async function () {
+      await db
+        .dropCollection('useBigInt64Test')
+        .catch(() => expect.fail('failed to drop collection'));
+      await client.close().catch(() => expect.fail('failed to close client'));
+    });
+
+    it('deserializes to a bigint', async function () {
```
baileympearson, line 118: This isn't right.  What we expect is that it is possible to override more global `useBigInt64` configurations when we set it on the operation.  We don't always expect it to enable bigint deserialization.

So this test should be similar to the others. Except that we need to round trip information to confirm that the values are what we expect instead of inspecting options.  Alternatively, we could spy on the call to serialize and confirm that the option is set.  My preference would be the former

test/tools/unified-spec-runner/match.ts
```diff
@@ -131,7 +131,10 @@ TYPE_MAP.set(
 );
 TYPE_MAP.set(
   'long',
-  actual => (typeof actual === 'number' && Number.isInteger(actual)) || Long.isLong(actual)
```
baileympearson, line 134: Out of curiosity, is this to fix any failing tests or is this an improvement for the future? I don't see any unified tests being run with bigint enabled, so I'm surprised this is necessary

baileympearson (changes requested):

test/integration/node-specific/bson-options/useBigInt64.test.ts
W-A-James: Would it suffice to explicitly set the global option to false and change it to true at the operation level, then check that the option made its way through by checking the type of the result in addition to doing it the other way around (globally set to true, set to false at the operation level)?

test/integration/node-specific/bson-options/useBigInt64.test.ts
W-A-James: This was an artefact of copying over the test structure from one of the other bson option tests. I can go ahead and remove these.

test/tools/unified-spec-runner/match.ts
W-A-James: If I remember correctly, this was done to fix tests that were failing when `useBigInt64` was globally defaulted to true. @nbbeeken correct me if I'm wrong, but I believe it was to fix a failure in one of the client-side encryption tests.

nbbeeken: Correct, this was done when we were setting `useBigInt64` to true globally, this just helped cut out a lot of expected failures so we could better sort through the actual issues

baileympearson: OK. To be clear here, I'm in favor of this change because it's needed if we ever want to enable bigint serialization in our unified runner.

But I want to confirm that we are _not_ setting this option globally for our tests, correct?

W-A-James: Just to document it, we are _not_ setting the option to `true` globally, but we ran evergreen patches where that was the case just to document the test failures that occurred and make changes to the code/tests that made those failures more transparent, so rather than just resulting in an obscure error, it would result in something more readable, like this example: https://parsley.mongodb.com/test/mongo_node_driver_next_rhel80_large_hydrogen_test_3.6_replica_set_patch_e544d880b1731fa59ad6a0d740b946e79478cfa6_63efd6123627e0388b69d98e_23_02_17_19_31_41/0/41766c6f7a9248879f6067a39be845a5?bookmarks=0,8&shareLine=1

baileympearson (approved):

durran (approved):

PR #3520, opened 2023-01-11
feat(NODE-4738)!: remove dot notation support by default

baileympearson:
#### What is changing?

The `Filter` and `UpdateFilter` types have been reverted to pre-dot notation support.
Additionally, new types `StrictUpdateFilter` and `StrictFilter` are exported for consumption.  These types are experimental and provide the same dot-notation functionality as version 4.13.0 of the driver but they are not used in any driver CRUD APIs.

Yes.


etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+
+declare const collection: Collection<Schema>;
+// compiles pre-v4.3.0, fails in v4.3.0+
+collection.find({ 'user.name': 4 })
+```
+
+This change caused many problems for users, including slow compilation times and compile errors for
+valid dot notation queries.
+
+Driver 5.0 removes type checking on filter predicates.  The preceding example compile will compile with
```
durran, line 40:
```suggestion
Driver 5.0 removes type checking on filter predicates.  The preceding example will compile with
```

src/mongo_types.ts
```diff
@@ -541,3 +525,60 @@ export type NestedPathsOfType<TSchema, Type> = KeysOfAType<
...
+  $push?: PushOperator<TSchema>;
+  $pullAll?: PullAllOperator<TSchema>;
+  $bit?: OnlyFieldsOfType<
+    TSchema,
+    NumericType | undefined,
+    { and: IntegerType } | { or: IntegerType } | { xor: IntegerType }
+  >;
+} & Document;
+
+/** @public */
```
durran, line 571: Do we want this to be public as well or should we mark internal?

durran (changes requested):

src/mongo_types.ts
durran: The 2 should be public but I was referring specifically to `DotNotationMatchKeysAndValues` - does it need to be public because the other 2 are?

baileympearson: Oh, sorry.  I misread your message.

Yeah, if we mark this as internal, we get api-extractor warnings.

durran: Ah ok - thanks for clarifying.

durran (dismissed):

src/mongo_types.ts
```diff
@@ -541,3 +525,60 @@ export type NestedPathsOfType<TSchema, Type> = KeysOfAType<
   },
   Type
 >;
+
+/**
+ * @public
+ * @experimental
+ */
+export type DotNotationFilter<TSchema> =
```
nbbeeken, line 533: Since the expectation would now be that folks can use this annotation directly (instead of via the find call) Can we add more to the docs here? I'm thinking we can copy some of the language from the migration guide + 4.3.0 release notes w/ a code snippet

nbbeeken, line 533: We should throw around some more ideas for a new name here, if we want to iron this type out more and eventually reintroduce it I wouldn't want folks to go through another rename change, is there something more forward looking? StrictFilter? SmartFilter? MongoDBFilter? We can thread in slack and resolve this comment with team consensus

nbbeeken (changes requested):

src/mongo_types.ts
baileympearson: My thoughts are that the existing documentation and release notes are enough, and that we don't want to provide documentation on an experimental (read: unsupported) feature.

baileympearson: Well, I think that regardless of the name of this type, the actual filter type should (and will always) be named `Filter` because that aligns with every piece of documentation I've ever seen about mongodb.

This is intended to be a temporary workaround, I don't think a rename change is a big deal.

If down the line, we decide we don't want to force users to rename, we can always just alias the current name to the `Filter` type:
```typescript
export type DotNotationFilter<T> = Filter<T>;
```

nbbeeken: We're still going to have [this section](https://www.mongodb.com/docs/drivers/node/current/fundamentals/typescript/#type-safety-and-dot-notation) on the docs page (updated for this change) so we can link to that. In terms of code snippet, it is new for us to have a "feature" that's opt-in TS, I was thinking something small that just demonstrates it being used as an annotation.

We should get clarity on messaging here, we have other features marked experimental that are supported so if unsupported is the plan here we might want a different marker

nbbeeken: > actual filter type should ... be named `Filter`

Yes I agree

> that aligns with every piece of documentation I've ever seen about mongodb

It used to be named FilterQuery in 3.x community types, and query has been interchangeable with filter, but we unified it all across the driver in 4.0.

> we can always just alias the current name to the Filter

Yep although we wouldn't want to maintain a duplicate type forever. I think if we consider a more generic name the type can prove useful as an opt-in TS feature for folks that has more than dot-notation.

I am thinking about how typescript offers increasing safety with each version sometimes introducing new behavior that require code changes. The approach taken by TS is to put these changes behind the strict flag, non-strict TS does not break between compiler versions (to my knowledge). Along that line of thinking, I'd like us to consider `StrictFilter` as this annotation does provide assertions on nested types. In the future, we could consider enhancements that increase the hinting & assertions to improve DX, possibly [taking ideas from here](https://docs.google.com/document/d/1IwjKAiBUXLm9WlpJdw6eSgGPdGp4i-a809vsVK3S6FI/edit#heading=h.5qau76drh2m1).

baileympearson: Are we going to have that docs page?  I assumed that would be removed for 5.x

nbbeeken: If that's the case can you clarify that we want it removed in the "Documentation Changes Summary", I thought it was remaining because it says we're keeping the type (which is true) so I thought the docs stayed too

From slack, `@unsupported - xxx` would be a preferred doc tag where xxx is some language that describes what that means

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+This change caused many problems for users, including slow compilation times and compile errors for
+valid dot notation queries.
+
+Driver 5.0 removes type checking on filter predicates.  The preceding example will compile with
+driver v5.
+
+#### Dot Notation Helper Types Exported
+
+Although we removed support for type checking on dot notation filters by default, we now export these types.
+These helper types can be used for type checking.  We export the `DotNotationUpdateFilter` and the `DotNotationFilter`
```
nbbeeken, line 46:
```suggestion
These helper types can be used for type checking.  We export the `StrictUpdateFilter` and the `StrictFilter`
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+types for type safety in updates and finds.
+
+To use one of the new types, simply create a predicate that uses dot notation and assign it the type of `DotNotationFilter<your schema>`.
```
nbbeeken, line 49:
```suggestion
To use one of the new types, simply create a predicate that uses dot notation and assign it the type of `StrictFilter<your schema>`.
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+interface Schema {
+  user: {
+    name: string
+  }
+}
+
+declare const collection: Collection<Schema>;
+
+// fails to compile, 4 is not assignable to type "string"
+const filterPredicate: DotNotationFilter<Schema> = { 'user.name': 4 };
```
nbbeeken, line 60:
```suggestion
const filterPredicate: StrictFilter<Schema> = { 'user.name': 4 };
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
 
 ## Changes
 
+### Dot Notation Typescript Support Removed By Default
+
+**NOTE** This version only removes our **Typescript** compile-time support for dot notation in queries.
```
nbbeeken, line 21:
```suggestion
**NOTE** This is a **Typescript compile-time only** change. Dot notation in filters sent to MongoDB still work the same. 
```

nbbeeken (changes requested):

nbbeeken (dismissed):

durran (dismissed):

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
 
 ## Changes
 
+### Dot Notation Typescript Support Removed By Default
+
+**NOTE** This is a **Typescript compile-time only** change. Dot notation in filters sent to MongoDB still work the same.
```
dariakp, line 21:
```suggestion
**NOTE** This is a **Typescript compile-time only** change. Dot notation in filters sent to MongoDB will still work the same. 
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+    name: string
+  }
+}
+
+declare const collection: Collection<Schema>;
+// compiles pre-v4.3.0, fails in v4.3.0+
+collection.find({ 'user.name': 4 })
+```
+
+This change caused many problems for users, including slow compilation times and compile errors for
```
dariakp, line 37:
```suggestion
This change caused a number of problems for users, including slow compilation times and compile errors for
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+valid dot notation queries.
```
dariakp, line 38:
```suggestion
valid dot notation queries. While we have tried to mitigate this issue as much as possible
in v4, ultimately we do not believe that this feature is fully production ready for all use cases.
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+
+Driver 5.0 removes type checking on filter predicates.  The preceding example will compile with
```
dariakp, line 40: Not all type checking, right, just dot notation type checking?

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+driver v5.
+
+#### Dot Notation Helper Types Exported
+
+Although we removed support for type checking on dot notation filters by default, we now export these types.
```
dariakp, line 45:
```suggestion
Although we removed support for type checking on dot notation filters by default, we have preserved the
corresponding types in an experimental capacity.
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,53 @@ The following is a detailed collection of the changes in the major v5 release of
...
+}
+
+declare const collection: Collection<Schema>;
+
+// fails to compile, 4 is not assignable to type "string"
+const filterPredicate: StrictFilter<Schema> = { 'user.name': 4 };
+collection.find(filterPredicate);
+```
+
+**NOTE** These types are also now marked experimental and can be changed at any time.
```
dariakp, line 64:
```suggestion
**NOTE** As an experimental feature, these types can change at any time and are not recommended for production settings.
```

dariakp (changes requested):

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,55 @@ The following is a detailed collection of the changes in the major v5 release of
...
+declare const collection: Collection<Schema>;
+// compiles pre-v4.3.0, fails in v4.3.0+
+collection.find({ 'user.name': 4 })
+```
+
+This change caused a number of problems for users, including slow compilation times and compile errors for
+valid dot notation queries. While we have tried to mitigate this issue as much as possible
+in v4, ultimately we do not believe that this feature is fully production ready for all use cases.
+
+Driver 5.0 removes type checking on filter predicates.  The preceding example will compile with
```
baileympearson, line 41:
```suggestion
Driver 5.0 removes type checking for dot notation on filter predicates.  The preceding example will compile with
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -16,6 +16,55 @@ The following is a detailed collection of the changes in the major v5 release of
...
+declare const collection: Collection<Schema>;
+// compiles pre-v4.3.0, fails in v4.3.0+
+collection.find({ 'user.name': 4 })
+```
+
+This change caused a number of problems for users, including slow compilation times and compile errors for
+valid dot notation queries. While we have tried to mitigate this issue as much as possible
+in v4, ultimately we do not believe that this feature is fully production ready for all use cases.
+
+Driver 5.0 removes type checking for dot notation on filter predicates.  The preceding example will compile with
```
baileympearson, line 41:
```suggestion
Driver 5.0 removes type checking for dot notation in filter predicates.  The preceding example will compile with
```

etc/notes/CHANGES_5.0.0.md
baileympearson: Yeah, just dot notation support

dariakp (approved):

dariakp:
@avaly Yes, we are aware - unfortunately there were a lot of unintended consequences in making this the default behavior, which is why we are preserving it in an opt-in capacity and will consider options for integrating it more naturally in the future. We are leaving it marked experimental because we have not had the opportunity to test it as thoroughly as something with such wide-reaching scope would merit. In retrospect, we should have taken the time to do a technical design to fully define the intended boundaries of the functionality - but we were excited and it did seem to "just work". We would like to take a step back now and give this feature the attention it deserves, which we cannot do right this moment. Until we are able to prioritize this work, we are definitely open to feedback from the community to help guide this in the right direction.

avaly:
Thanks @dariakp for the message!

Will the mongodb team accept pull requests regarding these strict filter types? Or would you like to keep those as-is until the team has resources to make those a priority?

dariakp:
@avaly We are open to accepting PRs, with the caveat that as experimental features, triage and review of changes to these types will fall lower on our priority list than bugs or improvements to our officially supported features. However, our ultimate goal with these experimental types is to get them into a form that our users find most helpful, so we will consider all feedback we get: whether through PRs, Jira tickets, or comments.

One thing to keep in mind is that the current `StrictFilter` is intended to be strict on known schema fields, but to allow for unknown `$-`operators (because the driver does not attempt to exhaustively catalogue all available server-side operators), so improvements to that type would need to adhere to that design intent. If there is a use case for restricting the set of allowed operators, we are open to considering a separate experimental type.

avaly:
Thanks again for the honest response.

We (at Plex) have build [Papr](https://github.com/plexinc/papr) with full type-safety in mind from day 0. That's why we opted to submit all these features upstream to the driver. But given these shifted priorities in the driver project, we will copy these types in our library code to be able to move faster (https://github.com/plexinc/papr/issues/410).

dariakp:
@avaly We strive for transparency :) As far as dot notation support: we'll be sure to follow your version of the types and consider them in any future updates to ours. Thank you again for making the original contribution!

avaly:
As mentioned previously, we're adopting these strict types in our Papr library to be able to iterate on them faster and introduce more strictness: https://github.com/plexinc/papr/pull/430

@dariakp Would your team be interesting in getting some of these changes submitted in PRs here?

I know some changes will probably not be accepted (e.g. removing recursive types support), so I won't bother submitting those. But how about the other parts?

dariakp:
@avaly As long as they come with tests and don't run counter to the objectives I mentioned above for `StrictFilter`, we welcome any improvements to functionality :)

PR #3524, opened 2023-01-16
feat(NODE-2014)!: return the result of provided callback in withTransaction and withSession

durran:
### Description

Updates `withTransaction` and `withSession` helpers to return the return value of the user provided callback.

#### What is changing?

- `session.withTransaction()` now properly returns the return value of the callback.
- `client.withSession()` updates its signature to return `Promise<any>` instead of `Promise<void` and returns the return value of the provided callback.
- Updates migration guide.

None

NODE-2014

### DIFF
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -61,7 +61,7 @@ export interface ClientSessionOptions {
 }
 
 /** @public */
-export type WithTransactionCallback<T = void> = (session: ClientSession) => Promise<T>;
+export type WithTransactionCallback<T = any> = (session: ClientSession) => Promise<T>;
 
 /** @public */
 export type ClientSessionEvents = {
@@ -470,10 +470,7 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    * @param options - optional settings for the transaction
    * @returns A raw command response or undefined
    */
-  withTransaction<T = void>(
-    fn: WithTransactionCallback<T>,
-    options?: TransactionOptions
-  ): Promise<Document | undefined> {
+  withTransaction<T>(fn: WithTransactionCallback<T>, options?: TransactionOptions): Promise<T> {
     const startTime = now();
     return attemptTransaction(this, startTime, fn, options);
   }
@@ -615,12 +612,14 @@ function attemptTransaction<TSchema>(
   }
 
   return promise.then(
-    () => {
+    value => {
       if (userExplicitlyEndedTransaction(session)) {
-        return;
+        return value;
       }
 
-      return attemptTransactionCommit(session, startTime, fn, options);
+      return attemptTransactionCommit(session, startTime, fn, options).then(() => {
+        return value;
+      });
     },
     err => {
       function maybeRetryOrThrow(err: MongoError): Promise<any> {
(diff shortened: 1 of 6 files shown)


src/collection.ts
```diff
@@ -86,7 +86,6 @@ import {
   Callback,
   checkCollectionName,
   DEFAULT_PK_FACTORY,
-  emitWarningOnce,
```
durran, line 89: This has nothing to do with this ticket specifically, but was failing the lint task on CI.

src/sessions.ts
```diff
@@ -470,10 +470,7 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    * @param options - optional settings for the transaction
    * @returns A raw command response or undefined
    */
-  withTransaction<T = void>(
-    fn: WithTransactionCallback<T>,
-    options?: TransactionOptions
-  ): Promise<Document | undefined> {
+  withTransaction<T>(fn: WithTransactionCallback<T>, options?: TransactionOptions): Promise<T> {
```
nbbeeken, line 473: The docs need a little update too:
```
   * @remarks
   * This function:
   * - Will return the result of the withTransaction callback if every operation is successful
   * - Will return `undefined` if the transaction is intentionally aborted with `await session.abortTransaction()`
   * - Will throw if one of the operations throws or `throw` statement is used inside the `withTransaction` callback
```

Which makes me question, should we require that the callback return something truthy (via TS) or default to something truthy (maybe `true`) if the callback returns `undefined`? Otherwise, how can someone differentiate an intentionally aborted transaction and a committed one if they don't have a return value?

[Spec ref](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst#the-callback-function-has-a-flexible-signature)

nbbeeken (changes requested):
Actual changes LGTM, but outstanding question about our approach here

src/sessions.ts
durran: I think returning `true` sounds good if they haven't returned anything.

nbbeeken: Sounds good to me, I'm thinking we can just add `value ?? true` and a test for that case along with docs. I'll update the ticket AC just so we keep track

src/sessions.ts
```diff
@@ -615,12 +612,14 @@ function attemptTransaction<TSchema>(
...
 
   return promise.then(
-    () => {
+    value => {
       if (userExplicitlyEndedTransaction(session)) {
-        return;
+        return value;
       }
 
-      return attemptTransactionCommit(session, startTime, fn, options);
```
addaleax, line 623: Will it still be possible to get the legacy return value somehow (i.e. the `Document` returned by `session.commitTransaction()`)?

I think it would be an acceptable breaking change in mongosh if that goes away (right now our own `withTransaction` helper returns it), but it might be nice to avoid this. 

To avoid this, you could do some hackery to attach this secondary result to the returned Promise itself or return a tuple, e.g.
- `withTransaction<T>(fn: WithTransactionCallback<T>, options?: …): Promise<T> & { transactionResult: Promise<Document | undefined> }`
- `withTransaction<T>(fn: WithTransactionCallback<T>, options?: …): Promise<[callbackResult: T, transactionResult: Document|undefined]>`

(that would also maybe solve the problem mentioned above about distinguishing the callback returning `undefined` from an aborted transaction?)

durran: My only worry here is with this change `withTransaction` and `withSession` now diverge, and one of the goals of the ticket was to keep those two APIs the same for consistency. But I'm flexible.

addaleax: Maybe it could just be a separate function or item in `options` then that makes it return a tuple?

nbbeeken: We decided to proceed with returning the user's callback return value and not provide a way to get the old behavior

nbbeeken:
From slack: We're moving this effort down the road to the next major to get ahead of documentation updates that we would want to have ready before committing to this change.

PR #3525, opened 2023-01-17
feat(NODE-4756)!: ok 1 with write concern failure event changes

durran:
### Description

For commands with write concern errors but an `{ ok: 1 }` in the command reply, the driver now emits a command succeeded event.

#### What is changing?
- Command monitoring now emits a `CommandSucceededEvent` when a reply has `{ ok: 1 }` with write concern errors.
- Added a second test in the retryable writes prose tests that confirms the above behaviour.
- Added a unified test for DRIVERS-2468 since that ticket was still in the backlog. This test also confirms the behaviour.

Specs PR: https://github.com/mongodb/specifications/pull/1367

None

NODE-4756/DRIVERS-2468

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -681,7 +681,7 @@ function write(
     operationDescription.started = now();
     operationDescription.cb = (err, reply) => {
       // Command monitoring spec states that if ok is 1, then we must always emit
-      // a command suceeded event, even if there's and error. Write concern errors
+      // a command suceeded event, even if there's an error. Write concern errors
       // will have an ok: 1 in their reply.
       if (err && reply?.ok !== 1) {
         conn.emit(
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -416,7 +416,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
 
       if (operationDescription.command) {
         if (document.writeConcernError) {
-          callback(new MongoWriteConcernError(document.writeConcernError, document));
+          callback(new MongoWriteConcernError(document.writeConcernError, document), document);
           return;
         }
 
@@ -680,7 +680,10 @@ function write(
 
     operationDescription.started = now();
     operationDescription.cb = (err, reply) => {
-      if (err) {
+      // Command monitoring spec states that if ok is 1, then we must always emit
+      // a command suceeded event, even if there's and error. Write concern errors
+      // will have an ok: 1 in their reply.
+      if (err && reply?.ok !== 1) {
         conn.emit(
           Connection.COMMAND_FAILED,
           new CommandFailedEvent(conn, command, err, operationDescription.started)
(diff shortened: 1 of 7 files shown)


durran:
Lint failure unrelated to this PR but addressed in https://github.com/mongodb/node-mongodb-native/pull/3524

src/sdam/server.ts
```diff
@@ -500,7 +500,8 @@ function makeOperationHandler(
 ): Callback {
   const session = options?.session;
   return function handleOperationResult(error, result) {
-    if (result != null) {
+    // We should not swallow an error if it is present.
+    if (error == null && result != null) {
```
durran, line 504: If we are not montoring commands the callback doesn't get wrapped and thus both an error and result are present here.

nbbeeken (changes requested):

test/integration/retryable-writes/retryable_writes.spec.prose.test.ts
```diff
@@ -299,5 +294,54 @@ describe('Retryable Writes Spec Prose', () => {
...
+            writeConcernError: {
+              code: 91,
+              errorLabels: ['RetryableWriteError']
+            },
+            failCommands: ['insert']
+          }
+        });
+
+        client.addListener('commandSucceeded', successListener);
+        client.addListener('commandFailed', failureListener);
```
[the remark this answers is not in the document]
durran: Ah very good idea. I'll update.

test/integration/retryable-writes/retryable_writes.spec.prose.test.ts
```diff
@@ -299,5 +296,48 @@ describe('Retryable Writes Spec Prose', () => {
...
+              errorLabels: ['RetryableWriteError']
+            },
+            failCommands: ['insert']
+          }
+        });
+
+        const willBeCommandSucceeded = once(client, 'commandSucceeded').catch(error => error);
+        const willBeCommandFailed = once(client, 'commandFailed', {
+          signal: AbortSignal.timeout(1000)
+        }).catch(error => error);
```
nbbeeken, line 327: I looked at the docs before suggesting this and it said AbortSignal was available in 14.17+ and we're testing on 14.21 so not sure why it's erroring 😕 


Stealing this from our unified runner: https://github.com/mongodb/node-mongodb-native/blob/main/test/tools/unified-spec-runner/operations.ts#L464-L476
```typescript
const willBeCommandFailed = Promise.race([
  once(client, 'commandFailed'), 
  sleep(1000).then(() => Promise.reject(new Error('timeout'))
]).catch(error => error);
```

nbbeeken (changes requested):

test/integration/retryable-writes/retryable_writes.spec.prose.test.ts
durran: Updated and rebased.

nbbeeken (approved):
LGTM

nbbeeken (changes requested):
just got a lint error, I thought it was the one that was on main for a bit, but that's been fixed

nbbeeken (dismissed):

src/cmap/connection.ts
```diff
@@ -680,7 +680,10 @@ function write(
 
     operationDescription.started = now();
     operationDescription.cb = (err, reply) => {
-      if (err) {
+      // Command monitoring spec states that if ok is 1, then we must always emit
+      // a command suceeded event, even if there's and error. Write concern errors
```
dariakp, line 684:
```suggestion
      // a command suceeded event, even if there's an error. Write concern errors
```

nbbeeken:
note: red CI is unrelated to changes here, I've filed a ticket about the new failures and we know about the FLE CI script issue

PR #3528, opened 2023-01-18
test(NODE-4962): sync range index spec tests

baileympearson:
#### What is changing?

New spec tests for range index support are synced.

No.

### DIFF
--- a/test/integration/client-side-encryption/client_side_encryption.spec.test.ts
+++ b/test/integration/client-side-encryption/client_side_encryption.spec.test.ts
@@ -96,6 +96,11 @@ describe('Client Side Encryption (Legacy)', function () {
 
       return !isSkippedTest;
     }
+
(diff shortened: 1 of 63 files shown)


test/integration/client-side-encryption/client_side_encryption.spec.test.ts
```diff
@@ -75,7 +75,7 @@ describe('Client Side Encryption (Legacy)', function () {
   const testSuites = gatherTestSuites(
     path.join(__dirname, '../../spec/client-side-encryption/tests/legacy'),
     testContext
-  );
+  ).filter(({ name }) => !name.includes('fle2-Range'));
```
baileympearson, line 78: Not our usual method for skipping tests because they do not appear in test output - but because I am about to implement the logic for these tests, it was easier than listing out all tests to skip into an array.

PR #3532, opened 2023-01-23
feat(NODE-4992): Deprecate methods and options that reference legacy logger

W-A-James:
### Description

Deprecating methods and options that are being removed in the driver v5.

#### What is changing?

Deprecating the following methods

- [x] MongoClient.getLogger()
- [x] MongoClient.logger()
- [x] Db.getLogger()
- [x] Db.logger()
- [x] GridFSBucket.getLogger()
- [x] GridFSBucketReadStream.abort()

Deprecating the following options

- [x] MongoClient.logger
- [x] MongoClient.loggerLevel

No, documentation is in the tsdoc comments

Getting rid of old logger cleanly to make way for spec-compliant logger.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -774,7 +774,8 @@ export class Db {
     return new ChangeStream<TSchema, TChange>(this, pipeline, resolveOptions(this, options));
   }
 
-  /** Return the db logger
+  /**
+   * Return the db logger
    * @deprecated The Legacy Logger is deprecated and will be removed in the next major version.
    */
   getLogger(): Logger {
--- a/src/db.ts
+++ b/src/db.ts
@@ -775,12 +775,12 @@ export class Db {
   }
 
   /** Return the db logger
-   * @deprecated The Legacy Logger is deprecated and will be removed in the next major version. See [mongodb-legacy](https://github.com/mongodb-js/nodejs-mongodb-legacy) for migration assistance */
+   * @deprecated The Legacy Logger is deprecated and will be removed in the next major version.*/
   getLogger(): Logger {
     return this.s.logger;
   }
 
-  /** @deprecated The Legacy Logger is deprecated and will be removed in the next major version. See [mongodb-legacy](https://github.com/mongodb-js/nodejs-mongodb-legacy) for migration assistance */
+  /** @deprecated The Legacy Logger is deprecated and will be removed in the next major version.*/
   get logger(): Logger {
     return this.s.logger;
   }
--- a/src/db.ts
+++ b/src/db.ts
@@ -775,12 +775,13 @@ export class Db {
   }
 
   /** Return the db logger
-   * @deprecated The Legacy Logger is deprecated and will be removed in the next major version.*/
+   * @deprecated The Legacy Logger is deprecated and will be removed in the next major version.
+   */
   getLogger(): Logger {
     return this.s.logger;
   }
 
-  /** @deprecated The Legacy Logger is deprecated and will be removed in the next major version.*/
+  /** @deprecated The Legacy Logger is deprecated and will be removed in the next major version. */
   get logger(): Logger {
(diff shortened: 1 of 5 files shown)


src/gridfs/download.ts
```diff
@@ -184,6 +184,8 @@ export class GridFSBucketReadStream extends Readable implements NodeJS.ReadableS
    * the 'close' event once the cursor is successfully killed.
    *
    * @param callback - called when the cursor is successfully closed or an error occurred.
+   *
+   * @deprecated Callbacks are deprecated and will be removed in the next major version. See [mongodb-legacy](https://github.com/mongodb-js/nodejs-mongodb-legacy) for migration assistance
```
dariakp, line 188: you should make this match the way we do it in the upload abort, by first splitting off the overload, and then deprecating just the callback one, because we don't want to deprecate the entire method, just the callback version of it

dariakp (changes requested):

dariakp (changes requested):
This looks good, but let's also deprecate the `LoggerLevel`, `LoggerFunction`, and `LoggerOptions` defined in `logger.ts` - I noticed they weren't deprecated either

src/db.ts
```diff
@@ -774,11 +774,13 @@ export class Db {
     return new ChangeStream<TSchema, TChange>(this, pipeline, resolveOptions(this, options));
   }
 
-  /** Return the db logger */
+  /** Return the db logger
+   * @deprecated The Legacy Logger is deprecated and will be removed in the next major version.*/
```
nbbeeken, line 778:
```suggestion
  /** Return the db logger
   * @deprecated The Legacy Logger is deprecated and will be removed in the next major version.
   */
```

src/db.ts
```diff
@@ -774,11 +774,13 @@ export class Db {
...
   getLogger(): Logger {
     return this.s.logger;
   }
 
+  /** @deprecated The Legacy Logger is deprecated and will be removed in the next major version.*/
```
nbbeeken, line 783:
```suggestion
  /** @deprecated The Legacy Logger is deprecated and will be removed in the next major version. */
```

src/gridfs/index.ts
```diff
@@ -226,7 +226,9 @@ export class GridFSBucket extends TypedEventEmitter<GridFSBucketEvents> {
     }, callback);
   }
 
-  /** Get the Db scoped logger. */
+  /** Get the Db scoped logger.
+   * @deprecated Legacy Logger is deprecated and will be removed in the next major version.
+   */
```
nbbeeken, line 231:
```suggestion
  /** 
   * Get the Db scoped logger.
   * @deprecated Legacy Logger is deprecated and will be removed in the next major version.
   */
```

src/logger.ts
```diff
@@ -27,13 +28,16 @@ export const LoggerLevel = Object.freeze({
   debug: 'debug'
 } as const);
 
-/** @public */
+/** @public
+ * @deprecated The Legacy Logger is deprecated and will be removed in the next major version.*/
```
nbbeeken, line 32:
```suggestion
/** 
 * @public
 * @deprecated The Legacy Logger is deprecated and will be removed in the next major version.
 */
```

nbbeeken: I won't comment on them all but the public tag does not get picked up if its on the same line as the `**`

nbbeeken (changes requested):

src/logger.ts
nbbeeken: The comments have to either be all one line, or with a line break after the `**`

dariakp: Good catch

dariakp (approved):

dariakp (approved):

nbbeeken (approved):

PR #3533, opened 2023-01-23
allow setting ClientSession to null in TypeScript types

vkarpov15:
### Description

Quick suggestion. Setting `session` option to `null` seems to work fine, but TypeScript types don't allow that. Re: https://github.com/Automattic/mongoose/pull/12933


durran:
Hi @vkarpov15 - thanks for the suggestion. We had a discussion about this and it seems if we were going to treat any optional property as acceptable with both `null` or `undefined` we'd need to make it consistent across all options in the driver. For that reason we are inclined to say the Mongoose itself should just adhere to the value here accepting only `undefined` and not permitting `null`.

PR #3538, opened 2023-01-24
fix(NODE-4998): use ipv4 in socks5 tests for node 18

nbbeeken:
#### What is changing?

* Changes the proxy host addresses to be IPv4 to avoid node 18 dns issue
* Splits socks5 driver only run from socks5 csfle run

Green ci

### DIFF
--- a/.evergreen/config.in.yml
+++ b/.evergreen/config.in.yml
@@ -434,6 +434,7 @@ functions:
           export DRIVERS_TOOLS="${DRIVERS_TOOLS}"
           export NODE_LTS_NAME='${NODE_LTS_NAME}'
           export MONGODB_URI="${MONGODB_URI}"
+          export TEST_SOCKS5_CSFLE="${TEST_SOCKS5_CSFLE}"
           export SSL="${SSL}"
 
           # Disable xtrace (just in case it was accidentally set).
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -398,6 +398,8 @@ functions:
 
           export MONGODB_URI="${MONGODB_URI}"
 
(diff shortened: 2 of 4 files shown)


durran (approved):
NODE-4897 addresses the timeout on the release test and the other failure is NODE-4646 flaky test ticket.

baileympearson (changes requested):
Currently, the two tasks are

- socks5 tests only
- socks5 tests and csfle + socks5 tests

We want to break them apart entirely, so 

- one task for socks5 driver tests
- one task for csfle tests w/ socks5 enabled

durran:
> Currently, the two tasks are
> 
>     * socks5 tests only
> 
>     * socks5 tests and csfle + socks5 tests
> 
> 
> We want to break them apart entirely, so
> 
>     * one task for socks5 driver tests
> 
>     * one task for csfle tests w/ socks5 enabled

Ah good catch.

baileympearson (approved):
optional change - we could break apart the auth tests too, and have four total tasks here

- socks5 no auth
- socks5 auth
- csfle socks5 no auth
- csfle socks5 auth

I'm partial to this because I when evergreen shows the results of the test run (which it doesn't if you have multiple test runs in a task), but this is an optional change

durran (approved):

durran:
> optional change - we could break apart the auth tests too, and have four total tasks here
> 
>     * socks5 no auth
> 
>     * socks5 auth
> 
>     * csfle socks5 no auth
> 
>     * csfle socks5 auth
> 
> 
> I'm partial to this because I when evergreen shows the results of the test run (which it doesn't if you have multiple test runs in a task), but this is an optional change

Maybe a new ticket for that effort?

PR #3539, opened 2023-01-24
test(NODE-4974): backport range index tests to 4x

baileympearson:
#### What is changing?

This PR backports tests for range index support to the 4x branch.

Changes should be easy to review, the commits were cherry picked straight from main.

No.


durran (changes requested):
I think we need to change the FLE tests to not run on `LOWEST_LTS` as Node 12 does not support importing 'fs/promises'

baileympearson:
@durran Instead of bumping the node version, I changed the import.  We bumped to Node14 as a part of dropping support for Node12, which won't be a part of the 4.x branch.

durran:
> @durran Instead of bumping the node version, I changed the import. We bumped to Node14 as a part of dropping support for Node12, which won't be a part of the 4.x branch.

Better solution. Thanks!

durran (approved):

nbbeeken (approved):

PR #3540, opened 2023-01-25
docs(NODE-4916, NODE-4523): update readme and upgrade guide with callback removal information

baileympearson:
#### What is changing?

- References to callbacks have been removed from the driver readme.
- Migration examples have been added to the migration guide for callbacks

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,11 +136,6 @@ operations using the MongoDB driver.
 
 Add code to connect to the server and the database **myProject**:
 
-> **NOTE:** All the examples below use async/await syntax.
->
-> However, all async API calls support an optional callback as the final argument,
-> if a callback is provided a Promise will not be returned.
-
 > **NOTE:** Resolving DNS Connection issues
 >
 > Node.js 18 changed the default DNS resolution ordering from always prioritizing ipv4 to the ordering
@@ -267,7 +262,7 @@ It is our recommendation to use `instanceof` checks on errors and to avoid relyi
 We guarantee `instanceof` checks will pass according to semver guidelines, but errors may be sub-classed or their messages may change at any time, even patch releases, as we see fit to increase the helpfulness of the errors.
 
 Any new errors we add to the driver will directly extend an existing error class and no existing error will be moved to a different parent class outside of a major release.
-This means `instanceof` will always be able to accurately capture the errors that our driver throws (or returns in a callback).
+This means `instanceof` will always be able to accurately capture the errors that our driver throws.
 
 ```typescript
 const client = new MongoClient(url);
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 The official [MongoDB](https://www.mongodb.com/) driver for Node.js.
 
+**Upgrading to version 5? Take a look at our [upgrade guide here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_5.0.0.md)!**
+
 **Upgrading to version 4? Take a look at our [upgrade guide here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_4.0.0.md)!**
 
 ## Quick Links
@@ -15,6 +17,7 @@ The official [MongoDB](https://www.mongodb.com/) driver for Node.js.
 | mongodb       | [www.mongodb.com](https://www.mongodb.com)                                                              |
 | changelog     | [HISTORY.md](https://github.com/mongodb/node-mongodb-native/blob/HEAD/HISTORY.md)                       |
 | upgrade to v4 | [etc/notes/CHANGES_4.0.0.md](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_4.0.0.md) |
+| upgrade to v5 | [etc/notes/CHANGES_5.0.0.md](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_5.0.0.md) |
 | contributing  | [CONTRIBUTING.md](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md)             |
 
 ### Bugs / Feature Requests
@@ -53,7 +56,7 @@ If you run into any unexpected compiler failures against our supported TypeScrip
 
 ## Installation
 
-The recommended way to get started using the Node.js 4.x driver is by using the `npm` (Node Package Manager) to install the dependency in your project.
+The recommended way to get started using the Node.js 5.x driver is by using the `npm` (Node Package Manager) to install the dependency in your project.
 
 After you've created your own project using `npm init`, you can run:
 
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@ The official [MongoDB](https://www.mongodb.com/) driver for Node.js.
 
 **Upgrading to version 5? Take a look at our [upgrade guide here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_5.0.0.md)!**
 
-**Upgrading to version 4? Take a look at our [upgrade guide here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_4.0.0.md)!**
-
 ## Quick Links
 
 | what          | where                                                                                                   |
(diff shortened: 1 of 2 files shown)


etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,79 @@ The following is a detailed collection of the changes in the major v5 release of
 
 ### Optional callback support migrated to `mongodb-legacy`
 
+Node v5 drops support for callbacks in favor of a pure async-await API.  Below are some strategies for
+callback users to adopt driver v5 in recommendation order.
+
+### Migrate to promise based api (recommended!)
+
+The Node team strongly encourages anyone who is able to migrate from callbacks to promises to do so.  The
```
dariakp, line 26: I think we should mention *why* that's a better experience for users (and/or mention why we are doing async/await in the heading above: better types, better stack traces, etc)

etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,79 @@ The following is a detailed collection of the changes in the major v5 release of
 
 ### Optional callback support migrated to `mongodb-legacy`
 
+Node v5 drops support for callbacks in favor of a pure async-await API.  Below are some strategies for
+callback users to adopt driver v5 in recommendation order.
+
+### Migrate to promise based api (recommended!)
```
dariakp, line 24: Should this and the other subheadings in this section use the `####` instead of `###`?

etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,79 @@ The following is a detailed collection of the changes in the major v5 release of
...
+  const result = await collection.findOne({ name: 'john snow'});
+  doSomethingWithResult(result);
+} catch (error) {
+  doSomethingWithError(error);
+}
+```
+
+### Use the promise-based API and `util.callbackify`
+
+If you are unable to adopt the promise API fully, we recommend using the promise API and Nodejs's `callbackify`
```
dariakp, line 62:
```suggestion
If you only have a few callback instances where you are currently unable to adopt the promise API, we recommend using the promise API and Nodejs's `callbackify`
```

etc/notes/CHANGES_5.0.0.md
```diff
@@ -26,6 +99,10 @@ We have migrated it to a new package:
 
 The package wraps all of the driver's asynchronous operations that previously supported both promises and callbacks. All the wrapped APIs offer callback support via an optional callback argument alongside a Promise return value so projects with mixed usage will continue to work.
 
+`mongodb-legacy` is intended to preserve driver v4 behavior to enable an smoother transition between
```
dariakp, line 102:
```suggestion
`mongodb-legacy` is intended to preserve driver v4 behavior to enable a smoother transition between
```

dariakp (changes requested):

etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,84 @@ The following is a detailed collection of the changes in the major v5 release of
...
+  doSomethingWithError(error);
+}
+```
+
+#### Use the promise-based API and `util.callbackify`
+
+If you only have a few callback instances where you are currently unable to adopt the promise API, we recommend using the promise API and Nodejs's `callbackify`
+utility to adapt the promise-based API to use callbacks.
+
+**Note** Manually converting a promise-based api to a callback-based API is error prone.  We strongly encourage the user of `callbackify`.
```
W-A-James, line 70:
```suggestion
**Note** Manually converting a promise-based api to a callback-based API is error prone.  We strongly encourage the use of `callbackify`.
```

W-A-James (changes requested):
LGTM apart from one typo

README.md
```diff
@@ -2,6 +2,8 @@
 
 The official [MongoDB](https://www.mongodb.com/) driver for Node.js.
 
+**Upgrading to version 5? Take a look at our [upgrade guide here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_5.0.0.md)!**
+
 **Upgrading to version 4? Take a look at our [upgrade guide here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_4.0.0.md)!**
```
nbbeeken, line 7: We can remove this now, and same with the table entry. Or if we want to surface it still perhaps a section more towards the bottom of the readme that also mentions the 3.x version of that doc? (An "upgrading?" section)

etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,84 @@ The following is a detailed collection of the changes in the major v5 release of
 
 ### Optional callback support migrated to `mongodb-legacy`
 
+Node v5 drops support for callbacks in favor of a pure async-await API.  Below are some strategies for
```
nbbeeken, line 21:
```suggestion
Node v5 drops support for callbacks in favor of an async-await API.  Below are some strategies for
```

I don't think this has general meaning externally, its terminology we invented to differentiate using the syntax and making promises ourselves. It might also be more accurate to change "favor of an async-await API" to "favor of a Promise only API"

etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,84 @@ The following is a detailed collection of the changes in the major v5 release of
...
+callback users to adopt driver v5 in recommendation order.
```
nbbeeken, line 22:
```suggestion
callback users to adopt driver v5 in most to least recommended order.
```

This is optional, the phrasing just caught me up on first read.

etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,84 @@ The following is a detailed collection of the changes in the major v5 release of
...
+The promise-based API is identical to the callback API except:
+
+- no callback is accepted as the last argument
+- a promise is always returned
+
+For example, with a findOne query:
+
+```typescript
+// callback-based API
+collection.findOne({ name: 'john snow'}, (error, result) => {
```
nbbeeken, line 41: Can you run prettier on this readme

etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,84 @@ The following is a detailed collection of the changes in the major v5 release of
...
+  }
+
+  doSomethingWithResult(result);
+})
+
+// promise-based API
+collection.findOne({ name: 'john snow'}).then(
+  (result) => doSomethingWithResult(result),
+  (error) => doSomethingWithError(error)
+);
```
nbbeeken, line 54: Do we want to be recommending this manual promise wrapping? If `async` `await` isn't available, `.then`/`.catch` is what _most_ people should use attaching to an existing promise chain (possibly by just returning it: `return c.findOne()`) 

If folks need to inline this code inside of a place that is callback based they should prefer callbackify because of how it correctly does the error handling for promise rejection

etc/notes/CHANGES_5.0.0.md
```diff
@@ -18,6 +18,84 @@ The following is a detailed collection of the changes in the major v5 release of
...
+#### Use the promise-based API and `util.callbackify`
+
+If you only have a few callback instances where you are currently unable to adopt the promise API, we recommend using the promise API and Nodejs's `callbackify`
+utility to adapt the promise-based API to use callbacks.
+
+**Note** Manually converting a promise-based api to a callback-based API is error prone.  We strongly encourage the use of `callbackify`.
+
+There are two approaches to using `callbackify` with a driver api.
+
+First, the driver method can be callbackified.  If this approach is used, care must be taken to bind the
```
nbbeeken, line 74: I think we should make the anonymous function approach first, and potentially consider dropping the bind one. The anonymous one allows folks to continue omitting the options argument, when you use bind I think you need to provide options to make the callback the third argument, which I don't think is worth explaining.

nbbeeken (changes requested):

etc/notes/CHANGES_5.0.0.md
baileympearson: I'm fine removing the `bind` approach.  Although, it does preserve optional arguments.

etc/notes/CHANGES_5.0.0.md
baileympearson:
```suggestion
// promise-based API
collection.findOne({ name: 'john snow'})
  .then((result) => doSomethingWithResult(result))
  .catch((error) => doSomethingWithError(error));
```

is this what you're suggesting?

nbbeeken: Yep! 

Maybe take it a step further and don't invoke handlers: `doSomethingWithResult` and instead leave comments: `/* Do something with result */`. We know how mixing the two paradigms gets tricky so I don't want to have code samples that suggest a way that could be error prone, but I don't feel very strongly at least if we're demonstrating usage with the typical `.catch` handling.

nbbeeken (changes requested):
viz for promise chain thread

nbbeeken (approved):

W-A-James (approved):
LGTM

dariakp (approved):

PR #3546, opened 2023-01-30
ci(NODE-5012): break apart mongosh integration tests

baileympearson:
#### What is changing?

This PR breaks apart the mongosh integration task PR in CI.  

specifically

- the mongosh integration test task is now broken into three categories of tasks.  there is a task that compiles mongosh, there are a number of tasks that test various mongosh packages (these packages depend on the driver), and there is a new manual test task that asserts that there are no new mongosh tasks that we need to consider

No.

More easily digestible output from the mongosh tasks in CI.


.evergreen/install-and-bootstrap-mongosh.sh
```diff
@@ -0,0 +1,33 @@
...
+
+	export REPLACE_PACKAGE="mongodb:${PROJECT_DIRECTORY}/${TARBALL_FILENAME}"
+	npm run replace-package
+
+	npm run bootstrap -- --no-ci
+}
+
+bootstrap_mongosh
+
+npm run compile-ts
```
addaleax, line 29: Just be aware that this creates a greater dependency on the exact scripts in mongosh and you may need to adjust this more frequently :)

.evergreen/generate_mongosh_tasks.js
```diff
@@ -0,0 +1,66 @@
...
+	'logging',
+	'mongosh',
+	'node-runtime-worker-thread',
+	'service-provider-core',
+	'service-provider-server',
+	'shell-api',
+	'shell-evaluator',
+	'snippet-manager',
+	'types',
+	'docker-build-scripts']
```
addaleax, line 25: - Keeping this as a hardcoded list isn’t going to be very maintainable; if you go with this approach, I would at least add a check somewhere that compares this list against the actual list
- There’s only a subset of these packages that is _really_ affected by driver changes (`lerna ls --scope @mongosh/service-provider-server --scope @mongosh/connectivity-tests --include-dependents` in mongosh)

baileympearson: - I did consider this, but I did make the assumption that the package structure of mongosh doesn't change much.  Is that incorrect?  Adding a check would make sense, but I could also just generate them each time the script is built (clone mongosh and lerna list).
- My intention here was to break apart the existing task just by scope.  If you think that the existing task was testing more than was necessary, then we can reconsider 😄

.evergreen/generate_evergreen_tasks.js
```diff
@@ -1,5 +1,6 @@
 const fs = require('fs');
```
baileympearson, line 1: I did accidentally format and commit this file.  If this proves to be a pain to review, I'm happy to revert it.

.evergreen/install-and-bootstrap-mongosh.sh
```diff
@@ -0,0 +1,38 @@
...
+
+npm pack | tee npm-pack.log
+
+npm cache clear --force || true
+npm i -g npm@8.x || true
+
+TARBALL_FILENAME="$(tail -n1 npm-pack.log)"
+
+MONGOSH_DIRECTORY="/tmp/$TASK_ID"
+git clone --depth=10 -b misc-changes-for-node-driver-ci https://github.com/baileympearson/mongosh.git $MONGOSH_DIRECTORY
```
baileympearson, line 32: This script and `run-mongosh-scope-test` pull in from my branch because this PR won't have a green CI until my PR in mongosh merges.  Once it does, I'll update these references.

.evergreen/run-serverless-tests.sh
```diff
@@ -10,7 +10,7 @@ if [ -z ${MONGODB_URI+omitted} ]; then echo "MONGODB_URI is unset" && exit 1; fi
 if [ -z ${SERVERLESS_ATLAS_USER+omitted} ]; then echo "SERVERLESS_ATLAS_USER is unset" && exit 1; fi
 if [ -z ${SERVERLESS_ATLAS_PASSWORD+omitted} ]; then echo "SERVERLESS_ATLAS_PASSWORD is unset" && exit 1; fi
 
-npm install mongodb-client-encryption@"2.4.0-alpha.2"
```
baileympearson, line 13: Unrelated, but necessary for a clean CI now that we've released mongodb-client-encryption@2.4.0.

.evergreen/generate_evergreen_tasks.js
```diff
@@ -1,5 +1,6 @@
...
 const yaml = require('js-yaml');
+const { mongoshTasks } = require('./generate_mongosh_tasks');
 
 const MONGODB_VERSIONS = ['latest', 'rapid', '6.0', '5.0', '4.4', '4.2', '4.0', '3.6'];
```
durran, line 5: Since the mongosh tasks were broken out into their own generation script, would it make sense to do the same with the versioning code? This way it could be imported here and in the mongosh script, and then for example `LOWEST_LTS` or `LATEST_LTS` could be used in both places, and only need to ever be updated in 1 file and not 2.

.evergreen/run-mongosh-integration-tests.sh
```diff
@@ -12,9 +26,13 @@ npm pack | tee npm-pack.log
...
 
-export TARBALL_FILENAME="$(tail -n1 npm-pack.log)"
-cd /tmp
-git clone --depth=10 https://github.com/mongodb-js/mongosh.git
-cd mongosh
-export REPLACE_PACKAGE="mongodb:${PROJECT_DIRECTORY}/${TARBALL_FILENAME}"
+TARBALL_FILENAME="$(tail -n1 npm-pack.log)"
+
+MONGOSH_DIRECTORY="/tmp/$TASK_ID"
+git clone --depth=10 -b misc-changes-for-node-driver-ci https://github.com/baileympearson/mongosh.git $MONGOSH_DIRECTORY
```
durran, line 32: Just a mental note that this line will need to change once the mongosh PR merges.

.evergreen/run-mongosh-scope-test.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
+
+if [ -z ${TASK_ID+omitted} ]; then echo "TASK_ID is unset" && exit 1; fi
+
+source "${PROJECT_DIRECTORY}/.evergreen/init-nvm.sh"
+
+MONGOSH_DIRECTORY="/tmp/$TASK_ID"
+git clone --depth=10 -b misc-changes-for-node-driver-ci https://github.com/baileympearson/mongosh.git $MONGOSH_DIRECTORY
```
durran, line 8: Same note as other to mark it will change.

test/readme.md
```diff
@@ -419,6 +419,41 @@ The following steps will walk you through how to run the tests for CSFLE.
...
+##### Point mongosh to the driver
+
+mongosh contains a script that does this.  To use the script, create an environment
+ variable `REPLACE_PACKAGE` that contains a string in the form
+`mongodb:<path to your local instance of the driver>`.  The package replacement script will replace
+all occurrences of `mongodb` with the local path of your driver.
+
+An alternative, which can be useful for
+testing a release, is to first run `npm pack` on the driver.  This generates a tarball containing all the code
+would be uploaded to npm if it were released.  Then set the environment variable `REPLACE_PACKAGE`
```
durran, line 440:
```suggestion
that would be uploaded to npm if it were released.  Then set the environment variable `REPLACE_PACKAGE`
```

durran (changes requested):

.evergreen/generate_evergreen_tasks.js
baileympearson: Yeah, I can do that.  Were you thinking all the constants?

```javascript
const MONGODB_VERSIONS = ['latest', 'rapid', '6.0', '5.0', '4.4', '4.2', '4.0', '3.6'];
const versions = [
  { codeName: 'fermium', versionNumber: 14 },
  { codeName: 'gallium', versionNumber: 16 },
  { codeName: 'hydrogen', versionNumber: 18 }
];
const NODE_VERSIONS = versions.map(({ codeName }) => codeName)
NODE_VERSIONS.sort();
const LOWEST_LTS = NODE_VERSIONS[0];
const LATEST_LTS = NODE_VERSIONS[NODE_VERSIONS.length - 1];

const TOPOLOGIES = ['server', 'replica_set', 'sharded_cluster'];
const AWS_AUTH_VERSIONS = ['latest', '6.0', '5.0', '4.4'];
const TLS_VERSIONS = ['latest', '6.0', '5.0', '4.4', '4.2'];

const DEFAULT_OS = 'rhel80-large';
```

.evergreen/generate_mongosh_tasks.js
```diff
@@ -0,0 +1,64 @@
+const scopes = [
+  'service-provider-server',
+  'browser-runtime-electron',
+  'cli-repl',
+  'node-runtime-worker-thread',
+  'mongosh',
+  'browser-repl'
```
addaleax, line 7:
```suggestion
  'browser-repl',
  'connectivity-tests'
```

test/manual/mongosh_scopes.test.ts
```diff
@@ -0,0 +1,24 @@
+import { expect } from 'chai';
+
+const expectedMongoshScopes = [
```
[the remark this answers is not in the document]
addaleax: I think this is an okay approach, we can go with it and see if this becomes too annoying at some point.

I’ll say though, I’d recommend using “package” instead of “scope”. The latter is kind of a lerna-specific term and, in particular, nobody is going to know that it refers to packages if there is no context to clarify it.

.evergreen/generate_evergreen_tasks.js
baileympearson: Done.

.evergreen/run-mongosh-integration-tests.sh
baileympearson: fixed! mongosh PR merged.

baileympearson:
@durran I also backported Neal's windows build fix into this PR.  I can break it out into a separate PR if you would prefer

durran (dismissed):

durran (approved):

PR #3547, opened 2023-01-30
docs(NODE-4990): comment about serverNonce format

nbbeeken:
#### What is changing?

Add a comment for a confusing bit of code

### DIFF
--- a/src/cmap/auth/mongodb_aws.ts
+++ b/src/cmap/auth/mongodb_aws.ts
@@ -108,8 +108,8 @@ export class MongoDBAWS extends AuthProvider {
           return;
         }
 
-        // TODO(NODE-4990)
         if (!ByteUtils.equals(serverNonce.subarray(0, nonce.byteLength), nonce)) {
+          // The serverNonce's leading 32 bytes must equal the client nonce's 32 bytes
           // TODO(NODE-3483)
           callback(new MongoRuntimeError('Server nonce does not begin with client nonce'));
           return;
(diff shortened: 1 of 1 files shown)


src/cmap/auth/mongodb_aws.ts
```diff
@@ -108,8 +108,8 @@ export class MongoDBAWS extends AuthProvider {
           return;
         }
 
-        // TODO(NODE-4990)
         if (!ByteUtils.equals(serverNonce.subarray(0, nonce.byteLength), nonce)) {
+          // The serverNonce's leading 32 bytes must equal the client nonce's 32 bytes
```
dariakp, line 112: should this comment be above this block? and also, I thought you were going to at least mention the spec in case someone else is wondering where this comes from

dariakp (changes requested):

src/cmap/auth/mongodb_aws.ts
nbbeeken: I usually prefer to put comments inside the `if`, just a habit, trying to make the code read like "if X is true, do Y" so if I'm following my own practice I should write something more like: `// throw because the server's nonce must begin with...` kinda like how mocha test the "it" is supposed to imply "the function under test should..."

will add a link to the spec

nbbeeken: https://github.com/mongodb/specifications/blob/875446db44aade414011731840831f38a6c668df/source/auth/auth.rst#id11

I used the latest commit hash incase a future version changes the header link

dariakp (approved):

PR #3553, opened 2023-02-02
test(NODE-5028): assert MongoClients are garbage collectable

nbbeeken:
### Description

Using both self process memory reporting and heapsnapshots, we can check that after 100 clients have been created and closed they are removed from memory by the GC.


test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as path from 'node:path';
+
+import { expect } from 'chai';
+import { parseSnapshot } from 'v8-heapsnapshot';
+
+const HEAPSNAPSHOT_BEFORE = './before.heapsnapshot.json';
+const HEAPSNAPSHOT_AFTER = './after.heapsnapshot.json';
+
+const REPRO_SCRIPT = uri => `
+const { MongoClient } = require('${path.resolve(__dirname, '../../../lib')}');
```
addaleax, line 13:
```suggestion
const { MongoClient } = require(${JSON.stringify(path.resolve(__dirname, '../../../lib'))});
```

(would always do this when building template strings containing JS codes, handles all the edge cases like "what if `__dirname` includes a forward slash" which is extremely unlikely here, but, you know 🙂 )

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,108 @@
...
+const process = require('node:process');
+const v8 = require('node:v8');
+
+async function run(i) {
+  const mongoClient = new MongoClient(
+    '${uri}',
```
addaleax, line 19:
```suggestion
    ${JSON.stringify(uri)},
```

test/integration/node-specific/resource_clean_up.test.ts
nbbeeken: Good call, I was thinking this or inspect, but that would just be for getting single quotes

baileympearson (changes requested):
I think we can use [AsyncResources](https://nodejs.org/api/async_context.html#class-asyncresource) for this.

If we attach an async resource to each mongo client and enable async hooks, then we can track when the async resources (and by proxy, the MongoClient) is actually garbage collected without relying on measuring the heap size.

baileympearson (changes requested):
CI is pretty red, a number of failures are unrelated (the fle failures on node18 appear on other PRs, I wonder if Node18.14 broke that test)

Should this test be a part of the revert PRs maybe?  Otherwise, we can't demonstrate with a test that the revert fixes the memory lead issue, and we can't merge this first because it will fail without the revert.

Also, the async resource tracking always records exactly one client (even with the revert).  If we keep the async hooks logic, we'll need to update the assertion (it's currently asserting against 0).  But should we keep that logic if we don't know why exactly one mongo client is always reported as being left?

nbbeeken:
We should merge the reverts in isolation for organizational sake, and then we can merge this when it's able to be green. I can post an evg patch showing this running against the driver with the revert

nbbeeken:
https://spruce.mongodb.com/version/63dd2b6ad6d80a7f9d1b7c21

nbbeeken:
We can't be sure that the tracking is accurate because as the test shows it is still showing one MongoClient in memory, while the heap snapshot shows none. So which is the true source of truth?

durran:
> We can't be sure that the tracking is accurate because as the test shows it is still showing one MongoClient in memory, while the heap snapshot shows none. So which is the true source of truth?

Yeah that's really odd and has the feeling it could be a flaky test in the future as well as checking the difference between start and end memory... Maybe we only need the test on line 137 and that's enough?

nbbeeken:
I do permit a wiggle room of 3 megabytes which is pretty friendly but that was just a guesstimate, we could make the range much larger to avoid flakes. The leak was growing unbounded easily reaching 90+ MB with 100 iterations of client.connect/close so we could do a range of 20MB? it just depends on how much allocation would be beyond our expected amount (relative to startup amount)

durran:
> I do permit a wiggle room of 3 megabytes which is pretty friendly but that was just a guesstimate, we could make the range much larger to avoid flakes. The leak was growing unbounded easily reaching 90+ MB with 100 iterations of client.connect/close so we could do a range of 20MB? it just depends on how much allocation would be beyond our expected amount (relative to startup amount)

I think that's ok. I guess the only thing I wonder about now based on the Slack thread is do we then need the async resource tracking as it's showing behaviour we don't expect. Can we just remove that portion and rely on the heap alone?

nbbeeken:
Sounds good to me, I bumped the range to 5MB and cleaned up the script, should be more manageable (& hopefully reusable!)

durran (approved):

durran (approved):

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * This 5MB range is selected arbitrarily and should likely be raised if failures are seen intermittently.
+ *
+ * The goal here is to catch unbounded memory growth, currently runScript defaults to 100 iterations of whatever test code is passed in
+ * More than a 5MB growth in memory usage after the script has finished and _should have_ cleaned up all its resources likely indicates that
+ * the growth will continue if the script is changed to iterate the the test code more.
+ */
+const MB_PERMITTED_OFFSET = 5;
+
+describe('Driver Resources', () => {
```
dariakp, line 14: I would nest everything in here inside a `context` that explains the leak scenario being tested; this set up is specifically for MongoClient cleanup on close, so that should be clear from the logged test output when it runs

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    );
+  });
+
+  it('heapsnapshot has 0 MongoClients in memory', async () => {
+    const clients = heap.nodes.filter(n => n.name === 'MongoClient' && n.type === 'object');
+    // lengthOf crashes chai b/c it tries to print out a gigantic diff
+    // Allow GC to miss a few
+    expect(
+      clients.length,
+      `expected no MongoClients in the heapsnapshot found ${clients.length}`
```
dariakp, line 87:
```suggestion
      `expected no MongoClients in the heapsnapshot, found ${clients.length}`
```

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    ).to.be.within(
+      startingMemoryUsed - MB_PERMITTED_OFFSET,
+      startingMemoryUsed + MB_PERMITTED_OFFSET
+    );
+  });
+
+  it('heapsnapshot has 0 MongoClients in memory', async () => {
+    const clients = heap.nodes.filter(n => n.name === 'MongoClient' && n.type === 'object');
+    // lengthOf crashes chai b/c it tries to print out a gigantic diff
+    // Allow GC to miss a few
```
dariakp, line 84: what is this comment about?

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      return;
+    }
+    const res = await runScript(
+      'no_resource_leak_connect_close',
+      this.configuration,
+      async function run({ MongoClient, uri }) {
+        const mongoClient = new MongoClient(uri);
+        await mongoClient.connect();
+        const db = mongoClient.db();
+        await Promise.all([
```
dariakp, line 43: why `db.collections()` and why this many times? also, are we interested in ensuring no leak when there is only one client or should we test multiples? (your helper script runs multiple iterations, but I am trying to understand what the actual scenario under test is)

test/integration/node-specific/resource_tracking_script_builder.ts
```diff
@@ -0,0 +1,132 @@
...
+  uri: string;
+  iteration: number;
+}) => Promise<void>;
+
+export const testScriptFactory = (
+  name: string,
+  uri: string,
+  iterations: number,
+  func: ResourceTestFunction
+) => `'use strict';
```
dariakp, line 23: I've been thinking about how to make this script itself easier to maintain, and I wonder if we could factor it out into a stand-alone `.js` file, where the things currently in `${}` scopes are marked with `'CLIENT_PATH'`, `'FUNCTION'` and `'URI'` respectively, then you could read this code in from the corresponding file here, and use a simple `.replace` to insert the `JSON.stringify` and `.toString()` values into the file.

test/integration/node-specific/resource_tracking_script_builder.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const starting = await messages.next();
+  const ending = await messages.next();
+
+  const startingMemoryUsed = starting.value[0].startingMemoryUsed;
+  const endingMemoryUsed = ending.value[0].endingMemoryUsed;
+
+  // make sure the process ended
+  const [exitCode] = await willClose;
+  expect(exitCode).to.equal(0);
```
dariakp, line 118: add a message just in case to provide the error context

test/integration/node-specific/resource_tracking_script_builder.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const heap = await readFile(heapsnapshotFile, { encoding: 'utf8' }).then(c =>
+    parseSnapshot(JSON.parse(c))
+  );
+
+  await unlink(scriptName);
```
dariakp, line 124: this cleanup isn't going to happen if any of those awaits reject

dariakp (changes requested):

test/integration/node-specific/resource_clean_up.test.ts
nbbeeken: `expect(clients).to.have.lengthOf(0)` is typically what we assert, but chai's methodology cannot handle printing that array out if it's got items in it.

test/integration/node-specific/resource_clean_up.test.ts
nbbeeken: The operation is arbitrary, the parallelism is to get "some" number of connections in the connection pool, this bug does reproduce without this though, I can make it one db.collections call

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,90 @@
...
+          db.collections(),
+          db.collections(),
+          db.collections(),
+          db.collections()
+        ]);
+        await mongoClient.close();
+      }
+    ).catch(error => error);
+
+    if (res instanceof Error) {
```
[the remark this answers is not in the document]
nbbeeken: Sounds good, I can do that, the expectation is that the script never exits with failure so this is just to help the engineer get pointed in the right direction if they see this in CI. But pulling it up locally is likely going to be required to figure out the cause.

Instead of assertions in the before hook I could add a test that fails if the process exited with failure, but I think I'd have to define hooks/it blocks inside runScript 🤔

test/integration/node-specific/resource_clean_up.test.ts
dariakp: I mean you are still asserting that the length is 0, so you aren't allowing it to miss a few as far as I see, so maybe updating the line above to `if GC misses a few`

test/integration/node-specific/resource_clean_up.test.ts
dariakp: ok but why db.collections? why isn't the `connect` call enough? if you just want connections in the connection pool, then setting a minpoolsize feels like a more straightforward approach

test/integration/node-specific/resource_tracking_script_builder.ts
dariakp: can you add a comment that explains this?

test/integration/node-specific/resource_clean_up.test.ts
nbbeeken: I minimized the reproduction more, it's a connect and ping, although I guess the connect is redundant with any operation. I tried only setting a minPoolSize, but the leak stopped reproducing so it would seem that running an operation over the connections is required.

@baileympearson FYI ^

test/integration/node-specific/resource_tracking_script_builder.ts
nbbeeken: Broke it out to a file with replace calls, let me know how it looks maintainability wise

dariakp: I think that works, thanks!

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        console.log(`runScript error message: ${error.message}`);
+        console.log(`runScript error stack: ${error.stack}`);
+        console.log(`runScript error cause: ${error.cause}`);
+        // Fail the test
+        this.test?.error(error);
+      }
+    });
+
+    describe('ending memory usage', () => {
+      it(`must within ${MB_PERMITTED_OFFSET}MB of starting amount`, async () => {
```
dariakp, line 65:
```suggestion
      it(`is within ${MB_PERMITTED_OFFSET}MB of starting amount`, async () => {
```

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,89 @@
...
+          endingMemoryUsed,
+          `script started with ${startingMemoryUsed}MB heap but ended with ${endingMemoryUsed}MB heap used`
+        ).to.be.within(
+          startingMemoryUsed - MB_PERMITTED_OFFSET,
+          startingMemoryUsed + MB_PERMITTED_OFFSET
+        );
+      });
+    });
+
+    describe('heap snapshot', () => {
```
dariakp, line 78:
```suggestion
    describe('ending heap snapshot', () => {
```

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      it('heapsnapshot has 0 MongoClients in memory', async () => {
```
dariakp, line 79:
```suggestion
      it('has 0 MongoClients in memory', async () => {
```

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+    if (typeof this.configuration.serverApi === 'string') {
+      if (this.currentTest) {
+        this.currentTest.skipReason = 'runScript does not support serverApi settings';
+      }
+      this.currentTest?.skip();
+    }
+  });
+
+  describe('MongoClient.close() should not leak resources', () => {
```
dariakp, line 32:
```suggestion
  context('on MongoClient.close()', () => {
```

dariakp (changes requested):

test/integration/node-specific/resource_tracking_script_builder.ts
```diff
@@ -0,0 +1,108 @@
...
+  const [exitCode] = await willClose;
+  expect(exitCode, 'process should have exited with zero').to.equal(0);
+
+  const heap = await readFile(heapsnapshotFile, { encoding: 'utf8' }).then(c =>
+    parseSnapshot(JSON.parse(c))
+  );
+
+  // If any of the above throws we won't reach these unlinks that clean up the created files.
+  // This is intentional so that when debugging the file will still be present to check it for errors
+  // await unlink(scriptName);
```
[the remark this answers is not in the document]
nbbeeken: Yes, thank you 😅

dariakp (approved):

PR #3557, opened 2023-02-06
feat(NODE-5034): support OIDC auth options

durran:
### Description

Add support for OIDC connection string and `MongoClient` options.

Reference PR: https://github.com/mongodb/specifications/pull/1365

#### What is changing?
- Syncs all the new auth spec tests and moves them to the legacy folder.
- Adds `PRINCIPAL_NAME`, `DEVICE_NAME`, `REQUEST_TOKEN_CALLBACK` and `REFRESH_TOKEN_CALLBACK` auth mechanism properties and types.
- Handles all the new OIDC option validation.

None

NODE-5034/DRIVERS-2415


src/cmap/auth/mongodb_oidc.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/** @public */
+export interface OIDCRequestFunction {
+  (idl: OIDCMechanismServerStep1): Promise<OIDCRequestTokenResult>;
+}
+
+/** @public */
+export interface OIDCRefreshFunction {
+  (idl: OIDCMechanismServerStep1, result: OIDCRequestTokenResult): Promise<OIDCRequestTokenResult>;
+}
```
nbbeeken, line 26: I think the only reason to do a callable interface is to support overloads which we could always do in the future, but for now I think defining a strict type suffices:
```suggestion
export type OIDCRefreshFunction = (idl: OIDCMechanismServerStep1, result: OIDCRequestTokenResult) => Promise<OIDCRequestTokenResult>;
```

And same for the other callback type.

src/cmap/auth/mongo_credentials.ts
```diff
@@ -32,6 +33,10 @@ export interface AuthMechanismProperties extends Document {
   SERVICE_REALM?: string;
   CANONICALIZE_HOST_NAME?: GSSAPICanonicalizationValue;
   AWS_SESSION_TOKEN?: string;
+  DEVICE_NAME?: string;
+  PRINCIPAL_NAME?: string;
+  REQUEST_TOKEN_CALLBACK?: OIDCRequestFunction;
+  REFRESH_TOKEN_CALLBACK?: OIDCRefreshFunction;
```
nbbeeken, line 39: Perhaps a small bit of documentation would be nice here
```suggestion
  /** A name for the device workflow that is one of "aws", "azure", or "gcp". */
  DEVICE_NAME?: string;
  /** This is required by the server when more than one Identity Provider is configured */
  PRINCIPAL_NAME?: string;
  REQUEST_TOKEN_CALLBACK?: OIDCRequestFunction;
  REFRESH_TOKEN_CALLBACK?: OIDCRefreshFunction;
```

etc. more for our reference than anything. Also do we want to put 'aws' or the other cloud providers as the type for DEVICE_NAME?

src/cmap/auth/mongo_credentials.ts
```diff
@@ -137,6 +142,44 @@ export class MongoCredentials {
       throw new MongoMissingCredentialsError(`Username required for mechanism '${this.mechanism}'`);
     }
 
+    if (this.mechanism === AuthMechanism.MONGODB_OIDC) {
+      if (this.username) {
+        throw new MongoAPIError(
```
nbbeeken, line 147: I think its valid to make these all `MongoInvalidArgumentError` slightly more specific about a misconfiguration issue.

nbbeeken (changes requested):

src/cmap/auth/mongodb_oidc.ts
durran: Changed to strict types.

src/cmap/auth/mongo_credentials.ts
durran: Added comments.

src/cmap/auth/mongo_credentials.ts
durran: Changed all to `MongoInvalidArgumentError`.

nbbeeken (approved):
LGTM

src/cmap/auth/mongo_credentials.ts
```diff
@@ -137,6 +150,44 @@ export class MongoCredentials {
       throw new MongoMissingCredentialsError(`Username required for mechanism '${this.mechanism}'`);
     }
 
+    if (this.mechanism === AuthMechanism.MONGODB_OIDC) {
```
nbbeeken, line 153: Sorry, just thought of something, will we merge this now or hold off until the implementation is ready (I think the latter is annoying process wise). 

Should we throw an error here to prevent this from appearing to be "supported" before it actually is if we ship this logic in v5.1.0?

nbbeeken (changes requested):

src/cmap/auth/mongo_credentials.ts
durran: I tagged the fix version in the tickets for 5.1.0 since I expected it to go in that release. But I guess it would be odd if the actual spec wasn't completed. But this will error if the user tries to use it due to the auth provider not being there yet. See: https://github.com/mongodb/node-mongodb-native/blob/main/src/cmap/connect.ts#L38 and https://github.com/mongodb/node-mongodb-native/blob/main/src/cmap/connect.ts#L111

nbbeeken: Interesting, a runtime error seems good to me, and we can callout in release notes if that it doesn't work yet if we reach 5.1.0 without the functionality

nbbeeken (approved):

src/cmap/auth/providers.ts
```diff
@@ -4,6 +4,7 @@ export const AuthMechanism = Object.freeze({
   MONGODB_CR: 'MONGODB-CR',
   MONGODB_DEFAULT: 'DEFAULT',
   MONGODB_GSSAPI: 'GSSAPI',
+  MONGODB_OIDC: 'MONGODB-OIDC',
```
nbbeeken, line 7: adding `@internal` here should make it not exist in the types (it will secretly be present at runtime 🤫 but I think that's okay)

nbbeeken (changes requested):
Typescript publicness fixes

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -0,0 +1,27 @@
+/** @public */
```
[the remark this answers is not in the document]
durran: Done

src/cmap/auth/mongo_credentials.ts
```diff
@@ -32,6 +37,14 @@ export interface AuthMechanismProperties extends Document {
   SERVICE_REALM?: string;
   CANONICALIZE_HOST_NAME?: GSSAPICanonicalizationValue;
   AWS_SESSION_TOKEN?: string;
+  /** Name for the OIDC device workflow */
+  DEVICE_NAME?: 'aws' | 'azure' | 'gcp';
+  /** Similar to a username, is require by OIDC when more than one IDP is configured. */
```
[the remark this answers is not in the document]
durran: Done

src/cmap/auth/providers.ts
durran: Done.

nbbeeken (approved):

nbbeeken (approved):

PR #3556, opened 2023-02-03
feat(NODE-3445): add support for AssumeRoleWithWebIdentity in aws authentication

baileympearson:
#### What is changing?

Adds testing for AssumeRoleWithIdentity into CI.

##### Is there new documentation needed for these changes?

No.

### Double check the following

- [ ] Ran `npm run check:lint` script
- [ ] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [ ] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`
- [ ] Changes are covered by tests
- [ ] New TODOs have a related JIRA ticket

PR #3566, opened 2023-02-13
feat(NODE-5054): add AssumeRoleWithWebIdentity support to 4x driver

baileympearson:
### Description

This PR backports testing for AWS EKS requests.

No.


.evergreen/config.yml
```diff
@@ -649,6 +654,53 @@ functions:
         script: |
           ${PREPARE_SHELL}
           ${PROJECT_DIRECTORY}/.evergreen/run-mongodb-aws-test.sh
+  run aws auth test AssumeRoleWithWebIdentity with AWS_ROLE_SESSION_NAME set:
```
dariakp, line 657: Noticed that this set of changes doesn't have the addition of the `install aws-credential-providers` script that #3556 did - why the discrepancy?

package.json
```diff
@@ -136,4 +136,4 @@
       "moduleResolution": "node"
     }
   }
-}
+}
```
dariakp, line 139: why was this file touched?

.evergreen/config.yml
baileympearson: In version 4x, the AWS SDK was an optional dependency so it was installed by default. In version five, it is no longer installed by default so we must manually install it.

package.json
baileympearson: Just a bad merge. I'll fix it

PR #3572, opened 2023-02-15
fix(NODE-5064): consolidate connection cleanup logic and ensure socket is always closed

baileympearson:
#### What is changing?

This PR primarily does three things:

- ensures that the message stream is always destroyed and ensure that the socket is always destroyed
- consolidate the connection clean up logic into a single function that is called from each point where the connection is cleaned out
- improves the unit tests to properly assert on the state of the connection after _cleanup has been called. Additionally, the unit tests now in for synchronous versus asynchronous behavior for the various error and clean up scenarios.

Example unit test output:
```
  new Connection()
    ✔ should support fire-and-forget messages
    ✔ should destroy streams which time out (55ms)
    ✔ should throw a network error with kBeforeHandshake set to false on timeout after handshake (56ms)
    ✔ should throw a network error with kBeforeHandshake set to true on timeout before handshake (53ms)
    #onMessage
      when the connection is a monitoring connection
        when multiple hellos exist on the stream
          ✔ calls the callback with the last hello document
        when requestId/responseTo do not match
          ✔ calls the operation description callback with the document
        when requestId/reponseTo match
          ✔ calls the operation description callback with the document
        when no operation description is in the queue
          ✔ does not error
        when more than one operation description is in the queue
          ✔ calls all operation description callbacks with an error
    when the socket times out
      delayed timeout for lambda behavior
        ✔ should delay timeout errors by one tick
        ✔ should clear timeout errors if more data is available
      when the timeout expires and no more data has come in
        ✔ destroys the message stream
        ✔ ends the socket
        ✔ destroys the socket after ending it
        ✔ passes the error along to any callbacks in the operation description queue (asynchronously)
        ✔ emits a Connection.CLOSE event (asynchronously)
        ✔ removes all listeners on the message stream
        ✔ removes all listeners on the socket
    when the message stream errors
      ✔ destroys the message stream synchronously
      ✔ ends the socket
      ✔ destroys the socket after ending it (asynchronously)
      ✔ passes the error along to any callbacks in the operation description queue (asynchronously)
      ✔ emits a Connection.CLOSE event (asynchronously)
      ✔ removes all listeners on the message stream
      ✔ removes all listeners on the socket
    when the underlying socket closes
      ✔ destroys the message stream synchronously
      ✔ ends the socket
      ✔ destroys the socket after ending it (asynchronously)
      ✔ calls any callbacks in the queue with a MongoNetworkError (asynchronously)
      ✔ emits a Connection.CLOSE event (asynchronously)
      ✔ removes all listeners on the message stream
      ✔ removes all listeners on the socket
    .hasSessionSupport
      when logicalSessionTimeoutMinutes is present
        ✔ returns true
      when logicalSessionTimeoutMinutes is not present
        when in load balancing mode
          ✔ returns true
        when not in load balancing mode
          ✔ returns false
    destroy()
      when a callback is provided
        when the connection was already destroyed
          ✔ does not attempt to cleanup the socket again
          ✔ calls the callback (asynchronously)
        when the connection was not destroyed
          ✔ cleans up the connection
          ✔ calls the callback (asynchronously)
      when options.force == true
        ✔ destroys the tcp socket (synchronously)
        ✔ does not call stream.end
        ✔ destroys the messageStream (synchronously)
        ✔ when destroy({ force: true }) is called multiple times, it calls stream.destroy exactly once
      when options.force == false
        ✔ destroys the tcp socket (asynchronously)
        ✔ ends the tcp socket (synchronously)
        ✔ destroys the messageStream (synchronously)
        ✔ calls stream.end exactly once when destroy is called multiple times
```

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -292,8 +292,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this[kLastUseTime] = now();
   }
 
-  _cleanup(options: { force: boolean; errorFactory?: () => Error }) {
-    const { errorFactory, force } = options;
+  _cleanup(force: boolean, error?: Error) {
     if (this.closed) {
       return;
     }
@@ -302,7 +301,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
 
     const completeCleanup = () => {
       for (const op of this[kQueue].values()) {
-        op.cb(errorFactory?.());
+        op.cb(error);
       }
 
       this[kQueue].clear();
@@ -337,12 +336,12 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     if (this.closed) {
       return;
     }
-    this._cleanup({ force: false, errorFactory: () => error });
+    this._cleanup(false, error);
   }
 
   onClose() {
     const message = `connection ${this.id} to ${this.address} closed`;
-    this._cleanup({ force: false, errorFactory: () => new MongoNetworkError(message) });
+    this._cleanup(false, new MongoNetworkError(message));
   }
 
   onTimeout() {
@@ -353,10 +352,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this[kDelayedTimeoutId] = setTimeout(() => {
       const message = `connection ${this.id} to ${this.address} timed out`;
       const beforeHandshake = this.hello == null;
-      this._cleanup({
-        force: false,
-        errorFactory: () => new MongoNetworkTimeoutError(message, { beforeHandshake })
-      });
+      this._cleanup(false, new MongoNetworkTimeoutError(message, { beforeHandshake }));
     }, 1).unref(); // No need for this timer to hold the event loop open
   }
 
@@ -377,10 +373,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
 
       // First check if the map is of invalid size
       if (this[kQueue].size > 1) {
-        this._cleanup({
-          force: true,
-          errorFactory: () => new MongoRuntimeError(INVALID_QUEUE_SIZE)
-        });
+        this._cleanup(true, new MongoRuntimeError(INVALID_QUEUE_SIZE));
       } else {
         // Get the first orphaned operation description.
         const entry = this[kQueue].entries().next();
@@ -466,7 +459,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     }
     this.once('close', () => process.nextTick(() => callback?.()));
     const message = `connection ${this.id} to ${this.address} closed`;
-    this._cleanup({ ...options, errorFactory: () => new MongoNetworkError(message) });
+    this._cleanup(options.force, new MongoNetworkError(message));
   }
 
   command(
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -154,8 +154,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   address: string;
   socketTimeoutMS: number;
   monitorCommands: boolean;
+  /** Indicates that the connection (including underlying TCP socket) has been closed. */
   closed: boolean;
-  destroyed: boolean;
   lastHelloMS?: number;
   serverApi?: ServerApi;
   helloOk?: boolean;
@@ -204,7 +204,6 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this.monitorCommands = options.monitorCommands;
     this.serverApi = options.serverApi;
     this.closed = false;
-    this.destroyed = false;
     this[kHello] = null;
     this[kClusterTime] = null;
 
@@ -297,10 +296,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     if (this.closed) {
       return;
     }
-
-    this[kStream].destroy(error);
-
-    this.closed = true;
+    this.destroy({ force: false });
 
     for (const op of this[kQueue].values()) {
       op.cb(error);
@@ -314,8 +310,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     if (this.closed) {
       return;
     }
-
-    this.closed = true;
+    this.destroy({ force: false });
 
     const message = `connection ${this.id} to ${this.address} closed`;
     for (const op of this[kQueue].values()) {
@@ -332,9 +327,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     }
 
     this[kDelayedTimeoutId] = setTimeout(() => {
-      this[kStream].destroy();
-
-      this.closed = true;
+      this.destroy({ force: false });
 
       const message = `connection ${this.id} to ${this.address} timed out`;
       const beforeHandshake = this.hello == null;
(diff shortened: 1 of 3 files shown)


src/cmap/connection.ts
```diff
@@ -293,37 +292,57 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     this[kLastUseTime] = now();
   }
 
-  onError(error: Error) {
+  _cleanup(options: { force: boolean; errorFactory?: () => Error }) {
```
baileympearson, line 295: errorFactory could be changed, this maintains the old behavior each callback receives a different instance of the error. We could also just passed an error indirectly.

src/cmap/connection.ts
```diff
@@ -444,34 +460,13 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
       return;
     }
-
-    this[kStream].end(() => {
-      this.destroyed = true;
-      if (typeof callback === 'function') {
-        callback();
-      }
-    });
+    this.once('close', () => process.nextTick(() => callback?.()));
```
baileympearson, line 467: I am happy to revert this change. By listening for the close event, the centralized cleanup handler does not need to worry about an optional callback. None of the event handlers rely on it, so it's only necessary in this method.

test/unit/cmap/connection.test.ts
```diff
@@ -362,81 +389,324 @@ describe('new Connection()', function () {
...
 
-        it('calls all operation description callbacks with an error', function () {
+        it('calls all operation description callbacks with an error', function (done) {
           expect(spyOne).to.be.calledOnce;
           expect(spyTwo).to.be.calledOnce;
           const errorOne = spyOne.firstCall.args[0];
           const errorTwo = spyTwo.firstCall.args[0];
           expect(errorOne).to.be.instanceof(MongoRuntimeError);
           expect(errorTwo).to.be.instanceof(MongoRuntimeError);
+          done();
```
addaleax, line 399: Is this change intentional?

src/cmap/connection.ts
addaleax: Yeah, it’s not immediately obvious to me why this isn’t `error: Error` instead of `errorFactory`… is it for the stack trace? That would make sense but is probably something that should be noted in a comment

addaleax (approved):

src/cmap/connection.ts
baileympearson: This technically preserves the existing behavior where each callback receives a distinct instance of the error. I think it's a reasonable argument to change this behavior (and good point about to call stack) so if we're OK passing the same instance to each call back, then i'll make that change.

_Technically_ there should only be one call back in there ever though.

src/cmap/connection.ts
```diff
@@ -293,37 +292,57 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+        op.cb(errorFactory?.());
+      }
+
+      this[kQueue].clear();
+
+      this.emit(Connection.CLOSE);
+    };
+
+    this.removeAllListeners(Connection.PINNED);
+    this.removeAllListeners(Connection.UNPINNED);
```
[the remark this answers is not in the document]
baileympearson: We need to leave the close listener on but that's it. I can remove them all if you'd like

src/cmap/connection.ts
nbbeeken: I was thinking we could make cleanup an async function and we can invoke the callback from a `.then`. If that's reasonable let me know. Otherwise, we can keep this but add a gate: `if (callback) this.once ///...`

src/cmap/connection.ts
addaleax: Right, if object identity for the `Error` objects is the concern here, I’m okay with passing the same `Error` object to multiple callbacks.

test/unit/cmap/connection.test.ts
```diff
@@ -362,81 +389,324 @@ describe('new Connection()', function () {
...
+        cleanupSpy = sinon.spy(connection, '_cleanup');
+        timeoutSpy = sinon.spy(connection, 'onTimeout');
+      });
+
+      afterEach(() => {
+        sinon.restore();
+        timerSandbox.restore();
+        clock.restore();
+        sinon.restore();
+      });
```
nbbeeken, line 466: dupe clean up, double sinon.restore and overlap with the above afterEach, I assume this is left from debuggin and not intentional?

test/unit/cmap/connection.test.ts
```diff
@@ -362,81 +389,324 @@ describe('new Connection()', function () {
...
+        expect(messageStream.eventNames()).to.deep.equal([]);
+      });
+
+      it('removes all listeners on the socket', () => {
+        expect(driverSocket.eventNames()).to.deep.equal([]);
+      });
+    });
+  });
+
+  describe('when the message stream errors', () => {
```
nbbeeken, line 548:
```suggestion
  describe('when the MessageStream errors', () => {
```

test/unit/cmap/connection.test.ts
```diff
@@ -362,81 +389,324 @@ describe('new Connection()', function () {
...
+      expect(closeCount).to.equal(0);
+      clock.runAll();
+      expect(closeCount).to.equal(1);
+    });
 
-      driverSocket.emit('timeout');
-      expect(connection.onTimeout).to.have.been.calledOnce;
-      expect(connection).to.have.property(kDelayedTimeoutId).that.is.instanceOf(NodeJSTimeoutClass);
+    it('removes all listeners on the message stream', () => {
+      expect(messageStream.eventNames()).to.deep.equal([]);
```
nbbeeken, line 622: minor: but listenerCount instead of deep equal empty array?

test/unit/cmap/connection.test.ts
```diff
@@ -362,81 +389,324 @@ describe('new Connection()', function () {
...
-      expect(connection).to.have.property('closed', false);
-      expect(connection).to.have.property(kDelayedTimeoutId, null);
+    it('emits a Connection.CLOSE event (asynchronously)', () => {
+      expect(closeCount).to.equal(0);
+      clock.runAll();
+      expect(closeCount).to.equal(1);
+    });
+
+    it('removes all listeners on the message stream', () => {
+      expect(messageStream.eventNames()).to.deep.equal([]);
```
nbbeeken, line 705: listenerCount + assertion message?

nbbeeken (changes requested):

src/cmap/connection.ts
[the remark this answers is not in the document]
nbbeeken: Yea, seems like a good call

src/cmap/connection.ts
```diff
@@ -444,34 +413,73 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+
+      this.emit(Connection.CLOSE);
+    };
+
+    for (const event of [
+      Connection.MESSAGE,
+      Connection.PINNED,
+      Connection.UNPINNED,
+      Connection.COMMAND_STARTED,
+      Connection.COMMAND_FAILED,
```
nbbeeken, line 457: I think CI is red b/c of this change, commandFailed is emitted when op.cb is called. I think we should move this loop into complete callback after op.cb is called

nbbeeken (changes requested):

src/cmap/connection.ts
nbbeeken: Actually then we can simplify this back to removeAllListeners if we call it after the emit(CLOSE)

nbbeeken (approved):

PR #3541, opened 2023-01-26
fix(NODE-4999): Write Concern 0 Must Not Affect Read Operations

W-A-James:
### Description

Update operations that create cursors to set their writeConcern to undefined.

#### What is changing?

- `ChangeStream` now silently unsets the `writeConcern` option
-  `AggregateOperation` now silently unsets the `writeConcern` option
-  `FindOperation`  now silently unsets the `writeConcern` option
-  `ListIndexesOperation` now silently unsets the `writeConcern` option
-  `ListCollections` now silently unsets the `writeConcern` option

##### Is there new documentation needed for these changes?

No

#### What is the motivation for this change?

This fixes the bug where read operations that have a writeConcern of 0 fail when getMore is called.

### Double check the following

- [x] Ran `npm run check:lint` script
- [x] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [x] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`
- [x] Changes are covered by tests
- [ ] New TODOs have a related JIRA ticket

PR #3575, opened 2023-02-17
fix(NODE-5044): Write Concern 0 Must Not Affect Read Operations (#3541)

baileympearson:
#### What is changing?

Cherry-picked changes from the 4.x branch with the same fix.

no.


src/operations/list_collections.ts
```diff
@@ -28,7 +28,8 @@ export class ListCollectionsOperation extends CommandOperation<string[]> {
   constructor(db: Db, filter: Document, options?: ListCollectionsOptions) {
```
nbbeeken, line 28: TS update

src/operations/indexes.ts
```diff
@@ -395,7 +395,8 @@ export class ListIndexesOperation extends CommandOperation<Document> {
   constructor(collection: Collection, options?: ListIndexesOptions) {
```
nbbeeken, line 395: TS update

nbbeeken (changes requested):

src/change_stream.ts
```diff
@@ -586,7 +586,8 @@ export class ChangeStream<
     super();
 
     this.pipeline = pipeline;
-    this.options = options;
+    this.options = { ...options };
+    delete this.options.writeConcern;
```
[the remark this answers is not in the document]
nbbeeken: @baileympearson disregard, let's just focus on the TS fixes

src/operations/indexes.ts
```diff
@@ -395,7 +395,10 @@ export class ListIndexesOperation extends CommandOperation<Document> {
...
     super(collection, options);
 
-    this.options = options ?? {};
+    this.options = { ...options };
+    // @ts-expect-error Write concern is disallowed in types, but
+    // must still be removed in case the user is not using Typescript
+    delete this.options.writeConcern;
```
nbbeeken, line 401: Interesting that we need `ts-expect-error` in `src` code, is there another way we could organize this? (can be in a follow up)

src/change_stream.ts
nbbeeken, line 590: Should changeStream also omit WC from the options?

nbbeeken (changes requested):

src/operations/indexes.ts
baileympearson: Not with strong typing. I mean, ultimately the real fix is to pass the full operation all the way down to the command layer, so that when we apply the session, we have access to the operation class and can check if it has a read aspect.

nbbeeken (changes requested):
requested changes already ^ bump

src/operations/indexes.ts
nbbeeken: If we're creating a new object, then can their be a new type that allows deleting the writeConcern without a ts-expect-error?

baileympearson: Do you mean like an internal options type? Does that not seem like overkill?

baileympearson: Also, how / why would we type something such that it allows the existence of a property that we will always delete?

baileympearson: What do you mean by types not being true? The type is correct. The argument could be made that we should omit this logic entirely because if someone passes a write concern to read operation and our types no longer support it, they are operating outside the bounds of our API.

nbbeeken: I thought the issue was with the option being inherited rather than being directly passed, right? 
I think if it was a correct type then we wouldn't need ts-expect-error

baileympearson: Good point about options being inherited.

I don't think the value of modifying the type to include `writeConcern?: never` is worthwhile.  I don't think that's the correct fix (I mention what I believe the correct fix to be above), so I'm going to leave this as-is.

baileympearson: Alternative option, again outside the scope of this work, is to consider a generic approach to constructing the options objects we use internally. Each command would construct an options object, only assigning the options that are actually supported for that command, instead of spreading all the existing options (maybe add an abstract `parseOptions(userOptions, inheritedOptions) : OptionInterface` method that each command implements). But that is a larger change that impacts the driver as a whole, so I won't address it here.

nbbeeken: I seem to recall a ticket related to that idea was a part of the Operation Layer Unification epic we used to have but I can't seem to find it, seems worth filing something for esp if we've considered it before.

I'd prefer if we made the types accurate rather than putting in compiler overrides

nbbeeken (approved):
Lgtm

PR #3577, opened 2023-02-20
feat(NODE-5035): enable OIDC authentication

durran:
### Description

Implements OIDC auth.

Spec PR: https://github.com/mongodb/specifications/pull/1365

Latest tools fix: https://github.com/mongodb-labs/drivers-evergreen-tools/pull/272

#### What is changing?
- Adds a new MongoDBOIDC auth provider.
- Introduces the concept of workflows for additional cloud providers in the future.
- Adds the spec prose tests for OIDC except reauthentication. (NODE-5036)

None

NODE-5035/DRIVERS-2415


src/cmap/auth/mongodb_oidc.ts
```diff
@@ -11,29 +20,131 @@ export interface OIDCMechanismServerStep1 {
...
+  return { accessToken: '' };
+}
+
+/**
+ * Generate the saslStart command document.
+ */
+function saslStartCommand(credentials: MongoCredentials): Document {
+  const payload: Document = {};
+  if (credentials.username || credentials.mechanismProperties.PRINCIPAL_NAME) {
+    payload.n = credentials.username || credentials.mechanismProperties.PRINCIPAL_NAME;
```
addaleax, line 131: Small heads up, `PRINCIPAL_NAME` got dropped from the spec text, with `username` being the relevant field now

src/cmap/auth/mongo_credentials.ts
```diff
@@ -30,24 +30,16 @@ function getDefaultAuthMechanism(hello?: Document): AuthMechanism {
...
   CANONICALIZE_HOST_NAME?: GSSAPICanonicalizationValue;
   AWS_SESSION_TOKEN?: string;
-  /** @internal Name for the OIDC device workflow */
   DEVICE_NAME?: 'aws' | 'azure' | 'gcp';
-  /** @internal Similar to a username, is require by OIDC when more than one IDP is configured. */
   PRINCIPAL_NAME?: string;
-  /** @internal User provided callback to get OIDC auth credentials */
   REQUEST_TOKEN_CALLBACK?: OIDCRequestFunction;
-  /** @internal User provided callback to refresh OIDC auth credentials */
   REFRESH_TOKEN_CALLBACK?: OIDCRefreshFunction;
```
addaleax, line 43: Also fyi, these functions now also take the principal name (i.e. username passed to the initial MongoClient) as an argument, as well as a timeout mechanism. In the spec, the example signature uses an integer, but referring to the spec text

> This can be as a timeout argument to the callback, a cancellation context passed to the callback, or some other language-appropriate mechanism. 

I would say that the “language-appropriate mechanism” for this is an `AbortSignal`.

durran: Yeah I just saw the spec was updated there. Thanks for the recommendation!

src/cmap/auth/mongodb_oidc.ts
durran: Uses only username now.

src/cmap/auth/mongo_credentials.ts
durran: Updated.

src/cmap/auth/mongodb_oidc/aws_device_workflow.ts
```diff
@@ -0,0 +1,36 @@
+import { readFile } from 'node:fs';
```
addaleax, line 1: Similar to NODE-5048, you can do this but some bundlers will choke on it by default, we can work around it but it might save us some trouble if the driver doesn’t use the `node:` prefix for now :)

durran: Updated.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -11,29 +17,91 @@ export interface OIDCMechanismServerStep1 {
...
+      .then(result => {
+        return callback(undefined, result);
+      })
+      .catch(error => {
+        callback(error);
+      });
+  }
+
+  /**
+   * Add the specualtive auth for the initial handshake.
```
addaleax, line 90:
```suggestion
   * Add the speculative auth for the initial handshake.
```

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -11,29 +17,91 @@ export interface OIDCMechanismServerStep1 {
...
+  result: OIDCRequestTokenResult,
+  timeout: AbortSignal | number
 ) => Promise<OIDCRequestTokenResult>;
+
+/** @internal */
+export const OIDC_WORKFLOWS = {
+  callback: new CallbackWorkflow(),
+  aws: new AwsDeviceWorkflow(),
+  azure: undefined,
+  gcp: undefined
```
addaleax, line 47:
```suggestion
  gcp: undefined,
  __proto__: null
```

since otherwise the `OIDC_WORKFLOWS[deviceName || 'callback']` below would return something for e.g. `deviceName === 'hasOwnProperty'`… :)

src/cmap/auth/mongodb_oidc.ts
durran: Updated.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -11,29 +17,91 @@ export interface OIDCMechanismServerStep1 {
   requestScopes?: string[];
 }
```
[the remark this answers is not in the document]
durran: Updated.

.evergreen/config.in.yml
```diff
@@ -107,9 +107,9 @@ functions:
         script: |
           ${PREPARE_SHELL}
 
-          # TODO(NODE-5035): Remove when merged - need to replace with branch just for OIDC.
+          # TODO(NODE-5035): Remove these 2 lines when PR merged.
           rm -rf "${DRIVERS_TOOLS}"
-          git clone --branch DRIVERS-2415 https://github.com/blink1073/drivers-evergreen-tools.git "${DRIVERS_TOOLS}"
+          git clone --branch DRIVERS-2415-orch https://github.com/mongodb-labs/drivers-evergreen-tools.git "${DRIVERS_TOOLS}"
```
addaleax, line 112: Just leaving a comment to remind about this

.evergreen/config.yml
```diff
@@ -77,27 +77,44 @@ functions:
...
           # than we have set in our evergreen project config. As these are
+
           # now specific to the OIDC ARN, we re-export for the python
+
           # scripts.
+
           export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+
           export AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+
```
addaleax, line 108: Are the extra line inserts here intentional?

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -0,0 +1,201 @@
...
+      }
+    } else {
+      // No entry means to start with the step one saslStart.
+      const executeCommand = promisify(connection.command.bind(connection));
+      const result = await executeCommand(
+        ns(credentials.source),
+        startCommandDocument(credentials),
+        undefined
+      );
+      // What to do about the payload?
```
addaleax, line 88: Is this comment still an open question? It seems right to me :)

src/cmap/auth/mongodb_oidc/aws_device_workflow.ts
```diff
@@ -0,0 +1,47 @@
...
+    return super.authenticate(connection, credentials, token);
+  }
+
+  /**
+   * Get the document to add for speculative authentication.
+   */
+  async speculativeAuth(): Promise<Document> {
+    const token = await getToken();
+    return { speculativeAuthenticate: commandDocument(token) };
+  }
```
addaleax, line 34: I’m (maybe wrongly?) assuming all other device workflows would also follow this pattern – `execute` calls `getToken()` and `super.authenticate()`, `speculativeAuth` calls `getToken()` and wraps it using `commandDocument()`.

Would it make sense to move this logic to `DeviceWorkflow` and instead only have an abstract `getToken()` method?

addaleax (commented):
Didn’t quite get around to the `test/` changes right now, will try to do that later, but what’s in `src/` looks good to me :+1:

.evergreen/config.yml
durran: No - our script that generates this file is doing this for some reason.

src/cmap/auth/mongodb_oidc/aws_device_workflow.ts
durran: Yeah that's a good suggestion.

durran: Updated per suggestion.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: Oh yes it's correct I just need to remove that comment.

durran: Removed the comment.

.evergreen/config.in.yml
durran: Removed as PR is merged.

test/manual/mongodb_oidc.test.ts
```diff
@@ -0,0 +1,450 @@
...
+        const doc = await collection.findOne();
+        expect(doc).to.equal(null);
+      });
+    });
+
+    // Drivers MUST be able to authenticate using either authentication or device
+    // type if there are multiple principals configured on the server.  Note that
+    // ``directConnection=true`` and ``readPreference=secondaryPreferred`` are needed
+    // because the server is a secondary on a replica set, on port ``27018``.
+    describe('3. Multiple Principles', function () {
```
W-A-James, line 114:
```suggestion
    describe('3. Multiple Principals', function () {
```

test/unit/cmap/auth/mongodb_oidc/aws_device_workflow.test.ts
```diff
@@ -0,0 +1,31 @@
...
+      });
+
+      after(function () {
+        process.env.AWS_WEB_IDENTITY_TOKEN_FILE = file;
+      });
+
+      it('returns an error in the callback', function () {
+        workflow.execute(connection, credentials, error => {
+          expect(error.message).to.include('AWS_WEB_IDENTITY_TOKEN_FILE');
+        });
```
W-A-James, line 27: The function signature for this does not take an error callback, so this passes because the assertion never occurs.

test/manual/mongodb_oidc.test.ts
```diff
@@ -0,0 +1,450 @@
...
+        });
+      });
+
+      context('when not providing a user', function () {
+        it('fails on option parsing', async function () {
+          expect(async () => {
+            new MongoClient(
+              'mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred'
+            );
+          }).to.throw;
```
W-A-James, line 257:
```suggestion
          }).to.throw();
```

The `throw` property must be called to execute the assertion.

For example, this will pass:

```typescript
it('throws', function() {
	expect(()=> {}).to.throw;
});
```

test/manual/mongodb_oidc.test.ts
```diff
@@ -0,0 +1,450 @@
...
+        });
+
+        after(async () => {
+          client?.close();
+        });
+
+        it('raises an error', async function () {
+          expect(async () => {
+            await client.connect();
+          }).to.throw;
```
W-A-James, line 285: Same as above

test/manual/mongodb_oidc.test.ts
```diff
@@ -0,0 +1,450 @@
...
```
W-A-James, line 311: Same as above

W-A-James (changes requested):

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -1,39 +1,136 @@
...
+          callback(error);
+        });
+    });
+  }
+}
+
+/**
+ * Gets either a device workflow or callback workflow.
+ */
+function getWorkflow(credentials: MongoCredentials, callback: Callback<Workflow>): void {
```
baileympearson, line 125: Is there any reason for a callback in this method?  There's no async work, so we could directly throw / return

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -1,39 +1,136 @@
...
+        return callback(error);
+      }
+      workflow
+        .speculativeAuth()
+        .then(result => {
+          return callback(undefined, { ...handshakeDoc, ...result });
+        })
+        .catch(error => {
+          callback(error);
+        });
```
baileympearson, line 117: This applies to all `then` chaining.

We should use the overload `.then(<success cb>, <error cb>)`.  If we use a `.then().catch()`, if `callback` throws in the `then` block, the callback will be called a second time in the catch block when that error is caught.

We pass this callback in so we can guarantee it never throws, but as a best practice we should opt for the overload that takes both the then and catch handlers in a single overload.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -0,0 +1,209 @@
...
+import { ns } from '../../../utils';
+import type { Connection } from '../../connection';
+import type { MongoCredentials } from '../mongo_credentials';
+import type { OIDCMechanismServerStep1, OIDCRequestTokenResult } from '../mongodb_oidc';
+import { AuthMechanism } from '../providers';
+import { TokenEntryCache } from './token_entry_cache';
+import type { Workflow } from './workflow';
+
+/* 5 minutes in milliseconds */
+const TIMEOUT = 300000;
```
baileympearson, line 14:
```suggestion
const TIMEOUT_MS = 300000;
```

maybe more clear in other places that this is a MS timeout?

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -0,0 +1,209 @@
...
+      if (entry.isValid()) {
+        // Skip step one and execute the step two saslContinue.
+        try {
+          const result = await finishAuth(entry.tokenResult, undefined, connection, credentials);
+          return result;
+        } catch (error) {
+          // If authentication errors when using a cached token we remove it from
+          // the cache.
+          this.cache.deleteEntry(connection.address, credentials.username);
+          return Promise.reject(error);
```
baileympearson, line 75:
```suggestion
          throw error;
```
any reason we can't directly throw the error, since we're in an async function?

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -0,0 +1,209 @@
...
+        return this.refreshAndFinish(
+          connection,
+          credentials,
+          entry.serverResult,
+          entry.tokenResult
+        );
+      }
+    } else {
+      // No entry means to start with the step one saslStart.
+      const executeCommand = promisify(connection.command.bind(connection));
```
baileympearson, line 90: Can we consider using `promisify` in the Connection constructor instead of per-invocation?  See https://github.com/mongodb/node-mongodb-native/blob/a45e0e1fc74a03ea43eb8cff058d758c3590b820/src/sdam/topology.ts#L235

Storing the promisified version makes the cost of using promisify once per-connection, instead of each time we execute a command from this module

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -0,0 +1,209 @@
...
+function continueCommandDocument(token: string, conversationId?: number): Document {
+  if (conversationId) {
+    return {
+      saslContinue: 1,
+      conversationId: conversationId,
+      payload: new Binary(BSON.serialize({ jwt: token }))
+    };
+  }
+  return {
+    saslStart: 1,
```
baileympearson, line 205: can we add a comment explaining why we're returning a saslStart command from `continueCommandDocument`?  I remember discussing this last thursday during tech chat but it would be nice to provide context for future devs (including myself)

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -0,0 +1,110 @@
+import type { OIDCMechanismServerStep1, OIDCRequestTokenResult } from '../mongodb_oidc';
+
+/* 5 minutes in milliseonds */
+const EXPIRATION_BUFFER = 300000;
+/* 5 hours in seconds */
+const DEFAULT_EXPIRATION = 18000;
```
baileympearson, line 6: Are these values pulled directly from the spec? If not, can we consider standardizing our unit of time measurement? It would be nice if OIDC, all the timeouts and durations were in milliseconds. (or seconds)

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Delete all expired entries from the cache.
+   */
+  deleteExpiredEntries(): void {
+    this.entries.forEach((entry, key, entries) => {
+      if (!entry.isValid()) {
+        entries.delete(key);
+      }
+    });
```
baileympearson, line 94:
```suggestion
    for (const [entry, key] of this.entries) {
      if (!entry.isValid()) {
        this.entries.delete(key);
      }
    }
```

I don't feel strongly about this, but in general for-each loops have superceeded forEach(cb) loops as best practice.

baileympearson (changes requested):
I didn't look at tests here, but have some suggestions on the implementation.  I like the workflow abstraction, and I love to see more usage of async-await in the driver.  Nice work 🙂

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * Set an entry in the token cache.
+   */
+  addEntry(
+    tokenResult: OIDCRequestTokenResult,
+    serverResult: OIDCMechanismServerStep1,
+    address: string,
+    username = ''
```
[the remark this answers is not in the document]
durran: Updated.

test/manual/mongodb_oidc.test.ts
durran: Updated.

test/unit/cmap/auth/mongodb_oidc/aws_device_workflow.test.ts
durran: Updated to use promises now.

test/manual/mongodb_oidc.test.ts
durran: Fixed.

test/manual/mongodb_oidc.test.ts
durran: Fixed.

src/cmap/auth/mongodb_oidc.ts
durran: The code got really odd when combining the callback from the caller and potentially throwing from this method. So the caller had both a try/catch and a callback call with the error. The combination of the two just seemed weird so I stuck with one - but happy to change if you feel strongly about it.

baileympearson: That's a good point, I don't feel strongly.

A potential alternative is to make this async, and then just then-chain onto it where we call it.  Then, the error this function rejects with will be caught by the `catch` handler for the promise chain.

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
durran: Yes the spec does define these so I kept them in the same units as the spec.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: Updated.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: Updated.

src/cmap/auth/mongodb_oidc.ts
durran: Updated - thanks for catching that.

src/cmap/auth/mongodb_oidc.ts
nbbeeken: Just some ideas, we aren't using this object as an enum per se, it serves as a lookup, so the losing the `as const`-ness of it wouldn't harm the usability

```typescript
export const OIDC_WORKFLOWS = Object.create(null) as { [deviceName: string]: Workflow | undefined };
OIDC_WORKFLOWS.callback = new CallbackWorkflow();
OIDC_WORKFLOWS.aws = new AwsDeviceWorkflow();
```

Alternatively, making a Map, and relying on `.get` returning undefined / defaulting the key to `'callback'` also serves us here.

```typescript
export const OIDC_WORKFLOWS = new Map<string, Workflow>([
  ['callback', new CallbackWorkflow()],
  ['aws', new AwsDeviceWorkflow()],
])
```

Would we prefer one of these to the proto setter?

addaleax: > Would we prefer one of these to the proto setter?

So, no strong feelings here, but maybe to clarify: `__proto__` inside object literals is special syntax and entirely distinct from the legacy `Object.prototype.__proto__` getter/setter (other than using the same identifier).

(Fun JS trivia fact: `eval('({ "__proto__": null })')` and `JSON.parse('{ "__proto__": null }')` are different things…)

I’d say it’s fairly standard to go with `{ ..., __proto__: null }` for lookup objects, especially when it’s a constant list and you don’t need all the flexibility that a map gives you.

src/cmap/auth/mongodb_oidc.ts
durran: I think I'd just prefer to keep it as-is, and then once we come back when async/awaiting all the internals just make it async then.

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
addaleax: If you think the units here might be confusing, I’d recommend to just incorporate them into the constant name (e.g. `EXPIRATION_BUFFER_MS`, `DEFAULT_EXPIRATION_SECS`, etc.)

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -0,0 +1,110 @@
...
+  addEntry(
+    address: string,
+    username: string,
+    tokenResult: OIDCRequestTokenResult,
+    serverResult: OIDCMechanismServerStep1
+  ): TokenEntry {
+    const entry = new TokenEntry(
+      tokenResult,
+      serverResult,
+      expirationTime(tokenResult.expiresInSeconds)
```
addaleax, line 59: Just a note, from the spec:

> If there is no `expiresInSeconds`, the token must be consider expired after first usage. 

`expirationTime(0)` is `Date.now()`, but there might be some question about how to interpret “no `expiresInSeconds`” in JS (e.g. do null and undefined fall into that category?).

Fwiw, from the devtools code we’re planning to always make use of this line and provide `expirationTime: 0` or whatever else allows us to bypass the global cache.

durran: I've added the units to the constants and decided to default `null` and `undefined` to equal `0` in this case.

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
durran: Updafted

test/manual/mongodb_oidc.test.ts
```diff
@@ -0,0 +1,456 @@
...
+        });
+      });
+
+      context('when not providing a user', function () {
+        it('fails on option parsing', function () {
+          expect(() => {
+            new MongoClient(
+              'mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred'
+            );
+          }).to.throw(/DEVICE_NAME|REQUEST_TOKEN_CALLBACK/);
```
W-A-James, line 257: nit: it would be nice if we asserted that we throw a specific error class in each of these error cases.

Don't need to block on this.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -71,12 +70,12 @@ export class CallbackWorkflow implements Workflow {
         } catch (error) {
           // If authentication errors when using a cached token we remove it from
           // the cache.
-          this.cache.deleteEntry(connection.address, credentials.username);
-          return Promise.reject(error);
+          this.cache.deleteEntry(connection.address, credentials.username || '');
```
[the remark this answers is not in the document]
durran: Yes - since we moved username to the second parameter of the method it could no longer be optional in the signature without some ugly overloads and type checking, which I find just messy. So we just ensure here it's always set to something.

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
baileympearson: Where is the defaulting happening?

src/cmap/auth/mongodb_oidc/device_workflow.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export abstract class DeviceWorkflow implements Workflow {
+  /**
+   * Execute the workflow. Looks for AWS_WEB_IDENTITY_TOKEN_FILE in the environment
+   * and then attempts to read the token from that path.
+   */
+  async execute(connection: Connection, credentials: MongoCredentials): Promise<Document> {
+    const token = await this.getToken();
+    const command = commandDocument(token);
+    const executeCommand = promisify(connection.command.bind(connection));
```
baileympearson, line 22: one missed `promisify` 🙂

test/unit/cmap/auth/mongodb_oidc/aws_device_workflow.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+      after(function () {
+        process.env.AWS_WEB_IDENTITY_TOKEN_FILE = file;
+      });
+
+      it('returns an error in the callback', async function () {
+        const result = await workflow.execute(connection, credentials).catch(error => {
+          expect(error.message).to.include('AWS_WEB_IDENTITY_TOKEN_FILE');
+        });
+        expect(result).to.equal(undefined, 'must fail without AWS_WEB_IDENTITY_TOKEN_FILE');
```
baileympearson, line 28: This will always be true, because of the success case the value will be passed to the callback, not returned from the function

baileympearson (changes requested):

src/cmap/auth/mongodb_oidc.ts
nbbeeken: the _proto_ that keeps on giving! TIL, thanks for the knowledge share, also no strong feelings, this gets the job done. @durran just offering here that, now that we have Maps we don't need to create lookup objects with a special consideration for inherited keys, if we want to take that approach

W-A-James (commented):
Note: Serverless failures are caused by issue documented in https://jira.mongodb.org/browse/DRIVERS-2564

Filing flaky test ticket for time series unified spec tests

W-A-James (approved):
Just need to address Bailey's comment re the always-passing test, but LGTM

src/cmap/auth/mongodb_oidc/device_workflow.ts
durran: Updated.

test/unit/cmap/auth/mongodb_oidc/aws_device_workflow.test.ts
durran: Changed to use try/catch

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
durran: https://github.com/mongodb/node-mongodb-native/pull/3577/files/6cfd1d5f36a9a9363e7c29b9bac3f37e32aac4ce#diff-bd6cb7449c5cb3124bcfb31066ba2c2374aea822f8d1326b4ba111e501105e04R102

baileympearson: _if_ we're provided a `null` value (which is disallowed by TS - but maybe something we should consider anyways because this will be arbitrary dynamic user input.  It never hurts to be safe.  I mentioned here because you were talking about it), it only works because of the way JavaScript coerces numbers when performing multiplication.

```typescript
function foo(x = "default") {
	return x
}

foo(undefined) // "default"
foo() // "default"
foo(null) // null
```

This function works when provided null because `<number> + null * number` is always zero, because `null` will get coerced to `0`.

_if_ we're going to support a `null` return value here, I think we should make it explicit (remove the default value, make it optional and use `??` to provide a default value inside the function).  

_if_ we decide not to support `null` here, then I think it's fine as-is

durran: I've changed to set the default in the function now to support null getting passed in.

test/manual/mongodb_oidc.test.ts
durran: Add check for error type.

baileympearson (approved):

W-A-James (approved):

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -0,0 +1,110 @@
+import type { OIDCMechanismServerStep1, OIDCRequestTokenResult } from '../mongodb_oidc';
+
+/* 5 minutes in milliseonds */
+const EXPIRATION_BUFFER_MS = 300000;
+/* 5 hours in seconds */
+const DEFAULT_EXPIRATION_SECS = 0;
```
dariakp, line 6: the comment above this line says it's supposed to be 5 hours, and the comment in L99 says the same thing - is the default to expire immediately? what's the intended spec behavior?

dariakp (changes requested):

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
durran: Updated the comment. The default in the spec is 0/expire immediately but I forgot to change the comment when I fixed it.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -1,39 +1,136 @@
...
+    callback: Callback<HandshakeDocument>
+  ): void {
+    const { credentials } = authContext;
+
+    if (!credentials) {
+      return callback(new MongoMissingCredentialsError('AuthContext must provide credentials.'));
+    }
+
+    getWorkflow(credentials, (error, workflow) => {
+      if (error || !workflow) {
```
dariakp, line 107: is `!workflow` possible without error? the `getWorkflow` function is designed not to return `null`, so if `!workflow`, it would be more appropriate to generate a `MongoRuntimeError` here instead of calling back with `null`

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -0,0 +1,110 @@
...
+    for (const [key, entry] of this.entries) {
+      if (!entry.isValid()) {
+        this.entries.delete(key);
+      }
+    }
+  }
+}
+
+/**
+ * Get an expiration time in milliseconds past epoch. Defaults to 5 hours.
```
dariakp, line 99: this still claims the expiration is 5 hours by default

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -1,39 +1,136 @@
...
+    super();
+  }
+
+  /**
+   * Authenticate using OIDC
+   */
+  override auth(authContext: AuthContext, callback: Callback): void {
+    const { connection, credentials, response } = authContext;
+
+    if (response && response.speculativeAuthenticate) {
```
dariakp, line 69: any reason not to use `response?.`?

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -1,39 +1,136 @@
...
+      return callback();
+    }
+
+    if (!credentials) {
+      return callback(new MongoMissingCredentialsError('AuthContext must provide credentials.'));
+    }
+
+    getWorkflow(credentials, (error, workflow) => {
+      if (error || !workflow) {
```
dariakp, line 78: same comment here about `!workflow` being a runtime error as in L107

src/cmap/auth/mongodb_oidc/aws_device_workflow.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * Get the token from the environment.
+   */
+  async getToken(): Promise<string> {
+    const tokenFile = process.env.AWS_WEB_IDENTITY_TOKEN_FILE;
+    if (tokenFile) {
+      return readFile(tokenFile, 'utf8');
+    } else {
```
dariakp, line 23: technically, the else wrapper isn't necessary here since L22 returns; I recommend `if (!tokenFile) { throw... }` and then `return readFile` on the bottom so that we follow a pattern of "validate then execute" as we do in the majority of the codebase

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -0,0 +1,227 @@
...
+      }
+      // Cache a new entry and continue with the saslContinue.
+      this.cache.addEntry(
+        connection.address,
+        credentials.username || '',
+        tokenResult,
+        stepOneResult
+      );
+      return finishAuth(tokenResult, conversationId, connection, credentials);
+    } else {
```
dariakp, line 164: can we invert this? `if (!request) { throw }` then un-nest what's current L149-163?

dariakp (changes requested):
Some notes

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      it('adds the token result', function () {
+        expect(entry.tokenResult).to.deep.equal(tokenResult);
+      });
+
+      it('adds the server result', function () {
+        expect(entry.serverResult).to.deep.equal(serverResult);
+      });
+
+      it('creates an expiration', function () {
+        expect(entry.expiration).to.be.above(Date.now());
```
dariakp, line 38: I think we could narrow the assertion range here using the `within` operator, or with a companion `below`

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      const expiredResult = { accessToken: 'test' };
+
+      before(function () {
+        cache.addEntry('localhost', 'user', expiredResult, serverResult);
+        entry = cache.getEntry('localhost', 'user');
+      });
+
+      it('sets an immediate expiration', function () {
+        expect(entry.expiration).to.be.below(Date.now() + 1);
```
dariakp, line 54: is the `+1` actually needed?

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        expiresInSeconds: null
+      };
+
+      before(function () {
+        cache.addEntry('localhost', 'user', expiredResult, serverResult);
+        entry = cache.getEntry('localhost', 'user');
+      });
+
+      it('sets an immediate expiration', function () {
+        expect(entry.expiration).to.be.below(Date.now() + 1);
```
dariakp, line 73: is the `+1` necessary?

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  describe('#deleteExpiredEntries', function () {
+    const cache = new TokenEntryCache();
+
+    const nonExpiredResult = {
+      accessToken: 'test',
+      expiresInSeconds: 600
+    };
+
+    before(function () {
+      cache.addEntry('localhost', 'user', tokenResult, serverResult);
```
dariakp, line 100: I'm confused, the `tokenResult` has an expiration of 100s, how is it already expired in L105?

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      cache.addEntry('localhost', 'user2', nonExpiredResult, serverResult);
+      cache.deleteExpiredEntries();
+    });
+
+    it('deletes all expired tokens from the cache', function () {
+      expect(cache.entries.size).to.equal(1);
```
dariakp, line 106: can we also assert on which entry is present or not present? (i.e., query the cache for each entry and expect a result for the non-expired one but not for the expired one)?

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    const cache = new TokenEntryCache();
+
+    before(function () {
+      cache.addEntry('localhost', 'user', tokenResult, serverResult);
+      cache.addEntry('localhost', 'user2', tokenResult, serverResult);
+    });
+
+    context('when there is a matching entry', function () {
+      it('returns the entry', function () {
+        expect(cache.getEntry('localhost', 'user').tokenResult).to.deep.equal(tokenResult);
```
dariakp, line 133: do we also expect it to plain `equal` the reference? looking at the code, it looks like the cache stores and returns the references directly as opposed to clones (I assume that's intentional); I think it may be worthwhile to make the `tokenResult` readonly for the test so that we can indirectly assert that the cache methods do not mutate the entry; we could also add an explicit test for it to each method

test/manual/mongodb_oidc.test.ts
```diff
@@ -0,0 +1,456 @@
...
+import { MongoClient, MongoInvalidArgumentError, OIDC_WORKFLOWS } from '../mongodb';
+
+describe('MONGODB-OIDC', function () {
+  context('when running in the environment', function () {
+    it('contains AWS_WEB_IDENTITY_TOKEN_FILE', function () {
+      expect(process.env).to.have.property('AWS_WEB_IDENTITY_TOKEN_FILE');
+    });
+  });
+
+  describe('OIDC Auth Spec Prose Tests', function () {
```
dariakp, line 14: since these are the prose tests, can we name the file in accordance with the test convention? i.e., `mongodb_oidc.prose.test`?

dariakp (changes requested):

test/manual/mongodb_oidc.test.ts
durran: Renamed.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
durran: I've changed the test to check for equal by reference and also frozen all the objects in the suite to ensure they are not mutated.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
durran: Added those extra two assertions.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
durran: The spec states:

> A cached accessToken will expire 5 minutes before the expiresInSeconds time, if given. If there is no expiresInSeconds, the token must be consider expired after first usage. 

So 100s is below the 5 minute threshold.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
durran: Changed to test `at.most(Date.now())` - the +1 was there to be defensive about code running fast enough to where `Date.now()` could be the same for both the expiration and the call in the assertion.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
durran: Changed to test `at.most(Date.now())` - the +1 was there to be defensive about code running fast enough to where `Date.now()` could be the same for both the expiration and the call in the assertion.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
durran: Changed assertion to be `at.most` for the `Date.now + expiresInSeconds * 1000`

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,143 @@
+import { expect } from 'chai';
+
+import { TokenEntryCache } from '../../../../mongodb';
+
+describe('TokenEntryCache', function () {
+  const tokenResult = {
```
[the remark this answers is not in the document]
durran: Renamed

test/unit/cmap/auth/mongodb_oidc/aws_device_workflow.test.ts
```diff
@@ -0,0 +1,34 @@
...
+      });
+
+      after(function () {
+        process.env.AWS_WEB_IDENTITY_TOKEN_FILE = file;
+      });
+
+      it('throws an error', async function () {
+        try {
+          await workflow.execute(connection, credentials);
+          fail('workflow must fail without AWS_WEB_IDENTITY_TOKEN_FILE');
```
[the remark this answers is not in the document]
durran: Changed to proper `expect.fail`

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: Inverted.

src/cmap/auth/mongodb_oidc/aws_device_workflow.ts
durran: Inverted.

src/cmap/auth/mongodb_oidc.ts
durran: Changed to pass a `MongoRuntimeError`

src/cmap/auth/mongodb_oidc.ts
durran: Changed.

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
durran: Fixed comment.

dariakp (changes requested):

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
dariakp: thanks for clarifying, I'm resolving this comment since this line is gone, but adding another one suggesting we clarify this in the test file itself

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
dariakp: we lost this test for explicit `null`, I added a comment in my following review

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
dariakp: there is a typo in the new line, but I added a new comment, so resolving this one

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      it('adds the token result', function () {
+        expect(entry.tokenResult).to.deep.equal(tokenResultWithExpiration);
+      });
+
+      it('adds the server result', function () {
+        expect(entry.serverResult).to.deep.equal(serverResult);
+      });
+
+      it('creates an expiration', function () {
+        expect(entry.expiration).to.be.at.most(Date.now() * 100 * 1000);
```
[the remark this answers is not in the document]
durran: Good catch. I've updated now.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+      const expiredResult = Object.freeze({ accessToken: 'test' });
+
+      before(function () {
+        cache.addEntry('localhost', 'user', expiredResult, serverResult);
+        entry = cache.getEntry('localhost', 'user');
+      });
+
+      it('sets an immediate expiration', function () {
+        expect(entry?.expiration).to.be.at.most(Date.now());
```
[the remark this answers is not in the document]
durran: Fixed.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      });
+    });
+  });
```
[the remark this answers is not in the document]
durran: Added the null test back.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      expiresInSeconds: 600
+    });
+
+    before(function () {
+      cache.addEntry('localhost', 'user', tokenResultWithExpiration, serverResult);
+      cache.addEntry('localhost', 'user2', nonExpiredResult, serverResult);
+      cache.deleteExpiredEntries();
+    });
+
+    it('deletes all expired tokens from the cache', function () {
```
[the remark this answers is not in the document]
durran: Updated the test to describe that now.

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -0,0 +1,110 @@
...
+function expirationTime(expiresInSeconds?: number): number {
+  return Date.now() + (expiresInSeconds ?? DEFAULT_EXPIRATION_SECS) * 1000;
+}
+
+/**
+ * Create a cache key from the address and username.
+ */
+function cacheKey(address: string, username: string): string {
+  return `${address}-${username}`;
+}
```
addaleax, line 110: Actually, just got reminded of this because of a conversation on our side of the OIDC auth, from [the spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#caching-credentials):

>  The cache key MUST also include a hashes of the callback function(s), if relevant and possible in the driver language.

I think we’d want that here as well, right? (Where I assume `hash` in JS is just the function identity, could be implemented e.g. by `WeakMap<Function, number>` which just assigns incremental numbers to the callbacks and then includes that in the string returned from `cacheKey`)

src/cmap/auth/mongo_credentials.ts
```diff
@@ -146,15 +145,18 @@ export class MongoCredentials {
     }
 
     if (this.mechanism === AuthMechanism.MONGODB_OIDC) {
-      if (this.username && this.mechanismProperties.DEVICE_NAME) {
+      if (this.username && this.mechanismProperties.SERVICE_NAME) {
```
dariakp, line 148: since this was previously `DEVICE_NAME` which was a TS enum `'aws' | 'azure' | 'gcp'`, do we want to do an analogous thing for `SERVICE_NAME` or does the spec specifically say that it needs to allow an arbitrary string? If it does not require that we keep it arbitrary, perhaps it would be better to limit `SERVICE_NAME` to just `'aws'` (based on the validation below), and then add other options once they are supported

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -40,14 +44,9 @@ export type OIDCRefreshFunction = (
...
 /** @internal */
-export const OIDC_WORKFLOWS = {
-  callback: new CallbackWorkflow(),
-  aws: new AwsDeviceWorkflow(),
-  azure: undefined,
-  gcp: undefined,
-  // Set the protype to null to avoid false returns and keep types.
-  __proto__: null
-};
+export const OIDC_WORKFLOWS: Map<string, Workflow> = new Map();
```
dariakp, line 47: should we provide a better type here and restrict the allowed keys in this map?

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
```diff
@@ -18,59 +26,55 @@ describe('TokenEntryCache', function () {
...
       before(function () {
-        cache.addEntry('localhost', 'user', expiredResult, serverResult);
-        entry = cache.getEntry('localhost', 'user');
+        cache.addEntry('localhost', 'user', fnOne, fnTwo, expiredResult, serverResult);
+        entry = cache.getEntry('localhost', 'user', fnOne, fnTwo);
       });
 
       it('sets an immediate expiration', function () {
-        expect(entry.expiration).to.be.below(Date.now() + 1);
+        expect(entry?.expiration).to.be.at.most(Date.now());
```
dariakp, line 77:
```suggestion
        expect(entry.expiration).to.be.at.most(Date.now());
```

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -105,6 +132,35 @@ function expirationTime(expiresInSeconds?: number): number {
...
+ */
+function hashFunctions(
+  requestFn: OIDCRequestFunction | null,
+  refreshFn: OIDCRefreshFunction | null
+): string {
+  let requestHash = FN_HASHES.get(requestFn || NO_FUNCTION);
+  let refreshHash = FN_HASHES.get(refreshFn || NO_FUNCTION);
+  if (!requestHash && requestFn) {
+    // Create a new one for the function and put it in the map.
+    FN_HASH_COUNTER++;
```
dariakp, line 155: how frequently do we expect to access this codepath and create new functions? if it's a human frequency (e.g., once per second or less) then this approach should be fine - just want to confirm that we don't need to build in a reuse mechanism for the generated numbers

dariakp (changes requested):

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
addaleax: It’s at most 2 numbers per MongoClient, that seems fine to me

src/cmap/auth/mongo_credentials.ts
durran: Needs to be an arbitrary string for when Kerberos uses that field.

test/unit/cmap/auth/mongodb_oidc/token_entry_cache.test.ts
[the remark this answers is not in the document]
durran: Fixed.

src/cmap/auth/mongodb_oidc.ts
durran: Needs to be a generic string here since that's what `SERVICE_NAME` is. If we do change it again to what Anna suggested in https://github.com/mongodb/specifications/pull/1383 then I'll make this more strict again in the NODE-5036 PR not to hold this up.

dariakp: I think it would be easier to just make this change here, since `PROVIDER_NAME` seems to be the consensus - I can merge it first thing tomorrow morning once it's all updated, since the semantics of the rest of this PR shouldn't be affected

durran: Change is made here now.

PR #3578, opened 2023-02-27
refactor(NODE-4685): unified spec runner to support logging tests

W-A-James:
### Description

Implement changes to unified spec runner to facilitate testing of logging spec implementation

#### What is changing?

- Implement `$$matchAsRoot` operator
- Implement `$$matchAsDocument` operator
- Implement matching and interpretation of new properties
	- `expectLogMessages`
	- `expectedLogMessages`
	- `expectedLogMessagesForClient`
	- `observeLogMessages`
- Implement unit tests for new operators
- Stub out mechanism for `UnifiedMongoClient` to be able to collect logs from `MongoLogger`

No

##### TODOs

- [x] Update AC of NODE-4849/file new ticket to include implementation of modifications to the unified spec runner that require the MongoLogger to be able to write logs to a Writable stream
- [ ] ~Update AC of NODE-4849/file new ticket to implement total ordering for `LogSeverity`~
- [x] Break out new ticket to add valid-pass and valid-fail tests to the unified-spec-format spec for `$$matchAsRoot` and `$$matchAsDocument` operators


test/integration/command-logging-and-monitoring/command_logging.spec.test.ts
```diff
@@ -0,0 +1,8 @@
+import * as path from 'path';
+
+import { loadSpecTests } from '../../spec';
+import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
+
+describe('Command Logging Spec (unified)', () => {
+  runUnifiedSuite(loadSpecTests(path.join('command-logging-and-monitoring', 'logging', 'command')));
```
W-A-James, line 7: Going to change this to actually run all of the command logging tests, but for only trying to run one file

test/tools/unified-spec-runner/entities.ts
```diff
@@ -169,6 +208,7 @@ export class UnifiedMongoClient extends MongoClient {
     this.observedSdamEvents = (description.observeEvents ?? [])
       .map(e => UnifiedMongoClient.SDAM_EVENT_NAME_LOOKUP[e])
       .filter(e => !!e);
+    this.observedLogEvents = new Map();
```
W-A-James, line 211: Will initialize this with the observedLogEvents passed in to the constructor in future, but need to implement NODE-4849 to have them take effect

test/tools/unified-spec-runner/entities.ts
```diff
@@ -429,6 +473,7 @@ export class EntitiesMap<E = Entity> extends Map<string, E> {
     const map = entityMap ?? new EntitiesMap();
     for (const entity of entities ?? []) {
       if ('client' in entity) {
+        // TODO(NODE-4849): check observeLogMessages to set the kinds of logs we should be collecting
```
W-A-James, line 476: See previous comment relating to NODE-4849

test/tools/unified-spec-runner/entities.ts
```diff
@@ -94,16 +102,40 @@ export type CmapEvent =
...
+class LogCollector extends Writable {
+  _accumulator: Log[];
+  constructor(options: WritableOptions) {
+    super({ ...options, objectMode: true });
+    this._accumulator = [];
+  }
+
+  _write(chunk: any, encoding: string, next: (error?: Error) => void) {
+    this._accumulator.push(chunk as Log);
+    next();
```
W-A-James, line 120: Need to check that chunk is actually a log message

test/tools/unified-spec-runner/entities.ts
```diff
@@ -94,16 +102,40 @@ export type CmapEvent =
...
   | ConnectionPoolClearedEvent;
 export type SdamEvent = ServerDescriptionChangedEvent;
+export type Log = ExpectedLogMessage;
 
 function getClient(address) {
   return new MongoClient(`mongodb://${address}`, getEnvironmentalOptions());
 }
 
+class LogCollector extends Writable {
+  _accumulator: Log[];
```
W-A-James, line 112: will change this to be named `_logs`

src/connection_string.ts
```diff
@@ -1282,7 +1282,9 @@ export const OPTIONS = {
   index: { type: 'any' },
   // Legacy Options, these are unused but left here to avoid errors with CSFLE lib
   useNewUrlParser: { type: 'boolean' } as OptionDescriptor,
-  useUnifiedTopology: { type: 'boolean' } as OptionDescriptor
+  useUnifiedTopology: { type: 'boolean' } as OptionDescriptor,
+  // MongoLogger
+  mongodbLogPath: { type: 'any' }
```
W-A-James, line 1287: This is added to allow hook the unified spec runner up with the client logger

test/tools/unified-spec-runner/entities.ts
```diff
@@ -94,16 +102,44 @@ export type CmapEvent =
   | ConnectionCheckedInEvent
   | ConnectionPoolClearedEvent;
 export type SdamEvent = ServerDescriptionChangedEvent;
+export type Log = ExpectedLogMessage;
```
W-A-James, line 105: Will replace with a more appropriate type

test/tools/unified-spec-runner/match.ts
```diff
@@ -294,7 +318,7 @@ export function specialCheck(
     // $$sessionLsid
     const session = entities.getEntity('session', expected.$$sessionLsid, false);
     expect(session, `Session ${expected.$$sessionLsid} does not exist in entities`).to.exist;
-    const entitySessionHex = session.id!.id.buffer.toString('hex').toUpperCase();
+    const entitySessionHex = session.id!.id.buffer.toString().toUpperCase();
```
W-A-James, line 321: Will open a new ticket to document this change. In summary, the Buffer.toString() does not take an argument, so this change doesn't modify any functionality.

test/tools/unified-spec-runner/match.ts
```diff
@@ -323,6 +347,23 @@ export function specialCheck(
         ejson`expected value at path ${path.join('')} NOT to exist, but received ${actual}`
       ).to.be.false;
     }
+  } else if (isMatchAsDocumentOperator(expected)) {
+    if (typeof actual === 'string') {
+      const actualDoc = EJSON.parse(actual, { relaxed: false });
```
W-A-James, line 352: Spec states that `$$matchAsDocument` should be used on EJSON strings

test/tools/unified-spec-runner/match.ts
```diff
@@ -323,6 +347,23 @@ export function specialCheck(
...
+      resultCheck(actualDoc, expected.$$matchAsDocument as any, entities, path, true);
+    } else {
+      expect.fail(`Expected value at path ${path.join('')} to be string, but received ${actual}`);
+    }
+  } else if (isMatchAsRootOperator(expected)) {
+    expect(
+      typeof actual,
+      `Expected value at path ${path.join('')} to be object, but received ${actual}`
+    ).to.equal('object');
+    expect(typeof expected.$$matchAsRoot, 'Value of $$matchAsRoot must be an object').to.equal(
```
W-A-James, line 362: Spec states that `$$matchAsDocument` should only be used when `expected` and `actual` are documents

test/tools/unified-spec-runner/schema.ts
```diff
@@ -307,6 +328,13 @@ export interface ExpectedError {
   expectResult?: unknown;
 }
 
+export interface ExpectedLogMessage {
+  level: string;
```
W-A-James, line 332: Will replace this `string` type with `ObservableLogSeverity`

test/unit/tools/unified_spec_runner.test.ts
```diff
@@ -0,0 +1,332 @@
...
+
+          it('throws AssertionError when failure is absent', function () {
+            actual = {
+              level: 'debug',
+              component: 'command',
+              data: {
+                message: 'some message'
+              }
+            };
+            expect(runCompareLogs).to.throw(AssertionError);
```
W-A-James, line 186: Will add an error message to match against where appropriate

src/connection_string.ts
W-A-James: Willing to revert this and defer this change to NODE-4849

src/mongo_logger.ts
```diff
@@ -19,6 +19,24 @@ export const SeverityLevel = Object.freeze({
...
+  [SeverityLevel.ERROR, 6],
+  [SeverityLevel.CRITICAL, 7],
+  [SeverityLevel.ALERT, 8],
+  [SeverityLevel.EMERGENCY, 9]
+]);
+
+for (const [level, value] of SeverityLevelMap.entries()) {
+  SeverityLevelMap.set(value, level);
+}
+
```
W-A-James, line 39: This is done to allow for easy comparison of log levels

W-A-James:
Once this goes through review, I will skip the CLAM spec tests

src/connection_string.ts
nbbeeken: This is fine, it doesn't exactly expose it as public API b/c the types aren't updated and we still have the symbol feat flag

nbbeeken (changes requested):

test/integration/command-logging-and-monitoring/command_logging.spec.test.ts
```diff
@@ -0,0 +1,8 @@
+import * as path from 'path';
+
+import { loadSpecTests } from '../../spec';
+import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
+
+describe('Command Logging Spec (unified)', () => {
```
nbbeeken, line 6:
```suggestion
describe('Command Logging Spec', () => {
```

test/tools/unified-spec-runner/entities.ts
```diff
@@ -149,13 +162,23 @@ export class UnifiedMongoClient extends MongoClient {
   } as const;
 
   constructor(uri: string, description: ClientEntity) {
+    const logCollector = new Transform({
+      objectMode: true,
+      transform(log: LogMessage, _: string, next: (e: Error | null, log: LogMessage) => void) {
+        next(null, log);
+      }
+    });
```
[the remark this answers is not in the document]
baileympearson: I was curious about the decision to use a transform stream and manually listen for data events on it.  What was the rationale behind this design?  

Why not make a log sink?

```typescript

class UnifiedLogCollector extends Writable {
  collectedEvents = [];

  _write(chunk) { // override syntax might not be right here but it's easy to do
    if (<conditions relevant to collecting logs>) {
      this.collectedEvents.push(chunk);
    }
  }
}

// and then in the UTR

export class UnifiedMongoClient extends MongoClient {
  // ...
  constructor() {
    this.logCollector = new UnifiedLogCollector();
  }

```

The new methods added to `UnifiedMongoClient` would then just become `getter`s that proxy onto methods on the `UnifiedLogCollector`.  This better encapsulates all the logging-related UTR logic into the `UnifiedLogCollector` (it doesn't pollute the `UnifiedMongoClient`) and we don't have to worry about manually consuming data from the stream.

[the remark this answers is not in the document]
baileympearson: I don't feel strongly about using a Nodejs stream or just implementing the interface.  But I don't like using a transform stream and listening to the data events, it feels like the wrong choice conceptually.  We're not passing the logs along anywhere here - so a transform or a passthrough stream is the wrong choice because we don't need the readable capability of duplex streams.

test/tools/unified-spec-runner/entities.ts
```diff
@@ -210,6 +233,10 @@ export class UnifiedMongoClient extends MongoClient {
     }
   }
 
+  getCapturedLogs(): LogMessage[] {
+    return this.logs;
+  }
```
baileympearson, line 238: can we name `this.logs` better and get rid of the getter?

test/unit/tools/unified_spec_runner.test.ts
```diff
@@ -0,0 +1,345 @@
...
+              a: { $$type: ['int'] }
+            }
+          };
+        });
+
+        context('when actual value is EJSON string', function () {
+          it('throws AssertionError when it finds extra keys', function () {
+            actual =
+              '{"data": {"$numberLong": "100"}, "a": {"$numberInt": "10"}, "b": {"$numberInt": "100"}}';
+            expect(runResultCheck).to.throw(AssertionError);
```
baileympearson, line 43: Is there a reason we're relying on the `runResultCheck` helper?  Usually we just inline the function call in an arrow function:  `expect(() => resultCheck(actual, expected, entitiesMap, [])).to.throw(AssertionError)`.

Also, we need to specifically assert that `resultCheck` fails for the correct reasons.  Can we add a regular expression match as well? `expect(() => ....).to.throw(AssertionError, /regex here/)`

baileympearson (changes requested):

test/tools/unified-spec-runner/entities.ts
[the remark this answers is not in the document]
nbbeeken: We can switch to the minimal interface when the code allows for it in NODE-4813. PassThrough + data even will be easy to replace since the real logic is inside pushLogMessage and it's a way to get the test runner past the current stream instanceof check to unlock debugging when the driver starts calling `.write`

test/tools/unified-spec-runner/match.ts
```diff
@@ -295,7 +319,7 @@ export function specialCheck(
     const session = entities.getEntity('session', expected.$$sessionLsid, false);
     expect(session, `Session ${expected.$$sessionLsid} does not exist in entities`).to.exist;
     const entitySessionHex = session.id!.id.buffer.toString('hex').toUpperCase();
-    const actualSessionHex = actual.id.buffer.toString('hex').toUpperCase();
+    const actualSessionHex = actual.id!.buffer.toString('hex').toUpperCase();
```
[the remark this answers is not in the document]
W-A-James: Yeah, this was mostly just because of TS warnings

test/unit/tools/unified_spec_runner.test.ts
W-A-James: The reason we're relying on the `runResultCheck` helper is just to make sure that no test can pass unless the function under test has been called. If you take a look at the top-level `beforeEach` and `afterEach` hooks, you can see where that's done.

baileympearson: Does `expect(() => ...).not.to.throw(AssertionError)` not cover the passing cases?

W-A-James: It would run, but my rationale behind putting that helper is just to ensure that if a new test is added that the only way it can pass is if it calls the function in question. I can go ahead and remove it since it was more of a development convenience than anything.

test/tools/unified-spec-runner/match.ts
```diff
@@ -523,6 +564,51 @@ export function matchesEvents(
...
 }
 
+export function compareLogs(
+  expected: ExpectedLogMessage[],
+  actual: ExpectedLogMessage[],
+  entities: EntitiesMap
+): void {
+  expect(actual).to.have.lengthOf(expected.length);
+
+  for (const [index, actualLog] of actual.entries()) {
```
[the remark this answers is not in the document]
W-A-James: We still need the index to set the root prefix appropriately, so I think it's better to keep it as it is since we'd just have to add an index variable either way

test/tools/unified-spec-runner/entities.ts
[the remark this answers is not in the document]
W-A-James: I think Bailey is right here and we should use a proper writable stream that just accumulates the logs for later access. It's not much more work and encapsulates the log filtering well.

test/unit/tools/unified_spec_runner.test.ts
baileympearson: My thought is that it's probably not necessary, but extra testing never hurts.  If we do want to enforce this with after hooks, I'd suggest using sinon.
```typescript
import * as sinon from 'sinon';

const resultCheckSpy = sinon.spy(resultCheck);

afterEach(() => {
  expect(resultCheckSpy.called).to.be.true;
});
```

test/tools/unified-spec-runner/match.ts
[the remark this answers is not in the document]
nbbeeken: Maybe we want to leave the code alone if it's just warnings then? or do we want to finish the improvement?

nbbeeken (changes requested):

test/integration/command-logging-and-monitoring/command_logging.spec.test.ts
```diff
@@ -0,0 +1,8 @@
+import * as path from 'path';
+
+import { loadSpecTests } from '../../spec';
+import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
+
+describe('Command Logging Spec', () => {
```
nbbeeken, line 6: Now that we've confirmed they run, let's add a skip & skipReason with a TODO

nbbeeken (changes requested):

test/tools/unified-spec-runner/entities.ts
```diff
@@ -110,15 +110,39 @@ function getClient(address) {
...
+  collectedLogs: LogMessage[] = [];
+  observeLogMessages: Record<ObservableLogComponent, ObservableLogSeverity>;
+
+  constructor(observeLogMessages?: Record<ObservableLogComponent, ObservableLogSeverity>) {
+    super({ objectMode: true });
+    this.observeLogMessages =
+      observeLogMessages ?? ({} as Record<ObservableLogComponent, ObservableLogSeverity>);
+  }
+
+  _write(log: LogMessage) {
```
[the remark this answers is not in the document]
baileympearson: Good catch.

https://nodejs.org/api/stream.html#writable_writechunk-encoding-callback

> The callback function must be called synchronously inside of writable._write() or asynchronously (i.e. different tick) to signal either that the write completed successfully or failed with an error. The first argument passed to the callback must be the Error object if the call failed or null if the write succeeded.

nbbeeken (approved):

nbbeeken (dismissed):

nbbeeken (approved):

test/tools/unified-spec-runner/match.ts
```diff
@@ -523,6 +567,51 @@ export function matchesEvents(
...
+
+    // Check that log levels match
+    expect(actualLog).to.have.deep.property('level', expectedLog.level);
+
+    // Check that components match
+    expect(actualLog).to.have.deep.property('component', expectedLog.component);
+
+    // NOTE: The spec states that if the failureIsRedacted flag is present, we
+    // must assert that a failure occurred.
+    if (expectedLog.failureIsRedacted !== undefined) {
```
dariakp, line 589: you may as well assert that `expectedLog.failureIsRedacted` is a boolean in the line below, we might not be running the `invalid` tests, but there's no reason not to have the guard

test/tools/unified-spec-runner/match.ts
```diff
@@ -523,6 +567,51 @@ export function matchesEvents(
...
+      if (expectedLog.failureIsRedacted) {
+        // Assert that a failure is present and has been redacted
+        expect(actualLog.data.failure, 'Can only use failureIsRedacted when a failure exists').to
```
dariakp, line 592: I think this error message is incorrect, if you can only use the flag when a failure exists that's about the expected value, not the actual. I think you can just say "Expected failure to exist" here.

test/tools/unified-spec-runner/match.ts
```diff
@@ -523,6 +567,51 @@ export function matchesEvents(
...
+          .exist;
+        expect(actualLog.data.failure, 'Expected failure to have been redacted').to.deep.equal({});
+      } else {
+        // Assert that a failure is present and has not been redacted
+        expect(actualLog.data.failure, 'Can only use failureIsRedacted when a failure exists').to
```
dariakp, line 597: This is unnecessarily repetitive, just move the assertion up to just below L589, since it's independent of the actual value of the flag

test/tools/unified-spec-runner/match.ts
```diff
@@ -523,6 +567,51 @@ export function matchesEvents(
...
+          .exist;
+        expect(
+          actualLog.data.failure,
+          'Expected failure to have not been redacted'
+        ).to.not.deep.equal({});
+      }
+    } else {
+      expect(
```
dariakp, line 605: is this correct? per the spec, `failureIsRedacted` is optional and *can* be present only on logs that have a `failure`; so anything with the `failureRedacted` flag will definitely have a `failure`, but anything without that flag could either not have a failure or have a failure where we don't care about whether or not it's redacted

the actual existence of the failure would be verified by the `resultCheck`, no? (since it would compare `actualLog.data.failure` against `expectedLog.data.failure`)

test/tools/unified-spec-runner/entities.ts
```diff
@@ -95,17 +104,54 @@ export type CmapEvent =
...
+    log: LogMessage,
+    _: string,
+    callback: (e: Error | null, l: LogMessage | undefined) => void
+  ) {
+    const minLogLevel = this.observeLogMessages[log.component] ?? 'off';
+    const numericMinLogLevel = SeverityLevelMap.get(minLogLevel) ?? 0;
+
+    const numericLogLevel = SeverityLevelMap.get(log.level);
+    if (typeof numericLogLevel !== 'number') {
+      expect.fail('Log level must be valid log level');
```
dariakp, line 134:
```suggestion
      expect.fail(`Invalid log level: ${log.level}`);
```

test/tools/unified-spec-runner/entities.ts
```diff
@@ -95,17 +104,54 @@ export type CmapEvent =
...
+      observeLogMessages ?? ({} as Record<ObservableLogComponent, ObservableLogSeverity>);
+  }
+
+  _write(
+    log: LogMessage,
+    _: string,
+    callback: (e: Error | null, l: LogMessage | undefined) => void
+  ) {
+    const minLogLevel = this.observeLogMessages[log.component] ?? 'off';
+    const numericMinLogLevel = SeverityLevelMap.get(minLogLevel) ?? 0;
```
dariakp, line 130: have we considered mapping the log level `off` to `Infinity` or `-Infinity` instead and aligning our numeric values with what the spec refers to as the `Code` for the corresponding severity?

src/mongo_logger.ts
```diff
@@ -19,6 +19,24 @@ export const SeverityLevel = Object.freeze({
 /** @internal */
 export type SeverityLevel = (typeof SeverityLevel)[keyof typeof SeverityLevel];
 
+/** @internal */
+export const SeverityLevelMap: Map<string | number, string | number> = new Map([
+  [SeverityLevel.OFF, 0],
```
dariakp, line 24: I think it might make sense to align these to the `code` corresponding to the severity instead, described here: https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#log-severity-levels

and maybe use `Infinity` / `-Infinity` for easier comparison where we'll be determining whether or not to log something

dariakp (changes requested):

src/mongo_logger.ts
nbbeeken: We may not need this data structure in the driver when we actually begin implementing log invocations depending on how severity filtering logic is handled. However, the tests do need this though to translate the severity level requested in the tests to what should be filtered, maybe we move this out of the driver? we could always copy it back if it proves useful.

dariakp: I don't see how we avoid having a numeric map considering we have to have some way of tracking which log levels to log for a given setting

dariakp: but I don't object to settling this later; if it's for tests only though, let's move it to test and still align it with the codes, if only to avoid confusion when reading the spec (since the numbers 0-8 already get assigned there)

nbbeeken: Yea I imagine we'll need a numeric mapping but the check for `off` could be different. Either `off=0` and the check is `currentLvl !== 0 && currentLvl >= setLvl` or you use `off=Infinity` to always make `off` larger `currentLvl >= setLvl`, or you can reverse it 😅. And I think we're not sure yet if this filtering is going to be in each severity helper or somehow determined at the setup step, we can avoid deciding now

dariakp: well, one of those reduces extra checks and reduces the confusion with log level 0 corresponding to "off" for us and to "emergency" in the spec, so, like I said, if we are going to have any sort of number map here or in the unified test, let's make that map match the spec.

test/tools/unified-spec-runner/match.ts
```diff
@@ -294,8 +318,11 @@ export function specialCheck(
     // $$sessionLsid
     const session = entities.getEntity('session', expected.$$sessionLsid, false);
     expect(session, `Session ${expected.$$sessionLsid} does not exist in entities`).to.exist;
-    const entitySessionHex = session.id!.id.buffer.toString('hex').toUpperCase();
-    const actualSessionHex = actual.id.buffer.toString('hex').toUpperCase();
+    const entitySessionHex = session.id?.id.toString('hex').toUpperCase();
+    const actualSessionHex = actual.id?.toString('hex').toUpperCase();
+    expect(entitySessionHex).to.be.a('string');
+    expect(actualSessionHex).to.be.a('string');
```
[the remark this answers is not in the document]
W-A-James: The reason these lines were changed is that TS was complaining about the call to `toString` not accepting an argument, which lead to some confusion on my part and an update to the structure of these lines.

I can go ahead and update the assertion to provide a more useful message.

W-A-James:
Documenting result of Slack discussion:

Going to change implementation of `UnifiedMongoClient`'s constructor to use the `ClientEntity` object passed into it to configure the wrapped `MongoClient`'s instance of `MongoLogger` rather than using it to filter logs on the test-runner's end. To facilitate this, we will need to modify the `parseOptions` function and the `OPTIONS` object in `src/connection_string.ts`.

src/connection_string.ts
```diff
@@ -521,6 +521,9 @@ export function parseOptions(
   const loggerFeatureFlag = Symbol.for('@@mdb.enableMongoLogger');
   mongoOptions[loggerFeatureFlag] = mongoOptions[loggerFeatureFlag] ?? false;
 
+  const internalLoggerConfigSymbol = Symbol.for('@@mdb.internalLoggerConfig');
```
W-A-James, line 524: Is there anywhere I should declare this symbol outside of this function? I didn't declare it anywhere else simply because it is something we only intend to use for testing.

test/tools/unified-spec-runner/entities.ts
```diff
@@ -149,13 +179,23 @@ export class UnifiedMongoClient extends MongoClient {
...
   constructor(uri: string, description: ClientEntity) {
+    const logCollector = new UnifiedLogCollector();
     super(uri, {
       monitorCommands: true,
       [Symbol.for('@@mdb.skipPingOnConnect')]: true,
+      [Symbol.for('@@mdb.enableMongoLogger')]: true,
+      [Symbol.for('@@mdb.internalLoggerConfig')]: {
+        componentSeverities: description.observeLogMessages,
+        logDestination: logCollector
+      },
```
W-A-James, line 190: Passing in the logDestination like this comes with the bonus that TS doesn't complain about options that don't exist

test/tools/unified-spec-runner/entities.ts
```diff
@@ -95,16 +101,40 @@ export type CmapEvent =
...
   | ConnectionPoolClearedEvent;
 export type SdamEvent = ServerDescriptionChangedEvent;
+export type LogMessage = Omit<ExpectedLogMessage, 'failureIsRedacted'>;
 
 function getClient(address) {
   return new MongoClient(`mongodb://${address}`, getEnvironmentalOptions());
 }
 
+// TODO(NODE-4813): Remove this class in favour of a simple object with a write method
+/* TODO(NODE-4813): Ensure that the object that we replace this with has logic to convert the
```
W-A-James, line 111: Added this TODO instead of adding the logic here since we are still solidifying what the shape of the logs will be

src/connection_string.ts
```diff
@@ -542,7 +545,7 @@ export function parseOptions(
     loggerClientOptions
   );
 
-  return mongoOptions;
+  return { ...mongoOptions, ...internalLoggerConfig };
```
nbbeeken, line 548: Instead of making changes here, a code path that will always be reached, can we instead make 
`internalLoggerConfig` spread into the `loggerClientOptions` only? That way the functionality is still only in the flagged code path

test/tools/unified-spec-runner/entities.ts
```diff
@@ -195,20 +179,20 @@ export class UnifiedMongoClient extends MongoClient {
...
+    const logCollector = new UnifiedLogCollector();
     super(uri, {
       monitorCommands: true,
       [Symbol.for('@@mdb.skipPingOnConnect')]: true,
       [Symbol.for('@@mdb.enableMongoLogger')]: true,
-      mongodbLogPath: logCollector,
+      [Symbol.for('@@mdb.internalMongoLoggerConfig')]: {
+        componentSeverities: { default: 'off', ...description.observeLogMessages },
+        logDestination: logCollector
+      },
```
nbbeeken, line 190: It's fine to continue passing the writable into the mongodbLogPath here since that option is for sure going to be part of the final API. The env var config is what we need to be able to configure from here, so I think internalMongoLoggerConfig should point to an object that has `components -> severity`. For now I think we can just pass through observerLogMessages as is, there will need to be some fixing at some point to connect the component names in observerLogMessages to the corresponding env variables.

nbbeeken (changes requested):

src/mongo_logger.ts
nbbeeken: I somehow entirely missed that the actual numbers were not equal to the one's in the spec, which explains some of our confusion last week, good catch

test/tools/unified-spec-runner/entities.ts
```diff
@@ -149,12 +179,39 @@ export class UnifiedMongoClient extends MongoClient {
...
+        }
+      }
+      return output.join('').toUpperCase();
+    };
+
+    // NOTE: this is done to override the logger environment variables
+    for (const key in description.observeLogMessages) {
+      componentSeverities['MONGODB_LOG_' + camelToUpperSnake(key)] =
+        description.observeLogMessages[key];
+    }
```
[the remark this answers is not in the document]
nbbeeken: I would hard code the mapping given its a know limited set of keys and then we can find out if there are ones we are unaware of during test sync. But I don't feel strongly

nbbeeken (approved):
nice work, lgtm!

dariakp (approved):
very nice work

test/unit/tools/unified_spec_runner.test.ts
```diff
@@ -0,0 +1,346 @@
+import { BSON } from 'bson';
+import { AssertionError, expect } from 'chai';
+import * as sinon from 'sinon';
+
+import { EntitiesMap } from '../../tools/unified-spec-runner/entities';
+import { compareLogs, resultCheck } from '../../tools/unified-spec-runner/match';
+import { ExpectedLogMessage } from '../../tools/unified-spec-runner/schema';
+
+describe('Unified Spec Runner', function () {
```
nbbeeken, line 9: some unit test failures, should be quick clean up

nbbeeken (changes requested):

nbbeeken (approved):
LGTM, just waiting on ci

PR #3579, opened 2023-02-27
test(NODE-5070): reduce fle prose 14 flakiness on sharded clusters

nbbeeken:
#### What is changing?

Trying a deleteMany before dropping and in the afterEach block.

We _may_ be running into this issue:
https://jira.mongodb.org/browse/SERVER-17397

> When dropping a database / collection in a sharded cluster, even if the drop is reported as successful it is possible the database / collection may still be present in some nodes in the cluster. In MongoDB 4.2 and later, rerunning the drop command should clean up the data. In MongoDB 4.0 and earlier, we do not recommend that users drop a database or collection and then attempt to reuse the namespace.

However we've seen this issue come up in before and after version 4.0. So I'm going with a deleteMany instead of doubling the drop operation. 

No flaky tests.


test/integration/client-side-encryption/client_side_encryption.prose.14.decryption_events.test.ts
```diff
@@ -218,6 +228,13 @@ describe('14. Decryption Events', metadata, function () {
       const collection = encryptedClient.db('db').collection('decryption_events');
       await collection.insertOne({ encrypted: malformedCiphertext });
 
+      /// Verify the malformedCiphertext was inserted with a plain client
+      const docCount = setupClient.db('db').collection('decryption_events').countDocuments({});
```
baileympearson, line 232: I think what we want here is not to count the documents, but rather to find all documents and assert that the malformed document is present.  If it is, and the aggregation fails, that indicates a bug in our FLE code. That's what I want to determine with this change

baileympearson (changes requested):

test/integration/client-side-encryption/client_side_encryption.prose.14.decryption_events.test.ts
baileympearson: Maybe we should add write concern majority to the insertion just in case too

baileympearson (changes requested):
missing await

baileympearson (approved):

baileympearson:
CI failures unrelated

PR #3582, opened 2023-02-28
fix(NODE-4854): set timeout on write and reset on message

nbbeeken:
#### What is changing?

The socket timeout:
- Is not set on freshly created connections
- Is set before writing to the socket
- Is cleared when recieving a message

Fixes a bug where connection in the pool were getting socket timeouts for being idle.

The commits are curated here to make reviewing the test changes easier, the file I worked in needed some TS love 💙 and some context blocks.  


src/cmap/connection.ts
```diff
@@ -316,6 +316,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       this[kDelayedTimeoutId] = null;
     }
 
+    this[kStream].setTimeout(0);
```
baileympearson, line 319: We can't only reset the time out here, because monitoring connections emit multiple messages for a single command.

Maybe we should move this logic into the if statement that checks for `moreToCome`.  That statement is already designed to handle the streaming protocol, so we can reset the socket time out there if needed

test/unit/cmap/connection.test.ts
```diff
@@ -31,10 +31,15 @@ const connectionOptionsDefaults = {
   loadBalanced: false
 };
 
-/** The absolute minimum socket API needed by Connection as of writing this test */
+/**
```
baileympearson, line 34: Where are the tests ensuring we unset the socket timeout on the connection?

nbbeeken: https://github.com/mongodb/node-mongodb-native/blob/NODE-4854-minPoolSizeReconnections/test/unit/cmap/connect.test.ts#L123-L246

baileympearson: Can we move those tests to `connection.test.ts`?  They're not connect related.

test/unit/cmap/connect.test.ts
```diff
@@ -0,0 +1,339 @@
...
+      try {
+        connection.onMessage(new BinMsg(msg, msgHeader, msgBody));
+      } catch {
+        // regardless of outcome
+      }
+      // timeout is still reset
+      expect(connection.stream).to.have.property('timeout', 0);
+    });
+
+    it('does not clear timeout after getting a message if moreToCome=true', async () => {
```
baileympearson, line 227: Can we also add tests demonstrating that for monitoring connections we do not clear the time out until the connection has no more messages (moreToCome is not present)?

baileympearson (changes requested):

test/unit/cmap/connect.test.ts
nbbeeken: Isn't that covered by the false case?

baileympearson: Oh yeah, I missed that. You're right

src/cmap/connection.ts
```diff
@@ -341,6 +341,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     }
 
     if (!operationDescription) {
```
baileympearson, line 343: Instead of unsetting the the socket timeout in two places, can we

- always unset it at the top of the method
- conditionally reset it from monitor connections with more responses?

nbbeeken: updated

baileympearson (approved):

nbbeeken:
Resolves NODE-2941

test/unit/cmap/connect.test.ts
```diff
@@ -0,0 +1,271 @@
...
+          },
+          //@ts-expect-error: Callbacks do not have mutual exclusion for error/result existence
+          callback
+        )
+      )().catch(error => error);
+
+      expect(error).to.match(/connection establishment was cancelled/);
+    });
+
+    it('interrupts connecting based on connectionTimeoutMS setting', async () => {
```
baileympearson, line 180:
```suggestion
   context('when connecting takes longer than connectTimeoutMS', () => {
      it('interrupts the connection with an error', async () => {
```

test/unit/cmap/connect.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    it('creates a connection with an infinite timeout', async () => {
+      expect(connection.stream).to.have.property('timeout', 0);
+    });
+
+    it('connection instance has property socketTimeoutMS equal to the value passed in the connectOptions', async () => {
+      expect(connection).to.have.property('socketTimeoutMS', 15000);
+    });
+
+    it.only('cancels connecting if provided cancellationToken emits cancel', async () => {
```
baileympearson, line 156:
```suggestion
   context('when the provided cancellation token emits cancel', () => {
     it('interrupts the connection with an error', () => {
```

baileympearson: and rogue .only

baileympearson (changes requested):

dariakp (approved):

baileympearson (changes requested):
failing unit tests

baileympearson (approved):

PR #3583, opened 2023-03-01
ci(NODE-5082): download node to local directory

nbbeeken:


.evergreen/install-dependencies.sh
```diff
@@ -1,105 +1,104 @@
...
-}
-
-NODE_MAJOR_VERSION=$(node_lts_to_version $NODE_LTS_NAME)
-NODE_VERSION=$(latest_version_for_node_major $NODE_MAJOR_VERSION)
-NODE_VERSION=${NODE_VERSION:1} # :1 gets rid of the leading 'v'
-
-echo "set version to $NODE_VERSION"
-
-# output node version to expansions file for use in subsequent scripts
+NODE_ARTIFACTS_PATH="${PROJECT_DIRECTORY:-$(pwd)}/node-artifacts"
```
baileympearson, line 5: Would it be better to make this required? If project directory is not specified, something is wrong. Defaulting to the current working directory silently doesn't feel like the best approach

.evergreen/install-dependencies.sh
```diff
@@ -1,105 +1,104 @@
...
+
+mkdir -p "$NODE_ARTIFACTS_PATH/npm_global"
+
+# Comparisons are all case insensitive
+shopt -s nocasematch
+
+# index.tab is a sorted tab separated values file with the following headers
+# 0       1    2     3   4  5  6    7       8       9   10
+# version date files npm v8 uv zlib openssl modules lts security
+curl --retry 8 -sS "https://nodejs.org/dist/index.tab" --max-time 300 --output node_index.tab
```
baileympearson, line 18: This is a long script, could we consider breaking it into smaller functions? Or moving the logic that determines the download URL into a separate file? That might help with the readability and modularity of the script.

baileympearson (changes requested):

.evergreen/install-dependencies.sh
nbbeeken: PROJECT_DIRECTORY is calculated from pwd. I don't think we need to make it a prereq, this makes it easier to run the script locally

.evergreen/install-dependencies.sh
nbbeeken: It's shorter than the script that proceeded it, I'd like to keep it in one file to make the porting and future updating easier

baileympearson: I don't really care about the length, if it's well organized. This file is just dense and the code is opaque. Some logical grouping might make it easier to work with in the future?

baileympearson: If we use bash functions, we can then source this locally and test each component

nbbeeken: Is the green CI not testing enough? we can't proceed if any step in this script fails, which part is opaque, can I improve variable names?

baileympearson: All of the logic related to finding the correct version of Node and downloading it is a little confusing, and since there's no clear boundary as to when it starts and stops, it's hard to tell when you're reading the script when that portion of code is finished. This is why we use functions, to logically group related functionality and to make understanding the higher-level purpose and logic of a script easier to GROK at a glance  

From a CI maintenance standpoint, ideally we never touch this code again. But if we do have to, I'd rather that it was logically group so that it's easier to reason about at the high level without reading and thoroughly understand the entire script.

And yes, you're right, a green CI means the script is passing. But if we ever have to touch this again, it would be nice if we could source the file and test it in a piecemeal fashion, as opposed to testing the script in its entirety or commenting out large chunks of code to test parts of it.

baileympearson (approved):

PR #3589, opened 2023-03-03
feat(NODE-5036): reauthenticate OIDC and retry

durran:
### Description

Enables reauthentication on the same connection with retry.

#### What is changing?

- Adds the ability to retry a command on the same connection when getting a reauthentication error.
- Adds reauthentication prose tests.
- Adds reauthentication unified auth tests.
- Marks OIDC as experimental

None

NODE-5036

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
     "check:atlas": "mocha --config test/manual/mocharc.json test/manual/atlas_connectivity.test.js",
     "check:adl": "mocha --config test/mocha_mongodb.json test/manual/atlas-data-lake-testing",
     "check:aws": "mocha --config test/mocha_mongodb.json test/integration/auth/mongodb_aws.test.ts",
-    "check:oidc": "TRACE_SOCKETS=true mocha --config test/manual/mocharc.json test/manual/mongodb_oidc.prose.test.ts",
+    "check:oidc": "mocha --config test/manual/mocharc.json test/manual/mongodb_oidc.prose.test.ts",
     "check:ocsp": "mocha --config test/manual/mocharc.json test/manual/ocsp_support.test.js",
     "check:kerberos": "mocha --config test/manual/mocharc.json test/manual/kerberos.test.js",
     "check:tls": "mocha --config test/manual/mocharc.json test/manual/tls_support.test.js",
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
     "check:atlas": "mocha --config test/manual/mocharc.json test/manual/atlas_connectivity.test.js",
     "check:adl": "mocha --config test/mocha_mongodb.json test/manual/atlas-data-lake-testing",
     "check:aws": "mocha --config test/mocha_mongodb.json test/integration/auth/mongodb_aws.test.ts",
-    "check:oidc": "mocha --config test/manual/mocharc.json test/manual/mongodb_oidc.prose.test.ts",
+    "check:oidc": "TRACE_SOCKETS=true mocha --config test/manual/mocharc.json test/manual/mongodb_oidc.prose.test.ts",
     "check:ocsp": "mocha --config test/manual/mocharc.json test/manual/ocsp_support.test.js",
     "check:kerberos": "mocha --config test/manual/mocharc.json test/manual/kerberos.test.js",
     "check:tls": "mocha --config test/manual/mocharc.json test/manual/tls_support.test.js",
--- a/src/cmap/auth/auth_provider.ts
+++ b/src/cmap/auth/auth_provider.ts
@@ -5,14 +5,20 @@ import type { HandshakeDocument } from '../connect';
 import type { Connection, ConnectionOptions } from '../connection';
 import type { MongoCredentials } from './mongo_credentials';
 
+/** @internal */
 export type AuthContextOptions = ConnectionOptions & ClientMetadataOptions;
 
-/** Context used during authentication */
+/**
+ * Context used during authentication
+ * @internal
+ */
 export class AuthContext {
   /** The connection to authenticate */
   connection: Connection;
   /** The credentials to use for authentication */
   credentials?: MongoCredentials;
+  /** If the context if for reauthentication. */
+  reauthenticating = false;
   /** The options passed to the `connect` method */
   options: AuthContextOptions;
 
--- a/src/cmap/auth/auth_provider.ts
+++ b/src/cmap/auth/auth_provider.ts
@@ -63,4 +63,18 @@ export class AuthProvider {
     // TODO(NODE-3483): Replace this with MongoMethodOverrideError
     callback(new MongoRuntimeError('`auth` method must be overridden by subclass'));
   }
+
+  /**
+   * Reauthenticate.
+   * @param context - The shared auth context.
+   * @param callback - The callback.
+   */
+  reauth(context: AuthContext, callback: Callback): void {
+    context.reauthenticating = true;
+    const cb: Callback = (error, result) => {
+      context.reauthenticating = false;
+      callback(error, result);
+    };
+    this.auth(context, cb);
+  }
 }
--- a/src/cmap/auth/auth_provider.ts
+++ b/src/cmap/auth/auth_provider.ts
@@ -17,7 +17,7 @@ export class AuthContext {
   connection: Connection;
   /** The credentials to use for authentication */
   credentials?: MongoCredentials;
-  /** If the context if for reauthentication. */
+  /** If the context is for reauthentication. */
   reauthenticating = false;
   /** The options passed to the `connect` method */
   options: AuthContextOptions;
@@ -70,6 +70,10 @@ export class AuthProvider {
    * @param callback - The callback.
    */
   reauth(context: AuthContext, callback: Callback): void {
+    // If we are already reauthenticating this is a no-op.
+    if (context.reauthenticating) {
+      return callback(new MongoRuntimeError('Reauthentication already in progress.'));
+    }
     context.reauthenticating = true;
     const cb: Callback = (error, result) => {
       context.reauthenticating = false;
--- a/src/cmap/auth/mongo_credentials.ts
+++ b/src/cmap/auth/mongo_credentials.ts
@@ -37,11 +37,20 @@ export interface AuthMechanismProperties extends Document {
   SERVICE_REALM?: string;
   CANONICALIZE_HOST_NAME?: GSSAPICanonicalizationValue;
   AWS_SESSION_TOKEN?: string;
-  /** @experimental */
+  /**
+   * @alpha
+   * @experimental
+   */
   REQUEST_TOKEN_CALLBACK?: OIDCRequestFunction;
-  /** @experimental */
+  /**
+   * @alpha
+   * @experimental
+   */
   REFRESH_TOKEN_CALLBACK?: OIDCRefreshFunction;
-  /** @experimental */
+  /**
+   * @alpha
+   * @experimental
+   */
   PROVIDER_NAME?: 'aws';
 }
 
--- a/src/cmap/auth/mongo_credentials.ts
+++ b/src/cmap/auth/mongo_credentials.ts
@@ -37,20 +37,11 @@ export interface AuthMechanismProperties extends Document {
   SERVICE_REALM?: string;
   CANONICALIZE_HOST_NAME?: GSSAPICanonicalizationValue;
   AWS_SESSION_TOKEN?: string;
-  /**
-   * @alpha
-   * @experimental
-   */
+  /** @experimental */
   REQUEST_TOKEN_CALLBACK?: OIDCRequestFunction;
-  /**
-   * @alpha
-   * @experimental
-   */
+  /** @experimental */
   REFRESH_TOKEN_CALLBACK?: OIDCRefreshFunction;
-  /**
-   * @alpha
-   * @experimental
-   */
+  /** @experimental */
   PROVIDER_NAME?: 'aws';
 }
 
--- a/src/cmap/auth/mongo_credentials.ts
+++ b/src/cmap/auth/mongo_credentials.ts
@@ -37,8 +37,11 @@ export interface AuthMechanismProperties extends Document {
   SERVICE_REALM?: string;
   CANONICALIZE_HOST_NAME?: GSSAPICanonicalizationValue;
   AWS_SESSION_TOKEN?: string;
+  /** @experimental */
   REQUEST_TOKEN_CALLBACK?: OIDCRequestFunction;
+  /** @experimental */
   REFRESH_TOKEN_CALLBACK?: OIDCRefreshFunction;
+  /** @experimental */
   PROVIDER_NAME?: 'aws';
 }
 
--- a/src/cmap/auth/mongodb_oidc/workflow.ts
+++ b/src/cmap/auth/mongodb_oidc/workflow.ts
@@ -8,7 +8,11 @@ export interface Workflow {
    * All device workflows must implement this method in order to get the access
    * token and then call authenticate with it.
    */
-  execute(connection: Connection, credentials: MongoCredentials): Promise<Document>;
+  execute(
+    connection: Connection,
+    credentials: MongoCredentials,
+    reauthenticate?: boolean
+  ): Promise<Document>;
 
   /**
    * Get the document to add for speculative authentication.
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -579,7 +579,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     fn: WithConnectionCallback,
     callback: Callback<Connection>
   ) {
-    if ((fnErr as MongoError).code === MONGODB_ERROR_CODES.Reauthenticate) {
+    if (fnErr instanceof MongoError && fnErr.code === MONGODB_ERROR_CODES.Reauthenticate) {
       this.reauthenticate(conn, fn, (error, res) => {
         if (error) {
           return callback(error);
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -548,8 +548,9 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
         if (fnErr) {
           return this.withReauthentication(fnErr, conn, fn, callback);
         }
-        return callback(undefined, result);
+        callback(undefined, result);
       });
+      return;
     }
 
     this.checkOut((err, conn) => {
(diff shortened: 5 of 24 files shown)


src/cmap/connection_pool.ts
```diff
@@ -544,7 +546,17 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       fn(undefined, conn, (fnErr, result) => {
         if (typeof callback === 'function') {
           if (fnErr) {
-            callback(fnErr);
+            if ((fnErr as MongoError).code === MONGODB_ERROR_CODES.Reauthenticate) {
+              this.reauthenticate(conn, fn, (error, res) => {
+                if (error) {
+                  callback(error);
```
baileympearson, line 552: I don't think this is quite right - when no callback is supplied, we don't attempt to re-authenticate?

We only use `withConnection` in server.command and a callback is always provided.  I think we should make `callback` required, which would make the guard on line 547 unnecessary and solve this issue (and remove a layer of nesting along the way 🙂).  thoughts?

src/cmap/connection_pool.ts
```diff
@@ -544,7 +546,17 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+                } else {
```
baileympearson, line 553: It would be nice to remove the duplicated logic and the unnecessary nesting.  From slack - Durran is looking into a potential `withReauthentication` approach that can be used to wrap `fn` inside of `withConnection`.

src/cmap/connection_pool.ts
```diff
@@ -572,6 +594,50 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+  private reauthenticate(
+    connection: Connection,
+    fn: WithConnectionCallback,
+    callback: Callback
+  ): void {
+    const authContext = connection.authContext;
+    if (!authContext) {
+      return callback(new MongoRuntimeError('No auth context found on connection.'));
+    }
+    authContext.reauthenticating = true;
```
baileympearson, line 609: just a thought to avoid leaking details about how the auth context manages state when re-authenticating.  what if we added a `reauth` method the `AuthProvider`?

```typescript
class MongoDBOIDC {
  reauth(authContext, callback) {
    authContext.reauthenticating = true;
    const _cb = (error, result) => {
      authContext.reauthenticating = false;
      _cb(error, result)
    }
    authContext.execute(authContext, _cb)
  }
}
```

another advantage of this approach is that it immediately becomes clear which auth mechanisms support re-authentication, because `reauth` could throw an error if we attempt to reauth using a mechanism that doesn't support it

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -452,5 +452,153 @@ describe('MONGODB-OIDC', function () {
...
+        }
+      };
+
+      before(function () {
+        // - Clear the cache
+        cache.clear();
+        // - Create a client with the callbacks and an event listener capable of
+        //   listening for SASL commands
+        //
+        // - TODO(NODE-3494): Driver does not observe sensitive commands.
```
baileympearson, line 515: Can you clarify this, comment? We don't observe sensitive commands, we do redact them, but we still emit command events for them (and that's how this test is passing).  given that the prose test only asserts that the events were emitted and nothing about the payload of the command, we could consider removing the comment entirely.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -452,5 +452,153 @@ describe('MONGODB-OIDC', function () {
...
+          authMechanismProperties: {
+            REQUEST_TOKEN_CALLBACK: requestCallback,
+            REFRESH_TOKEN_CALLBACK: refreshCallback
+          },
+          monitorCommands: true
+        });
+        client.on('commandStarted', commandStarted);
+        client.on('commandSucceeded', commandSucceeded);
+        client.on('commandFailed', commandFailed);
+        collection = client.db('test').collection('test');
```
baileympearson, line 526: General question about these test. Do you know why they are simply asserting on the correct number of events?  It seems like the ordering of events should be exact:

find start, find failed, sasl start, sasl suceeded, find start, find succeeded

asserting on the counts of events does nothing to assert on the correct behavior - this ordering would also pass (and is the ordering that would happen in a normal retry, not reauth)

sasl start, sasl succeeded, find start, find failed, find start, find succeeded

src/cmap/auth/mongodb_oidc/workflow.ts
```diff
@@ -8,7 +8,11 @@ export interface Workflow {
    * All device workflows must implement this method in order to get the access
    * token and then call authenticate with it.
    */
-  execute(connection: Connection, credentials: MongoCredentials): Promise<Document>;
+  execute(
+    connection: Connection,
+    credentials: MongoCredentials,
+    reauthenticate?: boolean
```
baileympearson, line 14: Personal preference - I think flags are more clear than booleans.  Feel free to ignore and close this comment
```suggestion
    mode: 'authenticate' | 'reauthenticate' = 'authenticate'
```

baileympearson: if we do address this though, we should make this change everywhere we use `reauthenticate` or `reauthenticating` in this PR

baileympearson (changes requested):

test/manual/mongodb_oidc.prose.test.ts
durran: I actually just need to delete the comment since this was implemented but the valid-pass tests in NODE-3494 aren't synced.

src/cmap/connection_pool.ts
durran: Added private `withReauthentication` now.

src/cmap/connection_pool.ts
durran: Added `reauth` to the `AuthProvider` but not throwing an error since all mechanisms can reauth - there's just a slight difference in behaviour when using OIDC with the cache. All other mechanisms just start auth over from the beginning.

test/manual/mongodb_oidc.prose.test.ts
durran: DRIVERS-2415 has a lot of changes to the prose tests around this area. May be better to wait for that to finish and then re-evaluate.

src/cmap/auth/mongodb_oidc/workflow.ts
durran: I think since it's an "on" or "off" scenario I feel like the boolean is fine since the property name explains what the state is.

src/cmap/connection_pool.ts
```diff
@@ -537,32 +540,29 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
       // use the provided connection, and do _not_ check it in after execution
       fn(undefined, conn, (fnErr, result) => {
-        if (typeof callback === 'function') {
-          if (fnErr) {
-            callback(fnErr);
-          } else {
-            callback(undefined, result);
-          }
+        if (fnErr) {
+          this.withRequthentication(fnErr, conn, fn, callback);
```
baileympearson, line 549:
```suggestion
          this.withReauthentication(fnErr, conn, fn, callback);
```
🙂

test/manual/mongodb_oidc.prose.test.ts
baileympearson: sounds good - I left a comment on the PR.

baileympearson: can you add a ticket to the OIDC support epic to incorporate changes from that PR (if we don't have one)?  I don't want to lose track of that work.

baileympearson (changes requested):
just putting some review up (still need to work out CI).

src/cmap/connection_pool.ts
durran: Updated.

durran:
https://github.com/mongodb/node-mongodb-native/pull/3589#discussion_r1139101249

NODE-5122

durran:
https://spruce.mongodb.com/task/mongo_node_driver_next_rhel80_large_node_latest_test_auth_oidc_patch_764972247737ddc868a213619522dfd8abba4c27_6414640f562343536ab6c6bb_23_03_17_12_59_20/tests?execution=5&sortBy=STATUS&sortDir=ASC

src/cmap/connection_pool.ts
```diff
@@ -572,6 +573,66 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+   * Reauthenticate on the same connection and then retry the operation.
+   */
+  private reauthenticate(
+    connection: Connection,
+    fn: WithConnectionCallback,
+    callback: Callback
+  ): void {
+    const authContext = connection.authContext;
+    if (!authContext) {
+      return callback(new MongoRuntimeError('No auth context found on connection.'));
```
nbbeeken, line 604: There's a number of synchronous callback invocations, are we sure that's the correct approach? I could see us implementing this using async/await and wrapping the old APIs in promisify.

- reauth could promisify auth, and use try/finally to set/unset the 
- we can promisify calling the `fn` 
- `withReauthentication` can return a promise and we can invoke the withConnection callback in a `.then`

thoughts?

src/index.ts
```diff
@@ -197,6 +197,7 @@ export type {
   ResumeToken,
   UpdateDescription
 } from './change_stream';
+export { AuthContext, AuthContextOptions } from './cmap/auth/auth_provider';
```
nbbeeken, line 200:
```suggestion
export type { AuthContext, AuthContextOptions } from './cmap/auth/auth_provider';
```

this will also remove the need to add these to the `index.test.ts` file.

package.json
```diff
@@ -128,7 +128,7 @@
     "check:atlas": "mocha --config test/manual/mocharc.json test/manual/atlas_connectivity.test.js",
     "check:adl": "mocha --config test/mocha_mongodb.json test/manual/atlas-data-lake-testing",
     "check:aws": "mocha --config test/mocha_mongodb.json test/integration/auth/mongodb_aws.test.ts",
-    "check:oidc": "mocha --config test/manual/mocharc.json test/manual/mongodb_oidc.prose.test.ts",
+    "check:oidc": "TRACE_SOCKETS=true mocha --config test/manual/mocharc.json test/manual/mongodb_oidc.prose.test.ts",
```
nbbeeken, line 131: Unfortunately the manual test config doesn't bring in the leak checker

you can add 
```json
"require": [
    "test/tools/runner/hooks/leak_checker.ts"
]
```
to `test/manual/mocharc.json`

nbbeeken (changes requested):

src/cmap/auth/mongo_credentials.ts
```diff
@@ -37,8 +37,20 @@ export interface AuthMechanismProperties extends Document {
   SERVICE_REALM?: string;
   CANONICALIZE_HOST_NAME?: GSSAPICanonicalizationValue;
   AWS_SESSION_TOKEN?: string;
+  /**
+   * @alpha
```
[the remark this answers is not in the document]
durran: The interesting thing was API extractor was failing when I used @experimental  - it said only @alpha, @beta, @public and @internal were valid.

package.json
durran: Going to remove that as I found the leaked socket.

src/cmap/auth/mongo_credentials.ts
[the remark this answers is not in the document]
nbbeeken: Strange that you'd be seeing this now, since we already have some, api-extractor respects the rules set up in tsdoc.json so we can add it there if you are still seeing errors

src/index.ts
durran: Updated.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -452,5 +452,172 @@ describe('MONGODB-OIDC', function () {
...
+          // Perform another find operation.
+          await collection.findOne();
+        });
+
+        after(async function () {
+          await client.db('admin').command({
+            configureFailPoint: 'failCommand',
+            mode: 'off'
+          });
+          console.log('clearing cache');
```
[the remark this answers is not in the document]
durran: console logs removed.

package.json
durran: Done

src/cmap/connection_pool.ts
```diff
@@ -572,6 +573,66 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     });
   }
 
+  private withReauthentication(
+    fnErr: AnyError,
+    conn: Connection,
+    fn: WithConnectionCallback,
+    callback: Callback<Connection>
+  ) {
+    if ((fnErr as MongoError).code === MONGODB_ERROR_CODES.Reauthenticate) {
```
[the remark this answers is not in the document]
durran: Updated.

src/cmap/auth/mongo_credentials.ts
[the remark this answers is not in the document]
durran: When I add to tsdoc.json it says the experimental tag already exists. When I just use it then the error is:

```
Error: src/cmap/auth/mongodb_oidc.ts:15:1 - (ae-missing-release-tag) "OIDCMechanismServerStep1" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
Error: src/cmap/auth/mongodb_oidc.ts:25:1 - (ae-missing-release-tag) "OIDCRequestTokenResult" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
Error: src/cmap/auth/mongodb_oidc.ts:32:1 - (ae-missing-release-tag) "OIDCRequestFunction" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
Error: src/cmap/auth/mongodb_oidc.ts:39:1 - (ae-missing-release-tag) "OIDCRefreshFunction" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
```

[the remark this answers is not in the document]
durran: Updated to @public and @experimental

src/cmap/connection_pool.ts
durran: I think that's a good idea but maybe we can put that in a separate refactoring ticket? When I first went down that path I started leaking connections and felt the refactor was getting a bit more out of scope from just this ticket.

src/index.ts
nbbeeken: the new fields in `index.test.ts` should be reverted now, they should be going red

src/cmap/connection_pool.ts
nbbeeken: I am targeting all the new code here with the suggested changes, just trying to make all new stuff async await and we can use .then & etc to put it inside of callback existing code. However, I don't feel super strongly so a follow up sounds good

dariakp: I think in the future this is something that we should try to capture in grooming or at least in the kickoff notes - the intent to use async/await or refactor - to help set the scope of the work and align with the team early on. As is, and considering that right now we want to have these changes in place and passing the new prose tests, a follow up ticket seems like the most viable approach - though I'm not certain where that work would live. In general, we do want to try to account for the async/await refactor when estimating new tickets because it really helps with the async stack traces (which was the original motivation for moving the driver to async/await).

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -452,5 +452,164 @@ describe('MONGODB-OIDC', function () {
         });
       });
     });
+
+    // The driver MUST test reauthentication with MONGODB-OIDC for a read operation.
+    describe('6. Reauthentication', function () {
+      let refreshInvokations = 0;
```
dariakp, line 458:
```suggestion
      let refreshInvocations = 0;
```

tsdoc.json
```diff
@@ -17,5 +20,11 @@
             "syntaxKind": "block",
             "tagName": "@category"
         }
-    ]
-}
+    ],
+    "supportForTags": {
```
dariakp, line 24: what are all these for? and why the update to tsdoc config?

dariakp (changes requested):

tsdoc.json
durran: I was chatting with @nbbeeken about the API extractor and that's what the updated config will look like. I'll remove from this PR.

nbbeeken: I originally thought we were getting errors with the experimental tag, we don't need this here, but would be a good improvement to add to get stricter validation on the permitted tags. It will pull in the settings supported by api-extractor. I can file a ticket, would be a quick win

test/manual/mongodb_oidc.prose.test.ts
durran: Updated

src/cmap/auth/auth_provider.ts
```diff
@@ -5,14 +5,20 @@ import type { HandshakeDocument } from '../connect';
...
+/**
+ * Context used during authentication
+ * @internal
+ */
 export class AuthContext {
   /** The connection to authenticate */
   connection: Connection;
   /** The credentials to use for authentication */
   credentials?: MongoCredentials;
+  /** If the context if for reauthentication. */
```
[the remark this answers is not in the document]
durran: Updated.

src/cmap/auth/auth_provider.ts
```diff
@@ -57,4 +63,18 @@ export class AuthProvider {
     // TODO(NODE-3483): Replace this with MongoMethodOverrideError
     callback(new MongoRuntimeError('`auth` method must be overridden by subclass'));
   }
+
+  /**
+   * Reauthenticate.
+   * @param context - The shared auth context.
+   * @param callback - The callback.
+   */
+  reauth(context: AuthContext, callback: Callback): void {
```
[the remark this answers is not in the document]
durran: Thanks for catching this. I now call the callback with a runtime error if it's already reauthenticating - it shouldn't enter it once already happening.

src/cmap/connection_pool.ts
durran: I generally try to async/await everything when introducing new code, and refactoring the surrounding code as well. (See recent PRs in this area). This particular area was a little bit more risky to me in this PR as everything in the driver runs through this path - once some leaks started happening I aborted. I think maybe this can go in a ticket/epic around refactoring the connection layer?

dariakp: yup, sounds good to me

nbbeeken (approved):

PR #3595, opened 2023-03-09
fix(NODE 4518)!: Bug Fix - Unclear Error Message AuthSource isn't set to $external. 

shreykumar18:
#### What is changing?

Instead of a "[object Object] can only have authSource set to '$external'", where obviously "[object Object]" error message, when a connection string is provided, the object is now clarified.

No new documentation needed.

It was a bug, that provided an unclear error message. The expected behaviour was to specify the object of the connection string when the authSource isn't set to $external.


baileympearson:
Hey @shreykumar18 - thanks for bringing our attention to this bug again.  We decided to go with a slightly different approach, implemented [here](https://github.com/mongodb/node-mongodb-native/pull/3597).  Thanks again!

PR #3596, opened 2023-03-10
fix(NODE-5106): prevent multiple mongo client connect()s from leaking topology

clemclx:
fix(NODE-5106): Add MongoClient connect lock mechanism to avoid topology leak

### Description

Related to the ticket opened here : https://jira.mongodb.org/browse/NODE-5106
As mentionned in the issue, not test provided because atm a test cannot be performed according to the beforeHooks awaiting the MongoClient connect.


src/mongo_client.ts
```diff
@@ -409,46 +411,57 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
-      await options.encrypter.connectInternalClient();
-    } else {
-      await topologyConnect();
-    }
+      return this;
+    })();
 
+    await this.connectionLock;
+    // release
+    this.connectionLock = undefined;
```
addaleax, line 464: The lock is never released if the wrapped block throws. That doesn't seem correct (you probably want try/finally here)?

addaleax (commented):
This generally seems like the right thing to do, but I think you'll need to add a test for this. Somebody from the team can probably confirm.

You probably don't need all the complexity of the `resource_clean_up.test.ts` file, I think you could just ensure that e.g. when you call `client.connect()` twice in the same tick, you'll stil only get one `open` event on the MongoClient.

clemclx:
Yes I clarified in the ticket why I did not add a test.
Maybe it is not necessary to add it to the `resource_clean_up.test.ts` yes, I let someone of the team with a much better knowledge of the code base decide where it belongs.

I remain available :relaxed:

src/mongo_client.ts
```diff
@@ -409,44 +411,59 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
       return this;
     }
 
-    const options = this[kOptions];
+    if (this.connectionLock) {
```
baileympearson, line 414: Hey @clemclx - thanks for the fix!  I think we're on the right track here.

We can't use them (yet) but I think a decorator might be a good fit here.  

```typescript
class MongoClient {
  ...
  @lock 
  connect() { 
    // no changes to method body here
  }
}
```

What do you think of a slight refactor here, to make the refactor to use a generator (when it's available) straightforward?  I was thinking we could

- move the entire (old) body of `connect` into a private `_connect` method
- make the current `connect` method enforce locking on the `_connect` method

Then, porting the logic to a `@lock` generator is straightforward because we just move the code from `_connect` back into `connect` and move the locking logic into the generator.  Thoughts?

---

as far as testing, I think Anna is correct.  It should be enough to parallelize a few connect invocations and assert that

- we successfully connect
- we only receive one `Topology.OPEN` event

These tests can live in `integration/mongo_client.test.ts`.

lmk if you want me to add the tests for you but feel free to add them if you'd like to!

baileympearson (changes requested):

src/mongo_client.ts
clemclx: Hey @baileympearson, thanks for your review !
I agree that a decorator would fit pretty well here, make sense. Did not took this way as I did not see any decorator in the driver codebase (you precises that it's not supported yet anyway).

I refactored the code to fit your needs, let me know if I correctly understood your request.

Moreover, I added the test suite as you suggested but I think there is something wrong with the leak_checker hook (`tools/runner/hooks/leak_checker.ts`)...
The `afterEach` hook fails on the first test as it seems to count the number of `connect` called on the prototype (instead maybe of couting the effective number of connections opens ?). The problem is that calling `client.connect` thrice to make concurrent calls increase the counter to 3 but calling `client.close` once make the assertion fails as it only decrease one.
A workaround could be to call the `client.close` as many times as `client.connect` but I really don't think it's the goal here.
I don't have the big picture of this hook to properly understand its purpose and constraints, could you check it please ?

Let me know if I can help out !

baileympearson: Hey @clemclx - yup, this is basically what I had in mind.

That's an interesting point about the leak checker.  Historically, we've leaked connections by not closing the client so the leak checker was added to ensure that we close every client we open in our tests.  I think it's fine to close it as many times as we need to for these tests (and add a comment explaining why), since this is a unique test where we explicitly connect a single client multiple times.

clemclx: Hello @baileympearson ! Thanks for the clarification about the leak_checker.
I applied the fix to tests in consequences.
We should be fine now, let me know if I can help for something else :smiley:

test/integration/mongo_client.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it('Concurrents client connect correctly locked (only one topology created)', async function () {
+    await Promise.all([clientConnect(), clientConnect(), clientConnect()]);
+
+    expect(topologyOpenEvents).to.have.lengthOf(1);
+    expect(client.topology?.isConnected()).to.be.true;
+  });
+
+  it('Failed client connect must properly release lock', async function () {
```
nbbeeken, line 45:
```suggestion
  it('when connect rejects lock is released regardless', async function () {
```

Just trying to clarify what "properly released" means here

test/integration/mongo_client.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    internalConnectStub.restore();
+
+    // second call should connect
+    try {
+      await clientConnect();
+    } catch (err) {
+      expect.fail(`client connect throwed unexpected error`);
+    }
```
nbbeeken, line 63:
```suggestion
    // second call should connect
    await clientConnect();
```

We can just let the potential rejection end the test, that way it will print out whatever error was thrown from this call

nbbeeken (changes requested):
Thanks so much @clemclx really appreciate your help with this! I've got some ideas to improve the testing a bit here:

test/integration/mongo_client.test.ts
```diff
@@ -0,0 +1,68 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+
+import { MongoClient } from '../../src';
+
+describe('MongoClient', () => {
```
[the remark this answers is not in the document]
clemclx: Yes could be better, will move it asap

test/integration/mongo_client.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    client = this.configuration.newClient();
+    topologyOpenEvents = [];
+    clientConnectCounter = 0;
+    client.on('open', event => topologyOpenEvents.push(event));
+  });
+
+  afterEach(async function () {
+    /** Close as many times as connect calls in the runned test (tracked by clientConnectCounter) */
+    const clientClosePromises = [...new Array(clientConnectCounter)].map(() => client.close());
+    await Promise.all(clientClosePromises);
```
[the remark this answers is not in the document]
nbbeeken: Appreciate the reference! I understand why we have to close more than once because of the leak checker, sorry about that limitation. 😅  

I'm wondering if `.close` also has a parallelism issue. Supposedly we could be calling `.end` on sockets more than once because as the first "`client.close`" promise is pending the I/O for socket closure the next promise could also synchronously reach the `.end` invocation. 

This may in fact _not_ be the case but _if_ it is we'll want to also fix that in a follow up ticket perhaps.  

I contrived an example, I have to use sleep 1ms to make the promises offset by a bit but I would imagine in real world scenario a delay like that could occur depending on what else is in the event loop.

<details><summary>Example script that calls .end in parallel</summary>

```js
import { once } from 'events';
import { Socket } from 'net';
import { setTimeout as sleep } from 'timers/promises';
import { promisify } from 'util';

const socket = new Socket();
const willConnect = once(socket, 'connect');
socket.connect({ host: '127.0.0.1', port: 2390 }).setDefaultEncoding('utf8').setEncoding('utf8');
await willConnect;

await promisify(callback => socket.write('hello\n', callback))();

// MongoClient .close in parallel
await Promise.all([
  promisify(callback => socket.end(callback))(),
  sleep(1).then(() => promisify(callback => socket.end(callback))())
]);
```

```txt
Error [ERR_STREAM_ALREADY_FINISHED]: Cannot call end after a stream was finished
    at new NodeError (node:internal/errors:393:5)
    at Writable.end (node:internal/streams/writable:637:11)
    at Socket.end (node:net:691:31)
  code: 'ERR_STREAM_ALREADY_FINISHED'
}
```
</details>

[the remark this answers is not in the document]
nbbeeken: https://jira.mongodb.org/browse/NODE-5138

clemclx:
Hey @nbbeeken,
I updated the PR according to your review, thanks !
Let me know if I can be useful for something else.

nbbeeken (approved):

nbbeeken (approved):

baileympearson (approved):

PR #3600, opened 2023-03-14
fix(NODE-3998): dont duplicate client metadata

durran:
### Description

Refactors handshake generation and adds tests for duplicate client metadata.

#### What is changing?
- Refactors handshake generation to use decorators.
- Adds unit tests around handshake generation.

None

NODE-3998


.eslintrc.json
```diff
@@ -87,6 +87,7 @@
       "global"
     ],
     "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/require-await": "off",
```
durran, line 90: As we discussed to make all the functions async without needing await in them.

src/cmap/connect.ts
```diff
@@ -117,157 +122,112 @@ function performInitialHandshake(
...
-      );
-    }
-    return provider.prepare(handshakeDoc, authContext, callback);
-  }
-  callback(undefined, handshakeDoc);
+export async function prepareHandshakeDocument(
+  authContext: AuthContext
+): Promise<HandshakeDocument> {
+  const { options } = authContext;
+  const generator = options.monitorHandshake
```
durran, line 224: The Monitor doesn't have an auth context - so we set an option to tell connect not to include the auth decorator.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -107,58 +98,45 @@ export class MongoDBOIDC extends AuthProvider {
...
+    } catch (error) {
+      callback(error);
+    }
   }
 
   /**
    * Add the speculative auth for the initial handshake.
    */
-  override prepare(
+  override async prepare(
```
durran, line 109: Made the `prepare` method in all auth providers async now.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -107,58 +98,45 @@ export class MongoDBOIDC extends AuthProvider {
...
+    const result = await workflow.speculativeAuth();
+    return { ...handshakeDoc, ...result };
   }
 }
 
 /**
  * Gets either a device workflow or callback workflow.
  */
-function getWorkflow(credentials: MongoCredentials, callback: Callback<Workflow>): void {
+function getWorkflow(credentials: MongoCredentials): Workflow {
```
durran, line 133: Removed the use of callbacks here.

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,94 @@
+import * as os from 'os';
```
durran, line 1: Cleanup to move all handshake related metadata into one location.

src/cmap/handshake/monitor_handshake_decorator.ts
```diff
@@ -0,0 +1,56 @@
...
+  ) {
+    this.monitorOptions = monitorOptions;
+    this.topologyVersion = topologyVersion;
+    this.helloOk = helloOk;
+  }
+  async decorate(handshake: HandshakeDocument): Promise<HandshakeDocument> {
+    // If the server response has helloOk: true, the monitor can switch to hello.
+    if (this.helloOk) {
+      delete handshake[LEGACY_HELLO_COMMAND];
+      handshake.hello = 1;
```
durran, line 35: Order matters here - `hello` needs to be the first command in the document.

src/cmap/handshake/handshake_document.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import type { ClientMetadata } from './client_metadata';
+
+/** @internal */
+export interface HandshakeDocument extends Document {
+  /**
+   * @deprecated Use hello instead
+   */
+  ismaster?: boolean | 0 | 1;
+  hello?: boolean | 0 | 1;
```
nbbeeken, line 11: We are careful to make sure these are set to 1 now, it was a v5 change, and also when would they be set to zero?

src/cmap/handshake/handshake_document.ts
```diff
@@ -0,0 +1,18 @@
...
+  helloOk?: boolean;
+  client?: ClientMetadata;
+  compression?: string[];
```
nbbeeken, line 14:
```suggestion
  compression?: CompressorName[];
```

src/cmap/handshake/handshake_document.ts
```diff
@@ -0,0 +1,18 @@
...
+  saslSupportedMechs?: string;
```
nbbeeken, line 15: union of string literals here?

src/cmap/handshake/handshake_document.ts
```diff
@@ -0,0 +1,18 @@
+import type { Document } from 'bson';
```
nbbeeken, line 1: Can we move this into another file? I would like to refrain from adding type only files as it will generate an empty javascript files. I suggest removing `default_` from the base handshake decorator and we can have these types exported from that file.

src/cmap/handshake/handshake_decorator.ts
```diff
@@ -0,0 +1,13 @@
+import type { AuthContext } from '../auth/auth_provider';
```
nbbeeken, line 1: type only file, same as other comment on the subject.

src/cmap/handshake/default_handshake_decorator.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      // Determine if we send hello or legacy hello.
+      handshake[serverApi?.version ? 'hello' : LEGACY_HELLO_COMMAND] = 1;
+
+      // Get compressor info.
+      const compressors = options.compressors ? options.compressors : [];
+
+      // Set metadata and compressor information.
+      handshake.client = options.metadata || makeClientMetadata(options);
+      handshake.compression = compressors;
```
nbbeeken, line 29: There's more than I would expect in this section, if an auth context exists why would it relate to setting compression, client, loadBalanced settings?

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,94 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const NODE_DRIVER_VERSION = require('../../../package.json').version;
+
+/**
+ * FaaS environment metadata.
+ * @public
+ */
+export interface FaasMetadata {
```
nbbeeken, line 10:
```suggestion
export interface FAASMetadata {
```

we went with all caps for OIDC, just think acronyms can stay in caps.

nbbeeken:
```suggestion
export interface FaaSMetadata {
```

or..

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,94 @@
...
+  driverInfo?: {
+    name?: string;
+    version?: string;
+    platform?: string;
+  };
+  appName?: string;
+}
+
+export function makeClientMetadata(options?: ClientMetadataOptions): ClientMetadata {
+  options = options ?? {};
```
nbbeeken, line 54:
```suggestion
export function makeClientMetadata(options: ClientMetadataOptions = {}): ClientMetadata {
```

might as well use default arg here.

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,94 @@
...
+      metadata.platform = `${metadata.platform}|${options.driverInfo.platform}`;
+    }
+  }
+
+  if (options.appName) {
+    // MongoDB requires the appName not exceed a byte length of 128
+    const buffer = Buffer.from(options.appName);
+    metadata.application = {
+      name: buffer.byteLength > 128 ? buffer.slice(0, 128).toString('utf8') : options.appName
+    };
```
nbbeeken, line 90:
```suggestion
    const name = Buffer.byteLength(options.appName, 'utf8') <= 128 ? options.appName : Buffer.from(options.appName, 'utf8').subarray(0, 128).toString('utf8')
    metadata.application = { name };
```

I suggest some extra reformatting here, but main points, let's use `Buffer.byteLength()` instead of converting, and use subarray instead of slice since slice is deprecated.

src/cmap/handshake/handshake_document.ts
```diff
@@ -0,0 +1,18 @@
...
+import type { ClientMetadata } from './client_metadata';
+
+/** @internal */
+export interface HandshakeDocument extends Document {
+  /**
+   * @deprecated Use hello instead
+   */
+  ismaster?: boolean | 0 | 1;
+  hello?: boolean | 0 | 1;
+  helloOk?: boolean;
```
nbbeeken, line 12: We are a hello capable driver so we should always send helloOk, could we make this required?

src/cmap/connect.ts
```diff
@@ -206,42 +197,11 @@ export interface HandshakeDocument extends Document {
...
-    if (!provider) {
-      throw new MongoInvalidArgumentError(`No AuthProvider for ${credentials.mechanism} defined.`);
-    }
-    return provider.prepare(handshakeDoc, authContext);
-  }
-  return handshakeDoc;
+  const generator = new HandshakeGenerator([
+    new DefaultHandshakeDecorator(),
+    new AuthHandshakeDecorator()
+  ]);
```
nbbeeken, line 203: These don't operate on any context, could we have one shared global instance?

test/unit/cmap/handshake/auth_handshake_decorator.test.ts
```diff
@@ -0,0 +1,127 @@
...
+          source: '$external',
+          mechanism: AuthMechanism.MONGODB_X509,
+          mechanismProperties: {}
+        }),
+        connection: mockConnection,
+        options: options
+      };
+      const decorator = new AuthHandshakeDecorator();
+
+      it('uses the machanism to prepare the handshake', async function () {
```
nbbeeken, line 115:
```suggestion
      it('uses the mechanism to prepare the handshake', async function () {
```

test/unit/cmap/handshake/default_handshake_decorator.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    generation: 2,
+    hostAddress: new HostAddress('127.0.0.1:27017'),
+    monitorCommands: false,
+    tls: false,
+    loadBalanced: false,
+    metadata: metadata
+  };
+
+  describe('#generate', function () {
+    context('when the auth context exists', function () {
```
nbbeeken, line 37: We identified we still have a bug, from the server logs:

```json
{"t":{"$date":"2023-03-28T17:15:09.807-04:00"},"s":"I",  "c":"NETWORK",  "id":51800,   "ctx":"conn2","msg":"client metadata","attr":{"remote":"127.0.0.1:65369","client":"conn2","doc":{"driver":{"name":"nodejs","version":"5.1.0"},"os":{"type":"Darwin","name":"darwin","architecture":"arm64","version":"21.6.0"},"platform":"Node.js v18.12.1, LE (unified)|Node.js v18.12.1, LE (unified)"}}}
```

They are tricky to setup but should we strive to add some integration tests where we use the driver from highlevel API only (runCommand, ping:1) and we use spys / FakeSocket to validate handshakes?

nbbeeken (changes requested):

PR #3605, opened 2023-03-20
Update HISTORY.md to add v4.14

Grubba27:
### Description

Updating HISTORY.md to add v4.14

#### What is changing?
Updating history.md to show what changed in MongoDB v4.14 is already in this [tag](https://github.com/mongodb/node-mongodb-native/blob/v4.14.0/HISTORY.md) but for some reason, was not in the main

No.

I was trying to check the diff between versions to see if what was reported in this [issue](https://github.com/meteor/meteor/issues/12567) may be a MongoDB TypeScript issue, but I was not able to find the changelog in the main. Made the PR with the diff, but I think a cherry-pick in the main would solve it as well

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -42,7 +42,7 @@ All notable changes to this project will be documented in this file. See [standa
 
 * **NODE-4998:** use ipv4 in socks5 tests for node 18 ([#3538](https://github.com/mongodb/node-mongodb-native/issues/3538)) ([425dbe0](https://github.com/mongodb/node-mongodb-native/commit/425dbe05d7bd1ebc1f77303332b67181aa5ab135))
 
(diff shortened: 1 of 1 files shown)


W-A-James:
Hi Gabriel! Thanks for your PR. Our 4.x and 5.x branches are maintained independently from one another which is what leads to the changelog on main not including the same changes that occurred in 4.14.

As far as the items that you added to the changelog, the deprecations were never on main as they were removed in the major version bump and the feature that was removed is mentioned in the Breaking Changes for v5.0.0-alpha.0 in NODE-4817. The revert of the MessageStream leak fix happened on main in NODE-5027 which is listed under the v5.0.1 bug fixes and the corresponding amended fix was introduced in NODE-5052 which is listed under the v5.1.0 Bug Fixes.

Grubba27:
Got it! Thanks for sharing that info!

PR #3607, opened 2023-03-22
refactor(NODE-5136): make authentication use async / await

baileympearson:
#### What is changing?

This PR cherry-picks unrelated refactors out of https://jira.mongodb.org/browse/NODE-3998.  This is done to simplify review, as well as limit the surface area that I need to debug while looking for the issues in CI for https://jira.mongodb.org/browse/NODE-3998.

No.


src/cmap/auth/mongodb_oidc.ts
```diff
@@ -88,77 +88,59 @@ export class MongoDBOIDC extends AuthProvider {
...
-      if (error) {
-        return callback(error);
-      }
-      if (!workflow) {
-        return callback(
-          new MongoRuntimeError(
-            `Could not load workflow for device ${credentials.mechanismProperties.PROVIDER_NAME}`
-          )
-        );
+    let workflow;
```
nbbeeken, line 91: We can use maybeCallback here, or just an async iife. Then invoke the callback in a `.then`. Then we don't need to special case the synchronous throw, and when this becomes async the refactor would be just unwrapping the iife

src/cmap/connect.ts
```diff
@@ -119,93 +119,92 @@ function performInitialHandshake(
...
-    }
-
-    const start = new Date().getTime();
-    conn.command(ns('admin.$cmd'), handshakeDoc, handshakeOptions, (err, response) => {
-      if (err) {
-        callback(err);
-        return;
+  prepareHandshakeDocument(authContext).then(
+    handshakeDoc => {
+      const handshakeOptions: Document = Object.assign({}, options);
```
nbbeeken, line 124:
```suggestion
      const handshakeOptions = { ...options };
```

src/cmap/connect.ts
```diff
@@ -119,93 +119,92 @@ function performInitialHandshake(
...
+        // The handshake technically is a monitoring check, so its socket timeout should be connectTimeoutMS
+        handshakeOptions.socketTimeoutMS = options.connectTimeoutMS;
       }
 
-      if (response?.ok === 0) {
-        callback(new MongoServerError(response));
-        return;
-      }
+      const start = new Date().getTime();
+      conn.command(ns('admin.$cmd'), handshakeDoc, handshakeOptions, (err, response) => {
```
nbbeeken, line 131: Should we use commandAsync here?

src/cmap/connect.ts
```diff
@@ -119,93 +119,92 @@ function performInitialHandshake(
...
+        }
 
-      if (!('isWritablePrimary' in response)) {
-        // Provide hello-style response document.
-        response.isWritablePrimary = response[LEGACY_HELLO_COMMAND];
-      }
+        if (response?.ok === 0) {
+          callback(new MongoServerError(response));
+          return;
+        }
```
nbbeeken, line 140: IIUC This is unreachable, command won't return a response that has ok:0 it will already have constructed an error

nbbeeken (changes requested):

src/cmap/connect.ts
baileympearson: doesn't this function take a callback? there's no reason to use the async version unless we're in a method that's explicitly async

src/cmap/connect.ts
```diff
@@ -92,120 +100,89 @@ function checkSupportedServer(hello: Document, options: ConnectionOptions) {
...
-    }
+  const handshakeDoc = await prepareHandshakeDocument(authContext);
 
-    const start = new Date().getTime();
-    conn.command(ns('admin.$cmd'), handshakeDoc, handshakeOptions, (err, response) => {
-      if (err) {
-        callback(err);
-        return;
-      }
+  // @ts-expect-error: TODO(NODE-XXXX): The options need to be filtered properly, Connection options differ from Command options
```
nbbeeken, line 123: file ticket

src/cmap/auth/scram.ts
```diff
@@ -363,14 +337,6 @@ function compareDigest(lhs: Buffer, rhs: Uint8Array) {
   return result === 0;
 }
 
-function resolveError(err?: AnyError, result?: Document) {
```
baileympearson, line 366: why is this logic removed?

src/cmap/connect.ts
```diff
@@ -61,7 +60,16 @@ export function connect(options: ConnectionOptions, callback: Callback<Connectio
     if (options.autoEncrypter) {
```
baileympearson, line 60: Just because we've come this far, do you want to refactor up one more function?

```typescript
const makeConnectionAsync = promisify(makeConnection);

async function connectAsync(..) {


}
export const connect = callbackify(connectAsync);
```

I don't think it makes sense to make `makeConnection` async, because ideally we'll refactor it eventually so it's not performing io at all (purely synchronous).  It also currently handles event listeners on the socket, so just wrapping it in a promisify seems more straightforward.

With this last refactor, the auth module is basically completely refactored.

src/cmap/auth/mongodb_aws.ts
```diff
@@ -35,35 +36,33 @@ interface AWSSaslContinuePayload {
...
-      makeTempCredentials(credentials, (err, tempCredentials) => {
-        if (err || !tempCredentials) return callback(err);
-
-        authContext.credentials = tempCredentials;
-        this.auth(authContext, callback);
-      });
-
-      return;
+      authContext.credentials = await makeTempCredentials(credentials);
+      return this.auth(authContext);
```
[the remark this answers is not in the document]
nbbeeken: This change still leaves the method recursive, was there something else to change here?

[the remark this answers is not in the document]
baileympearson: sorry, I meant to remove the call

src/cmap/auth/gssapi.ts
```diff
@@ -32,70 +26,59 @@ type MechanismProperties = {
...
+    const negotiatedPayload = await negotiate(client, 10, saslStartResponse.payload);
+
+    const saslContinueResponse = await externalCommand(
+      connection,
+      saslContinue(negotiatedPayload, saslStartResponse.conversationId)
+    );
+
+    const finalizePayload = await finalize(client, username, saslContinueResponse.payload);
+
+    await externalCommand(connection, {
```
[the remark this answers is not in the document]
nbbeeken: nope this is correct, the function will throw if we get ok: 0 indicating auth was unsuccessful

src/cmap/auth/gssapi.ts
```diff
@@ -32,70 +26,59 @@ type MechanismProperties = {
...
-        });
-      });
+
+    const { username } = credentials;
+
+    const client = await makeKerberosClient(authContext);
+
+    const payload = await client.step('');
+
+    const saslStartResponse = await externalCommand(connection, saslStart(payload));
```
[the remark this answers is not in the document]
nbbeeken: Yep, callback TS issue again

src/cmap/auth/mongocr.ts
```diff
@@ -1,47 +1,41 @@
...
+
+    const hashPassword = crypto
+      .createHash('md5')
+      .update(`${username}:mongo:${password}`, 'utf8')
+      .digest('hex');
+
+    // Final key
+    const key = crypto
+      .createHash('md5')
+      .update(nonce + username + hashPassword, 'utf8')
```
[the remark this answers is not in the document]
nbbeeken: done

src/cmap/auth/scram.ts
nbbeeken: https://github.com/mongodb/node-mongodb-native/blob/NODE-5136-misc-refactors/src/cmap/connection.ts#L410

The command fn throws if there is a server error, the result condition would never be used

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -77,88 +76,56 @@ export class MongoDBOIDC extends AuthProvider {
...
-        error => {
-          callback(error);
-        }
+    const workflow = getWorkflow(credentials);
+    if (!workflow) {
+      throw new MongoRuntimeError(
+        `Could not load workflow for provider ${credentials.mechanismProperties.PROVIDER_NAME}`
       );
-    });
+    }
```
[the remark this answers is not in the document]
nbbeeken: fixed and added a test

src/cmap/connect.ts
nbbeeken: I looked into making this change but it would only be usable here, the socks5 logic would need to be untangled from callbacks as well. Or we leave socks5 unchanged and it still uses the callback form at which point we haven't got much value out of making a promisified version. Let's stick with what we got so far

PR #3610, opened 2023-03-27
feat(NODE-5169): Implement emergency logger

W-A-James:
#### What is changing?

- Implement `MongoLogger.emergency` severity logger, the private `MongoLogger.log` method and related private helpers
- Change the requirement for `MongoLoggerMongoClientOptions.mongodbLogPath` to be a `MongoDBLogWritable`
- All command monitoring and connection monitoring events now have a public name field.
- `ConnectionClosedEvent`'s constructor now takes in an optional error that is used in the logger
- Adds a new integration test to ensure that the redaction of sensitive commands in the `CommandXXXXEvent` constructors works when the events are passed to the logger.

No

Get initial implementation of one severity logging method complete to enable quick completion of the remainder of the logging methods


src/cmap/connection_pool_events.ts
```diff
@@ -29,11 +29,18 @@ export class ConnectionPoolMonitoringEvent {
 export class ConnectionPoolCreatedEvent extends ConnectionPoolMonitoringEvent {
   /** The options used to create this connection pool */
   options?: ConnectionPoolOptions;
+  /** The size of the ConnectionPool's waitQueue */
+  waitQueueSize: number;
```
W-A-James, line 33: This field is required by the [CMAP logging spec](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#pool-created-message) if the driver supports the option and the user specifies it

src/cmap/connection_pool_events.ts
```diff
@@ -29,11 +29,18 @@ export class ConnectionPoolMonitoringEvent {
...
 
   /** @internal */
   constructor(pool: ConnectionPool) {
     super(pool);
     this.options = pool.options;
+    this.waitQueueSize = pool.waitQueueSize;
+  }
+
+  get name(): string {
+    return 'ConnectionPoolCreated';
```
W-A-James, line 43: A `name` getter method has been added to each Connection Pool Event and Command Monitoring Event to function as a discriminator between all the event types to make the `DEFAULT_LOG_TRANSFORM` easier to implement.

src/mongo_logger.ts
```diff
@@ -143,37 +180,234 @@ function resolveLogPath(
...
+    l.commandName = ev.commandName;
+    l.requestId = ev.requestId;
+    l.driverConnectionId = ev?.connectionId;
+    const { host, port } = getHostPort(ev.address);
+    l.serverHost = host;
+    l.serverPort = port;
+    l.serviceId = ev?.serviceId;
+
+    return l;
+  };
```
W-A-James, line 224: This centralizes the logic for the common fields for the command monitoring events

src/mongo_logger.ts
```diff
@@ -143,37 +180,234 @@ function resolveLogPath(
...
+
+function DEFAULT_LOG_TRANSFORM(logObject: Record<string, any>): Omit<Log, 's' | 't' | 'c'> {
+  let log: Omit<Log, 's' | 't' | 'c'> = {};
+
+  const getHostPort = (s: string): { host: string; port: number } => {
+    const lastColon = s.lastIndexOf(':');
+    const host = s.slice(0, lastColon);
+    const port = Number.parseInt(s.slice(lastColon + 1));
+    return { host, port };
+  };
```
W-A-James, line 209: This likely needs an update to handle IPV6 addresses

src/mongo_logger.ts
```diff
@@ -143,37 +180,234 @@ function resolveLogPath(
...
+  const attachConnectionFields = (l: any, ev: ConnectionPoolMonitoringEvent) => {
+    const { host, port } = getHostPort(ev.address);
+    l.serverHost = host;
+    l.serverPort = port;
+
+    return l;
+  };
+
+  let ev;
+  switch (logObject.name) {
```
W-A-James, line 235: Using this as a discriminant allows for arbitrary objects with a name field to match these branches and is likely not the direction we want to go with this.

test/unit/mongo_logger.test.ts
```diff
@@ -519,6 +526,391 @@ describe('class MongoLogger', function () {
...
+            const commandFailed = {
+              commandName: 'find',
+              requestId: 0,
+              duration: 0,
+              connectionId: 0,
+              address: '127.0.0.1:27017',
+              serviceId: '0x1234567890',
+              databaseName: 'db',
+              name: 'CommandFailed'
+            };
```
W-A-James, line 679: TODO: Replace these with a sinon Mock of each of the corresponding classes

test/unit/mongo_logger.test.ts
```diff
@@ -519,6 +526,391 @@ describe('class MongoLogger', function () {
...
+            const connectionCheckedOut = {
+              name: 'ConnectionCheckedOut',
+              connectionId: 0,
+              options
+            };
+            const connectionCheckedIn = {
+              name: 'ConnectionCheckedIn',
+              connectionId: 0,
+              options
+            };
```
W-A-James, line 815: TODO: Replace with mocks

test/unit/mongo_logger.test.ts
```diff
@@ -519,6 +526,391 @@ describe('class MongoLogger', function () {
...
+
+            logger[severityLevel]('command', message);
+            expect(stream.buffer).to.have.lengthOf(1);
+            expect(stream.buffer[0]).to.have.property('message', message);
+          });
+        });
+
+        context('spec-required logs', function () {
+          let stream: BufferingStream | undefined;
+          let logger: MongoLogger | undefined;
```
W-A-James, line 636: Doesn't need to be a union

src/cmap/command_monitoring_events.ts
```diff
@@ -50,6 +50,10 @@ export class CommandStartedEvent {
   get hasServiceId(): boolean {
     return !!this.serviceId;
   }
+
+  get name(): string {
+    return 'CommandStarted';
+  }
```
W-A-James, line 56: Change these getters to a readonly property

test/unit/mongo_logger.test.ts
```diff
@@ -519,6 +526,455 @@ describe('class MongoLogger', function () {
...
+              logDestination: stream
+            } as any);
+
+            logger[severityLevel]('command', message);
+            expect(stream.buffer).to.have.lengthOf(1);
+            expect(stream.buffer[0]).to.have.property('message', message);
+          });
+        });
+
+        context('spec-required logs', function () {
```
W-A-James, line 634: Can convert a lot of these to table tests

test/unit/mongo_logger.test.ts
```diff
@@ -95,7 +96,7 @@ describe('class MongoLogger', function () {
 
         context('when MONGODB_LOG_ALL is valid', () => {
           for (const [validOption, expectedValue] of validNonDefaultOptions) {
-            context(`{ MONGODB_LOG_ALL: '${validOption} }'`, () => {
+            context(`{ MONGODB_LOG_ALL: '${validOption}' }`, () => {
```
W-A-James, line 99: Drive by cleanup so our test output is more readable

test/unit/mongo_logger.test.ts
```diff
@@ -486,7 +493,11 @@ describe('class MongoLogger', function () {
         context('when mongodbLogPath is set to valid client option', function () {
           for (const validEnvironmentOption of validEnvironmentOptions) {
             for (const validValue of validClientOptions) {
-              it(`{environment: "${validEnvironmentOption}", client: "${validValue}"} uses the value from the client options`, function () {
+              it(`{environment: "${validEnvironmentOption}", client: ${
+                typeof validValue === 'object'
+                  ? 'new ' + validValue.constructor.name + '(...)'
+                  : '"' + validValue.toString() + '"'
+              }} uses the value from the client options`, function () {
```
W-A-James, line 500: Emits a more useful test output message. Would emit `[object Object]` when passed in a writable. now emits `new Writable(...)`

test/integration/command-logging-and-monitoring/command_logging.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  context('sensitive commands', function () {
+    it('should redact sensitive commands', async function () {
+      const log = (ev: CommandStartedEvent | CommandSucceededEvent | CommandFailedEvent) => {
+        client.mongoLogger.emergency('command', ev);
+      };
+      client.on('commandStarted', log);
+      client.on('commandSucceeded', log);
+      client.on('commandFailed', log);
```
W-A-James, line 60: Have to capture logs this way since we don't have the logger actually being called within the driver itself

src/mongo_logger.ts
W-A-James: Disregard as the logger is internal and will not be directly exposed to users.

src/connection_string.ts
```diff
@@ -1292,5 +1292,6 @@ export const DEFAULT_OPTIONS = new CaseInsensitiveMap(
  */
 export const FEATURE_FLAGS = new Set([
   Symbol.for('@@mdb.skipPingOnConnect'),
-  Symbol.for('@@mdb.enableMongoLogger')
+  Symbol.for('@@mdb.enableMongoLogger'),
+  Symbol.for('@@mdb.internalLoggerConfig')
```
W-A-James, line 1296: This may not be the correct place for this, so I'm open to moving the logic to accept the internal config elsewhere in this file/the `parseOptions` function

src/mongo_logger.ts
```diff
@@ -124,56 +176,247 @@ function resolveLogPath(
...
+  s: SeverityLevel;
+  t: Date;
+  c: MongoLoggableComponent;
+  message?: string;
+}
+
+/** @internal */
+export interface MongoDBLogWritable {
+  write(log: Log): unknown;
+}
```
W-A-James, line 217: Is this interface too simplistic? Should we explicitly accept a union of this interface and an actual `Writable` stream?

src/mongo_logger.ts
```diff
@@ -124,56 +176,247 @@ function resolveLogPath(
...
+      if (ev.options) {
+        const { maxIdleTimeMS, minPoolSize, maxPoolSize, maxConnecting, waitQueueTimeoutMS } =
+          ev.options;
+        log = {
+          ...log,
+          maxIdleTimeMS,
+          minPoolSize,
+          maxPoolSize,
+          maxConnecting,
+          waitQueueTimeoutMS
```
W-A-James, line 304: [The spec](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#pool-created-message) says these fields are only required if the user specified a value.

Are we okay with unconditionally having these fields, or do we think the extra information will be confusing if the user didn't configure it manually?

src/mongo_logger.ts
```diff
@@ -124,56 +176,247 @@ function resolveLogPath(
...
+        };
+        log.waitQueueSize = ev.waitQueueSize;
```
W-A-James, line 306: See above comment re this optional field

src/mongo_logger.ts
```diff
@@ -124,56 +176,247 @@ function resolveLogPath(
...
+      log.message = 'Connection checked in';
+      log.driverConnectionId = ev.connectionId;
+      break;
+    default:
+      for (const key in logObject) {
+        const value = logObject[key];
+        // eslint-disable-next-line no-restricted-syntax
+        if (value === undefined || value === null) continue;
+        log[key] = value;
+      }
```
W-A-James, line 379: Is there a better way to do this? The spec requires that for null or undefined fields, we delete the key itself, so optional chaining doesn't satisfy our requirements here.

src/mongo_logger.ts
```diff
@@ -124,56 +176,247 @@ function resolveLogPath(
...
 
   if (typeof MONGODB_LOG_PATH === 'string' && isValidLogDestinationString(MONGODB_LOG_PATH)) {
-    return MONGODB_LOG_PATH.toLowerCase() === 'stderr' ? process.stderr : process.stdout;
+    return (MONGODB_LOG_PATH.toLowerCase() === 'stderr'
+      ? process.stderr
+      : process.stdout) as unknown as MongoDBLogWritable;
   }
 
-  return process.stderr;
+  return process.stderr as unknown as MongoDBLogWritable;
```
W-A-James, line 203: This feels like an inelegant solution, so I'm open to changing this

src/mongo_logger.ts
```diff
@@ -122,58 +176,252 @@ function resolveLogPath(
...
   }
-): Writable {
+): MongoDBLogWritable {
   const isValidLogDestinationString = (destination: string) =>
     ['stdout', 'stderr'].includes(destination.toLowerCase());
   if (typeof mongodbLogPath === 'string' && isValidLogDestinationString(mongodbLogPath)) {
-    return mongodbLogPath.toLowerCase() === 'stderr' ? process.stderr : process.stdout;
+    return (mongodbLogPath.toLowerCase() === 'stderr'
+      ? process.stderr
+      : process.stdout) as unknown as MongoDBLogWritable;
```
W-A-James, line 187: This typecasting feels inelegant so I'm open to any suggestions on how to improve this.

src/mongo_logger.ts
```diff
@@ -122,58 +176,252 @@ function resolveLogPath(
...
+        log.message = 'Connection checked in';
+        log.driverConnectionId = ev.connectionId;
+        break;
+    }
+  } else {
+    for (const key in logObject) {
+      const value = logObject[key];
+      // eslint-disable-next-line no-restricted-syntax
+      if (value === undefined || value === null) continue;
+      log[key] = value;
```
W-A-James, line 383: Is there a better way to check this that isn't:

```typescript
if (value) {
	log[key] = value;
}
```

That solution doesn't work as we don't know if `value` is falsey (e.g. `false` or `0`) while not being nullish

src/cmap/command_monitoring_events.ts
```diff
@@ -65,6 +72,7 @@ export class CommandSucceededEvent {
   commandName: string;
   reply: unknown;
   serviceId?: ObjectId;
+  name: string = COMMAND_SUCCEEDED;
```
nbbeeken, line 75:
```suggestion
  name = COMMAND_SUCCEEDED;
```

If we leave out the string annotation then this gets to be a string literal type, which could be used for type narrowing, same for the rest.

src/mongo_logger.ts
```diff
@@ -122,58 +178,262 @@ function resolveLogPath(
...
+
+    return l;
+  };
+
+  let ev;
+  if (APM_EVENTS.includes(logObject.name)) {
+    log = attachCommandFields(log, logObject as any);
+    switch (logObject.name) {
+      case COMMAND_STARTED:
+        ev = logObject as CommandStartedEvent;
```
nbbeeken, line 272: I'm hoping that if we can improve the type information of `name` we can remove this casting. I think you'll have to make union out of Loggable

src/mongo_logger.ts
```diff
@@ -122,58 +178,262 @@ function resolveLogPath(
...
+        log.message = 'Command started';
```
nbbeeken, line 273: should we just set this to `logObject.name`? One less string to think about, I think the names, despite not having spaces are readable enough

src/cmap/connection_pool_events.ts
```diff
@@ -104,17 +127,21 @@ export class ConnectionClosedEvent extends ConnectionPoolMonitoringEvent {
...
   constructor(
     pool: ConnectionPool,
     connection: Pick<Connection, 'id' | 'serviceId'>,
-    reason: string
+    reason: string,
+    error?: Error
   ) {
     super(pool);
     this.connectionId = connection.id;
     this.reason = reason || 'unknown';
```
nbbeeken, line 142: We have a switch statement based on this value now, so maybe lets improve this a bit, it can never be nullish so we can remove the `|| 'unknown'` case and can we make the type of reason **in the constructor** a union of string literals, we can leave the reason prop on the class as a string for now, but this improvement would help us not make a mistake internally

src/mongo_logger.ts
```diff
@@ -122,58 +178,262 @@ function resolveLogPath(
...
+        log.driverConnectionId = ev.connectionId;
+        break;
+      case CONNECTION_CHECKED_IN:
+        ev = logObject as ConnectionCheckedInEvent;
+        log.message = 'Connection checked in';
+        log.driverConnectionId = ev.connectionId;
+        break;
+    }
+  } else {
+    for (const key in logObject) {
```
nbbeeken, line 381: Hm for...in is not typically what we would use b/c of enumerating keys from the prototype chain, but I suppose that is precisely what we want b/c ConnectionPoolMonitoringEvent uses inheritance. So just making sure that is intentional?

src/mongo_logger.ts
nbbeeken: This will filter undefined and null, but not 0 or empty string:
```ts
if (value != null) log[key] = value;
```

src/mongo_logger.ts
```diff
@@ -122,58 +178,262 @@ function resolveLogPath(
...
+  return log;
 }
 
 /** @internal */
 export class MongoLogger {
   componentSeverities: Record<MongoLoggableComponent, SeverityLevel>;
   maxDocumentLength: number;
-  logDestination: Writable;
+  logDestination: MongoDBLogWritable | Writable;
+  writeStringifiedLogs: boolean;
```
nbbeeken, line 396: from slack, we can remove if we wrap stdX

src/mongo_logger.ts
```diff
@@ -122,58 +178,262 @@ function resolveLogPath(
...
-  debug(component: any, message: any): void {}
+  /** @experimental */
+  emergency(component: MongoLoggableComponent, message: Loggable | string): void {
+    this.log(component, 'emergency', message);
+  }
 
-  trace(component: any, message: any): void {}
+  private log(
+    component: MongoLoggableComponent,
+    severity: SeverityLevel,
```
nbbeeken, line 413: Super minor, but I would've put severity first, component second, the thinking is that the sev is not _really_ a variable, all of log's invocations are going to have the first argument be a hardcoded string literal. 

Maybe it's worth keeping all our logger wrappers super slim? at least until there's a need for more than proxying the sev.
```ts
emergency = this.log.bind(this, 'emergency')
```

src/mongo_logger.ts
```diff
@@ -122,58 +178,262 @@ function resolveLogPath(
...
+    message: Loggable | string
+  ): void {
+    if (compareSeverity(severity, this.componentSeverities[component]) <= 0) {
```
nbbeeken, line 416: nit: Can we flip this and early return?

src/connection_string.ts
nbbeeken: LGTM, we need a way to get our settings in there for testing, as long as we capture this information. Can we add some basic unit tests to: `test/integration/node-specific/feature_flags.test.ts`. I'd like something that makes sure we don't drop things from here unintentionally thinking it's "unused".

src/cmap/connection_pool_events.ts
nbbeeken: We have a `waitQueueSize` property which represents how large the wait queue currently is, but I think that "only required to include if the driver supports this option and the user specified a value." clarifies that it's only if it is a configurable option, which it is not IIUC.

src/cmap/connection_pool_events.ts
```diff
@@ -104,17 +127,21 @@ export class ConnectionClosedEvent extends ConnectionPoolMonitoringEvent {
...
+  name: string = CONNECTION_CLOSED;
+  error: Error | undefined;
 
   /** @internal */
   constructor(
     pool: ConnectionPool,
     connection: Pick<Connection, 'id' | 'serviceId'>,
-    reason: string
+    reason: string,
+    error?: Error
```
nbbeeken, line 138: will this ever be something that is not a `MongoError` subclass?

nbbeeken (changes requested):

src/mongo_logger.ts
nbbeeken: We _could_ use the `DEFAULT_OPTIONS` map to filter some of these... I'm not sure the value is worth it, seeing defaults can be helpful too, inspect what you expect, maybe they would change across majors.

src/mongo_logger.ts
nbbeeken: `waitQueueSize` is different here, this will always be zero, it's not a configured max size but the current size, it will always be 0 here iiuc

src/mongo_logger.ts
W-A-James: The spec requires that we have that exact string for the message field, but if we get some more eyes on this I would be open to using the existing constant.

We could also attach the message to the event itself.

src/mongo_logger.ts
W-A-James: The intention here was just to pass through an arbitrary object that isn't a command event or a CMAP event excepting the top-level nullish keys. Would it be more appropriate to use `Object.entries` here?

src/cmap/connection_pool_events.ts
W-A-James: Good catch. There shouldn't be a case where this is true.

src/mongo_logger.ts
nbbeeken: Yep that works! Oh yes I see now that the CMAP event is handled earlier so inheritance isn't an issue here.

src/mongo_logger.ts
nbbeeken: If its the spec let's stick with it, I think it's fine to attach here, we don't need to add it to the events, particularly b/c the events now do have the name property.

src/cmap/connection_pool_events.ts
```diff
@@ -102,19 +137,23 @@ export class ConnectionClosedEvent extends ConnectionPoolMonitoringEvent {
   /** The id of the connection */
   connectionId: number | '<monitor>';
   /** The reason the connection was closed */
-  reason: string;
+  reason: 'idle' | 'stale' | 'poolClosed' | 'error';
   serviceId?: ObjectId;
+  name = CONNECTION_CLOSED;
+  error: Error | undefined;
```
nbbeeken, line 143:
```suggestion
  error: Error | null;
```

If we can using null is preferred to indicate the absence of something.
Also should this be MongoError? Following the logic it seems like the only non MongoError that could be set here is socket.authorizationError in connect.ts, maybe that's something we can add TODOs for but generally this type is more useful if it's a MongoError, b/c it will have a `.code` property

src/cmap/connection_pool_events.ts
```diff
@@ -102,19 +137,23 @@ export class ConnectionClosedEvent extends ConnectionPoolMonitoringEvent {
...
-    reason: string
+    reason: 'idle' | 'stale' | 'poolClosed' | 'error',
+    error?: AnyError
   ) {
     super(pool);
     this.connectionId = connection.id;
-    this.reason = reason || 'unknown';
+    this.reason = reason;
     this.serviceId = connection.serviceId;
+    this.error = error;
```
nbbeeken, line 156:
```suggestion
    this.error = error ?? null;
```

src/mongo_logger.ts
```diff
@@ -119,61 +189,273 @@ function parseSeverityFromString(s?: string): SeverityLevel | null {
...
+          log.reason = 'An error occurred while using the connection';
+          if (logObject.error) {
+            log.error = logObject.error;
+          }
+          break;
+        case 'poolClosed':
+          log.reason = 'Connection pool was closed';
+          break;
+        default:
+        // Omit if we have some other reason as it would be invalid
```
nbbeeken, line 395: MongoRuntimeError? or log.reason == `Unknown close reason: ${logObject.reason}`?  Kinda prefer the latter

src/mongo_logger.ts
```diff
@@ -119,61 +189,273 @@ function parseSeverityFromString(s?: string): SeverityLevel | null {
...
 
-  if (typeof MONGODB_LOG_PATH === 'string' && isValidLogDestinationString(MONGODB_LOG_PATH)) {
-    return MONGODB_LOG_PATH.toLowerCase() === 'stderr' ? process.stderr : process.stdout;
+  const host = hostAddress.host ?? '';
+  const port = hostAddress.port ?? 0;
+  return { host, port };
+}
+
+function attachCommandFields(
+  l: any,
```
nbbeeken, line 279:
```suggestion
  l: Record<string, any>,
```

It is at least an object, but I don't think much more typing beyond this is needed

src/mongo_logger.ts
```diff
@@ -119,61 +189,273 @@ function parseSeverityFromString(s?: string): SeverityLevel | null {
...
+/** @internal */
+export type Loggable = LoggableEvent | LogConvertible;
+
+function isLogConvertible(obj: Loggable): obj is LogConvertible {
+  const objAsLogConvertible = obj as LogConvertible;
+  // eslint-disable-next-line no-restricted-syntax
+  return objAsLogConvertible.toLog !== undefined && typeof objAsLogConvertible.toLog === 'function';
+}
+
+function getHostPort(address: string): { host: string; port: number } {
```
nbbeeken, line 265:
```ts
  /** @internal */
  toHostPort(): { host: string; port: number } {
    return { host: this.socketPath ?? this.host ?? '', port: this.port ?? 0 };
  }

  // Usage:
  const { host, port } = HostAddress.fromString(address).toHostPort();
```

Just in an effort to consolidate utilities, we can make this a method on HostAddress, it is supposed to be all our abstractions of address string parsing/handling

src/mongo_logger.ts
```diff
@@ -119,61 +189,273 @@ function parseSeverityFromString(s?: string): SeverityLevel | null {
...
+
+function attachConnectionFields(l: any, ev: ConnectionPoolMonitoringEvent) {
+  const { host, port } = getHostPort(ev.address);
+  l.serverHost = host;
+  l.serverPort = port;
+
+  return l;
+}
+
+function DEFAULT_LOG_TRANSFORM(logObject: LoggableEvent): Omit<Log, 's' | 't' | 'c'> {
```
nbbeeken, line 303:
```suggestion
function defaultLogTransform(logObject: LoggableEvent): Omit<Log, 's' | 't' | 'c'> {
```

nbbeeken (changes requested):

src/mongo_logger.ts
```diff
@@ -119,61 +189,273 @@ function parseSeverityFromString(s?: string): SeverityLevel | null {
...
   constructor(options: MongoLoggerOptions) {
     this.componentSeverities = options.componentSeverities;
     this.maxDocumentLength = options.maxDocumentLength;
     this.logDestination = options.logDestination;
   }
 
-  /* eslint-disable @typescript-eslint/no-unused-vars */
-  /* eslint-disable @typescript-eslint/no-empty-function */
-  emergency(component: any, message: any): void {}
+  /** @experimental */
```
baileympearson, line 437: Why is this type marked experimental?  The class itself is internal

baileympearson (changes requested):

src/cmap/connection_pool_events.ts
nbbeeken: https://github.com/mongodb/node-mongodb-native/pull/3610#discussion_r1163110240

src/mongo_logger.ts
```diff
@@ -111,6 +165,22 @@ function parseSeverityFromString(s?: string): SeverityLevel | null {
   return null;
 }
 
+type ClientLogPathOptions = {
```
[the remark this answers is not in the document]
nbbeeken: I asked this be broken out since the function has a concern about env vs client options, so the type helps communicate that but if we inlined the type and added tsdoc or named the parameters appropriately that's okay too

src/cmap/connection_pool_events.ts
W-A-James: The call to `connect` in connection_pool.ts where `ConnectionClosedEvent` is instantiated accepts a callback of the shape `(error?: AnyError, result?:Connection) => void`, so I think it would be appropriate to have this be `AnyError`

src/mongo_logger.ts
```diff
@@ -119,61 +189,273 @@ function parseSeverityFromString(s?: string): SeverityLevel | null {
...
+export interface Log extends Record<string, any> {
+  t: Date;
+  c: MongoLoggableComponent;
+  s: SeverityLevel;
+  message?: string;
+}
+
+/** @internal */
+export interface MongoDBLogWritable {
+  write(log: Log): unknown;
```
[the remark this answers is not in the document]
W-A-James: Good catch. When deciding on what the `MongoDBLogWritable` interface should look like, I wasn't sure if we wanted the write method to have a return value that would indicate success or failure that we'd have to handle inside the private log method, but I think we're good to change this to return `void`.

src/cmap/connection_pool_events.ts
nbbeeken: `AnyError` is a vestige of the `Callback` type which wasn't fully accurate. So in reality  `connect` will always return a MongoError subclass, besides one case: `authorizationError` -- I'm wondering if we can keep the TypeScript on this class accurate to what we _desire_ (MongoError) before making this public, and we can follow up with wrapping that one outlier later. 

I believe, wrapping it in a MongoError would be acceptable in semver per our policy on errors descending from the same parent + I'm not sure if `authorizationError` would be surfaced to user's via any other mechanism or if it would be converted to a server selection timeout error.

src/index.ts
```diff
@@ -314,7 +314,7 @@ export type {
   MongoLoggerOptions,
   SeverityLevel
 } from './mongo_logger';
-export { createStdLogger } from './mongo_logger';
+export { createStdioLogger } from './mongo_logger';
```
nbbeeken, line 317: This doesn't need to be exported, is this for testing? the test/mongodb.ts file imports all the things from all the files under the `src`

nbbeeken (changes requested):

src/mongo_logger.ts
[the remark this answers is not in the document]
nbbeeken: This type is internal for now but the intention is that it would be put on `MongoClientOptions` right?

I think `unknown` is correct here, `void` will prevent APIs that have a return value which is not a requirement we need to impose.

src/index.ts
W-A-James: npm install was complaining about this and since it's marked internal I assumed this was alright to do, but I can go ahead and fix this

src/utils.ts
```diff
@@ -1083,6 +1083,16 @@ export class HostAddress {
   static fromSrvRecord({ name, port }: SrvRecord): HostAddress {
     return HostAddress.fromHostPort(name, port);
   }
+
+  toHostPort(): { host: string; port: number } {
```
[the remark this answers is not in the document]
W-A-James: I think that would be useful to adopt that approach. Would `family` be set to `undefined` or `null` in the event that `socketPath` is defined?

src/mongo_logger.ts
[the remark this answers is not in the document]
nbbeeken: Yea because it would not matter to us I don't think we should have the enforcement of a return value

src/utils.ts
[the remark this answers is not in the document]
nbbeeken: I think family can't reliably be determined here, if the hostname is `localhost` that can either map to `ipv6` or `ipv4`, so if isIPv6 is false on the class that does not imply that the family is ipv4. And there's the question of what we would set it to when it is a socket path

src/cmap/connection_pool_events.ts
nbbeeken: From IRL, let's make the type MongoError, cast away issues, file a follow up to wrap `authorizationError`, add todos.

test/integration/command-logging-and-monitoring/command_logging.test.ts
W-A-James, line 1: Going to remove this as it just replicates the spec tests

nbbeeken (approved):

src/cmap/connection_pool_events.ts
```diff
@@ -8,11 +21,23 @@ import type { ConnectionPool, ConnectionPoolOptions } from './connection_pool';
  * @public
  * @category Event
  */
-export class ConnectionPoolMonitoringEvent {
+export abstract class ConnectionPoolMonitoringEvent {
```
baileympearson, line 24: I don't think we can make this class abstract, since it's public.

src/mongo_logger.ts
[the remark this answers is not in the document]
baileympearson: I'm in favor of inlining it and documenting it

src/utils.ts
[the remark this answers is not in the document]
nbbeeken: That default would not be true if the host was a socket path, and sIPv6=false does not imply an IPv4 address

src/cmap/connection_pool_events.ts
W-A-James: From what I understand, this class has a private constructor, so users shouldn't be constructing these in their own code since constructing this event (and any of the other subclasses for that matter) isn't a part of our public API. As such, making this abstract would only break things for users that are _already_ misusing our API.

src/constants.ts
```diff
@@ -23,16 +23,27 @@ export const SERVER_DESCRIPTION_CHANGED = 'serverDescriptionChanged' as const;
 export const TOPOLOGY_OPENING = 'topologyOpening' as const;
 export const TOPOLOGY_CLOSED = 'topologyClosed' as const;
 export const TOPOLOGY_DESCRIPTION_CHANGED = 'topologyDescriptionChanged' as const;
+/** @public */
```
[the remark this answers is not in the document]
W-A-James: This was done since the the `name` property that we have on our Command Events and ConnectionPool Events is public so we need to have these constants exported. I'm open to doing this in a more elegant way.

src/utils.ts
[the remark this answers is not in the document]
W-A-James: Just to clarify, what do we want our path forward here to be? As I see it we have a few options:

1. Change the function to a getter of name `address` and set `family` to `null` when `socketPath` is defined
2. Change the function to a getter of name `address` and set `family` to `socket` when `socketPath` is defined
3. Keep the function as is

src/constants.ts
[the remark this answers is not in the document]
baileympearson: Well one option is to make the `name` property on the command events internal.  There's no reason it needs to be public, right?  It's only used by the logger.

src/cmap/command_monitoring_events.ts
```diff
@@ -18,6 +24,7 @@ export class CommandStartedEvent {
   address: string;
   connectionId?: string | number;
   serviceId?: ObjectId;
+  name = COMMAND_STARTED;
```
nbbeeken, line 27: Make all `name` internal, and update constants to go back to not being public

src/cmap/connection_pool_events.ts
```diff
@@ -104,17 +136,21 @@ export class ConnectionClosedEvent extends ConnectionPoolMonitoringEvent {
   /** The reason the connection was closed */
   reason: string;
   serviceId?: ObjectId;
+  name = CONNECTION_CLOSED;
+  error: MongoError | null;
```
nbbeeken, line 140: Make error internal

src/cmap/connection_pool_events.ts
```diff
@@ -138,6 +176,7 @@ export class ConnectionCheckOutStartedEvent extends ConnectionPoolMonitoringEven
 export class ConnectionCheckOutFailedEvent extends ConnectionPoolMonitoringEvent {
   /** The reason the attempt to check out failed */
   reason: AnyError | string;
+  name = CONNECTION_CHECK_OUT_FAILED;
 
   /** @internal */
   constructor(pool: ConnectionPool, reason: AnyError | string) {
```
nbbeeken, line 182: Add error

nbbeeken (changes requested):

src/utils.ts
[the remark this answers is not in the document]
W-A-James: I think it makes sense to leave the name as is and leave it as a regular method for parity with the `fromHostPort` static function.

src/mongo_logger.ts
```diff
@@ -111,69 +165,293 @@ function parseSeverityFromString(s?: string): SeverityLevel | null {
...
+  connectionPoolEvent: ConnectionPoolMonitoringEvent
+) {
+  const { host, port } = HostAddress.fromString(connectionPoolEvent.address).toHostPort();
+  log.serverHost = host;
+  log.serverPort = port;
+
+  return log;
+}
+
+function defaultLogTransform(logObject: LoggableEvent): Omit<Log, 's' | 't' | 'c'> {
```
dariakp, line 291: Should this take more than just `LoggableEvent`? we seem to apply the transform to any object, so it may or may not be an actual event for our internal logging purposes.

dariakp (changes requested):

src/cmap/connection_pool.ts
```diff
@@ -812,7 +821,11 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
         const error = this.closed ? new PoolClosedError(this) : new PoolClearedError(this);
         this.emit(
           ConnectionPool.CONNECTION_CHECK_OUT_FAILED,
-          new ConnectionCheckOutFailedEvent(this, reason)
+          new ConnectionCheckOutFailedEvent(
+            this,
+            reason,
+            reason === 'connectionError' ? error : undefined
```
nbbeeken, line 827: rm condition

src/cmap/connection_pool_events.ts
```diff
@@ -137,12 +184,21 @@ export class ConnectionCheckOutStartedEvent extends ConnectionPoolMonitoringEven
  */
 export class ConnectionCheckOutFailedEvent extends ConnectionPoolMonitoringEvent {
   /** The reason the attempt to check out failed */
-  reason: AnyError | string;
+  reason: string;
+  /** @internal */
+  error?: AnyError;
```
nbbeeken, line 189: MongoError?

nbbeeken (changes requested):

src/mongo_logger.ts
W-A-James: This is a good catch. We had set the type to `LoggableEvent` to allow for type narrowing and get editor completions but I forgot to expand the type to include arbitrary records.

baileympearson (approved):

PR #3615, opened 2023-03-29
fix(NODE-3998): metadata duplication in handshake

baileympearson:
#### What is changing?

This PR prevents metadata duplication in the handshake. To do this, it removes custom transforms for the driverInfo and appName options and constructs the metadata once at the end of options parsing.

No.


test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -0,0 +1,83 @@
...
+            },
+            platform: `Node.js ${process.version}, ${os.endianness()} (unified)`
+          });
+        });
+      });
+
+      context('when app name is provided', function () {
+        context('when the app name is over 128 bytes', function () {
+          const longString = new Array(300).join('a');
+          const exactString = new Array(128 + 1).join('a');
```
nbbeeken, line 59: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat

Maybe use repeat here instead?

src/utils.ts
```diff
@@ -545,9 +545,7 @@ export interface ClientMetadataOptions {
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const NODE_DRIVER_VERSION = require('../package.json').version;
 
-export function makeClientMetadata(options?: ClientMetadataOptions): ClientMetadata {
-  options = options ?? {};
-
+export function makeClientMetadata(options: MongoOptions): ClientMetadata {
```
nbbeeken, line 548:
```ts
    const name = Buffer.byteLength(options.appName, 'utf8') <= 128 ? options.appName : Buffer.from(options.appName, 'utf8').subarray(0, 128).toString('utf8')
    metadata.application = { name };
```
From the other PR, can we make the improvement:

I suggest some extra reformatting here, but main points, let's use `Buffer.byteLength()` instead of converting everytime, and use subarray instead of slice since slice is deprecated.

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -0,0 +1,83 @@
...
+          const options = {
+            appName: longString
+          };
+          const metadata = makeClientMetadata(options);
+
+          it('truncates the application name to 128 bytes', function () {
+            expect(metadata.application?.name).to.equal(exactString);
+          });
```
nbbeeken, line 67: Can we add a test that uses the name `'€'.repeat(45)`. It adds coverage for the case where the string will no longer be wellformed because of the byte wise slicing.

nbbeeken (changes requested):

src/utils.ts
baileympearson: I'll make the refactor but to clarify - any "inefficiency" here isn't important.  This function is only invoked once per mongo client now

test/unit/cmap/handshake/client_metadata.test.ts
baileympearson: What coverage this this add?  It doesn't throw, and still produces a utf-8 string

src/utils.ts
nbbeeken: I still think it has value even though it is done once, thanks!

test/unit/cmap/handshake/client_metadata.test.ts
baileympearson: I added an assertion to confirm that the application string is `128` bytes

src/utils.ts
```diff
@@ -526,13 +529,14 @@ export interface ClientMetadata {
     version: string;
   };
   platform: string;
```
baileympearson, line 531: This was wrong, according to the spec: https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst#hello-command

this modifies a public interface but it's also technically a bug fix, so I think it's okay to remove

src/utils.ts
```diff
@@ -513,7 +513,10 @@ export function makeStateMachine(stateTable: StateTable): StateTransitionFunctio
   };
 }
 
-/** @public */
+/**
+ * @public
+ * https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst#hello-command
```
nbbeeken, line 518:
```suggestion
 * @see https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst#hello-command
```

src/utils.ts
```diff
@@ -526,13 +529,14 @@ export interface ClientMetadata {
...
-  version?: string;
   application?: {
     name: string;
   };
 }
 
-/** @public */
+/**
+ * @public
+ */
```
nbbeeken, line 539: nit, whitespace

src/utils.ts
```diff
@@ -545,44 +549,38 @@ export interface ClientMetadataOptions {
...
-  options = options ?? {};
+export function makeClientMetadata(
+  options: Pick<MongoOptions, 'appName' | 'driverInfo'>
+): ClientMetadata {
+  const name = options.driverInfo.name ? `nodejs|${options.driverInfo.name}` : 'nodejs';
+  const version = options.driverInfo.version
+    ? `${NODE_DRIVER_VERSION}|${options.driverInfo.version}`
+    : NODE_DRIVER_VERSION;
+  const platform = options.driverInfo.platform
+    ? `Node.js ${process.version}, ${os.endianness()} (unified)|${options.driverInfo.platform}`
```
nbbeeken, line 560: What does `(unified)` mean?

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -0,0 +1,99 @@
...
+import { makeClientMetadata } from '../../../mongodb';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const NODE_DRIVER_VERSION = require('../../../../package.json').version;
+
+describe('makeClientMetadata()', () => {
+  context('when driverInfo.platform is provided', () => {
+    it('appends driverInfo.platform to the platform field', () => {
+      const options = {
+        driverInfo: { platform: 'myPlatform' }
```
nbbeeken, line 13: Do we also want tests that pass combinations of the options? I was thinking we'd test both so we can ensure there's not mixing of dependency on the properties.

test/unit/cmap/handshake/client_metadata.test.ts
nbbeeken: It's not quite enough coverage for what I'm thinking, I want to make sure that we pass a string in that when sliced by UTF8 bytes results in a different value than it would be when sliced with the string slice API. Only asserting the output is 128 bytes doesn't confirm that way that slicing is being done is correct since `options.appName.slice(0, 128)` will still make this test pass.

nbbeeken (changes requested):

src/utils.ts
nbbeeken: https://github.com/mongodb/node-mongodb-native/commit/ade9a62301abfdb093fbbc2ff37c0a2e6776f977

We used to report whether the unified or legacy topology was being used, I think we can remove this now.

test/unit/cmap/handshake/client_metadata.test.ts
baileympearson: the options don't impact each other at all (ex: version -> append version, but doesn't impact name or platform). testing combinations of unrelated options doesn't add value

nbbeeken: They don't relate in the current implementation, would we fail a test if they became related in the future? I don't feel strongly if you think it isn't necessary.

src/utils.ts
baileympearson: alex is on board for this as well

baileympearson: addressed

src/utils.ts
baileympearson: fixed

src/utils.ts
baileympearson: fixed!

nbbeeken (dismissed):

nbbeeken (dismissed):

baileympearson:
evergreen retry

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -0,0 +1,150 @@
...
+          name: process.platform,
+          architecture: process.arch,
+          version: os.release()
+        },
+        platform: `Node.js ${process.version}, ${os.endianness()}`
+      });
+    });
+  });
+
+  context('when no custom driverInto is provided', () => {
```
W-A-James, line 76:
```suggestion
  context('when no custom driverInfo is provided', () => {
```

W-A-James (changes requested):

W-A-James (approved):
LGTM

PR #3621, opened 2023-04-04
feat(NODE-5189): deprecate tcp keepalive options

durran:
### Description

Updates TCP keepalive options to the spec where appropriate.

#### What is changing?
- Deprecates the `keepAlive` and `keepAliveInitialDelay` options.
- Defaults `keepAliveInitialDelay` to 300 seconds.

Node does not allow fine tuning of the keepalive options as per the spec, so we set the initial delay and then accept the other settings that Node sets internally. This results in:

-  `SO_KEEPALIVE=1`
- `TCP_KEEPIDLE=300000`
- `TCP_KEEPCNT=10`
- `TCP_KEEPINTVL=1`

https://github.com/mongodb/docs-node/pull/639

NODE-3607


dariakp (changes requested):
@durran No concerns with the deprecations (as long as there is also a clear ticket to remove the options), but changing the defaults is arguably breaking and I think that's why we framed NODE-3607 as a spike; can we take this back and discuss it as a team? I didn't see a kickoff comment or a link to an external doc summarizing the results of the investigation, but some notes addressing the 6 spec callouts point by point would help figure out what changes can happen now and which ones need a follow up ticket.

src/cmap/connect.ts
```diff
@@ -329,14 +329,13 @@ const SOCKET_ERROR_EVENTS = new Set(SOCKET_ERROR_EVENT_LIST);
...
   const connectTimeoutMS = options.connectTimeoutMS ?? 30000;
   const rejectUnauthorized = options.rejectUnauthorized ?? true;
-  const keepAliveInitialDelay =
-    ((options.keepAliveInitialDelay ?? 120000) > socketTimeoutMS
-      ? Math.round(socketTimeoutMS / 2)
-      : options.keepAliveInitialDelay) ?? 120000;
+  // Default to delay to 300 seconds. Node automatically then sets TCP_KEEPINTVL to 1 second
+  // which is acceptable to the recommendation of 10 seconds and also cannot be configured.
+  // TCP_KEEPCNT is also set to 10 in Node and cannot be configured. (Recommendation is 9)
+  const keepAliveInitialDelay = options.keepAliveInitialDelay || 300000;
```
nbbeeken, line 338:
```suggestion
  const keepAliveInitialDelay = options.keepAliveInitialDelay;
```

Because of the defaulting, we should always have a defined number here.

nbbeeken (changes requested):

src/cmap/connect.ts
durran: Updated.

src/connection_string.ts
```diff
@@ -864,12 +864,14 @@ export const OPTIONS = {
       return wc;
     }
   },
+  /** @deprecated - Will not be able to turn off in the future. */
```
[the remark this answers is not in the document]
durran: Added option with message.

src/cmap/connect.ts
```diff
@@ -329,14 +329,13 @@ const SOCKET_ERROR_EVENTS = new Set(SOCKET_ERROR_EVENT_LIST);
...
   const connectTimeoutMS = options.connectTimeoutMS ?? 30000;
   const rejectUnauthorized = options.rejectUnauthorized ?? true;
-  const keepAliveInitialDelay =
-    ((options.keepAliveInitialDelay ?? 120000) > socketTimeoutMS
-      ? Math.round(socketTimeoutMS / 2)
-      : options.keepAliveInitialDelay) ?? 120000;
+  // Default to delay to 300 seconds. Node automatically then sets TCP_KEEPINTVL to 1 second
+  // which is acceptable to the recommendation of 10 seconds and also cannot be configured.
+  // TCP_KEEPCNT is also set to 10 in Node and cannot be configured. (Recommendation is 9)
+  const keepAliveInitialDelay = options.keepAliveInitialDelay;
```
dariakp, line 338: We only want to do the deprecations; changing the defaults would be a potentially breaking change (at any rate it would change behavior), so we can do that in NODE-5190.

dariakp (changes requested):

src/cmap/connect.ts
durran: Removed the changes, only deprecations now.

nbbeeken (approved):

baileympearson (approved):

PR #3626, opened 2023-04-04
feat(NODE-4696): add FaaS env information to client metadata

nbbeeken:
#### What is changing?

- Adds getFaasEnv() that gathers environment information as specified
  - https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst#client-env
- Modifies makeClientMetadata to build the metadata document with an es map that checks the BSON size of every input preventing us from exceeding 512 bytes
  - Map offers a few conveniences over objects: key iteration, element count, set/delete is typical method syntax rather than bracket notation 
- Introduces an error if driverInfo properties name/version/platform use up the 512 bytes, this would prevent us from having a `helloDoc.client.driver.name` property which is required server side
- if metadata document is at capacity, we will attempt to remove fields from `os` until it can fit, followed by the same logic for `env`
  - We do predict this is unlikely, but the hard restriction is that we never construct a document larger than 512 bytes. 

No mongodb manual changes.

I changed the API docs above MongoOptions, this is made public by the client.options getter and represents the parsed options, the type isn't necessarily an accurate representation of the runtime value because there are properties that are set after SRV lookup. 

FaaS environments have a unique impact on the driver's approach to it's distributed systems logic (SDAM). If we can better understand the environments & resources being used we can provide the best FaaS experience possible.


nbbeeken:
https://spruce.mongodb.com/version/642d9be73e8e86b01cc1498f/tasks

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,227 @@
...
+    VERCEL_URL = '',
+    VERCEL_REGION = ''
+  } = process.env;
+
+  const isAWSFaaS = AWS_EXECUTION_ENV.length > 0 || AWS_LAMBDA_RUNTIME_API.length > 0;
+  const isAzureFaaS = FUNCTIONS_WORKER_RUNTIME.length > 0;
+  const isGCPFaaS = K_SERVICE.length > 0 || FUNCTION_NAME.length > 0;
+  const isVercelFaaS = VERCEL.length > 0;
+
+  // Note: order matters, name must always be the last key
```
nbbeeken, line 178: There's a test for this

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    it('client fails to connect with an error relating to size', async function () {
+      client = this.configuration.newClient({ serverSelectionTimeoutMS: 2000 });
+      const error = await client.connect().catch(error => error);
+      if (this.configuration.isLoadBalanced) {
+        expect(error).to.be.instanceOf(MongoServerError);
+      } else {
+        expect(error).to.be.instanceOf(MongoServerSelectionError);
+      }
+      expect(error).to.match(/client metadata document must be less/);
```
nbbeeken, line 42: client.driver.name is the minimum required field. But just to be sure the error relates to size and not missing fields, I check the string

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -0,0 +1,45 @@
...
+      });
+    });
+
+    after(() => sinon.restore());
+
+    it('client fails to connect with an error relating to size', async function () {
+      client = this.configuration.newClient({ serverSelectionTimeoutMS: 2000 });
+      const error = await client.connect().catch(error => error);
+      if (this.configuration.isLoadBalanced) {
+        expect(error).to.be.instanceOf(MongoServerError);
```
nbbeeken, line 38: Interesting that LB has this behavior, but I guess it makes sense

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,511 @@
...
+
+describe('client metadata module', () => {
+  describe('class LimitedSizeDocument', () => {
+    // For the sake of testing the size limiter features
+    // We test document: { _id: ObjectId() }
+    // 4 bytes + 1 type byte + 4 bytes for key + 12 bytes Oid + 1 null term byte
+    // = 22 bytes
+
+    it('sets key and value that fit within maxSize', () => {
+      const doc = new LimitedSizeDocument(22);
```
nbbeeken, line 24: `LimitedSizeDocument` is a private API meant only to serve the needs of clientMetadata. In an effort to help with keeping this utility focused on this use case I kept all fields private, so the driver can only interact via ifFitsSets and toObject

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,511 @@
...
-          version: NODE_DRIVER_VERSION
-        },
-        os: {
-          type: os.type(),
-          name: process.platform,
-          architecture: process.arch,
-          version: os.release()
-        },
-        platform: `Node.js ${process.version}, ${os.endianness()}`
+  describe('getFAASEnv()', function () {
```
nbbeeken, line 37: These tests are basically unchanged, I just updated the name fields

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,511 @@
...
+            try {
+              setupEnv();
+              const { env } = makeClientMetadata({ driverInfo: {} });
+              expect(env).to.deep.equal(outcome);
+            } finally {
+              cleanupEnv();
+            }
+          });
+
+          it('always places name as the last key', () => {
```
nbbeeken, line 391: here's that test for name being last, so if any new faas envs are added we'll keep this up

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,511 @@
...
+            }
+          });
+        }
+      });
+    }
+
+    context('when a numeric FAAS env variable is not numerically parsable', () => {
+      before(() => {
+        process.env.AWS_EXECUTION_ENV = 'non-empty-string';
+        process.env.AWS_LAMBDA_FUNCTION_MEMORY_SIZE = '123not numeric';
```
nbbeeken, line 407: The leading digits are an "interesting" part of this test, if we used parseInt we would get 123, but we use `Number`/`unary +` so we can test for NaN to see if the whole string was valid.

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,511 @@
...
+
+        it('only includes env.name', () => {
+          const metadata = makeClientMetadata({ driverInfo: {} });
+          expect(metadata).to.have.property('env');
+          expect(metadata).to.have.nested.property('env.region', 'abc');
+          expect(metadata.os).to.have.all.keys('type');
+        });
+      });
+
+      context('os.type too large', () => {
```
nbbeeken, line 481: The scenario is likely never ever possible (I have to fake an OS with a 512 length name), but it demonstrates that we abide the strictness of the 512 bytes limit.

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,227 @@
...
+export class LimitedSizeDocument {
+  private document = new Map();
+  constructor(private maxSize: number) {}
+
+  private getBsonByteLength() {
+    return BSON.serialize(this.document).byteLength;
+  }
+
+  /** Only adds key/value if the bsonByteLength is less than MAX_SIZE */
+  public ifFitsSets(key: string, value: Record<string, any> | string): boolean {
```
durran, line 60: Maybe a little perf nit here. Instead of serialising the entire document with each attempted property addition, can we just instead serialise the field being added? I'm proposing we save the current serialised document size as a property, then all we need to do on attempted addition is serialise the new property to get the proposed size being added to the existing doc. Then it's just a check of `propertySize + this.documentSize > this.maxSize`. But since we're never dealing with over 512 bytes maybe it doesn't matter, so feel free to ignore. Just thinking about cutting a little more time off.

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,227 @@
...
+
+  const isAWSFaaS = AWS_EXECUTION_ENV.length > 0 || AWS_LAMBDA_RUNTIME_API.length > 0;
+  const isAzureFaaS = FUNCTIONS_WORKER_RUNTIME.length > 0;
+  const isGCPFaaS = K_SERVICE.length > 0 || FUNCTION_NAME.length > 0;
+  const isVercelFaaS = VERCEL.length > 0;
+
+  // Note: order matters, name must always be the last key
+  const faasEnv = new Map();
+
+  if (isAWSFaaS && !(isAzureFaaS || isGCPFaaS || isVercelFaaS)) {
```
durran, line 181: Just a question here - have we tested on Vercel? Since Vercel runs on Lambda won't we have `AWS_EXECUTION_ENV`, `AWS_LAMBDA_RUNTIME_API` in the env as well as `VERCEL`? (Which would cause this condition to never be met?) I was thinking all the logic could be simplified and just check for Vercel first and then the others.

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,227 @@
...
+      faasEnv.set('memory_mb', new Int32(FUNCTION_MEMORY_MB));
+    }
+
+    if (FUNCTION_TIMEOUT_SEC.length > 0 && Number.isInteger(+FUNCTION_TIMEOUT_SEC)) {
+      faasEnv.set('timeout_sec', new Int32(FUNCTION_TIMEOUT_SEC));
+    }
+
+    faasEnv.set('name', 'gcp.func');
+    return faasEnv;
+  } else if (isVercelFaaS && !(isAWSFaaS || isAzureFaaS || isGCPFaaS)) {
```
durran, line 213: Same Lambda/Vercel comment as above.

test/integration/mongodb-handshake/mongodb-handshake.prose.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      expectedProvider: 'gcp.func',
+      env: [
+        ['K_SERVICE', 'servicename'],
+        ['FUNCTION_MEMORY_MB', '1024'],
+        ['FUNCTION_TIMEOUT_SEC', '60'],
+        ['FUNCTION_REGION', 'us-central1']
+      ]
+    },
+    {
+      context: '4. Valid Vercel',
```
durran, line 43: Per my comments before, I am assuming Vercel will also have the AWS Lambda env variables present as well but would need to verify on an actual Vercel function.

test/integration/mongodb-handshake/mongodb-handshake.test.ts
durran: Just a note on the comment - load balancer mode has its own custom server selector that just grabs the same server every time (the only server in the topology) so it never gets actual client side server selection errors.

durran (changes requested):

src/cmap/handshake/client_metadata.ts
nbbeeken: The spec has requirement:

> If none of those variables or variables for multiple names are populated the client.env value MUST be entirely omitted.

So this prevents any mutliple variables from getting added to the metadata, but yes I'm following up in the channel, I had the same question.

src/cmap/handshake/client_metadata.ts
nbbeeken: Yea I agree, and that's easy enough to manage.

test/integration/mongodb-handshake/mongodb-handshake.prose.test.ts
nbbeeken: We'll follow up from the channel but regardless these are the verbatim spec prose tests

src/cmap/handshake/client_metadata.ts
nbbeeken: This has been fixed and added a test, vercel does have the same env vars

src/cmap/handshake/client_metadata.ts
nbbeeken: fixed

test/integration/mongodb-handshake/mongodb-handshake.prose.test.ts
nbbeeken: I wanna leave the prose as is to match the spec, we can pull in an update to the spec when there is one. I did add a unit test for this case

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  MongoServerError,
+  MongoServerSelectionError
+} from '../../mongodb';
+
+describe('MongoDB Handshake Node tests', () => {
+  let client;
+
+  context('when hello is too large', () => {
+    before(() => {
+      sinon.stub(Connection.prototype, 'command').callsFake(function (ns, cmd, options, callback) {
```
baileympearson, line 16: Why is stubbing necessary?  And if we're stubbing, should these be unit? (my preference is no stubbing and leave as integration)

nbbeeken: The logic in our makeMetadata prevents me from constructing a metadata that's too large, so I'm using sinon to circumvent that. I still need a server, I'm invoking the real command method after I've made the 'too large' metadata document

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,511 @@
...
+        for (const { context, env: faasVariables, outcome } of testsForEnv) {
+          const setupEnv = () => {
+            for (const [k, v] of faasVariables) {
+              if (v != null) {
+                process.env[k] = v;
+              }
+            }
+          };
+
+          const cleanupEnv = () => {
```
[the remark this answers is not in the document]
nbbeeken: Tried it, seems to have the same issue, you need to call the hook creation inside the for loop so then all the tests share all the before hooks so the env isn't setup correctly for each test.

test/integration/mongodb-handshake/mongodb-handshake.test.ts
baileympearson: I'll be upfront and say I think this is unnecessary.  The changes you're proposing here are testing server behavior, not driver behavior.  If the server changes / removes the metadata limit, the driver doesn't care and we should continue to pass the prose tests (even if we silently pass them).

If we do care about checking whether or not the server throws if a document is too large, then we should directly test that using a `runCommand` with a manually constructed `hello`.  But I don't think we should test this.

nbbeeken: Oh good call on runCommand that would be better, I just didn't think of that

nbbeeken: Ah wait, building a hello isn't super straight forward, I'd have to limit the breadth of the server testing, by using the stubbing here I can cover auth and no-auth envs

test/integration/mongodb-handshake/mongodb-handshake.prose.test.ts
```diff
@@ -0,0 +1,107 @@
+import { expect } from 'chai';
+
+import { getFAASEnv, MongoClient } from '../../mongodb';
+
+describe('FAAS Environment Prose Tests', function () {
```
dariakp, line 5:
```suggestion
describe('Handshake Prose Tests', function () {
```

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as sinon from 'sinon';
+
+import {
+  Connection,
+  LEGACY_HELLO_COMMAND,
+  MongoServerError,
+  MongoServerSelectionError
+} from '../../mongodb';
+
+describe('MongoDB Handshake Node tests', () => {
```
dariakp, line 11:
```suggestion
describe('MongoDB Handshake', () => {
```

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -0,0 +1,45 @@
...
+            callback
+          );
+        }
+        return command(ns, cmd, options, callback);
+      });
+    });
+
+    after(() => sinon.restore());
+
+    it('client fails to connect with an error relating to size', async function () {
```
dariakp, line 34:
```suggestion
    it('should fail with an error relating to size', async function () {
```

test/tools/cmap_spec_runner.ts
```diff
@@ -370,7 +370,10 @@ async function runCmapTest(test: CmapTest, threadContext: ThreadContext) {
     delete poolOptions.backgroundThreadIntervalMS;
   }
 
-  const metadata = makeClientMetadata({ appName: poolOptions.appName, driverInfo: {} });
+  const metadata = makeClientMetadata({
```
dariakp, line 373: why did we need to touch this file?

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
+  getFAASEnv,
+  Int32,
+  LimitedSizeDocument,
+  makeClientMetadata,
+  MongoInvalidArgumentError,
+  ObjectId
+} from '../../../mongodb';
+
+describe('client metadata module', () => {
+  describe('class LimitedSizeDocument', () => {
```
dariakp, line 17:
```suggestion
  describe('new LimitedSizeDocument()', () => {
```

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
+    // For the sake of testing the size limiter features
+    // We test document: { _id: ObjectId() }
+    // 4 bytes + 1 type byte + 4 bytes for key + 12 bytes Oid + 1 null term byte
+    // = 22 bytes
+
+    it('sets key and value that fit within maxSize', () => {
```
dariakp, line 23:
```suggestion
    it('allows setting a key and value that fit within maxSize', () => {
```

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
+      const doc = new LimitedSizeDocument(22);
+      expect(doc.ifItFitsItSits('_id', new ObjectId())).to.be.true;
+      expect(doc.toObject()).to.have.all.keys('_id');
+    });
+
+    it('ignores ifItFitsItSits that are over size', () => {
```
dariakp, line 29:
```suggestion
    it('ignores attempts to set key-value pairs that are over size', () => {
```

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
+          delete process.env[envVariable];
+        });
+        it('determines the correct provider', () => {
+          expect(getFAASEnv()?.get('name')).to.equal(provider);
+        });
+      });
+    }
+
+    context('when there is no FAAS provider data in the env', () => {
+      it('parses no FAAS provider', () => {
```
dariakp, line 61:
```suggestion
      it('returns null', () => {
```

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
+          // aws
+          process.env.AWS_EXECUTION_ENV = 'non-empty-string';
+          // azure
+          process.env.FUNCTIONS_WORKER_RUNTIME = 'non-empty-string';
+        });
+        after(() => {
+          delete process.env.AWS_EXECUTION_ENV;
+          delete process.env.FUNCTIONS_WORKER_RUNTIME;
+        });
+        it('parses no FAAS provider', () => {
```
dariakp, line 78:
```suggestion
        it('returns null, () => {
```

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
-        },
-        os: {
-          type: os.type(),
-          name: process.platform,
-          architecture: process.arch,
-          version: os.release()
-        },
-        platform: `Node.js ${process.version}, ${os.endianness()}`
+    context('when driverInfo.platform is provided', () => {
+      it('appends driverInfo.platform to the platform field', () => {
```
dariakp, line 129: should there be a case for when it exceeds the allowed size?

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
+          },
+          platform: `Node.js ${process.version}, ${os.endianness()}|myPlatform`
+        });
       });
     });
 
-    it('does not set the application field', () => {
-      expect(metadata).not.to.have.property('application');
+    context('when driverInfo.name is provided', () => {
+      it('appends driverInfo.name to the driver.name field', () => {
```
dariakp, line 151: case for exceeding max size?

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
-      const metadata = makeClientMetadata(options);
-
-      it('truncates the application name to <=128 bytes', () => {
-        expect(metadata.application?.name).to.be.a('string');
-        // the above assertion fails if `metadata.application?.name` is undefined, so
-        // we can safely assert that it exists
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        expect(Buffer.byteLength(metadata.application!.name, 'utf8')).to.equal(128);
+    context('when driverInfo.version is provided', () => {
+      it('appends driverInfo.version to the version field', () => {
```
dariakp, line 173: case for exceeding max size?

test/unit/cmap/handshake/client_metadata.test.ts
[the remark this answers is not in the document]
dariakp: options wise, you can add the beforeEach hook outside the loop that generates the test cases; and you can switch on `currentTest.title` for super custom setup/cleanup

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,523 @@
...
+      it('does not attach it to the metadata', () => {
+        expect(makeClientMetadata({ driverInfo: {} })).not.to.have.nested.property('aws.memory_mb');
+      });
+    });
+  });
+
+  describe('metadata truncation', function () {
+    afterEach(() => sinon.restore());
+
+    context('when driverInfo is too large', () => {
```
dariakp, line 436: error throwing isn't truncation, these cases should live with the respective blocks for setting this info above

dariakp (changes requested):

src/cmap/handshake/client_metadata.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    if (FUNCTION_TIMEOUT_SEC.length > 0 && Number.isInteger(+FUNCTION_TIMEOUT_SEC)) {
+      faasEnv.set('timeout_sec', new Int32(FUNCTION_TIMEOUT_SEC));
+    }
+
+    faasEnv.set('name', 'gcp.func');
+    return faasEnv;
+  }
+
+  return null;
```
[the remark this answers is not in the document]
nbbeeken: From slack, this is intentional

test/unit/cmap/handshake/client_metadata.test.ts
nbbeeken: Moved up to the section on driverInfo

test/unit/cmap/handshake/client_metadata.test.ts
nbbeeken: the throwing tests now moved up here cover this

test/unit/cmap/handshake/client_metadata.test.ts
nbbeeken: the throwing tests now moved up here cover this

test/unit/cmap/handshake/client_metadata.test.ts
nbbeeken: the throwing tests now moved up here cover this

test/unit/cmap/handshake/client_metadata.test.ts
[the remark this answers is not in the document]
nbbeeken: Updated to put the context inside the for loop, sample output

```txt
FAAS metadata application to handshake
      aws - no additional metadata
        ✔ returns { name: 'aws.lambda' } under env property (3762ms)
        ✔ places name as the last key in map
      aws - AWS_REGION provided
        ✔ returns { name: 'aws.lambda', region: 'non-null' } under env property
        ✔ places name as the last key in map
      aws - AWS_LAMBDA_FUNCTION_MEMORY_SIZE provided
        ✔ returns { name: 'aws.lambda', memory_mb: new Int32(3) } under env property
        ✔ places name as the last key in map
```

durran:
@dariakp Are your comments all adequately addressed? And @nbbeeken I think we have just one merge conflict now.

test/unit/cmap/handshake/client_metadata.test.ts
```diff
@@ -1,149 +1,504 @@
...
-          version: NODE_DRIVER_VERSION
-        },
-        os: {
-          type: os.type(),
-          name: process.platform,
-          architecture: process.arch,
-          version: os.release()
-        },
-        platform: `Node.js ${process.version}, ${os.endianness()}`
+    context('when driverInfo is too large', () => {
```
dariakp, line 131: I can live with this, but it would be more clear if you just put the error cases with the appropriate properties where I left the original set of comments. That way it doesn't read "when the driverInfo is too large, throws a [single] error covering [all of] name, version, and platform".

durran (approved):

PR #3629, opened 2023-04-06
feat(NODE-4814): implement remaining severity logging methods

W-A-James:
#### What is changing?

Adds in the remainder of the severity loggers whose base designs were implemented in NODE-5169

No


W-A-James:
Question: Do we want to remove the entries in the `SeverityLevel` enum that correspond to the severities that we are not using?

These would be:

- `emergency`
- `alert`
- `critical`
- `notice`

W-A-James:
Question: Do we want to have the helper that corresponds with the `warn` severity be named `warn` or `warning`? Doing the former allows for easy accessing on the logger object by using the severity name directly as a property, so I skew towards that approach.

test/unit/mongo_logger.test.ts
```diff
@@ -683,20 +677,13 @@ describe('class MongoLogger', function () {
...
-            const EMERGENCY = 0;
             // Calls all severity logging methods with a level less than or equal to what severityLevel
-            for (
-              let l = SEVERITY_LEVEL_MAP.getNumericSeverityLevel(severityLevel);
-              l >= EMERGENCY;
-              l--
-            ) {
-              const severity = SEVERITY_LEVEL_MAP.getSeverityLevelName(l);
+            for (let i = index; i >= 0; i--) {
+              const severity = severities[i];
```
baileympearson, line 682: if we type `severities` as `SeverityLevel[]`, we should be able to remove the cast in line 683

baileympearson (changes requested):
Tiny change

baileympearson:
CI failures are unrelated.

src/mongo_logger.ts
```diff
@@ -430,14 +430,18 @@ export class MongoLogger {
   maxDocumentLength: number;
   logDestination: MongoDBLogWritable | Writable;
 
+  error = this.log.bind(this, 'error');
+  warn = this.log.bind(this, 'warn');
+  info = this.log.bind(this, 'info');
+  debug = this.log.bind(this, 'debug');
+  trace = this.log.bind(this, 'trace');
```
nbbeeken, line 437: Can we leave tsdoc about when to use each. For example an important detail about error from the spec is that it is to be used when there is an error that cannot be reported via API methods

nbbeeken (changes requested):

src/mongo_logger.ts
```diff
@@ -430,10 +430,30 @@ export class MongoLogger {
   maxDocumentLength: number;
   logDestination: MongoDBLogWritable | Writable;
 
+  /**
+   * This method should be used when logging errors that do not have a public driver API for
+   * reporting errors.
+   **/
```
nbbeeken, line 436: The doc comments here have a bit of formatting differences than the rest of the code, I don't think we have linting for this: `**/` -> `*/`

Ex.
```js
  /**
   * Close the client and its underlying connections
   *
   * @param force - Force close, emitting no events
   */
```

src/mongo_logger.ts
```diff
@@ -430,10 +430,30 @@ export class MongoLogger {
...
   error = this.log.bind(this, 'error');
+  /**
+   * This method should be used to log situations where undesirable application behaviour might
+   * occur. E.g.: encountering an unrecognized option in a connection string
```
nbbeeken, line 440:
```suggestion
   * occur. For example, encountering an unrecognized option in a connection string.
```

We can just complete the sentence. Also this example doesn't apply to our driver since we error on unrecognized options

nbbeeken (changes requested):

nbbeeken (approved):

baileympearson (approved):

PR #3631, opened 2023-04-10
fix(NODE-5171): allow `upsertedId` to be null in `UpdateResult`

baileympearson:
#### What is changing?

`UpdateResult` is now generic over a collection schema type.  `UpdateResult.upsertedId` has two changes: it can now be null, and the type is inferred from the collection schema.  

`Collection.update*` methods now return an `UpdateResult` that is generic over the collection's schema.

### DIFF
--- a/src/operations/update.ts
+++ b/src/operations/update.ts
@@ -23,8 +23,11 @@ export interface UpdateOptions extends CommandOperationOptions {
   let?: Document;
 }
 
-/** @public */
-export interface UpdateResult {
+/**
+ * @public
+ * `TSchema` is the schema of the collection
+ */
+export interface UpdateResult<TSchema = Document> {
   /** Indicates whether this write result was acknowledged. If not, then all other members of this result will be undefined */
   acknowledged: boolean;
   /** The number of documents that matched the filter */
@@ -34,7 +37,7 @@ export interface UpdateResult {
   /** The number of documents that were upserted */
   upsertedCount: number;
   /** The identifier of the inserted document if an upsert took place */
-  upsertedId: ObjectId | null;
+  upsertedId: InferIdType<TSchema> | null;
 }
 
 /** @public */
--- a/src/operations/update.ts
+++ b/src/operations/update.ts
@@ -28,7 +28,7 @@ export interface UpdateOptions extends CommandOperationOptions {
  * @public
  * `TSchema` is the schema of the collection
  */
-export interface UpdateResult<TSchema = Document> {
+export interface UpdateResult<TSchema extends Document = Document> {
   /** Indicates whether this write result was acknowledged. If not, then all other members of this result will be undefined */
   acknowledged: boolean;
   /** The number of documents that matched the filter */
--- a/src/operations/update.ts
+++ b/src/operations/update.ts
@@ -29,12 +29,12 @@ export interface UpdateResult {
   acknowledged: boolean;
   /** The number of documents that matched the filter */
   matchedCount: number;
-  /** The number of documents that were modified */
+  /** The number of documents ∂that were modified */
   modifiedCount: number;
   /** The number of documents that were upserted */
   upsertedCount: number;
   /** The identifier of the inserted document if an upsert took place */
-  upsertedId: ObjectId;
+  upsertedId: ObjectId | null;
 }
 
 /** @public */
--- a/src/operations/update.ts
+++ b/src/operations/update.ts
@@ -1,6 +1,7 @@
-import type { Document, ObjectId } from '../bson';
+import type { Document } from '../bson';
 import type { Collection } from '../collection';
 import { MongoCompatibilityError, MongoInvalidArgumentError, MongoServerError } from '../error';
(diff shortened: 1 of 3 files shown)


src/operations/update.ts
```diff
@@ -29,12 +29,12 @@ export interface UpdateResult {
...
   /** The number of documents that matched the filter */
   matchedCount: number;
-  /** The number of documents that were modified */
+  /** The number of documents ∂that were modified */
   modifiedCount: number;
   /** The number of documents that were upserted */
   upsertedCount: number;
   /** The identifier of the inserted document if an upsert took place */
-  upsertedId: ObjectId;
+  upsertedId: ObjectId | null;
```
dariakp, line 37: This doesn't account for [setOnInsert](https://www.mongodb.com/docs/v6.0/reference/operator/update/setOnInsert/#example)

src/operations/update.ts
```diff
@@ -29,12 +29,12 @@ export interface UpdateResult {
   acknowledged: boolean;
   /** The number of documents that matched the filter */
   matchedCount: number;
-  /** The number of documents that were modified */
+  /** The number of documents ∂that were modified */
```
dariakp, line 32: stray symbol here

dariakp (changes requested):

src/operations/update.ts
baileympearson: What are you saying the correct type should be?  Either way, we have no way of knowing what the actual id type is.  Should we type it as unknown?

dariakp: Both the filter for the operation and the `$set` and `$setOnInsert` operators match against `TSchema`, meaning that if an `_id` is passed it at all, it would be constrained to `InferIdType`, so it would make sense to do that here, too. It's true that this does not account for the `_id` being missing entirely resulting in the `ObjectId` being generated server-side, but this is a similar issue to what we have in, for example, `insertOne` and `insertMany` (if the user does not provide an `_id` or a `pkFactory`), with the only difference being that a `pkFactory` can't be applied here as a backup. So if we follow our existing pattern, we would call the use of the upsert without specifying an `_id` in the case of a custom `_id` `TSchema` "incorrect usage", and advise the users in our documentation because TypeScript can't effectively catch that. Our operating assumption is that users performing upserts do not want to introduce inconsistencies into their schema, if one exists. But we can discuss that further with the team if you believe there is a stronger argument to be made for `unknown`.

src/operations/update.ts
```diff
@@ -23,8 +24,11 @@ export interface UpdateOptions extends CommandOperationOptions {
   let?: Document;
 }
 
-/** @public */
-export interface UpdateResult {
+/**
+ * @public
+ * `TSchema` is the schema of the collection
+ */
+export interface UpdateResult<TSchema = Document> {
```
nbbeeken, line 31:
```suggestion
export interface UpdateResult<TSchema extends Document = Document> {
```

We should have the same constraint here that `Collection` has so that this type doesn't support something more than it would via the `Collection<TSchema>` generic.

nbbeeken (changes requested):

test/types/community/collection/updateX.test-d.ts
```diff
@@ -423,3 +424,16 @@ export async function testPushWithId(): Promise<void> {
     }
   );
 }
+
+{
+  // NODE-5171 - UpdateResult is generic over the collection schema and infers the id type
```
dariakp, line 429: Can we also add the corresponding not-assignable tests for the two cases?

dariakp (changes requested):

nbbeeken (approved):

test/types/community/collection/updateX.test-d.ts
```diff
@@ -431,9 +431,18 @@ export async function testPushWithId(): Promise<void> {
...
+
   const collectionWithSchema = {} as any as Collection<{ _id: number }>;
   expectAssignable<Promise<{ upsertedId: number | null }>>(
     collectionWithSchema.updateOne({ _id: 1234 }, {})
   );
   expectAssignable<Promise<{ upsertedId: number | null }>>(collectionWithSchema.updateMany({}, {}));
+  expectNotAssignable<Promise<{ upsertedId: ObjectId | null }>>(
+    collectionWithSchema.updateOne({ _id: 1234 }, {})
+  );
+  expectNotAssignable<Promise<{ upsertedId: ObjectId | null }>>(
```
dariakp, line 445: I think this one should be `number` no?

dariakp (changes requested):

test/types/community/collection/updateX.test-d.ts
baileympearson: No?  The schema mandates that _id is of type `number`, so we expect `upsertedId` must be `number | null` (not `ObjectId`)

dariakp: sorry I misread that

dariakp (approved):

PR #3635, opened 2023-04-12
feat(NODE-4815): stringify and truncate BSON documents in log messages

W-A-James:
#### What is changing?

Adds the `maybeTruncate` helper and calls it in the `defaultLogTransform` helper function to produce spec-compliant logs.

### DIFF
--- a/src/mongo_logger.ts
+++ b/src/mongo_logger.ts
@@ -54,6 +54,8 @@ export const SeverityLevel = Object.freeze({
 } as const);
 
 /** @internal */
+export const DEFAULT_MAX_DOCUMENT_LENGTH = 1000;
+/** @internal */
 export type SeverityLevel = (typeof SeverityLevel)[keyof typeof SeverityLevel];
 
 /** @internal */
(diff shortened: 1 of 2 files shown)


src/mongo_logger.ts
```diff
@@ -292,22 +306,23 @@ function attachConnectionFields(
 }
 
 function defaultLogTransform(
-  logObject: LoggableEvent | Record<string, any>
+  logObject: LoggableEvent | Record<string, any>,
+  maxDocumentLength: number = DEFAULT_MAX_DOCUMENT_LENGTH
```
nbbeeken, line 310: We always will default this in the options do we need to redo that here and in truncation?

src/mongo_logger.ts
```diff
@@ -254,6 +256,18 @@ export interface LogConvertible extends Record<string, any> {
   toLog(): Record<string, any>;
 }
 
+/** @internal */
+export function maybeTruncate(
```
nbbeeken, line 260: Can we move EJSON stringify calls into the same function that worries about enforcing the size limit? maybeTruncate -> stringifyWithLength (or whatever name seems cooler)

src/mongo_logger.ts
```diff
@@ -254,6 +256,18 @@ export interface LogConvertible extends Record<string, any> {
...
+  ejson: string,
+  maxDocumentLength: number = DEFAULT_MAX_DOCUMENT_LENGTH
+): string {
+  if (maxDocumentLength === 0) {
+    return ejson;
+  }
+
+  return ejson.length > maxDocumentLength ? `${ejson.slice(0, maxDocumentLength)}...` : ejson;
```
nbbeeken, line 268:
```suggestion
  return maxDocumentLength !== 0 && ejson.length > maxDocumentLength ? `${ejson.slice(0, maxDocumentLength)}...` : ejson;
```

nbbeeken (changes requested):

src/mongo_logger.ts
W-A-James: Do we have a preference for where we default it? I would skew to doing it in the log transform/logger itself

nbbeeken: The logger class seems good

nbbeeken (approved):
LGTM

PR #3636, opened 2023-04-12
feat(NODE-5188): add alternative runtime detection to client metadata

nbbeeken:
#### What is changing?

Changes platform metadata based on globals in the driver's runtime environement.

Node.js compatibility in Deno and Bun have made it possible to run the driver, it is useful to denote the difference between the runtimes in the platform information.

### DIFF
--- a/src/cmap/handshake/client_metadata.ts
+++ b/src/cmap/handshake/client_metadata.ts
@@ -237,17 +237,13 @@ export function getFAASEnv(): Map<string, string | Int32> | null {
 
 /**
  * @internal
- * This type represents the global Deno object and the minimal type contract we
- * expect it to satisfy. In order to not ship code in the driver that would break
- * future versions of this runtime assume all properties are nullish.
+ * This type represents the global Deno object and the minimal type contract we expect it to satisfy.
  */
 declare const Deno: { version?: { deno?: string } } | undefined;
 
 /**
  * @internal
- * This type represents the global Bun object and the minimal type contract we
- * expect it to satisfy. In order to not ship code in the driver that would break
- * future versions of this runtime assume all properties are nullish.
+ * This type represents the global Bun object and the minimal type contract we expect it to satisfy.
  */
 declare const Bun: { (): void; version?: string } | undefined;
 
@@ -255,9 +251,9 @@ declare const Bun: { (): void; version?: string } | undefined;
  * @internal
  * Get current JavaScript runtime platform
  *
- * NOTE: The version information fetching is intentionally written aggressively defensive
+ * NOTE: The version information fetching is intentionally written defensively
  * to avoid having a released driver version that becomes incompatible
- * with a future change to these global objects
+ * with a future change to these global objects.
  */
 function getRuntimeInfo(): string {
   if ('Deno' in globalThis) {
--- a/src/cmap/handshake/client_metadata.ts
+++ b/src/cmap/handshake/client_metadata.ts
@@ -277,7 +277,10 @@ function getRuntimeInfo(): string {
 
   if ('Bun' in globalThis) {
     const version =
-      Bun != null && typeof Bun === 'object' && 'version' in Bun && typeof Bun.version === 'string'
+      Bun != null &&
+      (typeof Bun === 'function' || typeof Bun === 'object') &&
+      'version' in Bun &&
+      typeof Bun.version === 'string'
         ? Bun.version
         : '0.0.0';
 
--- a/src/cmap/handshake/client_metadata.ts
+++ b/src/cmap/handshake/client_metadata.ts
@@ -116,12 +116,12 @@ export function makeClientMetadata(options: MakeClientMetadataOptions): ClientMe
     );
   }
 
-  const platformInfo =
-    platform.length > 0
-      ? `Node.js ${process.version}, ${os.endianness()}|${platform}`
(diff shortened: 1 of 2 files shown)


nbbeeken:
Server log lines:
```json
{"t":{"$date":"2023-04-12T15:53:58.738-04:00"},"s":"I",  "c":"NETWORK",  "id":51800,   "ctx":"conn6","msg":"client metadata","attr":{"remote":"127.0.0.1:64794","client":"conn6","doc":{"driver":{"name":"nodejs","version":"5.2.0"},
"platform":"Deno v1.32.3, LE",
"os":{"name":"darwin","architecture":"arm64","version":"22.4.0","type":"Darwin"}}}}

{"t":{"$date":"2023-04-12T16:03:31.216-04:00"},"s":"I",  "c":"NETWORK",  "id":51800,   "ctx":"conn18","msg":"client metadata","attr":{"remote":"127.0.0.1:59421","client":"conn18","doc":{"driver":{"name":"nodejs","version":"5.2.0"},
"platform":"Bun v0.5.4, LE",
"os":{"name":"darwin","architecture":"arm64","version":"22.4.0","type":"Darwin"}}}}
```

alexbevi (approved):
LGTM

src/cmap/handshake/client_metadata.ts
```diff
@@ -234,3 +234,58 @@ export function getFAASEnv(): Map<string, string | Int32> | null {
...
+/**
+ * @internal
+ * Get current JavaScript runtime platform
+ *
+ * NOTE: The version information fetching is intentionally written aggressively defensive
+ * to avoid having a released driver version that becomes incompatible
+ * with a future change to these global objects
+ */
+function getRuntimeInfo(): string {
+  if ('Deno' in globalThis) {
```
durran, line 263: Why are all the checks necessary? Can't we assume if `Deno` or `Bun` are present in the global namespace that they conform to the type we expect? And that if a user was to randomly put them there all bets are off?

durran (changes requested):

src/cmap/handshake/client_metadata.ts
```diff
@@ -234,3 +234,58 @@ export function getFAASEnv(): Map<string, string | Int32> | null {
...
+    const version =
+      Deno != null &&
+      typeof Deno === 'object' &&
+      'version' in Deno &&
+      Deno.version != null &&
+      typeof Deno.version === 'object' &&
+      'deno' in Deno.version &&
+      typeof Deno.version.deno === 'string'
+        ? Deno.version.deno
+        : '0.0.0';
```
nbbeeken, line 273: Add `-unknown`

src/cmap/handshake/client_metadata.ts
```diff
@@ -234,3 +234,58 @@ export function getFAASEnv(): Map<string, string | Int32> | null {
...
+ * with a future change to these global objects
+ */
+function getRuntimeInfo(): string {
+  if ('Deno' in globalThis) {
+    const version =
+      Deno != null &&
+      typeof Deno === 'object' &&
+      'version' in Deno &&
+      Deno.version != null &&
+      typeof Deno.version === 'object' &&
```
nbbeeken, line 269: use optional chaining

src/cmap/handshake/client_metadata.ts
nbbeeken: We just want to write code that won't end up throwing if there's a potential change. We can reduce the checks here to optional chains though

baileympearson (approved):

src/cmap/handshake/client_metadata.ts
```diff
@@ -234,3 +234,39 @@ export function getFAASEnv(): Map<string, string | Int32> | null {
...
+ * @internal
+ * Get current JavaScript runtime platform
+ *
+ * NOTE: The version information fetching is intentionally written defensively
+ * to avoid having a released driver version that becomes incompatible
+ * with a future change to these global objects.
+ */
+function getRuntimeInfo(): string {
+  if ('Deno' in globalThis) {
+    const version = typeof Deno?.version?.deno === 'string' ? Deno?.version?.deno : '0.0.0-unknown';
```
durran, line 260: `typeof new String(‘’)` returns “object” so a safer check: https://github.com/mongodb/node-mongodb-native/pull/3637/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R69

durran (changes requested):

src/cmap/handshake/client_metadata.ts
nbbeeken: Is this something we want to support? Generally, none of our code supports string objects / number objects / etc. I would be suspicious of code that uses String instances.

Unrelated but here is how String objects serialize in BSON, just to exemplify the degree to which we assume these constructors are assumed to not be used:
```js
> BSON.deserialize(BSON.serialize({ a: new String('abc') }))
{ a: { '0': 'a', '1': 'b', '2': 'c' } }
```

durran: Didn't know if we wanted to or not, given the serialisation example it seems not then, although unexpected. But I guess if no one has ever reported that no one really uses them.

durran (approved):

PR #3637, opened 2023-04-13
feat(NODE-5191): OIDC Auth Updates

durran:


src/cmap/auth/mongodb_oidc.ts
```diff
@@ -11,11 +11,8 @@ import type { Workflow } from './mongodb_oidc/workflow';
  * @experimental
  */
 export interface OIDCMechanismServerStep1 {
-  authorizationEndpoint?: string;
-  tokenEndpoint?: string;
-  deviceAuthorizationEndpoint?: string;
+  issuer?: string;
```
addaleax, line 14:
```suggestion
  issuer: string;
```

it's not marked optional in the [server IDL](https://github.com/10gen/mongo-enterprise-modules/blob/8cdcc2da55448c3f85f9a28aad958ae11f7dc55f/src/sasl/oidc_parameters.idl#L13-L19) either, and the callback needs this value :)

src/utils.ts
```diff
@@ -59,6 +61,15 @@ export const ByteUtils = {
   }
 };
 
+/**
+ * Use this to test if an object is a string. This is because
+ * typeof new String('test') is 'object' and not 'string'.
```
addaleax, line 66: Just curious, why do you want to accept `new String('test')` here?

durran: Just defensive if a user decided to define their strings in that manner.

src/cmap/auth/mongodb_oidc.ts
durran: Updated.

src/connection_string.ts
```diff
@@ -310,6 +310,17 @@ export function parseOptions(
     );
   }
 
+  const uriMechanismProperties = urlOptions.get('authMechanismProperties');
+  if (uriMechanismProperties) {
+    for (const property of uriMechanismProperties) {
+      if (property.includes('ALLOWED_HOSTS:')) {
```
addaleax, line 316:
```suggestion
      if (/(^|,)ALLOWED_HOSTS:/.test(property)) {
```

so that you don’t accidentally run into the (of course very unlikely) case that `ALLOWED_HOSTS:` is part of the value of another auth mechanism property?

src/utils.ts
```diff
@@ -59,6 +61,15 @@ export const ByteUtils = {
...
+ * @internal
+ */
+export function isString(value: any): boolean {
```
addaleax, line 69:
```suggestion
export function isString(value: any): value is (string | String) {
```

src/utils.ts
addaleax: It’s your call ultimately but I think it’s highly unlikely that users would do this (or at least would do this intentionally).

src/connection_string.ts
durran: Update to suggestion.

src/utils.ts
durran: Per conversation with Neal last week I'll change this to "typeof" since we don't generally handle `new String()` in the driver or BSON as a string, but rather an object.

src/utils.ts
durran: This method removed.

src/utils.ts
```diff
@@ -58,6 +58,23 @@ export const ByteUtils = {
   }
 };
 
+/**
+ * Determines if a connection's address matches a user provided list
+ * of domain wildcards.
+ */
+export function hostMatchesWildcards(address: string, wildcards: string[]): boolean {
+  const host = HostAddress.fromString(address).host;
```
addaleax, line 66: This check should probably always pass for UNIX sockets? (If yes, that can be a follow-up as far as I am concerned.)

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -95,19 +98,32 @@ export class MongoDBOIDC extends AuthProvider {
...
-    if (!credentials) {
-      throw new MongoMissingCredentialsError('AuthContext must provide credentials.');
+    const credentials = getCredentials(authContext);
+    const { connection } = authContext;
+    const mechanismProperties = credentials.mechanismProperties;
+    const allowedHosts = mechanismProperties.ALLOWED_HOSTS || DEFAULT_ALLOWED_HOSTS;
+    const providerName = mechanismProperties.PROVIDER_NAME;
+    // Cloud provider automatic auth does not validate allowed hosts.
+    if (!providerName && !hostMatchesWildcards(connection.address, allowedHosts)) {
+      throw new MongoInvalidArgumentError('Host does not match provided ALLOWED_HOSTS values');
```
addaleax, line 108: Does it make sense to include the hostname in the error message? That feels like it could improve debugging quite a bit.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -95,19 +98,32 @@ export class MongoDBOIDC extends AuthProvider {
...
-    const { credentials } = authContext;
-
-    if (!credentials) {
-      throw new MongoMissingCredentialsError('AuthContext must provide credentials.');
+    const credentials = getCredentials(authContext);
+    const { connection } = authContext;
+    const mechanismProperties = credentials.mechanismProperties;
+    const allowedHosts = mechanismProperties.ALLOWED_HOSTS || DEFAULT_ALLOWED_HOSTS;
+    const providerName = mechanismProperties.PROVIDER_NAME;
+    // Cloud provider automatic auth does not validate allowed hosts.
```
addaleax, line 106: This seems reasonable but I’m not quite seeing where this behavior is written down in the spec – it should probably be mentioned there?

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -27,219 +42,226 @@ export class CallbackWorkflow implements Workflow {
...
     const entry = this.cache.getEntry(
       connection.address,
       credentials.username,
-      request || null,
-      refresh || null
+      requestCallback,
+      refreshCallback || null
     );
+    let result;
+    // Reauthentication must go through all the steps again regards of a cache entry
```
addaleax, line 79: Might be a typo?

```suggestion
    // Reauthentication must go through all the steps again regardless of a cache entry
```

src/cmap/auth/mongodb_oidc.ts
durran: I've updated to include the hostname and ALLOWED_HOSTS in the error message.

src/cmap/auth/mongodb_oidc.ts
durran: I don't think the behaviour was explicitly called out in the spec but there is a prose test for it: https://github.com/mongodb/specifications/pull/1381/files#diff-93f94aaea2345c5e72ced4859fe1cefea928499f5b4019961c568256dbf72b62R136

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: I actually moved this comment down and changed the wording as the reauth behaviour changed slightly.

src/utils.ts
durran: I added the ability to check the unix socket now from the host address. Since it's not defined in the spec I've just assumed the simple "*/" leading wildcard for now and think it's fine for a follow up on the spec after.

addaleax: Alright, makes sense to me as a temporary solution to allow UNIX sockets :+1:

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -236,13 +236,13 @@ export class CallbackWorkflow implements Workflow {
...
-        result = await refreshCallback(serverInfo, context);
+        result = await withLock(refreshCallback)(serverInfo, context);
       } else {
-        result = await requestCallback(serverInfo, context);
+        result = await withLock(requestCallback)(serverInfo, context);
       }
     } else {
       // With no token in the cache we use the request callback.
-      result = await requestCallback(serverInfo, context);
+      result = await withLock(requestCallback)(serverInfo, context);
```
addaleax, line 245: This calls the callbacks with a new lock every time – you probably do need to store the lock as part of the cache entry (or that’s how I’m reading the spec text)

src/cmap/auth/mongodb_oidc/callback_lock_cache.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Get a cache key based on connection and credentials.
+ */
+function cacheKey(
+  connection: Connection,
+  credentials: MongoCredentials,
+  callbackHash: string
+): string {
+  return `${connection.address}-${credentials.username}=${callbackHash}`;
```
addaleax, line 101: Obviously this is not a concern in practice, but what if `me@example.com` authenticates to `cluster.net-bar.net` and `bar.net-me@example.com` authenticates to `cluster.net`?

```suggestion
  return JSON.stringify([connection.address, credentials.username, callbackHash}]);
```

addaleax (commented):
Locking looks good! :rocket:

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -73,19 +81,10 @@ export class MongoDBOIDC extends AuthProvider {
    * Authenticate using OIDC
    */
   override async auth(authContext: AuthContext): Promise<void> {
-    const { connection, credentials, response, reauthenticating } = authContext;
-
-    if (response?.speculativeAuthenticate) {
```
durran, line 78: Speculative auth is implemented in the the workflows.

src/cmap/auth/mongodb_oidc/callback_lock_cache.ts
```diff
@@ -0,0 +1,135 @@
...
+interface CallbacksEntry {
+  requestCallback: OIDCRequestFunction;
+  refreshCallback?: OIDCRefreshFunction;
+  callbackHash: string;
+}
+
+/**
+ * A cache of request and refresh callbacks per server/user.
+ */
+export class CallbackLockCache {
```
durran, line 38: This cache is now responsible for the hashing of the request and refresh functions, which is also then used by the `TokenEntryCache`. It also is responsible for wrapping the callbacks in a lock and storing them by the same cache key as the token results in the other cache. Main reasoning for not storing these in the other cache and creating a new one is that the callback functions have no expiration.

src/cmap/auth/mongodb_oidc/callback_lock_cache.ts
```diff
@@ -0,0 +1,135 @@
...
+    await lock;
+    lock = lock.then(() => callback(info, context));
+    return lock;
+  };
+}
+
+/**
+ * Get the hash string for the request and refresh functions.
+ */
+function hashFunctions(requestFn: OIDCRequestFunction, refreshFn?: OIDCRefreshFunction): string {
```
durran, line 119: This is the same as before, just moved to this cache impl from the other.

src/cmap/auth/mongodb_oidc/aws_service_workflow.ts
```diff
@@ -3,6 +3,9 @@ import { readFile } from 'fs/promises';
 import { MongoAWSError } from '../../../error';
 import { ServiceWorkflow } from './service_workflow';
 
+/** Error for when the token is missing in the environment. */
+const TOKEN_MISSING_ERROR = 'AWS_WEB_IDENTITY_TOKEN_FILE must be set in the environment.';
```
durran, line 7: This and other string constants just a simple refactoring of any non-interpolated string being a constant.

src/cmap/auth/mongodb_oidc/callback_lock_cache.ts
durran: I've updated to suggestion in both caches.

test/unit/cmap/auth/mongodb_oidc/callback_lock_cache.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      context('when a refresh callback exists', function () {
+        let requestCount = 0;
+        let refreshCount = 0;
+
+        const request = async () => {
+          requestCount++;
+          if (requestCount > 1) {
+            throw new Error('Cannot execute request simultaneously.');
+          }
+          await new Promise(resolve => setTimeout(() => resolve(), 1000));
```
nbbeeken, line 48: We have a `sleep` util, in `test/tools/utils.ts`

test/unit/cmap/auth/mongodb_oidc/callback_lock_cache.test.ts
```diff
@@ -0,0 +1,144 @@
...
+        username: 'test_user',
+        password: 'pwd',
+        source: '$external',
+        mechanismProperties: {}
+      });
+      const cache = new CallbackLockCache();
+
+      it('raises an error', function () {
+        try {
+          cache.getCallbacks(connection, credentials);
```
nbbeeken, line 28: This test does not fail if it *does not* raise an error
We've typically done:
```ts
const error = cache.getCallbacks(connection, credentials).catch(error => error);
expect(error).to.be.instanceOf(MongoInvalidArgumentError);
```
With the assumption that our APIs would never _return_ an instance of an error so we'd fail if the call succeeds.

test/unit/cmap/auth/mongodb_oidc/callback_lock_cache.test.ts
```diff
@@ -0,0 +1,144 @@
...
+        let refreshCount = 0;
+
+        const request = async () => {
+          requestCount++;
+          if (requestCount > 1) {
+            throw new Error('Cannot execute request simultaneously.');
+          }
+          await new Promise(resolve => setTimeout(() => resolve(), 1000));
+          requestCount--;
+          return Promise.resolve({ accessToken: '' });
```
nbbeeken, line 50: redundant promise here, it's already an async function

test/unit/cmap/auth/mongodb_oidc/callback_lock_cache.test.ts
```diff
@@ -0,0 +1,144 @@
...
+          }
+        });
+        const cache = new CallbackLockCache();
+        const { requestCallback, refreshCallback, callbackHash } = cache.getCallbacks(
+          connection,
+          credentials
+        );
+
+        it('puts a new entry in the cache', function () {
+          expect(cache.entries.size).to.equal(1);
```
nbbeeken, line 79: nit: you can use `lengthOf` assertions on maps / sets - it will print correctly:
`expected Map{} to have a size of 1 but got +0`

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -12,602 +27,1208 @@ describe('MONGODB-OIDC', function () {
...
-          });
-
-          // - Perform a ``find`` operation.
-          // - Clear the cache.
-          it('successfully authenticates', async function () {
-            const doc = await collection.findOne();
-            expect(doc).to.equal(null);
+          it('fails validation', async function () {
+            try {
+              await collection.findOne();
```
nbbeeken, line 238: This test won't fail if findOne returns successfully, see my other comment about adding a catch handler

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -12,602 +27,1208 @@ describe('MONGODB-OIDC', function () {
...
-          const testTokenFile = process.env.AWS_WEB_IDENTITY_TOKEN_FILE;
-          let client;
-          let collection;
+        // Create a client that uses the url ``mongodb://localhost/?authMechanism=MONGODB-OIDC&ignored=example.com`` a request callback, and an
+        // ``ALLOWED_HOSTS`` that contains ["example.com"].
+        // Assert that a ``find`` operation fails with a client-side error.
+        // Close the client.
+        context('when ALLOWED_HOSTS does not match', function () {
+          beforeEach(function () {
+            this.currentTest.skipReason = 'Will fail URI parsing as ignored is not a valid option';
```
nbbeeken, line 254: I don't follow the skipReason here, is this something we will be fixing or innate to our driver? like how we throw instead of warn on uri options?

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -12,602 +27,1208 @@ describe('MONGODB-OIDC', function () {
...
             collection = client.db('test').collection('test');
           });
 
-          // - Perform a ``find`` operation.
-          it('successfully authenticates', async function () {
-            const doc = await collection.findOne();
-            expect(doc).to.equal(null);
+          it('fails validation', async function () {
+            try {
+              await collection.findOne();
```
nbbeeken, line 295: no catch for success

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -12,602 +27,1208 @@ describe('MONGODB-OIDC', function () {
...
-            client?.close();
+        const requestCallback = async () => {
+          requestCounter++;
+          if (requestCounter > 1) {
+            throw new Error('Request callback was entered simultaneously.');
+          }
+          const token = await readFile(path.join(process.env.OIDC_TOKEN_DIR, 'test_user1'), {
+            encoding: 'utf8'
           });
+          await new Promise(resolve => setTimeout(() => resolve(), 3000));
```
nbbeeken, line 321: use `sleep`

nbbeeken (changes requested):

src/cmap/auth/mongodb_oidc/callback_lock_cache.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Get a cache key based on connection and credentials.
+ */
+function cacheKey(
+  connection: Connection,
+  credentials: MongoCredentials,
+  callbackHash: string
+): string {
+  return JSON.stringify([connection.address, credentials.username, callbackHash]);
```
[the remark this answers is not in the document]
addaleax: Here's why I suggested this: https://github.com/mongodb/node-mongodb-native/pull/3637#discussion_r1180646535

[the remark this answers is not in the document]
nbbeeken: Ah thanks! missed that context

src/cmap/auth/mongodb_oidc/aws_service_workflow.ts
```diff
@@ -3,6 +3,9 @@ import { readFile } from 'fs/promises';
 import { MongoAWSError } from '../../../error';
```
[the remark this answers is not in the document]
durran: Update to suggestion.

src/cmap/auth/mongodb_oidc/callback_lock_cache.ts
```diff
@@ -0,0 +1,135 @@
...
+    return lock;
+  };
+}
+
+/**
+ * Get the hash string for the request and refresh functions.
+ */
+function hashFunctions(requestFn: OIDCRequestFunction, refreshFn?: OIDCRefreshFunction): string {
+  let requestHash = FN_HASHES.get(requestFn || NO_FUNCTION);
+  let refreshHash = FN_HASHES.get(refreshFn || NO_FUNCTION);
```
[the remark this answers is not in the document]
durran: Updated.

src/cmap/auth/mongodb_oidc/callback_lock_cache.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (!requestHash && requestFn) {
```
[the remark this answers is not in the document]
durran: Actually the intention is `requestHash != null` which I have updated to.

src/cmap/auth/mongodb_oidc/callback_lock_cache.ts
```diff
@@ -0,0 +1,135 @@
...
+  OIDCCallbackContext,
+  OIDCRefreshFunction,
+  OIDCRequestFunction
+} from '../mongodb_oidc';
+
+/** Error message for when request callback is missing. */
+const REQUEST_CALLBACK_REQUIRED_ERROR =
+  'Auth mechanism property REQUEST_TOKEN_CALLBACK is required.';
+/* Counter for function "hashes".*/
+let FN_HASH_COUNTER = 0;
```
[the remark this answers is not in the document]
durran: No need to end the counter here and think it's probably a bit overkill in this case to use a generator.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -257,3 +267,30 @@ function continueCommandDocument(token: string, conversationId?: number): Docume
...
   };
 }
+
+/**
+ * Determines if a result returned from a request or refresh callback
+ * function is invalid. This means the result is nullish, doesn't contain
+ * the accessToken required field, and does not contain extra fields.
+ */
+function isCallbackResultInvalid(tokenResult: any): boolean {
+  if (!tokenResult) return true;
```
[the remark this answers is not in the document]
durran: Updated to proposed check.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -257,3 +267,30 @@ function continueCommandDocument(token: string, conversationId?: number): Docume
...
+  if (!tokenResult.accessToken) return true;
+  return !Object.getOwnPropertyNames(tokenResult).every(prop => RESULT_PROPERTIES.includes(prop));
```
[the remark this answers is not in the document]
durran: The later is the intention - anything extra whether enumerable or not should be checked.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -257,3 +267,30 @@ function continueCommandDocument(token: string, conversationId?: number): Docume
     payload: new Binary(BSON.serialize({ jwt: token }))
   };
 }
+
+/**
+ * Determines if a result returned from a request or refresh callback
+ * function is invalid. This means the result is nullish, doesn't contain
+ * the accessToken required field, and does not contain extra fields.
+ */
+function isCallbackResultInvalid(tokenResult: any): boolean {
```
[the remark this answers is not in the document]
durran: Updated to unknown and updated the subsequent check for "property in".

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -1,245 +1,255 @@
...
 }
 
 /**
- * Generate the saslContinue command document.
+ * Generate the finishing command document for authentication. Will be a
+ * saslStart or saslContinue depending on the presence of a conversation id.
  */
-function continueCommandDocument(token: string, conversationId?: number): Document {
+function finishCommandDocument(token: string, conversationId?: number): Document {
   if (conversationId) {
```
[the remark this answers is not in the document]
durran: Updated to check for null and typeof number

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -1,245 +1,255 @@
...
-    const tokenResult = await request(credentials.username, stepOneResult, TIMEOUT_MS);
-    // Validate the result.
-    if (!tokenResult || !tokenResult.accessToken) {
-      throw new MongoMissingCredentialsError(
-        'REQUEST_TOKEN_CALLBACK must return a valid object with an accessToken'
-      );
+    // Validate that the result returned by the callback is acceptable. If it is not
+    // we must clear the token result from the cache.
+    if (isCallbackResultInvalid(result)) {
+      this.cache.deleteEntry(connection.address, credentials.username || '', callbackHash);
```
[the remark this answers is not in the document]
durran: I've updated the `MongoCredentials` type to make username optional (as in OIDC it is not always needed) - and adjusted relevant code usages to to `credentials.username ?? ''`.

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
```diff
@@ -132,35 +101,6 @@ function expirationTime(expiresInSeconds?: number): number {
...
-  if (!refreshHash && refreshFn) {
-    // Create a new one for the function and put it in the map.
-    FN_HASH_COUNTER++;
-    refreshHash = FN_HASH_COUNTER;
-    FN_HASHES.set(refreshFn, FN_HASH_COUNTER);
-  }
-  return `${requestHash}-${refreshHash}`;
+function cacheKey(address: string, username: string, callbackHash: string): string {
+  return JSON.stringify([address, username, callbackHash]);
 }
```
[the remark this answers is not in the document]
durran: I refactored now to an abstract `Cache` class that both caches extend from with common functionality moved there (constructor, clear, and cacheKey).

src/cmap/auth/mongodb_oidc/workflow.ts
```diff
@@ -11,11 +11,12 @@ export interface Workflow {
   execute(
```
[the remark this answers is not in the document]
durran: Moved.

src/mongo_client.ts
```diff
@@ -448,6 +452,31 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+    // to even attempt to contact a host that is not allowed. So this is placed
+    // here directly after SRV resolution and before the creation of the
+    // topology to short circuit as quickly as possible.
+    if (options.credentials?.mechanism === AuthMechanism.MONGODB_OIDC) {
+      const allowedHosts =
+        options.credentials?.mechanismProperties?.ALLOWED_HOSTS || DEFAULT_ALLOWED_HOSTS;
+      const isServiceAuth = !!options.credentials?.mechanismProperties?.PROVIDER_NAME;
+      if (!isServiceAuth) {
+        for (const host of options.hosts) {
+          if (!hostMatchesWildcards(host.host || host.socketPath || 'localhost', allowedHosts)) {
```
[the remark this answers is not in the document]
durran: Updated to suggestion.

test/unit/cmap/auth/mongodb_oidc/callback_lock_cache.test.ts
durran: `sleep` didn't exist so I created it in utils, and changed all these to use it.

test/unit/cmap/auth/mongodb_oidc/callback_lock_cache.test.ts
durran: `getCallbacks` doesn't return a promise (it's sync) so I added a fail assertion after the call in the try block.

test/unit/cmap/auth/mongodb_oidc/callback_lock_cache.test.ts
durran: Changed all these to use `lengthOf` assertions.

test/manual/mongodb_oidc.prose.test.ts
durran: We throw for invalid options in our driver - so this prose test is skipped and will never be unskipped. I can remove the test entirely but wanted it to be shown that the test itself was not missed.

test/manual/mongodb_oidc.prose.test.ts
durran: Fixed.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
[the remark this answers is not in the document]
nbbeeken: Should we assign `''` to the username in the MongoCredentials constructor so its still always a string?

This is changing public API if not, but perhaps its not possible / would impact other auth mechs?

src/cmap/auth/mongodb_oidc/token_entry_cache.ts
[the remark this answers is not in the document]
nbbeeken: Nice, I like it

test/manual/mongodb_oidc.prose.test.ts
nbbeeken: Yea that makes sense to me, ty for clarifying

nbbeeken (changes requested):
just username and sleep outstanding comments

test/unit/cmap/auth/mongodb_oidc/callback_lock_cache.test.ts
durran: Oh I missed that, thanks. I've updated to use the test tools sleep now.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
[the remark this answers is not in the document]
durran: Good call. I've changed back `username` to be required and defaulted it to an empty string in the constructor.

durran:
Patch on latest commit without all the other clutter: https://spruce.mongodb.com/version/6452c353a4cf47cc42d9b7f5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

nbbeeken (approved):

PR #3641, opened 2023-04-13
chore(NODE-5133): remove broken filter as binary from find operation

baileympearson:
#### What is changing?

Logic to support `Binary` objects as filter predicates has been removed from the `find` operation.

No.

### DIFF
--- a/src/operations/find.ts
+++ b/src/operations/find.ts
@@ -93,16 +93,6 @@ export class FindOperation extends CommandOperation<Document> {
       throw new MongoInvalidArgumentError('Query filter must be a plain object or ObjectId');
     }
 
-    // If the filter is a buffer, validate that is a valid BSON document
-    if (Buffer.isBuffer(filter)) {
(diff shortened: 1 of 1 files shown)


src/operations/find.ts
```diff
@@ -93,16 +93,6 @@ export class FindOperation extends CommandOperation<Document> {
       throw new MongoInvalidArgumentError('Query filter must be a plain object or ObjectId');
```
durran, line 93: Should we also be throwing here if we passed a buffer or do we just rely on the server error?

baileympearson: I think we should rely on the server error.  What do you think?

If we do throw for binary specifically, we should add guards for each BSON type because our other BSON classes will pass this check: `typeof filter !== 'object' || Array.isArray(filter)`.

nbbeeken: I would agree with relying on the server error, filters are bson documents -> bson documents to _us_ are plain JS objects / Maps, passing in a Binary is breaking that API contract

nbbeeken (approved):

durran (approved):

PR #3642, opened 2023-04-14
feat(NODE-5203): emit `INIT` from `ChangeStream`

alecgibson:
#### What is changing?

Emit `INIT` event from `ChangeStream`

No, the docs [already list this as a valid event](https://mongodb.github.io/node-mongodb-native/5.2/classes/ChangeStream.html#INIT), even though it's not currently emitted.

We currently have a test that runs against MongoDB and looks like this:

```ts
it('subscribes to a change stream', async () => {
  const mongo = await new MongoClient(...).connect();
  const changeStream = mongo.collection(...).watch(...);
 
  mongo.collection(...).insertOne({...});
  await new Promise((resolve) => changeStream.once('change', resolve));
})
```
Most of the time (~99%), this test passes fine, even with the synchronous calls of `ChangeStream.watch()` followed immediately by `Collection.insertOne()`

However, this occasionally fails. I've done some digging and the cases where it fails are when the change stream's `cursor` is initialised later than the insertion completes (which is understandable).

What I'd like to do is change the test (and similar code that relies on a stream being initialised) to look like this:

```ts
it('subscribes to a change stream', async () => {
  const mongo = await new MongoClient(...).connect();
  const changeStream = mongo.collection(...).watch(...);
 
  // This line was added and would ensure our stream is subscribed
  // before trying to manipulate the collection:
  await new Promise((resolve) => changeStream.once('init', resolve));
 
  mongo.collection(...).insertOne({...});
  await new Promise((resolve) => changeStream.once('change', resolve));
});
```

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -43,7 +43,7 @@ const CHANGE_DOMAIN_TYPES = {
   CLUSTER: Symbol('Cluster')
 };
 
-const CHANGE_STREAM_EVENTS = [RESUME_TOKEN_CHANGED, END, CLOSE];
+const CHANGE_STREAM_EVENTS = [INIT, RESUME_TOKEN_CHANGED, END, CLOSE];
 
 const NO_RESUME_TOKEN_ERROR =
   'A change stream document has been received that lacks a resume token (_id).';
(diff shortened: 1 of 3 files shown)


alecgibson:
I'm not sure why the linter is failing. It claims:

Screenshot 2023-04-14 at 10 42 21

But as far as I can tell, `init` is already in this event map: https://github.com/mongodb/node-mongodb-native/blob/cc7c75a64720efe40b70f952e77c1d99d269bf06/src/change_stream.ts#L519

W-A-James:
Hi @alecgibson! Thank you for the PR!

Unfortunately we will not be accepting this change as the `INIT` event is a feature of the `Cursor` API which the `ChangeStream` is built upon, but we do not want that event to be a part of the `ChangeStream` API. We will revisit this work in future. You can track our work on this with [this ticket](https://jira.mongodb.org/browse/NODE-5208)

PR #3643, opened 2023-04-14
feat(NODE-5207): deprecate unsupported runCommand options and add spec tests

nbbeeken:
#### What is changing?

DRIVERS-2577
https://github.com/mongodb/specifications/pull/1389

- Adds runCommand spec tests.
- Improves the withTransaction operation runner to catch our undefined return type - which means the transaction did not commit
- Improves the runCommand operation runner to validate arguments
- Deprecates inherited options that are not supported by runCommand

No.


dariakp (dismissed):

dariakp (changes requested):
Need to address failures here and in the specs repo

dariakp (approved):

src/cmap/connection.ts
```diff
@@ -488,18 +488,19 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     options: CommandOptions | undefined,
     callback: Callback
   ): void {
-    const readPreference = getReadPreference(cmd, options);
+    cmd = { ...cmd };
```
durran, line 491: Although this leaves the supplied command untouched outside of the scope of this method can we assign it to a new const to avoid any potential confusion?

durran (changes requested):

dariakp (approved):

PR #3645, opened 2023-04-14
feat(NODE-4720): Add log messages to CMAP spec

W-A-James:
##### CMAP Logging updates
- Added `emitAndLog` method to `TypedEventEmitter`
- Added logging the following events:
  - `ConnectionPoolCreatedEvent`
  - `ConnectionPoolReadyEvent`
  - `ConnectionPoolClearedEvent`
  - `ConnectionPoolClosedEvent`
  - `ConnectionCreatedEvent`
  - `ConnectionReadyEvent`
  - `ConnectionClosedEvent`
  - `ConnectionCheckOutStartedEvent`
  - `ConnectionCheckOutFailedEvent`
  - `ConnectionCheckedOutEvent`
  - `ConnectionCheckedInEvent`

##### Changes to `Topology`
- Changed `Topology` constructor to accept `MongoClient` as first argument to guarantee that it always has a valid reference to the owning client

##### Changes to `Server`
- Moved `pool` and `topology` fields from fields on `Server.s` to top level fields of `Server` and updated all references to reflect this

##### Testing changes
- CMAP logging
	- Pulled in CMAP logging tests and fix from NODE-5212
- Topology
	- Converted `test/integration/node-specific/topology.test.js` to typescript and refactored test to get rid of flakiness introduced by other changes
	- Removed `TestConfiguration.newTopology` as it was unused
	- Updated unit tests making use of old `Topology` constructor to now make use of new `topologyWithPlaceHolderClient` helper function
- Unified Test Runner
	- Implemented close operation for client entities in Unified Test Runner and refactored this functionality to be less nested

No


test/integration/connection-monitoring-and-pooling/connection_monitoring_and_pooling.spec.test.ts
```diff
@@ -52,3 +53,19 @@ describe('Connection Monitoring and Pooling Spec Tests (Integration)', function
...
+
+  runUnifiedSuite(tests, (test: { description: string }) => {
+    if (
+      [
+        'waitQueueMultiple should be included in connection pool created message when specified',
+        'waitQueueSize should be included in connection pool created message when specified'
+      ].includes(test.description)
+    ) {
+      return 'not applicable: waitQueueSize not supported';
+    }
```
W-A-James, line 68: I am open to filing a new ticket to skip the `'Connection checkout fails due to error establishing connection'` spec test since this seems to be failing due to an existing event-ordering issue in the driver and not the logger itself.

test/tools/unified-spec-runner/entities.ts
```diff
@@ -171,7 +171,7 @@ export class UnifiedMongoClient extends MongoClient {
         const transformedLog = {
           level: log.s,
           component: log.c,
-          data: { ...log, s: undefined, c: undefined }
```
W-A-James, line 174: This is cleaner

test/tools/unified-spec-runner/entities.ts
```diff
@@ -171,7 +171,7 @@ export class UnifiedMongoClient extends MongoClient {
...
+          data: { ...log }
```
W-A-James, line 174: This is cleanup

test/tools/unified-spec-runner/operations.ts
```diff
@@ -206,8 +206,13 @@ operations.set('close', async ({ entities, operation }) => {
...
     await cursor.close();
   } catch (e) {
-    const changeStream = entities.getEntity('stream', operation.object);
-    await changeStream.close();
+    try {
+      const changeStream = entities.getEntity('stream', operation.object);
+      await changeStream.close();
+    } catch (e) {
+      const client = entities.getEntity('client', operation.object);
+      await client.close();
```
baileympearson, line 214:
```typescript
const entity = ['cursor', 'stream', 'client']
  .map(e => entities.get(e, operation.object, false))
  .find(e => e != null);

if (!entity) {
	throw new Error("attempted to close non-existant entity - good error message");
}
await entity.close();
```

completely optional suggestion - since we know it'll be exactly one type

src/sdam/server.ts
```diff
@@ -148,9 +148,10 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       options,
       state: STATE_CLOSED,
       topology,
-      pool: new ConnectionPool(this, poolOptions),
       operationCount: 0
-    };
+    } as any;
+    // NOTE: this is done so that the connection pool can access the topology in its constructor
```
baileympearson, line 153: I don't see anything in the constructor of the connection pool that would require the topology present on `server.s`.

How come this change is necessary?

test/tools/utils.ts
```diff
@@ -472,3 +472,27 @@ export const sorted = <T>(iterable: Iterable<T>, how: (a: T, b: T) => 0 | 1 | -1
   items.sort(how);
   return items;
 };
+
+/**
+ * This helper method is necessary as the changes in NODE-4720 required the Topology constructor to
+ * accept a client with access to a logger. This helper serves to keep old tests working with minimal
+ * changes*/
+export function topologyWithPlaceholderClient(
```
baileympearson, line 480: How come we decided to stub out a client instead of just creating a client inline here?

baileympearson (changes requested):

src/sdam/server.ts
W-A-James: The `ConnectionPool's` `client` getter accesses the `topology` field of `server.s` and since we have to make a call to the logger once the `ConnectionPool`'s constructor exits (in the `process.nextTick` callback), we were getting a `TypeError` for accessing the `client` field on the undefined `topology`

test/tools/utils.ts
W-A-James: This is only ever used in unit tests, so constructing a client felt like overkill.

test/tools/unified-spec-runner/operations.ts
W-A-James: This refactor would necessitate a refactor of the `getEntity` method since it throws whenever the type and object being retrieved do not match. That refactor seems out of scope for this ticket since I am unsure of the implications of changing that behaviour. Note that when a boolean is passed in as the third argument to `getEntity`, it modifies whether or not the runner throws when the entity requested does not exist. It does not, however, change whether or not it throws on type mismatch.

test/tools/runner/config.ts
```diff
@@ -250,7 +249,7 @@ export class TestConfiguration {
     options = Object.assign({}, options);
     const hosts =
       host == null ? [].concat(this.options.hostAddresses) : [new HostAddress(`${host}:${port}`)];
-    return new Topology(hosts, options);
+    return topologyWithPlaceholderClient(hosts, options);
```
nbbeeken, line 252: `newTopology` is used in one file: topology.test.js, can we delete this function and change that test to use `newClient`. TestConfiguration is for integration tests, so they'll almost always want a real client.

test/unit/sdam/server.test.ts
```diff
@@ -59,7 +58,11 @@ describe('Server', () => {
   describe('#handleError', () => {
     let server: Server, connection: Connection | undefined;
     beforeEach(() => {
-      server = new Server(new Topology([], {} as any), new ServerDescription('a:1'), {} as any);
+      server = new Server(
+        topologyWithPlaceholderClient([], {} as any),
```
nbbeeken, line 62: I'll just comment once but apply to all usages, can we make the type on the helper one where we don't need to cast to `as any`. maybe: `Partial<TopologyOptions>`?

test/tools/utils.ts
nbbeeken: I think we would wind up with less code, is there an issue with: `new MongoClient('mongodb://iLoveJavaScript')` as in will it fail for some reason? I am assuming we wouldn't reach connect logic in unit tests

test/tools/unified-spec-runner/operations.ts
nbbeeken: I agree, we don't want to make too many changes to shared parts of the test runner. Perhaps we can take this opportunity to at least remove the nesting:

```ts
  try {
    const cursor = entities.getEntity('cursor', operation.object);
    await cursor.close();
    return;
  } catch {
    // ignore
  }

  try {
    const changeStream = entities.getEntity('stream', operation.object);
    await changeStream.close();
    return;
  } catch {
    // ignore
  }

  try {
    const client = entities.getEntity('client', operation.object);
    await client.close();
    return;
  } catch {
    // ignore
  }
```

But as Bailey said, optional suggestion

test/integration/connection-monitoring-and-pooling/connection_monitoring_and_pooling.spec.test.ts
```diff
@@ -52,3 +53,22 @@ describe('Connection Monitoring and Pooling Spec Tests (Integration)', function
     )
   });
 });
+
+describe('Connection Monitoring and Pooling Spec Tests (Integration) - logging', function () {
+  const tests = loadSpecTests('connection-monitoring-and-pooling', 'logging');
+
+  runUnifiedSuite(tests, (test: { description: string }) => {
```
nbbeeken, line 60:
```suggestion
  runUnifiedSuite(tests, test => {
```

No need to annotate something that has type info already

src/mongo_client.ts
```diff
@@ -448,11 +448,10 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
       }
     }
 
-    const topology = new Topology(options.hosts, options);
+    const topology = new Topology(this, options.hosts, options);
     // Events can be emitted before initialization is complete so we have to
     // save the reference to the topology on the client ASAP if the event handlers need to access it
     this.topology = topology;
```
nbbeeken, line 454:
```suggestion
    this.topology = new Topology(this, options.hosts, options);
```

And update the following code, we don't need a comment that refers to this as a special case anymore I think. Or maybe keep the comment but still immediately assign to the MongoClient.

src/sdam/server.ts
nbbeeken: Can we consider a restructure? I know we have a ticket somewhere but we've wanted to migrate away from the `s` subobject pattern where we can, it doesn't play well with TS or I guess runtime contracts. The Server class should be able to advertise an always assigned Topology property as soon as it exists. I'm thinking we store it as a top-level class property

src/cmap/connection_pool.ts
```diff
@@ -872,17 +886,19 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
-              // TODO(NODE-5192): Remove this cast
-              new ConnectionCheckOutFailedEvent(this, 'connectionError', err as MongoError)
+            // TODO(NODE-5192): Remove this cast
+            const connectionCheckOutFailedEvent = new ConnectionCheckOutFailedEvent(
+              this,
+              'connectionError',
+              err as MongoError
             );
+            this.emit(ConnectionPool.CONNECTION_CHECK_OUT_FAILED, connectionCheckOutFailedEvent);
+            this.client.mongoLogger.debug('connection', connectionCheckOutFailedEvent);
```
nbbeeken, line 896: Is the spec specific about whether to log or give the user their event first?

src/cmap/connection_pool.ts
```diff
@@ -872,17 +886,19 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
           } else if (connection) {
             this[kCheckedOut].add(connection);
-            this.emit(
-              ConnectionPool.CONNECTION_CHECKED_OUT,
-              new ConnectionCheckedOutEvent(this, connection)
-            );
+            const connectionCheckedOutEvent = new ConnectionCheckedOutEvent(this, connection);
+            this.emit(ConnectionPool.CONNECTION_CHECKED_OUT, connectionCheckedOutEvent);
```
nbbeeken, line 900: Intead of editing each emit line would we want to consider an override? Perhaps there are events where we don't log? Just an idea.

```
  override emit<EventKey extends keyof ConnectionPoolEvents>(
    event: symbol | EventKey,
    ...args: Parameters<ConnectionPoolEvents[EventKey]>
  ): boolean {
    const res = super.emit(event, ...args);
    this.client.mongoLogger.debug('connection', args[0]);
    return res;
  }
```

nbbeeken (changes requested):

src/sdam/server.ts
W-A-James: Should we also do the same with the connection pool?

src/cmap/connection_pool.ts
W-A-James: I would imagine that if we were to introduce any new events to the CMAP spec, we would also want those to be logged as well.

The only issue I have with this is that it doesn't allow for logging at a different severity level, which I think may be useful. In which case, I think the first argument should be the severity.

src/cmap/connection_pool.ts
```diff
@@ -322,6 +324,15 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     return this[kCheckedOut];
   }
 
+  get client() {
+    // NOTE: This is here to get tests to pass when using MockServer. Looking into a way to modify the testing
+    // infrastructure so it is unnecessary.
+    if (this[kServer].s?.topology) {
+      return this[kServer].s.topology.client;
+    }
```
W-A-James, line 332: Note that this allows for testing to pass. Looking for where exactly it is failing so I can fix it there rather than in the production code

src/mongo_client.ts
nbbeeken: @W-A-James bump, the comment went away but not the code structure: `this.topology = new Topology`

src/sdam/server.ts
```diff
@@ -87,10 +87,8 @@ export interface ServerPrivate {
   options: ServerOptions;
   /** The current state of the Server */
   state: string;
-  /** The topology this server is a part of */
-  topology: Topology;
-  /** A connection pool for this server */
-  pool: ConnectionPool;
+  topology?: Topology;
+  pool?: ConnectionPool;
```
nbbeeken, line 91: These should just be gone now right?

test/unit/cmap/connection_pool.test.js
```diff
@@ -9,14 +9,32 @@ const { setImmediate } = require('timers');
...
+    mock.createServer().then(s => {
+      stubServer = {
+        topology: {
+          client: {
+            mongoLogger: {
+              debug: () => null
+            }
+          }
+        }
+      };
```
nbbeeken, line 28: This is fixed, can we just asign it to stubServer and make the variable const

src/cmap/connection_pool.ts
nbbeeken: From slack, lets gather emitting and logging into a shared method, it can always use debug, and be defined on TypedEE, internally

src/mongo_client.ts
```diff
@@ -448,23 +448,19 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+      this.topology.on(event, (...args: any[]) => this.emit(event, ...(args as any)));
     }
 
     const topologyConnect = async () => {
       try {
-        await promisify(callback => topology.connect(options, callback))();
+        await promisify(callback => this.topology!.connect(options, callback))();
       } catch (error) {
-        topology.close({ force: true });
+        this.topology!.close({ force: true });
```
W-A-James, line 463: @nbbeeken Is it appropriate to have the non-null assertion here? i.e., am I clear to ignore this lint failure or should I add the assertion to the field declaration itself instead?

nbbeeken (changes requested):
LGTM, just need small CI fixes

nbbeeken (dismissed):

dariakp (changes requested):

src/mongo_types.ts
```diff
@@ -397,8 +398,21 @@ export declare interface TypedEventEmitter<Events extends EventsDescription> ext
...
+  protected mongoLogger?: MongoLogger;
+  /** @internal */
+  protected component?: MongoLoggableComponent;
+  /** @internal */
+  protected emitAndLog<EventKey extends keyof Events>(
+    event: EventKey | symbol,
+    ...args: Parameters<Events[EventKey]>
+  ): void {
+    this.emit(event, ...args);
+    if (this.component) this.mongoLogger?.debug(this.component, args[0]);
```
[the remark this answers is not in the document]
W-A-James: At the moment, all of the spec-required logs are at the debug level.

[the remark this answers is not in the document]
nbbeeken:
```
  /** @internal */
  protected emitAndLog
```
The internal / protected modifier should afford us a lot of freedom for changes, but likely a new optional arg would get us what we need if we did need it.

nbbeeken (approved):

PR #3655, opened 2023-04-25
feat(NODE-5019): add runCursorCommand API

nbbeeken:
#### What is changing?

- Adds a new Db API: `await db.runCursorCommand(command, options)`
- A new `RunCommandCursor` subclass of AbstractCursor
  - The subclass overrides methods that do not apply to this kind of cursor throwing errors
  - Adds getMore configurations as required by the spec
  - overrides getMore execution in order to pass through the configurations correctly

At least API docs for the new command, possibly new docs site page.

DRIVERS-2533: https://github.com/mongodb/specifications/pull/1412

- [/] New TODOs have a related JIRA ticket


dariakp (changes requested):
Partial feedback, need to circle back to finish review

test/tools/unified-spec-runner/operations.ts
```diff
@@ -741,9 +790,10 @@ export async function executeOperationAndCheck(
   const opFunc = operations.get(operation.name);
   expect(opFunc, `Unknown operation: ${operation.name}`).to.exist;
 
-  if (operation.arguments?.session) {
-    const session = entities.getEntity('session', operation.arguments.session, false);
-    operation.arguments.session = session;
+  if (typeof operation.arguments?.session === 'string') {
+    // Cannot do this b/c operations pass through unsanitized options to command construction:
```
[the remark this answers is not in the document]
nbbeeken: for debugging, since we get the session here I was not able to see what the session's name was inside the operation function, so I was attaching the name as an underscore prop just to see the name. Turns out that will make some operations fail b/c it will take the options and directly map them to the command construction. Just sharing context, removed the comment.

test/tools/unified-spec-runner/operations.ts
```diff
@@ -659,6 +672,42 @@ operations.set('runCommand', async ({ entities, operation }: OperationFunctionPa
...
+  if (opts.comment !== undefined) cursor.setComment(opts.comment);
+
+  return cursor.toArray();
+});
+
+operations.set('createCommandCursor', async ({ entities, operation }: OperationFunctionParams) => {
+  const collection = entities.getEntity('db', operation.object);
+  const { command, ...opts } = operation.arguments!;
+  const cursor = collection.runCursorCommand(command, {
+    readPreference: ReadPreference.fromOptions(opts),
```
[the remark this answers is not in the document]
nbbeeken: Indeed, extracted

test/spec/run-command/runCommand.json
```diff
@@ -123,6 +123,60 @@
         }
       ]
     },
+    {
```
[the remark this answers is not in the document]
nbbeeken: fully sync-ed to latest

test/spec/unified-test-format/invalid/entity-createRunCursorCommand.json
```diff
@@ -0,0 +1,71 @@
+{
```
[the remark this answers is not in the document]
nbbeeken: removed

test/tools/unified-spec-runner/operations.ts
```diff
@@ -741,9 +790,8 @@ export async function executeOperationAndCheck(
   const opFunc = operations.get(operation.name);
   expect(opFunc, `Unknown operation: ${operation.name}`).to.exist;
 
-  if (operation.arguments?.session) {
-    const session = entities.getEntity('session', operation.arguments.session, false);
-    operation.arguments.session = session;
+  if (typeof operation.arguments?.session === 'string') {
```
dariakp, line 793: why did this logic need to change?

test/tools/unified-spec-runner/operations.ts
```diff
@@ -659,6 +672,42 @@ operations.set('runCommand', async ({ entities, operation }: OperationFunctionPa
   return db.command(command, options);
 });
 
+operations.set('runCursorCommand', async ({ entities, operation }: OperationFunctionParams) => {
+  const db = entities.getEntity('db', operation.object);
+  const { command, ...opts } = operation.arguments!;
+  const cursor = db.runCursorCommand(command, {
+    readPreference: ReadPreference.fromOptions(opts),
```
dariakp, line 679: this `ReadPreference` construction should follow the same pattern as in `createCommandCursor`

etc/sync-wip-spec.sh
```diff
@@ -0,0 +1,25 @@
+#! /usr/bin/env bash
```
dariakp, line 1: did you mean to add this file?

src/db.ts
```diff
@@ -523,6 +524,19 @@ export class Db {
 
     return new ChangeStream<TSchema, TChange>(this, pipeline, resolveOptions(this, options));
   }
+
+  /**
+   * A low level cursor API providing basic driver functionality.
```
dariakp, line 529:
```suggestion
   * A low level cursor API providing basic driver functionality:
```

test/integration/run-command/run_command.spec.test.ts
```diff
@@ -2,5 +2,13 @@ import { loadSpecTests } from '../../spec';
 import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
 
 describe('RunCommand spec', () => {
-  runUnifiedSuite(loadSpecTests('run-command'));
+  runUnifiedSuite(loadSpecTests('run-command'), test => {
+    if (test.description.includes('timeoutMS') || test.description.includes('timeoutMode')) {
```
dariakp, line 6: Are there still some tests left for this in run-command? I thought they were all moved to the CSOT spec folder. If there are some of these tests left, we should reference a TODO(NODE-xxxx) here where we would do the part of CSOT implementation that would unskip these.

For the tests that are in the CSOT spec folder, we should still sync these new tests now, and create the placeholder to run & skip CSOT spec tests with the same `TODO` reference because otherwise we might miss them when going to implement CSOT (we could add a callout to the corresponding JIRA to make sure to include them, but I think just syncing them is more straightforward and less hoops to jump).

test/integration/run-command/run_cursor_command.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    commandsStarted = [];
+    client.on('commandStarted', started => commandsStarted.push(started));
+  });
+
+  afterEach(async function () {
+    commandsStarted = [];
+    await client.close();
+  });
+
+  it('should only run init command once', async () => {
```
dariakp, line 26: I don't really understand the structure of this test. Can you break it apart and/or clarify the test case title so that it's clear what conditions you are testing? Feel free to ping me on slack to discuss this further.

test/integration/run-command/run_cursor_command.test.ts
```diff
@@ -0,0 +1,53 @@
+import { expect } from 'chai';
+
+import { CommandStartedEvent, Db, MongoClient } from '../../mongodb';
+
+describe('class RunCommandCursor', () => {
```
dariakp, line 5: This is supposed to be an integration test, so you should be describing an integration-level API. The file name is `run_cursor_command`, and what you are calling in the test case is `runCursorCommand`, so I'm guessing the title here should be something along the lines of `Db.runCursorCommand()` or `runCursorCommand API`

test/unit/cursor/run_command_cursor.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  context('setMaxTimeMS()', () => {
+    it('stores the maxTimeMS value in getMoreOptions.maxAwaitTimeMS', () => {
+      const cursor = client.db().runCursorCommand({ a: 1 });
+      cursor.setMaxTimeMS(2);
+      expect(cursor).to.have.nested.property('getMoreOptions.maxAwaitTimeMS', 2);
+    });
+
+    it('does not throw on incorrect maxTimeMS type', () => {
```
dariakp, line 38: suggestion: -> "does not validate maxTimeMS type"

why is this the desired/required behavior?

test/unit/cursor/run_command_cursor.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  context('setBatchSize()', () => {
+    it('stores the batchSize value in getMoreOptions', () => {
+      const cursor = client.db().runCursorCommand({ a: 1 });
+      cursor.setBatchSize(2);
+      expect(cursor).to.have.nested.property('getMoreOptions.batchSize', 2);
+    });
+
+    it('does not throw on incorrect batchSize type', () => {
```
dariakp, line 53: same suggestion, -> "does not validate ..."
and same question as maxTimeMS

test/unit/cursor/run_command_cursor.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      const cursor = client.db().runCursorCommand({ a: 1 });
+      // @ts-expect-error: testing for incorrect type
+      cursor.setBatchSize('abc');
+      expect(cursor).to.have.nested.property('getMoreOptions.batchSize', 'abc');
+    });
+  });
+
+  context('clone()', () => {
+    it('returns a new RunCursorCommand instance', () => {
```
dariakp, line 62: should there be some more tests here about the effect of cloning a cursor with set comment, maxTimeMS, and batchSize?

src/cursor/run_command_cursor.ts
```diff
@@ -0,0 +1,138 @@
...
+    );
+  }
+
+  public override addCursorFlag(_: string, __: boolean): never {
+    throw new MongoAPIError(
+      'RunCommandCursor does not support cursor flags, they must be attached to the command being run'
+    );
+  }
+
+  public override maxTimeMS(_: number): never {
```
dariakp, line 80: this error is confusing because there is a public `setMaxTimeMS` method, same regarding batchSize and `setBatchSize` below

dariakp (changes requested):

etc/sync-wip-spec.sh
nbbeeken: I have a note in the PR description, I intend to remove it when the spec merges. We can keep this thread to make sure it does get removed

src/cursor/run_command_cursor.ts
nbbeeken: added documentation and improved error messages

src/db.ts
nbbeeken: done

test/integration/run-command/run_command.spec.test.ts
nbbeeken: Should we sync all of CSOT now? I'm just thinking we're going to need to sync all of CSOT, and the state the tests are in now may not be what they are when it is time to do that, so having a snapshot or portion of the tests might be confusing later.

This is going to be the case for all drivers other than go and python that can sync and run the CSOT tests now.

test/integration/run-command/run_cursor_command.test.ts
nbbeeken: fixed, went with "runCursorCommand API"

test/integration/run-command/run_cursor_command.test.ts
nbbeeken: This is general to our abstract cursor, but also it may not be a test we need to keep. taking this offline

test/tools/unified-spec-runner/operations.ts
nbbeeken: fixed

test/tools/unified-spec-runner/operations.ts
nbbeeken: it did not! debugging cruft, fixed

test/unit/cursor/run_command_cursor.test.ts
nbbeeken: updated title

- We already have validation / filtering in command construction, so only typeof number will be sent

Additional hypothetical scenarios:
- If someone put the wrong value the server throws, and as a build your own command API we shouldn't intervene (future fractional batchSize?)
- Any of the BSON numeric types are permitted as long as they are positive and do not have a fractional component, we don't need further validation that works against that reality (typeof number blocks Int32/Long instances)

test/unit/cursor/run_command_cursor.test.ts
nbbeeken: Updated title
same answer in the batchSize comment

test/unit/cursor/run_command_cursor.test.ts
nbbeeken: Good call, currently the getMore options would be lost in the clone. I had considered making the clone API unsupported for this cursor type, the API is not super valuable (you can put your db.runCursorCommand API into a recallable function), and has these edge cases we need to account for. Thoughts?

test/integration/run-command/run_command.spec.test.ts
dariakp: the tests you introduced are in their own files, so I think it's fine not to sync the others; when we go to work on CSOT, there will be tickets referencing specific commits which we can sync and then update any corresponding tests accordingly; as for the other drivers - we should call out in downstream changes or the ticket description that drivers should ensure that they add these new tests now or in their own CSOT follow up tickets

test/unit/cursor/run_command_cursor.test.ts
dariakp: is clone spec for all cursors? if not, I don't see why we need to implement it if no one is asking for it

dariakp (commented):
Reviewed & resolved the addressed comments

test/unit/cursor/run_command_cursor.test.ts
nbbeeken: Not spec, sounds good to me

test/integration/run-command/run_cursor_command.test.ts
nbbeeken: Refactored test to be clear about the interleaving of iterators

etc/sync-wip-spec.sh
nbbeeken: Removed

src/cursor/run_command_cursor.ts
```diff
@@ -0,0 +1,140 @@
...
+import { GetMoreOperation } from '../operations/get_more';
+import { RunCommandOperation } from '../operations/run_command';
+import type { ReadConcernLike } from '../read_concern';
+import type { ReadPreferenceLike } from '../read_preference';
+import type { ClientSession } from '../sessions';
+import { Callback, ns } from '../utils';
+import { AbstractCursor } from './abstract_cursor';
+
+/** @public */
+export type RunCommandCursorOptions = {
```
dariakp, line 14: -> RunCursorCommandOptions to align with spec

dariakp (changes requested):

dariakp (approved):

test/integration/run-command/run_cursor_command.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    await db
+      .collection<{ _id: number }>('collection')
+      .insertMany([{ _id: 0 }, { _id: 1 }, { _id: 2 }]);
+  });
+
+  afterEach(async function () {
+    await client.close();
+  });
+
+  it('returns each document only once across multiple iterators', async () => {
```
baileympearson, line 22: Should this test be on `AbstractCursor` instead of `runCommandCursor`, since the behavior comes from the abstract class?  This test is not specific to our RunCommandCursor.

Why was it added now?

test/tools/unified-spec-runner/operations.ts
```diff
@@ -349,27 +349,40 @@ operations.set('insertMany', async ({ entities, operation }) => {
...
 
-operations.set('iterateUntilDocumentOrError', async ({ entities, operation }) => {
-  function getChangeStream(): UnifiedChangeStream | null {
-    try {
-      const changeStream = entities.getEntity('stream', operation.object);
-      return changeStream;
-    } catch (e) {
-      return null;
-    }
+function getChangeStream({ entities, operation }): UnifiedChangeStream | null {
```
baileympearson, line 352: Could this potentially be a method on the `entities` object?

test/tools/unified-spec-runner/operations.ts
```diff
@@ -349,27 +349,40 @@ operations.set('insertMany', async ({ entities, operation }) => {
...
+operations.set('iterateOnce', async ({ entities, operation }) => {
+  const changeStream = getChangeStream({ entities, operation });
+  if (changeStream == null) {
+    // iterateOnce is used for changes streams and regular cursors.
+    // we have no other way to distinguish which scenario we are testing when we run an
+    // iterateOnce operation, so we first try to get the changeStream and
+    // if that fails, we know we need to get a cursor
+    const cursor = entities.getEntity('cursor', operation.object);
+    return cursor.tryNext();
   }
```
baileympearson, line 383:
```suggestion
  const cursorlike = getChangeStream(..) ?? entities.getEntity('cursor');
  return cursorLike.tryNext();
```

test/tools/unified-spec-runner/operations.ts
```diff
@@ -659,6 +672,42 @@ operations.set('runCommand', async ({ entities, operation }: OperationFunctionPa
...
+    session: opts.session
+  });
+
+  if (!Number.isNaN(+opts.batchSize)) cursor.setBatchSize(+opts.batchSize);
+  if (!Number.isNaN(+opts.maxTimeMS)) cursor.setMaxTimeMS(+opts.maxTimeMS);
+  if (opts.comment !== undefined) cursor.setComment(opts.comment);
+
+  // The spec dictates that we create the cursor and force the find command
+  // to execute, but the first document must still be returned for the first iteration.
+  const result = await cursor.tryNext();
```
baileympearson, line 704: Why not just call `initialize` like our change stream helper does for unified change stream tests?

test/tools/unified-spec-runner/operations.ts
baileympearson: same suggestion above

baileympearson (changes requested):

dariakp (approved):

baileympearson (approved):

PR #3660, opened 2023-05-04
feat(NODE-5004): dont create or drop ecc collections

durran:
### Description

Does not create or drop ecc collections when using QEv2.

#### What is changing?
- No longer creates ecc collection during create collection.
- No longer drops ecc collection during drop.
- Throws an error when server < 7.0 when encrypted fields supplied.
- Syncs to NODE-5195 de-dup FLE tests.

None

NODE-5004/NODE-5195


durran:
Failures in CI unrelated to these changes, CSFLE all passing here: https://spruce.mongodb.com/version/6453b7cf57e85ab511b55d44/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Erhel8-custom-dependency-tests%24

src/operations/create_collection.ts
```diff
@@ -98,6 +100,10 @@ export interface CreateCollectionOptions extends CommandOperationOptions {
   changeStreamPreAndPostImages?: { enabled: boolean };
 }
 
+/* @internal */
+const INVALID_QE_VERSION =
+  'Driver support of Queryable Encryption is incompatible with server. Upgrade server to use Queryable Encryption.';
```
addaleax, line 105: Do we want to mention 7.0 as a version here? That might be good because users may think that QE was introduced in 6.0 (and they wouldn’t be _wrong_ about that, either).

addaleax (approved):

src/operations/create_collection.ts
nbbeeken: I agree with mentioning 7.0 here

src/operations/create_collection.ts
```diff
@@ -127,12 +133,15 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
         db.s.client.options.autoEncryption?.encryptedFieldsMap?.[`${db.databaseName}.${name}`];
 
       if (encryptedFields) {
+        // Creating a QE collection required min server of 7.0.0
+        if (server.description.maxWireVersion < MIN_SUPPORTED_QE_WIRE_VERSION) {
+          throw new MongoCompatibilityError(INVALID_QE_VERSION);
+        }
```
nbbeeken, line 139: Not at all a necessary change, but there is a class: `ServerCapabilities`: https://github.com/mongodb/node-mongodb-native/blob/main/src/sdam/topology.ts#L945-L989 That is _supposed_ to encode what the current wireVersion supports as a boolean method. 

I can find like 20 instances of this class not being used where it should be though, so maybe something we'd want to refactor as a whole later if at all.

nbbeeken (changes requested):
Just the one change about server version, mentioned an improvement if we want that as well

nbbeeken (approved):

src/operations/create_collection.ts
durran: I've updated to mention the server version required in the error now. With respect to `ServerCapabilities` - I think it's better to refactor as a whole later on... It's better existing as a property on `Server` instead of `Topology` (hence the name and we can have mixed topologies) and the command execution at the operation layer of the codebase executes on a server, so it would be more readily accessible and relevant there. I've created https://jira.mongodb.org/browse/NODE-5253 for a spike to create the ticket.

src/operations/create_collection.ts
durran: Updated to mention 7.0.

addaleax (approved):

nbbeeken (approved):

PR #3661, opened 2023-05-04
test(NODE-4772): Test that mongocryptd is not spawned when shared library is loaded

W-A-James:
#### What is changing?

- Adds CSFLE prose test from DRIVERS-2451

No

Ensuring we keep up with spec.


test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1106,6 +1110,75 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
 
     it.skip('Via bypassAutoEncryption', () => {}).skipReason =
       'TODO(NODE-2422): Implement "Bypass spawning mongocryptd" tests';
+
+    describe('via loading shared library', function () {
+      let clientEncrypted;
```
baileympearson, line 1115: We're missing logic to skip this test when the shared library is not present: 

>   IMPORTANT: This test requires the crypt_shared_ library be loaded. If the crypt_shared_ library is
   not available, skip the test.

I think we can use the `CRYPT_SHARED_LIB_PATH` environment variable to determine whether or not we're testing with the shared library.

baileympearson (changes requested):

.evergreen/run-kms-servers.sh
baileympearson, line 1: did you mean to change this file?

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1106,6 +1110,79 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+          .collection(dataCollName)
+          .insertOne({ unencrypted: 'test' });
+        expect(insertResult).to.have.property('insertedId');
+      });
+
+      afterEach(async function () {
+        await clientEncrypted.close();
+        if (client) {
+          await client.close();
+        }
```
baileympearson, line 1171:
```suggestion
        await clientEncrypted?.close();
        await client?.close();
```

baileympearson (changes requested):

.evergreen/run-kms-servers.sh
W-A-James: Yes, this was intentional. It didn't have execute permissions by default and whenever I test locally, I have to give it execute permissions, so I updated that here. I can add it to the "What is changing?" section of the PR description.

baileympearson: Any other reason you can't just `bash .evergreen/run-kms-servers.sh`?  I'm not opposed to adding execute permissions but we don't have them currently because in CI we intentionally run these scripts with `bash` because the default shell in evergreen is different on different hosts.

Screenshot 2023-05-09 at 3 09 15 PM

I was also confused by the changes because I thought Github was displaying the file size, not the permissions

W-A-James: No other reason than convenience.

On further inspection, if we're going to keep the permissions change, we should also add a shebang line to the top of that file so if it is invoked with `./.evergreen/run-kms-servers.sh` it uses bash automatically.

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1106,6 +1110,77 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
...
+
+    describe('via loading shared library', function () {
+      let clientEncrypted;
+      let client;
+      // Setup
+      beforeEach(async function () {
+        const { cryptSharedLibPath } = getEncryptExtraOptions();
+        if (!cryptSharedLibPath) {
+          this.skip('Skipped because CRYPT_SHARED_LIB_PATH is missing');
+        }
```
baileympearson, line 1122:
```suggestion
        if (!cryptSharedLibPath) {
          this.currentTest.skipReason = 'Skipped because CRYPT_SHARED_LIB_PATH is missing';
          this.skip();
        }
```

To attach skip reasons, we assign the reason to the `skipReason` on either `currentTest` or `test` (depending on the place where the test is being skipped from).

PR #3671, opened 2023-05-15
docs(NODE-4778): update tls option notes

durran:
### Description

Updates the TLS comments in `MongoClient` to note the spec compliant options, that others will be deprecated and removed, and fixed the type comments on the existing options.

NODE-4778

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -790,7 +790,7 @@ export interface MongoOptions
    * | see note below        | `tlsInsecure`                                 | N/A                | `boolean`          |
    *
    * If `tlsInsecure` is set to `true`, then it will set the node native options `checkServerIdentity`
-   * to a no-op and `rejectUnauthorized` to `false.
+   * to a no-op and `rejectUnauthorized` to `false`.
    *
    * If `tlsInsecure` is set to `false`, then it will set the node native options `checkServerIdentity`
    * to a no-op and `rejectUnauthorized` to the inverse value of `tlsAllowInvalidCertificates`. If
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -778,17 +778,23 @@ export interface MongoOptions
    *
    * ### Additional options:
    *
-   * | nodejs native option | driver spec compliant option name             | legacy option name | driver option type |
-   * |:---------------------|:----------------------------------------------|:-------------------|:-------------------|
-   * | `ca`                 | `tlsCAFile`                                   | `sslCA`            | `string`           |
-   * | `crl`                | N/A                                           | `sslCRL`           | `string`           |
-   * | `cert`               | `tlsCertificateFile`, `tlsCertificateKeyFile` | `sslCert`          | `string`           |
-   * | `key`                | `tlsCertificateKeyFile`                       | `sslKey`           | `string`           |
-   * | `passphrase`         | `tlsCertificateKeyFilePassword`               | `sslPass`          | `string`           |
-   * | `rejectUnauthorized` | `tlsAllowInvalidCertificates`                 | `sslValidate`      | `boolean`          |
-   * | N/A                  | `tlsAllowInvalidHostnames`                    | N/A                | `boolean`          |
-   * | N/A                  | `tlsInsecure`                                 | N/A                | `boolean`          |
+   * | nodejs native option  | driver spec compliant option name             | legacy option name | driver option type |
+   * |:----------------------|:----------------------------------------------|:-------------------|:-------------------|
+   * | `ca`                  | `tlsCAFile`                                   | `sslCA`            | `string`           |
+   * | `crl`                 | N/A                                           | `sslCRL`           | `string`           |
+   * | `cert`                | `tlsCertificateFile`, `tlsCertificateKeyFile` | `sslCert`          | `string`           |
+   * | `key`                 | `tlsCertificateKeyFile`                       | `sslKey`           | `string`           |
+   * | `passphrase`          | `tlsCertificateKeyFilePassword`               | `sslPass`          | `string`           |
+   * | `rejectUnauthorized`  | `tlsAllowInvalidCertificates`                 | `sslValidate`      | `boolean`          |
+   * | `checkServerIdentity` | `tlsAllowInvalidHostnames`                    | N/A                | `boolean`          |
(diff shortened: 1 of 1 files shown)


src/mongo_client.ts
```diff
@@ -774,14 +774,17 @@ export interface MongoOptions
    *
    * If set TLS enabled, equivalent to setting the ssl option.
    *
+   * Spec complicant options: ssl, tls, tlsCAFile, tlsCertificateKeyFile, tlsCertificateKeyFilePassworda, tlsAllowInvalidCertificates,
```
dariakp, line 777: This comment will be public facing, since it's on the public interface, so we probably want to rephrase it in a way that makes sense to a user reading this (and take care of the typo in the `tlsCertificateKeyFilePassword`). Additionally, `tlsAllowInvalidCertificates`, `tlsAllowInvalidCertificates`, and `tlsInsecure` aren't in the table below, so we should document them. 

It might be most user-friendly to update the table heading from `MongoDB equivalent` to `MongoDB driver spec equivalent` and then add a column for `Legacy option name` (or something along those lines) instead of just listing them in the paragraph above, that way it's easy to see what maps to what.

Lastly, `crl` will no longer have an equivalent config option if we deprecate the sslCRL option - are we ok not letting users set it at all, or do we still need to provide *some* way for users to set it even if the drivers spec doesn't account for it?

dariakp (changes requested):

addaleax (commented):
Are you sure you want to remove support for passing Node.js-specific types (`Buffer | Buffer[] | KeyObject[]`)? IIUC drivers have some liberty to provide options in a more language-specific idiomatic way if you want.

(I do share the concern about `crl`. I don’t know how many people use it but it’s something we expose in the shell.)

dariakp:
> Are you sure you want to remove support for passing Node.js-specific types (Buffer | Buffer[] | KeyObject[])? IIUC drivers have some liberty to provide options in a more language-specific idiomatic way if you want.

@addaleax This is not removing support for buffer, it's just updating the documentation to match reality. The buffer way of passing these into the client is not currently supported either in the types (which only list `string`) or in the actual parsing logic, which assumes it's a string pointing to the location of the files `fs.readFileSync(String(value), { encoding: 'ascii' })`. Moreover, NODE-3924 is going to further update this logic to make the certificate file reading happen async.

src/mongo_client.ts
durran: I've update the table per the suggestion. As for `crl`/`sslCRL` - @addaleax @dariakp  if shell feels that we do still need support for it, even though it's not in the spec I think it's fine to not deprecate it. Maybe we can update the deprecation ticket to also add a more "spec compliant looking" name like `tlsCRL` so they all have the same naming convention?

addaleax:
@dariakp Mh yeah, I guess the table is a bit confusing in that the Node.js options don’t match 1:1 in behavior to the MongoDB options?

For example, we do rely on the fact that we can pass a string _with the CA contents_ as the `ca` option (not `tlsCAFile`).

src/mongo_client.ts
```diff
@@ -774,18 +774,20 @@ export interface MongoOptions
...
+   * | nodejs native option | driver spec compliant option name             | legacy option name | driver option type |
+   * |:---------------------|:----------------------------------------------|:-------------------|:-------------------|
+   * | `ca`                 | `tlsCAFile`                                   | `sslCA`            | `string`           |
+   * | `crl`                | N/A                                           | `sslCRL`           | `string`           |
+   * | `cert`               | `tlsCertificateFile`, `tlsCertificateKeyFile` | `sslCert`          | `string`           |
+   * | `key`                | `tlsCertificateKeyFile`                       | `sslKey`           | `string`           |
+   * | `passphrase`         | `tlsCertificateKeyFilePassword`               | `sslPass`          | `string`           |
+   * | `rejectUnauthorized` | `tlsAllowInvalidCertificates`                 | `sslValidate`      | `boolean`          |
+   * | N/A                  | `tlsAllowInvalidHostnames`                    | N/A                | `boolean`          |
```
dariakp, line 789: this controls the `checkServerIdentity` nodejs option

src/mongo_client.ts
```diff
@@ -774,18 +774,20 @@ export interface MongoOptions
...
+   * | N/A                  | `tlsInsecure`                                 | N/A                | `boolean`          |
```
dariakp, line 790: this will modify `rejectUnauthorized` and `checkServerIdentity` depending on what other options are set, though I think this is fine to omit from the table, N/A is less confusing than trying to record the interaction

src/mongo_client.ts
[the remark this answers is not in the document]
durran: Typo fixed in table.

src/mongo_client.ts
durran: Updated.

src/mongo_client.ts
durran: I've changed from N/A to "see note below", and provided details below the table on the behaviour.

src/mongo_client.ts
```diff
@@ -772,19 +772,29 @@ export interface MongoOptions
...
+   * | `crl`                 | N/A                                           | `sslCRL`           | `string`           |
+   * | `cert`                | `tlsCertificateFile`, `tlsCertificateKeyFile` | `sslCert`          | `string`           |
+   * | `key`                 | `tlsCertificateKeyFile`                       | `sslKey`           | `string`           |
+   * | `passphrase`          | `tlsCertificateKeyFilePassword`               | `sslPass`          | `string`           |
+   * | `rejectUnauthorized`  | `tlsAllowInvalidCertificates`                 | `sslValidate`      | `boolean`          |
+   * | `checkServerIdentity` | `tlsAllowInvalidHostnames`                    | N/A                | `boolean`          |
+   * | see note below        | `tlsInsecure`                                 | N/A                | `boolean`          |
+   *
+   * If `tlsInsecure` is set to `true`, then it will set the node native options `checkServerIdentity`
+   * to a no-op and `rejectUnauthorized` to `false.
```
dariakp, line 793:
```suggestion
   * to a no-op and `rejectUnauthorized` to `false`.
```

dariakp (changes requested):

dariakp (approved):

dariakp (approved):

PR #3672, opened 2023-05-17
feat(NODE-5314): add search index helpers

baileympearson:
### Description

POC for https://github.com/mongodb/specifications/pull/1423.

#### What is changing?

- 5 new operations are added to the unified test runner
- new collection helpers are added to enable search index management

No.


.evergreen/config.in.yml
```diff
@@ -45,6 +45,15 @@ functions:
       params:
         file: src/expansion.yml
 
+  "run search index management tests":
```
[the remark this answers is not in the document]
baileympearson: I was originally thinking that these tests would eventually be a special environment test and run against an Atlas cluster so I was just setting it up for the eventual work.

src/utils.ts
```diff
@@ -337,6 +334,19 @@ export class MongoDBNamespace {
...
 }
 
+/**
+ * @public
+ *
+ * A class representing a collection's namespace.  This class enforces (through Typescript) that
+ * the `collection` portion of the namespace is defined and should only be
+ * used in scenarios where this can be guaranteed.
+ */
+export class MongoDBCollectionNamespace extends MongoDBNamespace {
```
baileympearson, line 344: This change is not necessary but without it, the operations added in this PR must either assert on the namespace's collection string being truthy or provide a default value, but in all cases the namespace has a collection defined (they're all operations defined on the collection).

package.json
```diff
@@ -120,7 +120,8 @@
     "check:tsd": "tsd --version && tsd",
     "check:dependencies": "mocha test/action/dependency.test.ts",
     "check:dts": "node ./node_modules/typescript/bin/tsc --noEmit mongodb.d.ts && tsd",
-    "check:test": "mocha --config test/mocha_mongodb.json test/integration",
+    "check:search-indexes": "mocha --config test/mocha_mongodb.json test/integration/index-management",
```
durran, line 123: I think since it will have its separate setup in the future for the Atlas cluster, it probably makes more sense to put this in /test/manual so we don't potentially set up a pattern of individually excluding integration tests in the check:test script.

durran (changes requested):

src/collection.ts
```diff
@@ -975,12 +998,45 @@ export class Collection<TSchema extends Document = Document> {
...
+    descriptions: ReadonlyArray<SearchIndexDescription>
+  ): Promise<string[]> {
+    return executeOperation(this.client, new CreateSearchIndexesOperation(this, descriptions));
+  }
+
+  async dropSearchIndex(name: string): Promise<void> {
+    return executeOperation(this.client, new DropSearchIndexOperation(this, name));
+  }
+
+  async updateSearchIndex(name: string, definition: Document): Promise<void> {
```
lerouxb, line 1039: Sorry to jump in here, but wasn't there also going to be a `get` operation to just get one by name?

lerouxb: I saw someone mention it in passing in slack _somewhere_, but now can't find it and I'm kinda wondering if I'm imagining it. If it isn't listed somewhere, then it probably isn't a thing.

baileympearson: Are you thinking of the optional overload for `create`?  I chose not to implement this in Node to align with the existing `createIndex` API in the driver

```
createSearchIndex(name: String, definition: Document)
```

lerouxb: I checked again and I'm now pretty sure I was imagining it.

durran (changes requested):

src/operations/search_indexes/drop.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    const command: Document = {
+      dropSearchIndex: namespace.collection
+    };
+
+    if (typeof this.name === 'string') {
+      command.name = this.name;
+    }
+
+    server.command(namespace, command, { session }, err => {
```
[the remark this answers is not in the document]
baileympearson: I thought about it, but if it's alright with you, we can hold off.  This will go away once we're using async await in favor of:
```typescript
await server.command(namespace, command, { session });
```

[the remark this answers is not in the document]
nbbeeken: We have added `connection.commandAsync` we can do the same for server, just a suggestion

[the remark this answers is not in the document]
baileympearson: I thought about it - but it doesn't make much of a difference unless `operation.execute` is also async because then we'd only be then chaining, instead of callback handling.

However, we can't easily do the same for operation.execute because we already have it - but the asyncified form calls the`operation.execute` callback form, so it needs to be defined in the callback overload.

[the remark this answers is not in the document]
baileympearson: This is why I asked about refactoring `operation.execute` to be async throughout the driver.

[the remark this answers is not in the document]
nbbeeken: Yea I agree, it is not much difference, you'd end up with `commandAsync().then` to invoke the callback. Using the xAsync alternatives makes it a bit easier for someone with no context to convert to `await` later, but all these operations are fairly thin so it's not a big deal

lerouxb (approved):

durran (approved):

package.json
```diff
@@ -120,6 +120,7 @@
     "check:tsd": "tsd --version && tsd",
     "check:dependencies": "mocha test/action/dependency.test.ts",
     "check:dts": "node ./node_modules/typescript/bin/tsc --noEmit mongodb.d.ts && tsd",
+    "check:search-indexes": "mocha --config test/mocha_mongodb.json test/manual/search-index-management.spec.test.ts",
```
nbbeeken, line 123: Should we have `nyc` here?

src/collection.ts
```diff
@@ -975,12 +998,95 @@ export class Collection<TSchema extends Document = Document> {
...
+   * @param options - The options for the list indexes operation.
+   *
+   * @remarks Only available when used against a 7.0+ Atlas cluster.
+   */
+  listSearchIndexes(options?: ListSearchIndexesOptions): ListSearchIndexesCursor;
+  listSearchIndexes(indexName: string, options?: ListSearchIndexesOptions): ListSearchIndexesCursor;
+  listSearchIndexes(
+    indexNameOrOptions?: string | ListSearchIndexesOptions,
+    options?: ListSearchIndexesOptions
+  ): ListSearchIndexesCursor {
```
nbbeeken, line 1021: Since the return value is the same for all these methods so we should just have the union type argument and optional options

src/collection.ts
```diff
@@ -975,12 +998,95 @@ export class Collection<TSchema extends Document = Document> {
...
+   * Creates a single search index for the collection.
+   *
+   * @param description - The index description for the new search index.
+   * @returns A promise that resolves to the name of the new search index.
+   *
+   * @remarks Only available when used against a 7.0+ Atlas cluster.
+   */
+  async createSearchIndex(description: SearchIndexDescription): Promise<string> {
+    const indexes = await this.createSearchIndexes([description]);
+    return indexes[0];
```
nbbeeken, line 1046:
```suggestion
    const [index] = await this.createSearchIndexes([description]);
    return index;
```

src/operations/search_indexes/drop.ts
```diff
@@ -0,0 +1,35 @@
+import type { Document } from 'bson';
+
+import type { Collection } from '../../collection';
+import type { Server } from '../../sdam/server';
+import type { ClientSession } from '../../sessions';
+import type { Callback } from '../../utils';
+import { AbstractOperation } from '../operation';
+
+export class DropSearchIndexOperation extends AbstractOperation<void> {
+  /** @internal */
```
nbbeeken, line 10: Did you mean to make the class internal?

src/collection.ts
baileympearson: That's not quite the same.  Collapsing them into a single overload implies that `listSearchIndexes(AggregateOptions, AggregateOptions)` is a valid signature.  Overloading the method clarifies which combinations of arguments are valid.

package.json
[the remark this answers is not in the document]
baileympearson: I'd prefer to leave them here.  The currently run against regular configurations but soon they'll be running in a separate variant that tests against Atlas and they'll never run in regular CI runs.

r.e. nyc - we don't run `nyc` for special environment tests.  I'd follow the existing pattern of not invoking it for non-integration suite tests.

src/collection.ts
```diff
@@ -975,12 +998,95 @@ export class Collection<TSchema extends Document = Document> {
...
+   * Creates multiple search indexes for the current collection.
+   *
+   * @param descriptions - An array of `SearchIndexDescription`s for the new search indexes.
+   * @returns A promise that resolves to an array of the newly created search index names.
+   *
+   * @remarks Only available when used against a 7.0+ Atlas cluster.
+   * @returns
+   */
+  async createSearchIndexes(
+    descriptions: ReadonlyArray<SearchIndexDescription>
```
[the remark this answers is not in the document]
baileympearson: It's readonly internally, so it makes sense to also expose it publicly as "readonly" indicating that the driver never modifies the input.  But we don't follow that convention for our other helpers so I removed the change

src/operations/search_indexes/create.ts
```diff
@@ -0,0 +1,48 @@
...
+  name?: string;
+
+  /** The index definition. */
+  description: Document;
+}
+
+/** @internal */
+export class CreateSearchIndexesOperation extends AbstractOperation<string[]> {
+  constructor(
+    private readonly collection: Collection<any>,
```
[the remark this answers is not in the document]
baileympearson: There isn't a standard pattern we follow for our operations.  For example, the `Find*` operations take a collection and no namespace.  So do the bulk operations.

The reason we haven't had this issue in the past is because we rely on the default `Document` for the collection generic in collections

src/operations/search_indexes/create.ts
```diff
@@ -0,0 +1,48 @@
...
+    private readonly descriptions: ReadonlyArray<SearchIndexDescription>
+  ) {
+    super();
+  }
+
+  execute(server: Server, session: ClientSession | undefined, callback: Callback<any>): void {
```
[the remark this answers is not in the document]
baileympearson: Do you mean typing it in `Callback`?

src/operations/search_indexes/list.ts
```diff
@@ -0,0 +1,22 @@
...
+import { AggregationCursor } from '../../cursor/aggregation_cursor';
+import type { AggregateOptions } from '../aggregate';
+
+/** @public */
+export type ListSearchIndexesOptions = AggregateOptions;
+
+/** @public */
+export class ListSearchIndexesCursor extends AggregationCursor<{ name: string }> {
+  /** @internal */
+  static create<T extends Document>(
```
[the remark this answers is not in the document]
baileympearson: Preference, mostly.  It's effectively the same, but I prefer factory functions over constructor subclassing because they can abstract details of object's construction and move logic out of the constructor.

Would you prefer I move the logic to ListSearchIndexesCursor.constructor?

src/operations/search_indexes/drop.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor(private readonly collection: Collection<any>, private readonly name: string) {
```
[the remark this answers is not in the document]
baileympearson: I fixed it instead by removing the generics.  Same for all of these comments

package.json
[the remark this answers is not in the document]
nbbeeken: We can leave it and decide differently later, gcp/azure kms landed in the integration directory despite being different variants, so it's a matter of choice with no consistent pattern.

nyc runs for kerberos and aws so seems like we could add it here to right? should just work right, might as well take what's free

src/collection.ts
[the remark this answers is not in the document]
nbbeeken: Ah I see, sounds like a good change to me, your call

src/db.ts
```diff
@@ -164,6 +164,11 @@ export class Db {
     };
   }
 
+  /** @internal */
+  get client(): MongoClient {
+    return this.s.client;
+  }
```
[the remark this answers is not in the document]
nbbeeken: Sure I wouldn't say no to taking it further, looks like 17 more lines with a quick search and no usage in tests

src/operations/search_indexes/create.ts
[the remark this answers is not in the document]
nbbeeken: Resolved in another thread, no need for `Collection<any>`

src/operations/search_indexes/create.ts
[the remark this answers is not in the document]
nbbeeken: Yep, for now at least that's where the result goes, when someone changes this to an async function I think it would be nice to have a "documented" result shape

src/operations/search_indexes/list.ts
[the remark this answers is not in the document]
nbbeeken: It is effectively the same, so is it only different for preference reasons? I'm not against your preference, I just don't want code that's unique to one implementation without providing meaning to the next engineer who comes along

Also can we move this to our cursors directory?

src/collection.ts
baileympearson: Screenshot 2023-05-30 at 1 00 49 PM
Screenshot 2023-05-30 at 1 00 30 PM

Maybe this is my particular workflow, but I often peruse the overloads for methods by scrolling through the modal that pops up in vscode.  I think it's beneficial to provide clear intended usages.  I added an additional tsdoc comment for the new overload.

Also, `listSearchIndexes(options, options)` won't work because the actual pipeline stage sent will be:

```
{ $listSearchIndexes: { name: <aggregation options > } }
```

nbbeeken: Yea that's fair, I think the union of types is equally clear, the possible inputs are listed in one popup, no scrolling needed. 

> Also, listSearchIndexes(options, options) won't work because the actual pipeline stage sent will be: `{ $listSearchIndexes: { name: <aggregation options > } }`

From the option handling:
```js
   : typeof indexNameOrOptions === 'object'
        ? null
```
This case should set the name to `null` and then the pipeline will be constructed correctly in the cursor since we test for `name == null`.

test/unit/utils.test.ts
```diff
@@ -12,13 +12,14 @@ import {
...
   maybeCallback,
+  MongoDBCollectionNamespace,
   MongoDBNamespace,
   MongoRuntimeError,
   ObjectId,
   shuffle
 } from '../mongodb';
 
-describe('driver utils', function () {
+describe.only('driver utils', function () {
```
nbbeeken, line 22: rogue only

src/cursor/list_search_indexes_cursor.ts
```diff
@@ -0,0 +1,20 @@
+import type { Collection } from '../collection';
+import type { AggregateOptions } from '../operations/aggregate';
+import { AggregationCursor } from './aggregation_cursor';
+
+/** @public */
+export type ListSearchIndexesOptions = AggregateOptions;
+
+/** @public */
```
dariakp, line 8: Are we making an intentional choice to have this and the other types exported in `index.ts` being public even though the APIs using them are internal? Is there any risk that we may need to change these underlying cursors before we officially release?

src/collection.ts
```diff
@@ -975,12 +998,111 @@ export class Collection<TSchema extends Document = Document> {
...
+   * @internal
+   *
+   * Updates a search index by replacing the existing index definition with the provided definition.
+   *
+   * @param name - The name of the search index to update.
+   * @param definition - The new search index definition.
+   *
+   * @remarks Only available when used against a 7.0+ Atlas cluster.
+   */
+  async updateSearchIndex(name: string, definition: Document): Promise<void> {
```
dariakp, line 1102: Have we enumerated everything that we'll need to remove `@internal` tags from when we are ready to release the feature? And if so, which ticket documents it?

dariakp (commented):
Just some clarifying questions

src/collection.ts
baileympearson: [NODE-5319](https://jira.mongodb.org/browse/NODE-5319) has been filed.

src/collection.ts
```diff
@@ -975,12 +998,111 @@ export class Collection<TSchema extends Document = Document> {
...
-        filter,
-        resolveOptions(this, options)
-      )
+      this.client,
+      new CountOperation(this.fullNamespace, filter, resolveOptions(this, options))
+    );
+  }
+
+  /**
+   * @internal
```
lerouxb, line 1007: Should these be @internal? Once released we're going to have to use them from mongosh, right?

lerouxb: That's fine, I guess. I forked this branch so I can work with it locally in the meantime anyway.

nbbeeken (dismissed):

dariakp (approved):

PR #3673, opened 2023-05-17
test(NODE-5295): reduce error message match strictness

nbbeeken:
#### What is changing?

Fixes an error match assertion on node 20


nbbeeken:
[Passing socks5-csfle](https://spruce.mongodb.com/task/mongo_node_driver_next_rhel80_large_Node20_test_socks5_csfle_patch_abbe61afb54175c8e6441ba7ebc47f75770d2c23_646529839ccd4eac564c582c_23_05_17_19_22_44/tests?execution=0&sortBy=STATUS&sortDir=ASC)

durran (dismissed):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1178,7 +1178,11 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
           .to.be.instanceOf(Error)
           .to.have.property('name', 'MongoServerSelectionError');
 
-        expect(insertError).to.match(/connect ECONNREFUSED 127.0.0.1:27021/);
+        expect(insertError, 'Error must contain ECONNREFUSED').to.satisfy(
+          error =>
+            /ECONNREFUSED/.test(error.message) ||
+            error.cause.cause.errors.every(e => e.code === 'ECONNREFUSED')
```
dariakp, line 1184: will this fail gracefully if the message doesn't match but also there is no `cause.cause.errors` property? I'm not sure what happens if you throw inside a `satisfy` clause like that

dariakp: I tested locally and the answer is "no", so as a follow up, can we make sure that this does fail gracefully with a reasonable message?

dariakp (changes requested):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
nbbeeken: optional chaining to the rescue, fixed

dariakp (dismissed):

baileympearson (changes requested):
can we make the PR description a bit better?  `reduce error message match strictness` is pretty vague and won't be much help in our commit history

test/integration/client-side-encryption/client_side_encryption.prose.test.js
```diff
@@ -1178,7 +1178,11 @@ describe('Client Side Encryption Prose Tests', metadata, function () {
           .to.be.instanceOf(Error)
           .to.have.property('name', 'MongoServerSelectionError');
 
-        expect(insertError).to.match(/connect ECONNREFUSED 127.0.0.1:27021/);
+        expect(insertError, 'Error must contain ECONNREFUSED').to.satisfy(
```
baileympearson, line 1181: Rather than hard-coding logic for different Node versions, why not inspect the error and search for `ECONNREFUSED`?
```suggestion
        expect(inspec(insertError), 'Error must contain ECONNREFUSED').to.match(/.*ECONNREFUSED.*/)
```

baileympearson (changes requested):

test/integration/client-side-encryption/client_side_encryption.prose.test.js
nbbeeken: I would want to know if there was yet another shape that the error took, the information went missing from the message in node 20, we will return it to the message in some way in: https://jira.mongodb.org/browse/NODE-5296, added todos and updated the ticket

baileympearson: From our public docs about our errors:

> It is our recommendation to use instanceof checks on errors and to avoid relying on parsing error.message and error.name strings in your code. We guarantee instanceof checks will pass according to semver guidelines, but errors may be sub-classed or their messages may change at any time, even patch releases, as we see fit to increase the helpfulness of the errors.

We don't guarantee error messages.  The error information is still present, the error is just structured differently.  I don't think preserving the exact error info is important and so checking that the error itself is the correct error type is sufficient.

nbbeeken: Oh I was not thinking about public API / semver, just usefulness

baileympearson (approved):

PR #3678, opened 2023-05-19
feat(NODE-5272): do not create or drop ecc collections

baileympearson:
#### What is changing?

The driver no longer creates an ecc collection when creating an encrypted collection.

Additionally, the QE v2 protocol requires 7.0+ servers.  This means that driver 4.x would be incompatible with QE unless we bump the max wire version for 4.x to 21.

Backport of [#3660](https://github.com/mongodb/node-mongodb-native/pull/3660)


src/operations/create_collection.ts
```diff
@@ -126,12 +135,17 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
         db.s.client.options.autoEncryption?.encryptedFieldsMap?.[`${db.databaseName}.${name}`];
 
       if (encryptedFields) {
+        // Creating a QE collection required min server of 7.0.0
+        if (server.description.maxWireVersion < MIN_SUPPORTED_QE_WIRE_VERSION) {
```
nbbeeken, line 139: Daria had a question about whether this was necessary

nbbeeken (changes requested):

src/operations/create_collection.ts
nbbeeken: @baileympearson bump

nbbeeken (approved):

nbbeeken:
@baileympearson Is "fix" correct here? asking because the original PR was a feat

baileympearson:
I think so - why was the original PR a feat?  

- QE is experimental, so the breakage is okay
- this fixes a bug with QE and 7.0 servers, where we'd create extra unnecessary collections

PR #3679, opened 2023-05-22
test(NODE-5237): fix flaky deadlock tests and modernize deadlock test suite

baileympearson:
#### What is changing?

The flaky deadlock tests are fixed.  The root cause appears to be a typo in the original test, where the encryption key was inserted without `{ w: majority }`.  

There are a lot of cleanups in this PR too, since I was touching the files.  The cleanups include:

- conversion of the deadlock tests to Typescript
- symbol properties have been removed from the capturing client
- the deadlock tests have been factored out into their own top level test suite
- the deadlock tests now use async/await
- enables the deadlock tests when auth is enabled

The commits are can be reviewed one-at-a-time to easily diff the changes, although the commits are not perfect.  Some changes introduced in a commit might be removed in a later commit (test name improvements for debugging and .only for example).


test/integration/client-side-encryption/client_side_encryption.prose.deadlock.ts
```diff
@@ -35,8 +35,8 @@ const kClientsCreated = Symbol('clientsCreated');
 class CapturingMongoClient extends MongoClient {
   [kEvents]: Array<CommandStartedEvent> = [];
   [kClientsCreated] = 0;
-  constructor(url, options: MongoClientOptions = {}) {
-    options = { ...options, monitorCommands: true };
+  constructor(url: string, options: MongoClientOptions = {}) {
+    options = { ...options, monitorCommands: true, [Symbol.for('@@mdb.skipPingOnConnect')]: true };
```
W-A-James: Out of curiosity, why was this flag added?

baileympearson: Good question - I did forget to mention that in the PR description.  I'm happy to provide details if you want but the tl;dr is that when the Node driver connects with authentication enabled, it runs an extra `{ ping: 1 }` command.  Other drivers do not do this.  Consequently, if tests rely on command monitoring to assert on the correct ordering of events (like many driver tests do), the Node driver has an extra `ping` command so the assertions fail but only when authentication is enabled.

If you peruse the repo, you'll notice a lot of our tests are skipped when auth is enabled.  More often than not, that's why.

`skipPingOnConnect` was added last year by Neal, and it disables this ping for the purposes of testing.

baileympearson: This was just a drive by cleanup to enable the tests for regular CI runs, instead of only the auth disabled tests.

W-A-James: Cool! Thanks for the info. If you could just mention that in the '*_What's changing_*' section of the PR description, then I'd say we're good to go.

W-A-James (approved):

nbbeeken (approved):

PR #3680, opened 2023-05-23
specify custom retry timeout for transactions in milliseconds

bhavya3024:
#### What is changing?
We can set custom timeout for transaction rewrites in milli seconds.

yes

We can use our custom timeout in milliseconds and that is optional. setting default by 120 seconds can be very huge in some cases for transactions (In our use case, it retried 15 times for some requests during load testing). So we can set custom timeout (smaller or larger) for multiple use cases.


bhavya3024:
@nbbeeken I understood the JIRA comment, but why the default timeout is twice the `transactionLifetimeLimitSeconds` ? It can be 10 or 20 seconds more for exceptional cases if the retry causes lets say 65 seconds (if I am not wrong). But I need the reason for setting the "twice" of `transactionLifetimeLimitSeconds`.

alexbevi:
@bhavya3024 I've replied to your question on NODE-5307 for better transparency in the event others have a similar question.

PR #3682, opened 2023-05-23
fix(NODE-5296): construct error messages for AggregateErrors in Node16+ 

baileympearson:
#### What is changing?

MongoErrors now appropriately handle AggregateErrors and construct a message from the error messages of each sub-error.

No.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -108,6 +108,16 @@ export interface ErrorDescription extends Document {
   errInfo?: Document;
 }
 
+function isAggregateError(e: unknown): e is Error & { errors: Error[] } {
+  return (
+    typeof e === 'object' &&
+    e != null &&
+    e instanceof Error &&
+    'errors' in e &&
+    Array.isArray(e.errors)
+  );
+}
+
 /**
  * @public
  * @category Error
@@ -132,7 +142,13 @@ export class MongoError extends Error {
 
   constructor(message: string | Error) {
     if (message instanceof Error) {
-      super(message.message);
+      if (isAggregateError(message)) {
+        const combinedMessage = 'Aggregate Error: ' + message.errors.map(e => e.message).join(', ');
+        super(combinedMessage);
+      } else {
+        super(message.message);
+      }
+
       this.cause = message;
     } else {
       super(message);
--- a/test/integration/node-specific/errors.ts
+++ b/test/integration/node-specific/errors.ts
@@ -1,0 +1,40 @@
+import { expect } from 'chai';
+
+import { MongoClient, MongoError } from '../../mongodb';
+
+describe('Error (Integration)', function () {
+  describe('AggregateErrors', function () {
+    it('constructs the message properly', { requires: { nodejs: '>=16' } }, () => {
+      for (const { errors, message } of [
+        { errors: [], message: 'Aggregate Error: ' },
+        { errors: [new Error('message 1')], message: 'Aggregate Error: message 1' },
+        {
+          errors: [new Error('message 1'), new Error('message 2')],
+          message: 'Aggregate Error: message 1, message 2'
+        }
+      ]) {
+        const error = new AggregateError(errors);
+        const mongoError = new MongoError(error);
+
+        expect(
+          mongoError.message,
+          `built the message properly with an array of ${errors.length} errors`
+        ).to.equal(message);
+      }
+    });
+
+    it('sets the aggregate error to the cause property', { requires: { nodejs: '>=16' } }, () => {
+      const error = new AggregateError([new Error('error 1')]);
+      const mongoError = new MongoError(error);
+      expect(mongoError.cause).to.equal(error);
+    });
+  });
+
+  it('NODE-5296: handles aggregate errors from dns lookup', async function () {
+    const error = await MongoClient.connect('mongodb://localhost:27222', {
+      serverSelectionTimeoutMS: 1000
+    }).catch(e => e);
+    expect(error).to.be.instanceOf(Error);
+    expect(error.message).to.match(/ECONNREFUSED/);
+  });
+});
--- a/global.d.ts
+++ b/global.d.ts
@@ -12,6 +12,7 @@ declare global {
       serverless?: 'forbid' | 'allow' | 'require';
       auth?: 'enabled' | 'disabled';
       idmsMockServer?: true;
+      nodejs?: string;
     };
 
     sessions?: {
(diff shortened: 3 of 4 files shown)


test/integration/node-specific/errors.ts
```diff
@@ -0,0 +1,40 @@
+import { expect } from 'chai';
+
+import { MongoClient, MongoError } from '../../mongodb';
+
+describe('Error (Integration)', function () {
```
baileympearson, line 5: Because the behavior of the test depends on Node version (AggregateError is only defined Node16+), I added these as integration tests instead of unit tests.

src/error.ts
```diff
@@ -132,7 +142,13 @@ export class MongoError extends Error {
 
   constructor(message: string | Error) {
     if (message instanceof Error) {
-      super(message.message);
+      if (isAggregateError(message)) {
+        const combinedMessage = 'Aggregate Error: ' + message.errors.map(e => e.message).join(', ');
+        super(combinedMessage);
```
nbbeeken, line 147:
```suggestion
        super(message.errors.map(({ message }) => message).join(', '));
```

Is indication of "Aggregate Error" important? I would expect us to have the error messages directly

test/integration/node-specific/errors.ts
nbbeeken: Our isAggregateError check is not strict enough to need us to test with actual AggregateErrors, so we could make these unit tests, I don't feel very strongly but we just wouldn't need a node filter if we moved these

src/error.ts
```diff
@@ -108,6 +108,16 @@ export interface ErrorDescription extends Document {
...
 
+function isAggregateError(e: unknown): e is Error & { errors: Error[] } {
+  return (
+    typeof e === 'object' &&
+    e != null &&
+    e instanceof Error &&
+    'errors' in e &&
+    Array.isArray(e.errors)
+  );
+}
```
nbbeeken, line 119: Do we need to confirm `instanceof Error`?

I think we should change the check here a bit, instead of focusing on "AggregateError" specifically, try and confirm more of the contract we _need_ from the object we're given

- [x] It's a non-null object
- [ ] If it has a `message` property, then it is an empty string
  - AggregateErrors can have a message, and if they do have one it should be preferred over building the joined string.
- [x] It has an `errors` property
- [ ] `errors` is a non-zero array
  - The non-zero check is to make sure that our `errors.map` actually returns something, but also we have no information to go off of, so maybe if we reach this case we can put some non-empty default, "see cause" or "good luck".

nbbeeken (changes requested):

test/integration/node-specific/errors.ts
baileympearson: AggregateErrors are currently the only errors that fit this description that we handle - I'd prefer to test with the actual errors we expect to see than creating an error that conforms to our contract for unit testing.

src/error.ts
```diff
@@ -131,15 +135,28 @@ export class MongoError extends Error {
...
+
     this[kErrorLabels] = new Set();
   }
 
+  /** @internal */
+  private static buildErrorMessage(e: Error | string): string {
+    if (typeof e === 'string') {
+      return e;
+    }
+    if (isAggregateError(e) && e.message.length === 0) {
```
nbbeeken, line 151:
```suggestion
    if (error.message.length === 0 && 'errors' in error && Array.isArray(error.errors)) {
```

Maybe we don't need the type assertion any more? b/c you need to make the argument `unknown` vs staying in this function means you've already narrowed to `Error`, less things to check

src/error.ts
```diff
@@ -131,15 +135,28 @@ export class MongoError extends Error {
...
       this.cause = message;
-    } else {
-      super(message);
     }
+
     this[kErrorLabels] = new Set();
   }
 
+  /** @internal */
+  private static buildErrorMessage(e: Error | string): string {
```
nbbeeken, line 147:
```suggestion
  private static buildErrorMessage(error: Error | string): string {
```

src/error.ts
```diff
@@ -131,15 +135,28 @@ export class MongoError extends Error {
...
+    if (typeof e === 'string') {
+      return e;
+    }
+    if (isAggregateError(e) && e.message.length === 0) {
+      return e.errors.length === 0
+        ? 'AggregateError: no suberrors have error messages.  please check the `cause` property for more error information.'
```
nbbeeken, line 153:
```suggestion
        ? 'AggregateError has an empty errors array. Please check the `cause` property for more information.'
```

nbbeeken (changes requested):

test/integration/node-specific/errors.ts
```diff
@@ -0,0 +1,51 @@
+import { expect } from 'chai';
+
+import { MongoClient, MongoError } from '../../mongodb';
+
+describe.only('Error (Integration)', function () {
```
nbbeeken, line 5:
```suggestion
describe('Error (Integration)', function () {
```

GO_TO_JAIL

nbbeeken (changes requested):

src/error.ts
baileympearson: I actually chose `e` intentionally, because it's not always an error (it can be a string).  Would you still like this changed?

nbbeeken: ah fair enough, I don't feel strongly

src/error.ts
baileympearson: the type assertion provides type safety on the `.errors` property inside the if-block, and it provides better clarify.  I updated the type assertion to only assert on Error types

nbbeeken (approved):

test/integration/node-specific/errors.ts
```diff
@@ -0,0 +1,54 @@
...
+          const error = new AggregateError(errors);
+          const mongoError = new MongoError(error);
+
+          expect(mongoError.message).to.equal(message);
+        }
+      );
+    }
+
+    context('when the message on the AggregateError is non-empty', () => {
+      it(`uses the AggregateError's message`, () => {
```
nbbeeken, line 32: Need a filter setting here ([failure](https://parsley.mongodb.com/evergreen/mongo_node_driver_next_rhel8_no_auth_tests_test_3.6_replica_set_noauth_patch_bf413e5beca843b11a4fb1159e5ce0738ee292cb_646e7342850e61eb51b3db65_23_05_24_20_27_47/0/task?bookmarks=0,7854&shareLine=7808))

nbbeeken (changes requested):

nbbeeken (approved):

PR #3684, opened 2023-05-25
ci(NODE-4698): test csfle with mongocryptd

W-A-James:
#### What is changing?
*Summary*
Adding new CI variants to run CSFLE tests with mongocryptd instead of crypt_shared, which we already had coverage for.

##### changes to `config.in.yml`
- script in "run tests" now conditionally executes `.evergreen/prepare-crypt-shared-lib.sh`
- script in "run tests" now sets the `TEST_NPM_SCRIPT` env variable

##### changes to `generate_evergreen_tasks.js`
- Added new tasks that only run the csfle integration suite on relevant server versions and topologies using mongocryptd

No

Spec compliance

### DIFF
--- a/.evergreen/generate_evergreen_tasks.js
+++ b/.evergreen/generate_evergreen_tasks.js
@@ -1,5 +1,6 @@
 const fs = require('fs');
 const yaml = require('js-yaml');
+const semver = require('semver');
 const { mongoshTasks } = require('./generate_mongosh_tasks');
 
 const {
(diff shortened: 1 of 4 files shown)


.evergreen/generate_evergreen_tasks.js
```diff
@@ -461,6 +461,54 @@ for (const {
   }
 }
 
+// Running CSFLE tests with mongocryptd
+const MONGOCRYPTD_CSFLE_TASKS = MONGODB_VERSIONS
+  .filter((mongoVersion) => { // Only run on server versions >= 4.2
+    const numericVersion = Number(mongoVersion);
+    return !Number.isNaN(numericVersion) && numericVersion >= 4.2;
+  })
```
baileympearson, line 469: We actually have a package called `semver` that we use for version comparison:

also, I think we should test on latest / rapid as well

```suggestion
  .filter((version) => ['latest', 'rapid'].includes(version) || semver.gte(version, '4.2'))
```

baileympearson (changes requested):
one small comment, otherwise looks great

baileympearson (approved):

PR #3688, opened 2023-05-26
test(NODE-5265): fix flaky operation count test

durran:
### Description

Fixes operation count flaky test.

#### What is changing?

Wraps the increment/decrement of the server's operation count in 2 methods, that then can be spied on to assert the number of times each was called.

None

NODE-5265

### DIFF
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -343,13 +343,13 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       return;
     }
 
-    this.s.operationCount += 1;
+    this.incrementOperationCount();
 
     this.pool.withConnection(
       conn,
       (err, conn, cb) => {
         if (err || !conn) {
-          this.s.operationCount -= 1;
+          this.decrementOperationCount();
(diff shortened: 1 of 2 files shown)


src/sdam/server.ts
```diff
@@ -420,6 +420,20 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
+  private decrementOperationCount(): number {
+    return (this.s.operationCount -= 1);
+  }
+
+  /**
+   * Increment the operation count, returning the new count.
+   */
+  private incrementOperationCount(): number {
+    return (this.s.operationCount += 1);
+  }
```
baileympearson, line 436: Any reason to return the new count, if we're not chaining it or using the returned value?

baileympearson (changes requested):
one question, just requesting changes for visibility.

src/sdam/server.ts
durran: No was just a general practice I like to use - if the method would normally be void try potentially returning something useful if possible.

PR #3692, opened 2023-05-30
fix(NODE-5197): fall back to polling monitoring in faas env

W-A-James:

### DIFF
--- a/src/sdam/monitor.ts
+++ b/src/sdam/monitor.ts
@@ -76,6 +76,7 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
     Pick<MonitorOptions, 'connectTimeoutMS' | 'heartbeatFrequencyMS' | 'minHeartbeatFrequencyMS'>
   >;
   connectOptions: ConnectionOptions;
+  isInFAASEnv: boolean;
   [kServer]: Server;
   [kConnection]?: Connection;
   [kCancellationToken]: CancellationToken;
@@ -105,6 +106,9 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
       heartbeatFrequencyMS: options.heartbeatFrequencyMS ?? 10000,
       minHeartbeatFrequencyMS: options.minHeartbeatFrequencyMS ?? 500
     });
+
+    // eslint-disable-next-line eqeqeq
+    this.isInFAASEnv = server.topology.clientMetadata?.env?.name != undefined;
 
     const cancellationToken = this[kCancellationToken];
     // TODO: refactor this to pull it directly from the pool, requires new ConnectionPool integration
@@ -283,7 +287,7 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
 
       // if we are using the streaming protocol then we immediately issue another `started`
       // event, otherwise the "check" is complete and return to the main monitor loop
-      if (isAwaitable && hello.topologyVersion) {
+      if (isAwaitable && hello.topologyVersion && !monitor.isInFAASEnv) {
         monitor.emit(
           Server.SERVER_HEARTBEAT_STARTED,
           new ServerHeartbeatStartedEvent(monitor.address)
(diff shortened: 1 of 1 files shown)


src/sdam/monitor.ts
```diff
@@ -310,6 +319,7 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
     }
 
     if (conn) {
+      // TODO: Check if this is correct
```
durran, line 322: I think what we should probably do here is rename the property on the connection to `isStreamingMonitoringConnection` and then set the value here to `monitor.heartbeatProtocol === 'streaming`.

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -62,3 +64,89 @@ describe('MongoDB Handshake', () => {
...
+        }
+      );
+    });
+
+    it('uses the polling protocol', metadata, async function () {
+      await client.connect();
+      const servers = client?.topology.s?.servers;
+      expect(servers).to.exist;
+      for (const server of servers.values()) {
+        const symbolMap = {};
```
durran, line 101: Here and also in the other test instead of constructing an object and putting all the symbols in it, we could just use `find` on the iterable:

```ts
const monitorSymbol = Object.getOwnPropertySymbols(server).find(s => s.toString() === 'Symbol(monitor)');
```

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -62,3 +64,89 @@ describe('MongoDB Handshake', () => {
...
+        {},
+        {
+          heartbeatFrequencyMS: HEARTBEAT_FREQUENCY_MS,
+          minHeartbeatFrequencyMS: MIN_HEARTBEAT_FREQUENCY_MS
+        }
+      );
+    });
+
+    it('uses the streaming protocol', metadata, async function () {
+      await client.connect();
```
durran, line 137:
```suggestion
      await client?.connect();
```

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -62,3 +64,89 @@ describe('MongoDB Handshake', () => {
...
+        {},
+        {
+          heartbeatFrequencyMS: HEARTBEAT_FREQUENCY_MS,
+          minHeartbeatFrequencyMS: MIN_HEARTBEAT_FREQUENCY_MS
+        }
+      );
+    });
+
+    it('uses the polling protocol', metadata, async function () {
+      await client.connect();
```
durran, line 97:
```suggestion
      await client?.connect();
```

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -62,3 +64,89 @@ describe('MongoDB Handshake', () => {
...
+  const metadata = { requires: { mongodb: '>= 4.4', topology: '!load-balanced' } };
+  let cachedEnv: NodeJS.ProcessEnv;
+  let client: MongoClient;
+
+  beforeEach(async function () {
+    cachedEnv = { ...process.env };
+  });
+
+  afterEach(async function () {
+    await client.close().catch(() => null);
```
durran, line 79:
```suggestion
    await client?.close().catch(() => null);
```

durran (changes requested):

src/sdam/monitor.ts
W-A-James: If I'm understanding correctly, at this point we are carrying out our initial connection, so we wouldn't know whether or not the server is capable of streaming yet, right? In that case we should check the server hello, use that to determine whether or not we should be polling or streaming and then set `isStreamingMonitoringConnection` appropriately.

test/integration/mongodb-handshake/mongodb-handshake.test.ts
nbbeeken: We have the [getSymbolFrom](https://github.com/mongodb/node-mongodb-native/blob/main/test/tools/utils.ts#L115) helper that I think helps here

PR #3696, opened 2023-06-01
fix(NODE-5316): prevent parallel topology creation in MongoClient.connect

baileympearson:
#### What is changing?

Backports the mongo client connect log fix from https://github.com/mongodb/node-mongodb-native/pull/3596.

No.

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -449,54 +449,65 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
     }
 
     return maybeCallback(async () => {
-      if (this.topology && this.topology.isConnected()) {
+      try {
+        if (this.connectionLock) return await this.connectionLock;
+
+        this.connectionLock = this._connect();
+        await this.connectionLock;
         return this;
+      } finally {
+        this.connectionLock = undefined;
       }
-
-      const options = this[kOptions];
-
-      if (typeof options.srvHost === 'string') {
-        const hosts = await resolveSRVRecord(options);
-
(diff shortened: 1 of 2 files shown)


src/mongo_client.ts
```diff
@@ -447,54 +449,63 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
     }
 
     return maybeCallback(async () => {
-      if (this.topology && this.topology.isConnected()) {
+      try {
+        this.connectionLock = this.connectionLock ?? this._connect();
```
nbbeeken, line 453:
```suggestion
        this.connectionLock ??= this._connect();
```
might as well short hand it

However, there is now a difference in the 5.x and 4.x implementations. This now always `await`s inside this function as opposed to the early returning of the existing promise without awaiting it. I am not sure we should be introducing a divergence to 4.x here

baileympearson: What is your concern with the modification here?  It works slightly differently internally, but from the user's perspective our API still provides a promise that resolves to a connected client

baileympearson: actually ??= is only available Node 15+.

nbbeeken: > actually ??= is only available Node 15+.

`??` is not in node 12 either, but typescript to the rescue!

> API still provides a promise that resolves to a connected client

That is correct, but it does so in more steps than needed, this refactor circumvents the `@typescript-eslint/return-await` lint rule we have to catch these situations.

nbbeeken (approved):

PR #3698, opened 2023-06-01
chore(NODE-5328): add dry-run nightly release automation

nbbeeken:
#### What is changing?

Add nightly release automation.

With `--dry-run` flag 

A script checks the current nightly sha against main's sha, and will modify package.json if they differ. npm publish will only run if this script exits successfully.

Version string format: `MAJ.MIN.PAT-dev+YYYYMMDD.sha.##########`
- [x] finalized
- [ ] Setup credentials

- [ ] TODO mention nightly tag in README.md

Allow downstream projects to continuosly pull in driver changes


.github/workflows/release-nightly.yml
```diff
@@ -0,0 +1,29 @@
+on:
+  schedule:
+    # https://crontab.guru/#0_12_*_*_1-5
+    # At 12:00 on every day-of-week from Monday through Friday.
+    - cron: '0 12 * * 1-5'
```
baileympearson, line 5: This runs the script at noon, no?  Don't we want it to run sometime during the night? 

Or does this run in UTC time, and you accounted for the timezone difference?

.github/workflows/release-nightly.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          ref: main
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 'lts/*'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm install -g npm
+      - run: npm clean-install
+      - run: npm run build:nightly
+        continue-on-error: true
```
baileympearson, line 25: Why are we continuing on error?  iiuc, the script intentionally exists with code 1 when there is no release needed, so we should _not_ continue executing the action, right?

baileympearson (changes requested):
High-level question: how can we test this action works properly without releasing?

.github/scripts/nightly.mjs
```diff
@@ -0,0 +1,74 @@
+// @ts-check
+import * as url from 'node:url';
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+import * as process from 'node:process';
+import * as child_process from 'node:child_process';
+import * as util from 'node:util';
```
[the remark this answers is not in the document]
nbbeeken: Consistency, I had a mix and just decided to unified them

.github/scripts/nightly.mjs
```diff
@@ -0,0 +1,74 @@
...
+const exec = util.promisify(child_process.exec);
+
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url));
+const pkgFilePath = path.join(__dirname, '..', '..', 'package.json');
+
+process.env.TZ = 'Etc/UTC';
```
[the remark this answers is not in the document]
nbbeeken: The cron job runs in UTC time, so I want to make sure that the node's Date() is the same, I am fairly certain TZ would already be utc up in GH action, but might as well encode it to be sure.

.github/workflows/release-nightly.yml
nbbeeken: It will run at 8am EDT, 2pm CET currently. I stuck a time in there expecting to finalize it here.

> High-level question: how can we test this action works properly without releasing?

We cannot beyond the fact that I've at least confirmed the node script runs fine locally. I was thinking we should put in a time that is shortly after this PR merges, and we can make fixes if we hit an issue or if it works then we can change it to the hour we want it to release at.

.github/workflows/release-nightly.yml
nbbeeken: `continue-on-error` means that GH won't mark the whole workflow as failing when this exits with a 1. Then we use the `if` to publish or not

.github/workflows/release-nightly.yml
dariakp: I think we should put the time that we want it to run, check whether it does work the following morning, and disable or make adjustments as needed if it doesn't work the first time

dariakp: We should also wait to merge it Monday, ideally

nbbeeken: Any preference for a time? I was thinking 8am would mean most of the team is waking up so we could address issues that same day (if any)

dariakp: Alternative suggestion, per slack: we can keep the time of expected merge and just make this a dry run, then do a follow-up PR to remove the dry run flag and set the time to the proper nightly time

nbbeeken: I've added the --dry-run flag to publish, so we can set a time to trigger, and see all the output to verify, then we can go back remove dry-run and set the final time we want to use

baileympearson: I don't follow.  Is the idea to, prior to merging, set the cron to run at a time just in the future after we intend to merge it?

nbbeeken: Yep, when we're approved on everything else I can adjust the time to be what ever hour is closest after our merge, then we can pick up NODE-5336 to set the time and remove dry-run

baileympearson (dismissed):

baileympearson (approved):

PR #3701, opened 2023-06-02
test(NODE-5198): fix kms request failure

durran:
### Description

CI was getting frequent errors "KMS Request Failed" due to the KMIP server not being started by the time the tests that needed it ran. This was a common happening across all drivers, and while each driver addressed it in its own manner the common solution was that the KMIP server needed to be started in its own task.

#### What is changing?
- Splits the startup of the KMIP and KMS HTTP servers in 2 separate subprocess tasks that run in the background.
- Splits out the activation of the kms virtual env into its own subprocess task that does not run in the background, to ensure all the python requirements are fulfilled synchronously.

None

NODE-5198

### DIFF
--- a/.evergreen/run-kms-servers.sh
+++ b/.evergreen/run-kms-servers.sh
@@ -1,9 +1,0 @@
-#!/bin/bash
-
-cd ${DRIVERS_TOOLS}/.evergreen/csfle
-. ./activate-kmstlsvenv.sh
-# by default it always runs on port 5698
-python -u kms_kmip_server.py &
-python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/expired.pem --port 8000  &
-python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/wrong-host.pem --port 8001  &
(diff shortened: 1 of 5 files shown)


.evergreen/run-kms-servers.sh
```diff
@@ -1,9 +1,15 @@
...
-# by default it always runs on port 5698
-python -u kms_kmip_server.py &
-python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/expired.pem --port 8000  &
-python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/wrong-host.pem --port 8001  &
-python -u kms_http_server.py --ca_file ../x509gen/ca.pem --cert_file ../x509gen/server.pem --port 8002 --require_client_cert &
+if [ "Windows_NT" = "$OS" ]; then
+  PYTHON="kmstlsvenv/Scripts/python.exe"
+else
+  PYTHON="./kmstlsvenv/bin/python3"
+fi
```
W-A-James, line 9: Why do we duplicate this logic across `run-kmip-server.sh` and `run-kms-servers.sh` instead of exporting this in `activate-kms-venv.sh` or a similarly named script?

durran: Moved into single spot in activate-kms-venv.sh

W-A-James (approved):

PR #3720, opened 2023-06-09
test(NODE-5342): modernize explain tests

malikj2000:
#### What is changing?
explain tests modernized
removed metadata
converted to typescript
no
new things are nice and look clean


test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      expect(explanation).nested.property('executionStats.allPlansExecution').to.exist;
+    });
+
+    it('findOne returns queryPlanner explain result', async function () {
+      const explanation = await collection.findOne({ a: 1 }, { explain: true });
+      expect(explanation).property('queryPlanner').to.exist;
+    });
+  });
+
+  context('when explain is not set to true', () => {
```
W-A-James, line 85: You can get rid of this context block since the value of `explain` is specified in each of the blocks inside this one

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+    context('when explain is set to false', () => {
+      it('only queryPlanner property is used in explain result', async function () {
+        const explanation = await collection.deleteOne({ a: 1 }, { explain: false });
+        expect(explanation).property('queryPlanner').to.exist;
+        expect(explanation).to.not.have.property('executionStats');
```
W-A-James, line 90: For this assertion, since you want the explanation object to _only_ have the `queryPlanner` field, then it would be correct to use the `keys` assertion ([docs here](https://www.chaijs.com/api/bdd/#method_keys)), which would look something like.

```ts
expect(explanation).to.have.all.keys('queryPlanner')
```

This fix would apply to the other tests that check for this specific condition.

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+        const explanation = await collection.findOneAndReplace(
+          { a: 1 },
+          { a: 2 },
+          { explain: 'queryPlanner' }
+        );
+        expect(explanation).property('queryPlanner').to.exist;
+      });
+    });
+
+    context('when explain is set to executionStats', () => {
```
W-A-James, line 116:
```suggestion
    context('when explain is set to "executionStats"', () => {
```

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      });
+
+      it('findOne returns executionStats explain result', async function () {
+        const explanation = await collection.findOne({ a: 1 }, { explain: 'executionStats' });
+        expect(explanation).property('queryPlanner').to.exist;
+        expect(explanation).property('executionStats').to.exist;
+      });
+    });
+
+    context('when explain is set to allPlansExecution', () => {
```
W-A-James, line 145:
```suggestion
    context('when explain is set to "allPlansExecution"', () => {
```

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  it('should throw a catchable error with invalid explain string', async function () {
+    const db = client.db('shouldThrowCatchableError');
+    const collection = db.collection('test');
+    const error = await collection
+      .find({ a: 1 })
+      .explain('invalidExplain')
+      .catch(error => error);
+    expect(error).to.be.instanceOf(MongoServerError);
+  });
```
W-A-James, line 281: I would put this under a context block with a title like `when explain is set to an invalid value' and change the title of this test to follow suit.

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      });
+    });
+  });
+
+  it('find returns queryPlanner explain result', async () => {
+    const [explanation] = await collection.find({ a: 1 }, { explain: true }).toArray();
+    expect(explanation).property('queryPlanner').to.exist;
+  });
+
+  it('should honor boolean explain with aggregate', async function () {
```
W-A-James, line 166: The aggregate tests provide another opportunity to use a context block for organization and getting rid of duplicate information in the test title and test setup.

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+    const db = client.db('shouldHonorBooleanExplainWithAggregate');
+    const collection = db.collection('test');
+    await collection.insertOne({ a: 1 });
```
W-A-James, line 169: This setup is duplicated; you already have a client, database and collection created in the top-level describe block

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      expect(aggResult[0]).to.have.property('queryPlanner');
+      expect(aggResult[0]).to.have.property('executionStats');
+    }
+  });
+
+  it('should honor string explain with aggregate', async function () {
+    const db = client.db('shouldHonorStringExplainWithAggregate');
+    const collection = db.collection('test');
+
+    await collection.insertOne({ a: 1 });
```
W-A-James, line 192: duplicate setup

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,282 @@
...
+  let client: MongoClient;
+  let db: Db;
+  let collection: Collection;
+  let commandsStarted: CommandStartedEvent[];
+
+  beforeEach(async function () {
+    client = this.configuration.newClient({ monitorCommands: true });
+    db = client.db('queryPlannerExplainResult');
+    collection = db.collection('test');
+    commandsStarted = [];
```
W-A-James, line 21: (nitpick) since this is only used in one test, you can move the declaration of `commandsStarted` and the setup of the `'commandStarted'` listener into that test.

test/integration/crud/explain.test.ts
```diff
@@ -80,203 +69,182 @@ describe('Explain', function () {
...
 
-  it('should throw a catchable error with invalid explain string', async function () {
-    const db = client.db('shouldThrowCatchableError');
-    const collection = db.collection('test');
-    const error = await collection
-      .find({ a: 1 })
-      .explain('invalidExplain')
-      .catch(error => error);
-    expect(error).to.be.instanceOf(MongoServerError);
+  context('when explain is set to "invalidExplain", result returns MongoServerError', () => {
```
W-A-James, line 241: In this context, we don't necessarily care that explain is set to `'invalidExplain'` exactly. More so, we just want to know that when the value provided isn't something we expect, i.e. it's not `true`, `false`, `'allPlansExecution'`, `'queryPlanner'` or `'executionStats'`, so we can say that explain is set to 'an invalid value'.

Also, when using context blocks, we use them to describe exactly that: the context. We leave what the expectation is for our `it` blocks.

Think of reading the test like this, first you have your top level `describe` block and then you have your `context` block describing the scenario (which can be nested) and then you have the `it` block describing the expected outcome. Taking a look at what this would look like here, we'd have

> [describe] Explain: [context] when explain is set to an invalid value, result returns MongoServerError, [it] should throw a catchable error with invalid explain string

which is clunky to read as it duplicates information.

We want it to read more like:

> [describe] Explain: [context] when explain is set to an invalid value, [it] should throw a catchable error with invalid explain string

test/integration/crud/explain.test.ts
```diff
@@ -1,40 +1,29 @@
...
     await collection.drop();
     await client.close();
-    commandsStarted = [];
   });
 
   context('when explain is set to true', () => {
     it('deleteOne returns queryPlanner explain result', async function () {
       const explanation = await collection.deleteOne({ a: 1 }, { explain: true });
       expect(explanation).property('queryPlanner').to.exist;
-      expect(commandsStarted[0]).to.have.nested.property('command.explain.delete');
```
W-A-James: We still want to assert on this, but since it's the only test that relies on the command started listener and events, we want to move that setup into this test, not remove it entirely.

test/integration/crud/explain.test.ts
```diff
@@ -80,203 +69,182 @@ describe('Explain', function () {
...
-      });
-
-      it('find returns queryPlanner explain result specified on cursor', async function () {
-        const explanation = await collection.find({ a: 1 }).explain(false);
-        expect(explanation).property('queryPlanner').to.exist;
-      });
+  context('when explain is set to false', () => {
+    it('only queryPlanner property is used in explain result', async function () {
+      const explanation = await collection.deleteOne({ a: 1 }, { explain: false });
+      expect(explanation).property('queryPlanner').to.exist;
```
W-A-James, line 82:
```suggestion
      expect(explanation).to.have.all.keys('queryPlanner');
```

test/integration/crud/explain.test.ts
```diff
@@ -80,203 +69,182 @@ describe('Explain', function () {
...
-
-      it('findOne returns executionStats explain result', async function () {
-        const explanation = await collection.findOne({ a: 1 }, { explain: 'executionStats' });
-        expect(explanation).property('queryPlanner').to.exist;
-        expect(explanation).property('executionStats').to.exist;
-      });
+  context('when explain is set to "queryPlanner"', () => {
+    it('only queryPlanner property is used in explain result', async function () {
+      const explanation = await collection.deleteOne({ a: 1 }, { explain: 'queryPlanner' });
+      expect(explanation).property('queryPlanner').to.exist;
```
W-A-James, line 94:
```suggestion
      expect(explanation).to.have.all.keys('queryPlanner');
```

W-A-James (changes requested):

W-A-James (dismissed):

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    it('find returns queryPlanner explain result', async () => {
+      const [explanation] = await collection.find({ a: 1 }, { explain: true }).toArray();
+      expect(explanation).property('queryPlanner').to.exist;
+    });
+  });
+
+  context('when explain is set to false', () => {
+    it('only queryPlanner property is used in explain result', async function () {
+      const explanation = await collection.deleteOne({ a: 1 }, { explain: false });
+      expect(explanation).property('queryPlanner').to.exist;
```
dariakp, line 82: Two comments on this test - first, the title of the test doesn't reflect the actual api used (`deleteOne`) - this would be fine if the assertion is intended to be generic across all APIs, but if that were the case, we would want to call at least a couple of different APIs and assert on the outcome of each. Second, it's odd to me that the assertion in the "when explain is set to true" and "when explain is set to false" is identical (as is the "when explain is set to queryPlanner" lower down). Is the behavior actually identical for all these values of explain? If not, then we should make sure that the assertions clearly differentiate the expected behavior.

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    it('find returns "queryPlanner" explain result specified on cursor', async function () {
+      const explanation = await collection.find({ a: 1 }).explain(false);
+      expect(explanation).property('queryPlanner').to.exist;
+    });
+  });
+
+  context('when explain is set to "queryPlanner"', () => {
+    it('only queryPlanner property is used in explain result', async function () {
+      const explanation = await collection.deleteOne({ a: 1 }, { explain: 'queryPlanner' });
+      expect(explanation).property('queryPlanner').to.exist;
```
dariakp, line 94: this assertion by itself does not guarantee that *only* `queryPlanner` exists; this test case would pass even if other properties were present on the result

dariakp (changes requested):
I did not review the entire set of tests, but I think it would be worthwhile to double check all test cases and make sure that 1) the "it" blocks accurately capture the behavior of interest and 2) the assertions fully test the described behavior. I called out one example of what I mean, let me know if you have any questions.

test/integration/crud/explain.test.ts
W-A-James: The reason we expect that the same behaviour for `explain=false` and `explain='queryPlanner'` is that we [interpret our boolean values](https://github.com/mongodb/node-mongodb-native/blob/dc068ac1f6e91794836e103f8a06c2580405ad76/src/explain.ts#L32-L40) of explain more like numeric verbosity with `false` corresponding to the lowest verbosity which is `queryPlanner` and `true` to the highest which is `allPlansExecution`. It's admittedly a strange choice, but it's a choice that got made 😅

test/integration/crud/explain.test.ts
```diff
@@ -77,21 +77,24 @@ describe('Explain', function () {
   });
 
   context('when explain is set to false', () => {
-    it('only queryPlanner property is used in explain result', async function () {
+    it('deleteOne only returns queryPlanner property in explain result', async function () {
       const explanation = await collection.deleteOne({ a: 1 }, { explain: false });
       expect(explanation).property('queryPlanner').to.exist;
+      expect(explanation).to.not.have.property('executionStats');
```
dariakp, line 83: If this is the differentiating assertion for `explain: false` vs `explain: true`, I would expect the `explain: true` results to assert that this property does exist; however this still doesn't guarantee that ONLY `queryPlanner` is returned. We do want to make sure we aren't over-specifying the result of the server reply for `explain` if the server api response is not under semver (I do no recall whether or not it is for explain), but I am not convinced that we have found the sweet spot between overspecifying and underspecifying it here. It may be worthwhile to take a step back and ask "what is the behavior that is important for the *driver* to guarantee here?" If it's to make sure that we correctly set `explain`, then we could just use command monitoring. If it's to make sure we don't strip explain data from the reply, we need to define what data we expect to see for each case, and then capture that in the tests.

dariakp (changes requested):

test/integration/crud/explain.test.ts
dariakp: @W-A-James That's very good context - thank you - I think we can capture that in the test descriptions (or comments) to avoid any similar confusion in the future. This could be done, for example, via a wrapper `context('lowest verbosity')` and `context('highest verbosity')` for the other groupings

test/integration/crud/explain.test.ts
nbbeeken: The shape of the explain result changes based on server version and internal implementation of the query engine, so not semver for driver nor server. I agree with testing commands sent, although, it's good to have _some_ verification that the explain result can be correctly returned from our APIs since they do not match the expected return type of those methods. The verification has to be some definition of lightly asserted since we can't rely on explain having the same shape all the time.

I do think its scope creep to add command monitoring tests since we set out to refactor what exists, and confirming queryPlanner on the resulting object seems like the sweet spot for the moment to confirm explain response can make it out of our APIs. @malikj2000 thoughts?

malikj2000: I agree, I do think command monitoring tests would help make the tests much clearer and more accurate to what the tests were actually meant to do. However, I think this is a good stopping point for this PR since we are very close to being done. We should open up another ticket for adding command monitoring tests.

W-A-James: You're right when it comes to scope creep on this ticket. @malikj2000 and/or @nbbeeken can you file a follow-up ticket to address this?

W-A-James: Specifically capturing the changes we need to make to the explain tests to capture driver-specific behaviour with command monitoring.

In addition, if you could update the context blocks for `explain set to true` and `explain set to false` to note that these represent highest and lowest verbosity respectively, that would be great and I'd be clear to merge.

dariakp: Scope creep concerns aside, the tests right now are misleading. They claim that it's important to us to test that `queryPlanner` is returned, but that's not actually what the intention is. We should make sure our test case titles and assertions capture what's important. And adding command monitoring isn't anymore effort than filing a ticket that accurately describes the work that remains to be done. Yes, from an intern point of view, this refactor was an exercise in converting callbacks to async/await, but there's no point in shuffling these tests about if they aren't going to be useful to us as a team in the codebase going forward. I don't want to check in something that is going to create more work for us in the future (in the form of wasting more engineering time trying to figure out what the tests are or aren't supposed to test). When we discussed the test refactor project with Neal, we explicitly talked about the fact that the hardest part of working with legacy tests is figuring out what the tests are supposed to do and what the correct assertions should be. If we use `queryPlanner` as a sanity check to make sure that the explain portion of the response isn't stripped, that's fine, and we can leave that assertion in with a comment, but that can't be the main and only thing we test here. Ideally we would test the driver's branching logic for the handling of explain, which is different on an api by api basis.

W-A-James: @dariakp You're right that the tests generally are either unclear or misleading, whether they use the new syntax or not. My bad for not being as discerning as I should have been with that; I was focused a bit too closely on the asyncifying part of the PR.

@malikj2000 let's go through the docs ([here](https://www.mongodb.com/docs/v6.0/reference/command/explain/#mongodb-dbcommand-dbcmd.explain) and [here](https://www.mongodb.com/docs/v6.0/reference/explain-results/)) to get a solid idea of what explain is supposed to do in the first place and get the tests to reflect that.

malikj2000: test not passing on aggregate, so will go back to lines 231 on and see what aggregate returns and adjust testing based on that

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    'deleteMany',
+    'updateOne',
+    'updateMany',
+    'distinct',
+    'findOneAndDelete',
+    'findOne',
+    'find',
+    'findOneAndReplace',
+    'aggregate'
+  ];
```
W-A-James, line 49: For ease of maintenance later on, I think it would be good to have these in a single array shaped like:

```ts
const ops = [
	{name: 'deleteOne', op: ...},
	...
]
```

So if we were to ever add or remove an operation, we wouldn't run into problems with the arrays being different sizes or having an operation be given the wrong name in the test output which could be confusing.

In the loop over the operations, any reference to the name would then reference `op.name` and the operation itself would be `op.op`

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,129 @@
...
+                expect(response[0]).to.have.property('executionStats');
+                expect(response[0]).to.not.have.nested.property('executionStats.allPlansExecution');
+              } else {
+                expect(response).to.have.property('queryPlanner');
+                expect(response).to.have.property('executionStats');
+                expect(response).to.not.have.nested.property('executionStats.allPlansExecution');
+              }
+              break;
+            default:
+              expect(response).to.be.instanceOf(MongoServerError);
```
W-A-James, line 107: Add a comment here to explain that this is for invalid values of explain

W-A-James (changes requested):
Two minor changes and we'll be good to send this back to team review :)

W-A-James (dismissed):
LGTM

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  type Collection,
+  type CommandStartedEvent,
+  type Db,
+  type MongoClient,
+  MongoServerError
+} from '../../mongodb';
+
+const explain = [true, false, 'queryPlanner', 'allPlansExecution', 'executionStats', 'invalid'];
+
+describe('Explain', function () {
```
dariakp, line 14:
```suggestion
describe('CRUD API explain option', function () {
```

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  afterEach(async function () {
+    await collection.drop();
+    await client.close();
+  });
+
+  for (const explainValue of explain) {
+    context(`When explain is ${explainValue}`, function () {
+      for (const op of ops) {
```
dariakp, line 81: suggestion to move this loop and the name declaration above the context block and update the context block to read `when explain is ${explainvalue}, operation ${name}`, so that the `it` contains a proper predicate that can focus on what we really care about behavior-wise here, namely, `it('sets command verbosity to <value> and includes <prop of interest> in the return response')`

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,186 @@
...
+          const commandStartedEvent = await commandStartedPromise;
+          switch (explainValue) {
+            case true:
+            case 'allPlansExecution':
+              expect(commandStartedEvent[0].command.verbosity).to.be.equal('allPlansExecution');
+              if (name === 'aggregate') {
+                if (response.stages) {
+                  if (response.stages[0].$cursor) {
+                    expect(response.stages[0].$cursor).to.have.property('queryPlanner');
+                    expect(response.stages[0].$cursor).nested.property(
```
dariakp, line 94: you can eliminate the duplication of the expects throughout by conditionally setting a `responsePropertyToCheck` based on the various conditions and then just writing the expectation block once (`responsePropertyToCheck` set to `response`, `response.stages`, `response[0]`, etc); you may be able to even use a conditional `? x : y` assignment, whatever looks prettier

dariakp (changes requested):

test/integration/crud/explain.test.ts
```diff
@@ -0,0 +1,173 @@
...
+                    responsePropertyToCheck = response[0].stages[0].$cursor;
+                  } else {
+                    responsePropertyToCheck = response[0].stages[0];
+                  }
+                } else {
+                  responsePropertyToCheck = response[0];
+                }
+              } else {
+                responsePropertyToCheck = response;
+              }
```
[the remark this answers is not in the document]
W-A-James: [nit] adding a comment that says which server version each branch corresponds to would be a nice to have, but isn't mandatory. If that would be too much work, a comment saying that this is meant to handle different server versions should be enough to let anyone else who ends up working on this know that if there's an issue, the server version might be something to look at.

W-A-James (changes requested):

baileympearson (dismissed):

W-A-James (approved):

dariakp (dismissed):

W-A-James (approved):

W-A-James (approved):

PR #3723, opened 2023-06-12
test(NODE-5288): sync flev2 legacy spec tests

durran:
### Description

Updates FLE2 tests to run on serverless.

#### What is changing?
- Syncs FLE2 tests to latest.
- Allows wire version checks to skip load balanced mode.

None

NODE-5288

### DIFF
--- a/src/operations/create_collection.ts
+++ b/src/operations/create_collection.ts
@@ -137,8 +137,10 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
 
       if (encryptedFields) {
         // Creating a QE collection required min server of 7.0.0
-        if (server.description.maxWireVersion < MIN_SUPPORTED_QE_WIRE_VERSION) {
-          console.log('SERVER DESCRIPTION', server.description);
+        if (
+          !server.loadBalanced &&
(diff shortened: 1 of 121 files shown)


src/operations/create_collection.ts
```diff
@@ -137,7 +137,10 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
 
       if (encryptedFields) {
         // Creating a QE collection required min server of 7.0.0
-        if (server.description.maxWireVersion < MIN_SUPPORTED_QE_WIRE_VERSION) {
+        if (
+          !server.loadBalanced &&
```
dariakp, line 141: Given the conversation in drivers, would it be better to keep the logic as is and temporarily skip the failing test? Or, alternatively, keep this implementation and file a follow up to refactor the load balancer logic to correctly check the connection wire version? Either way, some sort of follow up ticket is likely good to create and note in this PR

dariakp (changes requested):

src/operations/create_collection.ts
durran: I've created https://jira.mongodb.org/browse/NODE-5353 and updated the comment here.

dariakp (approved):

addaleax (approved):

PR #3726, opened 2023-06-14
fix(NODE-4977): load snappy lazily

nbbeeken:
#### What is changing?

- added getSnappy
- use it in the compression path so its imported lazily
- added a test to assert a small hardcoded list of deps upon importing the driver
- added a test that launches a new node process to test lazy importing
  - wrote a utility to structure this like a mocha test

Perhaps, when we've completed the related investigations we should capture out approach to optional importing and how it can effect bundling.

We only want to import an optional dependency if our library is actually using it. Snappy may be present for other reasons, and node's dependency resolution doesn't differentiate if an optional peer installed is for "our" need or for coincidentally another sibling package.


test/tools/utils.ts
```diff
@@ -499,3 +504,58 @@ export function topologyWithPlaceholderClient(
...
+  title: string,
+  fn: (d: { expect: typeof import('chai').expect; mongodb: typeof import('../mongodb') }) => void
+) {
+  it(title, async () => {
+    const script = `
+      import { expect } from 'chai';
+      import * as mongodb from './test/mongodb';
+      const run = ${fn};
+      run({ expect, mongodb }).then(
+        () => process.exit(0),
```
addaleax, line 518:
```suggestion
        () => process.exitCode = 0,
```


Just a very minor note, but often it's preferable to use `.exitCode` over `.exit(code)` because the latter immediately ends the process and wouldn't e.g. wait for unhandled rejections or for I/O to finish

test/tools/utils.ts
```diff
@@ -499,3 +504,58 @@ export function topologyWithPlaceholderClient(
...
+      scriptInstance.stdout?.addListener('data', data => {
+        stdout += data;
+      });
+
+      let stderr = '';
+      scriptInstance.stderr?.addListener('data', (data: string) => {
+        stderr += data
+          .split('\n')
+          .filter(line => !line.startsWith('Debugger') && !line.startsWith('For help'))
+          .join('\n');
```
addaleax, line 550: Also just a very minor note, but `data` chunks could also come in as `Debu`, `gger\n` in two separate events (so it might be preferable to do this filtering after all data has been read)

test/tools/utils.ts
nbbeeken: I agree, the original implementation of this launched node with `mjs` and if you `import` process, you get this error:
```txt
TypeError: Cannot assign to read only property 'exitCode' of object '[object Module]'
```

Now it uses `ts-node` and `.cts` just curious if you know of how ^ this should be done in esm?

test/tools/utils.ts
nbbeeken: Good call

test/tools/utils.ts
addaleax: Hmm, what exactly did you try originally here?

```js
$ cat test.mjs
import process from 'process'; // optional line

process.exitCode = 42;
$ node test.mjs; echo $?
42
```

seems to work fine.

It doesn’t work with `import * as process from 'process';`, but then again star-import is probably not what you want to do if you just want normal access to the `process` object, right?

addaleax:
@nbbeeken I don’t think I have the time for a full review right now :)

test/tools/utils.ts
nbbeeken: I am in the (possibly bad) habit of always using `* as` so that's why it wasn't working, TIL, TY.

src/deps.ts
```diff
@@ -111,16 +112,21 @@ type SnappyLib = {
...
-  new MongoMissingDependencyError(
-    'Optional module `snappy` not found. Please install it to enable snappy compression'
-  )
-);
-
-try {
-  // Ensure you always wrap an optional require in the try block NODE-3199
-  Snappy = require('snappy');
-} catch {} // eslint-disable-line
+export async function getSnappy(): Promise<
```
baileympearson, line 115: why are we lazy importing?

nbbeeken: By returning a promise here I'm making calling code handle either `require` or dynamic `import`, so the internal import mechanism we choose is abstracted from the caller.

baileympearson: But none of the other imports are async.  Why does snappy need to be async?  We're not using `import`s yet.

nbbeeken: removed it

durran (approved):

PR #3727, opened 2023-06-15
fix(NODE-5289): prevent scram auth from throwing TypeError if saslprep is not a function

baileympearson:


src/cmap/auth/scram.ts
```diff
@@ -2,7 +2,7 @@ import * as crypto from 'crypto';
 import { promisify } from 'util';
 
 import { Binary, type Document } from '../../bson';
-import { saslprep } from '../../deps';
+import * as deps from '../../deps';
```
nbbeeken, line 5: What's the reason behind this change?

test/integration/auth/scram_sha_256.test.ts
```diff
@@ -0,0 +1,41 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+
+import { deps, type MongoClient } from '../../mongodb';
```
nbbeeken, line 4: You should be able to just import `{ saslprep }` here b/c we have `* from deps`

test/mongodb.ts
```diff
@@ -143,6 +143,7 @@ export * from '../src/cursor/list_indexes_cursor';
 export * from '../src/cursor/run_command_cursor';
 export * from '../src/db';
 export * from '../src/deps';
+export * as deps from '../src/deps';
```
nbbeeken, line 146: This isn't really the correct way for us to maintain this file, the list is generated from the function at the top of the file and it's intended to make all internally exported functions available at the top level. (so mongodb.saslprep is exported already). So this is supposed to mirror the driver,  adding this adds a `deps` named export to our tests that doesn't exist for the driver

src/cmap/auth/scram.ts
```diff
@@ -140,7 +143,10 @@ async function continueScramConversation(
 
   let processedPassword;
   if (cryptoMethod === 'sha256') {
-    processedPassword = 'kModuleError' in saslprep ? password : saslprep(password);
+    processedPassword =
+      'kModuleError' in deps.saslprep || typeof deps.saslprep !== 'function'
```
nbbeeken, line 147: interesting, now that we validate saslprep is a function  the kModuleError check is redundant b/c if kModuleError exists the type will not be function. 

We don't have to change it, esp not with the goals here, just noting

nbbeeken (changes requested):

test/integration/auth/scram_sha_256.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+describe('SCRAM_SHA_256', function () {
+  context('when saslprep is not a function', () => {
```
nbbeeken, line 7: Should we add the happy path, just a expect(saslprep).to.have.been.called when it is a function

nbbeeken (changes requested):

src/cmap/auth/scram.ts
nbbeeken: Slack: Needed for mocking, b/c otherwise this file gets a reference to saslprep before sinon can stub it.

Maybe it's worth leaving a comment that the import is intentionally different than our others.

test/integration/auth/scram_sha_256.test.ts
baileympearson: Since we have no SCRAM tests, I chose to only write regression tests.  What value do you think that test provides, in the context of this bug fix?

nbbeeken: its testing the inverse of the condition, real scram tests would cover that saslprep is being used / works as expected, I'm just looking to cover the code change here.

nbbeeken (changes requested):

test/mongodb.ts
nbbeeken:
```suggestion
```

Now that we have it in the test file

nbbeeken (changes requested):

test/integration/auth/scram_sha_256.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    afterEach(() => client.close());
+
+    it('calls saslprep', { requires: { auth: 'enabled' } }, async function () {
+      const spy = sinon.spy(deps, 'saslprep');
+
+      await client.connect();
+
+      expect(spy.called).to.be.true;
```
nbbeeken, line 61:
```suggestion
      expect(spy).to.have.been.called;
```

use sinon-chai here?

nbbeeken (changes requested):

nbbeeken (approved):

PR #3729, opened 2023-06-15
refactor(NODE-5352): refactor AbstractOperation to use async

malikj2000:
#### What is changing?
created callback
subclass of operation
and changed execute
no
first step in asyncifying the operations layer

### DIFF
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -25,13 +25,13 @@ import {
 import type { Topology } from '../sdam/topology';
 import type { ClientSession } from '../sessions';
 import { type Callback, maybeCallback, supportsRetryableWrites } from '../utils';
-import { AbstractOperation, Aspect } from './operation';
+import { AbstractCallbackOperation, Aspect } from './operation';
 
 const MMAPv1_RETRY_WRITES_ERROR_CODE = MONGODB_ERROR_CODES.IllegalOperation;
 const MMAPv1_RETRY_WRITES_ERROR_MESSAGE =
   'This MongoDB deployment does not support retryable writes. Please add retryWrites=false to your connection string.';
 
-type ResultTypeFromOperation<TOperation> = TOperation extends AbstractOperation<infer K>
+type ResultTypeFromOperation<TOperation> = TOperation extends AbstractCallbackOperation<infer K>
   ? K
   : never;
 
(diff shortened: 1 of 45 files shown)


src/operations/operation.ts
```diff
@@ -121,6 +117,25 @@ export abstract class AbstractOperation<TResult = any> {
   }
 }
 
+/** @internal */
+export abstract class AbstractCallbackOperation<TResult = any> extends AbstractOperation {
```
baileympearson, line 121:
```suggestion
export abstract class AbstractCallbackOperation<TResult> extends AbstractOperation<TResult> {
```

baileympearson (changes requested):
minor comments

src/operations/operation.ts
```diff
@@ -121,6 +105,25 @@ export abstract class AbstractOperation<TResult = any> {
   }
 }
 
+/** @internal */
+export abstract class AbstractCallbackOperation<TResult> extends AbstractOperation<TResult> {
```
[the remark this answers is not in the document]
W-A-James: Once you get it to build, you're probably going to see some test failures that I flagged below this.

W-A-James (changes requested):

src/operations/operation.ts
W-A-James: Why remove the default generic parameter?

baileympearson: Because it's already defaulted in `AbstractOperation`.

W-A-James: In that case I think we should keep it here and remove it from `AbstractOperation` because TypeScript doesn't understand that we have it defaulted for the subclasses of `AbstractCallbackOperation` that don't explicitly provide the parameter when they are declared so we're getting TypeScript errors.

src/operations/execute_operation.ts
```diff
@@ -25,7 +25,7 @@ import {
 import type { Topology } from '../sdam/topology';
 import type { ClientSession } from '../sessions';
 import { type Callback, maybeCallback, supportsRetryableWrites } from '../utils';
-import { AbstractOperation, Aspect } from './operation';
+import { AbstractCallbackOperation, type AbstractOperation, Aspect } from './operation';
```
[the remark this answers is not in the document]
malikj2000: Yeah we missed one, I changed it

src/operations/execute_operation.ts
```diff
@@ -25,13 +25,13 @@ import {
...
 import { type Callback, maybeCallback, supportsRetryableWrites } from '../utils';
-import { AbstractOperation, Aspect } from './operation';
+import { AbstractCallbackOperation, Aspect } from './operation';
 
 const MMAPv1_RETRY_WRITES_ERROR_CODE = MONGODB_ERROR_CODES.IllegalOperation;
 const MMAPv1_RETRY_WRITES_ERROR_MESSAGE =
   'This MongoDB deployment does not support retryable writes. Please add retryWrites=false to your connection string.';
 
-type ResultTypeFromOperation<TOperation> = TOperation extends AbstractOperation<infer K>
+type ResultTypeFromOperation<TOperation> = TOperation extends AbstractCallbackOperation<infer K>
```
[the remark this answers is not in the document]
W-A-James: @malikj2000 my mistake on asking you to change that then.

W-A-James (changes requested):
@malikj2000 There are test failures to address

test/unit/cmap/connection.test.ts
```diff
@@ -72,7 +73,7 @@ class FakeSocket extends EventEmitter {
   }
 }
 
-class InputStream extends Readable {
+class InputStream extends Duplex {
```
W-A-James, line 76: For my own understanding, why was this changed?

test/unit/sdam/server.test.ts
```diff
@@ -64,6 +64,25 @@ describe('Server', () => {
...
+          {} as any
+        );
+        serverSpy = sinon.spy(server, 'command');
+      });
+
+      it('calls the command function through commandAsync', async function () {
+        await server.commandAsync();
+        expect(serverSpy).to.have.been.calledOnce;
+      });
+    });
```
W-A-James, line 84: This test is failing.

W-A-James (changes requested):
In addition to the comments above, there are also some failing unit tests for operations that subclass `AbstractCallbackOperation`

test/unit/cmap/connection.test.ts
baileympearson: Malik and I changed this debugging, but now this is unnecessary.  We should revert any changes that aren't necessary for this PR

test/unit/operations/find.test.ts
```diff
@@ -43,7 +43,7 @@ describe('FindOperation', function () {
       it('should build basic find command with filter', async () => {
         const findOperation = new FindOperation(undefined, namespace, filter);
         const stub = sinon.stub(server, 'command').yieldsRight();
-        await promisify(findOperation.execute.bind(findOperation))(server, undefined);
+        await promisify(findOperation.executeCallback.bind(findOperation))(server, undefined);
```
W-A-James, line 46: You can get rid of the `promisify` here instead of replacing `execute` with `executeCallback`. Remember that what `promisify` does is take an asynchronous function that returns its results via callback and changes it into one that returns a promise instead. That's exactly what you did with `execute` itself, so you can just await that directly.

The same goes for all the other unit tests like this below.

W-A-James (changes requested):

W-A-James (dismissed):

baileympearson (changes requested):

baileympearson (approved):
waiting on CI after the rebase before merging but otherwise LGTM

PR #3737, opened 2023-06-21
docs(NODE-5368): fix includeResultMetadata docs

durran:
### Description

Fixes the docs around the `includeResultMetadata` option.

#### What is changing?

Updates the docs on the options.

None

NODE-5368

## findOneAndModify* helpers can now return only the affected document, not the server result

A new flag, `includeResultMetadata` has been added to `Collection.findOneAnd*` helpers.  This flag defaults to true.  When disabled, the collection helpers only return the modified document.  For example:

```typescript
const deletedDocument = collection.findOneAndDelete({ _id: <id> }, { includeResultMetadata: false });
// deletedDocument contains the deleted document from the server, not a `ModifyResult`
```

### DIFF
--- a/src/operations/find_and_modify.ts
+++ b/src/operations/find_and_modify.ts
@@ -29,7 +29,10 @@ export interface FindOneAndDeleteOptions extends CommandOperationOptions {
   sort?: Sort;
   /** Map of parameter names and values that can be accessed using $$var (requires MongoDB 5.0). */
   let?: Document;
-  /** Return the ModifyResult instead of the modified document. Defaults to true. */
+  /**
+   * Return the ModifyResult instead of the modified document. Defaults to true
+   * but will default to false in the next major version.
+   */
   includeResultMetadata?: boolean;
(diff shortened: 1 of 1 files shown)


src/operations/find_and_modify.ts
```diff
@@ -29,7 +29,7 @@ export interface FindOneAndDeleteOptions extends CommandOperationOptions {
   sort?: Sort;
   /** Map of parameter names and values that can be accessed using $$var (requires MongoDB 5.0). */
   let?: Document;
-  /** Return the raw result document instead of the ModifyResult */
+  /** Return the ModifyResult instead of the modified document. Defaults to true. */
```
dariakp, line 32: Can we add a note that this will default to false in the next major version?

durran: Updated.

dariakp (approved):

baileympearson (approved):

PR #3742, opened 2023-06-23
fix(NODE-5102): fix listDatabases nameOnly setting is sent as NaN

redixhumayun:
### Description
Fixes the issue described in NODE-5102 where `nameOnly` was being coerced into `NaN` using the `Number` constructor. 

#### What is changing?
Now, the `nameOnly` property is coerced to a boolean following the example laid out by `listCollections`. Accompanying unit tests have also been added
Also, the CommandOperation class required an optional chaining operator when reading the properties of the parent object otherwise it throws an error when the new unit tests are run. 

No

Fixes NODE-5102


src/operations/list_databases.ts
```diff
@@ -31,6 +31,7 @@ export class ListDatabasesOperation extends CommandOperation<ListDatabasesResult
   constructor(db: Db, options?: ListDatabasesOptions) {
     super(db, options);
     this.options = options ?? {};
+    this.options.nameOnly = !!this.options.nameOnly;
```
W-A-James, line 34:
```suggestion
    this.options.nameOnly = this.options.nameOnly ?? false;
```

src/operations/command.ts
```diff
@@ -86,7 +86,7 @@ export abstract class CommandOperation<T> extends AbstractOperation<T> {
       this.ns = new MongoDBNamespace(dbNameOverride, '$cmd');
     } else {
       this.ns = parent
-        ? parent.s.namespace.withCollection('$cmd')
+        ? parent.s?.namespace.withCollection('$cmd')
```
W-A-James, line 89: Why add optional chaining for the `s` field?

test/unit/operations/list_databases.test.js
W-A-James, line 1: When creating new test files, we have a strong preference for typescript, so please convert this to typescript.

W-A-James (changes requested):
Hi @redixhumayun, thanks so much for this PR. I have left some specific comments on the files above. In addition to the unit tests you have written, before we can approve this change, we would need to have integration tests that would test this behaviour against a mongod instance. It would be good to create a new file for these in `test/integration/enumerate_databases.test.ts`.

Please see the [running tests locally section](https://github.com/mongodb/node-mongodb-native/blob/main/test/readme.md#running-the-tests-locally) section of our test readme for details on how to get integration tests running locally.

For an example of how to write an integration test in this repo, please see the [findAndModify tests](https://github.com/mongodb/node-mongodb-native/blob/d9c2600daf3bb6c8106e0f3a27cb2f12b770c8eb/test/integration/crud/find_and_modify.test.ts).

W-A-James:
@redixhumayun We do want to merge this change fairly soon, so we will take over the remaining code changes on this PR. Thanks again for your submission!

test/unit/operations/list_databases.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+    }
+  }
+};
+
+describe('ListDatabasesOperation', function () {
+  describe('#constructor', function () {
+    context('when nameOnly is provided', function () {
+      context('when nameOnly is true', function () {
+        const operation = new ListDatabasesOperation(mockDB, { nameOnly: true });
```
baileympearson, line 19: nit: don't initialize variables outside of before hooks or tests.

Even when a test is skipped, any code outside the test blocks or hooks runs when the suite is loaded.  Consequently, if we ever need to run the debugger on a function that is called frequently outside of tests like this, we can't easily rely on just dropping breakpoints inside the relevant function, because we'll hit the same breakpoint for tests that are skipped.

baileympearson (changes requested):

test/integration/enumerate_databases.test.ts
```diff
@@ -141,6 +142,64 @@ describe('listDatabases()', function () {
...
+    function optionToExpecation(nameOnly?: boolean): string {
+      switch (nameOnly) {
+        case true:
+          return 'with nameOnly = true';
+        case false:
+          return 'with nameOnly = false';
+        case undefined:
+          return 'without nameOnly field';
+      }
+    }
```
baileympearson, line 201: optional - functions like this that just map one value to another I think are better expressed as maps.  they're effectively the same, so feel free to ignore.

```suggestion
    const optionToExpecation = { 
      true: 'with nameOnly = true',
      false: 'with nameOnly = false',
      undefined: 'without nameOnly field'
   }
   
   optionToExpectation(String(<option>))
```

test/integration/enumerate_databases.test.ts
```diff
@@ -141,6 +142,64 @@ describe('listDatabases()', function () {
...
+  describe('nameOnly option', function () {
+    let client: MongoClient;
+    const DBS = 10;
+    const nameOnlyOptions = [true, false, undefined];
+
+    beforeEach(async function () {
+      client = await this.configuration.newClient({}, { monitorCommands: true }).connect();
+      for (let i = 0; i < DBS; i++) {
+        const db = client.db(`testDb_${i}`);
+        await db.collection('test').insertOne({ a: 1 });
```
baileympearson, line 154: Are we inserting documents only to create the collection?  

```suggestion
await client.db(<>).createCollection('test');
```

baileympearson (changes requested):

test/integration/enumerate_databases.test.ts
```diff
@@ -141,6 +142,51 @@ describe('listDatabases()', function () {
...
+            case true:
+              expect(commandStarted.command).to.have.property('nameOnly', true);
+              break;
+            case false:
+              expect(commandStarted.command).to.have.property('nameOnly', false);
+              break;
+            case undefined:
+              expect(commandStarted.command).to.not.have.property('nameOnly');
+              break;
+          }
```
nbbeeken, line 185: Needs a `default: expect.fail` case to cover all paths if we want to stick with this control flow format.

nbbeeken (changes requested):

nbbeeken (approved):

src/operations/command.ts
redixhumayun: @W-A-James Hey Warren, the reason for the optional chaining here is that the tests for `list_databases.test.js` fail if the optional chaining is not present on that line. 

There is an error thrown indicating that `parent.s` is undefined

Below is the error message

```
TypeError: Cannot read properties of undefined (reading 'namespace')
    at new CommandOperation (/Users/zaidhumayun/Desktop/Development.nosync/JavaScript/node-mongodb-native-redixhumayun/src/operations/command.ts:89:20)
```

test/unit/operations/list_databases.test.js
redixhumayun: @W-A-James Okay, I created a `.js` file because the tests file I referenced `list_collection.test.js` also was written in JS. I can change it to TS if you need me to

baileympearson (approved):

W-A-James (approved):

src/operations/command.ts
W-A-James: @redixhumayun Hey Zaid, this was resolved by updating `test/unit/operations/list_databases.test.ts:5` to include a mock database object. This was not a bug in the source itself.

PR #3746, opened 2023-06-27
fix(NODE-5374): do not apply cursor transform in Cursor.hasNext

baileympearson:
#### What is changing?

Primarily this PR fixes two bugs in our AbstractCursor.

- transform should not be applied when using `hasNext`.
- when transform throws, we should catch the error and propagate it to the user.

To facilitate this, the generic `next<T>` was refactored to use async-await as well.  

The PR is broken down into commits which might be easier to view.  The first commits are the async-await refactor, then the bug fixes.

No.

### Cursor.map transform bug fixes 

The cursor API provides the ability to apply a `map` function to each document in the cursor:

```typescript
const cursor = collection.find({ name: 'john doe' }).map(({ name }) => name);
for await (const document of cursor) {
  console.error(document); // only prints the `name` field from each document
}
```

#### Cursor iteration properly catches errors thrown from the transform applied in `Cursor.map`

Starting in version 4.0 of the driver, if the transform function throws an error, there are certain scenarios where the driver does not correctly catch this error and an uncaught exception is thrown:

```typescript
const cursor = collection.find({ name: 'john doe' }).map(() => {
  throw new Error('oh no! error here'); // 
});
await cursor.next(); // process crashes with uncaught error
```

This release adds logic to ensure that whenever we transform a cursor document, we handle any errors properly.  Any errors thrown from a transform function are caught and returned to the user.

#### Cursor.hasNext no longer transforms documents if a transform has been applied to the cursor

Version 4.0 introduced a bug that would apply a `transform` function to documents in the cursor when the cursor was iterated using `Cursor.hasNext()`.  When combined with `Cursor.next()`, this would result in transforming documents multiple times.

```typescript
const cursor = collection.find({ name: 'john doe' }).map((document) => document.name);
while (await cursor.hasNext()) { // this transforms the first document in the cursor once
	const doc = await cursor.next(); // the second document in the cursor is transformed again
}
```

This release removes the transform logic from `Cursor.hasNext`, preventing cursor documents from being transformed twice when iterated using `hasNext`.


src/cursor/abstract_cursor.ts
```diff
@@ -680,88 +680,107 @@ export abstract class AbstractCursor<
...
     // All cursors must operate within a session, one must be made implicitly if not explicitly provided
-    cursor[kInit](err => {
-      if (err) return callback(err);
-      return next(cursor, blocking, callback);
-    });
-
-    return;
+    const init = promisify(cb => cursor[kInit](cb));
+    await init();
+    return next(cursor, blocking, transform);
```
baileympearson, line 726: We could also promisify on the AbstractCursor class, but if we do, I'd like to consider an approach similar to how we're handling `AbstractOperation` so that we can eventually migrate cursors over to being async one at a time.  I can tackle that in this PR if we like or we can leave it as-is.

test/integration/crud/misc_cursors.test.js
```diff
@@ -3589,7 +3584,7 @@ describe('Cursor', function () {
     await client.close();
   });
 
-  it('should return implicit session to pool when client-side cursor exhausts results after a getMore', function (done) {
+  it('should return implicit session to pool when client-side cursor exhausts results after a getMore', async function () {
```
baileympearson, line 3587: refactored this test while debugging it but it's not necessary.  we may as well keep it, since it's one more async-await test.

test/integration/node-specific/abstract_cursor.test.ts
```diff
@@ -1,114 +1,264 @@
...
+        const [doc] = await once(cursor, 'data');
+        expect(transformSpy).to.have.been.calledOnce;
+        expect(doc.name).to.equal('JOHN DOE');
       });
-    }
 
-    it('throws when mapping to `null` and cleans up cursor', async function () {
-      const cursor = collection.find();
-      cursor.map(() => null);
+      // skipped because these tests fail after throwing uncaught exceptions
```
baileympearson, line 120:
```suggestion
```

test/integration/crud/misc_cursors.test.js
```diff
@@ -3604,25 +3599,22 @@ describe('Cursor', function () {
...
-                client.close(done);
-              });
-            });
-          });
-        });
-      });
-    });
+    await collection.insertMany(docs);
+
+    // TODO - talk to Neal about this test
```
baileympearson, line 3604:
```suggestion
```

test/integration/node-specific/cursor_stream.test.js
```diff
@@ -297,10 +297,10 @@ describe('Cursor Streams', function () {
         stream.on('error', err => (error = err));
         cursor.on('close', function () {
           // NOTE: use `setImmediate` here because the stream implementation uses `nextTick` to emit the error
-          setImmediate(() => {
+          setTimeout(() => {
```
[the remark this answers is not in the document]
baileympearson: Yeah, same actually.  I think this test might have failed while I was refactoring `next<T>` and I assumed it was a timing issue.

test/integration/node-specific/abstract_cursor.test.ts
```diff
@@ -1,114 +1,263 @@
...
+          const cursor = collection.find().map(() => {
+            throw new Error('error thrown in transform');
+          });
+
+          const error = await func(cursor).catch(e => e);
+          expect(error)
+            .to.be.instanceOf(Error)
+            .to.match(/error thrown in transform/);
+          expect(cursor.closed).to.be.true;
+        });
```
durran, line 109:
```suggestion
        describe(`#${method}`, function () {
          context('when the transform succeeds', function () {
            it('calls the cursor transform', async () => {
              const cursor = collection.find().map(transformSpy);
    
              const doc = await func(cursor);
              expect(transformSpy).to.have.been.calledOnce;
              expect(doc.name).to.equal('JOHN DOE');
            });
          });
  
          context('when the transform throws', function () {
            it('propagates the error to the user', async () => {
              const cursor = collection.find().map(() => {
                throw new Error('error thrown in transform');
              });
    
              const error = await func(cursor).catch(e => e);
              expect(error)
                .to.be.instanceOf(Error)
                .to.match(/error thrown in transform/);
              expect(cursor.closed).to.be.true;
            });
          });
        });
```

test/integration/node-specific/abstract_cursor.test.ts
```diff
@@ -1,114 +1,263 @@
...
+      }
+
+      it('Cursor.stream() calls the cursor transform when iterated', async function () {
```
durran, line 112: Same is the above comment, but Github's diffs are making it problematic for the rest... Can we change the test structures to match as an example:

```ts
describe('#methodName", function () {
  context('when some condition exists', function () {
    it('does/returns/throws something', function () {
      // assertion
    });
  });
});
```

src/cursor/abstract_cursor.ts
```diff
@@ -680,88 +680,107 @@ export abstract class AbstractCursor<
...
+ * @returns the next document in the cursor, or `null`.  When `blocking` is `true`, a `null` document means
+ * the cursor has been exhausted.  Otherwise, it means that there is no document available in the cursor's buffer.
  */
-export function next<T>(
+async function next<T>(
   cursor: AbstractCursor<T>,
   blocking: boolean,
-  callback: Callback<T | null>
-): void {
+  transform = true
```
durran, line 697: I assume we are defaulting to `true` here as more methods would transform vs. methods that don't (`hasNext`)?

durran (changes requested):

src/cursor/abstract_cursor.ts
durran: I think it's probably fine to not tackle that in this PR.

test/integration/crud/misc_cursors.test.js
```diff
@@ -3589,7 +3584,7 @@ describe('Cursor', function () {
     await client.close();
   });
 
-  it('should return implicit session to pool when client-side cursor exhausts results after a getMore', function (done) {
+  it('should return implicit session to pool when client-side cursor exhausts results after a getMore', async function () {
     const configuration = this.configuration;
     const client = configuration.newClient({ w: 1 }, { maxPoolSize: 1 });
```
nbbeeken, line 3589: This file has a shared client the test could be updated to use

test/integration/node-specific/abstract_cursor.test.ts
```diff
@@ -1,114 +1,263 @@
...
 import { inspect } from 'util';
 
-import { type Collection, MongoAPIError, type MongoClient } from '../../mongodb';
-
-const falseyValues = [0, 0n, NaN, '', false, undefined];
+import { type Collection, type FindCursor, MongoAPIError, type MongoClient } from '../../mongodb';
 
 describe('class AbstractCursor', function () {
-  let client: MongoClient;
+  describe('regression tests NODE-5372', function () {
```
nbbeeken, line 9: We can mention the ticket in a comment but can we make the title about not double invoking transform

src/cursor/abstract_cursor.ts
baileympearson: Yeah, that was my logic.  I also considered two other options:

- providing no default
- combining `blocking` and `transform` into an options object and requiring all callsites to explicitly set them for clarity (i.e., `next(this, { blocking: true, transform: false })`

test/integration/crud/misc_cursors.test.js
baileympearson: Done.

test/integration/node-specific/abstract_cursor.test.ts
baileympearson: I've refactored the tests for both comments.

src/cursor/abstract_cursor.ts
nbbeeken: I lean towards providing no default, for internal I prefer to make each call site define its inputs and because this isn't saving us a lot of refactors. I don't feel strongly though, feel free to resolve if we agree to leave it as is.

nbbeeken (dismissed):

durran (approved):

PR #3752, opened 2023-07-03
feat(NODE-5274): deprecate write concern options

durran:
### Description

Deprecates old write concern options.

#### What is changing?
- Deprecates `wtimeout`, `j`, and `fsync` options.
- Adds new `WriteConcern.apply()` method to add a write concern to a command.
- Updates instances of adding write concern to commands to use the new apply method.

None

NODE-5274

#### Write Concern legacy options deprecated

`wtimeout`, `j`, and `fsync` options have been deprecated, please use `wtimeoutMS` and `journal` instead.

### DIFF
--- a/src/gridfs/upload.ts
+++ b/src/gridfs/upload.ts
@@ -458,8 +458,8 @@ function getWriteOptions(stream: GridFSBucketWriteStream): WriteConcernOptions {
   if (stream.writeConcern) {
     obj.writeConcern = {
       w: stream.writeConcern.w,
-      wtimeout: stream.writeConcern.wtimeout,
-      j: stream.writeConcern.j
+      wtimeout: stream.writeConcern.wtimeoutMS,
+      j: stream.writeConcern.journal
     };
   }
   return obj;
--- a/src/gridfs/upload.ts
+++ b/src/gridfs/upload.ts
@@ -263,9 +263,8 @@ async function checkChunksIndex(stream: GridFSBucketWriteStream): Promise<void> 
   });
 
   if (!hasChunksIndex) {
-    const writeConcernOptions = getWriteOptions(stream);
     await stream.chunks.createIndex(index, {
-      ...writeConcernOptions,
+      ...stream.writeConcern,
       background: true,
       unique: true
     });
@@ -292,7 +291,7 @@ function checkDone(stream: GridFSBucketWriteStream, callback?: Callback): boolea
       return false;
     }
 
-    stream.files.insertOne(filesDoc, getWriteOptions(stream)).then(
+    stream.files.insertOne(filesDoc, { writeConcern: stream.writeConcern }).then(
       () => {
         stream.emit(GridFSBucketWriteStream.FINISH, filesDoc);
         stream.emit(GridFSBucketWriteStream.CLOSE);
@@ -423,7 +422,7 @@ function doWrite(
         return false;
       }
 
-      stream.chunks.insertOne(doc, getWriteOptions(stream)).then(
+      stream.chunks.insertOne(doc, { writeConcern: stream.writeConcern }).then(
         () => {
           --stream.state.outstandingRequests;
           --outstandingRequests;
@@ -451,18 +450,6 @@ function doWrite(
   // to be compatible with node's `.pipe()` function.
   // False means the client should wait for the 'drain' event.
   return false;
-}
-
-function getWriteOptions(stream: GridFSBucketWriteStream): WriteConcernOptions {
-  const obj: WriteConcernOptions = {};
-  if (stream.writeConcern) {
-    obj.writeConcern = {
-      w: stream.writeConcern.w,
-      wtimeout: stream.writeConcern.wtimeoutMS,
-      j: stream.writeConcern.journal
-    };
-  }
-  return obj;
 }
 
 function waitForIndexes(
@@ -499,7 +486,7 @@ function writeRemnant(stream: GridFSBucketWriteStream, callback?: Callback): boo
     return false;
   }
 
-  stream.chunks.insertOne(doc, getWriteOptions(stream)).then(
+  stream.chunks.insertOne(doc, { writeConcern: stream.writeConcern }).then(
     () => {
       --stream.state.outstandingRequests;
       checkDone(stream);
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -698,7 +698,7 @@ function endTransaction(
   // apply a writeConcern if specified
   let writeConcern;
   if (session.transaction.options.writeConcern) {
-    writeConcern = WriteConcern.apply({}, session.transaction.options.writeConcern);
+    writeConcern = Object.assign({}, session.transaction.options.writeConcern);
   } else if (session.clientOptions && session.clientOptions.writeConcern) {
     writeConcern = { w: session.clientOptions.writeConcern.w };
   }
@@ -708,9 +708,10 @@ function endTransaction(
   }
 
   if (writeConcern) {
-    Object.assign(command, { writeConcern });
+    WriteConcern.apply(command, writeConcern);
   }
 
+  console.log(command);
   if (commandName === 'commitTransaction' && session.transaction.options.maxTimeMS) {
     Object.assign(command, { maxTimeMS: session.transaction.options.maxTimeMS });
   }
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -773,9 +773,9 @@ function endTransaction(
 
           WriteConcern.apply(
             command,
-            Object.assign({ wtimeoutMS: 10000 }, command.writeConcern, {
-              w: 'majority'
-            })
+            WriteConcern.fromOptions(
+              Object.assign({ wtimeout: 10000 }, command.writeConcern, { w: 'majority' })
+            ) || {}
           );
         }
 
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -45,6 +45,7 @@ import {
   now,
   uuidV4
 } from './utils';
+import { WriteConcern } from './write_concern';
 
 const minWireVersionForShardedTransactions = 8;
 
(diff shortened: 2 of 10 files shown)


src/write_concern.ts
```diff
@@ -1,3 +1,5 @@
+import { type Document } from 'bson';
```
nbbeeken, line 1:
```suggestion
import { type Document } from './bson';
```

src/write_concern.ts
```diff
@@ -35,41 +56,66 @@ export const WRITE_CONCERN_KEYS = ['w', 'wtimeout', 'j', 'journal', 'fsync'];
...
    * Constructs a WriteConcern from the write concern properties.
    * @param w - request acknowledgment that the write operation has propagated to a specified number of mongod instances or to mongod instances with specified tags.
-   * @param wtimeout - specify a time limit to prevent write operations from blocking indefinitely
-   * @param j - request acknowledgment that the write operation has been written to the on-disk journal
+   * @param wtimeoutMS - specify a time limit to prevent write operations from blocking indefinitely
+   * @param journal - request acknowledgment that the write operation has been written to the on-disk journal
    * @param fsync - equivalent to the j option
    */
-  constructor(w?: W, wtimeout?: number, j?: boolean, fsync?: boolean | 1) {
+  constructor(w?: W, wtimeoutMS?: number, journal?: boolean, fsync?: boolean | 1) {
```
nbbeeken, line 88: Will the fsync parameter be removed too?

src/write_concern.ts
```diff
@@ -35,41 +56,66 @@ export const WRITE_CONCERN_KEYS = ['w', 'wtimeout', 'j', 'journal', 'fsync'];
...
+  /**
+   * Apply a write concern to a command document.
+   */
+  static apply(command: Document, writeConcern: WriteConcern): Document {
+    const wc: CommandWriteConcernOptions = {};
+    // The write concern document sent to the server has w/wtimeout/j fields.
+    if (writeConcern.w != null) wc.w = writeConcern.w;
+    if (writeConcern.wtimeoutMS != null) wc.wtimeout = writeConcern.wtimeoutMS;
+    if (writeConcern.journal != null) wc.j = writeConcern.j;
+    return Object.assign(command, { writeConcern: wc });
```
nbbeeken, line 116:
```suggestion
    command.writeConcern = wc;
    return command;
```

`Object.assign` is useful when there's many keys to stick on to an existing object, since we just have the one here, might as well.

src/write_concern.ts
```diff
@@ -35,41 +56,66 @@ export const WRITE_CONCERN_KEYS = ['w', 'wtimeout', 'j', 'journal', 'fsync'];
...
+      this.journal = this.j = journal;
     }
     if (fsync != null) {
-      this.fsync = fsync;
+      this.journal = this.j = fsync ? true : false;
     }
   }
 
+  /**
+   * Apply a write concern to a command document.
```
nbbeeken, line 108:
```suggestion
   * Apply a write concern to a command document. Will modify and return the command.
```
Just a bit more docs, and double checking this is supposed to be public correct? 

I also notice we don't use the return value, just the modification, should we pick one instead? (also fine with both, similar to `Object.assign` behavior, just asking the Q)

test/unit/write_concern.test.ts
```diff
@@ -0,0 +1,106 @@
+import { expect } from 'chai';
+
+import { WriteConcern } from '../mongodb';
+
+describe('WriteConcern', function () {
+  describe('#constructor', function () {
+    context('when w is provided', function () {
```
nbbeeken, line 7: Can we add a case for w being a string of digits `'234'` should work b/c of the `!Number.isNaN(Number(w))` case we have in the ctor

I guess we should also at least make sure majority works, but I think the intention is that any string could be provided if other `w` settings were introduced in the future.

nbbeeken (commented):
Just wanted to get you some comments for tmo even though this is draft

test/unit/write_concern.test.ts
durran: Add tests for string number and majority.

src/write_concern.ts
durran: Updated to suggestion.

src/write_concern.ts
durran: Yes it will get removed.

src/write_concern.ts
durran: Updated.

src/write_concern.ts
nbbeeken: Maybe we should mention it in the docs for that option, I don't think the deprecation tag can be applied to a param, but at least there would be mention of it

durran: Updated.

test/unit/write_concern.test.ts
```diff
@@ -0,0 +1,124 @@
...
+          writeConcern: { w: 2, wtimeout: 30, j: true }
+        });
+      });
+    });
+
+    context('when fsync is in the write concern', function () {
+      const document = {};
+      const writeConcern = new WriteConcern(2, 30, true, false);
+
+      it('overrites j to the write concern document', function () {
```
nbbeeken, line 117:
```suggestion
      it('overwrites j to the write concern document', function () {
```

src/sessions.ts
```diff
@@ -703,11 +704,11 @@ function endTransaction(
...
-    writeConcern = Object.assign({ wtimeout: 10000 }, writeConcern, { w: 'majority' });
+    writeConcern = Object.assign({ wtimeoutMS: 10000 }, writeConcern, { w: 'majority' });
   }
 
   if (writeConcern) {
-    Object.assign(command, { writeConcern });
+    WriteConcern.apply(command, writeConcern);
   }
 
   if (commandName === 'commitTransaction' && session.transaction.options.maxTimeMS) {
```
nbbeeken, line 714: On line 774 in the retryable write handling there's more WriteConcern applying

src/sessions.ts
```diff
@@ -703,11 +704,11 @@ function endTransaction(
   }
 
   if (txnState === TxnState.TRANSACTION_COMMITTED) {
-    writeConcern = Object.assign({ wtimeout: 10000 }, writeConcern, { w: 'majority' });
+    writeConcern = Object.assign({ wtimeoutMS: 10000 }, writeConcern, { w: 'majority' });
```
nbbeeken, line 707: Should we call `WriteConcern.fromOptions` here? and above as well, These Object.assign calls would make something that's an instance of the WriteConcern class turn into a plain object.

src/gridfs/upload.ts
```diff
@@ -458,8 +458,8 @@ function getWriteOptions(stream: GridFSBucketWriteStream): WriteConcernOptions {
   if (stream.writeConcern) {
     obj.writeConcern = {
       w: stream.writeConcern.w,
-      wtimeout: stream.writeConcern.wtimeout,
-      j: stream.writeConcern.j
+      wtimeout: stream.writeConcern.wtimeoutMS,
```
nbbeeken, line 461: I think we should take the fix here a bit further, we're passing the legacy options into our own APIs here. The stream holds a WriteConcern instance already, I don't think we need any reshaping here, just pass it along

nbbeeken (changes requested):

src/sessions.ts
durran: It's expected in the sessions module that the `writeConcern` is not an instance of a `WriteConcern` object since it's going to be put directly into the options. That worked previously is it was an instance since the `WriteConcern` properties matched 1-1 with what was valid to pass to the server, but this is no longer the case. `wtimeoutMS` and `journal` are not valid (they must be `wtimeout` and `j` when passed to the server). The reason for the change on this line is that later when we call `WriteConcern.apply()`, which always happens if a write concern exists, it's looking for our non-deprecated client facing option names.

src/gridfs/upload.ts
durran: Updated to pass through. The `getWriteOptions` method is now removed.

src/sessions.ts
durran: Changed to use apply.

test/unit/write_concern.test.ts
durran: Fixed.

src/sessions.ts
durran: I think it's cleaner here to do it this way here as `writeConcern` in this context (from line 699) is not an instance of a `WriteConcern` object but rather a plain object.

nbbeeken (approved):

test/integration/crud/find_and_modify.test.ts
```diff
@@ -119,36 +119,36 @@ describe('Collection (#findOneAnd...)', function () {
         });
 
         it('passes through the writeConcern', async function () {
-          await collection.findOneAndDelete({}, { writeConcern: { fsync: 1 } });
-          expect(started[0].command.writeConcern).to.deep.equal({ fsync: 1 });
+          await collection.findOneAndDelete({}, { writeConcern: { j: 1 } });
+          expect(started[0].command.writeConcern).to.deep.equal({ j: 1 });
```
W-A-James, line 123: Just to double check, the server treats `fsync` and `j` as synonyms, correct?

test/unit/write_concern.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+      context('when w is a string number', function () {
+        const writeConcern = new WriteConcern('10');
+
+        it('sets the w property to a number', function () {
+          expect(writeConcern.w).to.equal(10);
+        });
+      });
+
+      context('when w is a string', function () {
```
W-A-James, line 24: Do we have any behaviour defined for when `w` is an arbitrary string? I can see it being a possibility that people consuming the library from regular JavaScript that we silently do something different from what they expected if they were to make a typo like `'majorit'` instead of `'majority'`.

test/integration/crud/find_and_modify.test.ts
durran: Yes, which is why we're removing fsync in 6.0

test/unit/write_concern.test.ts
durran: The server will return a write concern error if it is a string and not 'majority'.

W-A-James (approved):

PR #3767, opened 2023-07-10
refactor(NODE-5392): refactor cursor operations to use async/await

malikj2000:
### Description
refactored cursor operations to use async/await instead of callbacks
execute_operation extends AbstractOperation
cleaned up collections tests

#### What is changing?
more operations are using async/await now

to further improve async stack traces


src/operations/command.ts
```diff
@@ -111,7 +111,7 @@ export abstract class CommandOperation<T> extends AbstractCallbackOperation<T> {
     server: Server,
     session: ClientSession | undefined,
     cmd: Document
-  ): Promise<Document> {
+  ): Promise<any> {
```
baileympearson, line 114:
```suggestion
  ): Promise<T> {
```
Is there a reason we can't use the generic type?  Operations are parameterized by their return types.

src/operations/command.ts
```diff
@@ -111,7 +111,7 @@ export abstract class CommandOperation<T> extends AbstractCallbackOperation<T> {
...
     // TODO: consider making this a non-enumerable property
     this.server = server;
```
baileympearson, line 117: I can't comment on it because it's hidden in the diff but I think the inheritence hierarchy is off.  `CommandOperation` should not inherit from `AbstractCallbackOperation` - otherwise, all `CommandOperations` will still be using a callback version of `execute.  

Here's the hierarchy as I envision it.  Is there a technical reason we diverged from this approach or was it a miss in the previous PR? 

Screenshot 2023-07-11 at 7 21 52 AM

baileympearson (changes requested):

src/operations/command.ts
malikj2000: I tried the hierarchy that you pictured a while back and it was throwing errors, so I changed it to the one I currently have. However, I just tried it again and it works perfectly fine, so I will change it back. I can't remember what the error was, but I must've changed something (in execute_operation perhaps) to fix it.

malikj2000: Ok I am getting those same errors again now, just checked with lint checker. If CommandOperation inherits from AbstractOperation and not AbstractCallbackOperation, all of the unchanged operations throw an error because they don't implement the abstract execute method from AbstractOperation. We need to keep it at AbstractCallbackOperation for now so it doesn't throw these errors, then change it back to AbstractOperation once all the operations are done.

src/operations/list_collections.ts
```diff
@@ -47,17 +48,16 @@ export class ListCollectionsOperation extends CommandCallbackOperation<string[]>
...
-    return super.executeCommandCallback(
+  override execute(server: Server, session: ClientSession | undefined): Promise<string[]> {
+    return super.executeCommand(
       server,
       session,
-      this.generateCommand(maxWireVersion(server)),
-      callback
-    );
+      this.generateCommand(maxWireVersion(server))
+    ) as TODO_NODE_3286;
```
nbbeeken, line 56: The TODO does not apply here, the generic is just incorrect. (Sorry 😅). I know this because MongoDB cannot return arrays as the top-level result. 

Hop in the debugger to confirm the result of this operation. This is a cursor creating command so the result is going to be something like: `{cursor:{id: number, firstBatch: Document[]}}`, (not sure the exact shape) 

It would be more correct to use `Document` at the very least, or we can declare a new `@internal` type `ListCollectionsResult` that can more accurately capture the expected response.

src/operations/list_collections.ts
```diff
@@ -47,17 +48,16 @@ export class ListCollectionsOperation extends CommandCallbackOperation<string[]>
...
+  }
+
+  executeCallback(_server: Server, _session: ClientSession | undefined, _callback: Callback): void {
+    throw new Error('Method not implemented');
   }
```
nbbeeken, line 61: This is unnecessary now b/c of the changes to inheritance correct?

test/integration/collection-management/collection.test.ts
```diff
@@ -26,90 +26,38 @@ describe('Collection', function () {
...
-              });
-            });
-          });
-        });
-      });
+      const spiderman_collection = await db.createCollection('test.spiderman');
+      const mario_collection = await db.createCollection('test.mario');
+      // Insert test documents (creates collections)
+      await spiderman_collection.insertOne({ foo: 5 }, configuration.writeConcernMax());
+      await mario_collection.insertOne({ bar: 0 }, configuration.writeConcernMax());
```
nbbeeken, line 56: It's interesting that these inserts were part of the test, the createCollection command will have created the collections, we don't need the inserts to do the implicit creation server side. so we can remove these.

test/integration/collection-management/collection.test.ts
```diff
@@ -404,40 +352,19 @@ describe('Collection', function () {
       });
     });
 
-    it('should correctly handle namespace when using collections method', function (done) {
+    it('should correctly handle namespace when using collections method', async function () {
```
nbbeeken, line 355: expand on "correctly"

test/unit/operations/get_more.test.ts
```diff
@@ -87,12 +87,10 @@ describe('GetMoreOperation', function () {
...
         const opts = { ...options, session };
         const operation = new GetMoreOperation(namespace, cursorId, server1, opts);
-        const callback = error => {
+        await operation.execute(server2, session).catch(error => {
           expect(error).to.be.instanceOf(MongoRuntimeError);
           expect(error.message).to.equal('Getmore must run on the same server operation began on');
-          done();
-        };
-        operation.executeCallback(server2, session, callback);
+        });
```
nbbeeken, line 93: This is totally correct code, there's potentially just a nicer way to format it. See the patterns we've used for error assertions elsewhere ([example](https://github.com/mongodb/node-mongodb-native/blob/NODE-5392/test/integration/node-specific/mongo_client.test.ts#L610)). You can make the catch instead return the error, and assert on the result.

nbbeeken (changes requested):

test/unit/operations/get_more.test.ts
malikj2000: Got it, done

test/integration/collection-management/collection_db_management.test.ts
```diff
@@ -1,83 +1,51 @@
...
+
+import { Collection, type Db, type MongoClient } from '../../mongodb';
+import { setupDatabase } from '../shared';
 
 describe('Collection Management and Db Management', function () {
+  let client: MongoClient;
+  let db: Db;
   before(function () {
     return setupDatabase(this.configuration);
   });
```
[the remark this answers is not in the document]
malikj2000: Ok it's been removed

test/integration/collection-management/collection_db_management.test.ts
```diff
@@ -1,83 +1,51 @@
...
-      .then(function (col) {
-        test.ok(col != null);
-
-        client.close(done);
-      })
-      .catch(function (err) {
-        test.ok(err != null);
-      });
+  beforeEach(function () {
+    client = this.configuration.newClient({}, { maxPoolSize: 5 });
```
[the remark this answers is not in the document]
malikj2000: Makes sense.

src/operations/command.ts
baileympearson: I think it's acceptable to have duplicated logic, if it's only in two place, it's temporary, and it's logic that we most likely won't touch once this PR is merged.

What's the concern with duplicating it?

malikj2000: That makes sense to me, shouldn't be a problem if it's temporary.

nbbeeken: I do not agree, we should not have duplicate logic that can fall out of sync, we do not know exactly when we'll have gotten to the point of being able to remove it. If we are duplicating the logic we need tests to make sure they stay the same, I think that's more effort than boxing up the logic in helpers that just proxy the arguments to the existing functionality. We have many incoming changes for the major that could effect this code path, we can avoid doing things twice even temporarily.

nbbeeken: From slack, let's try
- to avoid dupe code by creating an instance of CommandOperation
- we have to create a small anonymous class in order to to do so

test/integration/collection-management/collection.test.ts
```diff
@@ -26,90 +26,38 @@ describe('Collection', function () {
...
+        'test_collection_methods2'
+      );
+      expect(renameCollection.collectionName).to.equal('test_collection_methods2');
+      // Drop the collection and check that it's gone
+      const drop = await db.dropCollection('test_collection_methods2');
+      expect(drop).to.be.true;
     });
 
-    it('should correctly access collection names', function (done) {
+    it('should correctly access collection names', async function () {
```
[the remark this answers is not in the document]
nbbeeken: With this title and the other one, they are a lot more descriptive but I think you can be more specific about the intention of the test. Take this part `access [...] collection namespace from emptyDb.collections()` when reading that title what does `emptyDb` mean to someone who has not read the contents of the test? We want titles that describe the goal of the coverage / intention of the code:

```md
- context: when collections() is called on a database with dotted collection names
  - it: returns an array
  - it: returns an array of Collections
  - it: returns an array of Collections where each .collectionName includes the dot
  - it: returns an array of Collections where each .namespace includes the db name and dot
```

I'm not asking for more break down of the test, but consider the structure above as an idea of what to try and describe in a title, if a test starts failing, the failing assertion tells us what value is unexpected, the title should answer the question "What should the value be? and what should the API under test be accomplishing?"

cc: @baileympearson if you have more advice to share, please chime in :)

nbbeeken (changes requested):
Requesting changes for viz, pending from me:
- test title improvements
- de dupe of executeCommand logic
- CI fixes

test/integration/collection-management/collection.test.ts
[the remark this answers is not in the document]
baileympearson: A useful mental exercise I use is "given, action, outcome" (not very catchy but it works for me 🤷 ).  

- given a collection object
- when we call list collections with a dotted collection name ('foo.bar')
- the driver should create a collection whose name is the full collection name, including dots ('foo.bar')

I like this mental model because it applies to unit tests and integration tests nicely.  For unit tests, it could be something like "given an instance of CommandOperation, when we call execute(), we expect server.commandAsync() to be called exactly once".  For integration tests, it could be like "given a collection and a 7.0+ atlas replica set and there are search indexes on the collection, when I run a `listSearchIndexes` command, I get back an array of all the search indexes".

I find that thinking this way, specifically thinking about actions / outcomes, maps nicely to good test descriptions.

```typescript
context(`when the name of the collection has a dot in it ('foo.bar')`, async function() {
  it(`creates a collection with the name as specified, including dots`, () => { ... })
});
```

nbbeeken (approved):
LGTM

PR #3771, opened 2023-07-12
feat(NODE-5190)!: remove deprecated keep alive options

durran:
### Description

Remove the deprecated keep alive options.

#### What is changing?
- Removes the `keepAlive` and `keepAliveInitialDelay` options.
- Sets TCP keep alive to 30000 on the socket.

None

NODE-5190

#### The deprecated `keepAlive` and `keepAliveInitialDelay` options have been removed.

TCP keep alive will always be on and now set to a value of 30000ms.


W-A-James (approved):

src/cmap/connect.ts
```diff
@@ -377,7 +371,7 @@ function makeConnection(options: MakeConnectionOptions, _callback: Callback<Stre
     socket = net.createConnection(parseConnectOptions(options));
   }
 
-  socket.setKeepAlive(keepAlive, keepAliveInitialDelay);
+  socket.setKeepAlive(true, 300000);
```
nbbeeken, line 374: LGTM, Just calling out a reminder the spec mentions linking documentation, since we cannot set `tcp_keepalive_cnt` is that something that can be configured at the system level?

nbbeeken (approved):

src/cmap/connect.ts
durran: I've updated the docs changes needed to note that we should document what we set them to now, and if they need to be configured at the system level we should link to https://www.mongodb.com/docs/manual/faq/diagnostics/#does-tcp-keepalive-time-affect-mongodb-deployments

PR #3772, opened 2023-07-14
fix(NODE-5228)!: remove unneeded fields from ConnectionPoolCreatedEvent.options

W-A-James:
#### What is changing?
- Remove unneeded fields from `ConnectionPoolCraetedEvent.options`
- Made `ConnectionPoolCreatedEvent.options` field required

### Removed unneeded fields from `ConnectionPoolCreatedEvent.options`

The `options` field of `ConnectionPoolCreatedEvent` now has the following shape:

```ts
{
	maxPoolSize: number,
	minPoolSize: number,
	maxConnecting: number,
	maxIdleTimeMS: number,
	waitQueueTimeoutMS: number
}
```

### DIFF
--- a/src/cmap/connection_pool_events.ts
+++ b/src/cmap/connection_pool_events.ts
@@ -54,14 +54,19 @@ export abstract class ConnectionPoolMonitoringEvent {
  */
 export class ConnectionPoolCreatedEvent extends ConnectionPoolMonitoringEvent {
   /** The options used to create this connection pool */
-  options?: ConnectionPoolOptions;
+  options?: Pick<
+    ConnectionPoolOptions,
+    'maxPoolSize' | 'minPoolSize' | 'maxConnecting' | 'maxIdleTimeMS' | 'waitQueueTimeoutMS'
+  >;
   /** @internal */
   name = CONNECTION_POOL_CREATED;
 
   /** @internal */
   constructor(pool: ConnectionPool) {
     super(pool);
-    this.options = pool.options;
+    const { maxConnecting, maxPoolSize, minPoolSize, maxIdleTimeMS, waitQueueTimeoutMS } =
+      pool.options;
+    this.options = { maxConnecting, maxPoolSize, minPoolSize, maxIdleTimeMS, waitQueueTimeoutMS };
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/cmap/connection_pool_events.ts
```diff
@@ -54,14 +54,19 @@ export abstract class ConnectionPoolMonitoringEvent {
  */
 export class ConnectionPoolCreatedEvent extends ConnectionPoolMonitoringEvent {
   /** The options used to create this connection pool */
-  options?: ConnectionPoolOptions;
+  options: Pick<
+    ConnectionPoolOptions,
+    'maxPoolSize' | 'minPoolSize' | 'maxConnecting' | 'maxIdleTimeMS' | 'waitQueueTimeoutMS'
+  >;
```
nbbeeken, line 60: Since we're not consuming this API we don't have a type check here if ConnectionPoolOptions is changed it will just automatically be pulled in, can we add a type test?

nbbeeken (changes requested):

nbbeeken (approved):
LGTM

src/cmap/connection_pool_events.ts
baileympearson: Although we do it in other places, perhaps we shouldn't use `Pick` in our public APIs.  Generally our TS APIs should be as static as possible (i.e., hard coded) to ensure we don't break them by accident.  If the goal here was to ensure that `options`' properties types always match `ConnectionPoolOptions`, then could we consider something like this instead?  This would require no type tests.

```typescript
options: ConnectionPoolOptions & { maxPoolSize: number, minPoolSize: number, maxConnecting: number, ... }
```

nbbeeken: Isn't that sample code different? IIUC you are including the `ConnectionPoolOptions` again by intersecting the two types. Changing the declaration to that format does not remove the usefulness of type tests, they codify the public API as a downstream user would consume them, this format would conveniently warn on type changes but it is a distinct scenario we are trying to cover.

dariakp:
Friendly reminder to not forget the release highlight

baileympearson (approved):

PR #3773, opened 2023-07-14
refactor(NODE-5437): admin operations to use async syntax

malikj2000:
### Description
admin operations were converted from callback to async/await format

#### What is changing?
executeCallback changed to execute and now returns a promise
because of inheritance from AbstractCallbackOperation, there is also an implementation of executeCallback

No

further converting the driver from callbacks to async/await

### DIFF
--- a/src/operations/profiling_level.ts
+++ b/src/operations/profiling_level.ts
@@ -1,15 +1,15 @@
+import { type Callback, MongoUnexpectedServerResponseError } from 'mongodb-legacy';
+
 import type { Db } from '../db';
-import { MongoRuntimeError } from '../error';
 import type { Server } from '../sdam/server';
 import type { ClientSession } from '../sessions';
-import type { Callback } from '../utils';
-import { CommandCallbackOperation, type CommandOperationOptions } from './command';
+import { CommandOperation, type CommandOperationOptions } from './command';
 
 /** @public */
 export type ProfilingLevelOptions = CommandOperationOptions;
 
 /** @internal */
-export class ProfilingLevelOperation extends CommandCallbackOperation<string> {
+export class ProfilingLevelOperation extends CommandOperation<string> {
   override options: ProfilingLevelOptions;
 
   constructor(db: Db, options: ProfilingLevelOptions) {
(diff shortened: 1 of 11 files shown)


src/operations/profiling_level.ts
```diff
@@ -1,39 +1,42 @@
...
-    callback: Callback<string>
+  override async execute(server: Server, session: ClientSession | undefined): Promise<string> {
+    const doc = await super.executeCommand(server, session, { profile: -1 });
+    if (doc.ok === 1) {
+      const was = doc.was;
+      if (was === 0) return 'off';
+      if (was === 1) return 'slow_only';
+      if (was === 2) return 'all';
+      // TODO(NODE-3483)
+      throw new MongoUnexpectedServerResponseError(`Illegal profiling level value ${was}`);
```
nbbeeken, line 28: The TODOs can be removed since they are addressed here, and I would mention this change in the PR description.

src/operations/rename.ts
```diff
@@ -38,29 +38,23 @@ export class RenameOperation extends RunAdminCommandOperation {
...
-        newColl = new Collection(coll.s.db, this.newName, coll.s.options);
-      } catch (err) {
-        return callback(err);
-      }
+    let newColl: Collection<Document>;
+    try {
+      newColl = new Collection(coll.s.db, this.newName, coll.s.options);
+    } catch (err) {
+      return err;
+    }
```
nbbeeken, line 55: This is try/caught because the error needed to be passed to the callback, so there would not be inconsistency btwn sync/async errors. Since we're in an async function there's no difference we can just let this throw. Also, catching and `return err;`-ing is distinct from throwing / calling a callback with an error argument, this would be a significant change in the API.

src/operations/validate_collection.ts
```diff
@@ -34,26 +34,26 @@ export class ValidateCollectionOperation extends CommandCallbackOperation<Docume
...
-    session: ClientSession | undefined,
-    callback: Callback<Document>
-  ): void {
+  override async execute(server: Server, session: ClientSession | undefined): Promise<Document> {
     const collectionName = this.collectionName;
 
-    super.executeCommandCallback(server, session, this.command, (err, doc) => {
-      if (err != null) return callback(err);
+    const doc = await super.executeCommand(server, session, this.command);
+    // TODO(NODE-3483): Replace these with MongoUnexpectedServerResponseError
```
nbbeeken, line 41: No longer relevant TODO

nbbeeken (changes requested):

src/operations/validate_collection.ts
malikj2000: removed

src/operations/profiling_level.ts
malikj2000: removed

src/operations/rename.ts
malikj2000: got it, i removed it

nbbeeken (approved):

PR #3774, opened 2023-07-18
refactor(NODE-5458): refactoring index operations to use async syntax

malikj2000:
### Description
index operations were converted from callback to async/await syntax

#### What is changing?
all operations in indexes.ts now use async syntax
DropIndexesOperation was removed, and Collection.dropIndexes now invoked dropIndex with a "*"

No

further converting the driver from callbacks to async/await


src/operations/indexes.ts
```diff
@@ -458,39 +419,31 @@ export class IndexExistsOperation extends AbstractCallbackOperation<boolean> {
...
+    // Check in list of indexes
+    for (let i = 0; i < indexes.length; i++) {
+      if (info[indexes[i]] == null) {
+        return false;
       }
-    );
+    }
+
+    // All keys found return true
+    return true;
```
baileympearson, line 441: (optional cleanup)

We could clean this up a bit if we checked in the constructor if `indexes` was an array and if it wasn't, converting it to a string.  Then we only handle the case where `indexes` is an array.  We can also use a for-of loop while we're here.

Another option would be to use array builtins (instead of the loop entirely):
```
return Array.every(indexName => info[indexName] != null)
```

or a third option would be to use `full: true` and sets:

```
const indexes = await indexInformation(coll.s.db, coll.collectionName, {
      ...this.options,
      readPreference: this.readPreference,
      session,
	  full: true
    });

const names = indexes.map(({ name }) => name);

// isSuperSet is defined in util.ts
return isSuperSet(names, this.indexes);

////// or we get rid of indexInformation here entirely 
const indexesOnCollection = this.collection
    .listIndexes({ 
        ...this.options, 
        readPreference: this.readPreference, session })
    .map(({ name }) => name)
    .toArray();

return isSuperSet(indexesOnCollection, this.indexes);
```


feel free to refactor this if you want but this is optional

src/operations/indexes.ts
```diff
@@ -318,38 +303,28 @@ export class EnsureIndexOperation extends CreateIndexOperation {
...
+      if (indexName && indexes.some(index => index.name === indexName)) return indexName;
+      return await super.execute(server, session);
+    } catch (error) {
+      if (error instanceof MongoError && error.code === MONGODB_ERROR_CODES.NamespaceNotFound) {
+        // ignore "NamespaceNotFound" errors
+        return super.execute(server, session);
       }
-    );
+      return error;
+    }
```
baileympearson, line 319: The original error handling was only for the `toArray`, not the create call.  I'm not sure if `createIndex` will ever throw namespace not found errors, but we should revert the change anyways.

We can also flatten the logic a bit if we use a `catch` handler instead of a try-catch on the `toArray()`:


```suggestion
    const indexes = this.db.collection(this.collectionName)
		.listIndexes({ session })
		.toArray()
		.catch(error => {
			if (error is ns not found error) return []
			throw error;
		});

    if (indexName && indexes.some(index => index.name === indexName)
		return indexName;
    return await super.execute(server, session);
```

test/integration/index_management.test.js
```diff
@@ -258,39 +271,63 @@ describe('Indexes', function () {
     }
   });
 
-  it('shouldCorrectlyDropIndexes', {
-    metadata: {
-      requires: { topology: ['single', 'replicaset', 'sharded', 'ssl', 'heap', 'wiredtiger'] }
-    },
+  context('resolved dropIndexes', function () {
```
baileympearson, line 274: `resolved dropIndexes` describes the code (it uses promises now, it didn't before), not the behavior.  tests should describe the behavior under test.

this applies to all the tests here
For example:
```typescript
context('when dropIndexes succeeds', function() {
	it('dropIndexes returns true', function() {

	});

	it('the index is no longer present on the collection', function () {

	})
})
```

baileympearson (changes requested):

test/integration/index_management.test.js
```diff
@@ -258,39 +271,63 @@ describe('Indexes', function () {
...
-              test.equal('a_1', indexName);
-              // Drop all the indexes
-              collection.dropIndexes(function (err, result) {
-                test.equal(true, result);
+    beforeEach(async function () {
+      collection = await db.createCollection('test_drop_indexes');
+      await collection.insert({ a: 1 });
+      // Create an index on the collection
+      await db.createIndex(collection.collectionName, 'a');
+      /**@type {import('../tools/utils').FailPoint} */
```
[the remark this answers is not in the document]
malikj2000: I wrote this for the type autocomplete since this is a .js file. I can remove it now.

src/operations/indexes.ts
malikj2000: I used the second option I think it's the cleanest

test/integration/index_management.test.js
malikj2000: I can't separate the two tests here because of the beforeEach block

src/operations/indexes.ts
```diff
@@ -458,39 +418,25 @@ export class IndexExistsOperation extends AbstractCallbackOperation<boolean> {
...
+    const info = await indexInformation(coll.s.db, coll.collectionName, {
+      ...this.options,
+      readPreference: this.readPreference,
+      session
+    });
+    // Let's check for the index names
+    if (!Array.isArray(indexes)) return info[indexes] != null;
+
+    // All keys found return true
+    return Array.prototype.every(indexName => info[indexName] != null);
```
baileympearson, line 434:
```suggestion
    return indexes.every(indexName => info[indexName] != null);
```

I can't find documentation about what this returns but in my testing it always returns true.  I don't see tests failing because of this - can you add some basic testing for the `index exists` operation that would have caught this bug?  I think 

- a test for an index that doesn't exist returns false
- a test for an index that does exist returns true
- a test for an array of indexes that exist returns true
- a test for an array of indexes where one does't exist returns false

is probably sufficient

baileympearson (changes requested):

test/integration/index_management.test.js
```diff
@@ -258,39 +271,62 @@ describe('Indexes', function () {
...
+  context('when dropIndexes fails', function () {
+    let collection;
+
+    beforeEach(async function () {
+      collection = await db.createCollection('test_drop_indexes');
+      await collection.insert({ a: 1 });
+      // Create an index on the collection
+      await db.createIndex(collection.collectionName, 'a');
+      /**@type {import('../tools/utils').FailPoint} */
+      await client
```
baileympearson, line 307: it looks like tests are failing here, because failCommands are only available on 4.0+ servers: https://github.com/mongodb/mongo/wiki/The-%22failCommand%22-fail-point

the test in question is testing driver error handling of the server error.  I think it's fine to restrict this test to run only on 4.0+ servers, with a comment explaining why the restriction is there.

We should still test `createIndex` on all servers though.

malikj2000: how do you restrict it to specific servers

test/integration/index_management.test.js
```diff
@@ -1,13 +1,26 @@
-'use strict';
+('use strict');
```
baileympearson, line 1:
```suggestion
'use strict';
```

test/integration/index_management.test.js
```diff
@@ -1,13 +1,26 @@
...
   before(function () {
     return setupDatabase(this.configuration);
   });
 
+  beforeEach(function () {
+    client = this.configuration.newClient();
+    db = client.db();
+  });
+
+  this.afterEach(async function () {
```
baileympearson, line 20:
```suggestion
 afterEach(async function () {
```

test/integration/index_management.test.js
```diff
@@ -258,39 +271,106 @@ describe('Indexes', function () {
...
+    });
+
+    afterEach(async function () {
+      await db.dropCollection('test_drop_indexes');
+    });
+
+    it('should return false', {
+      metadata: {
+        requires: {
+          mongodb: '>4.0'
```
baileympearson, line 329:
```suggestion
          mongodb: '>=4.0'
```

I think this will pass on 4.0 servers too

test/integration/index_management.test.js
```diff
@@ -258,39 +271,106 @@ describe('Indexes', function () {
...
+
+    it('should return true when an array of indexes exists', async function () {
+      const result = await collection.indexExists('c_1_d_1_e_1');
+      expect(result).to.equal(true);
+    });
+
+    it('should return false when an array of indexes does not exist', async function () {
+      const result = await collection.indexExists('d_1_e_1_f_1');
+      expect(result).to.equal(false);
+    });
```
baileympearson, line 373: `indexExists` can take an array of indexes (i.e., `indexExists(['a_1', 'b_2'])`) and returns true iff all the indexes in the array exist.  That is the scenario we want to test with these two tests

baileympearson (changes requested):

test/integration/index_management.test.js
malikj2000: When you create the index with an array, though, the index looks like `'c_1_d_1_e_1'`. So when you check it like `indexExists(['a_1', 'b_2'])`, wouldn't it always return false?

baileympearson (approved):

PR #3775, opened 2023-07-19
fix(NODE-4936)!: remove unsupported options from db.command and admin.command

nbbeeken:
#### What is changing?

- Removes unsupported options from db.command and admin.command

Internal:
- Changes the inheritance of Run, Admin, Rename, CommandOperations classes
  - This fixes the accidental support of more options than should be supported
- Unskip test for RC option in runCommand
- Changes the construction of AdminCommand, unnecessary params
- MongoClient close uses executeOperation b/c of the public options not supporting `noResponse`

Spec compliance

### db.command() and admin.command() unsupported options removed

These APIs allow for specifying a command BSON document directly, so the driver does not try to enumerate all possible commands that could be passed to this API in an effort to be as forward and backward compatible as possible. Supporting various server-side options that are encoded into the command document is difficult to correctly support for all use cases.

The `db.command()` and `admin.command()` APIs have their `options` types updated to accurately support options compatible on all commands that could be passed to either API. Perhaps most notably, [`readConcern`](https://www.mongodb.com/docs/manual/reference/read-concern/) and [`writeConcern`](https://www.mongodb.com/docs/manual/reference/write-concern/) options are no longer handled by the driver, users **must** attach these properties to the command that is passed to the `.command()` method. 

- [x] Fill out release hightlight


src/operations/operation.ts
```diff
@@ -49,7 +49,6 @@ const kSession = Symbol('session');
  */
 export abstract class AbstractOperation<TResult = any> {
   ns!: MongoDBNamespace;
-  cmd!: Document;
```
baileympearson, line 52: removed because this no longer exists?

src/operations/rename.ts
```diff
@@ -17,39 +15,41 @@ export interface RenameOptions extends CommandOperationOptions {
...
-    const toCollection = collection.s.namespace.withCollection(newName).toString();
-    const dropTarget = typeof options.dropTarget === 'boolean' ? options.dropTarget : false;
-    const cmd = { renameCollection: renameCollection, to: toCollection, dropTarget: dropTarget };
-
-    super(collection, cmd, options);
+    this.ns = new MongoDBNamespace('admin', '$cmd');
     this.options = options;
     this.collection = collection;
     this.newName = newName;
   }
```
baileympearson, line 32:
```suggestion
  constructor(public collection: Collection, public newName: string, override public options: RenameOptions) {
    // Check the collection name
    checkCollectionName(newName);
    super(collection, options);

    this.ns = new MongoDBNamespace('admin', '$cmd');
  }
```
optional cleanup

src/operations/run_command.ts
```diff
@@ -1,73 +1,60 @@
...
 
-  constructor(parent: OperationParent | undefined, command: Document, options?: RunCommandOptions) {
-    super(parent, options);
-    this.options = options ?? {};
+  constructor(parent: Db, command: Document, options: RunCommandOptions) {
+    super(options);
+    this.options = options;
     this.command = command;
+    this.ns = parent.s.namespace.withCollection('$cmd');
   }
```
baileympearson, line 28:
```suggestion
  constructor(parent: Db, public command: Document, public override options: RunCommandOptions) {
    super(options);
    this.ns = parent.s.namespace.withCollection('$cmd');
  }
```
optional

src/operations/run_command.ts
```diff
@@ -1,73 +1,60 @@
...
+    this.ns = new MongoDBNamespace('admin', '$cmd');
+  }
+
+  override async execute(server: Server, session: ClientSession | undefined): Promise<T> {
+    this.server = server;
+    return server.commandAsync(this.ns, this.command, {
+      ...this.options,
+      readPreference: this.readPreference,
+      session
+    }) as TODO_NODE_3286;
```
baileympearson, line 58:
```suggestion
  constructor(
    public command: Document,
    public override options: RunCommandOptions & { noResponse?: boolean; bypassPinningCheck?: boolean }
  ) {
    super(options);
    this.ns = new MongoDBNamespace('admin', '$cmd');
  }
```
optional

src/operations/run_command.ts
```diff
@@ -1,73 +1,60 @@
...
   }
 }
 
-export class RunAdminCommandOperation<T = Document> extends RunCommandOperation<T> {
-  constructor(parent: OperationParent | undefined, command: Document, options?: RunCommandOptions) {
-    super(parent, command, options);
-    this.ns = new MongoDBNamespace('admin');
+export class RunAdminCommandOperation<T = Document> extends AbstractOperation<T> {
+  command: Document;
```
baileympearson, line 41: do we not need to override the option type here?

baileympearson (changes requested):
We're missing the following test from the kickoff:

> Add WC not supported test to match the RC one. (The spec tests that RC/WC is not inherited from db/client, we supported it as a direct option)

src/utils.ts
```diff
@@ -1155,20 +1155,11 @@ export function shuffle<T>(sequence: Iterable<T>, limit = 0): Array<T> {
 
 // TODO(NODE-4936): read concern eligibility for commands should be codified in command construction
 // @see https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#read-concern
-export function commandSupportsReadConcern(command: Document, options?: Document): boolean {
+export function commandSupportsReadConcern(command: Document): boolean {
```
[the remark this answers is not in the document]
nbbeeken: Yes, and actually good thing you commented here, I recall we discussed specifically not removing this because of how it effected users of mapReduce via runCommand, now that runCommand doesn't support RC we can remove this. I will update the highlight to capture that.

src/operations/rename.ts
nbbeeken: sure, done

src/operations/run_command.ts
nbbeeken: done

src/operations/run_command.ts
nbbeeken: No, the inheritance is from `AbstractOperation` not RunCommand, there isn't shared logic and it ended up just being a special case of runCommand where the namespace is hardcoded to admin. I could also change the usages of this class to RunCommandOperation, and pass in admin from the callers, but the support for the internal options `noResponse`, `bypassPinningCheck` on this class and not the other is nice organizationally I think.

src/operations/run_command.ts
nbbeeken: donezo

src/operations/operation.ts
nbbeeken: Kinda, the `xx!` properties on this class are tech debt from the rapid TS conversion, we couldn't sort out which subclasses actually defined these and which didn't. Specifically cmd what only being attached by the  `RenameOperation` so we don't need it defined on this super class. But also Rename is somewhat special in that most of our commands are build in the exec function, not in the constructor, so even Rename doesn't define this property anymore.

baileympearson (approved):

PR #3776, opened 2023-07-19
feat(NODE-3924)!: read tls files async

W-A-James:
##### connection string options
- updated `tlsCAFile` and `tlsCertificateKeyFile` options to store the file names and defer reading to first invocation of `MongoClient.connect`
- We now throw at option-parse time when provided an empty string for `tlsCAFile` or `tlsCertificateKeyFile`

##### MongoClient
- Added logic to `MongoClient.constructor` and `MongoClient.connect` to store `options.tlsCAFile` and `options.tlsCertificateKeyFile` on construction and read them into `options.ca` and `options.key` respectively on connect.

##### TLS support test
- refactored to typescript
- Added test that ensures that tls files are read in exactly once
- Added test that ensures that tls files are read at connect time
- Added test that ensures that connection fails when connecting with empty tls filepaths

##### TLS support test evergreen task and mocha config
- Updated mocha config to timeout in 10s instead of 2s
- ~Updated `run tls tests` evergreen task to timeout in 10s~

##### API docs
- Updated API documentation to mention how and when files specified by paths in `tlsCertificateKeyFile` and `tlsCAFile` are read

Yes, docs ticket has been generated.

NODE-3924

### TLS certificate authority and certificate-key files are now read asynchronously

In this release, we have updated the `MongoClient` to store the file names provided to the existing `tlsCAFile` and `tlsCertificateKeyFile` options and only read these files the first time it connects. Prior to this change, the files were read synchronously on `MongoClient` construction.

**Note:** This has no effect on driver functionality when TLS configuration files are properly specified.  However, if there are any issues with the TLS configuration files (invalid file name), the error is now thrown when the `MongoClient` is connected instead of at construction time.

```ts
const client = new MongoClient(CONNECTION_STRING, {
		tls: true,
		tlsCAFile: 'caFileName',
		tlsCertificateKeyFile: 'certKeyFile'
}); // Files are not read here, but file names are stored on the MongoClient

await client.connect(); // Files are now read and their contents stored
await client.close();

await client.connect(); // Since the file contents have already been cached, the files will not be read again.
```

Take a look at our [TLS documentation](https://www.mongodb.com/docs/drivers/node/current/fundamentals/connection/tls/) for more information on the `tlsCAFile` and `tlsCertificateKeyFile` options.


src/connection_string.ts
```diff
@@ -1097,16 +1096,12 @@ export const OPTIONS = {
     }
   },
   tlsCAFile: {
-    target: 'ca',
-    transform({ values: [value] }) {
-      return fs.readFileSync(String(value), { encoding: 'ascii' });
-    }
+    target: 'caFileName',
+    type: 'string'
```
W-A-James, line 1100: Can change this to not redirect to `caFileName`

test/unit/mongo_client.test.js
```diff
@@ -44,23 +44,22 @@ describe('MongoOptions', function () {
...
-     * | `key`                 | `tlsCertificateKeyFile`                       | `string`           |
-     * | `passphrase`          | `tlsCertificateKeyFilePassword`               | `string`           |
-     * | `rejectUnauthorized`  | `tlsAllowInvalidCertificates`                 | `boolean`          |
-     * | `checkServerIdentity` | `tlsAllowInvalidHostnames`                    | `boolean`          |
-     * | see note below        | `tlsInsecure`                                 | `boolean`          |
+     * | nodejs native option    | driver spec compliant option name             | driver option type |
+     * |:------------------------|:----------------------------------------------|:-------------------|
+     * | `tlsCAFile`             | `tlsCAFile`                                   | `string`           |
+     * | `crl`                   | N/A                                           | `string`           |
+     * | `tlsCertificateKeyFile` | `tlsCertificateKeyFile`                       | `string`           |
```
baileympearson, line 51: This shouldn't be changed, right?

src/mongo_client.ts
```diff
@@ -433,6 +434,18 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
 
     const options = this[kOptions];
 
+    if (options.tls) {
+      if (!options.ca && typeof options.tlsCAFile === 'string' && options.tlsCAFile.length > 0) {
+        options.ca = await fs.readFile(options.tlsCAFile, { encoding: 'utf8' });
+      }
```
baileympearson, line 440: I have two related cleanups here (these apply to `tlsCertificateKeyFile` too)

first, we can use the nullish coelescing assignment operator here to avoid checking `!options.ca`:
```typescript
      if (typeof options.tlsCAFile === 'string' && options.tlsCAFile.length > 0) {
        options.ca ??= await fs.readFile(options.tlsCAFile, { encoding: 'utf8' });
      }
```

---

second, I'm not sure we should be checking `options.tlsCAFile.length` (even though the AC says to).  the user must manually provide the filepath, and if we receive an empty string, the user's configuration is likely messed up.  We should probably alert the user that we received an empty file path instead of silently ignoring it.  Attempting to read from a file named "" will throw an error, so that was the old behavior.  

If we do decide to ignore empty strings, I suggest we move the logic into the options parser.  The requirement for these options would actually be "a non-empty string", instead of just "a string", so that should be codified in the options parsing and tested there.

.evergreen/config.yml
```diff
@@ -586,6 +586,7 @@ functions:
     - command: shell.exec
       type: test
       params:
+        timeout_secs: 10
```
baileympearson, line 589: Why are we setting a timeout here, and why is it 10 seconds?  Also, the timeout you set in the mocharc sets a default timeout per-test of 10s, whereas here the timeout is on the whole evergreen task

baileympearson (changes requested):

.evergreen/config.yml
W-A-James: This was cruft left over from trying to get the 0-length file path test to run correctly

test/unit/mongo_client.test.js
W-A-James: I may have misinterpreted what this table is for. I though the leftmost column corresponded to the option as it exists in node driver and the second column was the spec compliant option name. Is it actually the case that the leftmost column corresponds to node's _internal_ tls options?

src/mongo_client.ts
W-A-James: Your suggestion to handle empty file names in option parsing is more user friendly as far as catching bad config is concerned, so that's the solution I'd prefer as well.

baileympearson: Daria wrote the ticket and included "ignore empty filenames" as an AC.  Can you ask her why that was included?

nbbeeken: IIRC we wanted to have a nicer experience for environment variables being the source of the file name which can be set to an empty string as a way of clearing the value. I do not feel strongly, if we do ignore the empty string wouldn't TLS be misconfigured and fail? Is there a scenario where a server with TLS configured allows a client without TLS to connect? Also happy with keeping the behavior as it was (validate string and not length)

baileympearson: Well, in both cases it fails to connect, but if we throw ourselves it's a much more actionable error than a generic connection error. 

Did we have a specific environment / usecase / testing variant we wanted this for?

addaleax: > Is there a scenario where a server with TLS configured allows a client without TLS to connect?

Not sure what impact that has here, but sure, the server’s TLS mode can be `disabled`, `allowTLS`, `preferTLS`, `requireTLS`, and two of these are modes where the server accepts connections with or without TLS.

src/connection_string.ts
```diff
@@ -1096,10 +1096,20 @@ export const OPTIONS = {
     }
   },
   tlsCAFile: {
-    type: 'string'
+    transform({ name, values: [value] }) {
+      if (typeof value === 'string' && value.length === 0) {
```
baileympearson, line 1100: As-is, the options parsing would permit non-string values for `tlsCAFile` and `tlsCertificateKeyFile`.  Do we have tests for this case? 

I think we should:

- if a non-string value is passed, throw an error
- if a non-empty string is passed, throw an error
- only set the option if it is a string and the string is non-empty

src/mongo_client.ts
```diff
@@ -435,15 +435,14 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+        options.ca ??= await fs.readFile(options.tlsCAFile, { encoding: 'utf8' });
       }
       if (
-        !options.key &&
         typeof options.tlsCertificateKeyFile === 'string' &&
         options.tlsCertificateKeyFile.length > 0
       ) {
-        options.key = await fs.readFile(options.tlsCertificateKeyFile, { encoding: 'utf8' });
+        options.key ??= await fs.readFile(options.tlsCertificateKeyFile, { encoding: 'utf8' });
       }
```
baileympearson: Once we clean up the parsing, the we're guaranteed that these options are non-empty strings, so we can rely on truthiness alone.

```suggestion
      if (options.tlsCAFile) {
        options.ca ??= await fs.readFile(options.tlsCAFile, { encoding: 'utf8' });
      }
      if (options.tlsCertificateKeyFile) {
        options.key ??= await fs.readFile(options.tlsCertificateKeyFile, { encoding: 'utf8' });
      }
```

test/unit/mongo_client.test.js
```diff
@@ -44,15 +44,15 @@ describe('MongoOptions', function () {
...
-     * | nodejs native option    | driver spec compliant option name             | driver option type |
-     * |:------------------------|:----------------------------------------------|:-------------------|
-     * | `tlsCAFile`             | `tlsCAFile`                                   | `string`           |
-     * | `crl`                   | N/A                                           | `string`           |
-     * | `tlsCertificateKeyFile` | `tlsCertificateKeyFile`                       | `string`           |
-     * | `passphrase`            | `tlsCertificateKeyFilePassword`               | `string`           |
-     * | `rejectUnauthorized`    | `tlsAllowInvalidCertificates`                 | `boolean`          |
-     * | `checkServerIdentity`   | `tlsAllowInvalidHostnames`                    | `boolean`          |
-     * | see note below          | `tlsInsecure`                                 | `boolean`          |
+     * | nodejs native option  | driver spec compliant option name             | driver option type |
```
baileympearson: should we fix the header here too?

baileympearson (changes requested):

test/unit/mongo_client.test.js
```diff
@@ -44,7 +44,7 @@ describe('MongoOptions', function () {
      *
      * ### Additional options:
      *
-     * | nodejs native option  | driver spec compliant option name             | driver option type |
+     * | nodejs native option  | driver spec equicalient option name           | driver option type |
```
baileympearson, line 47:
```suggestion
     * | nodejs native option  | driver spec equivalent option name           | driver option type |
```

baileympearson (changes requested):

baileympearson (approved):

src/connection_string.ts
```diff
@@ -1097,15 +1096,29 @@ export const OPTIONS = {
...
-    transform({ values: [value] }) {
-      return fs.readFileSync(String(value), { encoding: 'ascii' });
+    transform({ name, values: [value] }) {
+      if (typeof value !== 'string') {
+        throw new MongoParseError(`${name} must be of type string`);
+      }
+
+      if (value.length === 0) {
+        throw new MongoParseError(`${name} must be have non-zero length`);
+      }
```
nbbeeken, line 1106: This changes the validation a bit more than I would expect from what we had before. I do think we should remove the string coercion that was present here before, it unnecessarily limited the inputs that `fs.readFile` accepts. Can we lean on Node.js errors for both the case of incorrect type and incorrect length? The other indication that we're overdoing it here is that even Node.js does not validate 0 length strings, it is the `open` system call returning `ENOENT` (`errno == -2`) that indicates an issue.

nbbeeken (changes requested):
One comment about how we should error. And there's slack thread about aligning on "empty" vs "unset".

src/connection_string.ts
W-A-James: So as far as moving forward here, it seems that we only want to pass through strings to the `fs.readFile` call, but I'm not clear on whether or not we want to throw on options parsing or at connect time and whether or not we want to check in our code whether or not the file path is empty.

nbbeeken: I think the known breakage here was that the error was moving to connect time, but then we're introducing new validation to construct time (currently). I think we should instead pass along whatever input we're given to the FS API.

baileympearson (approved):

nbbeeken (changes requested):

nbbeeken (approved):

PR #3777, opened 2023-07-19
refactor(NODE-5471): refactor crud operations to use async/await syntax

malikj2000:
### Description
converted all crud operations to use async/await syntax

#### What is changing?
all crud operations now inherit from Abstract/Command Operations instead of their callback counterparts
their execute methods also return promises

None

further converting the driver to async/await syntax


src/operations/delete.ts
```diff
@@ -105,22 +108,18 @@ export class DeleteOneOperation extends DeleteOperation {
...
-      });
-    });
+    session: ClientSession | undefined
+  ): Promise<DeleteResult> {
+    const res = await super.execute(server, session);
+    if (res.code) throw new MongoServerError(res);
+    if (res.writeErrors) throw new MongoServerError(res.writeErrors[0]);
+    if (this.explain) return res;
+    return {
+      acknowledged: this.writeConcern?.w !== 0 ?? true,
```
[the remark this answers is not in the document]
W-A-James: Should be able to get rid of the nullish coalescing here.

if `w` doesn't exist then, the inequality will resolve to `true`

if `w` does exist and is equal to zero, the inequality comparison will resolve to `false`, which is non-nullish so the result will be `false`

if `w` does exist and is not equal to zero, the inequality will resolve to `true`

W-A-James (changes requested):

W-A-James (changes requested):
CI is red

src/operations/update.ts
```diff
@@ -142,26 +141,22 @@ export class UpdateOneOperation extends UpdateOperation {
...
-      });
-    });
+    session: ClientSession | undefined
+  ): Promise<UpdateResult | Document> {
+    const res = await super.execute(server, session);
+    if (this.explain != null) return res;
+    if (res.writeErrors) throw new MongoServerError(res.writeErrors[0]);
+
+    return {
+      acknowledged: this.writeConcern?.w !== 0 ?? true,
```
W-A-James, line 153:
```suggestion
      acknowledged: this.writeConcern?.w !== 0,
```

src/operations/update.ts
```diff
@@ -179,26 +174,23 @@ export class UpdateManyOperation extends UpdateOperation {
...
-    });
+    session: ClientSession | undefined
+  ): Promise<UpdateResult | Document> {
+    const res = await super.execute(server, session);
+    if (this.explain != null) return res;
+    if (res.code) throw new MongoServerError(res);
+    if (res.writeErrors) throw new MongoServerError(res.writeErrors[0]);
+
+    return {
+      acknowledged: this.writeConcern?.w !== 0 ?? true,
```
W-A-James, line 187:
```suggestion
      acknowledged: this.writeConcern?.w !== 0,
```

src/operations/update.ts
```diff
@@ -235,26 +227,23 @@ export class ReplaceOneOperation extends UpdateOperation {
...
+    session: ClientSession | undefined
+  ): Promise<UpdateResult | Document> {
+    const res = await super.execute(server, session);
+    if (this.explain != null) return res;
+    if (res.code) throw new MongoServerError(res);
+    if (res.writeErrors) throw new MongoServerError(res.writeErrors[0]);
+
+    return {
+      acknowledged: this.writeConcern?.w !== 0,
+      modifiedCount: res.nModified != null ? res.nModified : res.n,
```
W-A-James, line 241:
```suggestion
      modifiedCount: res.nModified ?? res.n,
```

W-A-James (changes requested):
Missed a couple more of those slight syntax cleanups.

One thing we should also get in this PR is the `BulkWriteShimOperation`. It still inherits from `AbstractCallbackOperation`.

src/operations/bulk_write.ts
```diff
@@ -48,14 +46,12 @@ export class BulkWriteOperation extends AbstractCallbackOperation<BulkWriteResul
         bulk.raw(operations[i]);
       }
     } catch (err) {
-      return callback(err);
+      return err;
```
W-A-James, line 49: I think this is what's causing the CI to be so red. We were getting a lot of failures surrounding bulk write validation and not just throwing the error here would account for that.

You can get rid of the try-catch entirely and just run the loop since any errors thrown will automatically get converted to rejected promises

W-A-James (changes requested):
Identified the source of the CI redness

src/bulk/common.ts
```diff
@@ -835,26 +837,22 @@ export interface BulkWriteOptions extends CommandOperationOptions {
...
+  execute(_server: Server, session: ClientSession | undefined): Promise<any> {
     if (this.options.session == null) {
       // An implicit session could have been created by 'executeOperation'
       // So if we stick it on finalOptions here, each bulk operation
       // will use this same session, it'll be passed in the same way
       // an explicit session would be
       this.options.session = session;
     }
-    return executeCommands(this.bulkOperation, this.options, callback);
+    return promisify(executeCommands)(this.bulkOperation, this.options);
```
nbbeeken, line 855: Can you create this promisified version at the module level?

```ts
const executeCommandsAsync = promisify(executeCommands);
```

nbbeeken (changes requested):

src/operations/delete.ts
```diff
@@ -25,7 +22,7 @@ export interface DeleteOptions extends CommandOperationOptions, WriteConcernOpti
 }
 
 /** @public */
-export interface DeleteResult {
+export interface DeleteResult extends Document {
```
baileympearson, line 25: Why are we extending `Document`?

W-A-James (dismissed):

src/operations/delete.ts
```diff
@@ -105,22 +108,18 @@ export class DeleteOneOperation extends DeleteOperation {
...
-        deletedCount: res.n
-      });
-    });
+    session: ClientSession | undefined
+  ): Promise<DeleteResult> {
+    const res = await super.execute(server, session);
+    if (res.code) throw new MongoServerError(res);
+    if (res.writeErrors) throw new MongoServerError(res.writeErrors[0]);
+    if (this.explain) return res;
+    return {
```
baileympearson, line 119: Can we move the logic from this `execute` and `DeleteManyOperation`'s execute into the `DeleteOperation`?  Then these classes don't need `execute` methods at all.

src/operations/distinct.ts
```diff
@@ -73,17 +69,20 @@ export class DistinctOperation extends CommandCallbackOperation<any[]> {
     try {
       decorateWithCollation(cmd, coll, options);
     } catch (err) {
-      return callback(err);
+      return err;
     }
```
baileympearson, line 74: we can remove the try-catch here, because `async` functions automatically wrap any errors that are thrown in promise rejections.

src/operations/insert.ts
```diff
@@ -135,21 +135,24 @@ export class InsertManyOperation extends AbstractCallbackOperation<InsertManyRes
...
-        insertedIds: res.insertedIds
-      });
-    });
+    }
+    return {
+      acknowledged: writeConcern?.w !== 0,
+      insertedCount: res.insertedCount,
+      insertedIds: res.insertedIds
+    };
   }
```
baileympearson, line 156: we can clean this up
- we only need to check if `res` is null when we're using callbacks, because the signature for callbacks is `(error?, result?) => void`.  
- that means we can pull `res` into the try block specifically and pull the `return` statement into the `try` block too
- we need to `throw err`, not `return err` on line 148

src/operations/delete.ts
malikj2000: When you do res.code, res.writeErrors, etc. in DeleteOneOperation and DeleteManyOperation, it errors saying that code and writeErrors do not exist on DeleteResult

src/operations/update.ts
```diff
@@ -142,26 +141,22 @@ export class UpdateOneOperation extends UpdateOperation {
...
-        upsertedId:
-          Array.isArray(res.upserted) && res.upserted.length > 0 ? res.upserted[0]._id : null,
-        upsertedCount: Array.isArray(res.upserted) && res.upserted.length ? res.upserted.length : 0,
-        matchedCount: Array.isArray(res.upserted) && res.upserted.length > 0 ? 0 : res.n
-      });
-    });
+    session: ClientSession | undefined
+  ): Promise<UpdateResult | Document> {
+    const res = await super.execute(server, session);
+    if (this.explain != null) return res;
```
baileympearson, line 149: Did we intentionally remove the check for `res.code`?

src/operations/update.ts
```diff
@@ -179,26 +174,23 @@ export class UpdateManyOperation extends UpdateOperation {
...
```
baileympearson, line 182: Do we need this `execute` method?  The logic looks identical to the superclass.

src/operations/update.ts
```diff
@@ -235,26 +227,23 @@ export class ReplaceOneOperation extends UpdateOperation {
...
+    if (res.code) throw new MongoServerError(res);
+    if (res.writeErrors) throw new MongoServerError(res.writeErrors[0]);
```
baileympearson, line 237: same question here as above

baileympearson (changes requested):

src/operations/delete.ts
baileympearson: What happens if you change the definition to something like the following (note: I'm guessing on the types of writeErrors and code but I think this should fix the error that the properties don't exist on the delete result):

```typescript
export class DeleteOperation<DeleteResult> extends CommandOperation<DeleteResult & { code?: number, writeErrors?: Error[] } > {
```

src/operations/delete.ts
malikj2000: how do we do that

src/operations/insert.ts
malikj2000: if we put the return in the try block, the function throws saying that it's lacking a return statement

baileympearson: sorry, yes, we can also remove the `if(err)` too and then we should always be throwing from the `catch` block

src/operations/delete.ts
baileympearson: We should be able to move the logic right after the call to `super.executeCommand()` in delete operation, right?

malikj2000: in delete operation, it returns super.executeCommand(), so it would never reach the logic

baileympearson:
```typescript
export class DeleteOperation extends CommandOperation<DeleteResult> {
...
  override async execute(
    server: Server,
    session: ClientSession | undefined
  ): Promise<DeleteResult> {
    const options = this.options ?? {};
    const ordered = typeof options.ordered === 'boolean' ? options.ordered : true;
    const command: Document = {
      delete: this.ns.collection,
      deletes: this.statements,
      ordered
    };
    if (options.let) {
      command.let = options.let;
    }
    // we check for undefined specifically here to allow falsy values
    // eslint-disable-next-line no-restricted-syntax
    if (options.comment !== undefined) {
      command.comment = options.comment;
    }
    const unacknowledgedWrite = this.writeConcern && this.writeConcern.w === 0;
    if (unacknowledgedWrite) {
      if (this.statements.find((o: Document) => o.hint)) {
        // TODO(NODE-3541): fix error for hint with unacknowledged writes
        throw new MongoCompatibilityError(`hint is not supported with unacknowledged writes`);
      }
    }

    const res = super.executeCommand(server, session, command) as TODO_NODE_3286;
    // new logic here
    if (res.code) throw new MongoServerError(res);
    if (res.writeErrors) throw new MongoServerError(res.writeErrors[0]);
    if (this.explain) return res;

    return {
      acknowledged: this.writeConcern?.w !== 0,
      deletedCount: res.n
    };
  }
}
```
^ something like this?

malikj2000: got it. would there just be no execute function for delete one and delete many?

src/operations/delete.ts
malikj2000: this still throws errors, specifically because there's a third property (n) with type number. when I include that property, the error is that "type "number | undefined" is not assignable to type "number""

malikj2000: oh wait with the cleanup from below it was able to work

src/operations/delete.ts
baileympearson: That's my thought.  You can double check that the logic is the same in `DeleteOne.execute` and `DeleteMany.execute` by copy-pasting one into the other and confirming there are no changes with Git.

This works because `DeleteOperation` defines the `execute` method, and then each specialized class just inherits it.

baileympearson (changes requested):

W-A-James (changes requested):
Changes mostly look good, but you're getting a lot of red in CI around bulk write operations

src/operations/update.ts
```diff
@@ -142,26 +153,11 @@ export class UpdateOneOperation extends UpdateOperation {
...
-        modifiedCount: res.nModified != null ? res.nModified : res.n,
-        upsertedId:
-          Array.isArray(res.upserted) && res.upserted.length > 0 ? res.upserted[0]._id : null,
-        upsertedCount: Array.isArray(res.upserted) && res.upserted.length ? res.upserted.length : 0,
-        matchedCount: Array.isArray(res.upserted) && res.upserted.length > 0 ? 0 : res.n
-      });
-    });
+    session: ClientSession | undefined
+  ): Promise<UpdateResult | Document> {
+    return super.execute(server, session);
```
W-A-James, line 160: You can remove this and any other method that just passes through the arguments to the same method on its parent class.

src/operations/update.ts
```diff
@@ -179,26 +175,11 @@ export class UpdateManyOperation extends UpdateOperation {
...
```
W-A-James, line 182: Same as above

src/operations/update.ts
```diff
@@ -235,26 +216,11 @@ export class ReplaceOneOperation extends UpdateOperation {
...
```
W-A-James, line 223: Same as above

src/operations/delete.ts
```diff
@@ -105,22 +116,11 @@ export class DeleteOneOperation extends DeleteOperation {
...
-      if (this.explain) return callback(undefined, res);
-
-      callback(undefined, {
-        acknowledged: this.writeConcern?.w !== 0 ?? true,
-        deletedCount: res.n
-      });
-    });
+    session: ClientSession | undefined
+  ): Promise<DeleteResult> {
+    return super.execute(server, session);
```
W-A-James, line 123: Can delete this  since it just passes through arguments to the parent class

src/operations/delete.ts
```diff
@@ -129,22 +129,11 @@ export class DeleteManyOperation extends DeleteOperation {
...
```
W-A-James, line 136: Can delete this since it passes through arguments to the parent class

W-A-James (changes requested):

src/operations/update.ts
```diff
@@ -142,26 +141,23 @@ export class UpdateOneOperation extends UpdateOperation {
...
-          Array.isArray(res.upserted) && res.upserted.length > 0 ? res.upserted[0]._id : null,
-        upsertedCount: Array.isArray(res.upserted) && res.upserted.length ? res.upserted.length : 0,
-        matchedCount: Array.isArray(res.upserted) && res.upserted.length > 0 ? 0 : res.n
-      });
-    });
+    session: ClientSession | undefined
+  ): Promise<UpdateResult | Document> {
+    const res = await super.execute(server, session);
+    if (this.explain != null) return res;
+    if (res.code) throw new MongoServerError(res);
```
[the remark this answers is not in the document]
W-A-James: `res.code` is only ever non-zero when an error is present

W-A-James (changes requested):
Once you go through and make sure the existing suggestions about the nullish coalescing are fixed, then this should be done.

nbbeeken (dismissed):

W-A-James (changes requested):
Still missing a couple

baileympearson (approved):

PR #3779, opened 2023-07-20
refactor(NODE-5422): convert FLE kms providers and errors to Typescript

baileympearson:
#### What is changing?

The first half of FLE code has been converted to TS and integrated into the driver.  Specifically, the kms credential refresh logic and the errors.  This PR adds `gcp-metadata` as an optional dependency to enable gcp credential refreshing for FLE.

The failing dependency check action is expected, because we have not released a 6.0 major of `mongodb-client-encryption` yet.

No.


src/deps.ts
```diff
@@ -76,6 +75,18 @@ export function getZstdLibrary(): typeof ZStandard | { kModuleError: MongoMissin
   }
 }
 
+/**
```
baileympearson, line 78: Now that we consume these types in two places, I moved the `AWSCredentials` type into the `deps` module.  You'll also notice that I made the secret and session required.  This aligns with what the Typescript for the AWS SDK says (see the comment I left) and what I see when testing, so I think we can make the type stricter.

An alternative is to leave the properties optional and add logic in the aws credential fetching for FLE that asserts that these required properties are there and errors if they are not.

package.json
```diff
@@ -37,7 +37,7 @@
     "@mongodb-js/zstd": "^1.1.0",
     "gcp-metadata": "^5.2.0",
     "kerberos": "^2.0.1",
-    "mongodb-client-encryption": ">=2.3.0 <3",
+    "mongodb-client-encryption": ">=6.0.0-alpha.0 <7",
```
baileympearson, line 40: Optional, but probably worth doing.  After my previous PR, the old peer dependency compatibility range wasn't accurate.

tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "moduleResolution": "node",
     "skipLibCheck": true,
     "lib": [
-      "es2021"
+      "es2021",
+      "ES2022.Error"
```
baileympearson, line 13: `Nodejs 16` supports error's cause property, but it's included in ES2022.  We can safely pull `ES2022.Error` into our supported libs.

src/client-side-encryption/providers/index.ts
```diff
@@ -0,0 +1,169 @@
...
+ * @internal - exposed for testing purposes only
+ */
+export function isEmptyCredentials(provider: KMSProvider, kmsProviders: KMSProviders) {
+  return (
+    provider in kmsProviders &&
+    kmsProviders[provider] != null &&
+    typeof kmsProviders[provider] === 'object' &&
+    // Typescript does not infer that kmsProviders[provider] is non-null,
+    // even though we check it two lines up.
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
nbbeeken, line 142: I do not feel strongly, rather than explain this to the next reader/dev, I would just pull `kmsProviders[provider]` out to a variable and narrow the variable.

src/client-side-encryption/providers/utils.ts
```diff
@@ -1,16 +1,18 @@
...
- *
- * @returns { Promise<{ body: string, status: number }> }
- * @ignore
+ * @internal
  */
-function get(url, options = {}) {
+export function get(
+  url: URL | string,
+  options: http.RequestOptions = {}
+): Promise<{ body: string; status: number | undefined }> {
```
nbbeeken, line 12: This can be deduped with the http request helper in src/utils.ts right?

src/client-side-encryption/errors.ts
```diff
@@ -1,63 +1,71 @@
...
+ * @public
  * An error indicating that something went wrong specifically with MongoDB Client Encryption
  */
 export class MongoCryptError extends Error {
-  constructor(message, options = {}) {
-    super(message);
-    if (options.cause != null) {
-      this.cause = options.cause;
-    }
+  constructor(message: string, options: { cause?: Error } = {}) {
```
nbbeeken, line 8:
```ts
export class MongoCryptError extends Error {
  cause?: Error;
}
```
This was the type of the class in the bindings, so the ctor was already internal, should we keep it the same here? 

Does the inheritance change the type of cause? I think we might want `override cause?: Error` here since if someone has narrowed to this error instance then the cause will be an Error per our ctor argument.

test/unit/client-side-encryption/requirements.helper.ts
```diff
@@ -0,0 +1,33 @@
...
+export const AWS_SECRET_ACCESS_KEY = process.env.AWS_SECRET_ACCESS_KEY;
+export const AWS_REGION = process.env.AWS_REGION;
+export const AWS_CMK_ID = process.env.AWS_CMK_ID;
+
+export const awsKmsProviders = {
+  aws: { accessKeyId: AWS_ACCESS_KEY_ID, secretAccessKey: AWS_SECRET_ACCESS_KEY }
+};
+export const awsDataKeyOptions = { masterKey: { key: AWS_CMK_ID, region: AWS_REGION } };
+
+export const SKIP_AWS_TESTS = [
```
nbbeeken, line 15: Can we integrate this logic into our typical testing env info? or are there more clean up tickets on the way w.r.t tests? I'm imagining these could be methods/properties on `TestConfiguration`.

nbbeeken (changes requested):

src/client-side-encryption/providers/azure.ts
```diff
@@ -0,0 +1,171 @@
...
+ */
+interface AzureTokenCacheEntry extends AccessToken {
+  accessToken: string;
+  expiresOnTimestamp: number;
+}
+
+/**
+ * @internal
+ */
+export class AzureCredentialCache {
```
nbbeeken, line 28: From slack: Let's capture the deduplication of this in relation to future OIDC efforts

.evergreen/run-kerberos-tests.sh
```diff
@@ -24,11 +24,6 @@ set -o xtrace
 npm install kerberos@">=2.0.0-beta.0"
 npm run check:kerberos
```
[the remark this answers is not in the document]
baileympearson: We dropped support for kerberos 1.x a month ago in a separate ticket.

src/client-side-encryption/providers/utils.ts
baileympearson: It's not entirely straightforward - you'll notice this method is specialized to throw a `MongoCryptKMSRequestNetworkTimeoutError` error on timeout and to return a parsed JSON body with the status code.  The auth helper throws an auth error and only returns a parsed JSON body.

They can be deduped, but sharing the logic requires changes in both places, no matter which request helper we keep.  I didn't want my changes to spill into auth, so I kept it as-is.

baileympearson: _if_ we are going to make this change, I'd suggest we just use fetch (and pull in node_fetch for shell), and then both implementations effectively go away

src/client-side-encryption/errors.ts
baileympearson: Does the inheritance change the type of cause? I think we might want override cause?: Error here since if someone has narrowed to this error instance then the cause will be an Error per our ctor argument.

I'm not sure I follow.  Can you elaborate / provide an example?  `Error.cause` is already an error, no?

src/client-side-encryption/providers/azure.ts
```diff
@@ -0,0 +1,171 @@
+import { type Document } from 'bson';
+
+import { MongoCryptAzureKMSRequestError, MongoCryptKMSRequestNetworkTimeoutError } from '../errors';
+import { type KMSProviders } from '.';
```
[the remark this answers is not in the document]
baileympearson: Sure.  Why?

test/unit/client-side-encryption/requirements.helper.ts
baileympearson: Yeah, there are all the remaining tests to be converted to TS. I wasn't planning on cleaning this up too much - these are unit tests, not integration, so we don't have our usual test requirement infrastructure.

src/client-side-encryption/providers/azure.ts
```diff
@@ -0,0 +1,171 @@
+import { type Document } from 'bson';
```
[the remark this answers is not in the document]
baileympearson: If this is something we need to enforce, then we should enforce it somehow.  Linting is probably easiest.

src/client-side-encryption/errors.ts
nbbeeken: The cause from TS is defined as `unknown` because it can be any value (just like how `throw 'plain string'` is valid). Our errors will always set an `Error`, so we can make the type more accurate once a variable has been narrowed to one of our error classes.

baileympearson: We can address this when we subclass from `MongoError`, because then it'll inherit the narrowed type.

.evergreen/run-kerberos-tests.sh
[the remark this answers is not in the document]
nbbeeken: from slack: we can comment this out here, follow up to fix versioning

src/client-side-encryption/providers/azure.ts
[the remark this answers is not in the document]
nbbeeken: Just opinion, consider it optional 🧅

src/client-side-encryption/providers/utils.ts
nbbeeken: from tech chat, file a follow up

nbbeeken (approved):

PR #3781, opened 2023-07-21
feat(NODE-4796)!: remove addUser and collection.stats APIs

nbbeeken:
#### What is changing?

Removing deprecations

https://github.com/mongodb-js/nodejs-mongodb-legacy/pull/22

### `db.addUser()` and `admin.addUser()` removed

The deprecate `addUser` APIs have been removed. The driver maintains support across many server versions and the `createUser` command has support for different features based on the server's version, since applications can generally write code to work against a uniform and perhaps more modern server the path forward is for applications to send the createUser command directly.

The associated options interface with this API has also been removed: `AddUserOptions`.

See `createUser` documentation: https://www.mongodb.com/docs/manual/reference/command/createUser/

```ts
const db = client.db('admin');
// Example addUser usage
await db.addUser(
  'myUsername', 
  'myPassword', 
  { roles: [ { role: 'readWrite', db: 'mflix' } ] }
);
// Example equivalent command usage
await db.command({ 
  createUser: 'myUsername', 
  pwd: 'myPassword', 
  roles: [ { role: 'readWrite', db: 'mflix' } ] 
});
```

### `collection.stats()` removed

The collStats command is deprecated starting in server v6.2 so the driver is removing it's bespoke helper in this major release. The collStats command is still usable from manually running it via `await db.command()`. However, the recommended migration is to use the `$collStats` aggregation stage. You can find more information here: [$collStats](https://www.mongodb.com/docs/upcoming/reference/operator/aggregation/collStats/#mongodb-pipeline-pipe.-collStats).

The associated interfaces with this API have also been removed: `CollStatsOptions` and `WiredTigerData`.

### DIFF
--- a/test/integration/connection-monitoring-and-pooling/connection.test.ts
+++ b/test/integration/connection-monitoring-and-pooling/connection.test.ts
@@ -1,5 +1,3 @@
-import { promisify } from 'node:util';
-
 import { expect } from 'chai';
 
 import {


W-A-James (changes requested):
I didn't flag all of them, but can we update the tests that got touched to use async syntax? Other than that, lgtm.

W-A-James (approved):

PR #3782, opened 2023-07-21
feat(NODE-5409)!: allow socks to be installed optionally

nbbeeken:
#### What is changing?

- Add getSocks method to deps
- update dependencies test for socks, and general imports

No

See highlight

### Allow socks to be installed optionally

The driver uses the `socks` dependency to connect to `mongod` or `mongos` through a [SOCKS5 proxy](https://en.wikipedia.org/wiki/SOCKS).  `socks` used to be a required dependency of the driver and was installed automatically.  Now, `socks` is a `peerDependency` that must be installed to enable `socks` proxy support. 

### DIFF
--- a/src/client-side-encryption/stateMachine.js
+++ b/src/client-side-encryption/stateMachine.js
@@ -4,12 +4,25 @@ import * as tls from 'tls';
 import * as net from 'net';
 import * as fs from 'fs';
 import { once } from 'events';
-import { SocksClient } from 'socks';
 import { MongoNetworkTimeoutError } from '../error';
 import { debug, databaseNamespace, collectionNamespace } from './common';
 import { MongoCryptError } from './errors';
 import { BufferPool } from './buffer_pool';
 import { serialize, deserialize } from '../bson';
+import { getSocks } from '../deps';
+
+/** @type {import('../deps').SocksLib | null} */
+let socks = null;
+function loadSocks() {
+  if (socks == null) {
+    const socksImport = getSocks();
+    if ('kModuleError' in socksImport) {
+      throw socksImport.kModuleError;
+    }
+    socks = socksImport;
+  }
+  return socks;
+}
 
 // libmongocrypt states
 const MONGOCRYPT_CTX_ERROR = 0;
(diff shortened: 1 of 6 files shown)


src/client-side-encryption/stateMachine.js
```diff
@@ -4,12 +4,25 @@ import * as tls from 'tls';
...
 import { MongoNetworkTimeoutError } from '../error';
 import { debug, databaseNamespace, collectionNamespace } from './common';
 import { MongoCryptError } from './errors';
 import { BufferPool } from './buffer_pool';
 import { serialize, deserialize } from '../bson';
+import { getSocks } from '../deps';
+
+/** @type {import('../deps').SocksLib | null} */
+let socks = null;
+function loadSocks() {
```
baileympearson, line 16: Why is this necessary?  `makeModuleError` returns a proxy that throws whenever a key that isn't `kModuleError` is accessed.  So I'd expect we would just do the following (where it's used):
```typescript
const socks = loadSocks();
await socks.SocksClient(...)
```

src/cmap/connect.ts
```diff
@@ -419,6 +419,18 @@ function makeConnection(options: MakeConnectionOptions, _callback: Callback<Stre
   }
 }
 
+let socks: SocksLib | null = null;
+function loadSocks() {
+  if (socks == null) {
+    const socksImport = getSocks();
```
baileympearson, line 425: same comment as in stateMachine.js.

baileympearson (changes requested):
questions about how we're importing socks

src/client-side-encryption/stateMachine.js
nbbeeken: I'm alinging with the approach taken for kerberos, zstd, snappy. The goal is to not invoke the import operation more than once, determined by the nullishness of the `socks` variable.

src/cmap/connect.ts
nbbeeken: ditto response

baileympearson (approved):

PR #3783, opened 2023-07-24
feat(NODE-2014)!: return executor result from withSession and withTransaction

nbbeeken:
#### What is changing?

- `withSession` & `withTransaction` return the value returned by the provided functions

Yes, API docs have been updated. Look for docs changes needed on the ticket

See highlight.

### `withSession` and `withTransaction` return the value returned by the provided function

The `await client.withSession(async session => {})` now returns the value that the provided function returns. Previously, this function returned `void` this is a feature to align with the following breaking change. 

The `await session.withTransaction(async () => {})` method now returns the value that the provided function returns. Previously, this function returned the server command response which is subject to change depending on the server version or type the driver is connected to. The return value got in the way of writing robust, reliable, consistent code no matter the backing database supporting the application. 

### ⚠️ **Attention! `withTransaction` breaking change**

When upgrading to this version of the driver audit the usages of `withTransaction` for `if` statements or other conditional checks on the return value of `withTransaction`. Previously, the return value was the command response if the transaction was committed, and `undefined` if it had been manually aborted. It would only throw if an operation or the author of the function threw an error. Since thus far there has not been the ability to get the result of the function passed to `withTransaction` we suspect most existing usages to return `void`, making `withTransaction` a `void` returning function in this major release. Take care to ensure that the return values of your function match the expectation of the code that follows the completion of `withTransaction`. 

- [x] Release Notes

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -609,7 +609,7 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
    * The session will always be ended when the executor finishes.
    *
    * @param executor - An executor function that all operations using the provided session must be invoked in
-   * @param options - optional settings for the command
+   * @param options - optional settings for the session
    */
   async withSession<T = any>(executor: WithSessionCallback<T>): Promise<T>;
   async withSession<T = any>(
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -605,32 +605,30 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
   }
 
   /**
-   * Runs a given operation with an implicitly created session. The lifetime of the session
-   * will be handled without the need for user interaction.
+   * A convenience method for creating and handling the clean up of a ClientSession.
+   * The session will always be ended when the executor finishes.
    *
-   * NOTE: presently the operation MUST return a Promise (either explicit or implicitly as an async function)
-   *
-   * @param options - Optional settings for the command
-   * @param callback - An callback to execute with an implicitly created session
+   * @param executor - An executor function that all operations using the provided session must be invoked in
+   * @param options - optional settings for the command
    */
-  async withSession<T = any>(callback: WithSessionCallback<T>): Promise<T>;
+  async withSession<T = any>(executor: WithSessionCallback<T>): Promise<T>;
   async withSession<T = any>(
     options: ClientSessionOptions,
-    callback: WithSessionCallback<T>
+    executor: WithSessionCallback<T>
   ): Promise<T>;
   async withSession<T = any>(
-    optionsOrOperation: ClientSessionOptions | WithSessionCallback<T>,
-    callback?: WithSessionCallback<T>
+    optionsOrExecutor: ClientSessionOptions | WithSessionCallback<T>,
+    executor?: WithSessionCallback<T>
   ): Promise<T> {
     const options = {
       // Always define an owner
       owner: Symbol(),
       // If it's an object inherit the options
-      ...(typeof optionsOrOperation === 'object' ? optionsOrOperation : {})
+      ...(typeof optionsOrExecutor === 'object' ? optionsOrExecutor : {})
     };
 
     const withSessionCallback =
-      typeof optionsOrOperation === 'function' ? optionsOrOperation : callback;
+      typeof optionsOrExecutor === 'function' ? optionsOrExecutor : executor;
 
     if (withSessionCallback == null) {
       throw new MongoInvalidArgumentError('Missing required callback parameter');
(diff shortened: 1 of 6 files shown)


src/mongo_client.ts
```diff
@@ -605,29 +605,30 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
-   * will be handled without the need for user interaction.
+   * A convenience method for creating and handling the clean up of a ClientSession.
+   * The session will always be ended when the executor finishes.
    *
-   * NOTE: presently the operation MUST return a Promise (either explicit or implicitly as an async function)
-   *
-   * @param options - Optional settings for the command
-   * @param callback - An callback to execute with an implicitly created session
+   * @param executor - An executor function that all operations using the provided session must be invoked in
+   * @param options - optional settings for the command
```
baileympearson, line 612:
```suggestion
   * @param options - optional settings for the session
```

test/types/sessions.test-d.ts
```diff
@@ -15,3 +15,13 @@ expectType<ClientSession>(
...
+
+let something: any;
+expectType<number>(await client.withSession(async () => 2));
+expectType<string>(await client.withSession<string>(async () => something));
+const untypedFn: any = () => 2;
+expectType<any>(await client.withSession(untypedFn));
+const unknownFn: () => Promise<unknown> = async () => 2;
+expectType<unknown>(await client.withSession(unknownFn));
+// Not a promise returning function
+expectError(await client.withSession(() => null));
```
baileympearson, line 27: We have tests in `sessions.test.ts` that explicitly test this scenario - is that an accident or do we intentionally silently support this synchronous executors?

src/sessions.ts
```diff
@@ -591,18 +589,18 @@ function attemptTransaction<TSchema>(
...
   return promise.then(
-    () => {
+    result => {
       if (userExplicitlyEndedTransaction(session)) {
-        return;
+        return result;
       }
 
-      return attemptTransactionCommit(session, startTime, fn, options);
+      return attemptTransactionCommit(session, startTime, fn, options).then(() => result);
```
baileympearson, line 603: What happens if we need to retry the transaction?  I think we'd get incorrect behavior, because the first call to `attemptTransaction` would always return the result of the first call to `fn` - but I'd assume that we want to return the `result` associated with the successful `attemptTransaction`.

baileympearson: Perhaps this is okay and expected behavior?  We document that `Client.withSession` may call its callback more than once - we should here too at the very least.  But I think since we may call the callback twice, it's an acceptable limitation that the callback should be idempotent.

src/sessions.ts
```diff
@@ -591,18 +589,18 @@ function attemptTransaction<TSchema>(
 
   if (!isPromiseLike(promise)) {
     session.abortTransaction().catch(() => null);
-    throw new MongoInvalidArgumentError(
-      'Function provided to `withTransaction` must return a Promise'
+    return Promise.reject(
```
baileympearson, line 592: Why'd you revert the `throw`?

src/sessions.ts
baileympearson: edit: it is noted for this method

baileympearson (changes requested):
some questions

src/sessions.ts
nbbeeken: pending slack convo

nbbeeken: Updated to return the last result and added a test

test/integration/transactions/transactions.test.ts
```diff
@@ -136,6 +136,47 @@ describe('Transactions', function () {
...
+        let counter = 0;
+        const withTransactionResult = await session
+          .withTransaction(async session => {
+            await collection.insertOne({ a: 1 }, { session });
+            counter += 1;
+            return counter;
+          })
+          .finally(async () => await session.endSession());
+
+        expect(withTransactionResult).to.equal(counter);
```
baileympearson, line 177:
```suggestion
        expect(withTransactionResult).to.equal(3);
```
either we do this, or we somehow assert that the transaction was retried.  this test would pass as-is if the failpoint were never triggered

src/sessions.ts
baileympearson: I was just curious, this change is fine.  It'll be moot once Malik gets to this code anyways

baileympearson (changes requested):

test/integration/transactions/transactions.test.ts
nbbeeken: fixed this and the manual abort test

baileympearson (approved):

W-A-James (approved):

PR #3787, opened 2023-07-27
feat(NODE-5470)!: convert remaining FLE to TS and drop support for `onKMSProvidersRefresh`

baileympearson:
#### What is changing?

The remaining FLE files and tests are converted to TS.  Additional changes:

- buffer_pool.js was removed and now FLE code imports the buffer pool from `utils.ts`.
- `once` and `child_process` are lazily imported in the state machine and mongodbcyptd manager, respectively 
- lint has been re-enabled and once again covers all tests and src
- `databaseNamespace` and `collectionNamespace` have been removed in favor of `MongoDBNamespace`.

One other notable change is that when we fail to get kms provider information, the error thrown _used to_ attach an `originalError` property.  I adjusted this to use the error's `cause`, since Node16+ and MongoCryptErrors support the `cause` property.

I also removed support for `onKMSProvidersRefresh` here.

No, documentation is coming in a follow up.

### `ClientEncryption.onKMSProvidersRefresh` function removed

`ClientEncryption.onKMSProvidersRefresh` was added as a public API in version 2.3.0 of `mongodb-client-encryption` to allow for automatic refresh of KMS provider credentials.  Subsequently, we've added the capability to automatically refresh KMS credentials using the KMS provider's preferred refresh mechanism, and `onKMSProviderRefresh` is no longer used.


src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+    this._metaDataClient = options.metadataClient || client;
+    this._proxyOptions = options.proxyOptions || {};
+    this._tlsOptions = options.tlsOptions || {};
+    this._kmsProviders = options.kmsProviders || {};
+
+    const mongoCryptOptions: MongoCryptOptions = {};
+    if (options.schemaMap) {
+      mongoCryptOptions.schemaMap = Buffer.isBuffer(options.schemaMap)
+        ? options.schemaMap
+        : (serialize(options.schemaMap) as Buffer);
```
baileympearson, line 300: You'll notice a lot of type casts here.  `schemaMap` explicitly needs to be a buffer for our bindings, but our bson serialize is platform-compatible and is typed to return `Uint8Arrays`.  At runtime, BSON returns `Buffer`s for Node, so it's still valid, but these casts are unavoidable.

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+        )
+      );
+    }
+    if (!this._mongocryptdClient) {
+      return callback(
+        new MongoRuntimeError(
+          'Reached impossible state: mongocryptdClient is undefined when neither bypassSpawn nor the shared lib are specified.'
+        )
+      );
+    }
```
baileympearson, line 385: Typescript complains here, because we (developers) know that `init` is a no-op when `this._bypassMongocryptdAndCryptShared || this.cryptSharedLibVersionInfo` is specified and that if neither of those are true, we always have a _mongocryptdManager` and client.  We can cast it away but I think errors are a good practice, because these provide actionable feedback if the assumption that these properties are always defined when we need to spawn a mongocryptd is ever violated.

baileympearson: Maybe a `MongoCryptImpossibleState` error?

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+    options: CommandOptions,
+    callback: Callback<Document | Uint8Array>
+  ): void;
+  /**
+   * Encrypt a command for a given namespace.
+   */
+  encrypt(
+    ns: string,
+    cmd: Document,
+    options?: CommandOptions | Callback<Document | Uint8Array>,
```
baileympearson, line 465: `encrypt` is only called in one location (CryptoConnection.command) and we pass the command arguments in.

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+    callback?: Callback<Document | Uint8Array>
+  ) {
+    if (typeof ns !== 'string') {
+      throw new TypeError('Parameter `ns` must be a string');
+    }
+
+    if (typeof cmd !== 'object') {
+      throw new TypeError('Parameter `cmd` must be an object');
+    }
```
baileympearson, line 474: I decided not to clean up argument checking like this - for our public APIs in the CE object, I think they're worthwhile to keep.  Private APIs like this I think could be cleaned up, but for consistency and fewer changes I left them as-is.

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+      context = this._mongocrypt.makeEncryptionContext(
+        MongoDBCollectionNamespace.fromString(ns).db,
+        commandBuffer
+      );
+    } catch (err) {
+      callback(err, undefined);
+      return;
+    }
+
+    // TODO: should these be accessors from the addon?
```
baileympearson, line 503: note: these TODOs were here before me.

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+    context.id = this._contextCounter++;
+    context.ns = ns;
+    context.document = cmd;
+
+    const stateMachine = new StateMachine({
+      ...options,
```
baileympearson, line 509: perhaps questionable if this is needed .. I think we only need the user's BSON serialization options (if that) but future work can investigate this further.

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+    // TODO: should this be an accessor from the addon?
+    context.id = this._contextCounter++;
+
+    const stateMachine = new StateMachine({
+      ...options,
+      proxyOptions: this._proxyOptions,
+      tlsOptions: this._tlsOptions
+    });
+
+    // @ts-expect-error unique symbol cannot be used as an index type
```
baileympearson, line 553: I was unsure how to get around these errors - any thoughts?

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+    const decorateResult = this[Symbol.for('@@mdb.decorateDecryptionResult')];
+    stateMachine.execute(this, context, function (error?: Error, result?: Document) {
+      // Only for testing/internal usage
+      if (!error && result && decorateResult) {
+        const error = decorateDecryptionResult(result, response);
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
baileympearson, line 559: These will go away after we refactor to async-await, but for now TS complains because technically `callback` could have been re-assigned by something before we execute this code.  In practice, we know it never is re-assigned.

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,641 @@
...
+
+  /**
+   * Ask the user for KMS credentials.
+   *
+   * This returns anything that looks like the kmsProviders original input
+   * option. It can be empty, and any provider specified here will override
+   * the original ones.
+   */
+  async askForKMSCredentials(): Promise<KMSProviders> {
+    return loadCredentials(this._kmsProviders);
```
baileympearson, line 575: removed `onKmsProvidersRefresh`.  

We could consider removing this entirely and just calling `loadCredentials` directly in the state machine.

src/client-side-encryption/clientEncryption.ts
```diff
@@ -0,0 +1,1015 @@
...
+import { type Filter } from '../mongo_types';
+import { type CreateCollectionOptions } from '../operations/create_collection';
+import { type DeleteResult } from '../operations/delete';
+import { type Callback, MongoDBCollectionNamespace } from '../utils';
+import { maybeCallback, promiseOrCallback } from './common';
+import * as cryptoCallbacks from './cryptoCallbacks';
+import { MongoCryptCreateDataKeyError, MongoCryptCreateEncryptedCollectionError } from './errors';
+import { type KMSProvider, type KMSProviders, loadCredentials } from './providers/index';
+import { type CSFLEKMSTlsOptions, StateMachine, type StateMachineExecutable } from './stateMachine';
+
```
baileympearson, line 24: Types in this file were directly copied in from https://github.com/mongodb/libmongocrypt/blob/8588825112a93145bf82b4b7519070c101c45a43/bindings/node/index.d.ts#L448, with the exception of a few modifications:

- TLS options 
-

src/client-side-encryption/clientEncryption.ts
```diff
@@ -0,0 +1,1015 @@
...
+    const context = this._mongoCrypt.makeDataKeyContext(dataKeyBson, {
+      keyAltNames,
+      keyMaterial
+    });
+    const stateMachine = new StateMachine({
+      proxyOptions: this._proxyOptions,
+      tlsOptions: this._tlsOptions
+    });
+
+    // @ts-expect-error We did not convert promiseOrCallback to TS
```
baileympearson, line 205: We will remove callback support once this PR is merged - I intentionally did not TSify `promiseOrCallback` or `maybePromise`.

src/client-side-encryption/clientEncryption.ts
```diff
@@ -0,0 +1,1015 @@
...
+        throw new TypeError(`"options" cannot contain both "keyId" and "keyAltName"`);
+      }
+      if (typeof keyAltName !== 'string') {
+        throw new TypeError(
+          `"options.keyAltName" must be of type string, but was of type ${typeof keyAltName}`
+        );
+      }
+
+      contextOptions.keyAltName = serialize({ keyAltName });
+    }
```
baileympearson, line 743: I refactored options construction here because the old approach gives TS errors.  For example, keyId.

```typescript
const contextOptions: ExplicitEncryptionContextOptions = Object.assign({}, options);
if (options.keyId) {
  contextOptions.keyId = options.keyId.buffer;
}
```

In the above, TS complains in the first line because we assign a `Binary` to the `keyId` option, which accepts a buffer, only to re-assign it in the if statement.  Most of the options were impacted by this issue.

src/client-side-encryption/common.js
```diff
@@ -1,33 +1,4 @@
-/**
```
baileympearson, line 1: `debug` was moved to the state machine, and the namespace helpers were removed in favor of `MongoDBNamespace`.  the remainder of the file is untouched and I disabled lint because we plan to remove this entirely in a follow-up-ticket.

src/client-side-encryption/stateMachine.ts
```diff
@@ -43,90 +53,135 @@ const INSECURE_TLS_OPTIONS = [
...
+}
 
-/**
- * @ignore
- * @callback StateMachine~fetchCollectionInfoCallback
- * @param {Error} [err] If present, indicates that fetching the collection info failed with the given error
- * @param {object} [result] If present, is the fetched collection info for the first collection to match the given filter
- * @returns {void}
- */
+declare module 'mongodb-client-encryption' {
```
baileympearson, line 66: These three properties are only used in the JS layer and are only added by CE and AE to communicate context to the state machine.  A future refactor could provide them as a "context" object to `StateMachine.execute` but I left as-is for now.

We use declaration merging to add the properties to the `MongoCryptContext` interface only in the driver.

src/client-side-encryption/stateMachine.ts
```diff
@@ -43,90 +53,135 @@ const INSECURE_TLS_OPTIONS = [
...
-  constructor(options) {
-    this.options = options || {};
-
-    this.executeAsync = promisify((autoEncrypter, context, callback) =>
-      this.execute(autoEncrypter, context, callback)
-    );
+export class StateMachine {
+  constructor(
+    private options: StateMachineOptions,
+    private bsonOptions = pluckBSONSerializeOptions(options)
```
baileympearson, line 125: This is new.  Previously, we simply passed the `options` object into `serialize` to provide bson options.  I figured it makes sense to pull off all supported BSON serialization options and store them separately for clarity.

src/client-side-encryption/stateMachine.ts
```diff
@@ -272,10 +336,10 @@ class StateMachine {
         reject(new MongoCryptError('KMS request timed out'));
       }
 
-      function onerror(err) {
+      function onerror(err: Error) {
         destroySockets();
-        const mcError = new MongoCryptError('KMS request failed');
-        mcError.originalError = err;
+        // TODO: make note of this
+        const mcError = new MongoCryptError('KMS request failed', { cause: err });
```
baileympearson, line 342: This is a behavior change.  We now have access to the `cause` property - may as well make proper use of it.

nbbeeken: I assume this PR comment is the todo: `// TODO: make note of this`?

src/client-side-encryption/stateMachine.ts
nbbeeken: SGTM

src/client-side-encryption/cryptoCallbacks.ts
```diff
@@ -1,11 +1,16 @@
...
 
-export function makeAES256Hook(method, mode) {
-  return function (key, iv, input, output) {
+type AES256Callback = (key: Buffer, iv: Buffer, input: Buffer, output: Buffer) => number | Error;
+
+export function makeAES256Hook(
+  method: 'createCipheriv' | 'createDecipheriv',
+  mode: 'aes-256-cbc' | 'aes-256-ctr'
+): AES256Callback {
+  return function (key: Buffer, iv: Buffer, input: Buffer, output: Buffer): number | Error {
```
nbbeeken, line 9: I think we should strive to lower the contract needed here, the APIs accept Uint8Array as readily as they do Buffers. Are there Buffer specific APIs needed here?

src/client-side-encryption/mongocryptdManager.ts
```diff
@@ -32,27 +34,30 @@ export class MongocryptdManager {
...
     // to ensure child can outlive parent.
     this._child = spawn(cmdName, this.spawnArgs, {
       stdio: 'ignore',
       detached: true
     });
 
-    this._child.on('error', () => {});
+    this._child.on('error', () => {
+      // perhaps questionable, but we swallow mongocryptd spawn errors.
+    });
```
nbbeeken, line 60: It is not exactly questionable, the source of "truth" for a successful launch is a mongocryptd process we can connect to. Since this channel of getting an error can occur at anytime we defer to the connection to mongocryptd being broken instead which is kind of our way of "monitoring" that it is still up.

src/client-side-encryption/stateMachine.ts
```diff
@@ -56,90 +65,149 @@ const INSECURE_TLS_OPTIONS = [
...
 /**
- * @ignore
- * @callback StateMachine~executeCallback
- * @param {Error} [err] If present, indicates that the execute call failed with the given error
- * @param {object} [result] If present, is the result of executing the state machine.
- * @returns {void}
+ * Helper function for logging. Enabled by setting the environment flag MONGODB_CRYPT_DEBUG.
+ * @param msg - Anything you want to be logged.
  */
+export function debug(msg: unknown) {
```
nbbeeken, line 71:
```suggestion
function debug(msg: unknown) {
```

Not used elsewhere right?

src/client-side-encryption/stateMachine.ts
```diff
@@ -56,90 +65,149 @@ const INSECURE_TLS_OPTIONS = [
 ];
```
nbbeeken, line 65:
```
'tlsDisableOCSPEndpointCheck',
'tlsDisableCertificateRevocationCheck'
```
These are not client options. Maybe we should leave them though but with a comment to help clear up potential confusion.

src/client-side-encryption/stateMachine.ts
```diff
@@ -56,90 +65,149 @@ const INSECURE_TLS_OPTIONS = [
...
+ * @public
+ *
+ * TLS options to use when connecting. The spec specifically calls out which insecure
+ * tls options are not allowed:
+ *
+ *  - tlsAllowInvalidCertificates
+ *  - tlsAllowInvalidHostnames
+ *  - tlsInsecure
+ *  - tlsDisableOCSPEndpointCheck
+ *  - tlsDisableCertificateRevocationCheck
```
nbbeeken, line 98: We should remove these options from documentation

src/index.ts
```diff
@@ -93,13 +93,21 @@ export {
   UnorderedBulkOperation
 };
 
+export type { AutoEncryptionExtraOptions } from './client-side-encryption/autoEncrypter';
+export type { KMSProviders } from './client-side-encryption/providers/index';
+export type {
+  CSFLEKMSTlsOptions,
+  CSFLETlsOptions,
+  StateMachineExecutable
+} from './client-side-encryption/stateMachine';
```
nbbeeken, line 102: Can you move these down to live with the other type exports? we have a line "type only exports below, these are removed from emitted JS" breaking up the runtime values from the types.

nbbeeken (changes requested):

src/utils.ts
```diff
@@ -346,6 +346,10 @@ export class MongoDBCollectionNamespace extends MongoDBNamespace {
   constructor(db: string, override collection: string) {
     super(db, collection);
   }
+
+  static override fromString(namespace?: string | undefined): MongoDBCollectionNamespace {
```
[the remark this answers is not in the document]
baileympearson: It has the signature as the parent method it's overriding.

src/client-side-encryption/clientEncryption.ts
```diff
@@ -0,0 +1,1009 @@
...
+  creationDate: Date;
+  updateDate: Date;
+  status: number;
+  masterKey: Document;
+}
+
+/**
+ * The public interface for explicit in-use encryption
+ */
+export class ClientEncryption implements StateMachineExecutable {
```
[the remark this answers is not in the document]
baileympearson: https://jira.mongodb.org/browse/NODE-5423

src/client-side-encryption/mongocryptdManager.ts
baileympearson: Right - but we could race `spawn` and `error` together, and once successfully spawned stop listening for errors. 

It's strange that there's no way to know if auto spawn was successful other than a failing auto encrypted command (which manifests itself as a failure to connect the mongocryptd client).

src/client-side-encryption/stateMachine.ts
```diff
@@ -357,33 +435,29 @@ class StateMachine {
...
    *
-   * @param {ClientEncryptionTLSOptions} tlsOptions The client TLS options for the provider.
-   * @param {Object} options The existing connection options.
+   * @param tlsOptions - The client TLS options for the provider.
+   * @param options - The existing connection options.
    */
-  setTlsOptions(tlsOptions, options) {
+  setTlsOptions(tlsOptions: CSFLETlsOptions, options: tls.ConnectionOptions) {
     if (tlsOptions.tlsCertificateKeyFile) {
       const cert = fs.readFileSync(tlsOptions.tlsCertificateKeyFile);
```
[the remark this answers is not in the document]
baileympearson: I added it as an AC in the ticket

src/index.ts
baileympearson: Is this something we can enforce (via linting or another measure)?  (if it matters...?)

src/client-side-encryption/autoEncrypter.ts
baileympearson: it isn't fixable for users if they encounter this error - we create a `mongocryptd` client so if they encounter this error, we have a bug

src/client-side-encryption/autoEncrypter.ts
baileympearson: Can you clarify?  Why would we remove guards against invalid arguments?

src/client-side-encryption/autoEncrypter.ts
baileympearson: The types declared allow `CommandOptions` because that's currently what we pass in, but we only should need the BSON serialization options (and even those are questionable).

src/client-side-encryption/autoEncrypter.ts
nbbeeken: I think the argument validation is fine, a little unnecessary if we trust our TS. But I thought "clean up" would include making this MongoError subclass (like InvalidArgument)

src/client-side-encryption/autoEncrypter.ts
nbbeeken: I think this actually the source of that old sessions bug NODE-4896, if we didn't spread here the session wouldn't have made it to the operations. Do we want to make a change here or file a follow up to investigate?

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,665 @@
...
+ * | 4 | Trace |
+ */
+export type AutoEncryptionLoggerLevel =
+  (typeof AutoEncryptionLoggerLevel)[keyof typeof AutoEncryptionLoggerLevel];
+
+// Typescript errors if we index objects with `Symbol.for(...)`, so
+// to avoid TS errors we pull them out into variables.  Then we can type
+// the objects (and class) that we expect to see them on and prevent TS
+// errors.
+/** @internal */
```
nbbeeken, line 218: nit: Can you put the comment into the tsdoc on this variable?

src/client-side-encryption/autoEncrypter.ts
```diff
@@ -0,0 +1,665 @@
...
+
+  /**
+   * Used by devtools to enable decorating decryption results.
+   *
+   * When set and enabled, `decrypt` will automatically recursively
+   * traverse a decrypted document and if a field has been decrypted,
+   * it will mark it as decrypted.  Compass uses this to determine which
+   * fields were decrypted.
+   */
+  [kDecorateResult]?: boolean;
```
nbbeeken, line 253: I don't feel strongly but a wise Bailey once said optional booleans make an enum with 3 values. Any harm in setting this to false?

src/client-side-encryption/autoEncrypter.ts
nbbeeken: Do we want to make this change?

src/client-side-encryption/mongocryptdManager.ts
nbbeeken: This might be something that is captured in the specification (or should be). We can discuss the finer points in tech chat but for now let's leave something more descriptive than "perhaps questionable". I'm thinking let's capture the intention regardless of our takes on the approach. 

Something like:
> Add an empty error listener to squash child process errors
> If the process encountered an error we will learn of it when we fail to connect

src/client-side-encryption/mongocryptdManager.ts
```diff
@@ -32,27 +34,30 @@ export class MongocryptdManager {
...
-   * @param {Function} callback Invoked when we think a mongocryptd is up
    */
-  spawn(callback) {
+  spawn(callback: Callback<void>) {
     const cmdName = this.spawnPath || 'mongocryptd';
 
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const { spawn } = require('child_process') as typeof import('child_process');
+
     // Spawned with stdio: ignore and detatched:true
```
nbbeeken, line 51:
```suggestion
    // Spawned with stdio: ignore and detached: true
```

src/utils.ts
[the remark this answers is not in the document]
nbbeeken: Can you clarify? Why wouldn't it be `fromString(namespace?: string)`?

src/client-side-encryption/autoEncrypter.ts
baileympearson: No, the root cause was explicitly passing the session into driver helpers in the StateMachine: https://github.com/mongodb/libmongocrypt/commit/c9e44e7c0842683659e2a068696e6a712c33ab43

I checked in the `StateMachine` and the only four values we're accessing on the options object are the values we pass in here, so I removed the spreading.

src/utils.ts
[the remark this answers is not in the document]
baileympearson: Oh I misunderstood your question.  fixed

src/index.ts
nbbeeken: We can look into it, our exports are kept sorted but the "sections" are manually maintained.

src/client-side-encryption/autoEncrypter.ts
nbbeeken: Seems overkill fair enough

nbbeeken (approved):

PR #3788, opened 2023-07-28
feat(NODE-3920)!: validate options are not repeated in connection string

nbbeeken:
#### What is changing?

- Add validation for repeated options that aren't readPreferenceTags
- Fix bug with empty readPreferenceTags

None

See highlights

### Repeated options in connection string are no longer allowed

In order to avoid accidental misonfiguration the driver will no longer prioritize the first instance of an option provided on the URI. Instread repeated options that are not permitted to be repeated wil throw an error. This is to that connection strings that contain `"tls=true&tls=false` are no longer ambiguous.

### Empty readPreferenceTags in connection string fixed

The following connection string will now produce the following readPreferenceTags:

```ts
"mongodb://host?readPreferenceTags=region:ny&readPreferenceTags=rack:r1&readPreferenceTags=";
// client.options.readPreference.tags
[{ region: 'ny' }, { rack: 'r1' }, { }];
```

The empty `readPreferenceTags` allows drivers to still select a server if the leading tag conditions are not met. 


src/connection_string.ts
```diff
@@ -291,9 +294,17 @@ export function parseOptions(
...
+
+    if (!isReadPreferenceTags && values.length > 1) {
+      throw new MongoInvalidArgumentError(
+        `URI option "${key}" cannot appear more than once in the connection string`
+      );
+    }
 
-    if (values.includes('')) {
+    if (!isReadPreferenceTags && values.includes('')) {
       throw new MongoAPIError('URI cannot contain options with no value');
```
durran, line 308: I think for being a bit more helpful to users we should provide the `key` in the error message to be more clear which option had the empty value.

test/unit/connection_string.spec.test.ts
```diff
@@ -11,7 +11,6 @@ const skipTests = [
 
 describe('Connection String spec tests', function () {
   // TODO(NODE-3920): validate repeated options
```
durran, line 13: Can we remove this comment now?

durran (changes requested):

durran (approved):

PR #3790, opened 2023-08-01
feat(NODE-5233)!: prevent session from one client from being used on another

nbbeeken:
#### What is changing?

- Add a check for the same client reference
- Throw if they do not match

API docs on startSession updated

See highlights

### Using a session from another client causes an error

ClientSession instances from one client cannot be used in operation on another. This has not been a supported use case prior. However, it would not have prevented an operation from completing so we are making this change in a major version to prevent any accidental breakages. 


src/operations/execute_operation.ts
```diff
@@ -118,6 +118,8 @@ async function executeOperationAsync<
     throw new MongoExpiredSessionError('Use of expired sessions is not permitted');
   } else if (session.snapshotEnabled && !topology.capabilities.supportsSnapshotReads) {
     throw new MongoCompatibilityError('Snapshot reads require MongoDB 5.0 or later');
+  } else if (session.client !== client) {
+    throw new MongoRuntimeError('ClientSession must be from the same MongoClient');
```
baileympearson, line 122: Should this be a `MongoInvalidArgument` error?  We this error should be thrown when the user provides an invalid session as an argument.

baileympearson (changes requested):
questions from kickoff:

Did we decide referential equality was okay?
- best way to identify client? - instanceof is brittle, we can generate a unique id and/or use a symbol property; needs to account for legacy client usage
- what other identifier is there?
- what problems are there with checking referential equality?

Do we want to throw if a session from a LegacyMongoClient is passed into a MongoClient or vice versa? 

---

Session prose test 5 says the following: 

> Repeat the above for all methods that take a session parameter.

Did we decide not to implement this and if so, should we add more APIs to the test?  I don't think we need to exhaustively test the driver, but we should probably test more than a single `insertOne`.

---

I see the following as tests in the kickoff but not in the PR:

- Unit test can check when sessions are and are not supported
- Check when a session is and is not from the client passed to executeOperation

I thought the plan was to unit test this in some way (specifically to unit test execute operation).  Did the plan change?

nbbeeken:
From huddle:
- reference check good
- Add test for legacy client usage
- Add cursor and gridfs operations to the prose testing
- No unit tests

baileympearson (approved):

PR #3791, opened 2023-08-02
feat(NODE-3989)!: only accept true and false for boolean options

W-A-James:
#### What is changing?

- Removed support for deprecated alternate boolean values in connection strings

NODE-3989

### Boolean options only accept 'true' or 'false' in connection strings

Prior to this change, we accepted the values `'1', 'y', 'yes', 't'` as synonyms for `true` and `'-1', '0', 'f', 'n', 'no'` as synonyms for `false`. These have now been removed in an effort to make working with connection string options simpler.

```ts
// Incorrect
const client = new MongoClient('mongodb://localhost:27017?tls=1'); // throws MongoParseError

// Correct
const client  = new MongoClient('mongodb://localhost:27017?tls=true');
```


test/unit/assorted/uri_options.spec.test.ts
```diff
@@ -14,7 +14,10 @@ describe('URI option spec tests', function () {
     'tlsDisableCertificateRevocationCheck can be set to true',
     'tlsDisableCertificateRevocationCheck can be set to false',
     'tlsDisableOCSPEndpointCheck can be set to true',
-    'tlsDisableOCSPEndpointCheck can be set to false'
+    'tlsDisableOCSPEndpointCheck can be set to false',
+
+    // Skipped because we now only accept true/false for boolean options
```
durran, line 19: Maybe nice to note around here that the URI options spec expects warnings currently for invalid options and the Node driver throws.

test/unit/connection_string.test.ts
```diff
@@ -208,13 +208,40 @@ describe('Connection String', function () {
...
+                parseOptions(connString);
+              }).to.throw(MongoParseError);
+            });
+            break;
+          default:
+            it(`parses as ${expectation}`, function () {
+              const options = parseOptions(connString);
+              expect(options).to.have.property('retryWrites', expectation);
+            });
+        }
```
durran, line 242:
```suggestion
        if (expectation === 'error') {
            it('throws MongoParseError', function () {
              expect(() => {
                parseOptions(connString);
              }).to.throw(MongoParseError);
            });
        } else {
            it(`parses as ${expectation}`, function () {
              const options = parseOptions(connString);
              expect(options).to.have.property('retryWrites', expectation);
            });
        }
```

I think as there are only 2 cases here `if...else` makes a bit more sense and is 1 less line of code.

durran (changes requested):

test/unit/connection_string.test.ts
```diff
@@ -208,13 +208,38 @@ describe('Connection String', function () {
...
+      { value: 'y', expectation: 'error' },
+      { value: 'yes', expectation: 'error' },
+      { value: 'no', expectation: 'error' },
+      { value: 'unknown', expectation: 'error' }
+    ];
+    for (const { value, expectation } of valuesExpectations) {
+      const connString = `mongodb://hostname?retryWrites=${value}`;
+      context(`when provided '${value}'`, function () {
+        if (expectation === 'error') {
+            it('throws MongoParseError', function () {
```
durran, line 230: Think there's lint errors here now.

durran (changes requested):
Lint errors.

durran (approved):

nbbeeken (approved):

PR #3792, opened 2023-08-02
feat(NODE-5444)!: emit deprecation warning for useNewUrlParser and useUnifiedTopology

nbbeeken:
#### What is changing?

- emit deprecation warning for useNewUrlParser and useUnifiedTopology
- unit test for warning

Yes. Docs changes is filed on the ticket.

See below

###  `useNewUrlParser` and `useUnifiedTopology` emit deprecation warnings

These options were removed in 4.0.0 but continued to be parsed and silently left unused. We have now added a deprecation warning through Node.js' [warning system](https://nodejs.org/api/process.html#event-warning) to prepare for finally removing these options in the _next_ major release. 


src/connection_string.ts
```diff
@@ -1223,9 +1223,15 @@ export const OPTIONS = {
   pfx: { type: 'any' },
   secureProtocol: { type: 'any' },
   index: { type: 'any' },
-  // Legacy Options, these are unused but left here to avoid errors with CSFLE lib
-  useNewUrlParser: { type: 'boolean' } as OptionDescriptor,
-  useUnifiedTopology: { type: 'boolean' } as OptionDescriptor,
+  // Legacy options from v3 era
+  useNewUrlParser: {
+    type: 'boolean',
+    deprecated: 'useNewUrlParser is no longer used it can be removed'
```
durran, line 1229: Just a little not for these 2 messages, but maybe something noting that it will be removed in the next major release?

"option is a no-op and will be removed in the next major release". Thoughts?

durran (changes requested):

alexbevi:
Fun fact - [Mongoose removed these options with Mongoose 6.0.0](https://mongoosejs.com/docs/migrating_to_6.html#no-more-deprecation-warning-options)

src/connection_string.ts
alexbevi: _"useNewUrlParser has been a no-op since Node.js Driver version x.y.z and will be removed in the next major version"_ maybe

nbbeeken: _`useNewUrlParser` has no effect since Node.js Driver version 4.0.0 and will be removed in the next major version_

Does this look good to y'all? I have a _very_ slight preference to not phrasing it as "no-op"

alexbevi (approved):

src/connection_string.ts
```diff
@@ -1223,9 +1223,17 @@ export const OPTIONS = {
...
   index: { type: 'any' },
-  // Legacy Options, these are unused but left here to avoid errors with CSFLE lib
-  useNewUrlParser: { type: 'boolean' } as OptionDescriptor,
-  useUnifiedTopology: { type: 'boolean' } as OptionDescriptor,
+  // Legacy options from v3 era
+  useNewUrlParser: {
+    type: 'boolean',
+    deprecated:
+      'useNewUrlParser has no effect since Node.js Driver version 4.0.0 and will be removed in the next major version'
+  } as OptionDescriptor,
```
baileympearson, line 1231: just curious - how come we need to cast `as OptionDescriptor` here?

test/unit/connection_string.test.ts
```diff
@@ -761,4 +764,40 @@ describe('Connection String', function () {
     expect(() => new MongoClient('mango://localhost:23')).to.throw(/Invalid scheme/i);
     expect(() => new MongoClient('mango+srv://localhost:23')).to.throw(/Invalid scheme/i);
   });
+
+  describe('when deprecated options are used', () => {
+    it('useNewUrlParser emits a warning', async () => {
```
baileympearson, line 769: Should we only be emitting a single warning per-option?  The current implementation will emit a warning every time the option is specified.

Related question: do we have a practice of when we use `emitWarning` versus `emitWarningOnce`?

src/connection_string.ts
nbbeeken: Because the key doesn't appear in MongoClientOptions it doesn't get typed here b/c OPTIONS is typed: `Record<keyof MongoClientOptions, OptionDescriptor>`

test/unit/connection_string.test.ts
nbbeeken: What do you mean by every time it is specified?

I think emitWarning makes sense here because, in the case of creating multiple MongoClients, you can use Node's `--trace-warning` argument to get a stack trace pointing to each usage.

baileympearson (approved):

PR #3793, opened 2023-08-02
refactor(NODE-5473): Remove Unused Callback Inheritance in Operations Layer

malikj2000:
### Description
removed unused callback inheritance in all operation files

#### What is changing?
asyncified eval.ts
removed unimplemented executeCallback method from all operation files
removed callback classes from AbstractOperation and CommandOperation

None

further asyncifying the driver

### DIFF
--- a/src/operations/command.ts
+++ b/src/operations/command.ts
@@ -7,7 +7,6 @@ import type { Server } from '../sdam/server';
 import { MIN_SECONDARY_WRITE_WIRE_VERSION } from '../sdam/server_selection';
 import type { ClientSession } from '../sessions';
 import {
-  type Callback,
   commandSupportsReadConcern,
   decorateWithExplain,
   maxWireVersion,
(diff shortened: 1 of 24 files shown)


src/operations/eval.ts
```diff
@@ -5,16 +5,15 @@ import { MongoServerError } from '../error';
 import { ReadPreference } from '../read_preference';
 import type { Server } from '../sdam/server';
 import type { ClientSession } from '../sessions';
-import type { Callback } from '../utils';
-import { CommandCallbackOperation, type CommandOperationOptions } from './command';
+import { CommandOperation, type CommandOperationOptions } from './command';
 
 /** @public */
 export interface EvalOptions extends CommandOperationOptions {
```
nbbeeken, line 11: This operation is unused but because these are public APIs we have to make sure the change is visible in the changelog. Can you file a ticket to remove the EvalOperation and related APIs.

But we can keep the changes we're making here so that we can still remove the callback super class.

src/operations/command.ts
```diff
@@ -156,22 +155,3 @@ export abstract class CommandOperation<T> extends AbstractCallbackOperation<T> {
     return server.commandAsync(this.ns, cmd, options);
   }
 }
-
-/** @internal */
-export abstract class CommandCallbackOperation<T = any> extends CommandOperation<T> {
```
nbbeeken, line 161: Commenting here, but unrelated. We can also remove AbstractCallbackOperation as well now correct?

nbbeeken (changes requested):

PR #3797, opened 2023-08-04
feat(NODE-4986)!: remove callbacks from ClientEncryption encrypt, decrypt, and createDataKey

nbbeeken:
#### What is changing?

- removed callbacks from encrypt, decrypt, and createDataKey
- Changed return type createDataKey to Binary which lines up with the runtime return value

Release highlights

We do not offer callbacks APIs anymore

### Callbacks removed from ClientEncryption's `encrypt`, `decrypt`, and `createDataKey` methods

Driver v5 dropped support for callbacks in asynchronous functions in favor of returning promises.  We are now removing support for callbacks from the `ClientEncryption` class.

#### `createDataKey` return type fix

The Typescript for `createDataKey` incorrectly declared the result to be a `DataKey` but the method actually returns the DataKey's `insertedId`.


src/client-side-encryption/client_encryption.ts
```diff
@@ -213,8 +192,8 @@ export class ClientEncryption implements StateMachineExecutable {
       tlsOptions: this._tlsOptions
     });
 
-    // @ts-expect-error We did not convert promiseOrCallback to TS
```
baileympearson, line 216: Just FYI - we have a promisified `executeAsync` on the state machine already.

.evergreen/run-custom-csfle-tests.sh
```diff
@@ -57,8 +57,6 @@ bash ./etc/build-static.sh
 
 npm run rebuild # just in case this is necessary?
 
-ls
-ls lib
```
nbbeeken, line 61: Easy fix, `ls lib` broken by the recent merge

src/index.ts
```diff
@@ -33,7 +33,8 @@ export {
   MaxKey,
   MinKey,
   ObjectId,
-  Timestamp
+  Timestamp,
+  UUID
```
nbbeeken, line 37: This seems like an oversight, now that we have an API that advertises this as a return type, it seems more worth it than before that this is available as a top-level export

src/client-side-encryption/client_encryption.ts
```diff
@@ -213,32 +191,17 @@ export class ClientEncryption implements StateMachineExecutable {
...
-    return promiseOrCallback(callback, cb => {
-      stateMachine.execute<DataKey>(this, context, (err, dataKey) => {
-        if (err || !dataKey) {
-          cb(err, null);
-          return;
-        }
+    const updatedDataKey = await stateMachine.executeAsync<DataKey>(this, context);
```
nbbeeken, line 194: I have to choose another name here because we were shadowing. Up for any suggestions

test/integration/node-specific/client_encryption.test.ts
```diff
@@ -496,6 +491,26 @@ describe('ClientEncryption integration tests', function () {
...
+
+    it('returns a UUID instance', async () => {
+      const dataKey = await clientEncryption.createDataKey('local', {
+        name: 'local',
+        kmsProviders: { local: { key: Buffer.alloc(96) } }
+      });
+
+      expect(dataKey).to.be.instanceOf(UUID);
+    });
+  });
```
nbbeeken, line 512: New test

src/client-side-encryption/client_encryption.ts
baileympearson: maybe just remove the first declaration and name this `dataKey`?  it's not proving much value imo and it actually makes this more confusing, because the data key is created by `stateMachine.execute`.

```typescript
    const dataKeyBson = serialize({ provider, masterKey: options.masterKey });
```

test/integration/node-specific/crypto_callbacks.test.ts
```diff
@@ -25,24 +25,35 @@ const SKIP_AWS_TESTS = [AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION, AW
...
 
   beforeEach(function () {
     if (SKIP_AWS_TESTS) {
       this.currentTest?.skip();
       return;
     }
 
-    sandbox.restore();
+    sandbox = sinon.createSandbox();
+
```
baileympearson, line 45: Is it possible to stub the whole object in one go? 

`sandbox.spy(cryptoCallbacks)`

baileympearson (changes requested):

package.json
```diff
@@ -98,7 +98,7 @@
     "js-yaml": "^4.1.0",
     "mocha": "^10.2.0",
     "mocha-sinon": "^2.1.2",
-    "mongodb-client-encryption": "^6.0.0-alpha.1",
+    "mongodb-client-encryption": "6.0.0-alpha.1",
```
durran, line 101: Can we bump this to alpha 2 now that it's out?

nbbeeken: alpha.2 introduced the bug that's making the CI red, rolling back fixes it locally but all of EVG disagrees 😅 I'm not sure how the code is getting a hold of the alpha.2 version, I thought we only installed from devDeps now.

nbbeeken: Context: https://jira.mongodb.org/browse/NODE-5516

Now that we track the package in our devDeps we can get rid of the peppering of manual installations. cc @baileympearson

PR #3691, opened 2023-05-30
feat(NODE-5008): add zstd and kerberos to peer deps

durran:
### Description

Adds zstd and kerberos to peer dependencies.

#### What is changing?

Sets @mongodb-js/zstd and kerberos in the peer dependencies to their latest versions and sets them as optional.

##### Is there new documentation needed for these changes?

None

#### What is the motivation for this change?

NODE-5008

### Double check the following

- [x] Ran `npm run check:lint` script
- [x] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [x] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`
- [] Changes are covered by tests
- [ ] New TODOs have a related JIRA ticket

PR #3798, opened 2023-08-07
feat(NODE-5490)!: bump kerberos compatibility to ^2.0.1

durran:
### Description

Updates kerberos dependency to ^2.0.1

#### What is changing?

The optional peer dependency has already changed, this updates the test script to install it instead of the beta. There were no specific tests targeting kerberos 1.x and the filter in the evergreen generation script was removed even though it was no longer actually doing anything.

None

NODE-5490

#### Kerberos optional peer dependency minimum version raised to 2.0.1

This was done in https://github.com/mongodb/node-mongodb-native/pull/3691/commits/5fcdf93df573739904f405177413fb327a5015d6 (https://github.com/mongodb/node-mongodb-native/pull/3691)

### DIFF
--- a/.evergreen/generate_evergreen_tasks.js
+++ b/.evergreen/generate_evergreen_tasks.js
@@ -774,13 +774,6 @@ for (const variant of BUILD_VARIANTS.filter(
   );
 }
 
-// TODO(NODE-5021): Drop support for Kerberos 1.x on in 6.0.0
-for (const variant of BUILD_VARIANTS.filter(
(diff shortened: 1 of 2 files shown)


dariakp:
@durran Just a note that the release highlight won't get pulled in by the automation on a non-feat/non-fix commit, so we either have to change this commit type or move the highlight to another breaking change PR; since there was never a `feat!` corresponding to this change, I think the best thing to do would be to make this commit a `feat(NODE-)!: ...`, and just point to the commit that actually introduced the change in the PR description.

PR #3800, opened 2023-08-07
feat(NODE-5484)!: mark MongoError for internal use and remove Node14 cause assignment logic

W-A-James:
#### What is changing?
- `MongoError` and all `MongoError` subclasses now have documentation stating that they are no longer covered under semver
- `MongoError` constructor now defers to Node16+'s Error constructor to set the `cause` field
- `MongoCryptError` is now a subclass of `MongoError`
- Add unit tests checking that all subclasses of `MongoCryptError` are subclasses of `MongoError`
- Update etc/notes/error.md to add `MongoCrypt` to error hierarchy

No

NODE-5484

### `MongoError` and its subclasses now have internal constructors

`MongoError` and its subclasses are not meant to be constructed by users as they are thrown within the driver on specific error conditions to allow users to react to these conditions in ways which match their use cases. The constructors for these types are now subject to change outside of major versions and their documentation updated to reflect this.

### `MongoCryptError` is now a subclass of `MongoError`

Since `MongoCryptError` made use of NodeJS 16's `Error` API, it has long supported setting the `Error.cause` field using options passed in via the constructor. Now that Node 16 is our minimum supported version, `MongoError` has been modified to make use of this API as well, allowing us to let `MongoCryptError` subclass from it directly.


src/error.ts
```diff
@@ -108,6 +108,10 @@ export interface ErrorDescription extends Document {
   errInfo?: Document;
 }
 
+interface OptionsWithCause {
+  cause?: Error;
+}
+
```
W-A-James, line 114: Remove this

src/error.ts
```diff
@@ -692,9 +713,9 @@ export class MongoMissingCredentialsError extends MongoAPIError {
  * @category Error
  */
 export class MongoMissingDependencyError extends MongoAPIError {
-  constructor(message: string, { cause }: { cause?: Error } = {}) {
-    super(message);
-    if (cause) this.cause = cause;
+  /** @internal */
+  constructor(message: string, options: { cause?: Error } = {}) {
+    super(message, options);
   }
```
addaleax, line 719: What exactly is it that adding `@internal` is supposed to achieve? It won’t prevent users from creating class instances, it’s only going to make the class look like it doesn’t have its own constructor, i.e. uses the `constructor(...args) { super(...args); }` one.

If you want to prevent users from creating instances of these classes, you may need to do something like

```js
export class MongoSomethingError extends MongoError  {
  private constructor(...) {}

  /** @internal */
  static create(...args) { return new this(...args); }
}
```

W-A-James: We're doing this in an attempt to communicate that these classes shouldn't be instantiated by users. It's a pattern we have a lot elsewhere in the driver that probably could use some work with regards to getting that message across.

Wouldn't this solution also have the same effect? The user might not have documentation for the static `create` method, but they would still be able to call it on their end, right?

addaleax: Yeah, I get the intention here, and I agree that it's a goal that's worth putting effort into. :)

The problem is specifically with marking constructors as `@internal`, which is _different_ from all other methods, because that it does _not_ make it appear to consumers of your .d.ts file that there is no publicly accessible constructor; instead, the TS definition effectively says that "the constructor of this class has the same signature as the  constructor of the parent class".

That's different with a static method; yeah, of course users could still call it by casting the class to `as any` or just not using TS validation at all, but at least it would not show up in the generated TS definitions.

W-A-James: Okay, I see what you're saying now. I'll check with the team to see if we want to start with that new pattern here or deal with it in a separate ticket that takes care of this across the driver. Thanks for the clarification!

baileympearson (changes requested):

etc/notes/errors.md
```diff
@@ -144,6 +136,21 @@ These are errors which originate from faulty environment setup.
...
+- #### MongoCryptInvalidArgumentError
+  - Thrown when an invalid argument has been provided to an encryption API
+- #### MongoCryptInvalidCreateDataKeyError
+  - Thrown when `ClientEncryption.createEncryptedCollection()` failed to create data keys
+- #### MongoCryptInvalidCreateEncryptedCollectionError
+  - Thrown when `ClientEncryption.createEncryptedCollection()` failed to create a collection
+- #### MongoCryptInvalidCreateAzureKMSRequestError
+  - Thrown when auto-fetching KMS credentials fails to auto-refresh Azure KMS credentials
+- #### MongoCryptKMSRequestNetworkTimeoutError
+  - Thrown when auto-fetching KMS credentials times out
```
baileympearson, line 152:
```suggestion
- #### MongoCryptInvalidCreateAzureKMSRequestError
  - Thrown when the driver encounters an error when fetching Azure KMS credentials
- #### MongoCryptKMSRequestNetworkTimeoutError
  - Thrown when the HTTP request to the IDMS server times out when fetching Azure KMS credentials
```

baileympearson (changes requested):
I think the error descriptions were a little clunky - can we reword them slighty?

baileympearson (dismissed):

nbbeeken (changes requested):

test/unit/error.test.ts
```diff
@@ -99,7 +99,7 @@ describe('MongoErrors', () => {
       expect(err).to.be.an.instanceof(Error);
       expect(err.name).to.equal('MongoError');
       expect(err.message).to.equal(errorMessage);
-      expect(err).to.have.property('cause', inputError);
+      expect(err).to.not.have.property('cause');
```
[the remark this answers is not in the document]
W-A-James: Updated errors. Going through the tests to update their assertions now.

src/error.ts
```diff
@@ -139,11 +139,11 @@ export class MongoError extends Error {
...
    * @remarks
    * This class is only meant to be constructed within the driver. This constructor is
-   * not subject to semantic versioning compatiblity guarantees and may change at any time.
+   * not subject to semantic versioning compatibility guarantees and may change at any time.
    *
    * @public
    **/
-  constructor(message: string | Error, options?: { cause?: Error }) {
+  constructor(message: string, options?: { cause?: Error }) {
     super(MongoError.buildErrorMessage(message), options);
```
nbbeeken, line 147: `buildErrorMessage` relied on being passed the potential AggregateError, this change should result in some test failures.

src/cmap/connect.ts
```diff
@@ -502,7 +502,7 @@ function makeSocks5Connection(options: MakeConnectionOptions, callback: Callback
 function connectionFailureError(type: ErrorHandlerEventName, err: Error) {
   switch (type) {
     case 'error':
-      return new MongoNetworkError(err);
+      return new MongoNetworkError('error', { cause: err });
```
nbbeeken, line 505: I was going to suggest stealing the message from the error here but I think this is our aggregate error case so the message will be an empty string.

Maybe we allow the string argument to be nullish, then buildMessage's job will be to obtain it from the cause, or we make this say "error during connection establishment"

nbbeeken (changes requested):

src/error.ts
baileympearson: Is this comment suggesting we fix the tests or that we revert the change?

src/cmap/connect.ts
baileympearson: I'm a fan of the "make all consumers specify the error message".  It's more explicit and it makes it easier to reason about what error messages are for given errors.

If we want to handle aggregate errors, there's nothing stopping us from using `buildErrorMessage` wherever we want:

```typescript
      return new MongoNetworkError(MongoError.buildErrorMessage(err), { cause: err });
```

src/error.ts
nbbeeken: Either, I don't feel strongly. I don't see test failures which is intriguing, maybe because the tests are still just passing in an error as the first argument?

nbbeeken (approved):

PR #3805, opened 2023-08-10
test(NODE-3049): drivers atlas testing

durran:
### Description

This PR makes the necessary changes in the driver for it to be able to be tested by Astrolabe.

#### What is changing?

- Adds a new task `test:drivers-atlas-testing` to run the unified test runner against the Astrolabe tests.
- Updates the test runner config to use the URI provided by Astrolabe when present in the environment.
- Creates an astrolabe results write to write the spec defined events.json and results.json
- Updates the entities map to handle new new entites for `iterations`, `successes`, `events`, `failures`, and `errors`.
- Creates a registry to handle mapping events between the unified client and the entities map.
- Adds a new unified operation `loop`, as defined per the [Unified Test Format spec](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#loop)

Related drivers-atlas-testing changes: https://github.com/mongodb-labs/drivers-atlas-testing/pull/173

None

NODE-3049

### DIFF
--- a/test/tools/runner/config.ts
+++ b/test/tools/runner/config.ts
@@ -159,6 +159,10 @@ export class TestConfiguration {
   }
 
   newClient(dbOptions?: string | Record<string, any>, serverOptions?: Record<string, any>) {
+    if (process.env.DRIVERS_ATLAS_TESTING_URI) {
+      return new MongoClient(process.env.DRIVERS_ATLAS_TESTING_URI);
+    }
+
     serverOptions = Object.assign({}, getEnvironmentalOptions(), serverOptions);
 
     // support MongoClient constructor form (url, options) for `newClient`
@@ -257,6 +261,10 @@ export class TestConfiguration {
       proxyURIParams: this.options.proxyURIParams,
       ...options
     };
+
+    if (process.env.DRIVERS_ATLAS_TESTING_URI) {
+      return process.env.DRIVERS_ATLAS_TESTING_URI;
+    }
 
     const FILLER_HOST = 'fillerHost';
 
--- a/test/tools/runner/config.ts
+++ b/test/tools/runner/config.ts
@@ -159,10 +159,6 @@ export class TestConfiguration {
   }
 
   newClient(dbOptions?: string | Record<string, any>, serverOptions?: Record<string, any>) {
-    if (process.env.DRIVERS_ATLAS_TESTING_URI) {
-      return new MongoClient(process.env.DRIVERS_ATLAS_TESTING_URI);
-    }
-
     serverOptions = Object.assign({}, getEnvironmentalOptions(), serverOptions);
 
     // support MongoClient constructor form (url, options) for `newClient`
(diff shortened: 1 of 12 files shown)


test/tools/runner/config.ts
```diff
@@ -159,6 +159,10 @@ export class TestConfiguration {
   }
 
   newClient(dbOptions?: string | Record<string, any>, serverOptions?: Record<string, any>) {
```
baileympearson, line 161: thoughts on a dedicated subclass here?

```typescript
class AstrolabeTestConfiguration extends TestConfiguration {
  override newClient() {
	return new MongoClient(process.env.DRIVERS_ATLAS_TESTING_URI);
  }

  override url() {
    return process.env.DRIVERS_ATLAS_TESTING_URI;
  }
}

// ... in configuration.js
const testConfigBeforeHook = async function () {
  if (process.env.DRIVERS_ATLAS_TESTING_URI) {
    this.configuration = new AstrolabeTestConfiguration(process.env.DRIVERS_ATLAS_TESTING_URI, {});
    return;
  }
  ...
}
```

test/tools/unified-spec-runner/operations.ts
```diff
@@ -377,6 +387,77 @@ operations.set('listIndexes', async ({ entities, operation }) => {
...
+        if (storeFailuresAsEntity && error instanceof AssertionError) {
+          entities
+            .getEntity('failures', storeFailuresAsEntity)
+            .push({ error: error.message, time: Date.now() });
+        } else if (storeErrorsAsEntity && !(error instanceof MongoError)) {
+          // Checking not a MongoError ensures it's coming from the test runner.
+          entities
+            .getEntity('errors', storeErrorsAsEntity)
+            .push({ error: error.message, time: Date.now() });
+        }
```
baileympearson, line 455: optional cleanup: the first condition is redundant
```suggestion
        if (storeFailuresAsEntity) {
       		if (error instanceof AssertionError) {
	          entities
	            .getEntity('failures', storeFailuresAsEntity)
	            .push({ error: error.message, time: Date.now() });
	         }
	         else if (!(error instanceof MongoError)) {
		         entities
	           		.getEntity('errors', storeErrorsAsEntity)
	            	.push({ error: error.message, time: Date.now() });
	         }
        }
```

baileympearson (changes requested):
changes look good, just a few small suggestions

test/tools/unified-spec-runner/operations.ts
durran: How is the first condition redundant? We only store `failures` when `storeFailuresAsEntity` is defined and only store `errors` when `storeErrorsAsEntity` is defined. The suggestion would fail if `storeErrorsAsEntity` was not defined.

baileympearson: Oh, I didn't read it closely enough.

test/tools/runner/config.ts
durran: Nice suggestion - I've changed to use a subclass now.

baileympearson (approved):

baileympearson (approved):

PR #3808, opened 2023-08-10
fix(NODE-4788)!: use implementer Writable methods for GridFSBucketWriteStream 

nbbeeken:
#### What is changing?

- Remove overrides of stream methods
- Add overrides of _X stream methods that control the underlying implmentations
- Remove manual event emitting
- Ensure callback is always invoked asynchronously (nextTick)

- TBD

Node.js Streams are supposed to be implemented with the _X methods, the write & end functions are maintained by Node.js, the _ methods define the implementation of a stream.

### Corrected `GridFSBucketWriteStream`'s `Writable` method overrides and event emission

Our implementation of a writeable stream for GridFSBucketWriteStream mistakenly overrode the `write()` and `end()` methods, as well as, manually emitted `'close'`, `'drain'`, `'finish'` events. Per Node.js documentation, these methods and events are intended for the Node.js stream implementation to provide, and an author of a stream implementation is supposed to override `_write`, `_final`, and allow Node.js to manage event emitting. 

Since the API is still a `Writable` stream most usages will continue to work with no changes, the `.write()` and `.end()` methods are still available and take the same arguments. The breaking change relates to the improper manually emitted event listeners that are now handled by Node.js. **The `'finish'` and `'drain'` events will no longer receive the `GridFSFile` document as an argument** (this is the document inserted to the bucket's files collection after all chunks have been inserted). Instead, it will be available on the stream itself as a property: `gridFSFile`. 

```ts
// If our event handler is declared as a `function` "this" is bound to the stream.
fs.createReadStream('./file.txt')
  .pipe(bucket.openUploadStream('file.txt'))
  .on('finish', function () {
    console.log(this.gridFSFile)
  })

// If our event handler is declared using big arrow notation,
// the property is accessible on a scoped variable
const uploadStream = bucket.openUploadStream('file.txt')
fs.createReadStream('./file.txt')
  .pipe(uploadStream)
  .on('finish', () => console.log(uploadStream.gridFSFile))
```

Since the class no longer emits its own events: static constants `GridFSBucketWriteStream.ERROR`, `GridFSBucketWriteStream.FINISH`, `GridFSBucketWriteStream.CLOSE` have been removed to avoid confusion about the source of the events and the arguments their listeners accept.

### Fix manually emitted events from `GridFSBucketReadStream`

The `GridFSBucketReadStream` internals have also been corrected to no longer emit events that are handled by Node's stream logic. Since the class no longer emits its own events: static constants `GridFSBucketReadStream.ERROR`, `GridFSBucketReadStream.DATA`, `GridFSBucketReadStream.CLOSE`, and `GridFSBucketReadStream.END` have been removed to avoid confusion about the source of the events and the arguments their listeners accept.

### DIFF
--- a/src/gridfs/upload.ts
+++ b/src/gridfs/upload.ts
@@ -138,7 +138,9 @@ export class GridFSBucketWriteStream extends Writable {
   }
 
   /**
-   * The stream is considered constructed when the indexes ßare done being created
+   * @internal
+   *
+   * The stream is considered constructed when the indexes are done being created
    */
   override _construct(callback: (error?: Error | null) => void): void {
     if (this.bucket.s.checkedIndexes) {
@@ -148,12 +150,12 @@ export class GridFSBucketWriteStream extends Writable {
   }
 
   /**
+   * @internal
    * Write a buffer to the stream.
    *
    * @param chunk - Buffer to write
-   * @param encodingOrCallback - Optional encoding for the buffer
+   * @param encoding - Optional encoding for the buffer
    * @param callback - Function to call when the chunk was added to the buffer, or if the entire chunk was persisted to MongoDB if this chunk caused a flush.
-   * @returns False if this write required flushing a chunk to MongoDB. True otherwise.
    */
   override _write(
     chunk: Buffer | string,
@@ -163,6 +165,7 @@ export class GridFSBucketWriteStream extends Writable {
     doWrite(this, chunk, encoding, callback);
   }
 
+  /** @internal */
   override _final(callback: (error?: Error | null) => void): void {
     if (this.state.streamEnd) {
       return process.nextTick(callback);
--- a/src/gridfs/upload.ts
+++ b/src/gridfs/upload.ts
@@ -3,7 +3,7 @@ import { Writable } from 'stream';
 import type { Document } from '../bson';
 import { ObjectId } from '../bson';
 import type { Collection } from '../collection';
-import { type AnyError, MongoAPIError, MONGODB_ERROR_CODES, MongoError } from '../error';
+import { MongoAPIError, MONGODB_ERROR_CODES, MongoError } from '../error';
 import type { Callback } from '../utils';
 import type { WriteConcernOptions } from '../write_concern';
 import { WriteConcern } from './../write_concern';
@@ -38,7 +38,7 @@ export interface GridFSBucketWriteStreamOptions extends WriteConcernOptions {
  * Do not instantiate this class directly. Use `openUploadStream()` instead.
  * @public
  */
-export class GridFSBucketWriteStream extends Writable implements NodeJS.WritableStream {
+export class GridFSBucketWriteStream extends Writable {
   bucket: GridFSBucket;
   chunks: Collection<GridFSChunk>;
   filename: string;
@@ -59,15 +59,7 @@ export class GridFSBucketWriteStream extends Writable implements NodeJS.Writable
   };
   writeConcern?: WriteConcern;
 
-  /** @event */
-  static readonly CLOSE = 'close';
-  /** @event */
-  static readonly ERROR = 'error';
-  /**
-   * `end()` was called and the write stream successfully wrote the file metadata and all the chunks to MongoDB.
-   * @event
-   */
-  static readonly FINISH = 'finish';
+  fileMetadata: GridFSFile | null = null;
 
   /**
    * @param bucket - Handle for this stream's corresponding bucket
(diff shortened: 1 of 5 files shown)


test/integration/gridfs/gridfs_stream.test.js
```diff
@@ -874,134 +798,6 @@ describe('GridFS Stream', function () {
     }
   });
 
-  /**
-   * NODE-822 GridFSBucketWriteStream end method does not handle optional parameters
-   */
-  it('should correctly handle calling end function with only a callback', {
```
nbbeeken, line 880: Optional arg handling is node's problem now

test/integration/gridfs/gridfs_stream.test.js
```diff
@@ -874,134 +798,6 @@ describe('GridFS Stream', function () {
...
-
-  describe('upload stream end()', () => {
-    let client, db;
-
-    afterEach(async () => {
-      sinon.restore();
-      await client.close();
-    });
-
-    it('should not call the callback on repeat calls to end', {
```
nbbeeken, line 940: Calling end more than once makes node throw an error, so we don't have special handling for either of these tests

test/integration/gridfs/gridfs_stream.test.js
```diff
@@ -545,51 +474,46 @@ describe('GridFS Stream', function () {
...
-
-        // Wait for stream to finish
-        uploadStream.once('finish', function () {
-          const id = uploadStream.id;
-          const downloadStream = bucket.openDownloadStream(id);
-          const finished = {};
-          downloadStream.on('data', function () {
-            expect.fail('Should be unreachable');
-          });
+      const bucket = new GridFSBucket(db, { bucketName: 'gridfsdestroy', chunkSizeBytes: 10 });
```
nbbeeken, line 477: This is an unindent for client hook clean up

test/integration/gridfs/gridfs_stream.test.js
```diff
@@ -432,106 +435,32 @@ describe('GridFS Stream', function () {
     }
   });
 
-  /**
-   * Aborting an upload
-   *
-   * @example-class GridFSBucketWriteStream
-   * @example-method abort
-   */
   it('Aborting an upload', {
```
durran, line 438:
```suggestion
  it('writing to an aborted stream throws API error', {
```

src/gridfs/upload.ts
```diff
@@ -115,6 +137,16 @@ export class GridFSBucketWriteStream extends Writable implements NodeJS.Writable
     }
   }
 
+  /**
+   * The stream is considered constructed when the indexes ßare done being created
```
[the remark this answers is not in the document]
nbbeeken: fixed

src/gridfs/upload.ts
```diff
@@ -159,76 +189,15 @@ export class GridFSBucketWriteStream extends Writable implements NodeJS.Writable
...
 }
 
-function __handleError(
-  stream: GridFSBucketWriteStream,
-  error: AnyError,
-  callback?: Callback
-): void {
+function handleError(stream: GridFSBucketWriteStream, error: Error, callback: Callback): void {
   if (stream.state.errored) {
+    process.nextTick(callback);
```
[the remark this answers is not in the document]
nbbeeken: That is more of a style/correctness improvement. Callbacks shouldn't be invoked synchronously within the function that was given the callback because it means it won't return until the callback has finished running.

```js
stream._write(chunk, encoding, (err) => {
  console.log('finished write')
  for (;;) {}
})
console.log('started write')
```

Take the code above, because `_write` is a callback-accepting function it should always return once it is done setting up the write, and the callback is invoked when the write is done. If `_write` doesn't move the callback's execution to a separate tick it is including the callback's execution in its own execution. In this example, we'd never reach "started write" even though we have started and potentially finished the write.

test/integration/gridfs/gridfs_stream.test.js
nbbeeken: fixed

test/integration/gridfs/gridfs_stream.test.js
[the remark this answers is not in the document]
nbbeeken: Updated, I had to async/await it to fully grok it, which made me realize we missed fixing the emits inside the download stream. I've updated the release highlights to include those changes.

src/gridfs/upload.ts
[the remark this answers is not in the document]
durran: Thanks for the explanation.

durran (approved):

PR #3810, opened 2023-08-10
fix(NODE-5495): do not emit deprecation warning when tlsCertificateKeyFile is specified and tlsCertificateFile is not

baileympearson:
#### What is changing?

This PR prevents the driver from emitting a deprecation warning when `tlsCertificateKeyFile` is specified and `tlsCertificateFile` is not.

No.

### Fixed accidental deprecation warning

Because of internal options handling, a deprecation was emitted for `tlsCertificateFile` when using `tlsCertificateKeyFile`. That has been corrected.

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -347,10 +347,10 @@ export function parseOptions(
     allProvidedOptions.set(key, values);
   }
 
-  if (
+  const didMapTLSCertificateFile =
     allProvidedOptions.has('tlsCertificateKeyFile') &&
-    !allProvidedOptions.has('tlsCertificateFile')
-  ) {
+    !allProvidedOptions.has('tlsCertificateFile');
+  if (didMapTLSCertificateFile) {
     allProvidedOptions.set('tlsCertificateFile', allProvidedOptions.get('tlsCertificateKeyFile'));
   }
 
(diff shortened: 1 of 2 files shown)


src/connection_string.ts
```diff
@@ -347,10 +347,10 @@ export function parseOptions(
...
 
-  if (
+  const didMapTLSCertificateFile =
     allProvidedOptions.has('tlsCertificateKeyFile') &&
-    !allProvidedOptions.has('tlsCertificateFile')
-  ) {
+    !allProvidedOptions.has('tlsCertificateFile');
+  if (didMapTLSCertificateFile) {
     allProvidedOptions.set('tlsCertificateFile', allProvidedOptions.get('tlsCertificateKeyFile'));
   }
```
nbbeeken, line 355: Possibly missing something here, why are we still setting the deprecated option? can't we change this to set `tlsCertificateKeyFile`

nbbeeken (changes requested):

src/connection_string.ts
baileympearson: If we hit this logic, `tlsCertificateKeyFile` is already set.  This logic was explicitly added in https://jira.mongodb.org/browse/NODE-3591

The bug here is specifically when the user sets `tlsCertificateKeyFile` and not `tlsCertificateFile `, because we assign `tlsCertificateKeyFile` to `tlsCertificateFile ` and then check if the user has set a deprecated option.

I'm not sure what would happen if just removed this logic but I'm guessing we'd reintroduce the bug Daria fixed in 3591.  Ultimately, I'd rather just fix it like this because this is the 5.x branch and the logic is removed entirely from 6.x.

nbbeeken (approved):

PR #3815, opened 2023-08-15
feat(NODE-5396): add `mongodb-js/saslprep` as a required dependency

baileympearson:
#### What is changing?

We no longer use the third-party `saslprep` package in favor of mongodb-js/saslprep (which can be bundled).  We also made `saslprep` a required dependency and it will now always be installed.

No.

`mongodb-js/saslprep` can be bundled

### `mongodb-js/saslprep` is now installed by default

Until v6, the driver included the `saslprep` package as an optional dependency for SCRAM-SHA-256 authentication.  `saslprep` breaks when bundled with webpack because it attempted to read a file relative to the package location and consequently the driver would throw errors when using SCRAM-SHA-256 if it were bundled.

The driver now depends on `mongodb-js/saslprep`, a fork of `saslprep` that can be bundled with webpack because it includes the necessary saslprep data in memory upon loading. This will be installed by default but will only be used if SCRAM-SHA-256 authentication is used.  


test/integration/auth/scram_sha_256.test.ts
addaleax, line 1: I guess this is covered through some other test?

addaleax (dismissed):

test/integration/auth/scram_sha_256.test.ts
baileympearson: These tests only tested whether or not the driver threw when `saslprep` was not a function.  Now that we rely on mongodb-js and it's safe to bundle, these tests are no longer relevant.

src/cmap/auth/scram.ts
```diff
@@ -141,13 +135,8 @@ async function continueScramConversation(
...
 
-  let processedPassword;
-  if (cryptoMethod === 'sha256') {
-    processedPassword =
-      'kModuleError' in saslprep || typeof saslprep !== 'function' ? password : saslprep(password);
-  } else {
-    processedPassword = passwordDigest(username, password);
-  }
+  const processedPassword =
+    cryptoMethod === 'sha256' ? saslprep(password) : passwordDigest(username, password);
```
nbbeeken, line 139: Unrelated to this line, The check deps github tests are failing, just need to update the expected list of deps.

package.json
```diff
@@ -26,10 +26,8 @@
   },
   "dependencies": {
     "bson": "^5.4.0",
-    "mongodb-connection-string-url": "^2.6.0"
-  },
-  "optionalDependencies": {
-    "saslprep": "^1.0.3"
+    "mongodb-connection-string-url": "^2.6.0",
+    "@mongodb-js/saslprep": "^1.1.0"
```
nbbeeken, line 30: (Unrelated just picking a line)

Just some highlight clarifying suggestions, I think it would be nice to complete the picture on what actually breaks:

> breaks when bundled with webpack

"breaks when bundled with webpack because it attempted to read a file relative to the package location"

> that can be bundled with webpack

"that can be bundled with webpack because it includes the necessary saslprep data in memory upon loading"

nbbeeken (changes requested):

nbbeeken (approved):

test/integration/auth/scram_sha_256.test.ts
addaleax: Right, that was my point here. Feel free to resolve if this is covered elsewhere.

nbbeeken: All our tests run with auth enabled and will use sha-256, we have follow-up work (mentioned in today's slack thread about tls options) that will fix the explicit auth prose tests.

PR #3818, opened 2023-08-15
feat(NODE-5399): use mongodb-js/saslprep instead of saslprep

baileympearson:
#### What is changing?

`@mongodb-js/saslprep` has replaced `saslprep` as the optional dependency our driver users on for SCRAM-SHA-256 authentication.

All existing passing tests are passing.  The failures for FLE are unrelated (they also are failing on `5.x`).

### `mongodb-js/saslprep` is now installed by default

Until this release, the driver included the `saslprep` package as an optional dependency for SCRAM-SHA-256 authentication.  `saslprep` breaks when bundled with webpack because it attempted to read a file relative to the package location and consequently the driver would throw errors when using SCRAM-SHA-256 if it were bundled.

The driver now depends on `mongodb-js/saslprep`, a fork of `saslprep` that can be bundled with webpack because it includes the necessary saslprep data in memory upon loading. This will be installed by default but will only be used if SCRAM-SHA-256 authentication is used.  

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,6 @@
         "@types/kerberos": "^1.1.2",
         "@types/mocha": "^10.0.1",
         "@types/node": "^20.1.0",
-        "@types/saslprep": "^1.0.1",
         "@types/semver": "^7.5.0",
         "@types/sinon": "^10.0.14",
         "@types/sinon-chai": "^3.2.9",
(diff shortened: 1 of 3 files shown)


nbbeeken (approved):

alexbevi:
Hi @ZengJun-cloud, thanks for your comment yesterday, which helped us identify and address this issue quickly and push out a patch release.

PR #3819, opened 2023-08-16
fix(NODE-5548): ensure that tlsCertificateKeyFile maps to cert and key

W-A-James:
#### What is changing?

- file specified by `tlsCertificateKeyFile` is read into both `options.key` and `options.cert` at connect time
- Updated API docs to reflect this change

- Updated API docs

NODE-5548


src/mongo_client.ts
```diff
@@ -438,7 +438,11 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
         options.ca ??= await fs.readFile(options.tlsCAFile, { encoding: 'utf8' });
       }
       if (typeof options.tlsCertificateKeyFile === 'string') {
-        options.key ??= await fs.readFile(options.tlsCertificateKeyFile, { encoding: 'utf8' });
+        if (typeof options.key !== 'string' || typeof options.cert !== 'string') {
```
addaleax, line 441: Are these the right checks? `key` or `cert` could be `Buffer`s as well.

(As a tiny suggestion, I’d maybe even drop the `{ encoding: 'utf8' }` argument here – in theory, this shouldn’t matter much, because the files need to be PEM files and PEM files are ASCII-only, but there’s no real point in decoding the files from UTF-8 because Node.js will just immediately re-encode them using UTF-8, and it might just make the most sense to pass the contents of the file in the most one-to-one fashion possible to Node.js.)

W-A-James: In that case would a truthiness check be more appropriate? i.e.

```ts
if (!options.key || !options.cert) {...}
```

I had changed it from this initially since specifying the encoding would make these options of `string` type which would be falsey if the file provided were empty and so would read the file again on a second call to `connect`.

If I'm understanding, dropping the encoding in the `readFile` call would make this work as it would always return a `Buffer` which should always be truthy whether the file is empty or not and so shouldn't read the file again on subsequent `connect` calls.

addaleax: Yeah, as far as I’m concerned a truthiness check sounds right.

addaleax (approved):

durran:
https://spruce.mongodb.com/version/64de26af57e85a38c18b68ef/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

durran (approved):

test/manual/tls_support.test.ts
```diff
@@ -106,6 +108,42 @@ describe('TLS Support', function () {
...
+      if (client) await client.close();
+    });
+
+    it('throws an error', async () => {
+      const err = await client.connect().catch(e => e);
+      expect(err).to.be.instanceOf(Error);
+    });
+  });
+
+  context('when tlsCAFile is provided, but tlsCertificateKeyFile is missing', () => {
```
durran, line 130: I actually think it should work when passing the CA but not the key file. I'm looking at a Python test that is expected to succeed in this case. https://github.com/mongodb/mongo-python-driver/blob/5bd444a604f85de082646322263031c98af8c213/test/test_ssl.py#L380

test/manual/tls_support.test.ts
```diff
@@ -106,6 +108,42 @@ describe('TLS Support', function () {
       });
     });
   });
+
+  context('when tlsCertificateKeyFile is provided, but tlsCAFile is missing', () => {
```
durran, line 112: Is this test timing out after 10 seconds because we are expecting a server selection timeout error after 30?

durran (changes requested):

test/manual/tls_support.test.ts
W-A-James: Do we know why that's the case? I was trying to find information on that yesterday, but didn't have any luck.

addaleax: You should always be able to specify a CA file when using TLS, yes. The CA file would be used to validate the server certificate; `tlsCertificateKeyFile` is something that is specific to X.509 authentication, but `tlsCAFile` isn’t.

test/manual/tls_support.test.ts
W-A-James: Yes, I think so. Is the option that governs that `connectTimeoutMS`?

W-A-James: Nevermind, I need both `connectTimeoutMS` and `serverSelectionTimeoutMS`

test/manual/tls_support.test.ts
```diff
@@ -106,6 +108,44 @@ describe('TLS Support', function () {
...
+        connectTimeoutMS: 5000
+      });
+    });
+    afterEach(async () => {
+      if (client) await client.close();
+    });
+
+    it('throws an error', async () => {
+      const err = await client.connect().catch(e => e);
+      expect(err).to.be.instanceOf(Error);
```
durran, line 128: I think one last thing here would be to check this is a `MongoServerSelectionError` instead of a generic `Error` (or at least something more specific)

durran (changes requested):

durran (approved):

dariakp (approved):

PR #3834, opened 2023-08-22
feat(NODE-5566): add ability to provide CRL file via tlsCRLFile

durran:
### Description

Adds the ability to provide a CRL file as an option in the connection string or client options.

#### What is changing?
- Adds a new option, `tlsCRLFile`, that takes a string as a path to the crl.pem.
- Refactors a bit of TLS testing variable names.

Yes, TLS options should be updated.

NODE-5549/NODE-5566

#### Users may now provide a CRL file name as a client option.

For users wanting to provide a TLS client revokation list as a filename and not a `Buffer`, a new option `tlsCRLFile` has been added and can be provided in either the connection string or as a `MongoClient` option.

```ts
new MongoClient('mongodb://localhost:27017/db?tls=true&tlsCRLFile=path/to/crl.pem'); // option in connection string.
new MongoClient('mongodb://localhost:27017/db', { tls: true, tlsCRLFile: 'path/to/crl.pem' }) // option in client options.
```

### DIFF
--- a/.evergreen/config.in.yml
+++ b/.evergreen/config.in.yml
@@ -636,8 +636,6 @@ functions:
           export PROJECT_DIRECTORY="$(pwd)"
           export NODE_LTS_VERSION=${NODE_LTS_VERSION}
           export DRIVERS_TOOLS="${DRIVERS_TOOLS}"
-          export SSL_CA_FILE="${SSL_CA_FILE}"
-          export SSL_KEY_FILE="${SSL_KEY_FILE}"
           export MONGODB_URI="${MONGODB_URI}"
 
           bash ${PROJECT_DIRECTORY}/.evergreen/run-tls-tests.sh
--- a/.evergreen/run-tls-tests.sh
+++ b/.evergreen/run-tls-tests.sh
@@ -4,7 +4,8 @@ set -o errexit  # Exit the script with error if any of the commands fail
 
 source "${PROJECT_DIRECTORY}/.evergreen/init-node-and-npm-env.sh"
 
-export SSL_KEY_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/client.pem"
-export SSL_CA_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/ca.pem"
+export TLS_KEY_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/client.pem"
+export TLS_CA_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/ca.pem"
+export TLS_CRL_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/crl.pem"
 
 npm run check:tls
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -810,13 +810,13 @@ export interface MongoOptions
    * to a no-op and `rejectUnauthorized` to the inverse value of `tlsAllowInvalidCertificates`. If
    * `tlsAllowInvalidCertificates` is not set, then `rejectUnauthorized` will be set to `true`.
    *
-   * ### Note on `tlsCAFile` and `tlsCertificateKeyFile`
+   * ### Note on `tlsCAFile`, `tlsCertificateKeyFile` and `tlsCRLFile`
    *
-   * The files specified by the paths passed in to the `tlsCAFile` and `tlsCertificateKeyFile` fields
-   * are read lazily on the first call to `MongoClient.connect`. Once these files have been read and
-   * the `ca`, `cert` and `key` fields are populated, they will not be read again on subsequent calls to
+   * The files specified by the paths passed in to the `tlsCAFile`, `tlsCertificateKeyFile` and `tlsCRLFile`
+   * fields are read lazily on the first call to `MongoClient.connect`. Once these files have been read and
+   * the `ca`, `cert`, `crl` and `key` fields are populated, they will not be read again on subsequent calls to
    * `MongoClient.connect`. As a result, until the first call to `MongoClient.connect`, the `ca`,
-   * `cert` and `key` fields will be undefined.
+   * `cert`, `crl` and `key` fields will be undefined.
    */
   tls: boolean;
   tlsCAFile?: string;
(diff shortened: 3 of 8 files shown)


.evergreen/config.in.yml
```diff
@@ -636,8 +636,6 @@ functions:
           export PROJECT_DIRECTORY="$(pwd)"
           export NODE_LTS_VERSION=${NODE_LTS_VERSION}
           export DRIVERS_TOOLS="${DRIVERS_TOOLS}"
-          export SSL_CA_FILE="${SSL_CA_FILE}"
```
durran, line 639: These are actually not defined but set up in run-tls-tests.sh so I removed them from here.

.evergreen/run-tls-tests.sh
```diff
@@ -4,7 +4,8 @@ set -o errexit  # Exit the script with error if any of the commands fail
 
 source "${PROJECT_DIRECTORY}/.evergreen/init-node-and-npm-env.sh"
 
-export SSL_KEY_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/client.pem"
-export SSL_CA_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/ca.pem"
+export TLS_KEY_FILE="$DRIVERS_TOOLS/.evergreen/x509gen/client.pem"
```
durran, line 7: Just renamed while here to align with us preferring "tls" over "ssl".

src/mongo_client.ts
```diff
@@ -790,7 +795,7 @@ export interface MongoOptions
    * | nodejs native option  | driver spec equivalent option name            | driver option type |
    * |:----------------------|:----------------------------------------------|:-------------------|
    * | `ca`                  | `tlsCAFile`                                   | `string`           |
-   * | `crl`                 | N/A                                           | `string`           |
+   * | `crl`                 | `tlsCRLFile`                                  | `string`           |
```
durran, line 798: It's not really a spec option but didn't want to create another column in the table just for the one option.

test/manual/tls_support.test.ts
```diff
@@ -114,6 +115,29 @@ describe('TLS Support', function () {
     });
   });
 
+  context('when providing tlsCRLFile', () => {
```
durran, line 118: We don't have a CRL in drivers-tools that wouldn't reject the CA we have, so just have an integration test for the failure. Looked at the Python driver and they do the same.

nbbeeken (approved):

W-A-James (changes requested):

test/manual/tls_support.test.ts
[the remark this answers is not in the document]
durran: I've updated but needed to split them out for when the connection would succeed and fail.

src/mongo_client.ts
W-A-James, line 1: Can you update the `Note on \`tlsCAFile\` and \`tlsCertificateKeyFile\`` to also mention that `tlsCRLFile` also has the same behaviour.

test/manual/tls_support.test.ts
```diff
@@ -47,36 +48,59 @@ describe('TLS Support', function () {
...
+            const caFileAccessTime = (await fs.stat(TLS_CA_FILE)).atime;
+            const certKeyFileAccessTime = (await fs.stat(TLS_CERT_KEY_FILE)).atime;
 
-        expect(client.options).property('ca').to.exist;
-        expect(client.options).property('key').to.exist;
-        expect(client.options).property('cert').to.exist;
+            await client.connect();
+
+            expect((await fs.stat(TLS_CA_FILE)).atime).to.deep.equal(caFileAccessTime);
+            expect((await fs.stat(TLS_CERT_KEY_FILE)).atime).to.deep.equal(certKeyFileAccessTime);
```
W-A-James, line 81: Can we add assertions similar to these for the CRL file?

test/manual/tls_support.test.ts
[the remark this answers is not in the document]
W-A-James: Good on the test restructuring but I more so meant that we want to guarantee that no matter how many times we call connect, we only read in each of these files once. I added another comment directly on the assertions I'm making reference to.

W-A-James (changes requested):

test/manual/tls_support.test.ts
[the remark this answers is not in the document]
durran: Yes I just caught that and added the other test as well.

src/mongo_client.ts
durran: Updated.

test/manual/tls_support.test.ts
durran: Done

W-A-James (commented):
Parse error in `test/manual/tls_support.test.ts` but otherwise LGTM

durran:
> Parse error in `test/manual/tls_support.test.ts` but otherwise LGTM

Fixed.

nbbeeken (dismissed):

W-A-James (approved):

test/manual/tls_support.test.ts
```diff
@@ -42,41 +43,79 @@ describe('TLS Support', function () {
...
+          });
+        });
       });
 
-      context('when client has been opened and closed more than once', function () {
-        it('should only read files once', async () => {
-          await client.connect();
-          await client.close();
+      context('when the connection will fail', () => {
```
baileympearson, line 87: What do these tests mean?  I interpret them as "when the TLS settings are misconfigured" (resulting in the call to `connect` to fail.  And the first test does demonstrate this.

But what about the second?  I'd expect the call to `connect` on line 109 to throw, if the client were misconfigured, and that the test would fail.  How is it passing?

baileympearson (changes requested):

test/manual/tls_support.test.ts
durran: The connection will fail here because the CRL will cause a revokation. The second was failing - I just fixed that test.

nbbeeken (changes requested):

nbbeeken (approved):

baileympearson (approved):

PR #3835, opened 2023-08-22
fix(NODE-5563): unhandled exceptions during close()

DevBrent:
### Description
During sigfault/sigint or in cases where db.close is called twice there exist race conditions within db.close that cause unhandled exceptions.

#### What is changing?
- Additional guards to check if the topology.close function exists before calling it.
- Extracted decrypter.close from inside toplogy.close's promise which was not being called when the toplogy didn't exist due to the unhandled exception.

No

Preventing unhandled exceptions during sig fault/int/hup shutdowns where close may be called more than once or perhaps something else clears the topology.

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -549,19 +549,26 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
     const topology = this.topology;
     this.topology = undefined;
 
-    await new Promise<void>((resolve, reject) => {
-      topology.close({ force }, error => {
-        if (error) return reject(error);
-        const { encrypter } = this[kOptions];
-        if (encrypter) {
-          return encrypter.close(this, force, error => {
-            if (error) return reject(error);
-            resolve();
-          });
-        }
-        resolve();
+    // Prevent race condition during sighalt/int
+    if (toplogy) {
+      await new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


DevBrent:
> Hey @DevBrent thanks for helping us out here. Can you provide a small reproduction of the issue you're seeing? If it helps it might be worth taking a look at the parallel connect tests here: https://github.com/mongodb/node-mongodb-native/blob/main/test/integration/node-specific/mongo_client.test.ts#L519 to get an idea of how we might ensure this fixes the issue as intended.

I'll definitely look at writing new tests for this case. My best approximation might be calling close twice on the same tick. My trigger behavior is that we call a shutdown registry every time SIGHUP through SIGTERM are called. Every mongo connection registers a shutdown function with this shutdown registry. AWS sends multiple events on Amazon Linux 2 and I'm betting shutdown is being called concurrently twice.

Still, I don't think this code should throw unhandledExceptions. If there weren't promise wrappers without tries inside failing into reject callbacks I could at least catch the culprit.

Is optional chaining ok on this repo? `this.topology?.close?.()`

nbbeeken:
> Is optional chaining ok on this repo? this.topology?.close?.()

Yep! Thank you for the additional information that will help coming up with a reproduction. Does calling `.close()` more than once with out awaiting the promise cause this?

PR #3804, opened 2023-08-09
refactor(NODE-5379): cursor internals to use async-await

malikj2000:
### Description
refactored cursor internals to use async-await syntax instead of callbacks

#### What is changing?
getMore, _initialize, cleanupCursor, and [kInit] now use async-await syntax
all classes that inherit from AbstractCursor now use async-await

##### Is there new documentation needed for these changes?
None

#### What is the motivation for this change?
further converting the driver to async-await

### Double check the following

- [ ] Ran `npm run check:lint` script
- [ ] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [ ] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`
- [ ] Changes are covered by tests
- [ ] New TODOs have a related JIRA ticket

PR #3841, opened 2023-08-23
fix(NODE-5588): recursive calls to next cause memory leak

nbbeeken:
#### What is changing?

- Modified the isDead condition (moved it to a getter)
  - If the cursorId is zero, if it has been closed or has been killed, it's "dead"
  - if kId is nullish, it has not been initialized yet, a not-even-started-yet cursor is not dead
  - `(this[kId]?.isZero() ?? false) || this[kClosed] || this[kKilled];`

No

### Fix memory leak with ChangeStreams

In a previous release, 5.7.0, we refactored cursor internals from callbacks to async/await. In particular, the `next` function that powers cursors was written with callbacks and would recursively call itself depending on the cursor type. For `ChangeStreams`, this function would call itself if there were no new changes to return to the user. After converting that code to async/await each recursive call created a new promise that saved the current async context. This would slowly build up memory usage if no new changes came in to unwind the recursive calls. 

The function is now implemented as a loop, memory leak be gone!


src/cursor/abstract_cursor.ts
```diff
@@ -220,6 +220,11 @@ export abstract class AbstractCursor<
     return this[kId] ?? undefined;
   }
 
+  /** @internal */
+  get isDead() {
```
durran, line 224: This makes much more sense to me to be a property of the cursor itself instead of passing a cursor to a function.

durran (approved):
Heap looks good, using `bind` cleans it up nicely. Well done.

nbbeeken:
> using bind cleans it up nicely.

Thanks! I'd like us to pull back in #3804 after this is in to make them actually async functions

src/cursor/find_cursor.ts
```diff
@@ -101,7 +101,9 @@ export class FindCursor<TSchema = any> extends AbstractCursor<TSchema> {
         limit && limit > 0 && numReturned + batchSize > limit ? limit - numReturned : batchSize;
 
       if (batchSize <= 0) {
-        this.close().finally(() => callback());
+        this.close().finally(() =>
```
baileympearson, line 104: Is this change just for consistency in return from `_getMore` or is there a reason it's needed?

src/cursor/abstract_cursor.ts
baileympearson: Did we intentionally change the semantics of `isDead`?

baileympearson (approved):
two non-blocking questions, otherwise LGTM.

src/cursor/abstract_cursor.ts
baileympearson: Nevermind, I see that now in the description.  But I'm still unsure why the logic was changed

src/cursor/find_cursor.ts
nbbeeken: Yea, it is for result consistency, I may be able to revert it if I make sure the logic is nullish safe in next but the contract should have been either throw or result and when we make _getMore actually async that will be the case for the promise's resolution type. 

This code path also specifically claims it is for legacy servers which we don't support, I'm fairly certain this is only triggered by mock server tests.

src/cursor/abstract_cursor.ts
nbbeeken: I changed it to exit the loop and enter clean-up logic at the correct times. The prior logic didn't account for a cursor already being closed, or killed, so the loop would attempt a getMore on a killed cursor id. This same "exit" condition applies to all the places the cursorIsDead helper was being used.

PR #3842, opened 2023-08-23
fix(NODE-5587): recursive calls to next cause memory leak

nbbeeken:
#### What is changing?

- Modified the isDead condition (moved it to a getter)
  - If the cursorId is zero, if it has been closed or has been killed, it's "dead"
  - if kId is nullish, it has not been initialized yet, a not-even-started-yet cursor is not dead
  - `(this[kId]?.isZero() ?? false) || this[kClosed] || this[kKilled];`

No

### Fix memory leak with ChangeStreams

In a previous release, 5.7.0, we refactored cursor internals from callbacks to async/await. In particular, the `next` function that powers cursors was written with callbacks and would recursively call itself depending on the cursor type. For `ChangeStreams`, this function would call itself if there were no new changes to return to the user. After converting that code to async/await each recursive call created a new promise that saved the current async context. This would slowly build up memory usage if no new changes came in to unwind the recursive calls. 

The function is now implemented as a loop, memory leak be gone!

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -221,6 +221,11 @@ export abstract class AbstractCursor<
   }
 
   /** @internal */
+  get isDead() {
+    return (this[kId]?.isZero() ?? false) || this[kClosed] || this[kKilled];
+  }
+
+  /** @internal */
   get client(): MongoClient {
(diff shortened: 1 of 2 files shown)


durran (approved):
Same changes as 6.x.

PR #3859, opened 2023-09-06
test(NODE-4338): refactor auth prose tests

durran:
### Description

Re-enables SCRAM auth tests and refactors to not use `withClient`.

#### What is changing?

- Refactors the SCRAM auth tests to not use `withClient`
- Skips failing unicode username/password tests.

None

NODE-4338


test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+  return `mongodb://${username}:${password}@${config.host}:${config.port}/admin?`;
+}
+
+describe('Authentication Spec Prose Tests', function () {
+  beforeEach(function () {
+    // todo(NODE-5631): fix tests to run in load balancer mode.
+    if (
+      process.env.AUTH === 'noauth' ||
+      process.env.LOAD_BALANCER ||
+      !satisfies(this.configuration.version, '>=3.7.3')
```
durran, line 17: Note the create user commands will fail on < 3.7.3 since `mechanisms` cannot be provided to the command.

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+        await Promise.all(users.map(user => db.removeUser(user.username)));
+        await utilClient?.close();
+      });
+
+      [
+        { username: 'IX', password: 'IX' },
+        { username: 'IX', password: 'I\u00ADX' },
+        { username: '\u2168', password: 'IV' },
+        { username: '\u2168', password: 'I\u00ADV' }
+      ].forEach(({ username, password }) => {
```
baileympearson, line 347:
```suggestion
      for (const { username, password } of [
        { username: 'IX', password: 'IX' },
        { username: 'IX', password: 'I\u00ADX' },
        { username: '\u2168', password: 'IV' },
        { username: '\u2168', password: 'I\u00ADV' }
      ]) {
```

optional code modernization

test/integration/auth/auth.prose.test.ts
baileympearson: every test already specifies a min server version.  do we want to just move this logic into tests `requires`?  We can also use the generic predicate filter

```typescript
const metadata: MongoDBMetadataUI = {
  requires: { 
    auth: 'enabled',
    mongodb: '>=3.7.3',
    predicate: () => process.env.LOAD_BALANCER ? 'TODO(NODE-5631): fix tests to run in load balancer mode.' : true
  }
}

....

it('description', metadata, async function() { ...} );
```

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+      /**
+       * Step 2
+       * For each test user, verify that you can connect and run a command requiring
+       * authentication for the following cases:
+       *   - Explicitly specifying each mechanism the user supports.
+       *   - Specifying no mechanism and relying on mechanism negotiation.
+       */
+      for (const user of users) {
+        for (const mechanism of user.mechanisms) {
+          it(`authenticates ${user.description} when explicitly specifying ${mechanism}`, {
```
baileympearson, line 85:
```suggestion
          it(`authenticates ${user.description} when explicitly specifying ${mechanism} via the client options`, {
```

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+                  password: user.password
+                },
+                authMechanism: mechanism,
+                authSource: 'admin'
+              };
+
+              const client = this.configuration.newClient({}, options);
+              const stats = await client.db('test').stats();
+              expect(stats).to.exist;
+              await client.close();
```
baileympearson, line 100: this will leak a client if any step between line 97-99 fails.  This might not be a concern because if any of those steps fail, the test run also fails, but we can fix this by closing the client in an `after` hook or just using a finally block

applies to all tests that follow this pattern

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+          const spy = sinon.spy(ScramSHA256.prototype, 'auth');
+
+          const client = this.configuration.newClient({}, options);
+          const stats = await client.db('test').stats();
+          expect(stats).to.exist;
+          expect(spy.called).to.equal(true);
+          sinon.restore();
+          await client.close();
+        }
+      });
```
baileympearson, line 184:
```suggestion
      }).skipReason = '...'
```
can we attach a skip reason? 

We can also use the predicate filter if you'd prefer:

```typescript
{ requires: { predicate: () => 'TODO(..)...' } }
```

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+      it('fails for a nonexistent username with same error type as bad password', {
+        metadata: { requires: { mongodb: '>=3.7.3' } },
+        test: async function () {
+          const noUsernameOptions = {
+            auth: {
+              username: 'roth',
+              password: 'pencil'
+            },
+            authSource: 'admin'
+          };
```
baileympearson, line 232: Am I reading this test incorrectly?  Looks like this still specifies `username`

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+            .getCalls()
+            .filter(c => c.thisValue.id !== '<monitor>') // ignore all monitor connections
+            .filter(
+              c => c.args[1][process.env.MONGODB_API_VERSION ? 'hello' : LEGACY_HELLO_COMMAND]
+            );
+
+          expect(calls).to.have.length(1);
+          const handshakeDoc = calls[0].args[1];
+          expect(handshakeDoc).to.have.property('speculativeAuthenticate');
+          sinon.restore();
```
baileympearson, line 283: same comment about leaking clients with the `sinon.restore()`

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+          await client.close();
+        }
+      });
+    });
+
+    // todo(NODE-5621): fix the issue with unicode characters.
+    describe.skip('Step 4', function () {
```
baileympearson, line 290: same comment as above about attaching a skip reason

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,373 @@
...
+      it('fails to connect if incorrect auth mechanism is explicitly specified', {
+        metadata: { requires: { mongodb: '>=3.7.3' } },
+        test: async function () {
+          const options = {
+            auth: {
+              username: userMap.sha256.username,
+              password: userMap.sha256.password
+            },
+            authSource: 'admin',
+            authMechanism: 'SCRAM-SHA-1'
```
baileympearson, line 200: Should this be updated to use the current user, instead of always hard-coding to `sha256`?

baileympearson (changes requested):

test/integration/auth/auth.prose.test.ts
durran: Updated to suggestion.

test/integration/auth/auth.prose.test.ts
durran: Done

test/integration/auth/auth.prose.test.ts
durran: Added.

test/integration/auth/auth.prose.test.ts
durran: I've updated all tests to use try...finally for this pattern.

test/integration/auth/auth.prose.test.ts
durran: Added skip reason.

test/integration/auth/auth.prose.test.ts
durran: Removed username and updated test assertions with a comment since the Node driver fails validation when instantiating the client with no username present.

test/integration/auth/auth.prose.test.ts
durran: Use try...finally now.

test/integration/auth/auth.prose.test.ts
durran: Skip reason added.

test/integration/auth/auth.prose.test.ts
durran: It's basically explicitly stating that the `sha256` only user cannot authenticate when specifying `SCRAM-SHA-1` as the mechanism. I think it's clearer this way.

test/integration/auth/auth.prose.test.ts
```diff
@@ -239,26 +254,31 @@ describe('Authentication Spec Prose Tests', function () {
...
-      });
+      );
 
-      it('should send speculativeAuthenticate on initial handshake on MongoDB 4.4+', {
-        metadata: { requires: { mongodb: '>=4.4', topology: ['single'] } },
-        test: async function () {
+      it(
+        'sends speculativeAuthenticate on initial handshake on MongoDB 4.4+',
+        metadata,
+        async function () {
```
baileympearson, line 281: I had to read the spec here to clarify:

> Older servers will ignore the speculativeAuthenticate argument. New servers will participate in the standard authentication conversation if this argument is missing.

I don't think we have any branching logic determining when we use speculative authentication.  Maybe we should remove the server version from the test?

test/integration/auth/auth.prose.test.ts
baileympearson: >        * Step 3
>        * For test users that support only one mechanism, verify that explictly specifying
>        * the other mechanism fails.

The test implies that we also need to test that the `SCRAM-SHA-1` user cannot authenticate by specifying `SCRAM-SHA-256`.  Right now, this only covers the opposite.

baileympearson (changes requested):

test/integration/auth/auth.prose.test.ts
durran: I've removed the 4.4+ version from the description but left the metadata there as all tests still need 3.7.3+.

test/integration/auth/auth.prose.test.ts
durran: I've added the reverse test now.

baileympearson (dismissed):

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,416 @@
...
+            this.configuration.newClient({}, noUsernameOptions);
+            expect.fail(
+              'Creating a new client with a password and no username must fail validation.'
+            );
+          } catch (noUserError) {
+            // NOTE: This prose test fails Node's validation of the credentials object - a username
+            // MUST always be provided but it satisfies the test requirement of not getting a
+            // network or command error on the handshake.
+            expect(noUserError).to.match(/username/);
+          }
```
nbbeeken, line 290: The expect.fail has "username" in the string but I think our intention is to seek a username substring from a driver error correct?

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,416 @@
...
+          };
+
+          const spy = sinon.spy(ScramSHA256.prototype, 'auth');
+          const client = this.configuration.newClient({}, options);
+          try {
+            const stats = await client.db('test').stats();
+            expect(stats).to.exist;
+            expect(spy.called).to.equal(true);
+          } finally {
+            sinon.restore();
```
nbbeeken, line 194: Can we `sinon.restore` in the afterEach? newClient could throw preventing this cleanup. calling restore if there was no use of sinon is harmless so might as well put it where we can be sure it will run.

test/integration/auth/auth.prose.test.ts
```diff
@@ -0,0 +1,416 @@
...
+            authMechanism: 'SCRAM-SHA-1'
+          };
+
+          const client = this.configuration.newClient({}, options);
+          const error = await client
+            .db('test')
+            .stats()
+            .catch(e => e);
+          expect(error.message).to.match(/Authentication failed|SCRAM/);
+          await client.close();
```
nbbeeken, line 224: We use `finally` in a bunch of other tests to make sure we close the client, why not here? 

In my opinion, it would be easier to ensure this cleanup if we assign the client to a suite level `let` and use afterEach. If that sounds preferable you have to put the tests that are in the for loop into a function so they can remain bound to the same client they create

nbbeeken (changes requested):

test/integration/auth/auth.prose.test.ts
durran: Yes we are expecting the options validation to throw saying the username is missing. Was there a change request here?

nbbeeken: The `.match(/username/)` assertion will _always_ pass because the `expect.fail` _also_ satisfies `/username/`. Either we need to change the string in `expect.fail` or restructure this to store the error into a variable and assert outside of the `try/catch`

test/integration/auth/auth.prose.test.ts
durran: Moved to the `afterEach` now.

test/integration/auth/auth.prose.test.ts
durran: I've put the close calls now into `afterEach`, and moved the client to the top level to avoid needing to extract the tests in the for loop into a function.

test/integration/auth/auth.prose.test.ts
durran: Ahhh. Now I see. :)

durran: Fail message changed.

nbbeeken (approved):

baileympearson (approved):

PR #3801, opened 2023-08-08
chore(NODE-5446): update driver v6 dependencies

W-A-James:
#### What is changing?
Update the following dev dependencies
* "@microsoft/api-extractor": "^7.36.4"
* "@octokit/core": "^4.2.4"
* "@types/node": "^20.4.8"
* "@types/sinon": "^10.0.16"
* "@typescript-eslint/eslint-plugin": "^5.62.0"
* "@typescript-eslint/parser": "^5.62.0"
* "eslint": "^8.46.0"
* "eslint-config-prettier": "^8.10.0"
* "eslint-plugin-import": "^2.28.0"
* "gcp-metadata": "^5.3.0"
* "semver": "^7.5.4"
* "sinon": "^15.2.0"
* "v8-heapsnapshot": "^1.3.1"

##### Is there new documentation needed for these changes?

No

#### What is the motivation for this change?

NODE-5446; keeping tooling up to date

### Double check the following

- [x] Ran `npm run check:lint` script
- [x] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [x] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`
- [ ] Changes are covered by tests
- [ ] New TODOs have a related JIRA ticket

PR #3860, opened 2023-09-06
deps(NODE-5446): bump dependencies again for v6 

baileympearson:
### Description

In an interesting twist, upgrading dev dependencies again...no longer causes our aws tests to fail?  🤔 

This PR basically updates all the dependencies Warren updated as a part of https://github.com/mongodb/node-mongodb-native/pull/3801.  There are a few version differences (which could explain why the tests pass here but failed on Warren's initial PR).  The aws tests pass consistently for me - I've run them 4 times now with none of the failures we saw on the initial dependency bump.  I'm inclined to call this work "resolved", since I don't think it's worthwhile investigating why the dependency bump initially caused tests to fail if they pass consistently now.

This work includes a temporary fix to the `npm` version issues.  We can either block this on Durran's CI fixes or just merge this and have Durran remove it when he implements the permanent fix.

#### What is changing?

Bumps the dependencies originally updated in https://github.com/mongodb/node-mongodb-native/pull/3801.


durran (changes requested):
Just need to resolve the npm fix conflicts.

durran (approved):

PR #3867, opened 2023-09-14
fix(NODE-5628): bulkWriteResult.insertedIds does not filter out _ids that are not actually inserted

aditi-khare-mongoDB:
### Description for NODE-5628
Before this change, `BulkWriteResult.insertedIds` [(1)](https://www.mongodb.com/docs/manual/reference/method/BulkWriteResult/#mongodb-data-BulkWriteResult.insertedIds/) would return an Object containing all attempted insertedIds, regardless of if any of the inserts failed. This affects `BulkWrite` and `InsertMany` operations. After this PR is merged, `BulkWriteResult.insertedIds` will reflect which _ids were actually inserted.

#### Other Notes for Reviewers 
- The [JIRA ticket ](https://jira.mongodb.org/browse/NODE-5421) only mentions the bug on the `InsertMany` operation; however, I was able to reproduce it generally for any insert operation that returns `BulkWriteResult` namely `BulkWrite`.
- `BulkResult.insertedIds` will still contain unsuccessful inserted `_ids`. The `BulkResult` class is not user facing.
- In this PR, unsuccessful inserts are identified through `BulkResult.writeErrors` list property, rather than the actual direct insert response from the server, since the batch result does not contain` _id` information.

#### Is there new documentation needed for these changes?
Yes, it needs to be communicated to users that when an insert does not succeed through a `bulkwrite()` or `collection.insertMany()`, it will no longer appear in the error's `.result.insertedIds`

#### What is the motivation for this change?
See release highlights.

### Inserted ids in bulk write only contain successful insertions

Prior to this fix the bulk write error's `insertedIds` field contained ids of all attempted inserts in a bulk operation.

When a `bulkwrite()` or an `insertMany()` operation rejects one or more inserts, throwing an error, the error's `BulkWriteResult` property will be filtered to only contain successfully inserted ids (ex: `error.result.insertedIds`).

#### Previous Behavior
The error's `.result.insertedIds` does includes all attempted inserts, regardless of if they failed or not.

#### New Behavior
The error's `.result.insertedIds` does not include unsuccessful inserts.
Namely, 
- On an ordered bulk insert operation, after the first error, all future inserts fail, and therefore should not be included in insertedIds. (This is because the order would be invalid once one insert has failed).
[...]

PR #3870, opened 2023-09-15
fix(NODE-5627): BulkWriteResult.insertedIds does not filter out _ids that are not actually inserted

aditi-khare-mongoDB:
### Description for NODE-5627
Before this change, on Node Driver Version 5.x, `BulkWriteResult.insertedIds` [(1)](https://www.mongodb.com/docs/manual/reference/method/BulkWriteResult/#mongodb-data-BulkWriteResult.insertedIds/) would return an Object containing all attempted insertedIds, regardless of if any of the inserts failed. This affects `BulkWrite` and `InsertMany` operations. After this PR is merged, `BulkWriteResult.insertedIds` will reflect which _ids were actually inserted.

#### Other Notes for Reviewers 
- The [JIRA ticket ](https://jira.mongodb.org/browse/NODE-5421) only mentions the bug on the `InsertMany` operation; however, I was able to reproduce it generally for any insert operation that returns `BulkWriteResult` namely `BulkWrite`.
- `BulkResult.insertedIds` will still contain unsuccessful inserted `_ids`. The `BulkResult` class is not user facing.
- In this PR, unsuccessful inserts are identified through `BulkResult.writeErrors` list property, rather than the actual direct insert response from the server, since the batch result does not contain` _id` information.
- NOTE: this PR's code changes are identical to the [6.x NODE-5421 PR](https://github.com/mongodb/node-mongodb-native/pull/3867)

#### Is there new documentation needed for these changes?
Yes, it needs to be communicated to users that when an insert does not succeed through a `collection.bulkwrite()` or `collection.insertMany()` operation, it will no longer appear in the error's `.result.insertedIds`

See release highlights.

### Inserted ids in bulk write only contain successful insertions - 5.x 

Prior to this fix the bulk write error's `insertedIds` field contained ids of all attempted inserts in a bulk operation.

When a `bulkwrite()` or an `insertMany()` operation rejects one or more inserts, throwing an error, the error's `BulkWriteResult` property will be filtered to only contain successfully inserted ids (ex: `error.result.insertedIds`).

#### Previous Behavior
The error's `.result.insertedIds` does includes all attempted inserts, regardless of if they failed or not.

#### New Behavior
The error's `.result.insertedIds` does not include unsuccessful inserts.
Namely, 
- On an ordered bulk insert operation, after the first error, all future inserts fail, and therefore should not be included in insertedIds. (This is because the order would be invalid once one insert has failed).
- On an unordered bulk insert operation run on a transaction, after the first error, all future inserts fail, and therefore should not be included in insertedIds. (See [here](https://www.mongodb.com/docs/manual/reference/method/db.collection.bulkWrite/#std-label-bulkwrite-error-handling-txn) for more information.)
- On an unordered bulk insert operation that is not run on a transaction, all inserts that individually fail should be removed.

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -104,6 +104,25 @@
         "node": ">=6.0.0"
       }
     },
+    "node_modules/@aws-crypto/crc32": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/crc32/-/crc32-3.0.0.tgz",
+      "integrity": "sha512-IzSgsrxUcsrejQbPVilIKy16kAT52EwB6zSaI+M3xxIhKh5+aldEyvI+z6erM7TCLB2BJsFrtHjp6/4/sr+3dA==",
+      "optional": true,
+      "peer": true,
+      "dependencies": {
+        "@aws-crypto/util": "^3.0.0",
+        "@aws-sdk/types": "^3.222.0",
+        "tslib": "^1.11.1"
+      }
+    },
+    "node_modules/@aws-crypto/crc32/node_modules/tslib": {
+      "version": "1.14.1",
+      "resolved": "https://registry.npmjs.org/tslib/-/tslib-1.14.1.tgz",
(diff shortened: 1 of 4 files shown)


baileympearson (approved):
failing CI is unrelated

PR #3871, opened 2023-09-18
feat(NODE-5639): keep executing context on queued commands

david-luna:
### Description

Implementation for https://jira.mongodb.org/browse/NODE-5639

Make use of [AsyncResource.bind API](https://nodejs.org/api/async_context.html#static-method-asyncresourcebindfn-type-thisarg) to bind the current execution context of `WaitQueueMember` callback.

The change proposed will make callbacks to have the same context they where created. Useful if the consumer of the package is using [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) to keep state.

One use case is APM agents. They make use of these APIs to keep context between callbacks and async calls.

#### What is changing?

Bind the current execution context to `WaitQueueMember` callback property. 

No

Get span hierarchy right when `mongodb` driver is instrumented by an APM agent like Opentelemetry, Elastic, DataDog...


baileympearson:
This PR should be unnecessary now that the driver uses async-await internally.

PR #3872, opened 2023-09-19
Fix typo in options

drenbilalliii:
### Description
 A small typo fixed in find_and_modify.ts file with Nullish coalescing operator. It was throwing an error when trying to run the nodejs server.

### DIFF
--- a/src/operations/find_and_modify.ts
+++ b/src/operations/find_and_modify.ts
@@ -141,7 +141,7 @@ class FindAndModifyOperation extends CommandOperation<Document> {
       upsert: false
     };
 
-    options.includeResultMetadata ??= false;
+    options.includeResultMetadata ?? false;
 
     const sort = formatSort(options.sort);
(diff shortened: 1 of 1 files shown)


nbbeeken:
Hi @drenbilalliii, thanks for reaching out. The syntax used here is the nullish assignment operator `??=` it is intended to assign the default value of `false` to the `options.includeResultMetadata` option if the property is `null` or `undefined`. Making the proposed change would leave the `includeResultMetadata` option unset. If you are seeing a syntax error you might not be on a supported node version, logical assignment operators have been [available since Node.js v15.14](https://node.green/#ES2021-features-Logical-Assignment-----basic-support) and the driver's [minimum version is Node.js v16.20](https://github.com/mongodb/node-mongodb-native/blob/main/etc/notes/CHANGES_6.0.0.md#minimum-nodejs-version-is-now-v16201).

PR #3873, opened 2023-09-21
fix(NODE-5496): remove client-side collection and database name check validation

aditi-khare-mongoDB:
### Description for [NODE-5496](https://jira.mongodb.org/browse/NODE-5496)
The previous internal Node Driver `checkCollectionName` functions was more narrow with name check guidelines than server. This led `checkCollectionName` to invalidate legal collection names. This affects Mongosh and Compass since they rely on the Node Driver and need to be able to access all valid collection names.

The Node Driver has made the decision to only rely on server validation for collection name checking, and for database name checking. This is because collections and databases are so closely related.

In addition, string coercion for database names is removed for the Msg and Query class.

#### What is changing?
The Node Driver will no longer error-check for database naming and collection naming, and leave it to the server to validate.
Namely, `checkCollectionName` used to incorrectly invalidate collection names that started or ended with a '.' character.

However, for the case of the dot character in a database name, the Node Driver will not remove client-side validation for this case. This is because the dot character in a database name will cause ambiguous behavior before it is sent to the server.

Yes, the API now allows for more collection names.

The motivation is to ensure the driver is adjusted to match server naming restrictions for collections and databases.

#### Current Behavior 
 The following code leads to an error.
`db.createCollection('spider.');`
#### Expected Behavior
No error

### Release Highlight

Database and collection name checking will be in sync with the MongoDB server's [name requirements](https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Restriction-on-Collection-Names).
Specifically, users can now create collections that start or end with the '.' character.


test/integration/node-specific/db.test.js
```diff
@@ -1,37 +1,38 @@
...
+      await client.close();
+    });
+
+    it('should throw error on server only', async function () {
+      try {
+        db = client.db('a\x00b');
+        await db.createCollection('spider');
+        expect.fail('a MongoServerError was expected due to illegal db name');
+      } catch (error) {
+        expect(error instanceof MongoServerError);
```
nbbeeken, line 33: Let's use the chai `expect(error).to.be.instanceOf()` assertion here to get a more detailed error message

nbbeeken, line 33: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L110

Looks like the error we get back sets code to `73` I think that would be a good assertion to add, we want to know that the server error was due to the namespace and not something related to the operation.

test/integration/node-specific/db.test.js
```diff
@@ -1,37 +1,38 @@
...
+
+    afterEach(async function () {
+      db = undefined;
+      await client.close();
+    });
+
+    it('should throw error on server only', async function () {
+      try {
+        db = client.db('a\x00b');
+        await db.createCollection('spider');
```
nbbeeken, line 30: We can use the `.catch(error => error)` trick to remove the control flow from the test.

test/integration/collection-management/collection.test.ts
```diff
@@ -95,18 +95,14 @@ describe('Collection', function () {
...
-      );
-      expect(() => db.collection('test..t')).to.throw('Collection names cannot be empty');
-      done();
+    it('should fail on server due to illegal collection name', async function () {
+      try {
+        const illegalCollection = await db.createCollection('a\x00b');
+        await illegalCollection.insertOne({ a: 1 });
+        expect.fail('a MongoServerError was expected due to illegal collection name');
+      } catch (error) {
+        expect(error instanceof MongoServerError);
```
nbbeeken, line 104: chai assertion

test/integration/collection-management/collection.test.ts
```diff
@@ -95,18 +95,14 @@ describe('Collection', function () {
...
-      );
-      expect(() => db.collection('test.')).to.throw(
-        "Collection names must not start or end with '.'"
-      );
-      expect(() => db.collection('test..t')).to.throw('Collection names cannot be empty');
-      done();
+    it('should fail on server due to illegal collection name', async function () {
+      try {
+        const illegalCollection = await db.createCollection('a\x00b');
+        await illegalCollection.insertOne({ a: 1 });
```
nbbeeken, line 101: promise return error trick

test/integration/collection-management/collection.test.ts
```diff
@@ -95,18 +95,14 @@ describe('Collection', function () {
...
-        "Collection names must not start or end with '.'"
-      );
-      expect(() => db.collection('test.')).to.throw(
-        "Collection names must not start or end with '.'"
-      );
-      expect(() => db.collection('test..t')).to.throw('Collection names cannot be empty');
-      done();
+    it('should fail on server due to illegal collection name', async function () {
+      try {
+        const illegalCollection = await db.createCollection('a\x00b');
```
nbbeeken, line 100: Because we're using createCollection here which actually reaches out to the server to create the collection we aren't going to reach the next line performing the insert. I think we want the `db.collection` helper here, right?

test/integration/collection-management/collection.test.ts
```diff
@@ -95,18 +95,14 @@ describe('Collection', function () {
...
-      expect(() => db.collection('te$t')).to.throw("Collection names must not contain '$'");
-      expect(() => db.collection('.test')).to.throw(
-        "Collection names must not start or end with '.'"
-      );
-      expect(() => db.collection('test.')).to.throw(
-        "Collection names must not start or end with '.'"
-      );
-      expect(() => db.collection('test..t')).to.throw('Collection names cannot be empty');
-      done();
+    it('should fail on server due to illegal collection name', async function () {
```
nbbeeken, line 98: `Collection standard collection tests should fail on server due to illegal collection name`

If I am following the indentation of the tests correctly, the above is the full title of this test. I think it is important to mention the specific API under test and the intended behavior to understand what our goal was if we need to revisit the decisions made here. Also, nit, we can drop the "it ( should xx" for imperative "it ( fails xx"

nbbeeken (changes requested):
Some fixups, we also wanted to introduce API docs to MongoClient .db() and Db .collection() that mention the new behavior, something like: "namespace validation occurs at operation time"

test/integration/node-specific/db.test.js
aditi-khare-mongoDB: I'll add it in!

aditi-khare-mongoDB: Sounds good, as it turns out its not an instance of MongoServerError, but rather MongoServerError is the name of the error, so I added a check for that instead.

test/integration/node-specific/db.test.js
aditi-khare-mongoDB: Yes, I'll add it in.

test/integration/collection-management/collection.test.ts
aditi-khare-mongoDB: I'll fix the assertions for this one, too!

test/integration/collection-management/collection.test.ts
aditi-khare-mongoDB: I'll fix the error logic for this one, too.

test/integration/collection-management/collection.test.ts
aditi-khare-mongoDB: I see, I didn't realize it fails prior, I'll change it to `db.collection `

test/integration/collection-management/collection.test.ts
aditi-khare-mongoDB: Okay, sounds good

test/integration/collection-management/collection.test.ts
```diff
@@ -95,18 +95,14 @@ describe('Collection', function () {
...
-        "Collection names must not start or end with '.'"
-      );
-      expect(() => db.collection('test..t')).to.throw('Collection names cannot be empty');
-      done();
+    it('fails on server due to invalid namespace', async function () {
+      const error = await db
+        .collection('a\x00b')
+        .insertOne({ a: 1 })
+        .catch(error => error);
+      expect(error['name']).to.equal('MongoServerError');
```
nbbeeken, line 103:
```suggestion
      expect(error).to.have.property('name', 'MongoServerError');
```

Just another bit of chai assertion improvement advice. By checking for "to have prop" and passing the value as the second argument if the assertion fails it will print the whole error object, which provides a good reference for what might have gone wrong if the test fails.

test/integration/collection-management/collection.test.ts
```diff
@@ -95,18 +95,14 @@ describe('Collection', function () {
...
+      expect(error['code']).to.equal(73);
```
nbbeeken, line 104: You can apply the same chai advice above to all of these. But just some JS knowledge share: `error['code']` and `error.code` are the same thing. Generally, you only ever need to access an object property using computed key syntax (brackets with a string) if the key isn't valid js (ex. `object['key-with-dash']`) or if your key is actually computed (ex. `for (const key of ['a', b']) object[key];`)

test/integration/node-specific/db.test.js
```diff
@@ -2,36 +2,33 @@
...
+
+    afterEach(async function () {
+      db = undefined;
+      await client.close();
+    });
+
+    it('should throw error on server only', async function () {
+      db = client.db('a\x00b');
+      const error = await db.createCollection('spider').catch(error => error);
+      expect(error['name']).to.equal('MongoServerError');
```
nbbeeken, line 28: same chai improvements

src/db.ts
```diff
@@ -218,6 +216,8 @@ export class Db {
    * Create a new collection on a server with the specified options. Use this to create capped collections.
    * More information about command options available at https://www.mongodb.com/docs/manual/reference/command/create/
    *
+   * Collection namespace validation occurs at operation time
```
nbbeeken, line 219: This is fine, but perhaps a bit confusing. `createCollection` is an operation against the server, besides the error type returned there is not much of a distinction between the driver or the server validating the namespace. This sentence would fit better on `db.collection()` which is _not_ an operation against the sever so there will be no error until an operation is performed. If we want to keep a mention here about when validation is performed I would change this sentence to say "validation is performed server-side", rather than "operation time" as this method is operation time.

nbbeeken (changes requested):

test/integration/collection-management/collection.test.ts
nbbeeken: Also, I would go back to checking `instanceof` as it is the recommendation we give our users for filtering errors: https://github.com/mongodb/node-mongodb-native?tab=readme-ov-file#error-handling but I do think it would be great to keep the code number assertion as well.

aditi-khare-mongoDB: Sounds good, I'll push this in my next commit. I originally think ` instanceof` wasn't working because I didn't have my imports straight, but I've got it working fine now.

test/integration/collection-management/collection.test.ts
aditi-khare-mongoDB: Good to know this js advice, I'll keep it in mind!

test/integration/node-specific/db.test.js
aditi-khare-mongoDB: I'll add these changes to my next commit, as well.

src/db.ts
aditi-khare-mongoDB: @nbbeeken I was thinking something more verbose such as "Collection namespace validation is performed server-side once an collection operation is attempted," or does "Collection namespace validation is performed server-side" capture the same information?

src/db.ts
```diff
@@ -138,6 +138,7 @@ export class Db {
    *
    * @param client - The MongoClient for the database.
    * @param databaseName - The name of the database this instance represents.
+   * databaseName validation occurs at operation time
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good, I'll put it above the params in my next commit.

src/mongo_client.ts
```diff
@@ -308,6 +308,7 @@ const kOptions = Symbol('options');
  * client.db().collection('pets');
  * await client.insertOne({ name: 'spot', kind: 'dog' });
  * ```
+ * // Db namespace validation occurs at operation time
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Oh, I see, I put it in the right spot for my next commit.

src/db.ts
aditi-khare-mongoDB: I'll also be sure to add the db.collection API information in my next commit.

nbbeeken: > Collection namespace validation is performed server-side once an collection operation is attempted.

`an collection` -> `a collection`

But I think you can just say "Collection namespace validation is performed server-side" because `createCollection` is an operation, and it is kind of a database one. There is not really a concept of database vs. collection operations precisely but a collection is created inside a database so logically it is not really a collection operation.

nbbeeken (changes requested):

src/db.ts
aditi-khare-mongoDB: Sounds good, I'll change it the simpler wording.

nbbeeken (approved):
LGTM

src/db.ts
```diff
@@ -134,21 +134,20 @@ export class Db {
...
+   * @param options - Optional settings for Db construction.
    */
   constructor(client: MongoClient, databaseName: string, options?: DbOptions) {
     options = options ?? {};
 
     // Filter the options
     options = filterOptions(options, DB_OPTIONS_ALLOW_LIST);
 
-    // Ensure we have a valid db name
-    validateDatabaseName(databaseName);
```
nbbeeken, line 150: From slack db names should still prevent dots to avoid ambiguity

If we permit dots in the db name, code like this:
```ts
await client.db('aa.bb').command({ create: 'cc.dd.' })
```

Will create a collection named `"cc.dd."` in a database named `"aa"` (the `bb` section is removed silently)

nbbeeken (changes requested):

W-A-James (changes requested):
LGTM pending some API docs changes and a question

src/db.ts
```diff
@@ -294,6 +295,8 @@ export class Db {
   /**
    * Returns a reference to a MongoDB Collection. If it does not exist it will be created implicitly.
    *
+   * Collection namespace validation occurs at operation time.
+   *
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: I agree, I do not think server-side is clearer for the same reasons. I'll go ahead and change it for the next commit.

test/integration/node-specific/operation_examples.test.ts
```diff
@@ -1650,98 +1650,6 @@ describe('Operations', function () {
     }
   });
 
-  /**
-   * An example of illegal and legal renaming of a collection using a Promise.
-   *
-   * example-class Collection
-   * example-method rename
-   */
-  it('shouldCorrectlyRenameCollectionWithPromises', {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: This function tests our old checkCollectionName functionality. The error messages it checks for are not the same as the "{code: 73, codeName: 'Invalid Namespace'} error that the server throws.

src/db.ts
aditi-khare-mongoDB: I added these changes in, and I editted the PR description accordingly. I'm wondering if this is something we also want to note in the Release Highlight, and if so, how?

src/db.ts
```diff
@@ -148,6 +148,11 @@ export class Db {
     // Filter the options
     options = filterOptions(options, DB_OPTIONS_ALLOW_LIST);
 
+    // Ensure there are no dots in database name
+    if (typeof databaseName === 'string' && databaseName.indexOf('.') !== -1) {
+      throw new MongoAPIError(`database names cannot contain the character '.'`);
```
nbbeeken, line 153: Should probably be a MongoInvalidArgument error to match what it was before.

nbbeeken (changes requested):

src/db.ts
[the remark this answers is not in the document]
nbbeeken: LGTM

src/db.ts
nbbeeken: Release highlight LGTM

src/db.ts
aditi-khare-mongoDB: Sounds good, I'll add this into my next push.

nbbeeken (dismissed):

src/mongo_client.ts
```diff
@@ -571,6 +571,8 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
   /**
    * Create a new Db instance sharing the current socket connections.
    *
+   * Db name validation occurs at operation time
```
baileympearson, line 574: I suggest we strike this, this isn't fully accurate and it doesn't seem worthwhile explaining what validation occurs where.

aditi-khare-mongoDB: That's true, I'll go ahead and remove it

W-A-James (approved):

nbbeeken (approved):

baileympearson (changes requested):
question: should we at least check that the value passed to `new Db()` and `new Collection()` are of type string?

aditi-khare-mongoDB:
> question: should we at least check that the value passed to `new Db()` and `new Collection()` are of type string?

@baileympearson @nbbeeken @W-A-James 
If we use a number as a database name, no error is thrown. Since we now have to client-side error-check for 1) type checking, 2) ensuring no '.' character in db, it seems like edge cases might keep arising. Should we stick with our current release notes ("The Node Driver will no longer error-check for database naming and collection naming, except edge cases x and y") ?

nbbeeken:
I notice the database name does not encounter an error whereas the collection name will trigger a BadValue server error stating the collection name is the incorrect type. I think it is important to consider that our BSON library is the point at which types sent to the server matter, Javascript strings are not the only input that can be serialized as a BSON string, and if our driver does not _need_ access to the input then it should allow it to reach the BSON layer unmodified.

nbbeeken:
I notice the database name does not encounter an error whereas the collection name will trigger a `BadValue` server error stating the collection name is the incorrect type. I think it is important to consider that our BSON library is the point at which types sent to the server matter, javascript strings are not the only input that can be serialized as a BSON string, and if our driver does not _need_ access to the input then it should allow it to reach the BSON layer so it can be serialized to the correct value.

https://github.com/mongodb/node-mongodb-native/blob/main/src/cmap/connection.ts#L542
The reason database names do not error is because of this line that implicitly stringifies whatever value the db name is. I vaguely recall us discussing correcting this line to just pass in the db name directly without building the dotted string, should we consider correcting that instead?

aditi-khare-mongoDB:
https://github.com/mongodb/node-mongodb-native/blob/main/src/cmap/connection.ts#L542 Would correcting this code be part of this ticket? If so, how would we go about correcting this line / would it be more straight-forward to just error on non-string input for databases?

nbbeeken (approved):

nbbeeken (dismissed):

src/cmap/commands.ts
```diff
@@ -75,10 +74,13 @@ export class Query {
   partial: boolean;
   documentsReturnedIn?: string;
 
-  constructor(ns: string, query: Document, options: OpQueryOptions) {
+  constructor(public databaseName: string, query: Document, options: OpQueryOptions) {
```
baileympearson, line 77: for both constructors in this file - can we either declare all class properties in the constructor using `public/private` syntax or none?  whatever we choose, let's be consistent

src/cmap/command_monitoring_events.ts
```diff
@@ -181,9 +181,8 @@ const HELLO_COMMANDS = new Set(['hello', LEGACY_HELLO_COMMAND, LEGACY_HELLO_COMM
 
 // helper methods
 const extractCommandName = (commandDoc: Document) => Object.keys(commandDoc)[0];
-const namespace = (command: WriteProtocolMessageType) => command.ns;
-const databaseName = (command: WriteProtocolMessageType) => command.ns.split('.')[0];
-const collectionName = (command: WriteProtocolMessageType) => command.ns.split('.')[1];
+const namespace = (command: Query) => command.ns;
+const collectionName = (command: Query) => command.ns.split('.')[1];
```
baileympearson, line 185: Isn't this always just `$cmd`?

```ts
    const ns = `${databaseName}.$cmd`;
```

baileympearson (changes requested):

src/cmap/command_monitoring_events.ts
nbbeeken: Not quite because Query instances are never command monitored, this helper is never used. We have a lot of Query-handling logic that we can delete here but I'd say its worth another PR for that.

src/cmap/commands.ts
aditi-khare-mongoDB: I agree that will be easier for readability and consistency reasons.

nbbeeken (approved):

baileympearson (approved):

PR #3880, opened 2023-10-04
ci(NODE-5653): remove vars from ci config

nbbeeken:
#### What is changing?

- remove use of vars
- remove mongosh tests
- move vars to expansions

No

vars are being changed to not exist after functions this avoids any silent config issues.


.evergreen/ci_matrix_constants.js
```diff
@@ -5,7 +5,7 @@ const versions = [
   { codeName: 'hydrogen', versionNumber: 18 },
   { codeName: 'iron', versionNumber: 20 }
 ];
-const NODE_VERSIONS = versions.map(({ versionNumber }) => versionNumber).sort();
+const NODE_VERSIONS = versions.map(({ versionNumber }) => versionNumber).sort((a, b) => a - b);
```
baileympearson, line 8: why is this change necessary?

baileympearson (changes requested):

.evergreen/ci_matrix_constants.js
nbbeeken: It isn't the atlas search tests started failing b/c they were falling back to node 14 but they were set to use "latest lts" so I thought there was something wrong with the sorting. But they just used the wrong variable name NODE_LTS_NAME vs NODE_LTS_VERSION. 

However, without this comparator, sort doesn't perform what you would expect:
```
> [1,3,2,100].sort()
[ 1, 100, 2, 3 ]
```
I figure why not leave behind the correct thing so it never bites us

baileympearson:
```
> [1,3,2,100].sort()
[ 1, 100, 2, 3 ]
```
TIL! thanks.

baileympearson (approved):

PR #3886, opened 2023-10-10
Node 4849/log path client options

aditi-khare-mongoDB:

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -1224,8 +1224,11 @@ export const OPTIONS = {
       'useUnifiedTopology has no effect since Node.js Driver version 4.0.0 and will be removed in the next major version'
   } as OptionDescriptor,
   // MongoLogger
-  // TODO(NODE-4849): Tighten the type of mongodbLogPath
-  mongodbLogPath: { type: 'any' }
+  /**
+   * @internal
+   * TODO: NODE-5671 - remove internal flag
+   */
+  mongoLoggerClientOptions: { type: 'any' }
 } as Record<keyof MongoClientOptions, OptionDescriptor>;
 
 export const DEFAULT_OPTIONS = new CaseInsensitiveMap(
--- a/package-lock.json
+++ b/package-lock.json
@@ -123,6 +123,25 @@
         "node": ">=6.0.0"
       }
     },
+    "node_modules/@aws-crypto/crc32": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/crc32/-/crc32-3.0.0.tgz",
+      "integrity": "sha512-IzSgsrxUcsrejQbPVilIKy16kAT52EwB6zSaI+M3xxIhKh5+aldEyvI+z6erM7TCLB2BJsFrtHjp6/4/sr+3dA==",
+      "optional": true,
+      "peer": true,
+      "dependencies": {
+        "@aws-crypto/util": "^3.0.0",
+        "@aws-sdk/types": "^3.222.0",
+        "tslib": "^1.11.1"
+      }
+    },
+    "node_modules/@aws-crypto/crc32/node_modules/tslib": {
+      "version": "1.14.1",
+      "resolved": "https://registry.npmjs.org/tslib/-/tslib-1.14.1.tgz",
+      "integrity": "sha512-Xni35NKzjgMrwevysHTCArtLDpPvye8zV/0E4EyYn43P7/7qvQwPh9BGkHewbMulVntbigmcT7rdX3BNo9wRJg==",
+      "optional": true,
+      "peer": true
+    },
     "node_modules/@aws-crypto/ie11-detection": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/@aws-crypto/ie11-detection/-/ie11-detection-3.0.0.tgz",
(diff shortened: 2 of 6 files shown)


src/connection_string.ts
```diff
@@ -1224,8 +1224,11 @@ export const OPTIONS = {
       'useUnifiedTopology has no effect since Node.js Driver version 4.0.0 and will be removed in the next major version'
   } as OptionDescriptor,
   // MongoLogger
-  // TODO(NODE-4849): Tighten the type of mongodbLogPath
```
dariakp, line 1227: Do we believe that this implementation addresses NODE-4849?

src/connection_string.ts
```diff
@@ -1224,8 +1224,11 @@ export const OPTIONS = {
...
-  mongodbLogPath: { type: 'any' }
```
dariakp, line 1228: we still need to be able to specify mongodbLogPath in the connection string, so I don't think we can remove it from here;

we also need to make sure to add tests for specifying this option both in the connection string and as a constructor option

dariakp (changes requested):
It doesn't look like this PR is actually ready for review, so I am moving it back to draft. @aditi-khare-mongoDB Please make sure that any new PR stays in draft until it's actually ready for review.

src/connection_string.ts
aditi-khare-mongoDB: I'll fix this in my next commit, and unnest the option

test/unit/connection_string.test.ts
```diff
@@ -832,4 +834,47 @@ describe('Connection String', function () {
...
+    });
+
+    it('when option is invalid, it defaults to stderr', function () {
+      const client = new MongoClient('mongodb://a/?mongodbLogPath=stdnothing', {
+        [loggerFeatureFlag]: true
+      });
+      const log: Log = { t: new Date(), c: 'ConnectionStringInvalidOption', s: 'error' };
+      client.options.mongoLoggerOptions.logDestination.write(log);
+      expect(stderrStub.write).calledWith(inspect(log, { breakLength: Infinity, compact: true }));
+    });
```
durran, line 878: If we have cases where the language uses "when", prefer using surrounding context blocks for the tests to better organise the output.

```suggestion
  context('when the option is stderr', function () {
      it('it is accessible through mongoLogger.logDestination', function () {
        const client = new MongoClient('mongodb://a/?mongodbLogPath=stderr', {
          [loggerFeatureFlag]: true
        });
        const log: Log = { t: new Date(), c: 'ConnectionStringStdErr', s: 'error' };
        client.options.mongoLoggerOptions.logDestination.write(log);
        expect(stderrStub.write).calledWith(inspect(log, { breakLength: Infinity, compact: true }));
      });
    });

    context('when the option is stdout', function () {  
      it('it is accessible through mongoLogger.logDestination', function () {
        const client = new MongoClient('mongodb://a/?mongodbLogPath=stdout', {
          [loggerFeatureFlag]: true
        });
        const log: Log = { t: new Date(), c: 'ConnectionStringStdOut', s: 'error' };
        client.options.mongoLoggerOptions.logDestination.write(log);
        expect(stdoutStub.write).calledWith(inspect(log, { breakLength: Infinity, compact: true }));
      });
    });
 
   context('when the option is invalid', function () {
      it('it defaults to stderr', function () {
        const client = new MongoClient('mongodb://a/?mongodbLogPath=stdnothing', {
          [loggerFeatureFlag]: true
        });
        const log: Log = { t: new Date(), c: 'ConnectionStringInvalidOption', s: 'error' };
        client.options.mongoLoggerOptions.logDestination.write(log);
        expect(stderrStub.write).calledWith(inspect(log, { breakLength: Infinity, compact: true }));
      });
    });
```

durran: Can we also add tests for providing a `MongoLogWritable` as well?

test/unit/mongo_client.test.js
```diff
@@ -813,4 +815,60 @@ describe('MongoOptions', function () {
...
+    before(() => {
+      stdoutStub = sinon.stub(process.stdout);
+      stderrStub = sinon.stub(process.stderr);
+    });
+
+    after(() => {
+      sinon.restore();
+    });
+
+    it('when option is `stderr`, it is accessible through mongoLogger.logDestination', function () {
```
durran, line 834: Same here as the other comments to use context blocks for "when" clauses.

durran (changes requested):

test/unit/connection_string.test.ts
aditi-khare-mongoDB: Yes! I can, and I'll change the naming of the test blocks to fit conventions too.

durran (changes requested):

test/unit/mongo_client.test.js
```diff
@@ -813,4 +815,84 @@ describe('MongoOptions', function () {
...
+          [loggerFeatureFlag]: true,
+          mongodbLogPath: 'stderr'
+        });
+        const log = { t: new Date(), c: 'constructorStdErr', s: 'error' };
+        client.options.mongoLoggerOptions.logDestination.write(log);
+        expect(stderrStub.write).calledWith(inspect(log, { breakLength: Infinity, compact: true }));
+      });
+    });
+
+    context('when option is a Writable stream', function () {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: I didn't realize, I'll remove this test case.

durran (approved):

test/unit/mongo_client.test.js
```diff
@@ -1,4 +1,6 @@
-'use strict';
+import { inspect } from 'util';
+
+('use strict');
```
W-A-James, line 3: nit: should be the first line

test/unit/connection_string.test.ts
```diff
@@ -832,4 +834,53 @@ describe('Connection String', function () {
...
     });
   });
+
+  describe('when mongodbLogPath is in options', function () {
+    const loggerFeatureFlag = Symbol.for('@@mdb.enableMongoLogger');
+
+    let stderrStub;
+    let stdoutStub;
+
+    before(() => {
```
dariakp, line 844: This should probably be `beforeEach` and `afterEach` to be safe because otherwise your mocks won't get reset properly between test cases; your current assertions are resilient against this (so even though the 1st and 3rd cases are checking the same stderr stub, they are checking it for different things), but it would be easy to add a test case into this set that is inadvertently asserting against the first call into the stub instead of the test case's own call because the author didn't realize that the describe setup was using before and after hooks instead of beforeEach and afterEach.

test/unit/mongo_client.test.js
```diff
@@ -813,4 +815,73 @@ describe('MongoOptions', function () {
...
     });
   });
+
+  context('when mongodbLogPath is in options', function () {
+    const loggerFeatureFlag = Symbol.for('@@mdb.enableMongoLogger');
+
+    let stderrStub;
+    let stdoutStub;
+
+    before(() => {
```
dariakp, line 825: same comment here about preferring `beforeEach` and `afterEach`

dariakp (changes requested):

test/unit/connection_string.test.ts
aditi-khare-mongoDB: I see, my reasoning for having specific logs was so the tests didn't somehow access other tests data, but I'm now realizing that's because I was using before/after instead of `beforeEach/afterEach`. I'll change the functions.

test/unit/mongo_client.test.js
aditi-khare-mongoDB: sounds good, I'll change it!

test/unit/mongo_client.test.js
aditi-khare-mongoDB: Yeah, I was intrigued that the linter didn't have a problem with this, but I guess it sees it as a programmer judgement call.

W-A-James (approved):

dariakp (approved):

PR #3887, opened 2023-10-11
refactor(NODE-5675): refactor server selection and connection checkout to use abort signals for timeout management

baileympearson:
#### What is changing?

- A custom abort controller, `TimeoutController`, has been added that aborts its signal after a specified timeout
- Server selection and connection checkout have been refactored to use the `TimeoutController` instead of managing timeouts

No.

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -27,7 +27,7 @@ import {
 } from '../error';
 import { CancellationToken, TypedEventEmitter } from '../mongo_types';
 import type { Server } from '../sdam/server';
-import { type Callback, eachAsync, List, makeCounter } from '../utils';
+import { type Callback, eachAsync, List, makeCounter, TimeoutController } from '../utils';
 import { AUTH_PROVIDERS, connect } from './connect';
 import { Connection, type ConnectionEvents, type ConnectionOptions } from './connection';
 import {
@@ -101,7 +101,7 @@ export interface ConnectionPoolOptions extends Omit<ConnectionOptions, 'id' | 'g
 /** @internal */
 export interface WaitQueueMember {
   callback: Callback<Connection>;
-  timer?: NodeJS.Timeout;
+  timeoutController: TimeoutController;
   [kCancelled]?: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


src/cmap/connection_pool.ts
```diff
@@ -356,27 +356,29 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
-              ? this.waitQueueErrorMetrics()
-              : 'Timed out while checking out a connection from connection pool',
-            this.address
-          )
-        );
-      }, waitQueueTimeoutMS);
-    }
+    const waitQueueMember: WaitQueueMember = {
+      callback,
+      timeoutController: new TimeoutController(waitQueueTimeoutMS)
```
baileympearson, line 363: You might ask - why are we using AbortControllers and then manually consuming their signal, instead of making ConnectionPool.checkOut accept an abort signal?  Really, there are a few reasons but it's primarily with the eventual state of CSOT in mind (note: everything here applies to server selection as well):

- We need to clear any timeouts we set for connection checkout after we successfully obtain a connection.  Abort signals do not provide a "cancellation" interface - and if they did, it would break the Controller/Signal abstraction. Controllers control signals and signals notify other components of abortion.
- Given the above - in later work we will be introducing a TimeoutControllerFactory that is a property on the OperationContext.  So eventually it will look like:
```typescript
const waitQueueMember: WaitQueueMember = {
      callback,
      timeoutController: operationContext.timeoutFactory.controllerForConnectionCheckout()
}
```
- To answer "why don't we make connection checkout accept a signal instead and pass it in?" - it's two fold.  First - we need to cancel the controller like the first bullet mentions.  Second - mostly preference, but in terms of encapsulation, I'd rather have connection checkout be responsible for managing its own timeout.  If we pass a signal in as an option, we'd need to construct a timeout signal outside of every place we call ConnectionPool.checkout.

src/utils.ts
```diff
@@ -1254,3 +1255,30 @@ export async function request(
...
+ *
+ * If `timeout` is undefined or 0, the abort controller never aborts.
+ *
+ * This class provides two benefits over the built-in AbortSignal.timeout() method.
+ * - This class provides a mechanism for cancelling the timeout
+ * - This class supports infinite timeouts by interpreting a timeout of 0 as infinite.  This is
+ *    consistent with existing timeout options in the Node driver (serverSelectionTimeoutMS, for example).
+ * @internal
+ */
+export class TimeoutController extends AbortController {
```
baileympearson, line 1270: `TimeoutController` intentionally returns an "infinite" timeout when timeout is <=0 or undefined.  This unifies code paths for existing timeouts.  We now _always_ create a timeout controller to manage server selection and connection checkout, but only if a non-zero positive timeout was provided do we actually create a Nodejs timeout that cancels the request.

PR #3890, opened 2023-10-12
refactor(NODE-5675): refactor server selection and connection checkout to use abort signals for timeout management 

baileympearson:
#### What is changing?

- A custom abort controller, `TimeoutController`, has been added that aborts its signal after a specified timeout
- Server selection and connection checkout have been refactored to use the `TimeoutController` instead of managing timeouts

No.

### DIFF
--- a/src/cmap/connection_pool.ts
+++ b/src/cmap/connection_pool.ts
@@ -27,7 +27,7 @@ import {
 } from '../error';
 import { CancellationToken, TypedEventEmitter } from '../mongo_types';
 import type { Server } from '../sdam/server';
-import { type Callback, eachAsync, List, makeCounter } from '../utils';
+import { type Callback, eachAsync, List, makeCounter, TimeoutController } from '../utils';
 import { AUTH_PROVIDERS, connect } from './connect';
 import { Connection, type ConnectionEvents, type ConnectionOptions } from './connection';
 import {
@@ -101,7 +101,7 @@ export interface ConnectionPoolOptions extends Omit<ConnectionOptions, 'id' | 'g
 /** @internal */
 export interface WaitQueueMember {
   callback: Callback<Connection>;
-  timer?: NodeJS.Timeout;
+  timeoutController: TimeoutController;
   [kCancelled]?: boolean;
 }
 
@@ -356,27 +356,29 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       new ConnectionCheckOutStartedEvent(this)
     );
 
-    const waitQueueMember: WaitQueueMember = { callback };
     const waitQueueTimeoutMS = this.options.waitQueueTimeoutMS;
-    if (waitQueueTimeoutMS) {
-      waitQueueMember.timer = setTimeout(() => {
-        waitQueueMember[kCancelled] = true;
-        waitQueueMember.timer = undefined;
 
-        this.emitAndLog(
-          ConnectionPool.CONNECTION_CHECK_OUT_FAILED,
-          new ConnectionCheckOutFailedEvent(this, 'timeout')
-        );
-        waitQueueMember.callback(
-          new WaitQueueTimeoutError(
-            this.loadBalanced
-              ? this.waitQueueErrorMetrics()
-              : 'Timed out while checking out a connection from connection pool',
-            this.address
-          )
-        );
-      }, waitQueueTimeoutMS);
-    }
+    const waitQueueMember: WaitQueueMember = {
+      callback,
+      timeoutController: new TimeoutController(waitQueueTimeoutMS)
+    };
+    waitQueueMember.timeoutController.signal.onabort = () => {
+      waitQueueMember[kCancelled] = true;
+      waitQueueMember.timeoutController.clear();
+
+      this.emitAndLog(
+        ConnectionPool.CONNECTION_CHECK_OUT_FAILED,
+        new ConnectionCheckOutFailedEvent(this, 'timeout')
+      );
+      waitQueueMember.callback(
+        new WaitQueueTimeoutError(
+          this.loadBalanced
+            ? this.waitQueueErrorMetrics()
+            : 'Timed out while checking out a connection from connection pool',
+          this.address
+        )
+      );
+    };
 
     this[kWaitQueue].push(waitQueueMember);
     process.nextTick(() => this.processWaitQueue());
@@ -831,9 +833,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
           ConnectionPool.CONNECTION_CHECK_OUT_FAILED,
           new ConnectionCheckOutFailedEvent(this, reason, error)
         );
-        if (waitQueueMember.timer) {
-          clearTimeout(waitQueueMember.timer);
-        }
+        waitQueueMember.timeoutController.clear();
         this[kWaitQueue].shift();
         waitQueueMember.callback(error);
         continue;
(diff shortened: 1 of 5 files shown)


src/cmap/connection_pool.ts
```diff
@@ -356,27 +356,29 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
-          new WaitQueueTimeoutError(
-            this.loadBalanced
-              ? this.waitQueueErrorMetrics()
-              : 'Timed out while checking out a connection from connection pool',
-            this.address
-          )
-        );
-      }, waitQueueTimeoutMS);
-    }
+    const waitQueueMember: WaitQueueMember = {
```
baileympearson, line 361: You might ask - why are we using AbortControllers and then manually consuming their signal, instead of making ConnectionPool.checkOut accept an abort signal?  Really, there are a few reasons but it's primarily with the eventual state of CSOT in mind (note: everything here applies to server selection as well):

- We need to clear any timeouts we set for connection checkout after we successfully obtain a connection.  Abort signals do not provide a "cancellation" interface - and if they did, it would break the Controller/Signal abstraction. Controllers control signals and signals notify other components of abortion.
- Given the above - in later work we will be introducing a TimeoutControllerFactory that is a property on the OperationContext.  So eventually it will look like:
```typescript
const waitQueueMember: WaitQueueMember = {
      callback,
      timeoutController: operationContext.timeoutFactory.controllerForConnectionCheckout()
}
```
- To answer "why don't we make connection checkout accept a signal instead and pass it in?" - it's two fold.  First - we need to cancel the controller like the first bullet mentions.  Second - mostly preference, but in terms of encapsulation, I'd rather have connection checkout be responsible for managing its own timeout.  If we pass a signal in as an option, we'd need to construct a timeout signal outside of every place we call ConnectionPool.checkout.

src/utils.ts
```diff
@@ -1254,3 +1255,28 @@ export async function request(
...
+ *
+ * If `timeout` is undefined or \<=0, the abort controller never aborts.
+ *
+ * This class provides two benefits over the built-in AbortSignal.timeout() method.
+ * - This class provides a mechanism for cancelling the timeout
+ * - This class supports infinite timeouts by interpreting a timeout of 0 as infinite.  This is
+ *    consistent with existing timeout options in the Node driver (serverSelectionTimeoutMS, for example).
+ * @internal
+ */
+export class TimeoutController extends AbortController {
```
baileympearson, line 1270: `TimeoutController` intentionally returns an "infinite" timeout when timeout is <=0 or undefined.  This unifies code paths for existing timeouts.  We now _always_ create a timeout controller to manage server selection and connection checkout, but only if a non-zero positive timeout was provided do we actually create a Nodejs timeout that cancels the request.

src/cmap/connection_pool.ts
aditi-khare-mongoDB: I agree, considering the perspective how this piece of code relates to CSOT as a whole. What was the reasoning for not including this information in, say, the PR description so that team reviewers will be able to access it easier?

src/utils.ts
aditi-khare-mongoDB: This makes sense. Do you feel it would be beneficial to also capture this behavior in the PR description or the ticket information?

src/utils.ts
```diff
@@ -1254,3 +1255,28 @@ export async function request(
...
+  constructor(
+    timeout?: number,
```
nbbeeken, line 1272: I think we can default this argument to 0 and then remove the truth check before the greaterthan zero check right?

src/sdam/topology.ts
```diff
@@ -556,22 +556,20 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
-    const serverSelectionTimeoutMS = options.serverSelectionTimeoutMS;
-    if (serverSelectionTimeoutMS) {
-      waitQueueMember.timer = setTimeout(() => {
-        waitQueueMember[kCancelled] = true;
-        waitQueueMember.timer = undefined;
-        const timeoutError = new MongoServerSelectionError(
-          `Server selection timed out after ${serverSelectionTimeoutMS} ms`,
-          this.description
-        );
+    waitQueueMember.timeoutController.signal.onabort = () => {
```
nbbeeken, line 563: Can we use the addEventListener format? it is the recommended way to add events over the onXYZ format.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

src/cmap/connection_pool.ts
```diff
@@ -356,27 +356,29 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+      callback,
+      timeoutController: new TimeoutController(waitQueueTimeoutMS)
+    };
+    waitQueueMember.timeoutController.signal.onabort = () => {
```
nbbeeken, line 365: addEventListener here also

src/utils.ts
```diff
@@ -1254,3 +1255,28 @@ export async function request(
...
+    private timeoutId = timeout && timeout > 0 ? setTimeout(() => this.abort(), timeout) : null
```
nbbeeken, line 1273: abort accepts a reason, should we be giving it the error we want to reject with here?

src/utils.ts
```diff
@@ -1254,3 +1255,28 @@ export async function request(
...
+  ) {
```
nbbeeken, line 1274: Should all the timers we make be .unref()-ed? is there a scenario where this timer could outlast the resources that are actually meaningful to keeping the eventloop turning (sockets)?

src/utils.ts
```diff
@@ -1254,3 +1255,28 @@ export async function request(
...
+    super();
+  }
+
+  clear() {
+    this.timeoutId && clearTimeout(this.timeoutId);
```
nbbeeken, line 1279: I appreciate the brevity but mind switching to `if != null`? short circuiting in an expression is not as obvious as control flow

test/unit/utils.test.ts
```diff
@@ -1101,4 +1104,65 @@ describe('driver utils', function () {
...
+          timeoutController.clear();
+        });
+
+        it('it creates a timeout', () => {
+          // @ts-expect-error Accessing a private field on TimeoutController
+          expect(timeoutController.timeoutId).not.to.be.null;
+        });
+
+        it('times out after `timeout` milliseconds', () => {
+          expect(spy.called, 'spy was called after creation').to.be.false;
```
nbbeeken, line 1159: we can enhance this with the chai sinon assertions: to.be.called right?

nbbeeken (changes requested):

src/sdam/topology.ts
```diff
@@ -556,22 +556,20 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
+      waitQueueMember[kCancelled] = true;
```
nbbeeken, line 564: this handler is passed an DOM Event object, do we need to call preventDefault? (I have no idea, need to study up on web APIs)

src/utils.ts
baileympearson: No - notice the abort handling still happens at the callsite.  I think that's a cleaner SOC - the timeout controller is responsible with timing out and the callsite is responsible for handling the timeout.  Also, some errors rely on state that occur at the time of the error.  Server selection timeouts, for example, include a topology description of the topology at the time of the timeout.

test/unit/utils.test.ts
baileympearson: Would you prefer that I use the chai assertions?

src/sdam/topology.ts
baileympearson: No, prevent default stops the event handling chain.  In browser APIs, it's most commonly used to prevent the default behavior when a DOM element is clicked / handled (the `preventDefault` docs, for example, demonstrate how it can be used to enable a click handler from preventing a checkbox from being "checked" when it is clicked on)

src/utils.ts
baileympearson: No - when we close a client, we'll clear both the server selection queue and the connection checkout queue, cancelling all timeouts.

src/utils.ts
baileympearson: sure

test/unit/utils.test.ts
nbbeeken: sí, por favor

nbbeeken (approved):

PR #3896, opened 2023-10-17
refactor(NODE-5696): add async-iterator based socket helpers

baileympearson:
#### What is changing?

This PR is the first connection layer refactor.  It does three things

- Rename our existing commands (Response, Query, BinMsg and Msg) to more descriptive names.  Happy to revert this and put it in a separate PR.
- Creates a copy of the Connection class, called ModernConnection.  This class is unused and the connection layer refactor will be done on this class so that our existing connection layer can remain unaffected until the new connection layer is ready.
- Creates generic socket helpers.
-- read & readMany read wire protocol messages from a connection
-- writeCommand writes a wire protocol message to a connection 

These helpers are built on the `readWireProtocolMessages` async generator.  `readWireProtocolMessages` replaces the message stream - it buffers `data` emissions from the underlying socket until a full wire protocol message has been received.

I made some additional refactors but these are intentionally additive.  No existing functionality is impacted by the new code paths.

No

CSOT requires the ability to cancel socket read & socket write for each operation.  A generic socket read/write abstraction will make the CSOT implementation at the connection layer much more straightforward.

### DIFF
--- a/src/cmap/commands.ts
+++ b/src/cmap/commands.ts
@@ -4,7 +4,13 @@ import { MongoInvalidArgumentError, MongoRuntimeError } from '../error';
 import { ReadPreference } from '../read_preference';
 import type { ClientSession } from '../sessions';
 import type { CommandOptions } from './connection';
-import { OP_MSG, OP_QUERY } from './wire_protocol/constants';
+import {
+  compress,
+  Compressor,
+  type CompressorName,
+  uncompressibleCommands
+} from './wire_protocol/compression';
+import { OP_COMPRESSED, OP_MSG, OP_QUERY } from './wire_protocol/constants';
 
 // Incrementing request id
 let _requestId = 0;
@@ -25,7 +31,7 @@ const SHARD_CONFIG_STALE = 4;
 const AWAIT_CAPABLE = 8;
 
 /** @internal */
-export type WriteProtocolMessageType = Query | Msg;
+export type WriteProtocolMessageType = OpQueryRequest | OpMsgRequest;
 
 /** @internal */
 export interface OpQueryOptions extends CommandOptions {
@@ -52,7 +58,7 @@ export interface OpQueryOptions extends CommandOptions {
  * QUERY
  **************************************************************/
 /** @internal */
-export class Query {
+export class OpQueryRequest {
   ns: string;
   numberToSkip: number;
   numberToReturn: number;
@@ -96,7 +102,7 @@ export class Query {
     this.numberToSkip = options.numberToSkip || 0;
     this.numberToReturn = options.numberToReturn || 0;
     this.returnFieldSelector = options.returnFieldSelector || undefined;
-    this.requestId = Query.getRequestId();
+    this.requestId = options.requestId ?? OpQueryRequest.getRequestId();
 
     // special case for pre-3.2 find commands, delete ASAP
     this.pre32Limit = options.pre32Limit;
@@ -285,7 +291,7 @@ export interface OpResponseOptions extends BSONSerializeOptions {
 }
 
 /** @internal */
-export class Response {
+export class OpQueryResponse {
   parsed: boolean;
   raw: Buffer;
   data: Buffer;
@@ -472,7 +478,7 @@ export interface OpMsgOptions {
 }
 
 /** @internal */
-export class Msg {
+export class OpMsgRequest {
   requestId: number;
   serializeFunctions: boolean;
   ignoreUndefined: boolean;
@@ -502,7 +508,7 @@ export class Msg {
     this.options = options ?? {};
 
     // Additional options
-    this.requestId = options.requestId ? options.requestId : Msg.getRequestId();
+    this.requestId = options.requestId ? options.requestId : OpMsgRequest.getRequestId();
 
     // Serialization option
     this.serializeFunctions =
@@ -580,7 +586,7 @@ export class Msg {
 }
 
 /** @internal */
-export class BinMsg {
+export class OpMsgResponse {
   parsed: boolean;
   raw: Buffer;
   data: Buffer;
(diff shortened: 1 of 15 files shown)


src/cmap/commands.ts
```diff
@@ -709,3 +715,54 @@ export class BinMsg {
...
+
+const MESSAGE_HEADER_SIZE = 16;
+const COMPRESSION_DETAILS_SIZE = 9; // originalOpcode + uncompressedSize, compressorID
+
+/**
+ * @internal
+ *
+ * An OP_COMPRESSED request wraps either an OP_QUERY or OP_MSG message.
+ */
+export class OpCompressedRequest {
```
baileympearson, line 727: I like this abstraction but would be happy to remove it.  I think it makes sense conceptually - an OP_COMPRESSED wraps either an OP_QUERY or an OP_MSG.  However, our message _parsing_ logic is structured differently.  We first decompress a response into an OP_MSG or OP_QUERY, and then later we parse the response.  consequently, it's not as straightforward to create an OpCompresedResponse with the current structure.   If we'd prefer consistency, I can get rid of this class.

src/cmap/commands.ts
```diff
@@ -709,3 +715,54 @@ export class BinMsg {
...
+  // Return whether a command contains an uncompressible command term
+  // Will return true if command contains no uncompressible command terms
+  static canCompress(command: WriteProtocolMessageType) {
+    const commandDoc = command instanceof OpMsgRequest ? command.command : command.query;
+    const commandName = Object.keys(commandDoc)[0];
+    return !uncompressibleCommands.has(commandName);
+  }
+
+  async toBin(): Promise<Buffer[]> {
+    const concatenatedOriginalCommandBuffer = Buffer.concat(this.command.toBin());
```
nbbeeken, line 742: Is it necessary to concat here? Doesn't toBin return the chunks already sliced appropriately?

Same concern for `compressCommand`

It is not for this PR to fix but I think we identified that we should be writing chunks to the stream rather than reallocing and copying bytes. I would hope the code we introduce could facilitate that improvement.

src/cmap/commands.ts
```diff
@@ -709,3 +715,54 @@ export class BinMsg {
...
+  constructor(
+    private command: WriteProtocolMessageType,
+    private options: { zlibCompressionLevel: number; agreedCompressor: CompressorName }
+  ) {}
+
+  // Return whether a command contains an uncompressible command term
+  // Will return true if command contains no uncompressible command terms
+  static canCompress(command: WriteProtocolMessageType) {
+    const commandDoc = command instanceof OpMsgRequest ? command.command : command.query;
+    const commandName = Object.keys(commandDoc)[0];
```
nbbeeken, line 737: enumerating all keys and checking the 0th would be no different from checking for the key existing in the commandDoc right? Ex. a `find` can't have `saslContinue` as a top-level key.

src/cmap/connection.ts
```diff
@@ -757,3 +769,534 @@ function write(
...
+ * the stream by calling `return` on the generator.
+ *
+ * Note that `for-await` loops call `return` automatically when the loop is exited.
+ */
+export async function* readWireProtocolMessages(
+  connection: ModernConnection
+): AsyncGenerator<Buffer> {
+  const bufferPool = new BufferPool();
+  const maxBsonMessageSize = connection.hello?.maxBsonMessageSize ?? kDefaultMaxBsonMessageSize;
+  for await (const [chunk] of on(connection.socket, 'data')) {
```
nbbeeken, line 1219: `on` increases our footprint of needed dependencies from `events`. We went out of our way to import `events` in a nested path in FLE to avoid impacting our work to get the driver operational in other runtimes.

I believe we could use readable's asyncIterator method reducing the contract for what we need out of a socket. We can also follow up with an improvement if it is overly difficult to achieve.

src/cmap/connection.ts
```diff
@@ -757,3 +769,534 @@ function write(
...
+  /** @internal */
+  [kMessageStream]: MessageStream;
+  /** @internal */
+  socket: Stream;
+  /** @internal */
+  [kHello]: Document | null;
+  /** @internal */
+  [kClusterTime]: Document | null;
+
+  bufferPool: BufferPool;
```
nbbeeken, line 813: This doesn't exist on the Connection class so what other changes are there to this class?

src/cmap/connection.ts
```diff
@@ -757,3 +769,534 @@ function write(
...
+ *
+ * By itself, produces an infinite async generator of wire protocol messages and consumers must end
+ * the stream by calling `return` on the generator.
+ *
+ * Note that `for-await` loops call `return` automatically when the loop is exited.
+ */
+export async function* readWireProtocolMessages(
+  connection: ModernConnection
+): AsyncGenerator<Buffer> {
+  const bufferPool = new BufferPool();
```
nbbeeken, line 1217: Was this supposed to be the pool on the connection?

nbbeeken (changes requested):
@durran I already requested your review since I will be out soon, you might as well take up the primary review reins.

Left thoughts, didn't look at the tests wanted to get you something to spur discussion on some of the decisions.

test/unit/index.test.ts
```diff
@@ -133,10 +133,12 @@ const EXPECTED_EXPORTS = [
 ];
 
 describe('mongodb entrypoint', () => {
-  it('should export all and only the expected keys in expected_exports', () => {
-    expect(sorted(Object.keys(mongodb), alphabetically)).to.deep.equal(
-      sorted(EXPECTED_EXPORTS, alphabetically)
-    );
+  it('exports all the expected keys', () => {
+    expect(setDifference(EXPECTED_EXPORTS, Object.keys(mongodb))).to.be.empty;
```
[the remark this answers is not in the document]
baileympearson: The assertion confused me lol, I changed it for readability

src/cmap/commands.ts
baileympearson: No, `toBin()` currently returns an array of buffers.  There's no change in the existing serialization behavior here.

Is this something you'd want me to fix now?  I can, but I intentionally limited the changes to the connection layer (and some refactors).  The core serialization and compression logic is unchanged

src/cmap/commands.ts
baileympearson: Can we say that confidently for all commands? 🤔 

While maybe a worthwhile improvement, I'd rather file a follow up ticket for that.  Or even better, just pass the operation down into the connection layer and use an `INCOMPRESSIBLE` aspect on the operation here instead

src/cmap/connection.ts
baileympearson: None, besides renaming `kStream` to `socket`

src/cmap/connection.ts
baileympearson: no

src/cmap/commands.ts
nbbeeken: I am fairly (99.99999%) certain we can but fair enough if we want to file something and leave a TODO.

ismaster, saslStart, saslContinue, are the only commands we make in the driver but they are not operations. The rest can be run by someone else via runCommand so the decision does have to be made based on the command.

src/cmap/connection.ts
nbbeeken: Why not, I'm surprised we need to make a new pool on every read? I see the one on the class is removed now, was there an issue with preserving it across reads?

baileympearson: It's a leaky abstraction.  If `readWireProtocolMessages ` is the single abstraction we use to read full wire protocol messages from the socket, we shouldn't store the state it needs on the connection where it's not needed.

If this proves to be an impact on performance, we can reconsider

src/cmap/commands.ts
durran: I'd say a follow up is fine but it seems the server expects the first key to always be the command name and if another command name is in the top level the server will error:

```shell
Enterprise test> db.runCommand({ find: 'test', saslContinue:  {} })
MongoServerError: BSON field 'FindCommandRequest.saslContinue' is an unknown field.
Enterprise test> 
```

durran (approved):

src/cmap/commands.ts
baileympearson: Done: https://jira.mongodb.org/browse/NODE-5740

durran (approved):

PR #3899, opened 2023-10-19
feat(NODE-5197): add server monitoring mode

durran:


test/lambda/mongodb/app.mjs
W-A-James, line 1: Out of curiosity, where do we run this file?

src/sdam/monitor.ts
```diff
@@ -207,27 +223,37 @@ function resetMonitorState(monitor: Monitor) {
   monitor[kConnection] = undefined;
 }
 
+function useStreamingProtocol(monitor: Monitor, topologyVersion: TopologyVersion | null): boolean {
+  // If we have no topology version we always poll no matter
+  // what the user provided.
+  if (topologyVersion == null) return false;
```
W-A-James, line 229: For clarification, in this case, the server doesn't support streaming at all right?

src/mongo_client.ts
W-A-James, line 1: On account of this exposing a new URI and constructor option, can you mark the ticket as Docs Changes needed with the accompanying information?

test/lambda/mongodb/app.mjs
durran: This is run in our deployed lambda tests. We pass the lambda directory here https://github.com/mongodb/node-mongodb-native/blob/main/.evergreen/config.in.yml#L1278 and then when we `sam deploy` the function this is set up in the template to be called.

src/mongo_client.ts
durran: Ticket docs changes updated.

src/sdam/monitor.ts
durran: Yes, I've updated the comment to reflect that.

src/sdam/monitor.ts
```diff
@@ -286,15 +312,12 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
...
 
-      const awaited = isAwaitable && hello.topologyVersion != null;
       monitor.emit(
         Server.SERVER_HEARTBEAT_SUCCEEDED,
-        new ServerHeartbeatSucceededEvent(monitor.address, duration, hello, awaited)
+        new ServerHeartbeatSucceededEvent(monitor.address, duration, hello, isAwaitable)
       );
 
-      // if we are using the streaming protocol then we immediately issue another `started`
-      // event, otherwise the "check" is complete and return to the main monitor loop
```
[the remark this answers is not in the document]
durran: I've added back, was just missed.

W-A-James (dismissed):

src/sdam/monitor.ts
```diff
@@ -44,6 +45,11 @@ function isInCloseState(monitor: Monitor) {
   return monitor.s.state === STATE_CLOSED || monitor.s.state === STATE_CLOSING;
 }
 
+/** @public */
+export const ServerMonitoringModes = ['auto', 'poll', 'stream'];
```
nbbeeken, line 49: If this is to be a public API then wouldn't an enum be more ergonomic? Then we can use `Object.values()` to get the `.includes` in the options parsing working.

test/lambda/mongodb/package.json
```diff
@@ -5,6 +5,6 @@
   "main": "app.mjs",
   "license": "MIT",
   "dependencies": {
-    "mongodb": "^5.0.1"
+    "mongodb": "mongodb/node-mongodb-native#NODE-5197-server-monitoring-mode"
```
nbbeeken, line 8: This will fail when we merge this and GH auto deletes this branch. Since we've observed it run green, do we want to change it back here? Also, it seems wrong that this was installing 5.x, can we specify a file directory like ../.. so it uses the current driver? (we can defer that if there are setup issues)

nbbeeken (changes requested):

test/lambda/mongodb/package.json
durran: Yes I'll change this back to main. Pointing it to a file fails on `sam build` as it somehow gets into an infinite loop doing the npm install, so I needed to go to Github instead.

src/sdam/monitor.ts
durran: I've updated to be an enum now and updated the options parsing.

test/lambda/mongodb/package.json
durran: Updated to use main.

nbbeeken (approved):

PR #3902, opened 2023-10-19
feat(NODE-5452):  Logging Cosmos Document DB Info Message

aditi-khare-mongoDB:
### Description
A user connecting to a `ConnectionString` object while using a Cosmos/DocumentDB host will not have the same client functionality as a user using a genuine host. After these changes are merged, when a user provides a non-genuine host, a info-level Client component log will be outputted to the `mongodbLogPath`.

#### What is changing?
If Standardized Logging is turned on and environment settings allow for Client info-level logs, when a user provides one or more CosmosDB/DocumentDB hosts, an info-level client component log(s) will be outputted to the stream provided in `mongodbLogPath`.

Yes, it should be included in the documentation for Standardized Logging when Standardized Logging is released.

To notify users when we detect a CosmosDB/DocumentDB host, since they have limited functionality in MongoClient.


src/mongo_client.ts
```diff
@@ -252,6 +252,11 @@ export interface MongoClientOptions extends BSONSerializeOptions, SupportedNodeC
   srvPoller?: SrvPoller;
   /** @internal */
   connectionType?: typeof Connection;
+  /**
+   * @internal
+   * TODO: NODE-5671 - remove internal flag
+   */
```
aditi-khare-mongoDB, line 258: This and and mongo_client.test.js changes are from NODE-4849, which this branch branches off of. I don't know how to have it only show the NODE-5452 changes.

durran (changes requested):
Can we rebase against main and resolve the conflicts so we get a clean diff?

src/mongo_client.ts
```diff
@@ -364,6 +364,47 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
     };
+    this.checkForNonGenuineHosts();
+  }
+
+  /** @internal */
+  private checkForNonGenuineHosts() {
+    if (this.mongoLogger && this.mongoLogger.logDestination) {
+      const documentDBHostnames = this[kOptions].hosts.filter(
+        (hostAddress: HostAddress) =>
+          hostAddress.host &&
```
durran, line 375: I'd like to propose a few things here.

First, I think we should extract all the log messages to constants so we're not instantiating the same strings over and over again with each instantiation of a `MongoClient`, and since they are not dynamic.

Then we've got a pattern here of how we check if a host matches a specific domain, so we can extract a method to handle that common behaviour. And we can simplify that behaviour by using a Regex instead of potential multiple `endsWith` calls.

Example:
```typescript
const DOCUMENT_DB_CHECK = /(\.docdb\.amazonaws\.com$)|(\.docdb-elastic\.amazonaws\.com$)/
const COSMOS_DB_CHECK = /\.cosmos\.azure\.com$/

function isHostMatch(match: RegExp, host?: string): boolean {
  return host && match.test(host);
}
```

Then in the code we could refactor to do something like:

```typescript
const documentDbHostnames = this[kOptions].hosts.filter(
  (hostAddress: HostAddress) => isMostMatch(DOCUMENT_DB_CHECK, hostAddress)
);
const srvHostIsDocumentDB = isHostMatch(DOCUMENT_DB_CHECK, this[kOptions].srvHost);
```

Thoughts?

durran (changes requested):

src/mongo_client.ts
aditi-khare-mongoDB: I agree with both suggestions, I'll implement these changes.

aditi-khare-mongoDB: What are your thoughts on having Document_DB_Check and Cosmos_DB_Check as static class variables of MongoClient? Since they won't be used outside of the class, and classes cannot have const variables in JS.

src/mongo_client.ts
```diff
@@ -364,6 +364,45 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+    this.checkForNonGenuineHosts();
+  }
+
+  DOCUMENT_DB_CHECK = /(\.docdb\.amazonaws\.com$)|(\.docdb-elastic\.amazonaws\.com$)/;
+  COSMOS_DB_CHECK = /\.cosmos\.azure\.com$/;
+
+  static DOCUMENT_DB_MSG =
+    'You appear to be connected to a DocumentDB cluster. For more information regarding feature compatibility and support please visit https://www.mongodb.com/supportability/documentdb';
+  static COSMOS_DB_MSG =
+    'You appear to be connected to a CosmosDB cluster. For more information regarding feature compatibility and support please visit https://www.mongodb.com/supportability/cosmosdb';
```
durran, line 376: I'd prefer these to all be `const` outside of the `MongoClient` itself, since they aren't really properties of the class or its prototype. And now that I think about this more, I think this functionality is probably all better served in the utils module. So could we move the constants there as well as the `isHostMatch()` function?

durran (changes requested):

src/mongo_client.ts
durran: See other comment, I think they should actually be moved.

src/mongo_client.ts
aditi-khare-mongoDB: That makes sense!

durran (approved):

src/mongo_client.ts
```diff
@@ -364,6 +369,32 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
         return true;
       }
     };
+    this.checkForNonGenuineHosts();
+  }
+
+  /** @internal */
+  private checkForNonGenuineHosts() {
+    if (this.mongoLogger && this.mongoLogger.logDestination) {
```
W-A-James, line 377: The mongoLogger is always defined on the client (see src/mongo_client.ts:342). You also shouldn't need to check that the logDestination exists. It will always be set in the MongoLogger constructor (see src/mongo_logger.ts:479).

The severity logging methods check if a message is to be emitted on the component provided at the specified severity level, so we don't have to do these checks.

W-A-James (changes requested):

src/mongo_client.ts
aditi-khare-mongoDB: ahh, I see, I'll remove these.

src/mongo_client.ts
```diff
@@ -364,6 +369,32 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+      const cosmosDBHostnames = this[kOptions].hosts.filter((hostAddress: HostAddress) =>
+        isHostMatch(COSMOS_DB_CHECK, hostAddress.host)
+      );
+
+      const srvHostIsCosmosDB = isHostMatch(COSMOS_DB_CHECK, this[kOptions].srvHost);
+
+      if (cosmosDBHostnames.length !== 0 || srvHostIsCosmosDB) {
+        this.mongoLogger.info('client', COSMOS_DB_MSG);
+      }
+    }
```
W-A-James: Would it ever be the case that we would match both CosmosDB and DocumentDB? If not, I think we should make that clear with the control flow by making the two calls to `this.mongoLogger.info` live in separate branches of an if statement. Otherwise, disregard.

alexbevi: If you find one and print the message that should be it. Though it shouldn't ever be possible to match both based on the regex, even if it were the case only one message would need to be displayed.

aditi-khare-mongoDB: it is "technically" possible to match both CosmosDB and DocumentDB, but extremely unlikely, I'll refactor these to live in separate branches. We also want to comply with [DRIVERS-2583](https://jira.mongodb.org/browse/DRIVERS-2583)'s "Drivers should not emit more than one log message per MongoClient construction" directive, and not allow more than one log message per client anyway.

W-A-James (approved):

durran (approved):

PR #3905, opened 2023-10-30
feat(NODE-5590): deprecate GridFS fields

alenakhineika:
### Description
Deprecate unused GridFS fields.

#### What is changing?
Deprecates `contentType` and `aliases` options.

None

[NODE-5590](https://jira.mongodb.org/browse/NODE-5590)

### Release Highlight

The GridFS `contentType` and `aliases` options are deprecated. According to [the GridFS spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst), applications wishing to store `contentType` and `aliases` should add a corresponding field to the `metadata` document instead.

### DIFF
--- a/src/gridfs/download.ts
+++ b/src/gridfs/download.ts
@@ -46,13 +46,9 @@ export interface GridFSFile {
   filename: string;
   metadata?: Document;
   uploadDate: Date;
-  /**
-   * @deprecated Will be removed in the next major version.
-   */
+  /** @deprecated Will be removed in the next major version. */
   contentType?: string;
-  /**
-   * @deprecated Will be removed in the next major version.
-   */
+  /** @deprecated Will be removed in the next major version. */
   aliases?: string[];
 }
 
--- a/src/gridfs/download.ts
+++ b/src/gridfs/download.ts
@@ -44,10 +44,16 @@ export interface GridFSFile {
   length: number;
   chunkSize: number;
   filename: string;
-  contentType?: string;
-  aliases?: string[];
   metadata?: Document;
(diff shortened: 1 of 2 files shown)


src/gridfs/download.ts
```diff
@@ -44,10 +44,16 @@ export interface GridFSFile {
...
   chunkSize: number;
   filename: string;
-  contentType?: string;
-  aliases?: string[];
   metadata?: Document;
   uploadDate: Date;
+  /**
+   * @deprecated Will be removed in the next major version.
+   */
+  contentType?: string;
```
W-A-James, line 52: minor nit: Since we don't have any other inline documentation here, we can have the deprecation notice be on one line.
```suggestion
  /** @deprecated Will be removed in the next major version. */
  contentType?: string;
```

src/gridfs/download.ts
```diff
@@ -44,10 +44,16 @@ export interface GridFSFile {
...
+  /**
+   * @deprecated Will be removed in the next major version.
+   */
```
W-A-James, line 55:
```suggestion
  /** @deprecated Will be removed in the next major version. */
```

W-A-James (changes requested):

src/gridfs/download.ts
alenakhineika: Updated.

src/gridfs/download.ts
alenakhineika: Updated.

W-A-James (approved):

PR #3907, opened 2023-11-01
feat(NODE-3881): require hello command + OP_MSG when 'loadBalanced=True'

alenakhineika:
### Description
Send the `hello` command + `OP_MSG` when `loadBalanced` is true according to [spec](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst).

#### What is changing?
- When `loadBalanced` is true, we send the `hello` command over the `legacy hello` command.
- When `loadBalanced` is true we send `OP_MSG` over `OP_QUERY`.

None

[NODE-3881](https://jira.mongodb.org/browse/NODE-3881)


src/cmap/connection.ts
```diff
@@ -637,8 +637,17 @@ export function hasSessionSupport(conn: Connection): boolean {
...
 }
 
-function supportsOpMsg(conn: Connection) {
+/** @internal */
+export function supportsOpMsg(conn: Connection, options: { loadBalanced?: boolean }) {
+  // If loadBalanced is true, then we MUST send the initial hello command using OP_MSG.
+  // Since this is the first message and hello/legacy hello hasn't been sent yet,
+  // conn.description will be null and we can't rely on the server check to determine if
+  // the server supports OP_MSG.
   const description = conn.description;
```
baileympearson, line 646: The stream description is constructed in `Connection.constructor` - so it will never be null (the `description == null` check below is always false).  The stream description has a `loadBalanced` property that is already set and the connection has a `Connection.loadBalanced` property that returns whether the stream description is load balanced. Can we use the description here instead of providing a new option?

test/unit/cmap/connection.test.ts
```diff
@@ -894,6 +895,78 @@ describe('new Connection()', function () {
     });
   });
 
+  describe('supportsOpMsg', function () {
```
durran, line 898: Just nit-picky here to indicate a floating function. :)

```suggestion
  describe('.supportsOpMsg', function () {
```

durran (changes requested):

baileympearson (changes requested):

test/unit/cmap/connect.test.ts
```diff
@@ -242,6 +242,7 @@ describe('Connect Tests', function () {
           };
           const handshakeDocument = await prepareHandshakeDocument(authContext);
           expect(handshakeDocument).not.to.have.property('loadBalanced');
+          expect(handshakeDocument).to.have.property(LEGACY_HELLO_COMMAND, 1);
```
W-A-James, line 245: I think in this case, to surface that it is an explicit requirement that we have `LEGACY_HELLO_COMMAND` set to 1, we should break this assertion out into its own `it` block with the set up of the handshake document done in a `beforeEach`. That lets us have a test case in our output that says something like

```
when loadBalanced is not set as an option
	does not set loadBalanced on the handshake document ✅ 
	sets LEGACY_HELLO_COMMAND to 1			    ✅ 	
```

test/unit/cmap/connect.test.ts
```diff
@@ -256,6 +257,7 @@ describe('Connect Tests', function () {
...
```
W-A-James, line 260: Same change as above

test/unit/cmap/connect.test.ts
```diff
@@ -270,6 +272,7 @@ describe('Connect Tests', function () {
           };
           const handshakeDocument = await prepareHandshakeDocument(authContext);
           expect(handshakeDocument).to.have.property('loadBalanced', true);
+          expect(handshakeDocument).to.have.property('hello', 1);
```
W-A-James, line 275: Same change as above

W-A-James (changes requested):

src/cmap/connection.ts
durran: Per the Slack thread it seems this was the case - the issue was that we were sending `ismaster` over `OP_MSG` instead of `hello`.

baileympearson: Ah okay.  So the fix would be in `connect`, where we construct the handshake?

alenakhineika: It might be that this fix is required:
https://github.com/mongodb/node-mongodb-native/pull/3907/files#diff-7cc25e5d3247913cdf1fe1e7788e951e4435bb1091d6b8aef135c4b171c7c997R217

But the change in supportsOpMsg is not?

durran (approved):

W-A-James (approved):

src/cmap/connect.ts
```diff
@@ -214,7 +214,7 @@ export async function prepareHandshakeDocument(
   const { serverApi } = authContext.connection;
 
   const handshakeDoc: HandshakeDocument = {
-    [serverApi?.version ? 'hello' : LEGACY_HELLO_COMMAND]: 1,
+    [serverApi?.version || options.loadBalanced === true ? 'hello' : LEGACY_HELLO_COMMAND]: 1,
```
nbbeeken, line 217:
```ts
[serverApi?.version || options.loadBalanced ? 'hello' : LEGACY_HELLO_COMMAND]: 1,
```

Is there something specific we are looking for by explicitly checking for `true`? I've seen reasons to check for precisely `false`, to differentiate falsey from truly `false` but not the other way around 🤔

nbbeeken (changes requested):

src/cmap/connect.ts
alenakhineika: A bit lower in the code we explicitly check for `true` to then set the `handshakeDoc.loadBalanced` value:
https://github.com/mongodb/node-mongodb-native/blob/NODE-3881-hello-plus-op_msg/src/cmap/connect.ts#L223

And the [spec](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst) says `If a server API version is requested or loadBalanced: True`. I feel like this is a matching behaviour in this case.

nbbeeken (approved):

PR #3912, opened 2023-11-07
feat(NODE-5464): OIDC machine and callback workflow

durran:
### Description

Implements OIDC new machine and human callback workflows.

#### What is changing?
- Implements the OIDC callback workflow. Specified with `OIDC_CALLBACK` auth mech property.
- Implements the OIDC human callback workflow. Specified with `OIDC_HUMAN_CALLBACK` auth mech property.
- Implements the OIDC Test machine workflow. Specified with `ENVIRONMENT:test` auth mech property.
- Implements the OIDC Azure machine workflow. Specified with `ENVIRONMENT:azure` auth mech property.
- Implements the OIDC GCP machine workflow. Specified with `ENVIRONMENT:gcp` auth mech property.
- Uses a new generic `TokenCache` for all OIDC authentication that sits at the auth provider level.
- Removes the old complex callback workflow global caching.
- Removes the old global Azure token cache.

https://github.com/mongodb/specifications/pull/1471
https://github.com/mongodb/specifications/pull/1544
https://github.com/mongodb/specifications/pull/1513

### Support for MONGODB-OIDC Authentication

`MONGODB-OIDC` is now supported as an authentication mechanism for MongoDB server versions 7.0+. The currently supported facets to authenticate with are callback authentication, human interaction callback authentication, Azure machine authentication, and GCP machine authentication.

#### Azure Machine Authentication

The `MongoClient` must be instantiated with `authMechanism=MONGODB-OIDC` in the URI or in the client options. Additional required auth mechanism properties of `TOKEN_RESOURCE` and `ENVIRONMENT` are required and another optional username can be provided. Example:

```ts
const client = new MongoClient('mongodb+srv://<username>@<host>:<port>/?authMechanism=MONGODB-OIDC&authMechanismProperties=TOKEN_RESOURCE:<azure_token>,ENVIRONMENT=azure');
await client.connect();
```

#### GCP Machine Authentication

The `MongoClient` must be instantiated with `authMechanism=MONGODB-OIDC` in the URI or in the client options. Additional required auth mechanism properties of `TOKEN_RESOURCE` and `ENVIRONMENT` are required. Example:

```ts
const client = new MongoClient('mongodb+srv://<host>:<port>/?authMechanism=MONGODB-OIDC&authMechanismProperties=TOKEN_RESOURCE:<gcp_token>,ENVIRONMENT=gcp');
await client.connect();
```

#### Callback Authentication

The user can provide a custom callback to the `MongoClient` that returns a valid response with an access token. The callback is provided as an auth mechanism property an has the signature of:

```ts
const oidcCallBack = (params: OIDCCallbackParams): Promise<OIDCResponse> => {
  // params.timeoutContext is an AbortSignal that will abort after 30 seconds for non-human and 5 minutes for human.
  // params.version is the current OIDC API version.
  // params.idpInfo is the IdP info returned from the server.
  // params.username is the optional username.

  // Make a call to get a token.
  const token = ...;
  return {
     accessToken: token,
     expiresInSeconds: 300,
     refreshToken: token
  };
}

const client = new MongoClient('mongodb+srv://<host>:<port>/?authMechanism=MONGODB-OIDC', {
  authMechanismProperties: {
    OIDC_CALLBACK: oidcCallback
  }
});
await client.connect();
```

For callbacks that require human interaction, set the callback to the `OIDC_HUMAN_CALLBACK` property:

```ts
const client = new MongoClient('mongodb+srv://<host>:<port>/?authMechanism=MONGODB-OIDC', {
  authMechanismProperties: {
    OIDC_HUMAAN_CALLBACK: oidcCallback
  }
});
await client.connect();
```


addaleax (commented):
Looks all pretty good so far, started to do a review with reminders about TODO comments but then I just didn't have any other ones for now :)

.evergreen/prepare-shell.sh
```diff
@@ -31,7 +31,7 @@ export PATH="$MONGODB_BINARIES:$PATH"
 
 if [ ! -d "$DRIVERS_TOOLS" ]; then
   # Only clone driver tools if it does not exist
-  git clone --depth=1 "https://github.com/mongodb-labs/drivers-evergreen-tools.git" "${DRIVERS_TOOLS}"
+  git clone --depth=1 --branch DRIVERS-2601 "https://github.com/blink1073/drivers-evergreen-tools.git" "${DRIVERS_TOOLS}"
```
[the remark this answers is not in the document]
durran: Thanks. Yeah I needed to test GCP there on that branch but will move back to master once it merges.

test/unit/cmap/auth/mongodb_oidc/azure_machine_workflow.test.ts
```diff
@@ -0,0 +1,24 @@
...
+describe('AzureMachineFlow', function () {
+  describe('#execute', function () {
+    const workflow = new AzureMachineWorkflow();
+
+    context('when TOKEN_RESOURCE is not set', function () {
+      const connection = sinon.createStubInstance(Connection);
+      const credentials = sinon.createStubInstance(MongoCredentials);
+
+      it('throws an error', async function () {
+        try {
```
nbbeeken, line 15: A little confused here, are you looking for the execute to throw an error with "TOKEN_RESOURCE" in the message or not? the expect.fail cannot contain the string you are expecting to only come out of workflow.execute if so. 

Personally, I would remove the control flow risk from the test, but any solution works here.

```js
const error = await workflow.execute(connection, credentials).catch(error => error);
expect(error)...
```

test/unit/cmap/auth/mongodb_oidc/gcp_machine_workflow.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    const workflow = new GCPMachineWorkflow();
+
+    context('when TOKEN_RESOURCE is not set', function () {
+      const connection = sinon.createStubInstance(Connection);
+      const credentials = sinon.createStubInstance(MongoCredentials);
+
+      it('throws an error', async function () {
+        try {
+          await workflow.execute(connection, credentials);
+          expect.fail('workflow must fail without TOKEN_RESOURCE');
```
nbbeeken, line 17: same issue with expect fail containing the string we want to see come out of execute

test/unit/cmap/auth/mongodb_oidc/token_machine_workflow.test.ts
```diff
@@ -0,0 +1,34 @@
...
+      });
+
+      after(function () {
+        process.env.OIDC_TOKEN_FILE = file;
+      });
+
+      it('throws an error', async function () {
+        try {
+          await workflow.execute(connection, credentials);
+          expect.fail('workflow must fail without OIDC_TOKEN_FILE');
```
nbbeeken, line 27: string includes OIDC_TOKEN_FILE

test/unit/cmap/auth/mongodb_oidc/token_machine_workflow.test.ts
```diff
@@ -0,0 +1,34 @@
...
+      const connection = sinon.createStubInstance(Connection);
+      const credentials = sinon.createStubInstance(MongoCredentials);
+
+      before(function () {
+        file = process.env.OIDC_TOKEN_FILE;
+        delete process.env.OIDC_TOKEN_FILE;
+      });
+
+      after(function () {
+        process.env.OIDC_TOKEN_FILE = file;
```
nbbeeken, line 21: You cannot unconditionally assign back to process.env without potentially setting this to a non-empty string which can have side-effects on other tests. if OIDC_TOKEN_FILE is `undefined` then `process.env.OIDC_TOKEN_FILE` will be set to `"undefined"` the string.

src/index.ts
```diff
@@ -99,6 +100,7 @@ export {
 export { BatchType } from './bulk/common';
 export { AutoEncryptionLoggerLevel } from './client-side-encryption/auto_encrypter';
 export { GSSAPICanonicalizationValue } from './cmap/auth/gssapi';
+export { TokenCache, TokenEntry } from './cmap/auth/mongodb_oidc/token_cache';
```
nbbeeken, line 103: Since these are internal you can just export them as `type` we do below line 161 to make api-extractor happy. They will still be runtime accessible in test files.

test/unit/index.test.ts
```diff
@@ -124,6 +125,7 @@ const EXPECTED_EXPORTS = [
   'ServerType',
   'SrvPollingEvent',
   'Timestamp',
+  'TokenCache',
```
nbbeeken, line 128: This should be removed when you move this to a type only export.

nbbeeken (changes requested):

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Get the token from the environment.
+   */
+  async getToken(credentials?: MongoCredentials): Promise<AccessToken> {
+    const tokenAudience = credentials?.mechanismProperties.TOKEN_RESOURCE;
+    const username = credentials?.username;
+    if (!tokenAudience) {
+      throw new MongoAzureError(TOKEN_RESOURCE_MISSING_ERROR);
+    }
+    const response = await getAzureTokenData(tokenAudience, username);
```
baileympearson, line 35: (this applies to all mechanisms)

From the credential caching section of the spec:

> Drivers MUST ensure that only one call to the configured provider or OIDC callback can happen at a time. 

I don't see a mechanism for preventing concurrent requests to the provider.  Would caching the promise work?  We could cache the promise in the abstract class:

```typescript
class MachineWorkflow {

    private promise = null;
	getToken() {
      this.promise ??= this._getToken();
	  try {
        return await this.promise;
	  } finally {
		this.promise = null;
	  }
	}

	abstract _getToken() {} // subclasses actually implement this method, the parent class caches the promise to ensure only one request at a time.
}
```

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
```diff
@@ -0,0 +1,70 @@
...
+      throw new MongoAzureError(ENDPOINT_RESULT_ERROR);
+    }
+    return response;
+  }
+}
+
+/**
+ * Hit the Azure endpoint to get the token data.
+ */
+async function getAzureTokenData(tokenAudience: string, username?: string): Promise<AccessToken> {
```
baileympearson, line 46: (applies to this and GCP)

The spec says:

> If the response does not return a status code of 200, the driver MUST raise an error including the HTTP response body.

`request` doesn't return the status code or reject when the status is non-200.  If we use the `get` helper defined in azure.ts, we can access both the contents of the body and the status of the request.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -156,11 +96,10 @@ export class CallbackWorkflow implements Workflow {
   private async startAuthentication(
```
baileympearson, line 96: vscode says this is dead code.  Can we remove this?

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -193,79 +132,28 @@ export class CallbackWorkflow implements Workflow {
    * Fetches an access token using either the request or refresh callbacks and
    * puts it in the cache.
    */
-  private async fetchAccessToken(
+  protected async fetchAccessToken(
     connection: Connection,
     credentials: MongoCredentials,
```
baileympearson, line 137:
```suggestion
    _connection: Connection,
    _credentials: MongoCredentials,
```

I'm surprised lint didn't complain about these variables being unused

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -193,79 +132,28 @@ export class CallbackWorkflow implements Workflow {
...
-        result = await requestCallback(serverInfo, context);
-      }
-    } else {
-      // With no token in the cache we use the request callback.
-      result = await requestCallback(serverInfo, context);
+    callback: OIDCCallbackFunction,
+    idpInfo?: IdPInfo
+  ): Promise<TokenEntry> {
+    const params: OIDCCallbackParams = {
+      timeoutContext: AbortSignal.timeout(TIMEOUT_MS),
```
baileympearson, line 142: 5 minutes is only the default timeout for the human workflow.  The spec defines the timeout for machine callbacks as:

> The timeout value MUST be min(remaining connectTimeoutMS, remaining timeoutMS) as described in the Server Selection section of the CSOT spec.

But this isn't really helpful, since we haven't implemented CSOT.  Any idea what the correct timeout would be here?

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -193,79 +132,28 @@ export class CallbackWorkflow implements Workflow {
...
+      version: OIDC_VERSION
+    };
+    if (idpInfo) {
+      params.idpInfo = idpInfo;
     }
+    // With no token in the cache we use the request callback.
+    const result = await callback(params);
```
baileympearson, line 149: I'm surprised the spec has the callbacks handle timeouts.  If the callback doesn't handle the timeout, then it's possible that the driver hangs during auth?  

We can provide a timeout to the callback, but it might also be prudent to handle the timeout ourselves, so we don't hang on this line if the callback never resolves.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -278,19 +166,3 @@ function isCallbackResultInvalid(tokenResult: unknown): boolean {
   if (!('accessToken' in tokenResult)) return true;
   return !Object.getOwnPropertyNames(tokenResult).every(prop => RESULT_PROPERTIES.includes(prop));
```
baileympearson, line 167:
```suggestion
  return !Object.keys(tokenResult).every(prop => RESULT_PROPERTIES.includes(prop));
```

It's probably stricter than necessary to enforce that the callback returns an object with own properties.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -193,79 +132,28 @@ export class CallbackWorkflow implements Workflow {
...
- * saslStart or saslContinue depending on the presence of a conversation id.
- */
-function finishCommandDocument(token: string, conversationId?: number): Document {
-  if (conversationId != null && typeof conversationId === 'number') {
-    return {
-      saslContinue: 1,
-      conversationId: conversationId,
-      payload: new Binary(BSON.serialize({ jwt: token }))
-    };
+    return { idpServerResponse: result, idpInfo: idpInfo };
```
baileympearson, line 155: Is it necessary to return the idpInfo, if the caller is responsible for providing it in the first place?

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
```diff
@@ -0,0 +1,61 @@
...
+    cache?: TokenCache
+  ): Promise<Document> {
+    const callback = getCallback(credentials.mechanismProperties);
+    // If there is a cached access token, try to authenticate with it. If
+    // authentication fails with an Authentication error (18),
+    // invalidate the access token, fetch a new access token, and try
+    // to authenticate again.
+    // If the server fails for any other reason, do not clear the cache.
+    let tokenEntry: TokenEntry;
+    if (cache?.hasToken()) {
```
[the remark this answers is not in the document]
baileympearson: I don't feel strongly, but this is a direct port of the algorithm defined in the spec: https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#two-step

Cross-referencing an implementation with the spec is easier if they're line-for-line the same.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -83,20 +93,25 @@ export interface Workflow {
...
+OIDC_WORKFLOWS.set('test', new TokenMachineWorkflow());
+OIDC_WORKFLOWS.set('azure', new AzureMachineWorkflow());
+OIDC_WORKFLOWS.set('gcp', new GCPMachineWorkflow());
 
 /**
  * OIDC auth provider.
  * @experimental
  */
 export class MongoDBOIDC extends AuthProvider {
+  cache?: TokenCache;
```
baileympearson, line 107: Are there any scenarios where `cache` will not be provided?  If there aren't, it might be nicer to make it required.  This propagates through the workflows too, because if the cache is always present on the provider, it will always be provided to workflow.execute() or workflow.reauthenticate().

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -55,97 +42,50 @@ export class CallbackWorkflow implements Workflow {
...
+   * - Check if the connection's accessToken is not equal to the token manager's.
+   *   - If they are different, use the token from the manager and set it on the connection and finish auth.
+   *     - On success return, on error continue.
+   * - start auth to update the IDP information
+   *   - If the idp info has changed, clear access token and refresh token.
+   *   - If the idp info has not changed, attempt to use the refresh token.
+   * - if there's still a refresh token at this point, attempt to finish auth with that.
+   * - Attempt the full auth run, on error, raise to user.
+   */
+  async reauthenticate(
```
baileympearson, line 57: The doc comment here doesn't seem to match the implementation of this method

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -55,97 +42,50 @@ export class CallbackWorkflow implements Workflow {
     return { speculativeAuthenticate: document };
   }
 
+  /**
+   * Reauthenticate the callback workflow.
+   * For reauthentication:
+   * - Check if the connection's accessToken is not equal to the token manager's.
```
baileympearson, line 48: from the spec:

> Additionally, drivers MUST cache the access token used to authenticate a connection on the connection object (henceforth referred to as the Connection Cache).

It's not immediately clear how we're satisfying this requirement.  Do we cache the token on the connection anywhere?

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -193,79 +132,28 @@ export class CallbackWorkflow implements Workflow {
...
+      timeoutContext: AbortSignal.timeout(TIMEOUT_MS),
+      version: OIDC_VERSION
+    };
+    if (idpInfo) {
+      params.idpInfo = idpInfo;
     }
+    // With no token in the cache we use the request callback.
+    const result = await callback(params);
     // Validate that the result returned by the callback is acceptable. If it is not
     // we must clear the token result from the cache.
```
baileympearson, line 151: from the spec:

> Call the configured built-in provider integration or the OIDC callback to retrieve a new access token. Wait until it has been at least 100ms since the last callback invocation, to avoid overloading the callback.

Seems like we need to add some rate limiting to ensure we don't overload the callback.  Also, I'd like clarification that we only rate limit the callback workflow and not the machine workflow too.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -36,32 +39,30 @@ export interface IdPServerResponse {
...
 /**
  * @public
  * @experimental
  */
-export type OIDCRequestFunction = (
-  info: IdPServerInfo,
-  context: OIDCCallbackContext
-) => Promise<IdPServerResponse>;
+export interface OIDCCallbackParams {
+  timeoutContext: AbortSignal;
```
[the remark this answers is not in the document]
baileympearson: > is it possible that someone could want to make a DB call on another MongoClient

Highly unlikely.  This callback is intended to communicate with an IDP which is unrelated to the driver.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -71,10 +72,19 @@ export interface Workflow {
...
-    reauthenticating: boolean,
+    cache?: TokenCache,
     response?: Document
   ): Promise<Document>;
 
+  /**
+   * Each workflow should specify the correct custom behaviour for reauthentication.
+   */
+  reauthenticate(
+    connection: Connection,
```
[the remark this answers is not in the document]
baileympearson: No.

Authentication and reauthentication both happen after the initial handshake (with the exception of speculative auth, but we don't send any commands through Connection.command() when preparing a handshake document for speculative authentication.

auto encryption is irrelevant to authentication because user data isn't being stored (what data would FLE be encrypting for a saslStart or saslContinue?) and [authentication commands are bypass encryption](https://github.com/mongodb/libmongocrypt/blob/ba2443f1c940f46de41fededf9217ca1da0a19a9/src/mongocrypt-ctx-encrypt.c#L2780).

baileympearson (changes requested):

src/cmap/auth/mongodb_oidc.ts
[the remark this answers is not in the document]
nbbeeken: Oh, excellent thank you for digging up that line of code.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
blink1073: It applies to both

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -16,7 +19,7 @@ const MISSING_CREDENTIALS_ERROR = 'AuthContext must provide credentials.';
  * @public
  * @experimental
```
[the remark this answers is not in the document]
blink1073: Yes, these interfaces are GA.

src/cmap/auth/mongodb_oidc/command_builders.ts
```diff
@@ -0,0 +1,43 @@
...
+  };
+}
+
+/**
+ * Generate the saslStart command document.
+ */
+export function startCommandDocument(credentials: MongoCredentials): Document {
+  const payload: Document = {};
+  if (credentials.username) {
+    payload.n = credentials.username;
```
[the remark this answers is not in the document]
blink1073: By spec, yes.

src/cmap/auth/mongo_credentials.ts
```diff
@@ -57,15 +62,17 @@ export interface AuthMechanismProperties extends Document {
...
   AWS_SESSION_TOKEN?: string;
   /** @experimental */
-  REQUEST_TOKEN_CALLBACK?: OIDCRequestFunction;
+  OIDC_CALLBACK?: OIDCCallbackFunction;
   /** @experimental */
-  REFRESH_TOKEN_CALLBACK?: OIDCRefreshFunction;
+  OIDC_HUMAN_CALLBACK?: OIDCCallbackFunction;
   /** @experimental */
-  PROVIDER_NAME?: 'aws' | 'azure';
+  ENVIRONMENT?: 'test' | 'azure' | 'gcp';
```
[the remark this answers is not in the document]
blink1073: I think it is fine as long as the user-facing docs mention that "azure" and "gcp" are the only valid options.

test/spec/auth/unified/oidc-auth-without-retry.yml
```diff
@@ -1,47 +1,43 @@
...
+  authMechanism: "MONGODB-OIDC"
 createEntities:
+- client:
+    id: authClient
 - client:
     id: client0
     uriOptions:
-      retryReads: false
-      retryWrites: false
+      authMechanism: "MONGODB-OIDC"
```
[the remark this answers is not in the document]
blink1073: Yes, we consolidated the files down to one.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -117,17 +81,18 @@ describe('MONGODB-OIDC', function () {
 
       describe('1.2 Single Principal Explicit Username', function () {
         before(function () {
-          client = new MongoClient('mongodb://test_user1@localhost/?authMechanism=MONGODB-OIDC', {
+          // Create a client with ``MONGODB_URI_SINGLE``, a username of ``test_user1``, and the OIDC request callback.
+          const url = new URL(process.env.MONGODB_URI_SINGLE);
```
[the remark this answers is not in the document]
blink1073: The username should have the domain appended for human callbacks.  It is not included in `MONGODB_URI_SINGLE`.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
blink1073: The timeout given to the callback is meant as a hint, the driver is responsible for enforcing timeouts, if possible.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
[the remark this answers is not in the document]
nbbeeken: Sure but that's a pedantic degree of code formatting that may not last. I agree with you on the point of cross-referencing and don't mind leaving it this way. However, if it has no functional purpose and someone is driving by here, their instinct may be to collapse this to a `const` because TS and eslint encourage it and so engineers gravitate to it, and would they be wrong to? If you, I, or Durran are not there to review it then there's nothing stopping it from changing. 

So, perhaps, if we wanna keep it looking like the pseudo-code, let us mention it in a comment or "normalize" it to ward off those kinds of known drive-bys

src/cmap/auth/mongodb_oidc.ts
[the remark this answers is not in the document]
addaleax: > What is the expectation around mishandling timeouts? what if the promise never resolves should we mitigate that risk in the driver?

I'd place this firmly in the "not the driver's problem" section.

> Not all things out there support signals or ever will support signals. For example is it possible that someone could want to make a DB call on another MongoClient?

I think the real argument you're making here is that the Node.js driver should support `AbortSignal` in the same way as applying CSOT 🙂 

I'd stick with a simple API and idiomatic API here (which this is) and not worry about these more complex use cases until there is a genuine demand for them.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -36,32 +39,30 @@ export interface IdPServerResponse {
...
  */
-export type OIDCRequestFunction = (
-  info: IdPServerInfo,
-  context: OIDCCallbackContext
-) => Promise<IdPServerResponse>;
+export interface OIDCCallbackParams {
+  timeoutContext: AbortSignal;
+  version: number;
```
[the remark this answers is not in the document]
addaleax: > Should this be more aggressively informative: `version: 1;` meaning if we change it to `version: 2;` we can "break" TS on purpose for those have have opted into using this?

Yes.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -71,32 +74,42 @@ export interface Workflow {
...
+  workflow: Workflow;
+
   /**
    * Instantiate the auth provider.
    */
-  constructor() {
+  constructor(workflow?: Workflow) {
     super();
+    if (!workflow) {
+      throw new MongoDriverError('');
```
addaleax, line 110: TODO?

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
```diff
@@ -0,0 +1,100 @@
...
+import {
+  OIDC_VERSION,
+  type OIDCCallbackFunction,
+  type OIDCCallbackParams,
+  type OIDCResponse
+} from '../mongodb_oidc';
+import { AUTOMATED_TIMEOUT_MS, CallbackWorkflow } from './callback_workflow';
+import { type TokenCache } from './token_cache';
+
+/** Must wait at least 100ms between invokations */
```
addaleax, line 15:
```suggestion
/** Must wait at least 100ms between invocations */
```

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
```diff
@@ -0,0 +1,100 @@
...
+const CALLBACK_DELAY = 100;
+
+/**
+ * Class implementing behaviour for the non human callback workflow.
+ * @internal
+ */
+export class AutomatedCallbackWorkflow extends CallbackWorkflow {
+  private lastInvokationTime: number;
```
addaleax, line 23: Similarly here, s/lastInvokationTime/lastInvocationTime/ (unless this is intentional? it feels like it might be 😅)

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
```diff
@@ -0,0 +1,100 @@
...
+  OIDC_VERSION,
+  type OIDCCallbackFunction,
+  type OIDCCallbackParams,
+  type OIDCResponse
+} from '../mongodb_oidc';
+import { AUTOMATED_TIMEOUT_MS, CallbackWorkflow } from './callback_workflow';
+import { type TokenCache } from './token_cache';
+
+/** Must wait at least 100ms between invokations */
+const CALLBACK_DELAY = 100;
```
addaleax, line 16:
```suggestion
const CALLBACK_DEBOUNCE_MS = 100;
```

Very optional suggestion, I think `debounce` is a bit more unambiguous in terms of what exactly it refers to (and I'm just personally a big fan of adding units to my JS time variables)

src/cmap/auth/mongodb_oidc/callback_workflow.ts
addaleax: @baileympearson I think there might be a small misunderstanding here – `Object.keys()` also only returns the names of own properties, but excludes non-enumerable ones

(And I have no strong feelings about _that_ distinction, tbh. A callback should typically just return a good-old regular JS object most of the time anyway.)

src/cmap/auth/mongodb_oidc.ts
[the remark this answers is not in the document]
addaleax: > the driver is responsible for enforcing timeouts, if possible.

Well, the driver can choose to ignore the result of the callback and treat it as failed if it detects the timeout expiring, but it _can't_ actually enforce anything, right?

> I also think CSOT makes the driver responsible for enforcing it 

Well, the spec says "This is to account for the human interaction required to complete the callback. In this case, the callback is not subject to CSOT".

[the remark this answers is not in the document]
blink1073: If you're using the same interface for machine and human callbacks, you'll need to make a distinction based on the auth mechanism parameter (`OIDC_CALLBACK` vs `OIDC_HUMAN_CALLBACK`).  For the machine callback, it *is* subject to CSOT if available, and it *is* expected to be cancelled if possible in the driver's language.   For Python we currently cannot cancel, only check the remaining time after any given operation.

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -83,20 +93,25 @@ export interface Workflow {
...
  * @experimental
  */
 export class MongoDBOIDC extends AuthProvider {
+  cache?: TokenCache;
+
   /**
    * Instantiate the auth provider.
    */
-  constructor() {
+  constructor(cache?: TokenCache) {
```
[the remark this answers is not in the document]
durran: I've actually refactored this to make the workflows have state now, and the `TokenCache` is now provided to their constructors and is always present.

test/integration/auth/mongodb_oidc_azure.prose.test.ts
```diff
@@ -28,181 +20,68 @@ describe('OIDC Auth Spec Prose Tests', function () {
...
       beforeEach(function () {
-        client = this.configuration.newClient(process.env.MONGODB_URI, {
-          authMechanismProperties: {
-            ALLOWED_HOSTS: []
-          }
-        });
+        const options: MongoClientOptions = {};
+        // if (process.env.AZUREOIDC_USERNAME) {
+        //   options.auth = { username: process.env.AZUREOIDC_USERNAME, password: undefined };
+        // }
```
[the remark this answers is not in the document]
durran: Removed.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -1214,17 +817,20 @@ describe('MONGODB-OIDC', function () {
...
             expect.fail('Reauthentication must fail on the saslStart error');
           } catch (error) {
             // This is the saslStart failCommand bubbled up.
             expect(error).to.be.instanceOf(MongoServerError);
+            expect(requestSpy).to.have.been.calledTwice;
           }
         });
       });
     });
+    // describe('6. Separate Connections Avoid Extra Callback Calls', function () {});
```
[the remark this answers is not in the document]
durran: This manual test has been completely removed now as the prose tests are in the integration folder.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -9,66 +9,41 @@ import {
...
+  type OIDCResponse
 } from '../mongodb';
-import { sleep } from '../tools/utils';
 
-describe('MONGODB-OIDC', function () {
+describe('OIDC Auth Spec Prose Tests', function () {
   context('when running in the environment', function () {
     it('contains AWS_WEB_IDENTITY_TOKEN_FILE', function () {
       expect(process.env).to.have.property('AWS_WEB_IDENTITY_TOKEN_FILE');
     });
```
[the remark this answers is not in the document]
durran: This is no longer used, and is removed.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -9,66 +9,41 @@ import {
...
   });
 
-  describe('OIDC Auth Spec Prose Tests', function () {
-    // Set up the cache variable.
-    const cache = OIDC_WORKFLOWS.get('callback').cache;
-    const callbackCache = OIDC_WORKFLOWS.get('callback').callbackCache;
+  describe('1. Callback Authentication', function () {
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -78,36 +53,25 @@ describe('MONGODB-OIDC', function () {
...
-          client = new MongoClient('mongodb://localhost/?authMechanism=MONGODB-OIDC', {
+          // Create the default OIDC client.
+          client = new MongoClient(`${process.env.MONGODB_URI_SINGLE}?authMechanism=MONGODB-OIDC`, {
             authMechanismProperties: {
-              REQUEST_TOKEN_CALLBACK: createRequestCallback()
+              OIDC_CALLBACK: createRequestCallback()
             }
           });
-          collection = client.db('test').collection('test');
+          collection = client.db('test').collection('nodeOidcTest');
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -9,66 +9,41 @@ import {
...
-    ) => {
-      return async (info: IdPServerInfo, context: OIDCCallbackContext) => {
-        const token = await readFile(path.join(process.env.OIDC_TOKEN_DIR, username), {
-          encoding: 'utf8'
-        });
-        // Do some basic property assertions.
-        expect(context).to.have.property('timeoutSeconds');
-        expect(info).to.have.property('issuer');
-        expect(info).to.have.property('clientId');
+        expect(params).to.have.property('timeoutContext');
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -255,7 +208,7 @@ describe('MONGODB-OIDC', function () {
             //   {
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -281,10 +234,10 @@ describe('MONGODB-OIDC', function () {
             client = new MongoClient('mongodb://evilmongodb.com/?authMechanism=MONGODB-OIDC', {
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -469,55 +355,42 @@ describe('MONGODB-OIDC', function () {
       });
 
       describe('3.1 Valid Callbacks', function () {
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -920,53 +479,41 @@ describe('MONGODB-OIDC', function () {
...
+        //     ],
+        //     "errorCode": 18
+        //   }
+        // }
+        //
+        // Note
+        //
+        // The driver MUST either use a unique appName or explicitly remove the failCommand after the test to prevent leakage.
+        //
         await setupFailPoint();
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -1028,26 +573,25 @@ describe('MONGODB-OIDC', function () {
...
+            monitorCommands: true
           });
-          await client.db('test').collection('test').findOne();
-          expect(refreshSpy).to.not.be.called;
-          client.close();
+          // Perform a find operation that succeeds.
+          // Assert that the request callback has been called once.
+          // Clear the listener state if possible.
+          await client.db('test').collection('nodeOidcTest').findOne();
+          expect(requestSpy).to.have.been.calledOnce;
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -1068,20 +612,16 @@ describe('MONGODB-OIDC', function () {
...
-          addListeners();
           await setupFailPoint();
-          await client.db('test').collection('test').findOne();
-          expect(refreshSpy).to.have.been.calledOnce;
+          addListeners();
+          await client.db('test').collection('nodeOidcTest').findOne();
+          expect(requestSpy).to.have.been.calledTwice;
           expect(commandStartedEvents.map(event => event.commandName)).to.deep.equal([
             'find',
             'find'
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -1141,26 +745,27 @@ describe('MONGODB-OIDC', function () {
...
         //   }
         // }
         //
         // Perform a find operation that succeeds.
+        // Assert that the request callback has been called three times.
         // Close the client.
         it('successfully authenticates', async function () {
-          const result = await client.db('test').collection('test').findOne();
+          const result = await client.db('test').collection('nodeOidcTest').findOne();
+          expect(requestSpy).to.have.been.calledThrice;
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -1129,9 +671,71 @@ describe('MONGODB-OIDC', function () {
...
+        };
+        // Sets up the fail point for the find to reauthenticate.
+        const setupFailPoint = async () => {
+          return await client
+            .db()
+            .admin()
+            .command({
+              configureFailPoint: 'failCommand',
+              mode: {
+                times: 2
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

test/manual/mongodb_oidc.prose.test.ts
```diff
@@ -1141,26 +745,27 @@ describe('MONGODB-OIDC', function () {
...
         it('successfully authenticates', async function () {
-          const result = await client.db('test').collection('test').findOne();
+          const result = await client.db('test').collection('nodeOidcTest').findOne();
+          expect(requestSpy).to.have.been.calledThrice;
           expect(result).to.be.null;
         });
       });
 
-      describe('6.3 Retries and Fails with no Cache', function () {
+      describe('5.3 Fails', function () {
```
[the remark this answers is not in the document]
durran: Manual test is removed. Integration tests follow spec prose tests.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: This is now split to 5 minutes and 60 seconds.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: Updated.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: It's actually used now in the human callback.

src/cmap/auth/mongodb_oidc.ts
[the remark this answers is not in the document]
nbbeeken: Talked about this a bit in a team meeting: Either we leave the responsibility of interruption entirely to the callback or we set up a Promise.race which will let the driver regain control but it won't stop the callback from pending.

In the former case, we can generally leave things be, pass the signal but we don't use it ourselves. If it hangs so be it, and 

In the latter case, I am not sure about the importance of the details of the outcome there but the user would be left with a callback that is going to, ideally, eventually fulfill but the value/error will not be handled by the driver. 

CSOT (relevant context: https://jira.mongodb.org/browse/DRIVERS-2347) should work somewhat like this, CSOT will be raced with processing waitQueueMembers for connection checkOut. If CSOT expires and the pool was creating a connection to service an operation, that creation is not interrupted but the operation does reject, the otherwise healthy connection should be put into the connection pool.  

The AbortSignal is starting to make more sense to me if our goal is to provide a mechanism that _both_ returns control to the driver and notifies the user's callback of early termination. I think it will be an important detail that the user's abort handling runs before the driver's (so either the driver ensures its listener runs last or we chain signals to make sure our abort comes after theirs).

Providing a plain timeout ms value _may_ also still work, as long as we call setTimeout after we've started the user's callback, if we both use the same value for ms and we call setTimeout first the driver timeout will win. 

> but it can't actually enforce anything, right?

Yes, I do not believe it is possible to agnostically preempt the callback. Unless we used some kind of thread solution but that is well out of scope here.

test/spec/auth/unified/oidc-auth-without-retry.yml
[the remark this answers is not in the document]
durran: This has been synced.

test/spec/auth/legacy/connection-string.yml
```diff
@@ -350,107 +350,43 @@ tests:
     mechanism: MONGODB-AWS
```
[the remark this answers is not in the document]
durran: Synced.

test/unit/cmap/auth/mongodb_oidc/azure_machine_workflow.test.ts
durran: I've updated to the suggestion.

test/unit/cmap/auth/mongodb_oidc/gcp_machine_workflow.test.ts
durran: Updated.

test/unit/cmap/auth/mongodb_oidc/token_machine_workflow.test.ts
durran: Updated.

test/unit/cmap/auth/mongodb_oidc/token_machine_workflow.test.ts
durran: Nice catch, I've added a check if `file` exists before setting it back.

src/index.ts
durran: Changed to type export.

test/unit/index.test.ts
durran: Removed

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: I've refactored this and now the result is only returned.

src/cmap/auth/mongodb_oidc.ts
durran: This was also refactored to always be present and is now a property of the workflows themselves, and no longer needs to be passed though.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: I've updated the comments for this method now.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: As discussed the use of a connection cache seemed redundant as the client cache is the source of truth for all connections. It made more sense here to always use the client cache.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
durran: Updated to suggestion.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
durran: I think I was thinking of this as something magical and german. :) I've changed.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
durran: Updated.

src/cmap/auth/mongodb_oidc.ts
durran: Updated to proper error now.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export class AutomatedCallbackWorkflow extends CallbackWorkflow {
+  /**
+   * Execute the OIDC callback workflow.
+   */
+  async execute(
+    connection: Connection,
+    credentials: MongoCredentials,
+    cache?: TokenCache
+  ): Promise<Document> {
```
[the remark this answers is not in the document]
durran: This has been refactored and now result is used.

.evergreen/run-oidc-prose-tests.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -o errexit  # Exit the script with error if any of the commands fail
+set -o xtrace   # Write all commands first to stderr
+
+ENVIRONMENT=${ENVIRONMENT:-"aws"}
+PROJECT_DIRECTORY=${PROJECT_DIRECTORY:-"."}
+source "${PROJECT_DIRECTORY}/.evergreen/init-node-and-npm-env.sh"
+
+printenv
```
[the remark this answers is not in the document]
durran: Yes was for debugging. Now removed.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (cache?.hasToken()) {
+      tokenEntry = cache.get();
+      try {
+        return await this.finishAuthentication(
+          connection,
+          credentials,
+          tokenEntry.idpServerResponse
+        );
+      } catch (error) {
+        if (error.code === 18) {
```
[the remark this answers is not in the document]
durran: Updated per suggestion and added to the error codes.

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
durran: This is now implemented with the `withLock` function.

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
durran: I've updated Azure to use this, still need to do GCP.

nbbeeken (changes requested):

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
nbbeeken: Spoke with @baileympearson I see we are intentionally making sure not to share callback results via a cached promise. He mentioned confirming that was the intention. 

IIUC maxConnecting already reduces the amount of possible parallelization (of course we're supposed to prevent it completely) is it the case that we want to ensure that the maxConnecting(default=2) connections _each_ get a unique ("unique" as in instance not necessarily as in value) result from the callback?

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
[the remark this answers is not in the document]
durran: Ah I missed that. I went now and refactored the `Workflow` interface to make `execute` and `reauthenticate` return `Promise<void>` and then made the changes down the object model, since they don't need to return a document at all.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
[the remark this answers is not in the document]
durran: Yes I made the change in the human callback not the automated one. I've updated there now as well.

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
durran: All machine workflows now no this.

src/cmap/auth/mongodb_oidc/token_cache.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { IdPInfo, OIDCResponse } from '../mongodb_oidc';
+
+class MongoOIDCError extends MongoDriverError {}
+
+/** @internal */
+export class TokenCache {
+  private accessToken?: string;
+  private refreshToken?: string;
+  private idpInfo?: IdPInfo;
+  private expiresInSeconds?: number;
```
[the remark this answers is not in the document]
durran: It is not used currently but noted in the spec to have it there for potential future use.

src/index.ts
```diff
@@ -99,6 +100,8 @@ export {
 export { BatchType } from './bulk/common';
 export { AutoEncryptionLoggerLevel } from './client-side-encryption/auto_encrypter';
 export { GSSAPICanonicalizationValue } from './cmap/auth/gssapi';
+export { Workflow } from './cmap/auth/mongodb_oidc';
+export type { TokenCache } from './cmap/auth/mongodb_oidc/token_cache';
```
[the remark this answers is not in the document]
durran: Moved.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -175,97 +97,49 @@ export class CallbackWorkflow implements Workflow {
...
-        result = await refreshCallback(serverInfo, context);
-      } else {
-        result = await requestCallback(serverInfo, context);
-      }
-    } else {
-      // With no token in the cache we use the request callback.
-      result = await requestCallback(serverInfo, context);
+  protected async executeAndValidateCallback(params: OIDCCallbackParams): Promise<OIDCResponse> {
+    if (!this.callback) {
+      throw new MongoDriverError('');
```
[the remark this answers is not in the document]
durran: Yes, I've removed this now.

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
```diff
@@ -0,0 +1,70 @@
...
+  url.searchParams.append('api-version', '2018-02-01');
+  url.searchParams.append('resource', tokenAudience);
+  if (username) {
+    url.searchParams.append('client_id', username);
+  }
+  const data = await request(url.toString(), {
+    json: true,
+    headers: AZURE_HEADERS
+  });
+  return data as AccessToken;
```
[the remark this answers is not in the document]
durran: It's validated in the `getToken` method.

src/cmap/auth/mongodb_oidc.ts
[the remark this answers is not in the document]
durran: I've updated this comment to explain what the version number is there to indicate. The property isn't meant to be breaking but users can throw if their version doesn't match.

src/cmap/auth/mongodb_oidc.ts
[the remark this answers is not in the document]
durran: I renamed to `callbackTimeoutMS` like the spec and pass that, updating to call `setTimeout` now.

nbbeeken (changes requested):

nbbeeken (commented):
Did another round of resolving, I think I got everything so what remains open still has questions to be answered, tag me on any that I may have missed or if it isn't clear what's pending there

src/cmap/auth/mongodb_oidc.ts
```diff
@@ -16,7 +19,7 @@ const MISSING_CREDENTIALS_ERROR = 'AuthContext must provide credentials.';
  * @public
```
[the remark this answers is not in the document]
durran: API docs have been updated for all properties.

src/cmap/auth/mongo_credentials.ts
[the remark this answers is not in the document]
durran: Noted in the API docs now that "test" is internal only.

test/integration/auth/mongodb_oidc_azure.prose.test.ts
```diff
@@ -28,181 +20,68 @@ describe('OIDC Auth Spec Prose Tests', function () {
...
       });
 
-      // Create a client with a url of the form mongodb://localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:azure,TOKEN_AUDIENCE:<foo>,
-      //   and an ALLOWED_HOSTS that is an empty list.
-      // Assert that a find operation succeeds.
-      // Close the client.
       it('successfully authenticates', async function () {
         const result = await collection.findOne();
-        expect(result).to.be.null;
+        expect(result).to.not.be.null;
```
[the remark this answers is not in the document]
durran: There's a doc inserted in the setup script.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
```diff
@@ -0,0 +1,96 @@
...
+          return await this.execute(connection, credentials);
+        } else {
+          throw error;
+        }
+      }
+    }
+    let response: OIDCResponse;
+    const now = Date.now();
+    // Ensure a delay between invokations to not overload the callback.
+    if (now - this.lastInvocationTime > CALLBACK_DEBOUNCE_MS) {
```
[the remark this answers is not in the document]
durran: I've moved this comment down now.

blink1073:
> @blink1073 Is there ever a divergence between connection level and client level caching?

I can't find where to respond.  Yes, if another connection on the same client has already re-authenticated, we see that the client cache is different (or newer) than the connection cache, and use that instead.

durran:
> > @blink1073 Is there ever a divergence between connection level and client level caching?
> 
> I can't find where to respond. Yes, if another connection on the same client has already re-authenticated, we see that the client cache is different (or newer) than the connection cache, and use that instead.

@blink1073 I'm just curious as to what the point of the connection cache is at all? If the client cache is always the source of truth for all connections, isn't simpler just to always use that and not copy tokens down to all the connections and need to check their values against the client cache?

blink1073:
> @blink1073 I'm just curious as to what the point of the connection cache is at all? If the client cache is always the source of truth for all connections, isn't it simpler just to always use that and not copy tokens down to all the connections and need to check their values against the client cache?

It allows us to figure out how to handle a 391 error, whether we actually need to call the callback again.  If we get our 391 after another connection, then it would have already updated the client cache, making it different/newer than the connection cache.

src/cmap/auth/mongodb_oidc.ts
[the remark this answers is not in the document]
durran: As discussed we've used the abort signal here with the same pattern we've done with CSOT using a Timeout.

src/index.ts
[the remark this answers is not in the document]
durran: I've pushed everything and they are definitely moved now.

src/cmap/auth/mongo_credentials.ts
```diff
@@ -179,45 +191,48 @@ export class MongoCredentials {
...
+        throw new MongoInvalidArgumentError(
+          `No password is allowed in ENVIRONMENT '${this.mechanismProperties.ENVIRONMENT}' for '${this.mechanism}'.`
+        );
       }
 
       if (
-        this.mechanismProperties.PROVIDER_NAME &&
-        !ALLOWED_PROVIDER_NAMES.includes(this.mechanismProperties.PROVIDER_NAME)
+        (this.mechanismProperties.ENVIRONMENT === 'azure' ||
+          this.mechanismProperties.ENVIRONMENT === 'gcp') &&
```
nbbeeken, line 212: It does not seem accurate that if this is 'gcp' we're going to throw an "Azure" error

test/tools/uri_spec_runner.ts
```diff
@@ -69,7 +75,9 @@ export function executeUriValidationTest(
       new MongoClient(test.uri);
       expect.fail(`Expected "${test.uri}" to be invalid${test.valid ? ' because of warning' : ''}`);
     } catch (err) {
-      if (err instanceof MongoRuntimeError) {
+      if (err instanceof MongoAzureError) {
+        // Azure URI errors don't have an underlying cause.
```
nbbeeken, line 79: I am a bit surprised this isn't just `MongoInvalidArgumentError`, even though it relates to Azure it's not Azure's error, it is our validation.

test/unit/index.test.ts
nbbeeken: We are still exporting these, we need to make sure we use "export type"

test/integration/client-side-encryption/client_side_encryption.prose.18.azure_kms_mock_server.test.ts
```diff
@@ -119,7 +119,7 @@ context('Azure KMS Mock Server Tests', function () {
     it('returns an error after the request times out', async () => {
       const error = await fetchAzureKMSToken(new KMSRequestOptions('slow')).catch(e => e);
 
-      expect(error).to.be.instanceof(MongoCryptAzureKMSRequestError);
+      expect(error).to.be.instanceof(MongoNetworkTimeoutError);
```
nbbeeken, line 122: This is an error that could reach users I believe. What's the reason for this changing? Shouldn't this remain the same API?

nbbeeken (changes requested):

test/unit/assorted/auth.spec.test.ts
```diff
@@ -1,13 +1,18 @@
 import { loadSpecTests } from '../../spec';
 import { executeUriValidationTest } from '../../tools/uri_spec_runner';
 
+const SKIP = 'should handle a complicated url-encoded TOKEN_RESOURCE (MONGODB-OIDC)';
```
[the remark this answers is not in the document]
durran: Unfortunately not yet, as a final decision hasn't been made yet of how it's going to be handled.

test/integration/client-side-encryption/client_side_encryption.prose.18.azure_kms_mock_server.test.ts
durran: I've fixed to throw the current error now.

test/tools/uri_spec_runner.ts
durran: Changed to `MongoInvalidArgumentError`.

test/integration/auth/mongodb_oidc.prose.test.ts
```diff
@@ -0,0 +1,1392 @@
...
+      });
+
+      describe('1.1 Callback is called during authentication', function () {
+        const callbackSpy = sinon.spy(createCallback('test_machine'));
+        // Create an OIDC configured client.
+        // Perform a find operation that succeeds.
+        // Assert that the callback was called 1 time.
+        // Close the client.
+        beforeEach(function () {
+          console.log(process.env);
```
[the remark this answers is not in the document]
durran: Removed.

test/unit/assorted/auth.spec.test.ts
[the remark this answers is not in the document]
durran: I've just added a TODO pointing at the NODE-6172 placeholder ticket.

nbbeeken (dismissed):

src/cmap/auth/mongodb_oidc/machine_workflow.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Ensure the callback is only executed one at a time.
+   */
+  private withLock(callback: OIDCTokenFunction) {
+    let lock: Promise<any> = Promise.resolve();
+    return async (credentials: MongoCredentials): Promise<AccessToken> => {
+      await lock;
+      // eslint-disable-next-line github/no-then
+      lock = lock.then(() => callback(credentials));
```
addaleax, line 92:
```suggestion
      lock = callback(credentials);
```

If we already `await` before this line anyway, we shouldn't need to chain the Promises here

src/cmap/auth/mongodb_oidc/machine_workflow.ts
```diff
@@ -0,0 +1,101 @@
...
+      const token = await this.callback(credentials);
+      this.cache.put({ accessToken: token.access_token, expiresInSeconds: token.expires_in });
+      return token.access_token;
+    }
+  }
+
+  /**
+   * Ensure the callback is only executed one at a time.
+   */
+  private withLock(callback: OIDCTokenFunction) {
```
addaleax, line 87:
```suggestion
  private withLock(callback: OIDCTokenFunction): OIDCTokenFunction {
```

src/cmap/auth/mongodb_oidc/callback_workflow.ts
```diff
@@ -175,97 +97,46 @@ export class CallbackWorkflow implements Workflow {
...
-function finishCommandDocument(token: string, conversationId?: number): Document {
-  if (conversationId != null && typeof conversationId === 'number') {
-    return {
-      saslContinue: 1,
-      conversationId: conversationId,
-      payload: new Binary(BSON.serialize({ jwt: token }))
+  /**
+   * Ensure the callback is only executed one at a time.
+   */
+  protected withLock(callback: OIDCCallbackFunction) {
```
addaleax, line 130:
```suggestion
  protected withLock(callback: OIDCCallbackFunction): OIDCCallbackFunction {
```

src/error.ts
```diff
@@ -557,6 +586,34 @@ export class MongoAzureError extends MongoRuntimeError {
...
 }
 
+/**
+ * A error generated when the user attempts to authenticate
+ * via GCP, but fails.
+ *
+ * @public
+ * @category Error
+ */
+export class MongoGCPError extends MongoRuntimeError {
```
addaleax, line 596: Should this be

```suggestion
export class MongoGCPError extends MongoOIDCError {
```

?

src/utils.ts
```diff
@@ -1157,6 +1158,38 @@ interface RequestOptions {
...
+      .on('error', error => {
+        clearTimeout(timeoutId);
+        reject(error);
+      })
+      .end();
+    timeoutId = setTimeout(() => {
+      request.destroy(new MongoNetworkTimeoutError(`request timed out after 10 seconds`));
+    }, 10000);
+  });
+}
```
addaleax, line 1191: Probably a dumb question, but this seems to duplicate a lot from the `request()` function below, should this be re-using it?

If not, I'd consider making the error message thrown here a bit more verbose and contain the URL, similar to `request()` below

addaleax (commented):
Looks great! Left a few comments but no need to wait for me before merging :)

baileympearson (changes requested):

src/utils.ts
blink1073: Note: we're using a different endpoint for GCP: `computeMetadata/v1/instance/service-accounts/default/identity` vs `computeMetadata/v1/instance/service-accounts/default/token`.

baileympearson: Our `get` and `request` are HTTP utilities, but Steve's comment is definitely relevant for https://github.com/mongodb/node-mongodb-native/pull/3912#discussion_r1606986443

src/cmap/auth/mongodb_oidc/machine_workflow.ts
durran: Updated.

src/cmap/auth/mongodb_oidc/callback_workflow.ts
durran: Updated.

src/error.ts
durran: Updated this as well as `MongoAzureError`.

src/cmap/auth/mongodb_oidc/automated_callback_workflow.ts
```diff
@@ -0,0 +1,112 @@
...
+      response = await this.fetchAccessToken(credentials);
+    } else {
+      // Ensure a delay between invokations to not overload the callback.
+      const responses = await Promise.all([
+        setTimeout(CALLBACK_DEBOUNCE_MS - (now - this.lastInvocationTime)),
+        this.fetchAccessToken(credentials)
+      ]);
+      response = responses[1];
+    }
+    this.lastInvocationTime = now;
```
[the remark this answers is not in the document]
durran: I've moved the throttling code down into `executeAndValidateCallback` which has the lock around it, so that the logic is handling properly now in order.

src/cmap/auth/mongodb_oidc/gcp_machine_workflow.ts
```diff
@@ -0,0 +1,53 @@
...
+    return await getGcpTokenData(tokenAudience);
+  }
+}
+
+/**
+ * Hit the GCP endpoint to get the token data.
+ */
+async function getGcpTokenData(tokenAudience: string): Promise<AccessToken> {
+  const url = new URL(GCP_BASE_URL);
+  url.searchParams.append('audience', tokenAudience);
```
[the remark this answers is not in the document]
durran: I think we should defer to a potential DRIVERS ticket here.

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
```diff
@@ -0,0 +1,82 @@
+import { MongoAzureError } from '../../../error';
+import { get } from '../../../utils';
+import type { MongoCredentials } from '../mongo_credentials';
+import { type AccessToken, MachineWorkflow } from './machine_workflow';
+import { type TokenCache } from './token_cache';
+
+/** Base URL for getting Azure tokens. */
+const AZURE_BASE_URL = 'http://169.254.169.254/metadata/identity/oauth2/token?';
```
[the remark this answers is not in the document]
durran: I've refactored the URI generating code to be shared between them now.

src/cmap/auth/mongodb_oidc/azure_machine_workflow.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Determines if a result returned from the endpoint is valid.
+ * This means the result is not nullish, contains the access_token required field
+ * and the expires_in required field.
+ */
+function isEndpointResultValid(
+  token: unknown
+): token is { access_token: unknown; expires_in: unknown } {
```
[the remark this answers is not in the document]
durran: We now assert they are string and number now..

baileympearson (dismissed):
Just waiting on one last CI, otherwise LGTM.  Nice!

baileympearson (approved):

PR #3914, opened 2023-11-08
fix(NODE-4863): do not use RetryableWriteError for non-server errors

alenakhineika:
### Description

Use the `MongoErrorLabel.PoolRequstedRetry` label instead of misused `MongoErrorLabel.RetryableWriteError` for non-server errors according to [spec](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst).

#### What is changing?

- The `PoolClearedError` constructor adds the new `MongoErrorLabel.PoolRequstedRetry` label instead of misused `MongoErrorLabel.RetryableWriteError`.
- The `isRetryableWriteError` function checks if PoolRequstedRetry or RetryableWriteError is set to determine if `retryOperation` is required.

None

[NODE-4863](https://jira.mongodb.org/browse/NODE-4863)

### DIFF
--- a/src/cmap/errors.ts
+++ b/src/cmap/errors.ts
@@ -56,7 +56,7 @@ export class PoolClearedError extends MongoNetworkError {
     super(errorMessage, pool.serverError ? { cause: pool.serverError } : undefined);
     this.address = pool.address;
 
-    this.addErrorLabel(MongoErrorLabel.RetryableWriteError);
+    this.addErrorLabel(MongoErrorLabel.PoolRequstedRetry);
   }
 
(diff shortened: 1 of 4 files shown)


durran (approved):
Just notes for other reviewers, the testing requirements made it seem to me like the prose test #2 should change somehow, but checking the spec this test matches so no change actually needed there.

The second testing requirement I'm not sure where that came from but I think the additional unit test added here suffices.

nbbeeken (approved):

W-A-James (approved):

PR #3916, opened 2023-11-09
fix(NODE-4996): swallow any errors in maybeCallback

baileympearson:
#### What is changing?

If the `callback` in `maybeCallback` throws, then `maybeCallback` throws an uncatchable promise rejection.  This PR adds a catch that swallows any errors thrown in the callback.

Long-term, we should refactor the remaining usages of `maybeCallback` to use async-await and this problem will go away.  This currently is only used by the callback overload of executeOperation, which is used in bulk write and sessions.  I thought it was too much work for a quick fix to refactor that code to async-await in this PR.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -347,10 +347,17 @@ export function maybeCallback<T>(
     return promise;
   }
 
-  promise.then(
-    result => callback(undefined, result),
-    error => callback(error)
-  );
+  promise
+    .then(
+      result => callback(undefined, result),
+      error => callback(error)
+    )
+    .catch(() => {
+      // in the case that `callback` throws, another promise rejection will be thrown.  we don't really have
+      // many options here - we can't call the callback again with the second error so we just swallow it.
+      // In the long-term, we should refactor the two remaining usages of `maybeCallback` to
+      // be async functions and we and remove this helper.
+    });
   return;
 }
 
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -347,10 +347,17 @@ export function maybeCallback<T>(
...
-  promise.then(
-    result => callback(undefined, result),
-    error => callback(error)
-  );
+  promise
+    .then(
+      result => callback(undefined, result),
+      error => callback(error)
+    )
+    .catch(() => {
```
durran, line 355: Can we at least log something here so the user at least knows we are swallowing the error?

baileympearson: I don't think we should add logs here, at least not with a `console.log`.  But I can file a ticket in the structured logging epic to add logging to places we may be suppressing errors?  There's a list of potential locations in the ticket.

durran: I guess I'd like to think what others think since it feels strange to me we'd swallow exceptions at this point (everything going through `executeOperation` and not notify in some form). I understand swallowing elsewhere without notifying (like `endSessions` in `close`), but `executeOperation` is a critical path for the driver operations.

nbbeeken: What is the error being observed that is causing the issue for your other work? 

I'm wondering if we wouldn't be better off correcting that "throw" to be properly sent to the callback, rather than introducing this more global silencing to all our maybeCallack usage.

addaleax, line 355: The typical result here would be to actually re-throw the exception outside of the current Promise chain, just like a callback would, e.g. inside a `process.nextTick(() => { throw exception; })`

nbbeeken:
https://jira.mongodb.org/browse/NODE-3332

`maybeCallback` is gone but I added a note about fixing the txn transition function to this ticket.

PR #3921, opened 2023-11-10
fix(NODE-5749): RTTPinger always sends legacy hello

nbbeeken:
#### What is changing?

- Adds serverApi and helloOk logic to RTTPinger
- Calls destory on connection when an error is recv-ed
- Replaced symbol properties with string properties

No

Fixes a connection leak when serverApi is enabled.

### Fix connection leak when serverApi is enabled

When enabling serverApi the driver's RTT mesurment logic (used to determine the closest node) still sent the legacy hello command "isMaster" causing the server to return an error. Unfortunately, the error handling logic did not correctly destroy the socket which would cause a leak.

Both sending the correct hello command and the error handling connection clean up logic are fixed in this change.


durran (approved):

test/integration/connection-monitoring-and-pooling/rtt_pinger.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    if (semver.gte('4.4.0', this.configuration.version)) {
+      if (this.currentTest)
+        this.currentTest.skipReason =
+          'Test requires streaming monitoring, needs to be on MongoDB 4.4+';
+      return this.skip();
+    }
+  });
+
+  context('when serverApi is enabled', () => {
```
dariakp, line 48: what about when it's not?

test/integration/connection-monitoring-and-pooling/rtt_pinger.test.ts
```diff
@@ -0,0 +1,109 @@
...
+      await sleep(11); // allow for another ping after spies have been made
+
+      expect(spies).to.have.lengthOf.at.least(1);
+      for (const spy of spies) {
+        expect(spy).to.have.been.calledWith(sinon.match.any, { hello: 1 }, sinon.match.any);
+      }
+    });
+  });
+
+  context('when rtt hello receives an error', () => {
```
dariakp, line 82: it might be good to make sure that this happens both for serverApi and non-serverApi enabled clients

dariakp (changes requested):

test/integration/connection-monitoring-and-pooling/rtt_pinger.test.ts
nbbeeken: Currently, `serverApi` does not relate to this behavior, and since it is not the only variable involved in how the hello is sent (ex. `conn.helloOk` and the connectionOptions that made the connection) I don't think this would be a variant that adds coverage. 

Does that logic make sense?

I use `.yeildsRight(new Error())` in an attempt to demonstrate "any error" will cause the connection to be destroyed. I can update the test title to better call that out.

dariakp: fair enough, although explicitly calling out a non-mongo error in the test case error string makes me wonder whether a mongo error would behave differently; "any error" might be less misleading about the intent in this case; also, what is the significance of the force destroy flag? what would be different if we didn't specify `force: true`?

nbbeeken: Agreed, fixed the message.

Using `force:false` would require us to pass in a callback since that makes the closure an async operation, we want to end the connection immediately because there isn't a reason to wait for FIN / graceful closure for a socket that's only purpose is RTT monitoring, we want to rebuild a new one ASAP.

test/integration/connection-monitoring-and-pooling/rtt_pinger.test.ts
nbbeeken: Agreed so the last case we're missing here is "serverApi disabled, server supports hello". added!

dariakp (approved):

PR #3930, opened 2023-11-17
refactor(NODE-5741): use async io helpers

nbbeeken:
#### What is changing?

Note reviewing per commit will assist greatly with following the changes made:

- test: latest lts avoiding npm install issue
  - npm latest needs 18.17 or later
- test: fix spec auth test to spy on correct connection class
  - the speculative auth tests spied on Connection.command, I adjusted it to spy on ModernConnection when they are enabled
- test: session usage back down to 1! woo!
  - I suspected this would happen! Session usage is back down to 1 for maxPoolSize 1. This has to do with callbacks having inconsistent asynchronicity vs promises being consistently ordered
- test: fix utf8 validation option tests
  - The utf8 validation tests spied on `onMessage` which no longer exists. They can just as easily assert the same thing if they spy on the method that parses the utf8 bson options
- test: set env variable for new connection tests
  - An environment variable controls passing `ModernConnection` into the MongoClient constructor in newClient, this may not be 100% coverage, but substantial enough to get us to the next PRs where we will switch over to one connection class entirely.
- refactor: re-introduce listeners for signal aborting with error
  - I mistakenly removed these, the listeners hook up to `controller.abort()`
- test: add new variants enabling ModernConnection
  - New build variant that defines the env variable that will make the driver test with `ModernConnection`
- refactor: connect error handling to signal
  - The Connection class now has a controller property set to an AbortController instance
  - The read and write helpers are given the `signal` from this controller
  - `timeout` and `close` events will abort the controller, which will interrupt the I/O operations if they are in progress
  - An AbortError will have its cause set to one of our MongoError subclasses which the operation handler will extract
- refactor: remove drain event listener
  - The `drain` event only emits if the stream has to buffer the data to be written, the way to determine if data has been buffered is by inspecting the return value of `.write`
  - Since we have to call write to determine if we need to wait for drain, I think it is equivalent for us to await the `.write`'s callback
- refactor: remove force flag and support noResponse
  - Added `noResponse` support, I decided to return ok:1 here, now that we are using promises our types are better and I do not think there is value in having our command method return something nullish
  - Removed the force flag for debugging, can re-introduce but consider:
    - it was easier to reason about the socket's state when a synchronous function marks it as ended the callbackified end function introduced an unnecessary step in finishing the cleanup process. 
- refactor: update cleanup method
  - Updated the cleanup method to simplify the new paradigm of only dealing with a socket and not also a MessageStream
- The following are self-explanatory helper refactors to organize code
  - refactor: remove message stream
  - refactor: remove onMessage
  - refactor: add sendCommand helper for send and recv
  - refactor: make command async
  - refactor: move command preparing to helper
  - refactor(NODE-5741): use async io helpers

No

Migrating I/O to async enables us to use signals. Singals will be an essntial part of CSOT.


src/utils.ts
```diff
@@ -1299,3 +1299,8 @@ export const COSMOS_DB_MSG =
 export function isHostMatch(match: RegExp, host?: string): boolean {
   return host && match.test(host.toLowerCase()) ? true : false;
 }
+
+export async function mayAbort(signal?: AbortSignal) {
+  if (signal == null) return new Promise(() => null); // never ending story
+  return new Promise((_, reject) => signal.addEventListener('abort', reject));
+}
```
baileympearson, line 1306: can this just be consolidated?
```suggestion
export async function mayAbort(signal?: AbortSignal) {
  return new Promise((_, reject) => signal?.addEventListener('abort', reject));
}
```

baileympearson: (also) - would you mind adding a nice doc comment?  I think good doc comments would on our utilities would be really nice for future devs

test/unit/cmap/modern_connection.test.ts
```diff
@@ -23,16 +23,17 @@ import {
...
+    queueMicrotask(cb);
   }
 }
 
 class MockModernConnection {
   socket = new MockSocket();
 }
 
-describe('writeCommand', () => {
+describe.skip('writeCommand', () => {
```
baileympearson, line 36: intentionally skipped?

.evergreen/config.in.yml
```diff
@@ -200,6 +200,7 @@ functions:
       type: test
       params:
         working_dir: "src"
+        add_expansions_to_env: true
```
baileympearson, line 203: just curious - how come this addition is necessary?

src/cmap/commands.ts
```diff
@@ -705,7 +705,10 @@ export class OpMsgResponse {
     this.parsed = true;
   }
 
-  parseBsonSerializationOptions({ enableUtf8Validation }: BSONSerializeOptions): {
+  parseBsonSerializationOptions(
+    this: void,
```
baileympearson, line 709: just curious - why is this necessary?

baileympearson (changes requested):
first pass review - didn't look at `Connection.ts` yet and a few other misc files.  Just looked over CI, utils and test changes

src/utils.ts
baileympearson: ((also)) should we test this utility?

test/unit/cmap/modern_connection.test.ts
nbbeeken: For now, I will revisit these, I can remove all the drain-related tests, but generally, now that we have integration coverage do we need unit tests?

src/cmap/commands.ts
nbbeeken: It is not, I will revert. I did it when changing the utf8 tests over to spy on this method and noticed the method isn't instance-dependent.

src/utils.ts
nbbeeken: Added tests and improved the function, we also needed a throw if the signal already aborted condition.

src/cmap/connection.ts
```diff
@@ -1171,26 +1009,130 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
+  ): Promise<Document> {
+    // Temporary to make sure no callback usage:
+    // eslint-disable-next-line prefer-rest-params
+    if (typeof arguments[arguments.length - 1] === 'function') {
+      throw new Error('no callbacks allowed!!');
+    }
+
+    const message = this.prepareCommand(ns.db, command, options);
+
+    let started = 0;
```
baileympearson, line 1090: why not just initialize started to `now()`?

src/cmap/connection.ts
```diff
@@ -1171,26 +1009,130 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
+    if (this.monitorCommands) {
+      started = now();
+      this.emit(
+        ModernConnection.COMMAND_STARTED,
+        new CommandStartedEvent(this as unknown as Connection, message)
+      );
+    }
+
+    const document = await this.sendCommand(message, options);
```
baileympearson, line 1099: What happens with command monitoring if `sendCommand` rejects?  I think we need a `catch` block that records a command failed event

src/cmap/connection.ts
```diff
@@ -1258,8 +1207,11 @@ export async function writeCommand(
           zlibCompressionLevel: options.zlibCompressionLevel ?? 0
         });
   const buffer = Buffer.concat(await finalCommand.toBin());
-  connection.socket.push(buffer);
-  await drained;
+  options.signal?.throwIfAborted();
```
baileympearson, line 1210: can we remove this now that we call `throwIfAborted` in `aborted`?

src/cmap/connection.ts
```diff
@@ -1208,11 +1150,17 @@ const kDefaultMaxBsonMessageSize = 1024 * 1024 * 16 * 4;
...
 export async function* readWireProtocolMessages(
-  connection: ModernConnection
+  connection: ModernConnection,
+  { signal }: { signal?: AbortSignal } = {}
 ): AsyncGenerator<Buffer> {
   const bufferPool = new BufferPool();
   const maxBsonMessageSize = connection.hello?.maxBsonMessageSize ?? kDefaultMaxBsonMessageSize;
-  for await (const [chunk] of on(connection.socket, 'data')) {
+  for await (const [chunk] of on(connection.socket, 'data', { signal })) {
+    if (connection.delayedTimeoutId) {
```
baileympearson, line 1159: can this live in `connection.command` instead?

src/cmap/connection.ts
```diff
@@ -1171,26 +1009,130 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
+    return message;
+  }
+
+  private async sendCommand(
+    message: WriteProtocolMessageType,
+    options: CommandOptions
+  ): Promise<Document> {
+    const { signal } = this.controller;
 
+    signal.throwIfAborted();
```
baileympearson, line 1035: i commented on another `throwIfAborted()`, but I'm surprised we need them in every helper?  My thought was that it should be sufficient to only check when we actually read/write to/from the socket

src/cmap/connection.ts
```diff
@@ -1171,26 +1009,130 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
+        agreedCompressor: this.description.compressor ?? 'none',
+        zlibCompressionLevel: this.description.zlibCompressionLevel,
+        signal
+      });
+
+      if (options.noResponse) return { ok: 1 };
+
+      response = await read(this, { signal });
+    } finally {
+      if (!signal.aborted) this.controller = new AbortController();
```
baileympearson, line 1055: How come we're instantiating a new controller after every successful command?  I was expecting that `this.controller` would be tied to the lifetime of the connection and that when it did abort, it would indicate that the connection was closed with an error

src/cmap/connection.ts
```diff
@@ -1078,50 +951,17 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
-
-    if (!this.socket.writableEnded) {
-      this.socket.end(() => {
-        this.socket.destroy();
-        completeCleanup();
-      });
-    } else {
-      completeCleanup();
-    }
+    this.socket.destroy();
```
baileympearson, line 959: question: why did we remove the option to gracefully close our sockets (`force: false`)?

src/cmap/connection.ts
```diff
@@ -1145,12 +985,10 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
         clusterTime = session.clusterTime;
       }
 
-      const err = applySession(session, cmd, options);
-      if (err) {
-        return callback(err);
-      }
+      const sessionError = applySession(session, cmd, options);
```
baileympearson, line 988: `applySession` is only called here - could we just make `applySession` throw instead?

baileympearson (changes requested):
Overall changes look good!  Connection layer looking much nicer 🔥

durran (changes requested):
Bailey's comments, some notes.

src/cmap/connection.ts
nbbeeken: In this specific case I could remove this and because of how we have the io helpers set up there will be no difference.

This might be a practices thing we want to discuss in tech chat. I believe the thinking is that you want to check for aborted ahead of any `await` for at least two reasons, 1. you may not be able to trust the API taking the signal to handle an already aborted signal, so you may block forever as a result. 2. You yeild to the queue of tasks despite having indication of an error already.

Kinda thinking/talking off the cuff here so I'll do some research and gather thoughts for chat.

src/cmap/connection.ts
```diff
@@ -892,15 +864,6 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
     this[kHello] = response;
   }
 
-  // Set the whether the message stream is for a monitoring connection.
-  set isMonitoringConnection(value: boolean) {
```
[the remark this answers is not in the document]
nbbeeken: Agreed, and also this connection implementation doesn't support streaming for monitoring yet. That is the next ticket. When it does we could consider reintroducting some skipping of hellos if we deem it necessary, but since we have the mode setting likely we won't.

src/cmap/connection.ts
baileympearson: ( side note - that pattern looks remarkably similar to scope-based resource management....)

src/cmap/connection.ts
baileympearson: curious about best practices.  I assumed that the best practice would be to just ferry the abort controller along to whatever async function we're calling and letting them handle it, unless there's a strong reason to manually check it ourselves

src/cmap/connection.ts
baileympearson: 🤦 Duh.  ofc we still need it in the regular connection class

src/cmap/connection.ts
baileympearson: I guess I'd say that if we want this to be "closer to the data event", then `delayedTimeoutId` should not be associated with the connection class?

but we can look at this later.

test/unit/cmap/modern_connection.test.ts
baileympearson: I think there's value in keeping the unit tests, even if just as documentation for how the `writeCommand` method is intended to work

nbbeeken: I can add documentation more thorough documentation to the function, tests can be hard to tease out whether the assertions are 'must always be x' or 'worked like x when written'. 
Currently, the tests bind these functions to the data event from the MockSocket but we could rewrite the functions to rely on some other method of obtaining chunks, and it wouldn't be wrong as long as the actual I/O continues to work.

src/cmap/connection.ts
nbbeeken: This does not appear to change much functionally since we would still be creating a new controller, the difference is it does not assign it back to the Connection instance. I agree we probably need something like this, but I would like to take the time to investigate and design that helper tailored to the CSOT functionality when we're working on that rather than in this PR.

src/cmap/connection.ts
nbbeeken: I am not sure I understand what is cleaner in this context, using let vs const is not something to sweat the difference on. Not calling a function vs calling a function is a free win I do not see it as a pre-optimization, there's no special design here just code. 

Benchmarking is on the horizon before we merge ModernConnection w/ Connection, so there is opportunity to get insight.

src/cmap/connection.ts
nbbeeken: https://openjsf.org/blog/2022/04/29/using-abortsignal-in-node-js/

This was illuminating on the subject of signals

> Within this function, you should immediately check to see if the signal has already been triggered and, if it has, immediately abort the operation.

And this makes sense: if you support signals, before doing any work like setting up listeners and kicking off tasks you want to bail out before performing all that labor.

src/cmap/connection.ts
nbbeeken: [openjsf.org/blog/2022/04/29/using-abortsignal-in-node-js](https://openjsf.org/blog/2022/04/29/using-abortsignal-in-node-js/)

The article I found seems to have the solution we may be looking for. Also documented here: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener. `addEventListener` takes a cleanup signal

nbbeeken: The added abortable helper ensures we clean up our listeners for write. For reads, see slack, I will discuss in a thread tomorrow.

test/unit/utils.test.ts
```diff
@@ -1165,4 +1170,119 @@ describe('driver utils', function () {
...
+        const error = await Promise.race([
+          aborted().abort.catch(error => error),
+          aborted(AbortSignal.timeout(100)).abort
+        ]).catch(error => error);
+        expect(error.name).to.equal('TimeoutError');
+      });
+    });
+  });
+
+  describe('abortable()', () => {
```
dariakp, line 1210: we're missing some cases here; based on the functionality we want, the following needs to be covered:
1) already rejected promise with already aborted signal
2) already resolved promise with already aborted signal
3) already rejected promise with not yet aborted signal
4) already resolved promise with not yet aborted signal
5) unresolved promise with an already aborted signal
6) eventually rejecting promise with not yet aborted signal
7) eventually resolving promise with not yet aborted signal
8) unresolved promise with eventually aborted signal

and for eventually rejecting/resolving I would add a timeout to give it a full run of the event loop

dariakp (changes requested):

src/cmap/connection.ts
```diff
@@ -1232,11 +1228,12 @@ export async function writeCommand(
...
           zlibCompressionLevel: options.zlibCompressionLevel ?? 0
         });
+
   const buffer = Buffer.concat(await finalCommand.toBin());
-  await Promise.race([
-    promisify(connection.socket.write.bind(connection.socket))(buffer),
-    aborted(options.signal)
-  ]);
+
+  const socketWriteFn = promisify(connection.socket.write.bind(connection.socket));
```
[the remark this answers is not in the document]
nbbeeken: The callback will be called immediately if there is no buffering or later when the buffer has been written. [Here](https://github.com/nodejs/node/blob/main/lib/internal/streams/writable.js#L696-L702) the callback is invoked a few lines after the drain event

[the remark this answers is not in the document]
baileympearson: Generally, I don't think we can rely on Nodejs source code as a source of truth since we need to support 3+ major node versions and source code is likely to change.  We should rely on documentation and/or confirmation with nodejs contributors.

The docs make me _think_ that you're correct?  Although I'm unclear on what exactly "data is finally written out" means, given that it explicitly uses the terminology in the line above "the entire data was flushed successfully to the kernel buffer".

> Sends data on the socket. The second parameter specifies the encoding in the case of a string. It defaults to UTF8 encoding.
> 
> Returns true if the entire data was flushed successfully to the kernel buffer. Returns false if all or part of the data was queued in user memory. ['drain'](https://nodejs.org/api/net.html#event-drain) will be emitted when the buffer is again free.
> 
> The optional callback parameter will be executed when the data is finally written out, which may not be immediately.

.evergreen/config.in.yml
nbbeeken: I have undone this, it makes sense to only change what we need.

.evergreen/run-tests.sh
```diff
@@ -55,6 +55,11 @@ fi
 npm install @mongodb-js/zstd
 npm install snappy
 
+if [[ "$OS" = "Windows_NT" ]]; then
+  # TODO: Temp fix for add_expansions_to_env usage that breaks Windows
```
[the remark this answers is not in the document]
nbbeeken: Removed this. We do not test FLE on windows and the add_expansions_to_env (now removed) inadvertenly enabled it

src/utils.ts
```diff
@@ -1299,3 +1299,47 @@ export const COSMOS_DB_MSG =
 export function isHostMatch(match: RegExp, host?: string): boolean {
   return host && match.test(host.toLowerCase()) ? true : false;
 }
+
+/**
+ * Takes a promise and races it with a promise wrapping the abort event of the optionally provided signal.
+ * If the signal aborts the returned promise will reject, otherwise the given promises resolved value will be returned.
+ * - The given promises rejection will proceed the aborted signal rejection.
```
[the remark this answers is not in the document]
nbbeeken: Reworked this, and linked to `Promise.race` because it is essentially a feature of that API, but we need to never re-order the array we pass in

test/unit/utils.test.ts
nbbeeken: Added the test cases, thank you for helping lay them out.

2 is a little questionable in the current state - "already resolved promise with already aborted signal". The current behavior I have is that the promise resolution wins, but should that be the case? I think if the signal has already been interrupted it unfortunately doesn't matter that the promise has resolved. Thoughts?

src/utils.ts
```diff
@@ -1302,9 +1302,11 @@ export function isHostMatch(match: RegExp, host?: string): boolean {
 
 /**
  * Takes a promise and races it with a promise wrapping the abort event of the optionally provided signal.
- * If the signal aborts the returned promise will reject, otherwise the given promises resolved value will be returned.
- * - The given promises rejection will proceed the aborted signal rejection.
- * If given an already rejected promise and an aborted signal, the promise returned will reject with the given promises rejection value
+ * The given promise is _always_ ordered before the signal's abort promise.
+ * When given an already rejected promise and an already aborted signal, the promises rejection takes precedence.
```
dariakp, line 1306:
```suggestion
 * When given an already rejected promise and an already aborted signal, the promise's rejection takes precedence.
```

test/unit/utils.test.ts
dariakp: The new test cases look good to me now, but would it still be worthwhile to have some additional assertions on cleanup like you did in your original ones?

dariakp (changes requested):

src/cmap/connection.ts
```diff
@@ -1171,26 +1010,151 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
-      callback(err);
+      await writeCommand(this, message, {
+        agreedCompressor: this.description.compressor ?? 'none',
+        zlibCompressionLevel: this.description.zlibCompressionLevel,
+        signal
+      });
+
+      signal.throwIfAborted();
+
+      if (options.noResponse) return { ok: 1 };
```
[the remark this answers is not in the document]
nbbeeken: Yea I agree, if we're at this point the write has completed and if we are doing a "noResponse" we should return, we sucessfully completed.

test/unit/utils.test.ts
```diff
@@ -1171,46 +1169,60 @@ describe('driver utils', function () {
...
-        const error = await aborted(controller.signal).abort.catch(error => error);
-        expect(error).to.be.instanceOf(MongoError);
-        expect(error.message).to.equal('my error');
+  describe('abortable()', () => {
+    const goodError = new Error('good error');
+    const badError = new Error('unexpected bad error!');
+    const expectedValue = "don't panic";
+
+    context('when not given a signal', () => {
+      it('returns promise fulfillment', async () => {
```
dariakp, line 1178:
```suggestion
      it('returns promise fulfillment if the promise resolves or rejects', async () => {
```

test/unit/utils.test.ts
```diff
@@ -1171,46 +1169,60 @@ describe('driver utils', function () {
...
+        expect(await abortable(Promise.resolve(expectedValue))).to.equal(expectedValue);
+        expect(await abortable(Promise.reject(goodError)).catch(e => e)).to.equal(goodError);
+      });
+
+      it('pends indefinitely', async () => {
```
dariakp, line 1183:
```suggestion
      it('pends indefinitely if the promise is never settled', async () => {
```

test/unit/utils.test.ts
```diff
@@ -1171,46 +1169,60 @@ describe('driver utils', function () {
...
     });
 
-    context('when given a signal that aborts after some time', () => {
-      it('rejects the returned promise after the timeout', async () => {
-        const controller = new AbortController();
-        setTimeout(() => controller.abort(new MongoError('my error')), 10);
-        const start = +new Date();
-        const error = await aborted(controller.signal).abort.catch(error => error);
-        const end = +new Date();
+    context('always removes abort listener it attaches', () => {
```
dariakp, line 1192:
```suggestion
    context('always removes the abort listener it attaches', () => {
```

dariakp (changes requested):

durran (approved):

PR #3934, opened 2023-11-28
refactor(NODE-5764): add commandName to AbstractOperation subclasses

aditi-khare-mongoDB:
### Description
Before this change, the command name of an operation was only able to be accessed at execution time. Now, with the use of a getter in each subclass of AbstractOperation, the name of a command is able to be accessed once the operation is instantiated, or with the operation class name.

#### What is changing?
 Allows access command name of operation once operation is instantiated, or through operation's class name.

No this is an internal change, invisible to users implicitly since its within a private class.

[NODE-4687 (Server Selection Logging)](https://jira.mongodb.org/browse/NODE-4687) was blocked before this change, since the AC for the ticket required command/operation name to be logged before execution of the command.


nbbeeken (changes requested):
We don't want to change the package-lock here right? We can undo the changes.

src/operations/run_command.ts
```diff
@@ -44,6 +48,10 @@ export class RunAdminCommandOperation<T = Document> extends AbstractOperation<T>
     this.ns = new MongoDBNamespace('admin', '$cmd');
   }
 
+  override get commandName() {
+    return 'runAdminCommand' as const;
```
nbbeeken, line 52:
```suggestion
    return 'runCommand' as const;
```

These are the two "pseudo" command names. runAdminCommand is not something that is a MongoDB at large concept, just a differentiation in our driver that isn't relevant to the server selection logging. I think we should stick with the shared concept of "runCommand" here.

nbbeeken (changes requested):

test/integration/crud/abstract_operation.test.ts
```diff
@@ -0,0 +1,324 @@
...
+          const subclassInstance = subclassCreator();
+          const prototypeCommandName = Object.getOwnPropertyDescriptor(
+            subclassType.prototype,
+            'commandName'
+          )?.get?.call(null);
+          expect(prototypeCommandName).to.equal(subclassInstance.commandName);
+        });
+
+        if (!serverlessOperationSubclasses.includes(subclassType.name.toString())) {
+          it(`server.command's first key should equal operation.commandName`, async function () {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Is this because we secretly aren't checking that it is the first key....

test/integration/crud/abstract_operation.test.ts
```diff
@@ -0,0 +1,324 @@
...
+            const subclassInstance = subclassCreator();
+            const yieldDoc =
+              subclassType.name === 'ProfilingLevelOperation' ? { ok: 1, was: 1 } : { ok: 1 };
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: It actually won't be able to be moved there since the `beforeEach()` does not have access to which class is currently being read. Similar special cases are handled within this test case as well (see the lines in this test case with the `.includes(subclassType.name.toString())`.

src/operations/operation.ts
```diff
@@ -75,6 +75,8 @@ export abstract class AbstractOperation<TResult = any> {
     this.trySecondaryWrite = false;
   }
 
+  abstract get commandName(): string;
```
baileympearson, line 78: q: why are we using a `getter` instead of just a string property?

src/operations/delete.ts
```diff
@@ -100,6 +104,10 @@ export class DeleteOneOperation extends DeleteOperation {
     super(collection.s.namespace, [makeDeleteStatement(filter, { ...options, limit: 1 })], options);
   }
 
+  override get commandName() {
+    return super.commandName;
+  }
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: I worked with Neal, I'm able to remove the super.command getters.

src/operations/operation.ts
aditi-khare-mongoDB: We want the command name to be read-only, and since there's no setter, a getter will ensure this. In addition, across all instances there's only one definition of the getter, but as for a string it would be duplicated for every instance.

test/integration/crud/abstract_operation.test.ts
```diff
@@ -0,0 +1,324 @@
...
+    ];
+
+    const sameServerOnlyOperationSubclases = ['GetMoreOperation', 'KillCursorsOperation'];
+
+    let client;
+    let db;
+    let admin;
+    let collection;
+    let constructorServer;
+    const subclassArray: AbstractOperationSubclasses[] = [
```
baileympearson, line 34: First - nice work with these tests.  They're thorough and well structured.

but I was curious about the intention behind this exhaustive testing.  is it really necessary?  I generally think exhaustive testing is good when we have complex, important logic that we must ensure is correct.  This, however, is just a single property on each class.  It seems bit overkill.  

I was thinking about what would happen in the future when we add new operation classes.  A new operation class will not fail any test in this file without manual modification of this file because these tests are built from a hard-coded list of operations.

One argument in favor of exhaustive testing is that it aids future developers.  By writing an exhaustive test, if a new developer adds a new operation class that doesn't match the criteria outlined in these tests, it should fail and alert the developer.  This prevents "tribal knowledge" being important to ensuring that a codebase works as intended.  But these tests require modification when a new operation class is added and nothing would fail if a new operation class is added, _and_ a developer would need to know these tests exist.  So there's some level of "tribal knowledge" necessary regardless of whether or not we have these tests.

My 2-cents is that a descriptive comment on `get commandName()` in `AbstractOperation` is sufficient and we can remove all tests.  If we _really_ want to test this property, I'd say it's sufficient to just test string equality.  `it('AggregateOperation.commandName is `aggregate`', () => { ... })`.  Tests confirming that the string is as intended combined with a thorough comment on `get commandName()` is sufficient to make me confident that our command names are correct and won't change.

test/integration/crud/abstract_operation.test.ts
```diff
@@ -0,0 +1,324 @@
...
+      {
+        subclassCreator: () =>
+          new mongodb.FindOneAndUpdateOperation(collection, { a: 2 }, { $a: 1 }, {}),
+        subclassType: mongodb.FindOneAndUpdateOperation
+      },
+      {
+        subclassCreator: () =>
+          new mongodb.GetMoreOperation(
+            collection.fullNamespace,
+            Object.freeze(Long.fromNumber(1)),
```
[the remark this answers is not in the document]
nbbeeken: It is fine to leave it or change it, just wanted to understand, ty!

test/integration/crud/abstract_operation.test.ts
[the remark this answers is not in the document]
nbbeeken: Well, is it a secret or is it the truth? 😄 If the string we're looking for _stopped_ being the first key then it would fail once it reaches the server, I don't think it is critical we re-assert that here.

test/integration/crud/abstract_operation.test.ts
aditi-khare-mongoDB: I agree that are command names likely won't change.

The reasoning behind the exhaustive tests was to essentially ensure that the command name matches the first key of the command / to ensure I didn't make any mistakes whilst creating the getters / or miss any. I generated the subclasses by iterating through the mongodb classes and checking for subclasses of AbstractOperation, so as to ensure I didn't miss any subclasses during my implementation. So personally, I do think that the sinon test is more important than string equality, since the whole point of command name is that it matches the first key of the command.

If a new operation class is added, and we want to ensure it is in the test file, we can loop through the` subclasses of Abstract Operation === subclassesArray`, but seeing as the current tests may be overkill, that might not be favorable.

test/integration/crud/abstract_operation.test.ts
```diff
@@ -0,0 +1,324 @@
...
+    });
+
+    for (const { subclassCreator, subclassType } of subclassArray) {
+      context(`when subclass is ${subclassType.name}`, async function () {
+        it(`subclass prototype's commandName should equal operation.commandName`, async function () {
+          const subclassInstance = subclassCreator();
+          const prototypeCommandName = Object.getOwnPropertyDescriptor(
+            subclassType.prototype,
+            'commandName'
+          )?.get?.call(null);
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: subclassInstance.commandName does work, if you see the kickoff, this is to ensure that we can access the commandName without an instance of the class (that's why we use a getter call to the` subclassType.prototype `here). I will end up just removing this test case though, in favor of just string equality, so we can get rid of the funky `super.commandName' getters.

src/operations/operation.ts
aditi-khare-mongoDB: I guess it's a small optimization, but the getter won't be called on every instance necessarily (for ex: if logging is turned off), so it's not necessary to duplicate was my logic. @nbbeeken anything to add?

aditi-khare-mongoDB: @nbbeeken just sent me some context, here is a prior precedent for constant string names as getters in our codebase: https://github.com/mongodb/node-mongodb-native/blob/NODE-5741-use-io-async/src/error.ts#L688-L690

test/integration/crud/abstract_operation.test.ts
aditi-khare-mongoDB: Also, we can add a test in after logging is completed that goes like "for every commandStarted log there is a corresponding serverSelectionStarted log for a command of the same name" to ensure that commandName works for each new subclass as intended. This way we don't really need to change these tests. What do you think @baileympearson?

src/operations/operation.ts
nbbeeken: Not much, getter vs property is borderline stylistic, I think getters seem natural for this kind of const value and it does seem to follow our error classes and _bsontype approaches.

test/integration/crud/abstract_operation.test.ts
```diff
@@ -288,26 +339,22 @@ describe('abstract operation', async function () {
...
       context(`when subclass is ${subclassType.name}`, async function () {
-        it(`subclass prototype's commandName should equal operation.commandName`, async function () {
+        it(`operation.commandName equals correct string`, async function () {
           const subclassInstance = subclassCreator();
-          const prototypeCommandName = Object.getOwnPropertyDescriptor(
-            subclassType.prototype,
-            'commandName'
-          )?.get?.call(null);
-          expect(prototypeCommandName).to.equal(subclassInstance.commandName);
+          expect(correctCommandName).to.equal(subclassInstance.commandName);
```
nbbeeken, line 346:
```suggestion
          expect(subclassInstance.commandName).to.equal(correctCommandName);
```

Minor but when this prints an error you'll want it to say the "result" of commandName, did not equal the expected "correctCommandName". The value under test goes into `expect()`

src/operations/operation.ts
```diff
@@ -75,6 +75,7 @@ export abstract class AbstractOperation<TResult = any> {
     this.trySecondaryWrite = false;
   }
 
+  // Must match the first key of the command object sent to server
```
nbbeeken, line 78:
```suggestion
  /** Must match the first key of the command object sent to the server */
```

This format makes it a doc comment, then it will appear when you hover over it in VSCode

nbbeeken (approved):

test/integration/crud/abstract_operation.test.ts
baileympearson: > So personally, I do think that the sinon test is more important than string equality, since the whole point of command name is that it matches the first key of the command.

Are we testing this behavior now?  I see that we're checking that the command name exists in the command document, but no tests that assert it's the first key.

> If a new operation class is added, and we want to ensure it is in the test file, we can loop through the subclasses of Abstract Operation === subclassesArray, but seeing as the current tests may be overkill, that might not be favorable.

I don't fully understand what this is saying.  Are you suggesting we generate `subclassArray` programmatically?

aditi-khare-mongoDB: We're not asserting that it is the first key, we very well could. 

`subclassArray` is a bit confusingly named, it is an array of subclass names and instances. 
We can't really generate `subclassArray` in a neat programmatic way since subclasses do not have consistent constructor arguments. I'm suggesting that we can generate the names only of the subclasses programmatically, and detect if any subclasses are missing from `subclassArray`.

baileympearson: I think that's probably overkill too.  I don't feel strongly here, I'll go ahead and resolve this conversation.

PR #3942, opened 2023-12-05
refactor(NODE-5778): change kConnection to connection, split polling and streaming functions, define exhaustCommand

nbbeeken:
#### What is changing?

Some minor refactors to the monitoring class 

No

To pull them out of the "feature" work of adding streaming support to the new monitoring class.

### DIFF
--- a/src/cmap/commands.ts
+++ b/src/cmap/commands.ts
@@ -317,6 +317,9 @@ export class OpQueryResponse {
   bsonRegExp?: boolean;
   index?: number;
 
+  /** moreToCome is an OP_MSG only concept */
+  moreToCome = false;
+
   constructor(
     message: Buffer,
     msgHeader: MessageHeader,
@@ -598,6 +601,7 @@ export class OpMsgResponse {
   fromCompressed?: boolean;
   responseFlags: number;
   checksumPresent: boolean;
+  /** Indicates the server will be sending more responses on this connection */
   moreToCome: boolean;
   exhaustAllowed: boolean;
   useBigInt64: boolean;
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -563,13 +563,13 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     }
   }
 
-  moreToComeCommand(
+  exhaustCommand(
     ns: MongoDBNamespace,
     command: Document,
     options: CommandOptions | undefined,
-    callback: Callback
+    replyListener: Callback
   ) {
-    return this.command(ns, command, options, callback);
+    return this.command(ns, command, options, replyListener);
   }
 }
 
@@ -1164,6 +1164,15 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
     }
 
     return document;
+  }
+
+  exhaustCommand(
+    _ns: MongoDBNamespace,
+    _command: Document,
+    _options: CommandOptions,
+    _replyListener: Callback
+  ) {
+    throw new Error('NODE-5742: not implemented.');
   }
 }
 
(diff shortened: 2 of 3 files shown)


src/cmap/connection.ts
```diff
@@ -562,6 +562,15 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       callback(err);
     }
   }
+
+  moreToComeCommand(
```
baileympearson, line 566: Are we attached to the name `moreToComeCommand`?  The publicly documented many-to-one behavior is referrred to as "exhaust" (exhaust allowed, exhaust cursors, etc), and `moreToCome` is just an implementation detail of exhaust support.   `exhaustCommand` might make more sense

durran, line 566: Just out of curiosity, what was the need for this? Are there upcoming changes that will alter the behaviour of this method?

src/cmap/commands.ts
```diff
@@ -317,6 +317,9 @@ export class OpQueryResponse {
   bsonRegExp?: boolean;
   index?: number;
 
+  /** moreToCome is an OP_MSG only concept */
+  moreToCome = false;
```
baileympearson, line 321: why was this necessary?  I'd prefer to keep our existing `!('moreToCome' in response)` check than add an OpMsg-specific concept to our QpQuery classes.

also - it shouldn't be necessary on the request, right?

src/cmap/connection.ts
nbbeeken: No, this method will remain as is until we remove the old connection class entirely. This is so the monitor can be refactored to support either Connection or ModernConnection without needing to gaurd on which instance of the connection class it is using. I added the empty method to ModernConnection to make it a bit more clear, the next segment of work will fill out that method and make Monitors start testing with the new connection class.

nbbeeken: SGTM

src/cmap/commands.ts
nbbeeken: It is not necessary, but neither is the `in` check, which is artificially introduced by typescript. Since OP msg vs query isn't something our driver code generally needs to discriminate it would be ideal if we could create an interface that both classes implement, then functions that do not care which protocol is being used can depend on that interface. 

That interface idea would be nice but a bit more scope than is appropriate for one condition improvement. I do not feel too strongly.

src/cmap/connection.ts
durran: Ah ok, thanks for the explanation.

durran (approved):

src/sdam/monitor.ts
```diff
@@ -94,21 +92,17 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
   connectOptions: ConnectionOptions;
   isRunningInFaasEnv: boolean;
   [kServer]: Server;
-  [kConnection]?: Connection;
+  connection: Connection | null;
```
alenakhineika, line 95: What is the motivation behind changing kConnection to connection? Or the backstory of why it was a symbol?

alenakhineika (approved):

src/sdam/monitor.ts
baileympearson: In the pre-TS days of the driver, there were issues with users accessing driver internal properties because we had no way to mark them internal.  So we switched to using symbol properties, similar to how Nodejs handles its internal properties, because although they're still accessible to users, it's less accessible.

Now that we have TS with the `private` keyword, generated TS-doc that strips out anything with an `@internal` tag and someday JS private properties, we don't really need symbol properties anymore.  

We're slowly removing them because they cause headaches for us:

- harder to access in tests
- no cmd+click to jump to definition of a class property
- doc comments don't appear on class properties that use symbols when you hover over the property
- etc

baileympearson (approved):

src/sdam/monitor.ts
alenakhineika: Got it! Thx for the context!

PR #3944, opened 2023-12-07
 refactor(NODE-5742): add streaming responses support

nbbeeken:
#### What is changing?

- Using an async iterator return responses as they arrive
- Change command to return after reciving one response from the iterator
- Correct tests that spy on conneciton.command

No

- Support monitoring streaming protocol


src/cmap/connection.ts
```diff
@@ -1103,76 +1102,87 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
+    ns: MongoDBNamespace,
+    command: Document,
+    options: CommandOptions,
+    replyListener: Callback
   ) {
-    throw new Error('NODE-5742: not implemented.');
+    const exhaustLoop = async () => {
+      this.controller.signal.throwIfAborted();
+      for await (const reply of this.sendCommand(ns, command, options)) {
+        this.controller.signal.throwIfAborted();
```
durran, line 1179: What's the reasoning for throwing if aborted both before and after handling the reply?

src/sdam/monitor.ts
```diff
@@ -2,7 +2,7 @@ import { clearTimeout, setTimeout } from 'timers';
 
 import { type Document, Long } from '../bson';
 import { connect } from '../cmap/connect';
-import { Connection, type ConnectionOptions } from '../cmap/connection';
+import { type Connection, type ConnectionOptions } from '../cmap/connection';
```
durran, line 5:
```suggestion
import type { Connection, ConnectionOptions } from '../cmap/connection';
```
nit.

durran (changes requested):

src/cmap/connection.ts
```diff
@@ -1103,76 +1102,87 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
-        this.emit(
-          ModernConnection.COMMAND_FAILED,
-          new CommandFailedEvent(this as unknown as Connection, message, ioError, started)
-        );
-      }
-      throw ioError;
-    }
+      this.controller.signal.throwIfAborted();
+      for await (document of this.sendWire(message, options)) {
+        if (!Buffer.isBuffer(document) && document.writeConcernError) {
```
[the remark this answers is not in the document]
nbbeeken: That's correct, previously callback code "got away" with not checking for the type of document because of the lack of mutual exclusion between the error / document and just weak typing in general. For example, things like `document.writeConcernError` are just falsey if document is a Buffer. Now that we have better TS with promises we are forced to handle the potential for a Buffer, the code will continue to work as long as we pass it along to the next layer above. 

We should probably improve this, but I imagine it is a heavy lift that we will deal with in adopting on_demand bson anyway.

src/cmap/connection.ts
nbbeeken: Above the for-await loop handles the case where we are first entering this code, the signal may already be aborted at that point and we want to bail out before blocking on a promise. The one inside the loop is checking on each iteration if we have aborted the signal since the promise resolved. I think you may be "smelling" something here that I can try cleaning up. It is important to check the status of the signal before hitting an `await`, so the one at the end of loop is still required. However, this one at the top _should_ never throw, because we just got a resolved reply so there shouldn't have been the chance to abort the signal since that succeeded.

src/sdam/monitor.ts
nbbeeken: done

src/cmap/connection.ts
nbbeeken: Cleaned up a handful and locally looked good, CI will have the final word.

durran (approved):

PR #3946, opened 2023-12-11
feat(NODE-4687): Add logging to server selection

aditi-khare-mongoDB:
### Description
Add logging to server selection.

#### What is changing?
When logging is enabled at the debug level, users will be able to see "server selection started, succeeded and failed events". When logging is enabled at the info level, users will be able to see "waiting for suitable server events".

Not until logging is released (NODE-5672)

To help users see their server selection events, depending on their logging level enabled.


aditi-khare-mongoDB:
Reference to the failing tests: there are two spec tests with the same description 'server-selection-logging', which cause a MochaError. Anyone know a work around?

alenakhineika:
> Reference to the failing tests: there are two spec tests with the same description 'server-selection-logging', which cause a MochaError. Anyone know a work around?

@dariakp I am wondering if this should be changed in the spec itself?

durran:
> > Reference to the failing tests: there are two spec tests with the same description 'server-selection-logging', which cause a MochaError. Anyone know a work around?
> 
> @dariakp I am wondering if this should be changed in the spec itself?

We shouldn't be duplicating all the unified tests here. I'd update the name of one of the duplicates here in this PR, open a DRIVERS ticket to change the name, and point to this PR as the reference implementation for the change.

aditi-khare-mongoDB:
I found a related ticket in the logging epic filed recently about potentially removing Load Balancer mode in server selection logging, this would also mean removing one of the duplicate test descriptions. I messaged in the drivers-1204-logging channel for more information: 
https://mongodb.slack.com/archives/G015EB3QHED/p1702492490091139

src/change_stream.ts
```diff
@@ -934,10 +934,14 @@ export class ChangeStream<
       this.cursor.close().catch(() => null);
 
       const topology = getTopology(this.parent);
-      topology.selectServer(this.cursor.readPreference, {}, serverSelectionError => {
-        if (serverSelectionError) return this._closeEmitterModeWithError(changeStreamError);
-        this.cursor = this._createChangeStreamCursor(this.cursor.resumeOptions);
-      });
+      topology.selectServer(
+        this.cursor.readPreference,
+        { operationName: 'reconnect topology' },
```
alenakhineika, line 939: Which operations cause a server selection? Should this be the particular name of the operation for which a server is being selected?

src/mongo_logger.ts
```diff
@@ -294,7 +308,7 @@ export interface LogConvertible extends Record<string, any> {
 
 /** @internal */
 export function stringifyWithMaxLen(value: any, maxDocumentLength: number): string {
-  const ejson = EJSON.stringify(value);
+  const ejson = EJSON.stringify(value) ?? 'custom function';
```
alenakhineika, line 311: Is it the application-provided custom selector? Maybe then 'custom selector' would be more descriptive rather than 'custom function'? Can we assign the same identifier multiple times or it should be unique? Do we want it to be a plain string like this or this should be some sort of stringified object format? Not sure where this being used, maybe some parsing logic later is being involved?

src/sdam/server_selection_events.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @internal
+ * @category Log Type
+ */
+export abstract class ServerSelectionEvent {
+  /** String representation of the selector being used to select the server.
+   *  Defaults to 'custom selector' for application-provided custom selector case.
+   */
+  selector: string | ReadPreference | ServerSelector;
+  /** The name of the operation for which a server is being selected.  */
+  operation: string;
```
alenakhineika, line 22: Is it operationName that you have used in execute_operation.ts and change_stream.ts? If yes should we use one prop name for consistency?

src/sdam/server_selection_events.ts
```diff
@@ -0,0 +1,137 @@
...
+import { type ReadPreference } from '../read_preference';
+import { type ServerSelector } from './server_selection';
+import type { TopologyDescription } from './topology_description';
+
+/**
+ * The base export class for all logs published from server selection
+ * @internal
+ * @category Log Type
+ */
+export abstract class ServerSelectionEvent {
```
alenakhineika, line 16: I have noticed that bulk operations use `operationId`. Is it a different id? Or we should add operationId as an optional prop of the ServerSelectionEvent class in case it presents.

src/sdam/server_selection_events.ts
```diff
@@ -0,0 +1,137 @@
...
+  abstract message: string;
+
+  /** @internal */
+  constructor(
+    selector: string | ReadPreference | ServerSelector,
+    topologyDescription: TopologyDescription,
+    operation?: string
+  ) {
+    this.selector = selector;
+    this.operation = operation ?? 'n/a';
```
alenakhineika, line 42: Can server selection happen without operation that caused it?

src/sdam/server_selection_events.ts
```diff
@@ -0,0 +1,137 @@
...
+ * An event published when server selection succeeds
+ * @internal
+ * @category Event
+ */
+export class ServerSelectionSucceededEvent extends ServerSelectionEvent {
+  /** @internal */
+  name = SERVER_SELECTION_SUCCEEDED;
+  message = 'Server selection succeeded';
+  /** The address (host/port pair) of the pool */
+  address: string;
```
alenakhineika, line 101: The spec says `This message MUST contain the following key-value pairs: message, serverHost , serverPort `, so maybe we shout split address to those?

src/sdam/topology.ts
```diff
@@ -881,13 +931,33 @@ function processWaitQueue(topology: Topology) {
...
+    let selectedServer: Server | undefined;
     if (selectedDescriptions.length === 0) {
+      if (!waitQueueMember.waitingLogged) {
+        topology.client.mongoLogger.info(
+          MongoLoggableComponent.SERVER_SELECTION,
+          new WaitingForSuitableServerEvent(
+            waitQueueMember.serverSelector,
+            topology.description,
+            waitQueueMember.timeoutController.getRemainingTimeMS(),
+            waitQueueMember?.operationName
```
alenakhineika, line 956: If we always can access waitQueueMember.timeoutController does it mean that we don't need to check if waitQueueMember exists to access operationName? So basically operationName will always be present?

test/integration/server-selection/server_selection.spec.test.ts
```diff
@@ -0,0 +1,20 @@
+import * as path from 'path';
+
+import { loadSpecTests } from '../../spec';
+import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
+
+describe.skip('Server Selection Unified Tests (Spec)', function () {
```
alenakhineika, line 6: Is skipping here intentional? If yes we could leave a todo with a ticket to unskip it.

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,13 @@
+import { loadSpecTests } from '../../spec';
+import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
+
+describe.only('Server Selection Unified Tests (Node Driver)', function () {
```
alenakhineika, line 4: Is only here intentional? Or you are still debugging this PR?

src/sdam/server_selection_events.ts
aditi-khare-mongoDB: Good idea! I've added in these changes.

src/sdam/server_selection_events.ts
aditi-khare-mongoDB: Not that I know of, I'll make it a non-optional property.

src/sdam/topology.ts
aditi-khare-mongoDB: I don't think we need optional chaining here, I've removed it

src/change_stream.ts
alenakhineika: The spec says that operation is `The name of the operation for which a server is being selected. When server selection is being performed to select a server for a command, this MUST be the command name`. If different commands cause server selection, is there a way to pass the actual command name to this method?

src/mongo_logger.ts
alenakhineika: Can you give an example of what `value` is for `stringifyWithMaxLen(selector, maxDocumentLength)` and `stringifyWithMaxLen(topologyDescription, maxDocumentLength)` if those are not empty?

src/sdam/server_selection_events.ts
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * An event published when server selection succeeds
+ * @internal
+ * @category Event
+ */
+export class ServerSelectionSucceededEvent extends ServerSelectionEvent {
+  /** @internal */
+  name = SERVER_SELECTION_SUCCEEDED;
+  message = 'Server selection succeeded';
+  /** 	The hostname, IP address, or Unix domain socket path for the selected server. */
```
alenakhineika, line 101:
```suggestion
  /** The hostname, IP address, or Unix domain socket path for the selected server. */
```

aditi-khare-mongoDB: Sounds good!

src/change_stream.ts
aditi-khare-mongoDB: There's not a specific command name for iterating through the change stream as far as I'm aware. The reason for this server selection is that while iterating through the change stream, connection was lost. This is why I named it "reconnect topology." We could very well change it to something like "reconnect topology in change stream" to be more descriptive.

src/mongo_logger.ts
aditi-khare-mongoDB: Yes!

The selector can take on 3 different types, I've listed them with their example `stringifyWithMaxLen` inputs below.

- Example string: `'primary'`
- Example ReadPreference object: 
```
{
"mode":"primary", "tags":null, "hedge":null, "maxStalenessSeconds":null," "minWireVersion":null
}
```
- Example Function: 
```
(err, server) => {
  if (err) {
    return this.close({ force: false }, () => exitWithError(err));
  }

  stateTransition(this, STATE_CONNECTED);
  this.emit(Topology.OPEN, this);
  this.emit(Topology.CONNECT, this);

  callback?.(undefined, this);
}
```

As for the topologyDescription, it is always a TopologyDescription object.
Here is an example: 
```
{
"type":"Single","servers":{"localhost:27017":{"address":"localhost:27017","type":"Unknown","hosts":[],"passives":[],"arbiters":[],"tags":{},"minWireVersion":0,"maxWireVersion":0,"roundTripTime":-1,"lastUpdateTime":1718729665,"lastWriteDate":0,"error":{},"topologyVersion":null,"setName":null,"setVersion":null,"electionId":null,"logicalSessionTimeoutMinutes":null,"primary":null,"me":null,"$clusterTime":null}},"stale":false,"compatible":true,"heartbeatFrequencyMS":500,"localThresholdMS":15,"setName":null,"maxElectionId":null,"maxSetVersion":null,"commonWireVersion":0,"logicalSessionTimeoutMinutes":null
}
```

test/integration/server-selection/server_selection.test.ts
```diff
@@ -0,0 +1,13 @@
+import { loadSpecTests } from '../../spec';
+import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
+
+describe('Server Selection Unified Tests (Node Driver)', function () {
+  /* TODO(NODE-2471) ping on connect
+   * TODO(NODE-5774) duplicate server selection for bulkWrite and other wrapper operations
```
durran, line 6: I don't think we should be duplicating the tests in this PR - I'd skip them in the regular suite only and/or fix them in the regular spec tests and open a drivers ticket to resolve in the specs repo. This duplication here even with the TODO could be confusing.

durran (changes requested):

src/change_stream.ts
alenakhineika: In this case yes, the "reconnect topology in change stream" name or something like this would be more helpful to read which command the server selection was performed to.

test/integration/server-selection/server_selection.test.ts
aditi-khare-mongoDB: The node behavior currently performs extra server selections for certain operations (including bulkWrite, the spec test case), since this is node specific behavior, we don't want to open a drivers ticket. Similar duplication exists for SDAM spec v driver tests ([see here](https://github.com/mongodb/node-mongodb-native/tree/main/test/integration/server-discovery-and-monitoring/unified-sdam-node-specs)).

src/change_stream.ts
alenakhineika: This issue hasn't been resolved, right? I see the same command name here.

src/mongo_logger.ts
alenakhineika: Based on the function name the only task it should be doing is `stringify with max len`, it seems unexpected that it also provides a default value. Maybe you could set the default value before passing it to stringifyWithMaxLen. Additionally, as you mentioned, the selector can be either a string, function, or object, therefore each of them might have a different default value rather than `custom function`. How later selectors are being used? Are they parsed in any way? It would be nice to double-check and ensure that the generic plain string value won't break any existing parsing logic if it is expect to be an object-like value.

src/change_stream.ts
aditi-khare-mongoDB: Apologies for the late update, rebasing right now and will update with the new name.

test/integration/server-selection/server_selection.test.ts
durran: I didn't see a precedent was set for this situation, which I disagree with but am happy to concede if the rest of the team is ok with it. My opinion here is 2 concerns:

1. By duplicating the unified tests and modifying them to account for the extra ping, we now have a situation where the unified tests get updated, and we need to remember to sync them in two places and potentially modify them in one. I feel this is creating additional work for us in the future which I'd like to avoid.
2. We already have addressed this issue in the unified runner, [see here](https://github.com/mongodb/node-mongodb-native/blob/main/test/tools/unified-spec-runner/entities.ts#L225), so I'm curious as to why this is happening in the first place, and why we aren't addressing it as part of this ticket?

src/mongo_logger.ts
```diff
@@ -402,10 +436,7 @@ function attachCommandFields(
   return log;
 }
 
-function attachConnectionFields(
-  log: Record<string, any>,
-  event: ConnectionPoolMonitoringEvent | ServerOpeningEvent | ServerClosedEvent
-) {
+function attachConnectionFields(log: Record<string, any>, event: any) {
```
W-A-James, line 439: Why did we broaden they type of the `event` parameter?

src/operations/execute_operation.ts
```diff
@@ -251,7 +254,10 @@ async function retryOperation<
   }
 
   // select a new server, and attempt to retry the operation
-  const server = await topology.selectServerAsync(selector, { session });
+  const server = await topology.selectServerAsync(selector, {
+    session: session,
```
W-A-James, line 258:
```suggestion
    session,
```

W-A-James (changes requested):

src/mongo_logger.ts
aditi-khare-mongoDB: Yes that makes sense, I've changed it to use the `Function.toString() ` method Selectors as strings as turned back into objects if possible by unified spec runner when its running.

test/integration/server-selection/server_selection.test.ts
aditi-khare-mongoDB: 1. I can ask the team for more input. 

2. To my knowledge the `skipPingOnConnect` symbol doesn't stop server selection, but skips sending the ping to the server, is this the desired behavior. Here's the [code snippet](https://github.com/mongodb/node-mongodb-native/blob/b93d405275c3a8ce6b1eba0af335ffb2a309e34e/src/sdam/topology.ts#L455).

src/operations/execute_operation.ts
aditi-khare-mongoDB: Didn't catch that, should be changed now.

test/integration/server-selection/server_selection.test.ts
durran: Yeah I think we should discuss in tech chat or virtual office. It might be an easier approach then to update the matchers in the unified runner itself to ignore the pings when asserting logging spec events, since it currently doesn't look for them.

durran (changes requested):
There are failures in the unit tests now around max staleness. These tests are using the server selection test helper [here](https://github.com/mongodb/node-mongodb-native/blob/main/test/unit/assorted/server_selection_spec_helper.js#L83) so I suspect the changes in this PR are directly related.

aditi-khare-mongoDB:
Ah, I see, I'll be more mindful of running unit tests before I push in the future.

aditi-khare-mongoDB:
Current CI Failures are flaky tests:

- [Client Side Encryption Prose Tests 16 (NODE-4616)](https://jira.mongodb.org/browse/NODE-4616)
- [aws: custom endpoint - clientside encryption prose test [1] (NODE-5381)] (https://jira.mongodb.org/browse/NODE-5381)
- [Should properly handle a changeStream event being processed mid-close (NODE-5788)](https://jira.mongodb.org/browse/NODE-5788)

durran (approved):

alenakhineika (approved):

PR #3948, opened 2023-12-13
perf(NODE-5771): improve new connection

nbbeeken:
#### What is changing?

- Adds our own version of data event emitter -> asyncIterator helper
  - Notably changes from the standard library, removal of watermark functionality
  - Reduced the generic-ness to only work for data events and remove the need to destructure the value returned by the iterator
- Removes re-creating a AbortController per command
- Moves I/O helpers to class methods (to get Typescript privacy functionality)
- Creates promisified socketWrite helper _once_ at the Connection construction
- Connects a transform stream to socket that buffers until message size is received
- Reworks `abortable` helper to not create a "done" signal, instead simplify the promise wrapping and abort listener removal to lower performance cost

No.

Improves the performance of the new connection layer to be on par with our existing connection layer. The major cost was constructing many AbortSignals, however, going back to using a stream to _just_ collect chunks gets us within a margin of error of the old performance. 


src/cmap/connection.ts
```diff
@@ -1181,121 +1186,83 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
+   * @internal
+   *
+   * Returns an async generator that yields full wire protocol messages from the underlying socket.  This function
+   * yields messages until `moreToCome` is false or not present in a response, or the caller cancels the request
+   * by calling `return` on the generator.
+   *
+   * Note that `for-await` loops call `return` automatically when the loop is exited.
+   */
+  async *readMany(): AsyncGenerator<OpMsgResponse | OpQueryResponse> {
+    for await (const message of onData(this.messageStream, { signal: this.controller.signal })) {
```
baileympearson, line 1223: are we ever going to use such a specific helper `onData` anywhere else?  thoughts about making `onData` into a method on the MessageStream, such as `messages({ signal: AbortSignal }): AsyncGenerator<Buffer>`?

src/cmap/connection.ts
```diff
@@ -1181,121 +1186,83 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
-  const maxBsonMessageSize = connection.hello?.maxBsonMessageSize ?? kDefaultMaxBsonMessageSize;
-  for await (const [chunk] of on(connection.socket, 'data', { signal })) {
-    if (connection.delayedTimeoutId) {
-      clearTimeout(connection.delayedTimeoutId);
-      connection.delayedTimeoutId = null;
-    }
-
-    bufferPool.append(chunk);
-    const sizeOfMessage = bufferPool.getInt32();
+  /**
```
baileympearson, line 1190: where did `kDefaultMaxBsonMessageSize` go?

baileympearson (changes requested):
nothing blocking since I'm OOO starting tomorrow, just some comments

src/cmap/connection.ts
durran: Yeah I think we still need this check correct? Maybe the server won't send us something larger but something impersonating the server could.

src/cmap/wire_protocol/on_data.ts
```diff
@@ -0,0 +1,99 @@
+import { type EventEmitter } from 'events';
+
+import { List, promiseWithResolvers } from '../../utils';
+
+type PendingPromises = Omit<
+  ReturnType<typeof promiseWithResolvers<IteratorResult<Buffer>>>,
+  'promise'
+>;
+
+export function onData(emitter: EventEmitter, options: { signal: AbortSignal }) {
```
durran, line 10: All of in the virtual office know the reasoning for this but maybe we should still add some comments here on why we implemented this.

durran (changes requested):
One question - did we also profile memory consumption?

src/cmap/connection.ts
```diff
@@ -836,9 +839,18 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
 
     this.socket = stream;
     this.controller = new AbortController();
-    this.socket.on('error', this.onError.bind(this));
+
+    this.messageStream = this.socket
+      .on('error', this.onError.bind(this))
+      .pipe(new SizedMessageTransform({ connection: this }))
+      .on('error', this.onError.bind(this));
```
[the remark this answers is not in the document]
nbbeeken: There are two different streams here that I'm listening to error events from, so it is not before/after, it is a listener on the socket and a listener on the SizedMessageTransform.

src/utils.ts
```diff
@@ -1314,36 +1314,30 @@ export function isHostMatch(match: RegExp, host?: string): boolean {
...
+  { signal }: { signal: AbortSignal }
 ): Promise<T> {
-  const { abort, done } = aborted(signal);
+  const { promise: aborted, reject } = promiseWithResolvers<never>();
+
+  function rejectOnAbort() {
+    reject(signal.reason);
+  }
+
+  if (signal.aborted) rejectOnAbort();
```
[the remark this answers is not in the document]
nbbeeken: It is not necessary, we attach rejection handling synchronously because there is no interruption between this point and the line where we call `Promise.race`. We have test cases demonstrating what an already rejected abort signal does with this API. 

We want to create a rejected promise but we still want to race with the input promise so that if it is rejected the error from that promise takes priority. This is also why we can't use throwIfAborted here, it would make the signal's error come before the input promise's one.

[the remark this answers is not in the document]
nbbeeken: Answered: https://github.com/mongodb/node-mongodb-native/pull/3948/files/dd856540d6fd99032dac1090b696d38b754c47ac#r1431664719

I strived to use throwIfAborted where it makes sense, but in some cases, I am either inside a function that is not async but is supposed to return a promise, or I need to synchronously setup some logic based on the `aborted` flag.

src/cmap/connection.ts
nbbeeken: I won't predict the future here, but this method is still relatively generic, the `'data'` event is common to any Node.js stream, and it could be extended to take the event name as a parameter, it just is not necessary today.

I am not against making it a class method, but it also doesn't gain us anything at this time. It could be moved to the class later easily but moving it off the class later may be a heavier lift.

src/cmap/connection.ts
nbbeeken: > where did kDefaultMaxBsonMessageSize go?

Apologies I didn't capture this in the PR description, I removed it. I intended to make a follow-up, will do.

> Yeah I think we still need this check correct?

There is something very questionable about this flag:
1. We never respect the value that comes from the server because we [check the hello for the value](https://github.com/mongodb/node-mongodb-native/blob/main/src/cmap/connection.ts#L232), but the hello is [not available until the handshake is performed](https://github.com/mongodb/node-mongodb-native/blob/main/src/cmap/connect.ts#L162).
2. More importantly, `maxBsonMessageSize` is not a field in the hello reply:  https://www.mongodb.com/docs/manual/reference/command/hello/ We seem to have a mixture of `maxBsonObjectSize` and `maxMessageSizeBytes`.
  - `maxBsonObjectSize` and others are the fields used by bulkWrite to split a bulk op into correctly sized chunks
3. `maxMessageSizeBytes` is probably what was intended here but this default exceeds the server default by quite a bit, this is ~67mb the server's default is 48mb. In either case that is the cap on the server's input. 
  - I am not sure as a client we should be validating anything other than what OP_MSG supports which would be int32 max, we also know our BSON library cannot process a document more than 16mb

Happy to reintroduce an upper limit to make us feel "safe" but the actual answer to all of the above confusion may require a bit more research 🤔

src/cmap/connection.ts
```diff
@@ -1181,121 +1186,83 @@ export class ModernConnection extends TypedEventEmitter<ConnectionEvents> {
...
-  }
-): Promise<void> {
-  const finalCommand =
-    options.agreedCompressor === 'none' || !OpCompressedRequest.canCompress(command)
-      ? command
-      : new OpCompressedRequest(command, {
-          agreedCompressor: options.agreedCompressor ?? 'none',
-          zlibCompressionLevel: options.zlibCompressionLevel ?? 0
-        });
+/** @internal */
```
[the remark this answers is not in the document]
nbbeeken: > going back to using a stream to just collect chunks gets us within a margin of error of the old performance

It was necessary, the original benchmarks show a mix of regressions but the worst case was a 30% regression, fixing the issue by making fewer abort signals gains about 10% back and adding a minimal stream regained the remaining 20%. 

I did attempt just using a connection.bufferPool with the data events from the socket but it still showed measurably worse performance, I suspect that creating a promise per chunk rather than per message is what keeps us on par with the old connection class. 

> I'd prefer not to use a custom stream if we can help it.

I am curious why we would avoid streams, they are generally the well-trodden path for working with Node.js I/O APIs.

src/cmap/connection.ts
durran: I'm happy to have this as a follow up. Just wanted to make sure we don't remove it without some research after if we need it.

durran (approved):

src/cmap/connection.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-5800

andreialecu:
Hey @nbbeeken @durran - not sure how to easily bring this to attention, but we discovered this issue: https://github.com/Automattic/mongoose/issues/14877 and it seems to pin down to `mongodb` 6.4.0

6.3.0 is unaffected. There's a repro in the issue.

PR #3951, opened 2023-12-14
feat(NODE-3449): Add serverConnectionId to Command Monitoring Spec

aditi-khare-mongoDB:
### Description
Added serverConnectionId field to `CommandStartedEvent`, `CommandFailedEvent`,` CommandSucceededEvent`. The server generated connection id will be present on server versions 4.2+.

#### What is changing?
Command monitoring events emissions now include a serverConnectionId on versions 4.2+

Yes, we need to update the [following mongoDB manual page](https://www.mongodb.com/docs/drivers/node/current/fundamentals/monitoring/command-monitoring/) for versions 4.2+ to have it include `serverConnectionId` in its sample command events.

Command Events now support `serverConnectionId` to comply with spec change to Command Events API.
See [kickoff](https://docs.google.com/document/d/1dAVI05o62oouBe_SCYYz0jZ3fMPLqy6YvDb8F-Kf0lQ/edit) for more context.

### Release Highlight

Add `serverConnectionId` field to `CommandStartedEvent`, `CommandFailedEvent`,` CommandSucceededEvent`. The server generated connection id will be of JS type `bigint` and present on server versions 4.2+.


durran (changes requested):
We have unified tests (https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/tests/monitoring/server-connection-id.yml) that test that the server connection id field is present. I would have expected these to be failing before but they are not. So I'd expect some sort of update in the Unified Runner that would have failed before this change and passed after.

Also could we add unit tests that create the different events with various types of server connction ids and assert that we are converting them all correctly to bigints?

aditi-khare-mongoDB:
Before we had support for serverConnectionId, those tests were being skipped, but I added in support through a small change in schema.ts 

I’ll add in unit tests to test for correct bigint conversion.

nbbeeken (changes requested):

src/cmap/command_monitoring_events.ts
```diff
@@ -161,6 +195,15 @@ export class CommandFailedEvent {
   }
 }
 
+/* @internal */
+function parseServerConnectionID(serverConnectionId: number | Double | bigint | Long): bigint {
+  return Long.isLong(serverConnectionId)
+    ? serverConnectionId.toBigInt()
+    : typeof serverConnectionId === 'bigint' || typeof serverConnectionId === 'number'
+    ? BigInt(serverConnectionId)
+    : BigInt(serverConnectionId.valueOf());
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: It's not letting me cast a Double using the above code sample. Will using valueOf() instead lead to an incorrect implementation?

src/cmap/connection.ts
```diff
@@ -727,18 +730,36 @@ function write(
       if (err && reply?.ok !== 1) {
         conn.emit(
           Connection.COMMAND_FAILED,
-          new CommandFailedEvent(conn, command, err, operationDescription.started)
+          new CommandFailedEvent(
+            conn,
+            command,
+            err,
+            operationDescription.started,
+            conn.hello?.connectionId
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Good idea, just made this change.

durran (changes requested):
@aditi-khare-mongoDB I'm looking at our running of these tests in main and they are not being skipped and are passing. I expect them to fail before this PR: https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_next_rhel80_large_node_latest_test_latest_server_a3c02983bc9cead125373d86a5c47cf8f4e40f9e_23_12_14_15_29_05/0?type=T#L2527

I don't see how the change in schema.ts resolves this. The field in the spec test is `hasServerConnectionId` which I expect in the unified runner would need some special handling as that's not the name of the actual field. Maybe adding a method `hasServerConnectionId` on the events? Also notice these events get special handling in the unified runner (https://github.com/mongodb/node-mongodb-native/blob/main/test/tools/unified-spec-runner/match.ts#L416) and that field never gets checked (which is why they are passing on main).

src/cmap/command_monitoring_events.ts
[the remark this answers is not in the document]
nbbeeken: Can you elaborate on the issue you are seeing? 

It would not be incorrect, but unnecessary, a branching code path that does not need to exist could be removed so we should avoid adding that kind of debt to this logic.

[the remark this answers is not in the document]
aditi-khare-mongoDB: The JS BigInt constructor does not take in BSON Double's as input.
```
var BigInt: BigIntConstructor (value: string | number | bigint | boolean) => bigint
```
Were you thinking of using a different BigInt to support this?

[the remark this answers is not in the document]
nbbeeken: - https://tc39.es/ecma262/#sec-bigint-constructor-number-value
- https://tc39.es/ecma262/#sec-toprimitive
- https://tc39.es/ecma262/#sec-ordinarytoprimitive

That is really unfortunate that the TS library type declares such restrictive types for the input when it does not reflect the reality of what the runtime supports. 

This is an ideal place for using a [`ts-expect-error`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) comment to work around the limitation so we can get notified if this type is ever fixed.

test/tools/unified-spec-runner/match.ts
```diff
@@ -486,6 +486,9 @@ function compareEvents(
         expect.fail(`expected ${path} to be instanceof CommandStartedEvent`);
       }
       compareCommandStartedEvents(actualEvent, expectedEvent.commandStartedEvent, entities, path);
+      if (expectedEvent.commandStartedEvent.hasServerConnectionId) {
+        expect(actualEvent).property('serverConnectionId').to.exist;
+      }
```
nbbeeken, line 491: I think we need an explicit "not exist" assertion when `hasServerConnectionId` is false, not just a true case, right?

src/cmap/stream_description.ts
```diff
@@ -58,6 +59,9 @@ export class StreamDescription {
       return;
     }
     this.type = parseServerType(response);
+    if (response.connectionId) {
+      this.serverConnectionId = this.parseServerConnectionID(response.connectionId);
```
nbbeeken, line 63: Could we consider always defining the property and setting it to null by default? 

I don't think `connectionId` would ever be 0 but if it were this code would ignore it because it is falsey, just always assigning and defaulting to null removes the need to check for that.

Ref. https://jira.mongodb.org/browse/NODE-4068

src/cmap/command_monitoring_events.ts
```diff
@@ -153,6 +184,9 @@ export class CommandFailedEvent {
     this.commandName = commandName;
     this.duration = calculateDurationInMs(started);
     this.failure = maybeRedact(commandName, cmd, error) as Error;
+    if (serverConnectionId) {
+      this.serverConnectionId = serverConnectionId;
```
nbbeeken, line 188: Same question about defaulting to null here and the rest of the event classes.

nbbeeken (changes requested):

src/cmap/stream_description.ts
aditi-khare-mongoDB: What was the behavior before server v4.2 / did it have a connectionId back then? I feel it's much safer to check if a value exists in the `Document`, especially since it could contain anything. Currently, the parseServerConnectionId method cannot handle a nullish value and would throw an error.

test/tools/unified-spec-runner/match.ts
aditi-khare-mongoDB: We do, in the case that hasServerConnectionId exists but is set to false. I just added it in.

src/cmap/command_monitoring_events.ts
aditi-khare-mongoDB: No problem in defaulting to null here. It now defaults to null.

aditi-khare-mongoDB:
@durran Good catch, I've added proper support for the runner now.

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.spec.test.ts
```diff
@@ -12,7 +12,7 @@ import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
 const SKIP = ['A successful unordered bulk write with an unacknowledged write concern'];
 
 describe('Command Logging and Monitoring Spec', function () {
-  describe('Command Monitoring Spec (unified)', () => {
+  describe.only('Command Monitoring Spec (unified)', () => {
```
nbbeeken, line 15:
```suggestion
  describe('Command Monitoring Spec (unified)', () => {
```

src/cmap/stream_description.ts
nbbeeken: > What was the behavior before server v4.2 / did it have a connectionId back then?

There was no connectionId property on the hello response, so we can check for the presence of the key: `'connectionId' in response` or `response.connectionId == null` and set a default of null

src/cmap/command_monitoring_events.ts
```diff
@@ -139,7 +165,8 @@ export class CommandFailedEvent {
     connection: Connection,
     command: WriteProtocolMessageType,
     error: Error | Document,
-    started: number
+    started: number,
+    serverConnectionId?: bigint | undefined
```
nbbeeken, line 169:
```suggestion
    serverConnectionId: bigint | null
```

`null !== undefined` if we want to fix this to be defaulted to null then it must only support bigint or null and not allow optionality.

src/cmap/command_monitoring_events.ts
```diff
@@ -117,7 +137,13 @@ export class CommandSucceededEvent {
  */
 export class CommandFailedEvent {
   address: string;
+  /** Driver generated connection id */
   connectionId?: string | number;
+  /**
+   * Server generated connection id
+   * Distinct from the connection id and is returned by the hello or legacy hello response as "connectionId" from the server on 4.2+.
+   */
+  serverConnectionId?: bigint;
```
nbbeeken, line 146:
```suggestion
  serverConnectionId: bigint;
```

Same goes for the other event classes

test/tools/unified-spec-runner/match.ts
```diff
@@ -497,12 +502,22 @@ function compareEvents(
...
     } else if (expectedEvent.commandFailedEvent) {
       const path = `${rootPrefix}.commandFailedEvent`;
       if (!(actualEvent instanceof CommandFailedEvent)) {
         expect.fail(`expected ${path} to be instanceof CommandFailedEvent`);
       }
       compareCommandFailedEvents(actualEvent, expectedEvent.commandFailedEvent, entities, path);
+      if (expectedEvent.commandFailedEvent.hasServerConnectionId) {
+        expect(actualEvent).property('serverConnectionId').to.exist;
+      } else if (expectedEvent.commandFailedEvent.hasServerConnectionId === false) {
+        expect(actualEvent).property('serverConnectionId').to.not.exist;
```
nbbeeken, line 519:
```suggestion
        expect(actualEvent).property('serverConnectionId').to.be.null;
```

Now that we are making it default to null, might as well test for that explicitly to catch any mistakes.

test/tools/unified-spec-runner/match.ts
```diff
@@ -497,12 +502,22 @@ function compareEvents(
...
+      }
     } else if (expectedEvent.commandFailedEvent) {
       const path = `${rootPrefix}.commandFailedEvent`;
       if (!(actualEvent instanceof CommandFailedEvent)) {
         expect.fail(`expected ${path} to be instanceof CommandFailedEvent`);
       }
       compareCommandFailedEvents(actualEvent, expectedEvent.commandFailedEvent, entities, path);
+      if (expectedEvent.commandFailedEvent.hasServerConnectionId) {
+        expect(actualEvent).property('serverConnectionId').to.exist;
```
nbbeeken, line 517:
```suggestion
        expect(actualEvent).property('serverConnectionId').to.be.a('bigint');
```

nbbeeken (changes requested):

src/cmap/stream_description.ts
aditi-khare-mongoDB: Sounds good, just made these changes.

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.spec.test.ts
aditi-khare-mongoDB: Ah, yeah a stray only, just got rid of it.

test/tools/unified-spec-runner/match.ts
aditi-khare-mongoDB: Sounds good!

src/cmap/command_monitoring_events.ts
aditi-khare-mongoDB: Sounds good, just implemented this change.

src/cmap/stream_description.ts
```diff
@@ -73,4 +80,12 @@ export class StreamDescription {
       this.compressor = this.compressors.filter(c => response.compression?.includes(c))[0];
     }
   }
+
+  /* @internal */
+  parseServerConnectionID(serverConnectionId: number | Double | bigint | Long): bigint {
+    return Long.isLong(serverConnectionId)
+      ? serverConnectionId.toBigInt()
+      : // @ts-expect-error: Doubles are coercible to number
```
baileympearson, line 88: Can we also add a comment (and maybe link to the server code that supports this) that connection ids are always integral, and so coercing a double to an integer type is safe?

baileympearson (approved):
one small comment - LGTM otherwise!

durran (approved):

src/cmap/stream_description.ts
durran: I added the comment but did not link to server code as it's location can change over time.

durran (approved):

PR #3953, opened 2023-12-20
fix(NODE-5791): type error with $addToSet in bulkWrite

alenakhineika:
### Description
The `update` methods for `Collection` and `BulkWrite` with `$addToSet` should not show a _type_ error for collections with shema, collections extended from the `Document`/`any` type, and fields of `any` type.

#### What is changing?
- The type of `update` in `Collection.UpdateOneModel`/`Collection.UpdateManyModel`/`BulkWrite.updateOne`/`BulkWrite.updateMany` becomes `UpdateFilter<TSchema> | Document[]`.
- The `SetFields` type helper skips validation for fields of `any` type (provided by custom interface or index signatures).

None

[NODE-5647](https://jira.mongodb.org/browse/NODE-5647)
[NODE-4664](https://jira.mongodb.org/browse/NODE-4664)

### The type error is fixed for fields of `any` type in update commands.
Extending collection interfaces from `Document`/`any` type or specifying fields of `any` type doesn't cause a type error anymore.


src/collection.ts
```diff
@@ -340,12 +340,17 @@ export class Collection<TSchema extends Document = Document> {
...
    */
   async updateOne(
     filter: Filter<TSchema>,
-    update: UpdateFilter<TSchema> | Partial<TSchema>,
+    /**
+     * The value of update can be either:
+     * UpdateFilter<TSchema> - A document that contains update operator expressions,
+     * Document[] - an aggregation pipeline.
+     * */
+    update: UpdateFilter<TSchema> | Document[],
```
nbbeeken, line 353: A TSDoc comment on a parameter like this doesn't render in the docs, you can run `npm run build:docs` and `npm run preview:docs` to see for yourself

src/collection.ts
```diff
@@ -340,12 +340,17 @@ export class Collection<TSchema extends Document = Document> {
    * Update a single document in a collection
    *
```
nbbeeken, line 341: I would add the below comment up here and format it a bit differently to get it to render nicely, you can use markdown syntax (ex. bullets and fixed-width):

```ts
  /**
   * Update a single document in a collection
   *
   * The value of `update` can be either:
   * - UpdateFilter<TSchema> - A document that contains update operator expressions,
   * - Document[] - an aggregation pipeline.
   *
   * @param filter - The filter used to select the document to update
   * @param update - The modifications to apply
   */
```

nbbeeken (changes requested):

src/collection.ts
alenakhineika: TIL ✨

src/collection.ts
alenakhineika: Updated with your suggestion.

nbbeeken (changes requested):

nbbeeken (approved):

durran (approved):

PR #3955, opened 2023-12-20
feat(NODE-4686): Add log messages to CLAM

aditi-khare-mongoDB:
### Description
Add logging to CLAM events, un-skip + add support for CLAM logging spec tests, and add in prose tests.

#### What is changing?
CLAM logging messages will be enabled when loggingClientOptions for component `MongoLoggableComponent.COMMAND >= SeverityLevel.DEBUG`.

NOTE:
-  Command logging is enabled separate from command monitoring, and the command logging messages contain additional fields than command monitoring, resulting in the need for the additional `emitAndLogCommand` method.
- More context can be found within the [kickoff](https://docs.google.com/document/d/1t8Of6gEbpbhpD3wIyS3qPIGr5YroRTtdQHXw524D1O4/edit). 
- through inspection of the spec runner, `failureIsRedacted` is already supported

Not until logging is released (NODE-5762)

To allow users to see CLAM events through logging. 


durran (changes requested):
Can we rebase this quickly against main first so we have an accurate diff against the new connection class changes?

aditi-khare-mongoDB:
@durran rebase is done, TIA!

src/cmap/connection.ts
```diff
@@ -441,10 +449,16 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     const message = this.prepareCommand(ns.db, command, options);
 
     let started = 0;
-    if (this.monitorCommands) {
+    if (
+      this.monitorCommands ||
+      (typeof this.description.serverConnectionId === 'bigint' &&
```
durran, line 454:
```suggestion
      (this.description.serverConnectionId &&
```
Only need to check for existence here as we always cast it to bigint when set.

src/cmap/connection.ts
```diff
@@ -464,9 +478,15 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
           throw new MongoServerError(document);
         }
 
-        if (this.monitorCommands) {
-          this.emit(
+        if (
+          this.monitorCommands ||
+          (typeof this.description.serverConnectionId === 'bigint' &&
```
durran, line 483:
```suggestion
          (this.description.serverConnectionId &&
```

src/cmap/connection.ts
```diff
@@ -481,10 +501,16 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
         this.controller.signal.throwIfAborted();
       }
     } catch (error) {
-      if (this.monitorCommands) {
+      if (
+        this.monitorCommands ||
+        (typeof this.description.serverConnectionId === 'bigint' &&
```
durran, line 506:
```suggestion
        (this.description.serverConnectionId &&
```

src/mongo_logger.ts
```diff
@@ -285,6 +281,40 @@ function compareSeverity(s0: SeverityLevel, s1: SeverityLevel): 1 | 0 | -1 {
   return s0Num < s1Num ? -1 : s0Num > s1Num ? 1 : 0;
 }
```
durran, line 283: Are we missing a `LoggableCommandStartedEvent`?

src/mongo_types.ts
```diff
@@ -438,6 +441,27 @@ export class TypedEventEmitter<Events extends EventsDescription> extends EventEm
...
+  emitAndLogCommand<EventKey extends keyof Events>(
+    monitorCommands: boolean,
+    event: EventKey | symbol,
+    databaseName: string,
+    ...args: Parameters<Events[EventKey]>
+  ): void {
+    if (monitorCommands) {
+      this.emit(event, ...args);
+    }
+    if (this.component && typeof args[0]?.serverConnectionId === 'bigint') {
```
durran, line 454:
```suggestion
    if (this.component && args[0]?.serverConnectionId) {
```

durran (changes requested):
Note that the unit tests are failing because the test command failed event has no `failure` on it.

src/cmap/connection.ts
aditi-khare-mongoDB: Sounds good, didn't catch that originally!

src/mongo_logger.ts
aditi-khare-mongoDB: No, the Events and Logging Spec do not differ for `CommandStartedEvent`, so they both use one class defined in [src/cmap/command_monitoring_events.ts](https://github.com/mongodb/node-mongodb-native/blob/b93d405275c3a8ce6b1eba0af335ffb2a309e34e/src/cmap/command_monitoring_events.ts#L18)

aditi-khare-mongoDB:
Added in the failure field to the unit tests

durran (changes requested):
Looks like we have the same failures with the extra pings here now on this PR. An extra 2 events are getting sent and the unified tests are failing.

aditi-khare-mongoDB:
I had a syntax error in emitAndLogCommand that made the output invalid, is fixed now.

durran (changes requested):
There are still failures that need to be addressed both in unit tests:

https://spruce.mongodb.com/task/mongo_node_driver_next_lint_run_unit_tests_node_16_patch_efece79d95a2c9892915ffd8c04f229d9e84196d_659726d4a4cf475f0b6a7eb6_24_01_04_21_44_53/logs?execution=0&sortBy=STATUS&sortDir=ASC

And in the integration tests:

https://spruce.mongodb.com/task/mongo_node_driver_next_rhel8_no_auth_tests_test_4.0_server_noauth_patch_efece79d95a2c9892915ffd8c04f229d9e84196d_659726d4a4cf475f0b6a7eb6_24_01_04_21_44_53/logs?execution=0&sortBy=STATUS&sortDir=ASC

durran (dismissed):

src/mongo_logger.ts
```diff
@@ -383,7 +413,9 @@ export function stringifyWithMaxLen(
   maxDocumentLength: number,
   options: EJSONOptions = {}
 ): string {
-  let strToTruncate: string;
+  // eslint-disable-next-line @typescript-eslint/no-inferrable-types
+  let strToTruncate: string = '';
```
alenakhineika, line 417:
```suggestion
  let strToTruncate = '';
```
You don't need to disable the eslint here. There is just no need to use an explicit type annotation.

aditi-khare-mongoDB: Ah I see!

src/cmap/connection.ts
```diff
@@ -441,10 +449,16 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     const message = this.prepareCommand(ns.db, command, options);
 
     let started = 0;
-    if (this.monitorCommands) {
+    if (
+      this.monitorCommands ||
+      (this.description.hello && this.mongoLogger?.willLog(SeverityLevel.DEBUG, this.component))
```
nbbeeken, line 454: There is quite a bit of non-booleans here and it would be nice to be clear about the expectations, so rather than depending on the falsey nature of hello can you check for null, and same goes for the `emitAndLogCommand` function, instead of passing in the hello document, can it only take in a boolean "isBefore/AfterHandshake"

src/mongo_types.ts
```diff
@@ -442,6 +445,28 @@ export class TypedEventEmitter<Events extends EventsDescription> extends EventEm
...
   }
+  /** @internal */
+  emitAndLogCommand<EventKey extends keyof Events>(
+    monitorCommands: boolean,
+    event: EventKey | symbol,
+    databaseName: string,
+    hello: Document | null,
+    ...args: Parameters<Events[EventKey]>
+  ): void {
+    if (monitorCommands) {
```
nbbeeken, line 456: If monitorCommands is true, and hello is falsey, then wouldn't that mean we would emit command started events for handshakes?

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,238 @@
...
+           the "reply" value is a string of length <= 1000 + (length of trailing ellipsis).
+
+        6. Run find() on the collection where the document was inserted.
+
+        7. Inspect the resulting "command succeeded" log message and assert that 
+           the reply is a string of length 1000 + (length of trailing ellipsis).
+    */
+
+    it('should follow default truncation limit of 1000', {
+      metadata: {},
```
nbbeeken, line 30: If there is no metadata then you don't need to specify this, might as well leave it out.

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      test: async function () {
+        const writable = {
+          buffer: [],
+          write(log) {
+            this.buffer.push(log);
+          }
+        };
+
+        // 1.
+        const client = this.configuration.newClient(
```
nbbeeken, line 40: If any of the assertions in this test fails, the client won't close and the tests won't exit. Strive to always use beforeEach/afterEach hooks to avoid this:

```
let client;
beforeEach(async function () {
  client = this.configuration.newClient();
});

afterEach(async function () {
  await client.close();
});
```

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,238 @@
...
+        await client.db('admin').collection('test').insertMany(docs);
+
+        // remove sensitive commands for uniformity
+        const cleanedBuffer = [];
+        const afterInsertBufferLength = writable.buffer.length;
+        for (let i = 0; i < afterInsertBufferLength; i++) {
+          if (writable.buffer[i].command !== '{}' && writable.buffer[i].reply !== '{}') {
+            cleanedBuffer.push(writable.buffer[i]);
+          }
+        }
```
nbbeeken, line 67: The prose test does mention needing to do this, why do we have to?

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,238 @@
...
+        // 5.
+        const insertManyCommandSucceeded = cleanedBuffer[1];
+        expect(insertManyCommandSucceeded?.message).to.equal('Command succeeded');
+        expect(insertManyCommandSucceeded?.reply).to.be.a('string');
+        expect(insertManyCommandSucceeded?.reply?.length).to.be.at.most(
+          DEFAULT_MAX_DOCUMENT_LENGTH + ELLIPSES_LENGTH
+        );
+
+        // 6.
+        await client.db('admin').collection('test').find()._initialize();
```
nbbeeken, line 86: `_initialize()` is a private method, you can't call that here, plus it is missing the required session argument. 

I think you wanted to call `toArray()`

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  context('Truncation with multi-byte codepoints', function () {
+    /*
+    A specific test case is not provided here due to the allowed variations in truncation logic
+     as well as varying extended JSON whitespace usage. 
+     Drivers MUST write language-specific tests that confirm truncation of commands, replies, 
+     and (if applicable) server responses included in error messages 
+     work as expected when the data being truncated includes multi-byte Unicode codepoints.
+     
+     If the driver uses anything other than Unicode codepoints as the unit for max document length, 
+     there also MUST be tests confirming that cases
```
nbbeeken, line 187: Not sure what may be going on with the linter in CI, but on my machine prettier has a bunch of fixes to this file, like trailing whitespace fixes.

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    5. (Optional)
+        If the driver attaches raw server responses to failures
+        and can access these via log messages to assert on, 
+        run the command {"notARealCommand": true}. 
+
+        Inspect the resulting "command failed" log message 
+        and confirm that the server error is a string of length 5 + (length of trailing ellipsis).
+    */
+    it('should follow custom truncation limit', {
+      metadata: {},
```
nbbeeken, line 129: rm  `metadata`

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      test: async function () {
+        const writable = {
+          buffer: [],
+          write(log) {
+            this.buffer.push(log);
+          }
+        };
+
+        // 1.
+        const client = this.configuration.newClient(
```
nbbeeken, line 139: move client to before/after each hooks

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,238 @@
...
+            [loggerFeatureFlag]: true,
+            mongodbLogPath: writable,
+            mongodbLogComponentSeverities: {
+              command: 'debug'
+            },
+            mongodbLogMaxDocumentLength: 50
+          }
+        );
+
+        const docs: Array<Document> = [{ x: '&&&&&&&&&&&&&&&&' }];
```
nbbeeken, line 212: Ampersands are single byte unicode code points and won't ever be possible to split incorrectly. 

> where the max length falls in the middle of a multi-byte codepoint are handled gracefully.

This is not obvious in javascript, at least with the APIs we have, let's discuss on slack.

test/unit/tools/unified_spec_runner.test.ts
```diff
@@ -167,7 +167,7 @@ describe('Unified Spec Runner', function () {
               level: 'debug',
               component: 'command',
               data: {
-                failure: {}
+                failure: '{}'
```
nbbeeken, line 170: I have a related question about this in the mongo_logger file, leaving a comment here to track.

nbbeeken (changes requested):

src/mongo_logger.ts
```diff
@@ -490,20 +524,20 @@ function defaultLogTransform(
...
       log.durationMS = logObject.duration;
-      log.reply = stringifyWithMaxLen(logObject.reply, maxDocumentLength);
+      log.reply = stringifyWithMaxLen(logObject.reply, maxDocumentLength, { relaxed: true });
       return log;
     case COMMAND_FAILED:
       log = attachCommandFields(log, logObject);
       log.message = 'Command failed';
       log.durationMS = logObject.duration;
-      log.failure = logObject.failure;
+      log.failure = logObject.failure.message ?? '{}';
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: `failure = {}` when commands are redacted. So `failure.message` will be null.

test/unit/tools/unified_spec_runner.test.ts
aditi-khare-mongoDB: The precedent with logging event failure properties thus far is to use message strings. I can replace this to the empty string or `'[REDACTED]'` if that's clearer, too.

src/mongo_logger.ts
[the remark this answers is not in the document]
nbbeeken: TY! missed that. The strange bit here is that when failure is not redacted it is going to be the string returned by the server, which is something descriptive. Would it be wrong to put something like `'(redacted)'` as the fallback string? Using `'{}'` would make more sense to me if the log.failure field was being set to logObject.failure, but it is the message, so it seems we could use a message-like string.

src/mongo_types.ts
aditi-khare-mongoDB: Yes. We only want to not log hello commands, emitting is fine. See [here](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.rst#log-messages).

src/cmap/connection.ts
aditi-khare-mongoDB: sounds good, we talked about this, but just for an added paper trail: checking the hello document exists is not enough truthiness for knowing the handshake has been completed. I introduced a `established` property to the `Connection` class, that is now used in `emitAndLogCommand`.

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: Sounds good, I've removed it

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: That's true, just added in the `beforeEach` and `afterEach` function

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: Paper trail: This was before the issue with the truthiness of the hello was resolved, it's been removed now.

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: Ah I see, sounds good!

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: Sounds good

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: done! TIA

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: I think the linter problem was a stray only / smg along those lines. My machine doesn't have any trailing whitespace fixes on this file.

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: Paper trail: changed ampersand to snowman emoji!

test/spec/command-logging-and-monitoring/logging/redacted-commands.json
```diff
@@ -14,7 +14,8 @@
         "useMultipleMongoses": false,
         "observeLogMessages": {
           "command": "debug"
-        }
+        },
+        "observeSensitiveCommands": true
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: No need for this anymore since the hello truthiness issue is resolved. Best to not edit specs :slightly_smiling_face:

test/tools/unified-spec-runner/runner.ts
```diff
@@ -232,9 +238,19 @@ async function runUnifiedTest(
...
-        const filteredTestClientLogs = expectedLogsForClient.ignoreMessages
-          ? filterLogs(expectedLogsForClient.ignoreMessages, testClient!.collectedLogs, entities)
+        let filteredTestClientLogs = expectedLogsForClient.ignoreMessages
+          ? filterExtraLogs(
+              expectedLogsForClient.ignoreMessages,
+              testClient!.collectedLogs,
+              entities
+            )
           : testClient!.collectedLogs;
+        if (!testClient!.observeSensitiveCommands) {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Yes, this should be removed. Removed it now!

src/mongo_logger.ts
```diff
@@ -1,12 +1,8 @@
-import { type Document, EJSON, type EJSONOptions } from 'bson';
+import { type Document, EJSON, type EJSONOptions, type ObjectId } from 'bson';
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good, changed it and added in EJSON exports from bson.ts as well.

src/cmap/connect.ts
```diff
@@ -186,6 +186,10 @@ async function performInitialHandshake(
       throw error;
     }
   }
+
+  // Connection establishment is socket creatio (tcp handshake, tls handshake, MongoDB handshake (saslStart, saslContinue))
+  // Once connection is established, command logging and monitoring can emit and/or log events (if either are enabled)
```
nbbeeken, line 191:
```suggestion
  // Connection establishment is socket creation (tcp handshake, tls handshake, MongoDB handshake (saslStart, saslContinue))
```

src/cmap/connection.ts
```diff
@@ -165,6 +168,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   public delayedTimeoutId: NodeJS.Timeout | null = null;
   public generation: number;
   public readonly description: Readonly<StreamDescription>;
+  public established: boolean;
```
nbbeeken, line 171: We discussed leaving a TSDoc comment here, we should include what it means and "who" sets it. I think the comment inside `connect()` mostly fits well here.

src/mongo_types.ts
nbbeeken: Is this condition still necessary with the "established" condition & fix?

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
nbbeeken: It is still here

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,228 @@
...
+describe('Command Logging and Monitoring Prose Tests', function () {
+  const loggerFeatureFlag = Symbol.for('@@mdb.enableMongoLogger');
+  const ELLIPSES_LENGTH = 3;
+  let client;
+  let writable;
+
+  context('When no custom truncation limit is provided', function () {
+    /*
+        1. Configure logging with a minimum severity level of "debug" for the "command" component. 
+           Do not explicitly configure the max document length.
```
nbbeeken, line 14: This line has trailing whitespace, and there are other lines as well, double check prettier has run on this file.

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
nbbeeken: The new test still does not prove the the point we are trying to assert. We still use .slice which will leave a string with a dangling surrogate. Snowman was an unlucky choice 😅 it is a very old emoji. Try: `'\uD801\uDC37'`

nbbeeken (changes requested):

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: Ah there are some trailing whitespaces, fixed them now, my linter didn't catch any of them originally unfortunately. Good catch.

src/cmap/connection.ts
aditi-khare-mongoDB: Sounds good, I've updated the TSDoc now

test/integration/command-logging-and-monitoring/command_logging_and_monitoring.prose.test.ts
aditi-khare-mongoDB: Should be removed now!

src/cmap/message_stream.ts
```diff
@@ -0,0 +1,220 @@
...
+  zlibCompressionLevel?: number;
+  $clusterTime?: Document;
+}
+
+/**
+ * A duplex stream that is capable of reading and writing raw wire protocol messages, with
+ * support for optional compression
+ * @internal
+ */
+export class MessageStream extends Duplex {
```
nbbeeken, line 48: I think MessageStream should stay deleted 😅 (rebase issue?)

src/mongo_types.ts
nbbeeken: Ah ok, because this function can be entered if monitoring OR logging is enabled, we need to filter inside here. fair enough.

nbbeeken (changes requested):

src/cmap/message_stream.ts
aditi-khare-mongoDB: I imagine so.. 😅

baileympearson (changes requested):

package-lock.json
```diff
@@ -799,17 +799,89 @@
       }
     },
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: I believe I fixed this in a previous commit!

src/cmap/connection.ts
```diff
@@ -174,6 +177,10 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   private messageStream: Readable;
   private socketWrite: (buffer: Uint8Array) => Promise<void>;
   private clusterTime: Document | null = null;
+  /** @internal  */
+  override component = MongoLoggableComponent.COMMAND;
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: No, there won't. The extra this.component check should ensure that and emitAndLog or variant function (emitAndLogCommand, emitAndLogHeartbeat) is not called from any class for which the logging component is not defined already.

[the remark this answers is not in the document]
aditi-khare-mongoDB: I can add in an extra check such as: 'this.component === MongoLoggableComponent.COMMAND,' remove the this.component check altogether, or keep it the same. What do you think?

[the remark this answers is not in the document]
baileympearson: It's strange to me that we have a dedicated method for logging commands, but to use it we also need to set the component on the class. I would expect that either:

- we only have `emitAndLogCommand()`, and not the `component` property.
- we have a generic `emitAndLog()` method that uses the `component` property and no specialized methods on the TypedEventEmitter

> No, there won't. The extra this.component check should ensure that an emitAndLog or variant function (emitAndLogCommand, emitAndLogHeartbeat) is not called from any class for which the logging component is not defined already.

If we want to prevent the emitAndLog functions from being called, maybe we could use inheritance for this?

```
class TypedEventEmitter {
  private logger: MongoLogger;
}

class CommandEventEmitter extends TypedEventEmitter {
  emitAndLogCommand(..) {}
}


class HeartbeatEventEmitter extends TypedEventEmitter {
  emitAndLogHeartbeat(..) {}
}

class Connection extends CommandEventEmitter { ...} 
class Monitor extends HeartbeatEventEmitter { ...} 
```

src/mongo_logger.ts
```diff
@@ -383,7 +413,8 @@ export function stringifyWithMaxLen(
   maxDocumentLength: number,
   options: EJSONOptions = {}
 ): string {
-  let strToTruncate: string;
+  let strToTruncate = '';
```
addaleax, line 416: Non-blocking and definitely only tangentially related: the `.slice()` call below can end up truncating `strToTruncate` in the middle of a character. That won't happen a lot if the string mostly just contains ASCII data, but since this is a very low-level generic utility, it might be worth accounting for that case (e.g. `if (strToTruncate.length > maxDocumentLength && strToTruncate.charCodeAt(maxDocumentLength-1) !== strToTruncate.codePointAt(maxDocumentLength-1)) maxDocumentLength--;`).

src/cmap/connection.ts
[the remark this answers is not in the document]
aditi-khare-mongoDB: That's true. For the sake of this PR, I'll just hardcode in the component in `emitAndLogCommand` as `MongoLoggableComponent.COMMAND.` Since I realize `emitAndLogCommand` already implies that the command component is being utilized.

src/mongo_logger.ts
aditi-khare-mongoDB: This is something we will solve in [NODE-5839](https://jira.mongodb.org/browse/NODE-5839 ).

nbbeeken (dismissed):

src/cmap/connection.ts
```diff
@@ -165,6 +168,15 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
   public readonly description: Readonly<StreamDescription>;
+  /**
+   * @public
+   * Represents if the connection has been established:
+   *  - TCP handshake
+   *  - TLS negotiated
+   *  - mongodb handshake (saslStart, saslContinue), includes authentication
+   * Once connection is established, command logging can log events (if enabled)
+   */
+  public established: boolean;
```
baileympearson, line 179: What is the expected behavior with command logging when we re-authenticate?

baileympearson: Reauthentication will send commands we don't want to log I think.  We may either need to 

- add logic that more specifically determines whether or not a command should be logged
- set `established` to false when reauthenticating and then set it back to true after a successful reauthentication

baileympearson: Sorry I didn't notice this earlier ( I didn't actually check ) but instead of modifying `established` when we reauthenticate, could we just check the `context.reauthenticating` field?

one idea is a getter on the connection class (or we could just inline it):
```typescript
get shouldEmitAndLog(): boolean {
  return this.monitorCommands ||
      (this.established && !this.context.reauthenticating && this.mongoLogger?.willLog(SeverityLevel.DEBUG, MongoLoggableComponent.COMMAND))
}
```

aditi-khare-mongoDB: Isn't it more accurate to edit established as well, though? Since during authentication, the connection is not established.

baileympearson: I understood established to mean that the connection had been established, not that it was "ready"/currently established.  What I would like to avoid is duplicate data that might get out of sync.  If established=false can also represent a connection reauthenticating, then it's possible for us to end up in a scenario context.reauthenticating=true with established=false (or vice versa) - which is a poorly defined state for the driver.

One way to avoid this is to define established as whether or not the connection was initially established and rely on context.reauthenticating (my suggestion).

aditi-khare-mongoDB: That sounds good. I've removed the extra auth changes to `conn.established` and am using `this.authContext?.authenticating` in the new getter `shouldEmitAndLogCommand` instead.

baileympearson (approved):

nbbeeken (approved):

PR #3959, opened 2024-01-08
refactor(NODE-5809): remove executeOperation callback overload and maybeCallback

nbbeeken:
#### What is changing?

- Move .then/.catch chain into callers
- Remove maybeCallback
- Improve docs over executeOperation

No.

This change is effectively a no-op, but it removes the overloads advertized by executeOperation that are no longer used making future changes to executeOperation easier to make.

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -579,23 +579,18 @@ function executeCommands(
   }
 
   try {
-    if (isInsertBatch(batch)) {
-      executeOperation(
-        bulkOperation.s.collection.client,
-        new InsertOperation(bulkOperation.s.namespace, batch.operations, finalOptions),
-        resultHandler
-      );
-    } else if (isUpdateBatch(batch)) {
-      executeOperation(
-        bulkOperation.s.collection.client,
-        new UpdateOperation(bulkOperation.s.namespace, batch.operations, finalOptions),
-        resultHandler
-      );
-    } else if (isDeleteBatch(batch)) {
-      executeOperation(
-        bulkOperation.s.collection.client,
-        new DeleteOperation(bulkOperation.s.namespace, batch.operations, finalOptions),
-        resultHandler
+    const operation = isInsertBatch(batch)
+      ? new InsertOperation(bulkOperation.s.namespace, batch.operations, finalOptions)
+      : isUpdateBatch(batch)
+      ? new UpdateOperation(bulkOperation.s.namespace, batch.operations, finalOptions)
+      : isDeleteBatch(batch)
+      ? new DeleteOperation(bulkOperation.s.namespace, batch.operations, finalOptions)
+      : null;
+
+    if (operation != null) {
+      executeOperation(bulkOperation.s.collection.client, operation).then(
+        result => resultHandler(undefined, result),
+        error => resultHandler(error)
       );
     }
   } catch (err) {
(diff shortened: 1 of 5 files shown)


src/bulk/common.ts
```diff
@@ -579,23 +579,18 @@ function executeCommands(
...
-        bulkOperation.s.collection.client,
-        new DeleteOperation(bulkOperation.s.namespace, batch.operations, finalOptions),
-        resultHandler
+    const operation = isInsertBatch(batch)
+      ? new InsertOperation(bulkOperation.s.namespace, batch.operations, finalOptions)
+      : isUpdateBatch(batch)
+      ? new UpdateOperation(bulkOperation.s.namespace, batch.operations, finalOptions)
+      : isDeleteBatch(batch)
+      ? new DeleteOperation(bulkOperation.s.namespace, batch.operations, finalOptions)
+      : null;
```
baileympearson, line 588: It's not a case we currently handle - but what happens if we can't find a batch?  In theory that's an impossible state.  Should we throw a `MongoRuntimeError` if we are unable to create an operation to execute?

src/operations/execute_operation.ts
```diff
@@ -51,36 +51,24 @@ export interface ExecutionResult {
...
- * it allows for a single point of entry to provide features such as implicit sessions, which
+ * Allows for a single point of entry to provide features such as implicit sessions, which
  * are required by the Driver Sessions specification in the event that a ClientSession is
- * not provided
+ * not provided.
  *
- * @param topology - The topology to execute this operation on
+ * The expectation is that this function:
+ * - Connects the MongoClient if it has not already been connected
+ *   - This amounts to creating and starting the Monitors, something that serverSelection could also do
```
baileympearson, line 60:
```suggestion
```

nbbeeken: Fair enough, wishful thinking 😅 I did want to capture that it is not a strict requirement to call "connect" but rather start the logic of connecting... anyway too much for implication something I can save for how CSOT comes together.

src/bulk/common.ts
nbbeeken: I considered that but I decided against making functional changes. I agree it is an impossible state but do we know for sure that the "empty" code path is unused? if it is I'm not sure anything would crash, I can make a ticket to follow up on it.

baileympearson: I think it's a relatively safe assumption - `batch` is defined above as:

```typescript
    return callback(
      undefined,
      new BulkWriteResult(bulkOperation.s.bulkResult, bulkOperation.isOrdered)
    );
  }
  const batch = bulkOperation.s.batches.shift() as Batch;
```

so batch will never be empty

nbbeeken: https://jira.mongodb.org/browse/NODE-5813

nbbeeken: It looks like addToOperationsList does not validate the BatchType, so it could be input wrong (even though TS would error on it) and then at runtime this logic would silently skip. I don't think that is correct but it is worthy of a thorough look before it changes.

W-A-James (approved):

W-A-James:
Failing tests are known to be flaky or unrelated

alenakhineika (approved):

PR #3960, opened 2024-01-08
test(NODE-5724): heartbeat started is emitted before socket is connected

W-A-James:
#### What is changing?
Implement spec test added in DRIVERS-2677

No

NODE-5724

### DIFF
--- a/test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
+++ b/test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
@@ -14,9 +14,12 @@ describe('Heartbeat tests', function () {
     // Create TCP server that responds to hellos by closing the connection
     // and pushing "client connection created" to shared array
     server = createServer(clientSocket => {
-      console.log(`connection from ${clientSocket.remoteAddress}`);
       events.push('client connected');
-      clientSocket.end();
+
+      clientSocket.once('data', () => {
+        events.push('client hello received');
+        clientSocket.end();
+      });
     });
     server.listen(9999);
 
@@ -50,10 +53,11 @@ describe('Heartbeat tests', function () {
     if (server.listening) server.close();
   });
 
-  it('emits the first HeartbeatStartedEvent after the monitoring socket was created', async function () {
-    expect(events).to.have.lengthOf(3);
+  it('emits the first HeartbeatStartedEvent after the monitoring socket was created and before hello is sent', async function () {
+    expect(events).to.have.lengthOf(4);
     expect(events[0]).to.equal('client connection created');
     expect(events[1]).to.equal(SERVER_HEARTBEAT_STARTED);
+    expect(events[2]).to.equal('client hello received');
     expect(events[3]).to.equal(SERVER_HEARTBEAT_FAILED);
   });
 });
--- a/test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
+++ b/test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
@@ -1,4 +1,5 @@
 import { expect } from 'chai';
+import { once } from 'events';
 import { createServer, type Server } from 'net';
 
 import { MongoClient, SERVER_HEARTBEAT_FAILED, SERVER_HEARTBEAT_STARTED } from '../../mongodb';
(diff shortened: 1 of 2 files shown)


test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    // and pushing "client connection created" to shared array
+    server = createServer(clientSocket => {
+      events.push('client connected');
+
+      clientSocket.once('data', () => {
+        events.push('client hello received');
+        clientSocket.end();
+      });
+    });
+    server.listen(9999);
```
nbbeeken, line 24: Can we make a promise out of the [`'listening'`](https://nodejs.org/docs/latest/api/net.html#event-listening) event and await it? I think we're "getting lucky" that the server has bound to the 9999 port because we await client.connect()

test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Catch error
+    expect(maybeError).to.be.instanceOf(Error);
+  });
+
+  afterEach(async function () {
+    if (server.listening) server.close();
+  });
+
+  it('emits the first HeartbeatStartedEvent after the monitoring socket was created and before hello is sent', async function () {
+    expect(events).to.have.lengthOf(4);
```
nbbeeken, line 57: Similar to my comment on the spec PR, could we just deep eq check a hardcoded array? then the assertion will print out the whole list when it is wrong

test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    for (const e of [SERVER_HEARTBEAT_STARTED, SERVER_HEARTBEAT_FAILED]) {
+      client.on(e, () => {
+        events.push(e);
+      });
+    }
+
+    // Attempt to connect to mock server
+    const maybeError = await client.connect().catch(e => e);
+    // Catch error
+    expect(maybeError).to.be.instanceOf(Error);
```
nbbeeken, line 49: This can be a part of the `it` block right? That way the test fails rather than the setup hook

test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  beforeEach(async function () {
+    // Create TCP server that responds to hellos by closing the connection
+    // and pushing "client connection created" to shared array
+    server = createServer(clientSocket => {
+      events.push('client connected');
+
+      clientSocket.once('data', () => {
+        events.push('client hello received');
+        clientSocket.end();
```
nbbeeken, line 21:
```suggestion
        clientSocket.destroy();
```

This is a nit, and mostly opinion: if you aren't sending a final chunk and you aren't using the callback that `.end` accepts then I would reach for the API that "Ensures that no more I/O activity happens on this socket." vs the one that "Half-closes the socket. i.e., it sends a FIN packet. It is possible the server will still send some data."

nbbeeken (commented):
Looks great! a few clean up ideas

W-A-James:
Note: Bug ticket NODE-5840 filed to address incorrect heartbeat duration measurement

test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
nbbeeken: Bumping this (sorta) any reason to check length now that we deep eq? 

just thinking of the future engineer peeking at EVG logs, the deep eq check will fail if the length is wrong but it will print out the contents

nbbeeken (changes requested):

test/unit/sdam/server_discovery_and_monitoring.prose.test.ts
W-A-James: Oh my bad, I forgot to remove that when adding the deep equal check.

nbbeeken (approved):

nbbeeken:
Will merge soon, just double checking some new setup failures that I think are transient

PR #3963, opened 2024-01-12
feat(NODE-3470): retry selects another mongos

durran:
### Description

When retrying reads and writes on a sharded cluster, will attempt the retry on a different mongos if multiple mongoses are present.

#### What is changing?
- `executeOperation` now keeps track of the previously selected server.
- Server selection will now attempt to exclude the previous server when the topology is sharded and multiple mongoses are present.
- Adds unit tests for server selection and prose tests for retryable reads and writes.

None

NODE-3470

### DIFF
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -247,7 +247,7 @@ async function retryOperation<
   const server = await topology.selectServerAsync(selector, {
     session,
     operationName: operation.commandName,
-    previousServer: previousServer
+    previousServer
   });
 
   if (isWriteOperation && !supportsRetryableWrites(server)) {
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -18,6 +18,7 @@ import {
 import type { MongoClient } from '../mongo_client';
 import { ReadPreference } from '../read_preference';
 import type { Server } from '../sdam/server';
+import type { ServerDescription } from '../sdam/server_description';
 import {
   sameServerSelector,
   secondaryWritableServerSelector,
@@ -183,7 +184,8 @@ export async function executeOperation<
       return await retryOperation(operation, operationError, {
         session,
         topology,
-        selector
+        selector,
+        previousServer: server.description
       });
     }
     throw operationError;
@@ -199,6 +201,7 @@ type RetryOptions = {
   session: ClientSession;
   topology: Topology;
   selector: ReadPreference | ServerSelector;
+  previousServer: ServerDescription;
 };
 
 async function retryOperation<
@@ -207,7 +210,7 @@ async function retryOperation<
 >(
   operation: T,
   originalError: MongoError,
-  { session, topology, selector }: RetryOptions
+  { session, topology, selector, previousServer }: RetryOptions
 ): Promise<TResult> {
   const isWriteOperation = operation.hasAspect(Aspect.WRITE_OPERATION);
   const isReadOperation = operation.hasAspect(Aspect.READ_OPERATION);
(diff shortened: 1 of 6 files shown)


src/operations/execute_operation.ts
```diff
@@ -243,7 +246,8 @@ async function retryOperation<
   // select a new server, and attempt to retry the operation
   const server = await topology.selectServerAsync(selector, {
     session,
-    operationName: operation.commandName
+    operationName: operation.commandName,
+    previousServer: previousServer
```
alenakhineika, line 250:
```suggestion
    previousServer
```

src/sdam/server_selection.ts
```diff
@@ -287,13 +289,15 @@ export function readPreferenceServerSelector(readPreference: ReadPreference): Se
...
-    if (
-      topologyDescription.type === TopologyType.Single ||
-      topologyDescription.type === TopologyType.Sharded
-    ) {
+    if (topologyDescription.type === TopologyType.Single) {
       return latencyWindowReducer(topologyDescription, servers.filter(knownFilter));
     }
 
+    if (topologyDescription.type === TopologyType.Sharded) {
+      const selectable = servers.length > 0 ? servers : deprioritized;
```
alenakhineika, line 297: But wouldn't `servers` here contain also deprioritized ones? Do we filter out them somewhere before that?

src/sdam/topology.ts
```diff
@@ -930,8 +933,13 @@ function processWaitQueue(topology: Topology) {
     let selectedDescriptions;
     try {
       const serverSelector = waitQueueMember.serverSelector;
+      const previousServer = waitQueueMember.previousServer;
       selectedDescriptions = serverSelector
-        ? serverSelector(topology.description, serverDescriptions)
+        ? serverSelector(
+            topology.description,
+            serverDescriptions,
+            previousServer ? [previousServer] : []
```
alenakhineika, line 941: Do we basically set the `deprioritized` value here? Does it mean that we encounter only the most recent previousServer, not all that we tried and havend't succeeded before?

src/sdam/server_selection.ts
durran: Good catch. I've updated to filter them out now.

src/operations/execute_operation.ts
durran: Updated.

src/sdam/topology.ts
durran: Yes currently we only encounter the most recent previous server, as we only retry once right now. Once CSOT is implemented, then we update this to push each previous previous server to the deprioritised list since we then can retry multiple times. I made this an array in anticipation of the CSOT work even though right now it will contain only 1.

alenakhineika: Got it! Thx for the detailed explanation ✨

alenakhineika (approved):

src/sdam/topology.ts
```diff
@@ -175,6 +176,7 @@ export interface SelectServerOptions {
   serverSelectionTimeoutMS?: number;
   session?: ClientSession;
   operationName: string;
+  previousServer?: ServerDescription;
```
baileympearson, line 179: (non-blocking) 

the name `previousServer` is odd when we're performing server selection for the first time, since there's no previous server that has been selected.  what would you think of the more general name, `serversToIgnore: ServerDescription[]`?  We would set it to `[]` for non-retries.

durran: I chose `previousServer` since that matches the naming in the spec itself, and figured for clarity it would be fine to keep that name:  https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#pseudocode

baileympearson (approved):

alenakhineika:
Flaky tests are reported in Jira: [NODE-5803](https://jira.mongodb.org/browse/NODE-5803), [NODE-5845](https://jira.mongodb.org/browse/NODE-5845)

PR #3966, opened 2024-01-17
test(NODE-5823): CSOT unified runner changes

durran:
### Description

Implements unified test runner changes for the CSOT unified tests to run.

#### What is changing?

Implements the following operations:
- `listDatabaseNames`
- `listCollectionNames`
- `count`
- `listIndexNames`
- `dropIndexes`
- `drop`

Updates `find` to work on both a collection or gridfs bucket.

Patch for enabled CSOT unified tests on all topologies: https://spruce.mongodb.com/version/65a7fe30e3c33167dda24b5d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

None

NODE-5823

### DIFF
--- a/src/connection_string.ts
+++ b/src/connection_string.ts
@@ -1090,6 +1090,9 @@ export const OPTIONS = {
     target: 'tls',
     type: 'boolean'
   },
+  timeoutMS: {
+    type: 'uint'
+  },
   tls: {
     type: 'boolean'
   },
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -118,7 +118,7 @@ export type SupportedNodeConnectionOptions = SupportedTLSConnectionOptions &
 export interface MongoClientOptions extends BSONSerializeOptions, SupportedNodeConnectionOptions {
   /** Specifies the name of the replica set, if the mongod is a member of a replica set. */
   replicaSet?: string;
-  /** @experimental */
+  /** @internal This option is in development and currently has no behaviour. */
   timeoutMS?: number;
   /** Enables or disables TLS/SSL for the connection. */
   tls?: boolean;
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -118,6 +118,8 @@ export type SupportedNodeConnectionOptions = SupportedTLSConnectionOptions &
 export interface MongoClientOptions extends BSONSerializeOptions, SupportedNodeConnectionOptions {
   /** Specifies the name of the replica set, if the mongod is a member of a replica set. */
   replicaSet?: string;
+  /** @experimental */
+  timeoutMS?: number;
   /** Enables or disables TLS/SSL for the connection. */
   tls?: boolean;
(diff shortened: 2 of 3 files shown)


src/mongo_client.ts
```diff
@@ -118,6 +118,8 @@ export type SupportedNodeConnectionOptions = SupportedTLSConnectionOptions &
 export interface MongoClientOptions extends BSONSerializeOptions, SupportedNodeConnectionOptions {
   /** Specifies the name of the replica set, if the mongod is a member of a replica set. */
   replicaSet?: string;
+  /** @experimental */
+  timeoutMS?: number;
```
nbbeeken, line 122: We could leave this marked internal until it is ready, the tests should rely on the driver `src` type definitions rather than the public-only generated one. 

But this is fine by me, I would add "this option is in development and currently has undefined behavior" or something, because the driver may have partial support as we work to implement it. Thoughts?

test/tools/unified-spec-runner/operations.ts
```diff
@@ -680,6 +714,12 @@ operations.set('withTransaction', async ({ entities, operation, client, testConf
   }, options);
 });
 
+operations.set('count', async ({ entities, operation }) => {
```
nbbeeken, line 717: Is the count operation missing from the unified test format document? I was trying to confirm if this was supposed to be countDocument/estimatedCount or what you have. Pretty sure you've got the right API but where do we check?

src/connection_string.ts
```diff
@@ -1090,6 +1090,9 @@ export const OPTIONS = {
     target: 'tls',
     type: 'boolean'
   },
+  timeoutMS: {
+    type: 'uint'
+  },
```
nbbeeken, line 1095: Also fine with this, just bus factor++ we may want to assert more about this value than just unsigned int. We're building this functionality on `setTimeout` which has an upper limit.

```txt
(node:65848) TimeoutOverflowWarning: 2147483648 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
```
Node decides this is a warning which will manifest as a timeout of 1ms but potentially we want to consider this an error before even creating the timeout.

nbbeeken (changes requested):

test/tools/unified-spec-runner/operations.ts
durran: This was an odd one, but the fail command in the tests was on the `count` command so I figured they wanted the deprecated one. Eg: https://github.com/mongodb/node-mongodb-native/blob/main/test/spec/client-side-operations-timeout/retryability-timeoutMS.yml#L1893

src/mongo_client.ts
durran: I've updated to make internal and added a comment.

src/connection_string.ts
durran: I think we can leave that for later when we actually implement it. This was just to get through the unified tests.

PR #3968, opened 2024-01-17
fix(NODE-5723): emit TopologyDescriptionChangedEvent just before close

W-A-James:
#### What is changing?

- Unskip Topology life cycle tests
- Remove modified logging spec tests
- Pull in updated `logging-loadbalancer` tests
- Pull in new monitoring tests
- Fix bug in test runner that was exiting on encountering a single SDAM event
- Emit `TopologyDescriptionChangedEvent` transitioning to `Unknown` Topology type just before close

No

NODE-5723 / DRIVERS-2711


W-A-James:
Note: getting spec test failures, but they disappear when _only_ running the tests in question.

test/manual/socks5.test.ts
```diff
@@ -302,7 +302,7 @@ describe('Socks5 Connectivity', function () {
 async function testConnection(connectionString, clientOptions) {
   const client = new MongoClient(connectionString, clientOptions);
   let topologyType;
-  client.on('topologyDescriptionChanged', ev => (topologyType = ev.newDescription.type));
+  client.on('topologyDescriptionChanged', ev => (topologyType = ev.previousDescription.type));
```
W-A-James, line 305: This function was previously relying on the fact that the last `TopologyDescroptionChangedEvent` would be the one transitioning the topology to the fully known, desired operating state, with the description being in the `newDescription` field.

Now that we emit a final instance of that event transitioning to the `UNKNOWN` state, this final event would have the desired topology description in the `previousDescription` field, rather than the `newDescription` field.

W-A-James:
Note: this branch was rebased against #4089 which implements the UTR changes introduced in DRIVERS-2875

src/sdam/topology.ts
```diff
@@ -511,6 +513,13 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
     this.removeListener(Topology.TOPOLOGY_DESCRIPTION_CHANGED, this.s.detectShardedTopology);
 
     stateTransition(this, STATE_CLOSED);
+    this.s.description = new TopologyDescription(TopologyType.Unknown, new Map());
```
nbbeeken, line 516:
```suggestion
    this.s.description = new TopologyDescription(
        TopologyType.Unknown,
        new Map(),
        options.replicaSet,
        undefined,
        undefined,
        undefined,
        options
      );
```

Not sure if this is right, but feels like it would make sense to set it to unknown but pipe through the options that made the original starting `TopologyDescription`. Thoughts?

nbbeeken (changes requested):
code lgtm, one small question. 

You are calling this a fix, was the previous event emission technically "wrong"? can you make your release notes describe the change in terms of what it is fixing?

src/sdam/topology.ts
W-A-James: I'm not against this and it wouldn't go against the spec, so I'm happy to implement this

nbbeeken (changes requested):
From slack: 

> missed a TODO skip message to run some of the tests ... that are failing because of an extra event emission

Just keeping this here for viz.

dariakp:
Closing this for now because it is stale. We can reopen when we are ready to work on this.

PR #3970, opened 2024-01-19
feat(NODE-4847): Add config error handling to logging

aditi-khare-mongoDB:
### Description
Essentially, these changes ensure that the logger does not crash the user's application at runtime, at any point after options are parsed. 

During options parsing, if the argument passed in is of incorrect type we want to throw for client options, but not for an incorrect environment option (set to default in this case).

#### What is changing?
To ensure we don't crash the user's application through the logger, unless there is an invalid type client option passed in to the `MongoClient / ConnectionString`.

Invalid argument types are checked through the `transform` function for an element of the `OPTIONS` array. The `OPTIONS` array is used in parseOptions (which is called by `MongoClient` and `ConnectionString`), so it resolves the argument types for both classes. 

See all unit tests defined under config error handling in [Kickoff Document](https://docs.google.com/document/d/1gDJ9vpVdH7BalRxEvzC_UZz-gIYOLSTU7HTqSoa18FA/edit#heading=h.7t9v3ahij9a1), 

After NODE-5672 (Release Logger)

This ticket is a subtask of [NODE-4816](https://jira.mongodb.org/browse/NODE-4816) (Add Error Handling to the Logger).


src/connection_string.ts
```diff
@@ -1243,12 +1245,53 @@ export const OPTIONS = {
...
-  mongodbLogPath: { type: 'any' },
+  mongodbLogPath: {
+    transform({ values: [value] }) {
+      if (
+        !(
+          (typeof value === 'string' && ['stderr', 'stdout'].includes(value)) ||
+          (value &&
+            typeof value === 'object' &&
+            Object.keys(value).indexOf('write') >= 0 &&
+            typeof Object.values(value)[Object.keys(value).indexOf('write')] === 'function')
```
alenakhineika, line 1256: If `value` is an object, could we just check if `typeof value === 'object' && typeof value?.write === 'function'`?

alenakhineika:
(housekeeping): double check the checkboxes in the pr description

src/connection_string.ts
```diff
@@ -1243,12 +1245,53 @@ export const OPTIONS = {
    * @internal
    * TODO: NODE-5671 - remove internal flag
    */
-  mongodbLogPath: { type: 'any' },
+  mongodbLogPath: {
+    transform({ values: [value] }) {
+      if (
+        !(
+          (typeof value === 'string' && ['stderr', 'stdout'].includes(value)) ||
```
alenakhineika, line 1252: When `mongodbLogPath = string that is not 'stdout' | 'stderr'` shouldn't we `default to stderr` instead of throwing `MongoAPIError`.

Do we have MONGODB_LOG_PATH at this point already? We could also check for `mongodbLogPath=undefined` and `default to MONGODB_LOG_PATH value ?? stderr`.

And do I understand correctly that the rest of the cases like `Some point after construction, user-defined stream crashes` and  `mongodbLogPath= 'stdout', but stdout is closed`, etc will be handeled by another sub-task?

src/connection_string.ts
```diff
@@ -1243,12 +1245,53 @@ export const OPTIONS = {
...
+        }
+        if (!Object.values(SeverityLevel).some(val => val === v)) {
+          throw new MongoAPIError(
+            `Option 'mongodbLogComponentSeverities' does not support ${v} as a value for ${k}`
+          );
+        }
+      }
+      return value;
+    }
+  },
```
alenakhineika, line 1294: The connection_string.ts module handles only URI parsing, right? Where do we ensure the same behavior applies to MongoClientOptions? I thought this PR intends to prevent throwing when possible both in URI and options.

test/unit/connection_string.test.ts
```diff
@@ -912,13 +912,11 @@ describe('Connection String', function () {
...
-          [loggerFeatureFlag]: true
-        });
-        const log: Log = { t: new Date(), c: 'ConnectionStringInvalidOption', s: 'error' };
-        client.options.mongoLoggerOptions.logDestination.write(log);
-        expect(stderrStub.write).calledWith(inspect(log, { breakLength: Infinity, compact: true }));
+      it('should throw error at construction', function () {
+        expect(
+          () =>
+            new MongoClient('mongodb://a/?mongodbLogPath=stdnothing', { [loggerFeatureFlag]: true })
+        ).to.throw(MongoAPIError);
```
alenakhineika, line 919: Should we add more test cases here to check all edge cases for mongodbLogPath and mongodbLogComponentSeverities?

alenakhineika:
Could we be more specific in the PR title and description, and maybe AC to the ticket about what we expect from this PR? The current description is mostly about the parent ticket and also `"Most of the mongodbLogPath stream error handling must be done at runtime, so is in a different subtask."`. It is unclear what we are aiming to handle as part of this PR. Can we keep the link to the parent ticket in the `What is the motivation for this change?` section and put into `Description` and `What is changing?` more relevant to the current PR details? From the code, I see only changes made to ConnectionString options to validate those for incorrect values. Based on the kick-off doc, the same edge cases should be already handled by `MongoClient.parseOptions()` for MongoClientOptions. But I don't see the logic described in "Logging Error Handling Mini Design".

src/connection_string.ts
aditi-khare-mongoDB: We definitely can, but that exact approach throws a TS error (`TS: 2339`). I had to add in this unnecessarily long code to circumvent a TS error, but I've found a easier to look at workaround:

```
(
   value &&
  typeof value === 'object' &&
  'write' in value &&
  typeof value.write === 'function'
)
```

(I updated the code too!)

src/connection_string.ts
aditi-khare-mongoDB: The `OPTIONS` array that we are editing here is called in `parseOptions`, which is called by both `MongoClient` and `ConnectionString` when parsing options. `ConnectionString` can actually not support an object-like input to mongodbLogPath, and that check is only for `MongoClient`.

test/unit/connection_string.test.ts
aditi-khare-mongoDB: Which edge cases would you like to check?

I should have included this in my PR description, but if you expand this test file and the `mongo_client.test.js` test file, you can view some other test cases for the client options added in as part of the original logging client options implementation in [NODE-4849](https://jira.mongodb.org/browse/NODE-4849) and [NODE-4878](https://jira.mongodb.org/browse/NODE-4878).

src/connection_string.ts
aditi-khare-mongoDB: I apologize for the confusion with the test cases table. I've updated it for clarity and correctness. See [here](https://docs.google.com/document/d/1gDJ9vpVdH7BalRxEvzC_UZz-gIYOLSTU7HTqSoa18FA/edit#heading=h.797de2n1adjo).

The specified type of `mongodbLogPath: 'stderr' | 'stdout' | MongoDBLogWritable`, so we want to throw a type error when the type differs.

aditi-khare-mongoDB:
Updated the PR description and Logging Mini Design!

src/connection_string.ts
alenakhineika: We could also do `typeof value === 'object' && typeof (value as MongoDBLogWritable)?.write === 'function'`.

test/unit/connection_string.test.ts
alenakhineika: This comment is not relevant anymore after you clarified the test cases in the kick-off doc. Thx for the detailed answers, helpful to gather the full context of the problem!

src/connection_string.ts
aditi-khare-mongoDB: That is a shorter and more descriptive solution, thanks! I've updated the code again.

nbbeeken: I'm curious, is there a motivation for preferring casting away the TS error instead of writing the typescript code that correctly narrows the type? 

I often see the `in` check used so I don't think it something to be worked around

test/unit/mongo_client.test.js
```diff
@@ -976,35 +1019,121 @@ describe('MongoOptions', function () {
...
+            expect(
+              () =>
+                new MongoClient('mongodb://a/', {
+                  [loggerFeatureFlag]: true,
+                  mongodbLogMaxDocumentLength: -290
+                })
+            ).to.throw(MongoParseError);
+          });
+        });
+        context('when env option for MONGODB_LOG_MAX_DOCUMENT_LENGTH is provided', function () {
```
alenakhineika, line 1119: Should we add test cases when MONGODB_LOG_MAX_DOCUMENT_LENGTH is provided and mongodbLogMaxDocumentLength is negative, and also when MONGODB_LOG_MAX_DOCUMENT_LENGTH is provided but negative?

src/connection_string.ts
alenakhineika: We talked with Neal over Slack and agreed with him :) Sorry Aditi for the confusion, could you please change code accordingly to not cast it to MongoDBLogWritable. I was relying on the type validation of MongoClientOptions, but if users write JS then we can not be sure what object gets to MongoDBLogWritable. Can we also create a follow up ticket to clean up other casting that happens in the same file?

test/unit/mongo_client.test.js
```diff
@@ -976,35 +1019,121 @@ describe('MongoOptions', function () {
...
+              mongodbLogComponentSeverities: {
+                default: 'imFake'
+              }
+            });
+
+          context('when valid environment option is provided', function () {
+            it('should still throw error at construction', function () {
+              process.env['MONGODB_LOG_ALL'] = 'emergency';
+              expect(badClientCreator).to.throw(MongoAPIError);
+              process.env['MONGODB_LOG_ALL'] = 'undefined';
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Yes, sounds good!

test/unit/mongo_client.test.js
aditi-khare-mongoDB: Good idea, just added them in

src/connection_string.ts
aditi-khare-mongoDB: Sounds good, I've changed it back to the `in` check.

alenakhineika (dismissed):

test/unit/mongo_client.test.js
```diff
@@ -976,35 +1019,162 @@ describe('MongoOptions', function () {
...
+          it('should store value for default maxDocumentLength correctly', function () {
+            const client = new MongoClient('mongodb://a/', {
+              [loggerFeatureFlag]: true
+            });
+            expect(client.options.mongoLoggerOptions.maxDocumentLength).to.equal(1000);
+          });
+        });
+        context('when env option for MONGODB_LOG_MAX_DOCUMENT_LENGTH is provided', function () {
+          afterEach(function () {
+            process.env['MONGODB_LOG_MAX_DOCUMENT_LENGTH'] = undefined;
```
nbbeeken, line 1160:
```suggestion
            delete process.env.MONGODB_LOG_MAX_DOCUMENT_LENGTH;
```

Any value that is set on process.env is stringified, so you'll get "undefined" as a value of MONGODB_LOG_MAX_DOCUMENT_LENGTH. I believe what you want here is to remove the setting altogether. 

Also, you don't have to use computed key syntax here, dot access works too.

nbbeeken (changes requested):

test/unit/mongo_client.test.js
aditi-khare-mongoDB: Sounds good, I've changed all of them to dot access since they aren't computed, and also used the `delete` keyword instead of setting to `'undefined'`

nbbeeken (approved):

alenakhineika:
The failing test is reported in [NODE-5852](https://jira.mongodb.org/browse/NODE-5852).

alenakhineika (approved):

PR #3971, opened 2024-01-19
feat(NODE-4848): Add runtime error handling to logging

aditi-khare-mongoDB:
### Note for Reviewers
Look at changes commit by commit to easily separate async and sync changes

### Description
This ticket is a subtask of [NODE-4816](https://jira.mongodb.org/browse/NODE-4816) (Add Error Handling to the Logger).
Essentially, these changes ensure that the logger does not crash the user's application at runtime, at any point after options are parsed.

#### What is changing?
Add support for runtime error handling (ensure no constructors for loggers throw unless given type errors, log.write, stringifyWithMaxLen, defaultLogTransform).

Add async support for log.write function

See all unit tests defined under runtime error handling in [Kickoff Document](https://docs.google.com/document/d/1gDJ9vpVdH7BalRxEvzC_UZz-gIYOLSTU7HTqSoa18FA/edit#heading=h.7t9v3ahij9a1), and under mongodbLogPath stream handling (since some of this must be done at runtime).

Note: There are no changes to any of the constructors of `LoggableEvent` classes or calls to emitAndLog / variants because no `LoggableEvent` constructors have any explicit errors thrown. In addition, no arguments to `LoggableEvent` constructors are ever invalid at the time of construction. 

After NODE-5672 (Release Logger)

To ensure we don't crash the user's application through the logger.

### DIFF
--- a/src/mongo_logger.ts
+++ b/src/mongo_logger.ts
@@ -197,11 +197,12 @@ export interface MongoLoggerOptions {
 
 /**
  * Parses a string as one of SeverityLevel
+ * @internal
  *
  * @param s - the value to be parsed
  * @returns one of SeverityLevel if value can be parsed as such, otherwise null
  */
-function parseSeverityFromString(s?: string): SeverityLevel | null {
+export function parseSeverityFromString(s?: string): SeverityLevel | null {
   const validSeverities: string[] = Object.values(SeverityLevel);
   const lowerSeverity = s?.toLowerCase();
 
@@ -415,10 +416,15 @@ export function stringifyWithMaxLen(
 ): string {
   let strToTruncate = '';
 
-  if (typeof value === 'function') {
-    strToTruncate = value.toString();
-  } else {
-    strToTruncate = EJSON.stringify(value, options);
+  try {
+    strToTruncate =
+      typeof value !== 'function'
+        ? EJSON.stringify(value, options)
+        : value.name !== ''
+        ? value.name
+        : 'anonymous function';
+  } catch (e) {
+    return '... ESJON failed : Error ...';
   }
 
   return maxDocumentLength !== 0 && strToTruncate.length > maxDocumentLength
@@ -455,21 +461,21 @@ function attachCommandFields(
 ) {
   log.commandName = commandEvent.commandName;
   log.requestId = commandEvent.requestId;
-  log.driverConnectionId = commandEvent?.connectionId;
-  const { host, port } = HostAddress.fromString(commandEvent.address).toHostPort();
+  log.driverConnectionId = commandEvent.connectionId;
+  const { host, port } = HostAddress.fromString(commandEvent.address ?? '').toHostPort();
   log.serverHost = host;
   log.serverPort = port;
   if (commandEvent?.serviceId) {
     log.serviceId = commandEvent.serviceId.toHexString();
   }
   log.databaseName = commandEvent.databaseName;
-  log.serverConnectionId = commandEvent?.serverConnectionId;
+  log.serverConnectionId = commandEvent.serverConnectionId;
 
   return log;
 }
 
 function attachConnectionFields(log: Record<string, any>, event: any) {
-  const { host, port } = HostAddress.fromString(event.address).toHostPort();
+  const { host, port } = HostAddress.fromString(event.address ?? '').toHostPort();
   log.serverHost = host;
   log.serverPort = port;
 
@@ -491,13 +497,14 @@ function attachServerHeartbeatFields(
   const { awaited, connectionId } = serverHeartbeatEvent;
   log.awaited = awaited;
   log.driverConnectionId = serverHeartbeatEvent.connectionId;
-  const { host, port } = HostAddress.fromString(connectionId).toHostPort();
+  const { host, port } = HostAddress.fromString(connectionId ?? '').toHostPort();
   log.serverHost = host;
   log.serverPort = port;
   return log;
 }
 
-function defaultLogTransform(
+/** @internal */
+export function defaultLogTransform(
   logObject: LoggableEvent | Record<string, any>,
   maxDocumentLength: number = DEFAULT_MAX_DOCUMENT_LENGTH
 ): Omit<Log, 's' | 't' | 'c'> {
@@ -509,7 +516,7 @@ function defaultLogTransform(
       return log;
     case SERVER_SELECTION_FAILED:
       log = attachServerSelectionFields(log, logObject, maxDocumentLength);
-      log.failure = logObject.failure.message;
+      log.failure = logObject.failure?.message;
       return log;
     case SERVER_SELECTION_SUCCEEDED:
       log = attachServerSelectionFields(log, logObject, maxDocumentLength);
@@ -536,7 +543,7 @@ function defaultLogTransform(
       log = attachCommandFields(log, logObject);
       log.message = 'Command failed';
       log.durationMS = logObject.duration;
-      log.failure = logObject.failure.message ?? '(redacted)';
+      log.failure = logObject.failure?.message ?? '(redacted)';
       return log;
     case CONNECTION_POOL_CREATED:
       log = attachConnectionFields(log, logObject);
@@ -562,7 +569,7 @@ function defaultLogTransform(
       log = attachConnectionFields(log, logObject);
       log.message = 'Connection pool cleared';
       if (logObject.serviceId?._bsontype === 'ObjectId') {
-        log.serviceId = logObject.serviceId.toHexString();
+        log.serviceId = logObject.serviceId?.toHexString();
       }
       return log;
     case CONNECTION_POOL_CLOSED:
@@ -666,7 +673,7 @@ function defaultLogTransform(
       log = attachServerHeartbeatFields(log, logObject);
       log.message = 'Server heartbeat failed';
       log.durationMS = logObject.duration;
-      log.failure = logObject.failure.message;
+      log.failure = logObject.failure?.message;
       return log;
     case TOPOLOGY_OPENING:
       log = attachSDAMFields(log, logObject);
@@ -738,6 +745,28 @@ export class MongoLogger {
     return compareSeverity(severity, this.componentSeverities[component]) <= 0;
   }
 
+  turnOffSeverities() {
+    for (const key of Object.keys(MongoLoggableComponent)) {
+      this.componentSeverities[key as MongoLoggableComponent] = SeverityLevel.OFF;
+    }
+  }
+
+  logWriteFailureHandler(errorMessage: string) {
+    try {
+      const tempLogDestination = createStdioLogger(process.stderr);
+      tempLogDestination.write({
+        t: new Date(),
+        c: MongoLoggableComponent.CLIENT,
+        s: SeverityLevel.ERROR,
+        message: `User input for mongodbLogPath is now invalid. Now logging to stderr.`,
+        error: errorMessage
+      });
+      this.logDestination = tempLogDestination;
+    } catch (e) {
+      this.turnOffSeverities();
+    }
+  }
+
   private log(
     severity: SeverityLevel,
     component: MongoLoggableComponent,
@@ -755,7 +784,11 @@ export class MongoLogger {
         logMessage = { ...logMessage, ...defaultLogTransform(message, this.maxDocumentLength) };
       }
     }
-    this.logDestination.write(logMessage);
+    try {
+      this.logDestination.write(logMessage);
+    } catch (e) {
+      this.logWriteFailureHandler(e.message);
+    }
   }
 
   /**
--- a/src/mongo_logger.ts
+++ b/src/mongo_logger.ts
@@ -1,4 +1,3 @@
-import type { Writable } from 'stream';
 import { inspect } from 'util';
 
 import { type Document, EJSON, type EJSONOptions, type ObjectId } from './bson';
@@ -44,7 +43,6 @@ import {
   TOPOLOGY_OPENING,
   WAITING_FOR_SUITABLE_SERVER
 } from './constants';
-import { MongoError } from './error';
 import type {
   ServerClosedEvent,
   ServerOpeningEvent,
@@ -193,7 +191,9 @@ export interface MongoLoggerOptions {
   /** Max length of embedded EJSON docs. Setting to 0 disables truncation. Defaults to 1000. */
   maxDocumentLength: number;
   /** Destination for log messages. */
-  logDestination: Writable | MongoDBLogWritable;
+  logDestination: MongoDBLogWritable;
+  /** For internal check to see if error should stop logging. */
+  logDestinationIsStdErr: boolean;
 }
 
 /**
@@ -215,16 +215,14 @@ export function parseSeverityFromString(s?: string): SeverityLevel | null {
 }
 
 /** @internal */
-export function createStdioLogger(
-  stream: { write: NodeJS.WriteStream['write'] },
-  streamName: 'stderr' | 'stdout'
-): MongoDBLogWritable {
+export function createStdioLogger(stream: {
+  write: NodeJS.WriteStream['write'];
+}): MongoDBLogWritable {
   return {
     write: (log: Log): unknown => {
       stream.write(inspect(log, { compact: true, breakLength: Infinity }), 'utf-8');
       return;
-    },
-    streamName: streamName
+    }
   };
 }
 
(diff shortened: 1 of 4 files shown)


src/mongo_logger.ts
```diff
@@ -281,7 +282,7 @@ export interface Log extends Record<string, any> {
 
 /** @internal */
 export interface MongoDBLogWritable {
-  write(log: Log): void;
+  write(log: Log): any;
```
nbbeeken, line 285:
```suggestion
  write(log: Log): PromiseLike<unknown> | unknown;
```

If possible I think we would prefer `unknown` here, it forces you to write type assertion/narrowing code to make sure we're handling the right value. I also think there's an informational benefit to indicating the result can be a promise, thoughts?

src/mongo_logger.ts
```diff
@@ -455,21 +461,21 @@ function attachCommandFields(
 ) {
   log.commandName = commandEvent.commandName;
   log.requestId = commandEvent.requestId;
-  log.driverConnectionId = commandEvent?.connectionId;
-  const { host, port } = HostAddress.fromString(commandEvent.address).toHostPort();
+  log.driverConnectionId = commandEvent.connectionId;
+  const { host, port } = HostAddress.fromString(commandEvent.address ?? '').toHostPort();
```
nbbeeken, line 465: I see the `?.` may have accidentally been left in given the input shouldn't be nullish, but why did the address change?

src/mongo_logger.ts
```diff
@@ -732,12 +740,40 @@ export class MongoLogger {
...
+    }
+  }
+
+  private logWriteFailureHandler(error: Error) {
+    try {
+      const tempLogDestination = createStdioLogger(process.stderr);
+      tempLogDestination.write({
+        t: new Date(),
+        c: MongoLoggableComponent.CLIENT,
+        s: SeverityLevel.ERROR,
```
nbbeeken, line 762: Should this be done by using the `.error` helper?

src/mongo_logger.ts
```diff
@@ -732,12 +740,40 @@ export class MongoLogger {
...
 
+  turnOffSeverities() {
+    for (const key of Object.keys(MongoLoggableComponent)) {
+      this.componentSeverities[key as MongoLoggableComponent] = SeverityLevel.OFF;
+    }
+  }
+
+  private logWriteFailureHandler(error: Error) {
+    try {
+      const tempLogDestination = createStdioLogger(process.stderr);
```
nbbeeken, line 758: I think we want a condition here, if the logger was already using stderr then don't do this.

src/mongo_logger.ts
```diff
@@ -732,12 +740,40 @@ export class MongoLogger {
...
+      tempLogDestination.write({
+        t: new Date(),
+        c: MongoLoggableComponent.CLIENT,
+        s: SeverityLevel.ERROR,
+        message: `User input for mongodbLogPath is now invalid. Now logging to stderr.`,
+        error: error.message
+      });
+      this.logDestination = tempLogDestination;
+    } catch (e) {
+      this.turnOffSeverities();
```
nbbeeken, line 768: Should we be turning off logging for any error?

test/unit/mongo_logger.test.ts
```diff
@@ -1276,6 +1286,218 @@ describe('class MongoLogger', function () {
...
+              [2, 'two'],
+              [3, 'three']
+            ])
+          },
+          {
+            name: 'Object with invalid _bsontype',
+            input: { _bsontype: 'i will never be a real bson type' }
+          }
+        ];
+        for (let i = 0; i < errorInputs.length; i++) {
```
nbbeeken, line 1305:
```suggestion
        for (const errorInput of errorInputs) {
```

src/mongo_logger.ts
```diff
@@ -701,6 +708,7 @@ export class MongoLogger {
   componentSeverities: Record<MongoLoggableComponent, SeverityLevel>;
   maxDocumentLength: number;
   logDestination: MongoDBLogWritable | Writable;
+  pendingLog: any | undefined;
```
nbbeeken, line 711:
```suggestion
  pendingLog: unknown = null;
```

isPromiseLike will determine if this is nullish or not and narrow it to something that supports a .then method.

src/mongo_logger.ts
```diff
@@ -415,10 +416,15 @@ export function stringifyWithMaxLen(
...
-    strToTruncate = EJSON.stringify(value, options);
+  try {
+    strToTruncate =
+      typeof value !== 'function'
+        ? EJSON.stringify(value, options)
+        : value.name !== ''
+        ? value.name
+        : 'anonymous function';
+  } catch (e) {
+    return '... ESJON failed : Error ...';
```
nbbeeken, line 427: Should the error be included in the string here?

src/mongo_logger.ts
```diff
@@ -415,10 +416,15 @@ export function stringifyWithMaxLen(
...
-    strToTruncate = value.toString();
-  } else {
-    strToTruncate = EJSON.stringify(value, options);
+  try {
+    strToTruncate =
+      typeof value !== 'function'
+        ? EJSON.stringify(value, options)
+        : value.name !== ''
+        ? value.name
+        : 'anonymous function';
```
nbbeeken, line 425: Could you determine when we're passing in a function here? This is for server selection right? "anonymous function" is not going to be very descriptive so we should make a note of what functions are coming in here and give them informative names.

test/unit/mongo_logger.test.ts
```diff
@@ -1276,6 +1286,218 @@ describe('class MongoLogger', function () {
...
+
+        logger.debug('longer timeout');
+        logger.debug('shorter timeout');
+        logger.debug('no timeout');
+
+        expect(stream.buffer.length).to.equal(0);
+
+        await new Promise(r => setTimeout(r, 2100));
+        expect(stream.buffer.length).to.equal(1);
+        expect(stream.buffer[0]).to.equal('longer timeout');
```
nbbeeken, line 1494: Love this test!

Small improvement, instead of checking length and the item at an offset can you deep equal check with a hardcoded array?

```js
expect(stream.buffer).to.deep.equal(['longer timeout']);
```
And same below but with the subsequent logs. Then the test will print out the whole array if something goes wrong.

test/unit/mongo_logger.test.ts
```diff
@@ -1276,6 +1286,218 @@ describe('class MongoLogger', function () {
...
+        });
+      });
+    });
+    context('when async stream has multiple logs with different timeouts', async function () {
+      it('should preserve their order', async function () {
+        const stream = {
+          buffer: [],
+          async write(log) {
+            if (log.c === 'longer timeout') {
+              await new Promise(r => setTimeout(r, 2000));
```
nbbeeken, line 1473: I don't feel strongly but we have a sleep utility in test/tools/utils, we can reuse that for a bit nicer syntax here.

nbbeeken (changes requested):

test/unit/mongo_logger.test.ts
aditi-khare-mongoDB: Sounds good, it does look much neater, just changed it.

src/mongo_logger.ts
nbbeeken: Yes, the function can take `unknown` as input, it'll force a few more checks but we'll confirm any input is a promise.

aditi-khare-mongoDB: Sounds good, I updated the type

src/mongo_logger.ts
```diff
@@ -281,7 +285,8 @@ export interface Log extends Record<string, any> {
 
 /** @internal */
 export interface MongoDBLogWritable {
-  write(log: Log): void;
+  write(log: Log): PromiseLike<unknown> | any;
+  streamName?: string;
```
nbbeeken, line 289: This type is intended to go public and be the alternative to specifying stderr/out, I do not think we want users to be required to specify this property. Can the stream destination be tracked on the logger class instead to keep this out of the public API?

src/mongo_logger.ts
nbbeeken: So can you identify when you are seeing an anonymous function enter this code? like using the debugger or print stmt when you see an empty function name. Then we can trace back to the input function and find out why it doesn't have a name and give it one, possibly in a followup.

src/mongo_logger.ts
nbbeeken: How would this occur? It isn't our intention to make access to the logger public so only our "trusted" types should be able to make it into the logger.

src/mongo_logger.ts
```diff
@@ -738,6 +750,37 @@ export class MongoLogger {
...
 
+  turnOffSeverities() {
+    for (const key of Object.keys(MongoLoggableComponent)) {
+      this.componentSeverities[key as MongoLoggableComponent] = SeverityLevel.OFF;
+    }
+  }
+
+  private logWriteFailureHandler(error: Error) {
+    try {
+      if ((this.logDestination as any)?.streamName === 'stderr') {
```
nbbeeken, line 761: I think this points to a miss in the type of logDestination. I don't think we ever have Writable, it is always a MongoDBLogWritable, but I also asked in another comment that we track this stderr information on this logger class, not the destination.

src/mongo_logger.ts
```diff
@@ -738,6 +750,37 @@ export class MongoLogger {
...
+        throw MongoError;
```
nbbeeken, line 762: This was a little strange at first, but I follow, you are jumping down to the catch block. Since the error is going to be discarded anyway can you leave something descriptive: `throw new Error('...')` where ... is something about skipping over the stderr log. And I use `Error` here because that will also stand out to anyone reading this

Alternatively, switch this to an early return style if stmt, it incurs less scrutiny when others read this.

src/mongo_logger.ts
nbbeeken: I thought we wanted to stop logging after emitting this warning not to continue using stderr. It is strange to me that we would start logging to that destination for the remainder of the life of the process, the user probably has nothing set up to capture from there otherwise they would've configured stderr to be captured.

src/utils.ts
```diff
@@ -171,7 +171,10 @@ export function applyRetryableWrites<T extends HasRetryableWrites>(target: T, db
  * @param value - An object that could be a promise
  * @returns true if the provided value is a Promise
  */
-export function isPromiseLike<T = any>(value?: PromiseLike<T> | void): value is Promise<T> {
+export function isPromiseLike<T = any>(
+  value?: PromiseLike<T> | void | unknown
+): value is Promise<T> {
+  // @ts-expect-error since value exists, value.then check is valid ts
   return !!value && typeof value.then === 'function';
 }
```
nbbeeken, line 179:
```suggestion
export function isPromiseLike<T = any>(value?: unknown): value is PromiseLike<T> {
  return (
    value != null &&
    typeof value === 'object' &&
    'then' in value &&
    typeof value.then === 'function'
  );
}
```

We can just use `unknown` as the input, which forces us to prove a few more facts about the value but all worth doing since we do not know what we may be passing in.

nbbeeken (changes requested):

src/mongo_logger.ts
aditi-khare-mongoDB: Yes, it can be tracked in the logger class;  `streamName` is now removed from the public API.

src/mongo_logger.ts
aditi-khare-mongoDB: A user can pass in an anonymous/ nameless function as their serverSelector:

```
myTopology.selectServer(
    function (topologyDescription: TopologyDescription, servers: ServerDescription[], deprioritized?: ServerDescription[]): ServerDescription[] { .. do something }, 
    {}, 
    callback
);
```

I've included a test case for this as well under `describe('stringifyWithMaxLen', function ...`

src/mongo_logger.ts
aditi-khare-mongoDB: I see, I thought it was our intention to make the logger public to users, so that they could log anything they wished, especially since a user has access to the logger through `mongoClient.mongoLogger.`

I wrote a test case for this under `describe('defaultLogTransform', ...`

src/mongo_logger.ts
aditi-khare-mongoDB: I was thinking the same actually. I'll change the internal logger structure to just have `MongoDbLogWritable` as the type.

src/mongo_logger.ts
aditi-khare-mongoDB: I've changed it to an early return statement for clarity.

src/mongo_logger.ts
aditi-khare-mongoDB: We determined that this was our course of action through this [slack thread in drivers-1204](https://mongodb.slack.com/archives/G015EB3QHED/p1704916869488999), which points to this [spec section](https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#configurable-log-destination).

src/utils.ts
aditi-khare-mongoDB: Sounds good, I've changed it!

src/mongo_logger.ts
nbbeeken: Following that thread, I think single message to stderr and then ceasing future logs is the approach we want to go with.

The linked spec refers to putting a message on stderr when misconfigured, here we're concerned with a "runtime" during operation error.

src/mongo_logger.ts
```diff
@@ -281,7 +283,7 @@ export interface Log extends Record<string, any> {
 
 /** @internal */
 export interface MongoDBLogWritable {
-  write(log: Log): void;
+  write(log: Log): PromiseLike<unknown> | any;
```
nbbeeken, line 286:
```suggestion
  write(log: Log): PromiseLike<unknown> | unknown;
```

nbbeeken (changes requested):

src/mongo_logger.ts
nbbeeken: I do not think we intend to make `client.mongoLogger` public, it doesn't make sense for something outside the driver to use the driver's logger. So can we clean up these type changes?

src/mongo_logger.ts
```diff
@@ -415,10 +417,15 @@ export function stringifyWithMaxLen(
...
-    strToTruncate = EJSON.stringify(value, options);
+  try {
+    strToTruncate =
+      typeof value !== 'function'
+        ? EJSON.stringify(value, options)
+        : value.name !== ''
+        ? value.name
+        : 'anonymous function';
+  } catch (e) {
+    strToTruncate = `... ESJON failed : Error ${e.message}`;
```
addaleax, line 428: There's a typo here (`EJSON`) but it might be worth wording this a bit more explicitly anyway:

```suggestion
    strToTruncate = `Extended JSON serialization failed with: ${e.message}`;
```

src/mongo_logger.ts
aditi-khare-mongoDB: I was mistaken, client.mongoLogger is already internal. I've removed the extraneous `HostAddress.fromString(commandEvent.address ?? '').toHostPort();` and the matching test.

src/mongo_logger.ts
aditi-khare-mongoDB: Bookkeeping: We've clarified that the logger is not public / that the user cannot access it. 

We call server selection with anonymous functions [in execute operation](https://github.com/mongodb/node-mongodb-native/blob/7f97c2a2aeb88d7ee0394ade258e440ddfcb3b2f/src/operations/execute_operation.ts#L142). (When [secondaryWritableServerSelector]( https://github.com/mongodb/node-mongodb-native/blob/7f97c2a2aeb88d7ee0394ade258e440ddfcb3b2f/src/sdam/server_selection.ts#L57) or [sameServerSelector](https://github.com/mongodb/node-mongodb-native/blob/7f97c2a2aeb88d7ee0394ade258e440ddfcb3b2f/src/sdam/server_selection.ts#L39) are called)

src/mongo_logger.ts
aditi-khare-mongoDB: Sounds good, I'll configure it to turn off logging once an error with the stream is encountered and the warning message has been logged.

src/mongo_logger.ts
aditi-khare-mongoDB: Sounds good to me!

src/utils.ts
```diff
@@ -171,8 +171,13 @@ export function applyRetryableWrites<T extends HasRetryableWrites>(target: T, db
  * @param value - An object that could be a promise
  * @returns true if the provided value is a Promise
  */
-export function isPromiseLike<T = any>(value?: PromiseLike<T> | void): value is Promise<T> {
-  return !!value && typeof value.then === 'function';
+export function isPromiseLike<T = any>(value?: unknown): value is Promise<T> {
```
nbbeeken, line 174:
```suggestion
export function isPromiseLike<T = any>(value?: unknown): value is PromiseLike<T> {
```

src/mongo_logger.ts
nbbeeken: Can we file a ticket to give those functions names (or do that here?)

nbbeeken (approved):

src/utils.ts
```diff
@@ -171,8 +171,13 @@ export function applyRetryableWrites<T extends HasRetryableWrites>(target: T, db
  * @param value - An object that could be a promise
  * @returns true if the provided value is a Promise
  */
-export function isPromiseLike<T = any>(value?: PromiseLike<T> | void): value is Promise<T> {
-  return !!value && typeof value.then === 'function';
+export function isPromiseLike<T = any>(value?: unknown): value is PromiseLike<T> {
```
addaleax, line 174: It's a obviously a tradeoff, but I'd consider doing

```suggestion
export function isPromiseLike<T = unknown>(value?: unknown): value is PromiseLike<T> {
```

instead so one doesn't potentially introduce a hidden `any` by default each time they're using this function

addaleax (dismissed):

src/utils.ts
aditi-khare-mongoDB: Sounds good, changed it :)

PR #3973, opened 2024-01-23
fix(NODE-5840): heartbeat duration includes socket creation

nbbeeken:
#### What is changing?

- Refactor connect to be async
- Rename and refactor makeConnection -> makeSocket
- Add makeConnection function
- Refactor monitor to record start time after socket is created, before handshake
- socket failures set heartbeat failed duration to -1 
- Refactor tests using callback style connect
- Test duration does not include socket creation by artificially delaying it

The duration field of hb started and failed should match the definition in the spec.

### Fixed heartbeat duration including socket creation

The ServerHeartbeatSucceeded and ServerHeartbeatFailed event have a duration property that represents the time it took to perform the `hello` handshake with MongoDB. The Monitor responsible for issuing heartbeats mistakenly included the time it took to create the socket in this field, which inflates the value with the time it takes to perform a DNS lookup, TCP, and TLS handshakes.

### Double-check the following

### DIFF
--- a/src/cmap/connect.ts
+++ b/src/cmap/connect.ts
@@ -16,7 +16,7 @@ import {
   MongoRuntimeError,
   needsRetryableWriteLabel
 } from '../error';
-import { type Callback, HostAddress, ns } from '../utils';
+import { HostAddress, ns, promiseWithResolvers } from '../utils';
 import { AuthContext, type AuthProvider } from './auth/auth_provider';
 import { GSSAPI } from './auth/gssapi';
 import { MongoCR } from './auth/mongocr';
@@ -55,27 +55,26 @@ export const AUTH_PROVIDERS = new Map<AuthMechanism | string, AuthProvider>([
 /** @public */
 export type Stream = Socket | TLSSocket;
 
-export function connect(options: ConnectionOptions, callback: Callback<Connection>): void {
-  makeConnection({ ...options, existingSocket: undefined }, (err, socket) => {
-    if (err || !socket) {
-      return callback(err);
-    }
+export async function connect(options: ConnectionOptions): Promise<Connection> {
+  let connection: Connection | null = null;
+  try {
+    const socket = await makeSocket(options);
+    connection = makeConnection(options, socket);
+    await performInitialHandshake(connection, options);
+    return connection;
+  } catch (error) {
+    connection?.destroy({ force: false });
+    throw error;
+  }
+}
 
-    let ConnectionType = options.connectionType ?? Connection;
-    if (options.autoEncrypter) {
-      ConnectionType = CryptoConnection;
-    }
+export function makeConnection(options: ConnectionOptions, socket: Stream): Connection {
+  let ConnectionType = options.connectionType ?? Connection;
+  if (options.autoEncrypter) {
+    ConnectionType = CryptoConnection;
+  }
 
-    const connection = new ConnectionType(socket, options);
-
-    performInitialHandshake(connection, options).then(
-      () => callback(undefined, connection),
-      error => {
-        connection.destroy({ force: false });
-        callback(error);
-      }
-    );
-  });
+  return new ConnectionType(socket, options);
 }
 
 function checkSupportedServer(hello: Document, options: ConnectionOptions) {
@@ -103,7 +102,7 @@ function checkSupportedServer(hello: Document, options: ConnectionOptions) {
   return new MongoCompatibilityError(message);
 }
 
-async function performInitialHandshake(
+export async function performInitialHandshake(
   conn: Connection,
   options: ConnectionOptions
 ): Promise<void> {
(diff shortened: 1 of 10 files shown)


src/cmap/connect.ts
```diff
@@ -435,78 +414,71 @@ function loadSocks() {
...
+      destination: {
+        host: destination.host,
+        port: destination.port
+      },
+      proxy: {
+        // host and port are ignored because we pass existing_socket
+        host: 'iLoveJavaScript',
+        port: 0,
+        type: 5,
+        userId: options.proxyUsername || undefined,
```
alenakhineika, line 453: Is setting `undefined` explicitly necessary? If `proxyUsername` or `proxyPassword` are not in the options, their value will be `undefined` anyway.

src/error.ts
```diff
@@ -150,7 +150,7 @@ export class MongoError extends Error {
   }
 
   /** @internal */
-  static buildErrorMessage(e: Error | string): string {
+  static buildErrorMessage(e: unknown): string {
```
alenakhineika, line 153: What is the motivation for this change? I don't see where in the code it could be set differently rather than `Error` or `string`.

src/sdam/monitor.ts
```diff
@@ -352,37 +354,45 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
...
-    if (err) {
-      monitor.connection = null;
-
-      awaited = false;
-      onHeartbeatFailed(err);
-      return;
+  (async () => {
+    const socket = await makeSocket(monitor.connectOptions);
+    const connection = makeConnection(monitor.connectOptions, socket);
+    start = now();
```
alenakhineika, line 360: Maybe it would be nice to add a comment here and on the 338 line (with another `start = now()`) explaining what you were telling me about when we expect it to be called.

src/utils.ts
```diff
@@ -492,9 +492,9 @@ export function now(): number {
 }
 
 /** @internal */
-export function calculateDurationInMs(started: number): number {
+export function calculateDurationInMs(started: number | undefined): number {
```
alenakhineika, line 495: I also didn't find where started could be set as `undefined`.

test/integration/connection-monitoring-and-pooling/connection.test.ts
```diff
@@ -35,15 +33,15 @@ describe('Connection', function () {
     it('should execute a command against a server', {
       metadata: { requires: { apiVersion: false, topology: '!load-balanced' } },
       test: async function () {
-        const connectOptions: Partial<ConnectionOptions> = {
+        const connectOptions: ConnectionOptions = {
```
alenakhineika, line 36: Why do we remove `Partial` here? To get rid of those castings later `connectOptions as any as ConnectionOptions`? Because connectOptions has indeed partial type here.

test/unit/sdam/monitor.test.ts
```diff
@@ -579,4 +582,32 @@ describe('monitoring', function () {
...
+        socket.on('data', () => socket.destroy(new Error('I am not real!')));
+        return socket;
+      });
+
+      client = new MongoClient(`mongodb://localhost:1`, { serverSelectionTimeoutMS: 200 });
+      client.on('serverHeartbeatFailed', ev => (serverHeartbeatFailed = ev));
+    });
+
+    afterEach(function () {
+      sinon.restore();
```
alenakhineika, line 604: Do we need to destroy the socket in afterEach? In case something went wrong and we haven't destroyed it on data callback.

src/cmap/connect.ts
nbbeeken: If proxyUsername is an empty string we pass in `undefined` instead

src/error.ts
nbbeeken: I made the type in `connectionFailureError` more accurate which makes error nullish. To me it felt easier to propagate the handling of that to `buildErrorMessage`

alenakhineika: Isn't it then `Error | string | null`?

test/integration/connection-monitoring-and-pooling/connection.test.ts
nbbeeken: I rounded out the type so its no longer partial :)

Rather than cast away a problem we should mark why we are able to get away with violating TS but converting tests to TS is sometimes done to large files and we can leave behind type errors that don't actually break the code at runtime.

src/utils.ts
nbbeeken: Isn't it fine for a let to remain unassigned? typescript will know if you use it before assignment right?

src/sdam/monitor.ts
nbbeeken: Added doc comments

src/error.ts
nbbeeken: Any error caught in a catch block has the potential to be any type, we just disabled that rule temporarily, `useUnknownInCatchVariables`, I'm writing unknown here now so there is less work for NODE-3659 down the road.

alenakhineika (approved):

W-A-James (approved):

PR #3974, opened 2024-01-23
fix(NODE-5818): Add feature flagging to server selection logging

aditi-khare-mongoDB:
### Description
Add feature flagging to server selection logging, so that event instances are only created when `willLog` returns `true` for the logging configuration.

#### **EVG Perf Analysis Results**

Upon comparing the perf results of NODE-5818's changes to the commit before server selection logging was merged ([b93d405](https://evergreen.mongodb.com/task/mongo_node_driver_next_performance_tests_run_spec_benchmark_tests_node_18_server_6.0_b93d405275c3a8ce6b1eba0af335ffb2a309e34e_24_01_03_17_07_24), there are no significant regressions through this commit. 
Both commits were rerun at the same time to ensure no ci changes influence the results.

Here is the [Spreadsheet Link](https://docs.google.com/spreadsheets/d/1L_tlp-n5-tUFVuSAQrmIdC0aqMNvFsPLKqV8my-ACE8/edit#gid=0).

#### What is changing?
Same as above.

No.

Before: (Due to server selection logging merge [NODE-4687](https://jira.mongodb.org/browse/NODE-4687))
findOne - 6% throughput reduction
runCommand - 9% throughput reduction
smallDocInsertOne - 6.5% throughput reduction

### DIFF
--- a/src/sdam/topology.ts
+++ b/src/sdam/topology.ts
@@ -31,7 +31,7 @@ import {
   MongoTopologyClosedError
 } from '../error';
 import type { MongoClient, ServerApi } from '../mongo_client';
-import { MongoLoggableComponent, type MongoLogger } from '../mongo_logger';
+import { MongoLoggableComponent, type MongoLogger, SeverityLevel } from '../mongo_logger';
 import { TypedEventEmitter } from '../mongo_types';
 import { ReadPreference, type ReadPreferenceLike } from '../read_preference';
 import type { ClientSession } from '../sessions';
@@ -568,25 +568,36 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
     }
 
     options = { serverSelectionTimeoutMS: this.s.serverSelectionTimeoutMS, ...options };
-    this.client.mongoLogger?.debug(
-      MongoLoggableComponent.SERVER_SELECTION,
-      new ServerSelectionStartedEvent(selector, this.description, options.operationName)
-    );
+    if (
+      this.client.mongoLogger?.willLog(SeverityLevel.DEBUG, MongoLoggableComponent.SERVER_SELECTION)
+    ) {
+      this.client.mongoLogger?.debug(
+        MongoLoggableComponent.SERVER_SELECTION,
+        new ServerSelectionStartedEvent(selector, this.description, options.operationName)
+      );
+    }
 
     const isSharded = this.description.type === TopologyType.Sharded;
     const session = options.session;
     const transaction = session && session.transaction;
 
     if (isSharded && transaction && transaction.server) {
-      this.client.mongoLogger?.debug(
-        MongoLoggableComponent.SERVER_SELECTION,
-        new ServerSelectionSucceededEvent(
-          selector,
-          this.description,
-          transaction.server.pool.address,
-          options.operationName
+      if (
+        this.client.mongoLogger?.willLog(
+          SeverityLevel.DEBUG,
+          MongoLoggableComponent.SERVER_SELECTION
         )
-      );
+      ) {
+        this.client.mongoLogger?.debug(
+          MongoLoggableComponent.SERVER_SELECTION,
+          new ServerSelectionSucceededEvent(
+            selector,
+            this.description,
+            transaction.server.pool.address,
+            options.operationName
+          )
+        );
+      }
       callback(undefined, transaction.server);
       return;
     }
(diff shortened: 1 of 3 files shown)


durran:
Am I correct in analysing this that, for example, `findOne` on these changes is slower than the mainline?

Screenshot 2024-01-24 at 20 02 30

aditi-khare-mongoDB:
Bookkeeping: 
Screenshot 2024-01-24 at 3 13 03 PM

With newest merge to main, this is a perf improvement.

aditi-khare-mongoDB:
Perf Analysis Results:

Evergreen Trends:
Upon comparing the perf results of NODE-5818's changes to the commit before server selection logging was merged ([b93d405](https://evergreen.mongodb.com/task/mongo_node_driver_next_performance_tests_run_spec_benchmark_tests_node_18_server_6.0_b93d405275c3a8ce6b1eba0af335ffb2a309e34e_24_01_03_17_07_24). There are no regressions as high as 6% through this commit, but one `FindAndEmptyCursor` regressed by 3% . All other tests are perf improvements or regressions by les than 1%

[Spreadsheet Link](https://docs.google.com/spreadsheets/d/1L_tlp-n5-tUFVuSAQrmIdC0aqMNvFsPLKqV8my-ACE8/edit#gid=0)

src/sdam/topology.ts
```diff
@@ -568,25 +568,36 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
     }
 
     options = { serverSelectionTimeoutMS: this.s.serverSelectionTimeoutMS, ...options };
-    this.client.mongoLogger.debug(
-      MongoLoggableComponent.SERVER_SELECTION,
-      new ServerSelectionStartedEvent(selector, this.description, options.operationName)
-    );
+    if (
+      this.client.mongoLogger.willLog(SeverityLevel.DEBUG, MongoLoggableComponent.SERVER_SELECTION)
+    ) {
```
alenakhineika, line 573: (nit): Would it be beneficial to create a function similar to `shouldEmitAndLogCommand` and move the `willLog` check from here? True that there is only one condition that we check, but maybe with the function name it will look shorter and easier to read code. Plus, if we need to change it later, we will need to modify only one function rather than multiple `if`'s in the code. Looks good as is, just caught my eye since we are doing it quite a number of times in this pr.

aditi-khare-mongoDB: I was thinking something like that, but there are two differences between server selection logging and command logging to consider here:

1. Unlike CLAM, Server selection logging has different severity levels for its logs (debug and info), so regardless we will have to pass something into the function
2. in `shouldEmitAndLogCommand`, we check multiple booleans, here we are just calling one function.

alenakhineika (approved):

W-A-James (approved):

alenakhineika:
The PR is approved by the team, but blocked by [DEVPROD-4523](https://jira.mongodb.org/browse/DEVPROD-4523) and can be merged only when the waterfall is fixed.

Also flaky tests: NODE-5897, NODE-5498, NODE-5803

durran (changes requested):
I think we just need to rebase now, remove the `willLog` checks, and then make all calls to the mongologger user `topology.client.mongoLogger?.debug()` now.

aditi-khare-mongoDB:
@durran what is your reasoning for removing the `willLog` / feature flagging checks?

durran:
> @durran what is your reasoning for removing the `willLog` / feature flagging checks?

@aditi-khare-mongoDB The logger instantiation PR refactored the `log` method in the client logger to check `willLog` first and return if it won't log. `debug` is bound to `log` so it's going to check that first as well - I don't see the need for the redundant check to `willLog`. That's why I thought this PR would just go away.

aditi-khare-mongoDB:
@durran 
I see, while it might seem that way the log method always checked `willLog` or some similar method (`compareSeverities`) before logging. See [this PR](https://github.com/mongodb/node-mongodb-native/pull/3971/files#diff-e3cf2d38cd60718791de47479d524b551c336cf17919e9f285249d62adb374d5) (a random logging PR from before the logger instantiation PR), for reference. 

The focus of this PR is so that a `ServerSelectionEvent` instance is only created if `willLog` returns true, and the log call only occurs after the `ServerSelectionEvent` instance is created. This is why, I am of the opinion that we have to add a willLog check before we call `MongoLogger?.debug `or `MongoLogger?.info`.

durran:
> @durran I see. The `log` method always checked `willLog` or some similar method (`compareSeverities`) before logging the message. See [this PR](https://github.com/mongodb/node-mongodb-native/pull/3971/files#diff-e3cf2d38cd60718791de47479d524b551c336cf17919e9f285249d62adb374d5) (a random logging PR from before the logger instantiation PR), for reference.
> 
> The focus of this PR is so that a `ServerSelectionEvent` instance is only created if `willLog` returns true, and the `log` call only occurs after the `ServerSelectionEvent` instance is created. This is why, I am of the opinion that we should add a `willLog` check before we call `MongoLogger?.debug `or `MongoLogger?.info`.

Ah ok, thanks for clarifying that. Then I think we can proceed here.

alenakhineika (approved):

PR #3979, opened 2024-01-24
fix(NODE-5839): support for multibyte code-points in stringifyWithMaxLen

aditi-khare-mongoDB:
### Description
Support truncation of multibyte codepoints through `stringifyWithMaxLen` into well-formed strings. (round down to the nearest codepoint).

#### What is changing?

- Add codepoint check before truncation
- If maxDocLength = 1, and maxDocLength--, return ' '
- Un-skipped prose test #3 for command logging
- No longer wrap lone strings in an extra quotation while stringifying for logger (makes stringification confusing to read and is unnecessary) 

No

Prose test #3 command logging (see kickoff)

### DIFF
--- a/src/mongo_logger.ts
+++ b/src/mongo_logger.ts
@@ -420,10 +420,25 @@ export function stringifyWithMaxLen(
 ): string {
   let strToTruncate = '';
 
-  try {
-    strToTruncate = typeof value !== 'function' ? EJSON.stringify(value, options) : value.name;
-  } catch (e) {
-    strToTruncate = `Extended JSON serialization failed with: ${e.message}`;
+  if (typeof value === 'string') {
+    strToTruncate = value;
+  } else if (typeof value === 'function') {
+    strToTruncate = value.toString();
+  } else {
+    strToTruncate = EJSON.stringify(value, options);
+  }
+
+  // handle truncation that occurs in the middle of multi-byte codepoints
+  if (
+    maxDocumentLength !== 0 &&
+    strToTruncate.length > maxDocumentLength &&
+    strToTruncate.charCodeAt(maxDocumentLength - 1) !==
+      strToTruncate.codePointAt(maxDocumentLength - 1)
+  ) {
+    maxDocumentLength--;
+    if (maxDocumentLength === 0) {
+      return '';
+    }
   }
 
   return maxDocumentLength !== 0 && strToTruncate.length > maxDocumentLength
--- a/src/mongo_logger.ts
+++ b/src/mongo_logger.ts
@@ -425,7 +425,11 @@ export function stringifyWithMaxLen(
   } else if (typeof value === 'function') {
     strToTruncate = value.toString();
   } else {
-    strToTruncate = EJSON.stringify(value, options);
+    try {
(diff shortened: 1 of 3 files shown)


test/unit/mongo_logger.test.ts
```diff
@@ -1248,14 +1248,36 @@ describe('class MongoLogger', function () {
...
+          context('when maxDocumentLength = 1 but greater than 0', function () {
+            it('should return an empty string', function () {
+              expect(stringifyWithMaxLen(multiByteCodePoint, 1, { relaxed: true })).to.equal('');
+            });
+          });
+
+          context('when maxDocumentLength > 1', function () {
+            it('should round down maxDocLength to previous codepoint', function () {
+              const randomString = `random ${multiByteCodePoint}random random${multiByteCodePoint}${multiByteCodePoint}`;
+              const randomStringMinusACodePoint = `random ${multiByteCodePoint}random random${multiByteCodePoint}`;
```
addaleax, line 1274:
```suggestion
              const randomStringMinusACodePoint = `random ${multiByteCodePoint}random random${multiByteCodePoint}`;
              const randomString = `${randomStringMinusACodePoint}${multiByteCodePoint}`;
```

very optional, but this would make it a bit easier to see that `randomStringMinusACodePoint` really is a prefix of `randomString` by just looking at the test :)

addaleax (dismissed):

test/unit/mongo_logger.test.ts
aditi-khare-mongoDB: I see, that's a good idea! Just added it in

nbbeeken (approved):

nbbeeken (changes requested):
from slack: merge conflict, req changes for viz

src/mongo_logger.ts
```diff
@@ -415,12 +415,27 @@ export function stringifyWithMaxLen(
...
   } else {
     strToTruncate = EJSON.stringify(value, options);
   }
 
+  // handle truncation that occurs in the middle of multi-byte codepoints
+  if (
+    maxDocumentLength !== 0 &&
+    strToTruncate.length > maxDocumentLength &&
+    strToTruncate.charCodeAt(maxDocumentLength - 1) !==
+      strToTruncate.codePointAt(maxDocumentLength - 1)
```
W-A-James, line 431: I'm not the most well versed on Unicode, so a general question: Are there valid multi-byte codepoints where these two functions return the same value? I'd imagine there aren't but I just want to confirm.

aditi-khare-mongoDB: There shouldn't be. 
Here, we assume 1 unicode codepoint = 2 unicode code units. Valid multibyte codepoints may have `codePointAt(codeUnitIndex) === charCodeAt(codeUnitIndex) ` can be true for the second code unit in the second but not the first. `charCodeAt` only returns values as high as until 65535 (the number of code units), and for multibyte characters, the code point value must be higher than 65535 for the first code unit. 

There is more context [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt).

nbbeeken (approved):

PR #3984, opened 2024-02-02
perf(NODE-5854): Conditional logger instantiation and precompute willLog perf fix

aditi-khare-mongoDB:
### Description
Misc perf improvements to logger.

#### What is changing?
`new MongoLogger` is only instantiated on the MongoClient when at least one of the component severities is turned on. 
`willLog` is now a precomputed nested object rather than a function call (access is cheaper than a function)

Not until logging is released NODE-5672

To align with the company wide perf initiative and ensure that the logger is optimized.


alenakhineika:
Could you double-check the list in the PR description? To make sure that all requirements are met.

src/mongo_logger.ts
```diff
@@ -728,6 +728,10 @@ export class MongoLogger {
   logDestination: MongoDBLogWritable;
   logDestinationIsStdErr: boolean;
   pendingLog: PromiseLike<unknown> | unknown = null;
+  willLog: Record<MongoLoggableComponent, Record<SeverityLevel, boolean>> = {} as Record<
```
alenakhineika, line 731: Can we change a type here in a way that we don't need a casting?

src/mongo_logger.ts
```diff
@@ -760,15 +764,26 @@ export class MongoLogger {
...
-    return compareSeverity(severity, this.componentSeverities[component]) <= 0;
+  createWillLog() {
+    for (const component of Object.values(MongoLoggableComponent)) {
+      this.willLog[component] = Object.fromEntries(
+        Object.entries(SeverityLevel).map(([_key, value]) => [
+          value as SeverityLevel,
+          compareSeverity(value, this.componentSeverities[component]) <= 0
+        ])
+      ) as Record<SeverityLevel, boolean>;
+    }
```
alenakhineika, line 778: (nit): I am to sure if this is easier to read, but at least we can avoid casting.

```js
createWillLog() {
  for (const component of Object.values(MongoLoggableComponent)) {
    for (const severityLevel of Object.values(SeverityLevel)) {
      this.willLog[component] = {
        ...this.willLog[component],
        [severityLevel]: compareSeverity(severityLevel, this.componentSeverities[component]) <= 0
      };
    }
  }
}
```

Maybe there is even a better way to refactor it. Feel free to explore or ignore if this is a common way it is done elsewhere in the codebase.

src/mongo_logger.ts
aditi-khare-mongoDB: What do you suggest?

src/mongo_logger.ts
aditi-khare-mongoDB: I see. One worry I have about this way to refactor it is having to run the spread `...` complexity for each entry in `willLog`. Do you have any suggestions on how to avoid casting and avoid using spread?

alenakhineika: Would `Object.assign()` work?

```js
  for (const component of Object.values(MongoLoggableComponent)) {
    for (const severityLevel of Object.values(SeverityLevel)) {
      this.willLog[component] = Object.assign({}, this.willLog[component], {
        [severityLevel]: compareSeverity(severityLevel, this.componentSeverities[component]) <= 0
      });
    }
  }
}
```

src/mongo_logger.ts
aditi-khare-mongoDB: I tried this. While it gets rid of the type casting being necessary its also not an accurate type for our willLog object as it doesn't assert all the `MongoLoggableComponent`s will be in the object. I've found a workaround though, if you see my most recent push!

src/mongo_logger.ts
aditi-khare-mongoDB: I believe we're trying to use `Object.assign()` less in the Node Driver and it has the same O(N) complexity as spread. I've found a typecasting work around though (in my latest commit).

alenakhineika (approved):

src/mongo_logger.ts
```diff
@@ -728,6 +728,7 @@ export class MongoLogger {
   logDestination: MongoDBLogWritable;
   logDestinationIsStdErr: boolean;
   pendingLog: PromiseLike<unknown> | unknown = null;
+  willLog: Record<MongoLoggableComponent, Record<SeverityLevel, boolean>>;
```
durran, line 731:
```suggestion
  private severities: Record<MongoLoggableComponent, Record<SeverityLevel, boolean>>;
```

From a design perspective I don't think we should require other objects asking the logger to know about it's internal structure and this should be encapsulated behind an actual `willLog` method like before, since it was previously a public method and is now a publicly accessible attribute (even though it's called only internally for now). See other notes and suggestions on how I'd refactor this.

src/mongo_logger.ts
```diff
@@ -760,15 +761,27 @@ export class MongoLogger {
     this.maxDocumentLength = options.maxDocumentLength;
     this.logDestination = options.logDestination;
     this.logDestinationIsStdErr = options.logDestinationIsStdErr;
+    this.willLog = this.createWillLog();
```
durran, line 764:
```suggestion
    this.severities = this.createLoggingSeverities();
```
Suggest renaming the structure and method to create it.

src/mongo_logger.ts
```diff
@@ -760,15 +761,27 @@ export class MongoLogger {
...
+  createWillLog(): Record<MongoLoggableComponent, Record<SeverityLevel, boolean>> {
+    const willLog = Object();
+    for (const component of Object.values(MongoLoggableComponent)) {
+      willLog[component] = {};
+      for (const severityLevel of Object.values(SeverityLevel)) {
+        willLog[component][severityLevel] =
+          compareSeverity(severityLevel, this.componentSeverities[component]) <= 0;
+      }
+    }
+    return willLog;
```
durran, line 776:
```suggestion
  createSeverities(): Record<MongoLoggableComponent, Record<SeverityLevel, boolean>> {
    const severities = Object();
    for (const component of Object.values(MongoLoggableComponent)) {
      severities[component] = {};
      for (const severityLevel of Object.values(SeverityLevel)) {
        severities[component][severityLevel] =
          compareSeverity(severityLevel, this.componentSeverities[component]) <= 0;
      }
    }
    return severities;
```

src/mongo_logger.ts
```diff
@@ -760,15 +761,27 @@ export class MongoLogger {
...
   }
 
   turnOffSeverities() {
-    for (const key of Object.values(MongoLoggableComponent)) {
-      this.componentSeverities[key as MongoLoggableComponent] = SeverityLevel.OFF;
+    for (const component of Object.values(MongoLoggableComponent)) {
+      this.componentSeverities[component] = SeverityLevel.OFF;
+      for (const severityLevel of Object.values(SeverityLevel)) {
+        this.willLog[component][severityLevel] = false;
```
durran, line 783:
```suggestion
        this.severities[component][severityLevel] = false;
```

src/mongo_logger.ts
```diff
@@ -802,7 +815,7 @@ export class MongoLogger {
     component: MongoLoggableComponent,
     message: Loggable | string
   ): void {
-    if (!this.willLog(severity, component)) return;
+    if (!this.willLog[component][severity]) return;
```
durran, line 818:
```suggestion
    if (!this.willLog(component, severity)) return;
```

src/mongo_logger.ts
```diff
@@ -760,15 +761,27 @@ export class MongoLogger {
     this.maxDocumentLength = options.maxDocumentLength;
     this.logDestination = options.logDestination;
     this.logDestinationIsStdErr = options.logDestinationIsStdErr;
+    this.willLog = this.createWillLog();
   }
```
durran, line 766:
```suggestion
  willLog(component: MongoLoggableComponent, level: SeverityLevel): boolean {
    return this.severities[component][severity];
  }
```

test/unit/mongo_logger.test.ts
```diff
@@ -114,6 +115,61 @@ describe('class MongoLogger', async function () {
         expect(buffer).to.have.lengthOf(1);
       });
     });
+
+    describe('willLog', function () {
```
durran, line 119: If we refactor as my suggestions, then this becomes a unit test of the `willLog()` method and the test no longer needs to know about the internal structure of the severities. I think it makes the tests more easy to reason about here.

durran (changes requested):

src/mongo_client.ts
```diff
@@ -359,7 +360,14 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
     super();
 
     this[kOptions] = parseOptions(url, this, options);
-    this.mongoLogger = new MongoLogger(this[kOptions].mongoLoggerOptions);
+
+    const shouldSetLogger =
+      Object.values(this[kOptions].mongoLoggerOptions.componentSeverities).filter(
+        value => value !== SeverityLevel.OFF
+      ).length !== 0;
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good, I just changed it.

src/mongo_logger.ts
aditi-khare-mongoDB: I'm conflicted here, since we are using the willLog as a precomputed object in order to avoid the perf cost of calling a function every-time (access to a variable is quicker). How will this refactor affect perf?

aditi-khare-mongoDB: Additionally, I'm trying to think of a more precise name for `severities`, since it could get mixed up with `componentSeverities`.  In either case, I'll add a clarifying comment to both variables specifying their function.

durran: We would still be using a precomputed object to avoid the perf costs - you can see from my suggestions that still happens. The only difference here is that we access that through a method in my suggestions instead of accessing the precomputed object directly. This encapsulation of that access should not have an impact but also allows us to be able to refactor the underlying structure at any point in time without the need to refactor potential multiple calls to it in the future, it all goes through one method.

durran: I'm also completely fine using a different name than `severities` - I just called it that for the purposes of trying to explain in code what I was suggesting.

aditi-khare-mongoDB: I think since it's not a public facing variable, we don't have to be as careful with naming. The code comments should suffice imo.

test/unit/mongo_logger.test.ts
```diff
@@ -1548,4 +1549,70 @@ describe('class MongoLogger', async function () {
...
+            for (const severityLevel of Object.values(SeverityLevel)) {
+              expect(logger.willLog(component3, severityLevel)).to.equal(
+                expectedWillLogResults[component3][severityLevel]
+              );
+            }
+          }
+        });
+      });
+    }
+  });
```
durran, line 1617:
```suggestion
  describe('#willLog', function () {
    const severityLevels = Object.values(SeverityLevel);
    for (const severityLevel of severityLevels) {
      context(`when the severity level is ${severityLevel}`, function () {
        let logger: MongoLogger;
        let componentSeverities;
        const components = Object.values(MongoLoggableComponent);

        for (const component of components) {
          context(`when ${component} severity level <= ${severityLevel}`, function () {
            beforeEach(function () {
              const index = severityLevels.indexOf(severityLevel);
              componentSeverities = components.reduce((severities, value) => {
                component === value
                  ? (severities[component] = severityLevel)
                  : (severities[value] =
                      severityLevels[index + 1] === 'off'
                        ? severityLevel
                        : severityLevels[index + 1]);
                return severities;
              }, {});
              logger = new MongoLogger({
                componentSeverities: componentSeverities,
                logDestination: createStdioLogger(process.stderr),
                logDestinationIsStdErr: true
              } as any);
            });

            if (severityLevel === 'off') {
              it('returns false always for off', function () {
                expect(logger.willLog(component, severityLevel)).to.be.false;
              });
            } else {
              it('returns true', function () {
                expect(logger.willLog(component, severityLevel)).to.be.true;
              });
            }
          });

          context(`when ${component} severity level > ${severityLevel}`, function () {
            if (severityLevel !== 'emergency') {
              beforeEach(function () {
                const index = severityLevels.indexOf(severityLevel);
                componentSeverities = components.reduce((severities, value) => {
                  component === value
                    ? (severities[component] = severityLevels[index - 1])
                    : (severities[value] = severityLevel);
                  return severities;
                }, {});
                logger = new MongoLogger({
                  componentSeverities: componentSeverities,
                  logDestination: createStdioLogger(process.stderr),
                  logDestinationIsStdErr: true
                } as any);
              });

              it('returns false', function () {
                expect(logger.willLog(component, severityLevel)).to.be.false;
              });
            }
          });
        }
      });
    }
  });
});

```

Ok so I spent some time going over these tests, and I'll explain the rational behind this suggestions.

First, the previous tests had one `it` block with multiple assertions. When looking at the test output it would not be clear on which component or severity level failed, so I refactored the tests to have an assertion per `it` block per component and severity level, thus giving us a test output like so:

```
class MongoLogger
    #willLog
      when the severity level is emergency
        when command severity level <= emergency
          ✔ returns true
        when topology severity level <= emergency
          ✔ returns true
        when serverSelection severity level <= emergency
          ✔ returns true
        when connection severity level <= emergency
          ✔ returns true
        when client severity level <= emergency
          ✔ returns true
      when the severity level is alert
        when command severity level <= alert
          ✔ returns true
        when command severity level > alert
          ✔ returns false
        when topology severity level <= alert
          ✔ returns true
        when topology severity level > alert
          ✔ returns false
        when serverSelection severity level <= alert
          ✔ returns true
        when serverSelection severity level > alert
          ✔ returns false
        when connection severity level <= alert
          ✔ returns true
        when connection severity level > alert
          ✔ returns false
        when client severity level <= alert
          ✔ returns true
        when client severity level > alert
          ✔ returns false
      when the severity level is critical
        when command severity level <= critical
          ✔ returns true
        when command severity level > critical
          ✔ returns false
        when topology severity level <= critical
          ✔ returns true
        when topology severity level > critical
          ✔ returns false
        when serverSelection severity level <= critical
          ✔ returns true
        when serverSelection severity level > critical
          ✔ returns false
        when connection severity level <= critical
          ✔ returns true
        when connection severity level > critical
          ✔ returns false
        when client severity level <= critical
          ✔ returns true
        when client severity level > critical
          ✔ returns false
      when the severity level is error
        when command severity level <= error
          ✔ returns true
        when command severity level > error
          ✔ returns false
        when topology severity level <= error
          ✔ returns true
        when topology severity level > error
          ✔ returns false
        when serverSelection severity level <= error
          ✔ returns true
        when serverSelection severity level > error
          ✔ returns false
        when connection severity level <= error
          ✔ returns true
        when connection severity level > error
          ✔ returns false
        when client severity level <= error
          ✔ returns true
        when client severity level > error
          ✔ returns false
      when the severity level is warn
        when command severity level <= warn
          ✔ returns true
        when command severity level > warn
          ✔ returns false
        when topology severity level <= warn
          ✔ returns true
        when topology severity level > warn
          ✔ returns false
        when serverSelection severity level <= warn
          ✔ returns true
        when serverSelection severity level > warn
          ✔ returns false
        when connection severity level <= warn
          ✔ returns true
        when connection severity level > warn
          ✔ returns false
        when client severity level <= warn
          ✔ returns true
        when client severity level > warn
          ✔ returns false
      when the severity level is notice
        when command severity level <= notice
          ✔ returns true
        when command severity level > notice
          ✔ returns false
        when topology severity level <= notice
          ✔ returns true
        when topology severity level > notice
          ✔ returns false
        when serverSelection severity level <= notice
          ✔ returns true
        when serverSelection severity level > notice
          ✔ returns false
        when connection severity level <= notice
          ✔ returns true
        when connection severity level > notice
          ✔ returns false
        when client severity level <= notice
          ✔ returns true
        when client severity level > notice
          ✔ returns false
      when the severity level is info
        when command severity level <= info
          ✔ returns true
        when command severity level > info
          ✔ returns false
        when topology severity level <= info
          ✔ returns true
        when topology severity level > info
          ✔ returns false
        when serverSelection severity level <= info
          ✔ returns true
        when serverSelection severity level > info
          ✔ returns false
        when connection severity level <= info
          ✔ returns true
        when connection severity level > info
          ✔ returns false
        when client severity level <= info
          ✔ returns true
        when client severity level > info
          ✔ returns false
      when the severity level is debug
        when command severity level <= debug
          ✔ returns true
        when command severity level > debug
          ✔ returns false
        when topology severity level <= debug
          ✔ returns true
        when topology severity level > debug
          ✔ returns false
        when serverSelection severity level <= debug
          ✔ returns true
        when serverSelection severity level > debug
          ✔ returns false
        when connection severity level <= debug
          ✔ returns true
        when connection severity level > debug
          ✔ returns false
        when client severity level <= debug
          ✔ returns true
        when client severity level > debug
          ✔ returns false
      when the severity level is trace
        when command severity level <= trace
          ✔ returns true
        when command severity level > trace
          ✔ returns false
        when topology severity level <= trace
          ✔ returns true
        when topology severity level > trace
          ✔ returns false
        when serverSelection severity level <= trace
          ✔ returns true
        when serverSelection severity level > trace
          ✔ returns false
        when connection severity level <= trace
          ✔ returns true
        when connection severity level > trace
          ✔ returns false
        when client severity level <= trace
          ✔ returns true
        when client severity level > trace
          ✔ returns false
      when the severity level is off
        when command severity level <= off
          ✔ returns false always for off
        when command severity level > off
          ✔ returns false
        when topology severity level <= off
          ✔ returns false always for off
        when topology severity level > off
          ✔ returns false
        when serverSelection severity level <= off
          ✔ returns false always for off
        when serverSelection severity level > off
          ✔ returns false
        when connection severity level <= off
          ✔ returns false always for off
        when connection severity level > off
          ✔ returns false
        when client severity level <= off
          ✔ returns false always for off
        when client severity level > off
          ✔ returns false


  95 passing (71ms)

wrote xunit.xml                                                                                                                          
```

This gives us much better insight when looking at build failure logs as to what has failed specifically in these tests, and makes it easier to reason about what exactly the problem is before investigation.

Secondly, in my refactor I changed the setup to handle both sides of the comparison, <= and >. Setting up the logger options to use the indexes of the security level enum to dynamically generate test that cover every possible combination. This lead to...

Thirdly, the generation of the severities object was not quite correct for the way in which it was accessed, and the test refactor exposed a bug in the handling of the "off" severity. In the creation of the pre-generated object, "off" would be set to `true` if was set in the options for a component, causing `willLog` to return `true` if "off" was set on the component. This would have been the same outcome in the previous iterations of this PR. You will see in my new suggestion for the `willLog` method will address that. I do note that "off" would never be exposed via the public logging methods we use as it's not bound as a logging method, but this will future proof possible issues.

Apologies for the huge section here. I can if you like push to this branch with my local changes posted here, or feel free to change yourself. Great work so far BTW.

src/mongo_logger.ts
```diff
@@ -797,12 +810,16 @@ export class MongoLogger {
     this.pendingLog = null;
   }
 
+  willLog(component: MongoLoggableComponent, severity: SeverityLevel): boolean {
+    return this.severities[component][severity];
```
durran, line 814:
```suggestion
    if (severity === SeverityLevel.OFF) return false;
    return this.severities[component][severity];
```

durran (changes requested):

src/mongo_logger.ts
aditi-khare-mongoDB: I see where you're coming from here. I've made the change.

durran (approved):

nbbeeken (approved):

PR #3991, opened 2024-02-13
feat(NODE-5939): Implement 6.x: cache the AWS credentials provider in the MONGODB-AWS auth logic

alenakhineika:
### Description
Cashing credentials is required by the [spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#caching-credentials).

The AWS credentials provider instance itself is implemented as a smart credentials cache. It will issue at most one request to IMDS, and will only issue another request when the credentials are about to expire.

Instead of creating a new AWS provider for each authentication, we cache the AWS credentials provider per client.

#### What is changing?

- Refactor `AUTH_PROVIDERS` to have factory functions instead of creating providers
- Create instances of auth providers in `MongoClient` and cache them when requested per client to ensure that cached credentials are not shared with other clients

  - Expose the `getAuthProvider()` function per `ClientOptions`, so it can be requested by connection and monitoring to communicate with auth providers
  - A client can have more than one auth provider because while connecting we can e.g. switch from `MONGODB_DEFAULT` to `MONGODB_SCRAM_SHA256`

- Cache AWS provider per `MongoDBAWS`. AWS SDK will cache the first HTTP request as long as the provider is not reconstructed. We refactor the codebase to ensure that the AWS provider is stored per the `MongoDBAWS` instance to prevent overwhelming the auth endpoint

None

[NODE-5616](https://jira.mongodb.org/browse/NODE-5616)

### Caching AWS credentials provider per client

Instead of creating a new AWS provider for each authentication, we cache the AWS credentials provider per client to prevent overwhelming the auth endpoint and ensure that cached credentials are not shared with other clients.

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -6,20 +6,12 @@ import { promisify } from 'util';
 import { type BSONSerializeOptions, type Document, resolveBSONOptions } from './bson';
 import { ChangeStream, type ChangeStreamDocument, type ChangeStreamOptions } from './change_stream';
 import type { AutoEncrypter, AutoEncryptionOptions } from './client-side-encryption/auto_encrypter';
-import { type AuthProvider } from './cmap/auth/auth_provider';
-import { GSSAPI } from './cmap/auth/gssapi';
 import {
   type AuthMechanismProperties,
   DEFAULT_ALLOWED_HOSTS,
   type MongoCredentials
 } from './cmap/auth/mongo_credentials';
-import { MongoCR } from './cmap/auth/mongocr';
-import { MongoDBAWS } from './cmap/auth/mongodb_aws';
-import { MongoDBOIDC } from './cmap/auth/mongodb_oidc';
-import { Plain } from './cmap/auth/plain';
 import { AuthMechanism } from './cmap/auth/providers';
-import { ScramSHA1, ScramSHA256 } from './cmap/auth/scram';
-import { X509 } from './cmap/auth/x509';
 import type { LEGAL_TCP_SOCKET_OPTIONS, LEGAL_TLS_SOCKET_OPTIONS } from './cmap/connect';
 import type { Connection } from './cmap/connection';
 import type { ClientMetadata } from './cmap/handshake/client_metadata';
@@ -29,6 +21,7 @@ import { MONGO_CLIENT_EVENTS } from './constants';
 import { Db, type DbOptions } from './db';
 import type { Encrypter } from './encrypter';
 import { MongoInvalidArgumentError } from './error';
+import { MongoClientAuthProviders } from './mongo_client_auth_providers';
 import {
   type LogComponentSeveritiesClientOptions,
   type MongoDBLogWritable,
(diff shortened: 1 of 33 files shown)


src/mongo_client.ts
```diff
@@ -380,6 +404,15 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
       sessionPool: new ServerSessionPool(this),
       activeSessions: new Set(),
 
+      getAuthProvider(name: AuthMechanism | string) {
```
durran, line 407: I find it a little bit unexpected that this method resides on the client's internal state object. Could we possibly just make it a floating function in its own module and mark it @internal and then move the `AUTH_PROVIDERS` map into that module as well?

src/mongo_client.ts
```diff
@@ -380,6 +404,15 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
+        if (!clientAuthProviders[name]) {
+          const provider = AUTH_PROVIDERS.get(name)?.();
+          if (provider) {
+            clientAuthProviders[name] = provider;
+          }
```
durran, line 412: I think we should probably throw here if there's no provider to allude that an invalid name was passed in.

test/integration/auth/mongodb_aws.test.ts
```diff
@@ -250,9 +261,22 @@ describe('MONGODB-AWS', function () {
...
+          await client.close();
+          await client.connect();
+          await client
+            .db('aws')
+            .collection('aws_test')
+            .estimatedDocumentCount()
+            .catch(error => error);
+
+          expect(n).to.be.eql(1);
+        });
```
durran, line 279: The testing AC on the ticket also asks for a test that asserts `fromNodeProviderChain` is only called once - can we add that test as well?

durran (changes requested):

src/mongo_client.ts
alenakhineika: Considering that the method looks weird in the client's internal state object, I refactored it a bit and moved the auth providers cache to the separate MongoClientAuthProviders module. We can't completely get rid of it as a part of the MongoClient state, because we need to create providers per client instead of per module. We could make getAuthProvider a floating function, but then we need to pass the client reference to it from the connection pool and monitoring instances to get access to existing client providers, and then pass it all along across different helper functions like performInitialHandshake() and prepareHandshakeDocument(), which all have access to ConnectionOptions. Another thing to consider is that we can't decide on a single provider when we create a MongoClient instance because a client can have more than one auth provider, - while connecting we can e.g. switch from MONGODB_DEFAULT to MONGODB_SCRAM_SHA256. Let me know if this sounds reasonable or if you have thoughts on how to refactor it further.

test/integration/auth/mongodb_aws.test.ts
alenakhineika: I renamed the meaningless `n` counter to make it more descriptive :) The `numberOfFromNodeProviderChainCalls` verifies that the subbed `MongoDBAWS.credentialProvider` is called only once.

durran (approved):

aditi-khare-mongoDB (approved):

PR #3995, opened 2024-02-14
docs(NODE-5933): remove find-and-modify warning about defaults

I3uckwheat:
#### What is changing?

The previous version of the documentation (5.x) also had the lines that are removed in this commit. However, after testing and verifying, the default behavior is as-if the property "includeResultMetadata" was set to false.

The change is documentation 

Incorrect documentation 

### DIFF
--- a/src/operations/find_and_modify.ts
+++ b/src/operations/find_and_modify.ts
@@ -53,7 +53,6 @@ export interface FindOneAndReplaceOptions extends CommandOperationOptions {
   let?: Document;
   /**
    * Return the ModifyResult instead of the modified document. Defaults to true
-   * but will default to false in the next major version.
    */
   includeResultMetadata?: boolean;
 }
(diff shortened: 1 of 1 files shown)


durran (approved):

src/operations/find_and_modify.ts
```diff
@@ -53,7 +53,6 @@ export interface FindOneAndReplaceOptions extends CommandOperationOptions {
   let?: Document;
   /**
    * Return the ModifyResult instead of the modified document. Defaults to true
```
durran, line 55:
```suggestion
   * Return the ModifyResult instead of the modified document. Defaults to false
```

src/operations/find_and_modify.ts
```diff
@@ -78,7 +77,6 @@ export interface FindOneAndUpdateOptions extends CommandOperationOptions {
...
```
durran, line 79:
```suggestion
   * Return the ModifyResult instead of the modified document. Defaults to false
```

durran (changes requested):

durran (approved):

I3uckwheat:
Oh shoot, got confused. Thank you for fixing and merging!

PR #4005, opened 2024-02-27
feat(NODE-5968): Container and Kubernetes Awareness

aditi-khare-mongoDB:
### Description
Track user usage of containers. 

#### What is changing?
When kubernetes is in `process.env`, we add a  `{orchestrator: 'kubernetes'}` field to the `client.env` field of the handshake document (if this does not make the document exceed 512 bytes).

When '/.dockerenv' exists, we add a  `{runtime: 'docker'}` field to the `client.env` field of the handshake document (if this does not make the document exceed 512 bytes).

Note: We do not unit test docker, but I tested our logic for detecting docker on an EVG Host, ran Docker and detection succeeds for both root and non-root user cases.

No, non-user facing change. 

Product wants to track user usage of docker and kubernetes.

### Container and Kubernetes Awareness
The Node.js driver now keeps track of container metadata in the `client.env.container` field of the handshake document. 

If space allows, the following metadata will be included in `client.env.container`:

```
env?: { 
  container?: {
    orchestrator?: 'kubernetes' // if process.env.KUBERNETES_SERVICE_HOST is set
    runtime?: 'docker' // if the '/.dockerenv' file exists
  } 
}
```

Note: If neither Kubernetes nor Docker is present, `client.env` will not have the `container` property.


aditi-khare-mongoDB:
Unsure what's causing transaction unpin tests to magically start running again on this PR.....

test/unit/cmap/connect.test.ts
```diff
@@ -185,9 +185,100 @@ describe('Connect Tests', function () {
...
+          connection: {},
+          options: { ...CONNECT_DEFAULTS }
+        };
+
+        beforeEach(() => {
+          process.env.KUBERNETES_SERVICE_HOST = 'I exist';
+        });
+
+        afterEach(() => {
+          process.env.KUBERNETES_SERVICE_HOST = cachedEnv.KUBERNETES_SERVICE_HOST;
```
nbbeeken, line 203: This will not restore `KUBERNETES_SERVICE_HOST` to what it was before if it was `undefined` (the most common case), it will set it to `'undefined'` (string) because every assignment to process.env stringifies the value.

test/unit/cmap/connect.test.ts
```diff
@@ -185,9 +185,100 @@ describe('Connect Tests', function () {
...
+
+      context('when container nor FAAS env is not present (empty string case)', () => {
+        const authContext = {
+          connection: {},
+          options: { ...CONNECT_DEFAULTS }
+        };
+
+        context('when process.env.KUBERNETES_SERVICE_HOST = undefined', () => {
+          beforeEach(() => {
+            process.env.KUBERNETES_SERVICE_HOST = '';
```
nbbeeken, line 250: This test says undefined but is testing empty string

nbbeeken (changes requested):

test/unit/cmap/connect.test.ts
```diff
@@ -185,9 +185,156 @@ describe('Connect Tests', function () {
...
+          expect(handshakeDocument.client.env.name).to.equal('aws.lambda');
+        });
+
+        context('when 512 byte size limit is exceeded', async () => {
+          it(`should not have 'container' property in client.env`, async () => {
+            let longAppName = '';
+            // make metadata = 507 bytes, so it takes up entire LimitedSizeDocument
+            for (let i = 0; i < 485; i++) {
+              longAppName += 's';
+            }
```
nbbeeken, line 271:
```suggestion
            // make metadata = 507 bytes, so it takes up entire LimitedSizeDocument
            const appName = 's'.repeat(485);
```

test/unit/cmap/connect.test.ts
```diff
@@ -185,9 +185,156 @@ describe('Connect Tests', function () {
...
+              options: {
+                ...CONNECT_DEFAULTS,
+                metadata: {
+                  appName: longAppName,
+                  env: { name: 'aws.lambda' }
+                }
+              }
+            };
+            const handshakeDocument = await prepareHandshakeDocument(longAuthContext);
+            expect(handshakeDocument.client).to.not.have.property('env');
```
nbbeeken, line 283: The test name is "should not have 'container' property in client.env" but we're asserting there is no "env" at all. I don't think that is what we would expect here, I think the input metadata you've got here is already above 512 which shouldn't be possible real-world since we build the metadata with a LimitedSizeDocument of 512 in parseOptions.

nbbeeken (changes requested):

test/unit/cmap/connect.test.ts
aditi-khare-mongoDB: My mistake, I need to double check my copy pasting.

test/unit/cmap/connect.test.ts
aditi-khare-mongoDB: Good to know this trick.

nbbeeken (dismissed):

src/cmap/connect.ts
```diff
@@ -200,11 +200,12 @@ export async function prepareHandshakeDocument(
   const options = authContext.options;
   const compressors = options.compressors ? options.compressors : [];
   const { serverApi } = authContext.connection;
+  const clientMetadata = await addContainerMetadata(options.metadata);
```
baileympearson, line 203: Currently, we cache the result of checking for a `.dockerenv` file (i.e., whether or not we're running in a docker environment) but we don't cache the handshake's client metadata object.  Consequently, every connection attempt will attempt to add the environment metadata (looping over keys in the metadata document, calling `ifItFitsItSits` for each key).  

I was imagining that, instead of continuing to build the metadata document during options parsing and then adding the env metadata at connect time, we'd instead make the entirety of metadata document construction async, build the whole client metadata object during connect time but cache the result so we only calculate it once per-client.  ([durran's suggested approach](https://mongodb.slack.com/archives/GGWBN4ZNK/p1709068406115619) is another way to achieve the outcome of only building the env metadata once).

thoughts?

src/cmap/handshake/client_metadata.ts
```diff
@@ -152,8 +153,62 @@ export function makeClientMetadata(options: MakeClientMetadataOptions): ClientMe
...
+  const containerMetadata: Record<string, any> = {};
+  if (isDocker == null) {
+    dockerPromise ??= fs.access('/.dockerenv');
+    try {
+      await dockerPromise;
+      isDocker = true;
+    } catch {
+      isDocker = false;
+    }
+  }
```
baileympearson, line 172: there's a million ways to skin a cat - but just because I think it's a bit cleaner not to cache to variables:

```suggestion
  dockerPromise ??= fs.access('/.dockerenv').then(() => true, () => false);
  const isDocker = await dockerPromise;
```

baileympearson (changes requested):

src/cmap/connect.ts
```diff
@@ -183,7 +183,7 @@ export interface HandshakeDocument extends Document {
   ismaster?: boolean;
   hello?: boolean;
   helloOk?: boolean;
-  client: ClientMetadata;
+  client: Document;
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Our changes make the `client.env.name` property optional (in the case that env exists), this would be a breaking change to the `public ClientMetadata` type, so it's just easier to change it to document type.

src/cmap/handshake/client_metadata.ts
aditi-khare-mongoDB: `fs.access` has a return type of Promise<void> not Promise<boolean> so I'm more comfortable with the current implementation as it stands.

baileympearson: > fs.access has a return type of Promise not Promise

could you clarify?

src/cmap/connect.ts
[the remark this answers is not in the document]
baileympearson: wow, I guess that was an oversight.  `ClientMetadata` probably should be `internal`

thanks for the context

src/cmap/handshake/client_metadata.ts
nbbeeken: (The angle brackets were interpreted as html, fixed)

baileympearson: I'm fine leaving it but I don't really understand the objection.  It resolves to undefined if the file is accessible and rejects otherwise - that's why we map it to a boolean

addaleax: Just pointing out, there _is_ a real difference between the current code and @baileympearson's suggestion; if you call this function twice in a row without waiting for the result, i.e. `getContainerMetadata(); getContainerMetadata();`, then the current code will call `fs.access` twice and the suggestion here would only call it once

src/cmap/connect.ts
aditi-khare-mongoDB: Caching the entire metadata object is a good idea, I'm good to implement this and store it as a module scoped variable.

src/cmap/handshake/client_metadata.ts
aditi-khare-mongoDB: I see, I'm good to switch to Bailey's implementation.

baileympearson: @addaleax I don't think so - because we use the `??=` operator.  We'll enter the if-statement twice, but after the first call the promise will no longer be undefined so we'll only call `fs.access` once.  Right?

addaleax: ah, right – I did miss that the original code actually did two checks, one on `isDocker` and then another one on `dockerPromise`. ignore me then :)

aditi-khare-mongoDB: I'm still okay with switching to Bailey's suggestion since the code is smaller.

nbbeeken: https://jira.mongodb.org/browse/NODE-5916

The ticket above intends to make this kind of code discouraged, of course, we can always exempt cases where it makes sense but I'll add that this creates a new promise rather than using the one already returned by `fs.access()`.

Fine by me to leave it as is for this PR, but it may come into question when that ticket has to correct the lint errors.

nbbeeken (approved):

src/cmap/handshake/client_metadata.ts
aditi-khare-mongoDB: good to know!

src/connection_string.ts
```diff
@@ -551,6 +551,7 @@ export function parseOptions(
   );
 
   mongoOptions.metadata = makeClientMetadata(mongoOptions);
+  mongoOptions.extendedMetadata = addContainerMetadata(mongoOptions.metadata);
```
baileympearson, line 554: we might want to add a `catch` block, like Anna [did when parallelizing dns lookup and srv resolution](https://github.com/mongodb/node-mongodb-native/pull/4012), so that this doesn't result in an unhandled promise rejection on the off change it rejects

src/cmap/handshake/client_metadata.ts
```diff
@@ -152,8 +153,60 @@ export function makeClientMetadata(options: MakeClientMetadataOptions): ClientMe
...
+async function getContainerMetadata() {
+  const containerMetadata: Record<string, any> = {};
+  if (isDocker == null) {
+    dockerPromise ??= fs.access('/.dockerenv').then(
+      () => true,
+      () => false
+    );
+    isDocker = await dockerPromise;
+  }
+
```
baileympearson, line 171: (optional) the original suggestion was intended to make it unnecessary to cache two variables.  but we can leave it as-is too.
```suggestion
  dockerPromise ??= fs.access('/.dockerenv').then(
    () => true,
    () => false
  );
  const  isDocker = await dockerPromise;
```

baileympearson (changes requested):
two small comments, otherwise LGTM.

src/connection_string.ts
nbbeeken: Good catch.

Why are we calling this up here again? seems like `addContainerMetadata` has what it needs down in prepareHandshake no?

baileympearson: We cache the promise during options parsing so that we only construct the extended metadata once.  Then we just `await` it during every call to connect.  Aditi filed https://jira.mongodb.org/browse/NODE-5994 to make `clientMetadata` private, remove `extendedMetadata` and just store its result in `metadata` in driver v7

nbbeeken: Ah ok, thanks for the context sorry I missed that.

src/cmap/handshake/client_metadata.ts
aditi-khare-mongoDB: Sounds good, made the change.

src/connection_string.ts
aditi-khare-mongoDB: Added in the catch block.

baileympearson (approved):

nbbeeken (approved):

PR #4010, opened 2024-03-01
fix(NODE-5981): read preference not applied to commands properly

alenakhineika:
### Description
Add `$readPrefernce` to commands according to [spec](https://github.com/mongodb/specifications/blob/5de9d45528a07d51579253f57f7ba23dbbaa68d1/source/server-selection/server-selection.md#topology-type-single).

**Topology type: Single**

A deployment of topology type Single contains only a single server of any type. Topology type Single signifies a direct connection intended to receive all read and write operations.

Therefore, read preference is ignored during server selection with topology type Single. The single server is always suitable for reads if it is available. Depending on server type, the read preference is communicated to the server differently:
 * Type Mongos: the read preference is sent to the server using the rules for [Passing read preference to mongos and load balancers](https://github.com/mongodb/specifications/blob/5de9d45528a07d51579253f57f7ba23dbbaa68d1/source/server-selection/server-selection.md#passing-read-preference-to-mongos-and-load-balancers).
 * Type Standalone: clients MUST NOT send the read preference to the server
 * For all other types, using OP_QUERY: clients MUST always set the SecondaryOk wire protocol flag on reads to ensure that any server type can handle the request.
 * For all other types, using OP_MSG: If no read preference is configured by the application, or if the application read preference is Primary, then $readPreference MUST be set to \{ "mode": "primaryPreferred" } to ensure that any server type can handle the request. If the application read preference is set otherwise, $readPreference MUST be set following [Document structure](https://github.com/mongodb/specifications/blob/5de9d45528a07d51579253f57f7ba23dbbaa68d1/source/server-selection/server-selection.md#document-structure).

The single server is always suitable for write operations if it is available.

#### What is changing?

**Implementation**
- When preparing a command to the server / `prepareCommand()`, apply `$readPrefernce` to commands according to spec
- Pass the `directConnection` option when we create a connection in the connection pool

**Tests**
- Update test runner to allow overwriting host and port in configuration URL in order to connect separately to each host in replicaset.
- Convert `test/integration/server-selection/readpreference.test.js` to TypeScript
- Unskip `readPreference` tests (fixes[ NODE-5263](https://jira.mongodb.org/browse/NODE-5263))
- Sync recent `run-command` spec tests
- Run max staleness on replicaset, because accordion to spec, clients MUST NOT send the read preference to the standalone server

None

[NODE-5981](https://jira.mongodb.org/browse/NODE-5981)

### Fixed applying read preference to commands depending on topology
When connecting to a secondary in a replica set with a direct connection, if a read operation is performed, the driver attaches a read preference of `primaryPreferred` to the command.


test/integration/server-selection/readpreference.test.ts
```diff
@@ -463,4 +457,214 @@ describe('ReadPreference', function () {
       await client.close();
     }
   });
+
+  context('when connecting to a secondary in a replica set with a direct connection', () => {
+    context('when readPreference is primary', () => {
+      it('should attach a read preference of primaryPreferred to the read command for replicaset', {
+        metadata: { requires: { topology: 'replicaset', mongodb: '>=3.6' } },
```
nbbeeken, line 464: We don't test below 3.6 so we could clean up the: `, mongodb: '>=3.6'`

test/integration/server-selection/readpreference.test.ts
```diff
@@ -463,4 +457,214 @@ describe('ReadPreference', function () {
...
+        test: async function () {
+          if (this.configuration.topologyType !== 'ReplicaSetWithPrimary') {
+            return this.skip();
```
nbbeeken, line 467: Should we set a skipReason here? `this.skipReason = '...'` will print out why this test didn't run

test/integration/server-selection/readpreference.test.ts
```diff
@@ -463,4 +457,214 @@ describe('ReadPreference', function () {
...
+              }
+            });
+            let serverStatus;
+
+            try {
+              const admin = client.db().admin();
+              serverStatus = await admin.serverStatus();
+            } catch (serverStatusError) {
+              expect(serverStatusError).to.not.exist;
+            }
```
nbbeeken, line 603: We could just let the serverStatus function throw to fail the test instead of this assertion right?

test/integration/server-selection/readpreference.test.ts
```diff
@@ -463,4 +457,214 @@ describe('ReadPreference', function () {
...
+            if (serverStatus.repl.secondary) {
+              let serverError;
+
+              try {
+                await client.db('test').collection('test').findOne({ a: 1 });
+              } catch (error) {
+                serverError = error;
+              }
+
+              expect(serverError).to.not.exist;
```
nbbeeken, line 614: Same here, if findOne throws then the test fails already

test/integration/server-selection/readpreference.test.ts
```diff
@@ -463,4 +457,214 @@ describe('ReadPreference', function () {
...
+              expect(events[0]).to.containSubset({
+                commandName: 'find',
+                command: {
+                  $readPreference: { mode: 'secondary' }
+                }
+              });
```
nbbeeken, line 620: I don't love containSubset it could overlook mistakes that ideally we'd want to be more strict about in our assertions. If I add `events[0].command.$readPreference.funnyMode = 'haha'` this test still passes because it still satisfied the subset, but I think we want to not only make sure readpref is attached but that it is attached correctly. But I don't expect us to make a big manual copy of the command just to check eq, you can get something susinct with has deep property checks:

```ts
expect(events[0]).to.have.property('commandName', 'find');
expect(events[0]).to.have.deep.nested.property('command.$readPrefrence', { mode: 'secondary' })
```

test/integration/max-staleness/max_staleness.test.js
```diff
@@ -46,33 +46,31 @@ describe('Max Staleness', function () {
...
+    test: function () {
+      const self = this;
       const configuration = this.configuration;
       const client = configuration.newClient(
         `mongodb://${test.server.uri()}/test?readPreference=secondary&maxStalenessSeconds=250`,
         { serverApi: null } // TODO(NODE-3807): remove resetting serverApi when the usage of mongodb mock server is removed
       );
 
-      client.connect(function (err, client) {
+      client.connect(async function (err, client) {
```
nbbeeken, line 61: Passing an async function into connect I don't think will propogate any rejections correctly. you could either finish converting the test function to async/await or move this function to a `.then` chain, and make sure you `return` the promise

test/integration/max-staleness/max_staleness.test.js
```diff
@@ -81,36 +79,38 @@ describe('Max Staleness', function () {
...
       const client = configuration.newClient(`mongodb://${test.server.uri()}/test`, {
         serverApi: null // TODO(NODE-3807): remove resetting serverApi when the usage of mongodb mock server is removed
       });
-      client.connect(function (err, client) {
+
+      try {
+        await client.connect();
+      } catch (err) {
         expect(err).to.not.exist;
+      }
```
nbbeeken, line 96: If connect rejects then the test ends, redundant try/catch assertion

test/integration/max-staleness/max_staleness.test.js
```diff
@@ -157,35 +164,39 @@ describe('Max Staleness', function () {
...
+      } catch (err) {
+        expect(err).to.not.exist;
+      }
+
+      expect(test.checkCommand).to.containSubset({
+        $query: { find: 'test', filter: {} },
+        $readPreference: { mode: 'secondary', maxStalenessSeconds: 250 }
       });
+
+      client.close();
```
nbbeeken, line 199: need to await this right?

src/cmap/connection_pool.ts
```diff
@@ -634,7 +636,8 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       generation: this[kGeneration],
       cancellationToken: this[kCancellationToken],
       mongoLogger: this.mongoLogger,
-      authProviders: this[kServer].topology.client.s.authProviders
+      authProviders: this[kServer].topology.client.s.authProviders,
+      dierctConnection: this[kServer].topology.client.options.directConnection === true
```
nbbeeken, line 640:
```suggestion
      directConnection: this[kServer].topology.client.options.directConnection === true
```

src/cmap/connection.ts
```diff
@@ -120,6 +121,7 @@ export interface ConnectionOptions
   metadata: ClientMetadata;
   /** @internal */
   mongoLogger?: MongoLogger | undefined;
+  dierctConnection?: boolean;
```
nbbeeken, line 124:
```suggestion
  /** @internal */
  directConnection?: boolean;
```

src/cmap/connection_pool.ts
```diff
@@ -100,6 +100,8 @@ export interface ConnectionPoolOptions extends Omit<ConnectionOptions, 'id' | 'g
   waitQueueTimeoutMS: number;
   /** If we are in load balancer mode. */
   loadBalanced: boolean;
+  /** If topology contains only a single server of any type. */
```
nbbeeken, line 103:
```suggestion
  /** @internal If topology contains only a single server of any type. */
```

nbbeeken (changes requested):

src/cmap/connection.ts
```diff
@@ -122,6 +123,8 @@ export interface ConnectionOptions
   extendedMetadata: Promise<Document>;
   /** @internal */
   mongoLogger?: MongoLogger | undefined;
+  /** @internal */
+  directConnection: boolean;
```
baileympearson, line 127: Could we pass `directConnection` as an argument to `Connection.command()` from `Server.command()` instead of storing `directConnection` on the stream description, connection pool and (with a getter) on the Connection?

baileympearson: alternatively, we already have some read preference logic in `Server.command()` (that's where we apply omitReadPreference) - perhaps we could just attach the read preference to commands there instead of in the connection class.

nbbeeken (approved):

src/cmap/connection.ts
baileympearson: With the exception of `loadBalanced` (which should not live on the stream description in my opinion), the stream description contains information about the connection.  `directConnection` is a client option that contains information about the topology (which is why it's stored on the topology).  In terms of code organization, it makes more sense to store this information on the topology.  

That also removes the need to store it on the connection pool, because we wouldn't need to assign it on the connection's description when the connection is created.

nbbeeken: To this point, I wish the pool/monitor was accessible from within a connection so we could reach up for this setting rather than passing it down

But I'm also fine with it on the description, the description represents a state which controls certain things about command construction (like if hello is ok) so controlling rp from it makes sense but I admit it is inconsistently deployed (à la ServerCapabilities topology.ts)

alenakhineika: But connection.ts does not have access to topology, only from the session, but the session does not always exist. You would need to do first what Neal said - refactor to access the pool/monitor from within a connection.

baileympearson: Do we need to attach read preference to non-reads/writes (such as the authentication handshake)?  My thought is no - right?  

If so, the only places we need to attach read preference to commands are called from `Server.command()` - which does have access to the topology.

nbbeeken: I recall someone alluding to something about always attaching RP for simplicity, reads/writes etc all of the above so the driver doesn't need to discern between commands

nbbeeken (approved):

baileympearson (approved):

PR #4011, opened 2024-03-01
feat(NODE-5967): container and Kubernetes awareness in client metadata

aditi-khare-mongoDB:
**Note:** This PR's description and changes are  almost identical to the one on the 6.x branch (NODE-5968).

### Description
Track user usage of containers on 5.x branch.

#### What is changing?
When kubernetes is in `process.env`, we add a  `{orchestrator: 'kubernetes'}` field to the `client.env` field of the handshake document (if this does not make the document exceed 512 bytes).

When '/.dockerenv' exists, we add a  `{runtime: 'docker'}` field to the `client.env` field of the handshake document (if this does not make the document exceed 512 bytes).

Note: We do not unit test docker, but I tested our logic for detecting docker on an EVG Host, ran Docker and detection succeeds for both root and non-root user cases.

No, non-user facing change. 

Product wants to track user usage of docker and kubernetes.

### Container and Kubernetes Awareness

The Node.js driver now keeps track of container metadata in the `client.env.container` field of the handshake document. 

If space allows, the following metadata will be included in `client.env.container`:

```
env?: { 
  container?: {
    orchestrator?: 'kubernetes' // if process.env.KUBERNETES_SERVICE_HOST is set
    runtime?: 'docker' // if the '/.dockerenv' file exists
  } 
}
```

Note: If neither Kubernetes nor Docker is present, `client.env` will not have the `container` property.

### DIFF
--- a/src/cmap/connect.ts
+++ b/src/cmap/connect.ts
@@ -25,7 +25,6 @@ import {
   type ConnectionOptions,
   CryptoConnection
 } from './connection';
-import { addContainerMetadata } from './handshake/client_metadata';
 import {
   MAX_SUPPORTED_SERVER_VERSION,
   MAX_SUPPORTED_WIRE_VERSION,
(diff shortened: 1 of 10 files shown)


nbbeeken (dismissed):
LGTM, I figure the txn test skips will merge very soon and we can rerun this PR on top of that to double check. 

Side note, since this is merging into 5.x we don't need to mention 5.x in the title or release notes because of where it destined for it already has the context of the branch.

nbbeeken (approved):

PR #4014, opened 2024-03-04
fix(NODE-5985): throw Nodejs' certificate expired error when TLS fails to connect instead of `CERT_HAS_EXPIRED`

baileympearson:
#### What is changing?

When enabled, `rejectUnauthorized` configures Node's TLS API to reject / error when its unable to verify the certificates of the server.  This is enabled by default.  When the option is enabled and Node is unabled to verify the certificates of the server, an error event is emitted.  In our connect logic, this means the `connect` promise rejects.

Prior to https://github.com/mongodb/node-mongodb-native/pull/3973, the code removed in this PR was in the connect handler (called on successful TLS connection).  This was dead code because we will never receive an error that is triggered by `rejectUnauthorized` and successfully connect.  When https://github.com/mongodb/node-mongodb-native/pull/3973 refactored `connect()` to be async, we moved this logic into the `catch` block instead of the `try` block, so that it _did_ run when an error occurred during TLS connection.

The code throws `socket.authorizationError` instead of the caught error, which is a string value (https://nodejs.org/api/tls.html#tlssocketauthorizationerror).  As a result, instead of throwing the error Nodejs gave us, we throw the authorization error property from the socket.

No.

### Release Highlight

Driver `6.3.0` included an internal refactor to the driver's TLS connection logic that introduced logic that intercepted TLS connection errors.  In certain situations, the driver would erroneously throw the TLS socket's `authorizationError` property instead of the error thrown from Nodejs' TLS API.

This was observable in two ways:

- The error message for TLS connection issues changed.  For example, instead of `certificate has expired`, the driver threw an error with the message `CERT_HAS_EXPIRED`
- The error thrown from the driver had a `cause` property set to a string instead of an error.

The driver now correctly propagates TLS errors.

### DIFF
--- a/src/cmap/connect.ts
+++ b/src/cmap/connect.ts
@@ -375,10 +375,6 @@ export async function makeSocket(options: MakeConnectionOptions): Promise<Stream
     return socket;
   } catch (error) {
     socket.destroy();
-    if ('authorizationError' in socket && socket.authorizationError != null && rejectUnauthorized) {
-      // TODO(NODE-5192): wrap this with a MongoError subclass
-      throw socket.authorizationError;
-    }
     throw error;
   } finally {
     socket.setTimeout(0);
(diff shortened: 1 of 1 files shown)


nbbeeken (changes requested):
Change LGTM, can you make the title and release notes reflect the bug being fixed, which I think is mainly the error message, but also that there is a cause being set to a string (while not broken, highly unexpected)

nbbeeken (approved):

PR #4016, opened 2024-03-04
feat(NODE-5988): Provide access to raw results doc after MongoServerError

aditi-khare-mongoDB:
### Description
Provide access to raw results doc in MongoServerError.

#### What is changing?
See release highlight.

Yes, API docs. 

Provide support for all values returned by server error.

### Add property errorResponse to MongoServerError 
The `MongoServerError.errorResponse` property now contains the raw error document returned by the server.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -201,7 +201,7 @@ export class MongoError extends Error {
  */
 export class MongoServerError extends MongoError {
   /** Raw error result document returned by server. */
-  errorResponse?: ErrorDescription;
+  errorResponse: ErrorDescription;
   codeName?: string;
   writeConcernError?: Document;
   errInfo?: Document;
--- a/src/error.ts
+++ b/src/error.ts
@@ -200,6 +200,7 @@ export class MongoError extends Error {
  * @category Error
  */
 export class MongoServerError extends MongoError {
+  /** Raw error result document returned by server. */
   errorResponse?: ErrorDescription;
   codeName?: string;
   writeConcernError?: Document;
--- a/src/error.ts
+++ b/src/error.ts
@@ -200,6 +200,7 @@ export class MongoError extends Error {
  * @category Error
  */
 export class MongoServerError extends MongoError {
+  errorResponse?: ErrorDescription;
   codeName?: string;
   writeConcernError?: Document;
   errInfo?: Document;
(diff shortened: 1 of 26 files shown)


test/integration/crud/crud.spec.test.js
```diff
@@ -425,6 +425,10 @@ describe('CRUD spec v1', function () {
   }
 });
 
+// The Node driver does not have a Collection.modifyCollection helper.
+const SKIPPED_TESTS = ['findOneAndUpdate document validation errInfo is accessible'];
```
baileympearson, line 429: These tests are different from the collection management specs.  These tests use the `modifyCollection` operation to set up the collection for the test, but they're not testing the driver's `modifyCollection` helper.  

I think it might make sense to implement a unified test format helper for `modifyCollection` that uses `runCommand`to send a `collMod` to the server (i.e., a new operation in `operations.ts`).  Then we could unskip this test.  We'd still leave the `modifyCollection` tests alone though because we don't have a driver helper.  

what do you think? 

(also see https://github.com/mongodb/specifications/pull/1316/files#r994171436)

src/error.ts
```diff
@@ -200,6 +200,8 @@ export class MongoError extends Error {
  * @category Error
  */
 export class MongoServerError extends MongoError {
+  /** Raw error result document returned by server. */
+  errorResponse?: ErrorDescription;
```
baileympearson, line 204: Should this be required?  We always assign it in the constructor

baileympearson (changes requested):
two small comments, otherwise looks good.  Nice work!

test/integration/crud/crud.spec.test.js
aditi-khare-mongoDB: Perhaps we could merge these changes and make a separate subtask or even ticket for adding in the modifyCollection changes?

src/error.ts
aditi-khare-mongoDB: Yep, just changed it.

aditi-khare-mongoDB: Ideally it should be. There were some type errors since we feed `MongoError` instances into `MongoServerError` fields in parts of the driver, so I just casted these instances as `MongoServerError` as a fix.

baileympearson: Can you file a subtask to fix the type errors this uncovered (tackling it as follow-up is fine since it's a separate issue)?  I think it's a bug in the driver's types but we'll need to look to be sure - ideally we'd fix the type issue instead of just casting it away.

We have a number of instances of `MongoServerError` that are actually invoked with `MongoError`s, which worked before because `MongoError` was assignable to `MongoServerError`.

test/integration/crud/crud.spec.test.js
baileympearson: Sounds good to me - if we file a separate ticket, can we add a `TODO(NODE-XXXX)` comment here so we don't forget about it?

src/error.ts
aditi-khare-mongoDB: Sure, here's the link [to the ticket]( https://jira.mongodb.org/browse/NODE-5999).

baileympearson (approved):

PR #4020, opened 2024-03-05
refactor(NODE-5914): topology.selectServer to async-await

W-A-James:
#### What is changing?

- Refactored `Topology.selectServer` to async/await syntax
- Refactored `Topology.connect` to async/await syntax and add connection lock
- Refactored all references to `Topology.selectServer` in the driver including

No

NODE-5914


test/unit/sdam/topology.test.js
```diff
@@ -530,9 +509,7 @@ describe('Topology (unit)', function () {
         //     returning their value
         //   - verify that 10 callbacks were called
 
-        topology.connect(err => {
-          expect(err).to.not.exist;
-
+        topology.connect().then(async () => {
```
nbbeeken, line 512: I don't think this is the correct conversion iiuc. Before we would call finish which calls done to complete the test. But this promise is floating, the test isn't blocked on it.

src/sdam/topology.ts
```diff
@@ -459,40 +460,44 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
+      if (!skipPingOnConnect && server && this.s.credentials) {
+        try {
+          await server.command(ns('admin.$cmd'), { ping: 1 }, {});
+          stateTransition(this, STATE_CONNECTED);
+          this.emit(Topology.OPEN, this);
+          this.emit(Topology.CONNECT, this);
+
+          return this;
+        } catch (error) {
+          exitWithError(error);
```
nbbeeken, line 486: Previously, we conditionally switched between calling a callback or emitting error dependent on whether a callback was provided. Ever since we wrapped this in promisify we were always providing a callback. Is it correct that we now always emit an error? If not then I think we can get rid of the try catch and let things throw on their own.

src/sdam/topology.ts
```diff
@@ -964,7 +970,7 @@ function processWaitQueue(topology: Topology) {
           )
         );
       }
-      waitQueueMember.callback(e);
+      waitQueueMember.reject(e);
```
nbbeeken, line 973: Could we improve the variable name here? `selectorError` or something

src/sdam/topology.ts
```diff
@@ -911,7 +917,7 @@ function drainWaitQueue(queue: List<ServerSelectionRequest>, err?: MongoDriverEr
           );
         }
       }
-      waitQueueMember.callback(err);
+      waitQueueMember.reject(err);
```
nbbeeken, line 920: maybe `drainError`, this is actually always going to be `MongoTopologyClosedError` at least currently, maybe something alluding to that `shutdownError`

src/sdam/topology.ts
```diff
@@ -1027,7 +1033,7 @@ function processWaitQueue(topology: Topology) {
           )
         );
       }
-      waitQueueMember.callback(error);
+      waitQueueMember.reject(error);
```
nbbeeken, line 1036: Maybe `serverSelectionError`?

src/sdam/topology.ts
```diff
@@ -105,7 +105,8 @@ export interface ServerSelectionRequest {
   mongoLogger: MongoLogger | undefined;
   transaction?: Transaction;
   startTime: number;
-  callback: ServerSelectionCallback;
+  resolve: (server: Server) => void;
+  reject: (error?: AnyError) => void;
   [kCancelled]?: boolean;
```
nbbeeken, line 110: I don't think this should be optional we wouldn't want a promise to reject without an error, if there's anywhere where we may call this with something nullish we should make sure 1. is that actually true or does the TS need fixing 2. if it is true we should make a runtime error that says something about the unexpected nullishness of something.

test/unit/assorted/server_selection_latency_window_utils.ts
```diff
@@ -139,10 +138,7 @@ export async function runServerSelectionLatencyWindowTest(test: ServerSelectionL
   const selectedServers: Server[] = [];
 
   for (let i = 0; i < test.iterations; ++i) {
-    const server: Server = await promisify(topology.selectServer.bind(topology))(
-      ReadPreference.NEAREST,
-      {}
-    );
+    const server: Server = await topology.selectServer.bind(topology)(ReadPreference.NEAREST, {});
```
nbbeeken, line 141: We don't need the bind here anymore :)

nbbeeken (changes requested):
Some relatively small cleanups, this looks great!

nbbeeken (approved):

test/unit/sdam/topology.test.js
```diff
@@ -457,30 +437,33 @@ describe('Topology (unit)', function () {
...
-          expect(requestCheck).property('callCount').to.equal(1);
+            // When server is created `connect` is called on the monitor. When server selection
+            // occurs `requestCheck` will be called for an immediate check.
+            expect(requestCheck).property('callCount').to.equal(1);
 
-          topology.close();
-          done();
-        });
+            topology.close();
+            done();
```
nbbeeken, line 465: Still using done here we should switch to returing the promise. This test is crashing the unit runs because the call count is 5 🤔

nbbeeken (changes requested):

nbbeeken (approved):

test/unit/sdam/topology.test.js
```diff
@@ -450,38 +430,47 @@ describe('Topology (unit)', function () {
...
+            expect(err).to.exist;
+            expect(err).to.match(/Server selection timed out/);
+            expect(err).to.have.property('reason');
 
-          // When server is created `connect` is called on the monitor. When server selection
-          // occurs `requestCheck` will be called for an immediate check.
-          expect(requestCheck).property('callCount').to.equal(1);
+            // When server is created `connect` is called on the monitor. When server selection
+            // occurs `requestCheck` will be called for an immediate check.
+            expect(requestCheck).property('callCount').to.equal(1);
```
nbbeeken, line 466: Test failing, change req for viz.

nbbeeken (changes requested):

nbbeeken (approved):

test/unit/sdam/topology.test.ts
```diff
@@ -442,86 +408,81 @@ describe('Topology (unit)', function () {
...
+      // When server is created `connect` is called on the monitor. When server selection
+      // occurs `requestCheck` will be called for an immediate check.
+      expect(requestCheck).to.have.been.calledOnce;
+      topology.close();
     });
 
-    it('should disallow selection when the topology is explicitly closed', function (done) {
-      const topology = topologyWithPlaceholderClient('someserver:27019');
-      this.sinon.stub(Server.prototype, 'connect').callsFake(function () {
+    it('should disallow selection when the topology is explicitly closed', async function () {
```
nbbeeken, line 445: We are so close, 🤏🏻  unit failure here.

nbbeeken (changes requested):

src/sdam/topology.ts
```diff
@@ -411,17 +402,26 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
+      return this.connectionLock;
+    }
+
+    try {
+      this.connectionLock = this._connect(options);
+      await this.connectionLock;
+    } finally {
+      // release
+      this.connectionLock = undefined;
+    }
```
baileympearson, line 416: (optional) thoughts?

```suggestion
    this.connectionLock ??= this._connect(options);
    try {
    	await this.connectionLock;
    	return this;
    } finally {
    	this.connectionLock = undefined;
    }
```

src/sdam/topology.ts
```diff
@@ -883,7 +877,7 @@ function updateServers(topology: Topology, incomingServerDescription?: ServerDes
   }
 }
 
-function drainWaitQueue(queue: List<ServerSelectionRequest>, err?: MongoDriverError) {
+function drainWaitQueue(queue: List<ServerSelectionRequest>, drainError: MongoDriverError) {
```
baileympearson, line 880: 🔥 I always appreciate drive-by improvement of variable names, thanks for improving this one!

lack of attention to variable names leads to death of readability by a thousand cuts 🔪

test/unit/assorted/server_selection_spec_helper.js
```diff
@@ -102,104 +102,84 @@ function executeServerSelectionTest(testDefinition, testDone) {
...
+        continue;
       }
+      expect(selectedServerDescription[prop]).to.deep.equal(
+        value,
+        `Mismatched selected server "${prop}"`
+      );
+    }
+    return;
+  } catch (err) {
+    // are we expecting an error?
```
baileympearson, line 174: tbh I thought this was intended to be a question for the reviewers 😅 

I don't feel strongly but maybe making it a statement instead of a question would be a little more clear

baileympearson (changes requested):

src/sdam/topology.ts
nbbeeken:
```suggestion
    try {
      return await (this.connectionLock ??= this._connect(options));
    } finally {
      this.connectionLock = undefined;
    }
```

Let's play ⛳ (kidding... mostly 👀)

baileympearson:
```suggestion
    return await (this.connectionLock ??= this._connect(options)).finally(() => this.connectionLock = undefined);
```

😤

nbbeeken (changes requested):
just a formatting lint 🎉

nbbeeken (approved):

baileympearson (approved):

PR #4021, opened 2024-03-05
refactor(NODE-5915): refactor topology close logic to be synchronous

W-A-James:
#### What is changing?

* Refactored the following functions/methods to be synchronous:
	* `Connection.destroy`
	* `ConnectionPool.close`
	* `Server.destroy`
	* `Topology.close`
* Removed `DestroyOptions` as the `force` flag is unused in all the places it would have been passed in
* Reordered connection check in and call to `Connection.onError` in `ConnectionPool.interruptInUseConnections` due to failure on this test
* Updated relevant tests to accommodate newly synchronous functions 

No

NODE-5915


src/cmap/connection_pool.ts
```diff
@@ -493,25 +492,16 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
   private interruptInUseConnections(minGeneration: number) {
     for (const connection of this[kCheckedOut]) {
       if (connection.generation <= minGeneration) {
-        this.checkIn(connection);
         connection.onError(new PoolClearedOnNetworkError(this));
+        this.checkIn(connection);
```
W-A-James, line 496: @baileympearson this was done to fix a test failure where we were getting an uncaught error when running [this spec test](https://github.com/mongodb/specifications/blob/345c243748390f17072754a0b07bc2dd4580b8c9/source/server-discovery-and-monitoring/tests/unified/interruptInUse-pool-clear.yml#L232)

src/cmap/connection_pool.ts
aditi-khare-mongoDB, line 1: Let's file a follow-up ticket in the V7 epic to remove closeOptions since its no longer used anywhere in the code base after these changes.

src/cmap/connection_pool.ts
```diff
@@ -28,7 +28,6 @@ import { CancellationToken, TypedEventEmitter } from '../mongo_types';
 import type { Server } from '../sdam/server';
 import {
   type Callback,
-  eachAsync,
```
aditi-khare-mongoDB, line 31: This utils function is now also used nowhere in the codebase, do we want to remove it? It uses callbacks, and IIUC we no longer want to add any new callback code to the driver.

aditi-khare-mongoDB (changes requested):
left a few small comments, otherwise LGTM, nice work :)

src/cmap/connection_pool.ts
W-A-James: Good callout here. I can file the follow-up and the first subtask of that follow-up can be to mark it as deprecated.

src/cmap/connection_pool.ts
W-A-James: Good point. I agree that removing it is likely the way we want to go here. @nbbeeken @baileympearson thoughts?

nbbeeken: Go for it

src/cmap/connection_pool.ts
W-A-James: Looking a little closer, it seems the only places that interface was used are in `Topology.close` and `ConnectionPool.close` which are both internal classes, so users shouldn't be interacting with this interface at all. Seems like it wouldn't be a breaking change to just remove it in this PR along with `DestroyOptions`.

baileympearson: Unfortunately, even though it's only used in internal classes / types, it's public and exported so technically it's a part of our API. In situations like this in the past, we've considered it breaking to remove and done it in a major release

aditi-khare-mongoDB (approved):

W-A-James:
Follow-up ticket: NODE-6007

src/mongo_client.ts
```diff
@@ -614,19 +614,12 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> {
...
-        }
-        resolve();
-      });
-    });
+    topology.close();
+
+    const { encrypter } = this[kOptions];
+    if (encrypter) {
+      await encrypter.close(this, force);
+    }
```
baileympearson, line 622: can `close()` throw?  If so, both the new and old implementations potentially would leave the `encrypter` open if `topology.close()` throws

src/sdam/server.ts
```diff
@@ -257,13 +246,9 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
-    this.pool.close(options, err => {
-      stateTransition(this, STATE_CLOSED);
-      this.emit('closed');
-      if (typeof callback === 'function') {
-        callback(err);
-      }
-    });
+    this.pool.close();
+    stateTransition(this, STATE_CLOSED);
+    this.emit('closed');
```
baileympearson, line 251: same question as with `client.close()` - can `pool.close()` throw?

src/sdam/topology.ts
```diff
@@ -469,7 +469,8 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
       selectServerOptions,
       (err, server) => {
         if (err) {
-          return this.close({ force: false }, () => exitWithError(err));
+          this.close();
```
baileympearson, line 472: same question about errors

src/cmap/connection_pool.ts
baileympearson: Yeah, this is right.  Good catch.

I think the reason this worked in the past was because

1. `checkIn` calls `destroyConnection()`, which calls `connection.destroy()` in a process.nextTick()
2. then `connection.onError() runs and shuts down the connection 
3. then `connection.destroy()` is a no-op because the connection has already been destroyed

You made `destroyConnection` synchronous, so we need to run `onError` first to ensure the connection is closed with the correct error.  Then we run checkIn, which resets the pool state and calls `connection.destroy()`, which is a no-op.

baileympearson (changes requested):

src/sdam/topology.ts
W-A-James: |method|relevant direct dependencies|throws?|
|------|------------|-------|
|`Connection.destroy`| - | No |
|`ConnectionPool.close`| `Connection.destroy` | No |
|`Server.destroy` | `stateTransition`\*, `ConnectionPool.close` | No; `stateTransition`\* can technically throw, but `Server.destroy` is written such that it will never try to make an illegal state transition.|
|`destroyServer` | `Server.destroy` | No |
|`Topology.close` | `destroyServer`, `stateTransition`\*\* | No; `stateTransition`\*\* can technically throw, but like in `Server.destroy`, `Topology.close` is written such that it never tries to make an illegal state transition|

Note:

`stateTransition`\* refers to the local function using the static state machine defined for the `Server` class created by the `makeStateMachine` helper.

`stateTransition`\*\* refers to the local function using the static state machine defined for the `Topology` class created by the `makeStateMachine` helper.

nbbeeken (approved):

src/mongo_client.ts
baileympearson: Thanks for the investigation into whether these can throw or not.

doesn't have to be in this PR but do you think it might be worth handling errors anyways, since we could introduce errors into `close()` in the future?

baileympearson (approved):

PR #4022, opened 2024-03-06
fix(NODE-5993): memory leak in the `Connection` class

nbbeeken:
#### What is changing?

- Remove promise wrapping of signal
- Make writing async based on whether or not write returns true
- Remove signal

The Connection class stored a promise that would reject when the signal was aborted. This caused a leak that would create new promises unbounded. I've removed the promise wrapping opting for a simpler implementation that only makes writing async if there was buffering. If there was buffering we wait for the drain event and use try/finally to make sure we clean up event listeners.

### Fixed memory leak in Connection layer

The Connection class has recently been refactored to operate on our socket operations using promises. An oversight how we made async network operations interruptible made new promises for every operation. We've simplified the approach and corrected the leak.

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -36,7 +36,7 @@ import {
   maxWireVersion,
   type MongoDBNamespace,
   now,
-  promiseWithResolvers,
+  once,
   uuidV4
 } from '../utils';
 import type { WriteConcern } from '../write_concern';
@@ -333,6 +333,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     if (error) {
       this.error = error;
       this.dataEvents?.throw(error).then(undefined, () => null); // squash unhandled rejection
+    } else {
+      this.dataEvents?.return().then(undefined, () => null); // squash unhandled rejection
     }
     this.closed = true;
     this.emit(Connection.CLOSE);
@@ -599,17 +601,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     const buffer = Buffer.concat(await finalCommand.toBin());
 
     if (this.socket.write(buffer)) return;
-
-    const { promise: drained, resolve, reject } = promiseWithResolvers<void>();
-    const onDrain = () => resolve();
-    const onError = (error: Error) => reject(error);
-
-    this.socket.once('drain', onDrain).once('error', onError);
-    try {
-      return await drained;
-    } finally {
-      this.socket.off('drain', onDrain).off('error', onError);
-    }
+    return once(this.socket, 'drain');
   }
 
   /**
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -623,7 +623,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
    */
   private async *readMany(): AsyncGenerator<OpMsgResponse | OpQueryResponse> {
     try {
-      this.dataEvents = this.dataEvents = onData(this.messageStream);
+      this.dataEvents = onData(this.messageStream);
       for await (const message of this.dataEvents) {
         const response = await decompressResponse(message);
         yield response;
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -1,6 +1,5 @@
 import { type Readable, Transform, type TransformCallback } from 'stream';
 import { clearTimeout, setTimeout } from 'timers';
-import { promisify } from 'util';
 
 import type { BSONSerializeOptions, Document, ObjectId } from '../bson';
 import type { AutoEncrypter } from '../client-side-encryption/auto_encrypter';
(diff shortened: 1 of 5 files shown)


test/integration/connection-monitoring-and-pooling/connection.test.ts
```diff
@@ -197,6 +203,84 @@ describe('Connection', function () {
...
+
+          // Spy on the socket event listeners
+          const addedListeners: string[] = [];
+          const removedListeners: string[] = [];
+          socket
+            .on('removeListener', name => removedListeners.push(name))
+            .on('newListener', name => addedListeners.push(name));
+
+          // Make server sockets block
+          for (const s of mockServer.sockets) s.pause();
```
nbbeeken, line 262: @addaleax I did not end up proxying to MongoDB but I did figure out how to make the server-side socket hold up the flow, seems to work consistently, TY again

test/integration/node-specific/resource_clean_up.test.ts
```diff
@@ -86,4 +89,34 @@ describe('Driver Resources', () => {
...
+    it('does not leave behind additional promises', async () => {
+      const test = client.db('test').collection('test');
+      const promiseCountBefore = v8.queryObjects(Promise, { format: 'count' });
+      for (let i = 0; i < 100; i++) {
+        await test.findOne();
+      }
+      await sleep(10);
+      const promiseCountAfter = v8.queryObjects(Promise, { format: 'count' });
+
+      expect(promiseCountAfter).to.be.within(promiseCountBefore - 5, promiseCountBefore + 5);
```
nbbeeken, line 119: This test does not run in our CI, but it will someday, I used Node.js v22 locally and this passes.

src/cmap/connection.ts
```diff
@@ -632,13 +620,19 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
    * Note that `for-await` loops call `return` automatically when the loop is exited.
    */
   private async *readMany(): AsyncGenerator<OpMsgResponse | OpQueryResponse> {
-    for await (const message of onData(this.messageStream, { signal: this.signal })) {
-      const response = await decompressResponse(message);
-      yield response;
+    try {
+      this.dataEvents = this.dataEvents = onData(this.messageStream);
```
nbbeeken, line 624: Surprised there isn't a lint rule for this 😞

src/cmap/connection.ts
```diff
@@ -619,7 +596,18 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
 
     const buffer = Buffer.concat(await finalCommand.toBin());
 
-    return this.socketWrite(buffer);
+    if (this.socket.write(buffer)) return;
+
+    const { promise: drained, resolve, reject } = promiseWithResolvers<void>();
+    const onDrain = () => resolve();
+    const onError = (error: Error) => reject(error);
```
nbbeeken, line 603: Creating functions here to prevent extraneous arguments from reaching the promise completion functions.

addaleax (approved):

src/cmap/wire_protocol/on_data.ts
```diff
@@ -18,9 +18,7 @@ type PendingPromises = Omit<
  * Returns an AsyncIterator that iterates each 'data' event emitted from emitter.
  * It will reject upon an error event or if the provided signal is aborted.
```
baileympearson, line 19:
```suggestion
 * It will reject upon an error event.
```

test/integration/node-specific/resource_clean_up.test.ts
nbbeeken: Yes, we go from 100s of promises before the test starts to 400s

src/cmap/connection.ts
```diff
@@ -355,7 +328,11 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     }
 
     this.socket.destroy();
-    this.controller.abort(error);
+    if (error) {
+      this.error = error;
+      this.dataEvents?.throw(error).then(undefined, () => null); // squash unhandled rejection
+    }
```
baileympearson, line 336: what happens if we close a connection without an error when an operation is in progress?  Right now, I think we would do nothing.

Should we instead always set an error (creating one if one doesn't exist?)?
```typescript
this.error = error ?? new MongoNetworkError("connection shutdown");
this.dataEvents?.throw(this.error).then(undefined, () => null); // squash unhandled rejection
```

baileympearson (changes requested):

src/cmap/connection.ts
nbbeeken: Since we even make an error for onClose I'm not sure error can really ever be optional 🤔

src/cmap/connection.ts
```diff
@@ -619,7 +596,18 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+    const { promise: drained, resolve, reject } = promiseWithResolvers<void>();
+    const onDrain = () => resolve();
+    const onError = (error: Error) => reject(error);
+
+    this.socket.once('drain', onDrain).once('error', onError);
+    try {
+      return await drained;
+    } finally {
+      this.socket.off('drain', onDrain).off('error', onError);
+    }
```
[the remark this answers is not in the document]
nbbeeken: Just didn't think to use it, I made a util out of it! :)

I lean towards a util we write over the node one for runtime compat in case there's a change in a future version, and I can simplify the logic for our use case, single argument / no signal use.

src/cmap/wire_protocol/on_data.ts
nbbeeken: fixed

PR #4025, opened 2024-03-06
fix(NODE-5636): generate _ids using pkFactory in bulk write operations

baileympearson:
#### What is changing?

Bulk operations now generate _ids for inserted documents using the pkFactory, if the user has provided one.

No.

### Bulk Write Operations Generate Ids using `pkFactory`

When performing inserts, the driver automatically generates `_id`s for each document if there is no `_id` present.  By default, the driver generates `ObjectId`s.  An option, `pkFactory`, can be used to configure the driver to generate `_id`s that are not object ids.

For a long time, only `Collection.insert` and `Collection.insertMany` actually used the `pkFactory`, if configured.  Notably, `Collection.bulkWrite()`, `Collection.initializeOrderedBulkOp()` and `Collection.initializeOrderedBulkOp()` _always_ generated `ObjectId`s, regardless of what was configured on collection.

The driver always generates `_id`s for inserted documents using the `pkFactory`.

> [!CAUTION]
> If you are using a `pkFactory` and performing bulk writes, you may have inserted data into your database that does not have `_id`s generated by the `pkFactory`.  


src/bulk/common.ts
```diff
@@ -1093,21 +1094,16 @@ export abstract class BulkOperationBase {
...
-          (op.insertOne as Document)._id = new ObjectId();
-        }
-
-        return this.addToOperationsList(BatchType.INSERT, op.insertOne);
-      }
+      const forceServerObjectId = this.shouldForceServerObjectId();
+      const document =
+        op.insertOne && op.insertOne.document == null
+          ? // NOTE: provided for legacy support, but this is a malformed operation
+            (op.insertOne as Document)
```
nbbeeken, line 1101: This probably shouldn't have just been left as a comment, we should actually remove this, ideally, TS is already preventing this kind of code from existing for 3 major versions now. todo & ticket?

src/operations/insert.ts
```diff
@@ -131,7 +131,9 @@ export class InsertManyOperation extends AbstractOperation<InsertManyResult> {
     const writeConcern = WriteConcern.fromOptions(options);
     const bulkWriteOperation = new BulkWriteOperation(
       coll,
-      prepareDocs(coll, this.docs, options).map(document => ({ insertOne: { document } })),
+      maybeAddIdToDocuments(coll, this.docs, options).map(document => ({
+        insertOne: { document }
+      })),
```
nbbeeken, line 136: Is this redundant now since bulk will call the maybeAdd on the array of `{insertOne:{document}}`?

test/integration/crud/bulk.test.ts
```diff
@@ -65,16 +66,84 @@ describe('Bulk', function () {
...
+        collection = client.db('integration').collection('pk_factory_tests');
+        await collection.deleteMany({});
+      });
+
+      afterEach(() => client.close());
+
+      it('insertMany() generates _ids using the pkFactory', async function () {
+        await collection.insertMany([{ name: 'john doe' }]);
+        const result = await collection.findOne({ name: 'john doe' });
+        expect(result).to.have.property('_id').to.be.instanceOf(Double);
```
nbbeeken, line 116: Usually we check the `_bsontype` but I don't feel strongly.

src/operations/common_functions.ts
```diff
@@ -43,26 +43,37 @@ export async function indexInformation(
...
   options: { forceServerObjectId?: boolean }
-): Document[] {
+): Document[];
+export function maybeAddIdToDocuments(
+  coll: Collection,
+  docs: Document,
+  options: { forceServerObjectId?: boolean }
+): Document;
+export function maybeAddIdToDocuments(
+  coll: Collection,
```
nbbeeken, line 57: Just curious, not at all essential, if this could be made type-safe since we have a reference to the collection, could TSchema be threaded through along with InferId 🤔 is pkFactory parameterized on TSchema?

test/integration/crud/bulk.test.ts
```diff
@@ -65,16 +66,84 @@ describe('Bulk', function () {
...
+          .raw({ insertOne: { name: 'john doe' } })
+          .execute();
+        const result = await client.db('test').collection('test').findOne({ name: 'john doe' });
+        expect(result).to.exist;
+      });
     });
   });
 
   describe('Collection', function () {
+    describe('when a pkFactory is set on the client', function () {
```
nbbeeken, line 95: Can we combine or remove the tests in `test/integration/crud/pk_factory.test.js`?

src/bulk/common.ts
```diff
@@ -1268,6 +1264,18 @@ export abstract class BulkOperationBase {
...
+  private shouldForceServerObjectId(): boolean {
+    if (typeof this.s.options.forceServerObjectId === 'boolean') {
+      return this.s.options.forceServerObjectId;
+    }
+
+    if (typeof this.s.collection.s.db.options?.forceServerObjectId === 'boolean') {
+      return this.s.collection.s.db.options?.forceServerObjectId;
+    }
+
+    return false;
```
nbbeeken, line 1277:
```suggestion
    return this.s.options.forceServerObjectId === true || this.s.collection.s.db.options.forceServerObjectId === true;
```

Either option is either set precisely to `true` or we default to `false`, just a suggestion

src/bulk/common.ts
```diff
@@ -1268,6 +1264,18 @@ export abstract class BulkOperationBase {
...
     document: Document | UpdateStatement | DeleteStatement
   ): this;
+
+  private shouldForceServerObjectId(): boolean {
+    if (typeof this.s.options.forceServerObjectId === 'boolean') {
+      return this.s.options.forceServerObjectId;
+    }
+
+    if (typeof this.s.collection.s.db.options?.forceServerObjectId === 'boolean') {
+      return this.s.collection.s.db.options?.forceServerObjectId;
```
nbbeeken, line 1274:
```suggestion
      return this.s.collection.s.db.options.forceServerObjectId;
```

Not in scope but db.options is never nullish according to the db constructor, any concerns just cleaning that up?

nbbeeken (changes requested):
Nice work, a few tiny cleanups

src/bulk/common.ts
baileympearson: I don't know if it's safe to remove - this is available in mongosh.  In mongosh, TS doesn't prevent this 🙂

nbbeeken: I agree, not here of course, but perhaps in a major we could right?

src/operations/common_functions.ts
nbbeeken: Not essential but since you're "here". You could delete that TODO, js-bson is typed but I don't think it would ever be helpful to constrain the return type 🤔 at least not to a BSON type, maybe to the TSchema but that's a thought for another time.

test/integration/crud/bulk.test.ts
baileympearson: I removed the pk_factory test and moved it into `insert.test.js`

src/operations/insert.ts
baileympearson: fixed

src/bulk/common.ts
baileympearson: I filed a ticket in the v7 epic

src/bulk/common.ts
baileympearson: I chose to leave the `?` chaining.

src/bulk/common.ts
baileympearson: good catch! fixed.

test/integration/crud/bulk.test.ts
baileympearson: fixed

PR #4028, opened 2024-03-08
fix(Node 5999): Fix mongoServerError casting

aditi-khare-mongoDB:


src/sdam/topology_description.ts
```diff
@@ -307,13 +307,13 @@ export class TopologyDescription {
     );
   }
 
-  get error(): MongoServerError | null {
+  get error(): MongoError | null {
```
baileympearson, line 310: This is technically public.  I think this change is okay to make, because we usually permit breaking TS fixes in non-major releases.  What do you think?

aditi-khare-mongoDB: Is an expanded type considered a breaking change?

baileympearson: Hm on second thought - if we had made this change before https://github.com/mongodb/node-mongodb-native/commit/c02324218db18e7c51f5b775f35edacc084762b0, I don't think this would have been breaking.  but since `MongoServerError` is no longer assignable to `MongoError`, it's now breaking to make this change?  But I still think that's okay.  This is a correctness fix.

aditi-khare-mongoDB: `MongoServerError` inherits from` MongoError`, so `MongoServerError` can be assigned to `MongoError`, but not the other way around. This is why we had to make this change.

baileympearson:
```typescript
interface Foo {
  name: string;
}

interface Bar extends Foo {
  age: number;
}

const document: Bar = { name: 'bumpy', age: 7 };
```

If `Bar` is changed to `Foo`, this code fails to compile.  That's why this change is potentially problematic

aditi-khare-mongoDB: Hm I see. Should we wait until the next major version to make this change?

baileympearson (approved):

PR #4031, opened 2024-03-11
fix(NODE-6010): Avoid triple-emitting close events

vkarpov15:
### Description

If you kill and then restart the underlying replica set in the following script, you'll see that the `close` event is emitted 3 times:

```javascript
'use strict';
  
const { MongoClient } = require('mongodb');

void async function main() {
  const uri = 'mongodb://127.0.0.1:27017,127.0.0.1:27018/mongoose_test';
  const client = new MongoClient(uri, { useNewUrlParser: true, useUnifiedTopology: true });
  await client.connect();

  const db = client.db();
  const collection = db.collection('people');

  let resumeAfter = null;
  let changeStream = collection.watch();
  addChangeStreamListeners(changeStream);

  function addChangeStreamListeners(changeStream) {
    changeStream.on('change', data => console.log(new Date(), data));
    changeStream.on('resumeTokenChanged', data => {
      console.log(new Date(), 'resumeTokenChanged', data);
      resumeAfter = data._data;
    });
    changeStream.on('error', data => console.log(new Date(), 'error', data));
    changeStream.on('close', async data => {
      console.log(new Date(), 'close', data, new Error().stack);
      changeStream = collection.watch([], { resumeAfter });
      addChangeStreamListeners(changeStream);
    });
  }

  while (true) {
    await new Promise(resolve => setTimeout(resolve, 8000));
    // Insert a doc, will trigger the change stream handler above
    console.log(new Date(), 'Inserting doc');
    await collection.insertOne({ name: 'Axl Rose' });
    console.log(new Date(), 'Inserted doc');
  }
}();
```

First time because abstract cursor `next()` cleans up the cursor: https://github.com/mongodb/node-mongodb-native/blob/31f1eed293f96d9e2f9d64a07088700c522ec860/src/cursor/abstract_cursor.ts#L737-L743, which emits 'close' here: https://github.com/mongodb/node-mongodb-native/blob/31f1eed293f96d9e2f9d64a07088700c522ec860/src/cursor/abstract_cursor.ts#L846

2nd time because `ReadableCursorStream` calls `destroy()`: https://github.com/mongodb/node-mongodb-native/blob/31f1eed293f96d9e2f9d64a07088700c522ec860/src/cursor/abstract_cursor.ts#L943, which calls `close()` on the already closed cursor: https://github.com/mongodb/node-mongodb-native/blob/31f1eed293f96d9e2f9d64a07088700c522ec860/src/cursor/abstract_cursor.ts#L895

3rd time because `destroy` triggers an 'error' event for reasons that I haven't quite been able to figure out

This PR makes it so that only 1 'close' event is emitted, and it looks like all it does is prevent an extra unnecessary `close()` call?

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -940,7 +940,9 @@ class ReadableCursorStream extends Readable {
         //       working on 4.4 servers because the error emitted on failover was "interrupted at
         //       shutdown" while on 5.0+ it is "The server is in quiesce mode and will shut down".
         //       See NODE-4475.
-        return this.destroy(err);
+        if (!this._cursor.isDead) {
+          return this.destroy(err);
+        }
       }
     );
   }


baileympearson:
Hey @vkarpov15 - is there a specific bug you're trying to fix here or a specific issue this is addressing?  It seems like you're only touching cursor behavior in this PR, but the reproduction uses change streams so I'm curious about the actual bug reported.

I haven't been able to reproduce this behavior when testing locally.  How are you restarting the cluster when testing?

> 3rd time because destroy triggers an 'error' event for reasons that I haven't quite been able to figure out

Our cursor stream is a Nodejs stream, which emits an error on destroy(): https://nodejs.org/api/stream.html#writabledestroyerror

W-A-James:
Hi @vkarpov15, apologies for the delay in response. Thanks for reporting this bug. I've reproduced the results you reported and have run your patch against our CI. While your patch does fix the immediate issue you reported, it does lead to a number of other test failures as we have certain functionality that depends on how our change streams emit close events.
Fixing this issue will take a more concerted effort to refactor our cursors which we're not able to undertake at this time. I'll update the ticket (NODE-6010) with the details you've provided here and we'll keep track of it for future consideration.

PR #4034, opened 2024-03-12
fix(NODE-6019): indexExists always returns false when `full` is set to `true`

baileympearson:
#### What is changing?

In preparation for [NODE-5971](https://jira.mongodb.org/browse/NODE-5971), I took time to clean up our index management implementation.  

- `index_management.test.js` is converted to TS and async await.
- unused `EnsureIndexOperation` is removed.
- `CreateIndexOperation` has been removed in favor of `CreateIndexesOperation`.
- the `indexInformation` helper has been removed and the classes that used it as well: `Indexes`, `IndexExists` and `IndexInformation` operations have had their logic shifted into the collection class.  Additionally, tests were added to ensure the behavior was not broken for these functions.

### `indexExists()` no longer supports the `full` option

The `Collection.indexExists()` helper supported an option, `full`, that modified the internals of the method.  When `full` was set to `true`, the driver would always return `false`, regardless of whether or not the index exists.

The `full` option is intended to modify the return type of index enumeration APIs (`Collection.indexes()` and `Collection.indexInformation()`, but since the return type of `Collection.indexExists()` this option does not make sense for the `Collection.indexExists()` helper.  

We have removed support for this option.

### `indexExists()`, `indexes()` and `indexInformation()` support cursor options in Typescript

These APIs have supported cursor options at runtime since the 4.x version of the driver, but our Typescript has incorrectly omitted cursor options from these APIs.


src/operations/indexes.ts
```diff
@@ -176,42 +181,12 @@ function makeIndexSpec(
...
-export class CreateIndexesOperation<
-  T extends string | string[] = string[]
-> extends CommandOperation<T> {
+export class CreateIndexesOperation extends CommandOperation<string[]> {
   override options: CreateIndexesOptions;
   collectionName: string;
   indexes: ReadonlyArray<Omit<IndexDescription, 'key'> & { key: Map<string, IndexDirection> }>;
 
-  constructor(
+  private constructor(
```
baileympearson, line 189: I'm happy to revert this change - but since I added static factories I decided it made sense to restrict usage of the constructor.

src/collection.ts
```diff
@@ -804,10 +800,19 @@ export class Collection<TSchema extends Document = Document> {
...
-    );
+    const indexes = await this.listIndexes(options).toArray();
+    const full = options?.full ?? true;
+    if (full) {
+      return indexes;
+    }
+
+    const info: Record<string, Array<[string, unknown]>> = {};
+    for (const { name, key } of indexes) {
+      info[name] = Object.entries(key);
```
nbbeeken, line 811: If we use fromEntries here we can avoid accidentally setting `'__proto__'`

src/operations/indexes.ts
```diff
@@ -73,6 +71,13 @@ export type IndexSpecification = OneOrMore<
   | Map<string, IndexDirection>
 >;
 
+/** @public */
+export interface IndexInformationOptions {
+  full?: boolean;
+  readPreference?: ReadPreference;
```
nbbeeken, line 77: Some docs at least for `full`?

src/operations/create_collection.ts
```diff
@@ -167,7 +167,14 @@ export class CreateCollectionOperation extends CommandOperation<Collection> {
 
     if (encryptedFields) {
       // Create the required index for queryable encryption support.
-      const createIndexOp = new CreateIndexOperation(db, name, { __safeContent__: 1 }, {});
+      // We could use `this.db.collection(name).createIndex()` to create the index,
```
nbbeeken, line 170: I think you can remove the "we could" because now that we have this improvement (nice work!) we wouldn't want to anymore

src/operations/common_functions.ts
```diff
@@ -1,47 +1,5 @@
...
-  if (full) return indexes;
-
-  const info: Record<string, Array<[string, unknown]>> = {};
-  for (const index of indexes) {
-    info[index.name] = Object.entries(index.key);
-  }
-  return info;
-}
 
 export function maybeAddIdToDocuments(
```
nbbeeken, line 4: I would be inspired to move this to utils and drop this file but I do not feel strongly at all.

src/db.ts
```diff
@@ -426,10 +424,7 @@ export class Db {
     indexSpec: IndexSpecification,
     options?: CreateIndexesOptions
   ): Promise<string> {
-    return executeOperation(
-      this.client,
-      new CreateIndexOperation(this, name, indexSpec, resolveOptions(this, options))
-    );
+    return this.collection(name).createIndex(indexSpec, options);
```
nbbeeken, line 427: Why the throwaway collection vs just exec-ing your own createIndexOp?

src/db.ts
```diff
@@ -480,10 +475,7 @@ export class Db {
    * @param options - Optional settings for the command
    */
   async indexInformation(name: string, options?: IndexInformationOptions): Promise<Document> {
-    return executeOperation(
-      this.client,
-      new IndexInformationOperation(this, name, resolveOptions(this, options))
-    );
+    return this.collection(name).indexInformation(resolveOptions(this, options));
```
nbbeeken, line 478: same about collection vs op (just commenting so if we land on something different we don't forget to do both)

nbbeeken (changes requested):

src/operations/create_collection.ts
baileympearson: Thanks! 

Either way - but fwiw I think eventually we might want to run all this in a transaction, in which case we probably would want to use `db.collection(..).createIndex()`.  

Using public APIs feels cleaner, and has the added benefit of a bit of extra testing cause we're using a complete public API, but I don't feel strongly.

src/db.ts
baileympearson: Using public APIs feels cleaner, and has the added benefit of a bit of extra testing cause we're using a complete public API, but I don't feel strongly.

For `indexInformation`, `IndexInformationOperation` no longer exists so I call `Collection.indexInformation` to share code

src/operations/indexes.ts
baileympearson: sure.  but fwiw, it's not a new option 😓

src/collection.ts
```diff
@@ -684,10 +681,12 @@ export class Collection<TSchema extends Document = Document> {
     indexes: string | string[],
     options?: IndexInformationOptions
   ): Promise<boolean> {
-    return executeOperation(
-      this.client,
-      new IndexExistsOperation(this as TODO_NODE_3286, indexes, resolveOptions(this, options))
-    );
+    const indexNames: string[] = [indexes].flat();
```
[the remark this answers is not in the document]
baileympearson: Okay, I adjusted it to check if the input was an array.

src/collection.ts
```diff
@@ -684,10 +681,12 @@ export class Collection<TSchema extends Document = Document> {
...
+    const allIndexes: string[] = await this.listIndexes(options)
+      .map(({ name }) => name)
+      .toArray();
+
+    return indexNames.every(name => allIndexes.includes(name));
```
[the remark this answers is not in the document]
nbbeeken: Fine with all of the above, leaving it as-is seems least preferable since we're listing _all_ indexes I don't want to assume how many that could be, so seems worth paying an upfront cost of hashing the values to make `every` a bit snappier but I don't feel too strongly

Does this change make this PR a fix?

src/db.ts
nbbeeken: We should probably determine some consistency around using our own public API or circumventing it, I am not sure what the right call is or if there is one that applies universally.

I see this the opposite way of clean because we could call `CreateIndexesOperation.fromIndexSpecification` with the name and indexSpec given, and resolveOptions with the db only. I feel a little strongly here becasue we create an object and do work to make the collection that will just be lost and doesn't need to be done to perform this operation meanwhile using the helper I mention still shares a majority of the code.

src/operations/create_collection.ts
nbbeeken: As said in the other thread I hesitate to adopt public API because of the extra cost it adds to work that can be done with less. But I don't think there's a universal rule. 

Anyway something like what you suggest sounds like it would be a drivers wide change, want to file a ticket about it? I'd prefer the comment be paired down, suggestions like this don't have a enough context to clue the next dev into what the intention was here and it can imply that the change is small and simple when it may have larger impacts (like re-selecting a server)

src/collection.ts
```diff
@@ -806,13 +806,12 @@ export class Collection<TSchema extends Document = Document> {
       return indexes;
     }
 
-    const info: Record<string, Array<[string, unknown]>> = {};
-    for (const { name, key } of indexes) {
-      info[name] = Object.entries(key);
-    }
+    const object: Record<string, Array<[string, unknown]>> = Object.fromEntries(
```
nbbeeken, line 809:
```suggestion
    const object: Record<string, Array<[name: string, direction: IndexDirection]>> =
```

Just leaving a bit more TS context for the dev who fixes NODE-6029

nbbeeken (changes requested):

src/collection.ts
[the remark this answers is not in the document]
baileympearson: > Fine with all of the above, leaving it as-is seems least preferable since we're listing all indexes I don't want to assume how many that could be, so seems worth paying an upfront cost of hashing the values to make every a bit snappier but I don't feel too strongly

I modified the implementation to use a set.  But I do think it's worth keeping in mind that index management helpers are not typically used in hot code paths, they're usually administrative helpers that are used in tooling / scripting.

src/db.ts
nbbeeken: From slack, the indexInformation API and the way listIndexes work will create a collection anyway. We can leave this as is

nbbeeken (approved):

nbbeeken (approved):

PR #4038, opened 2024-03-14
docs(NODE-5721): thrown error expectations in withTransaction

aditi-khare-mongoDB:
### Description
Add API docs to `withTransaction` function to aid users with error handling in the case that the command passed into the transaction fails, and the transaction goes into an infinite loop.

#### What is changing?
Descriptive API doc with error handling steps is added to `withTransaction`

Yes, I've added "Docs Changes Needed" to the ticket description, asking the docs team to update this [page](https://www.mongodb.com/docs/drivers/node/current/fundamentals/transactions/#convenient-transaction-api).

Help users with error handling after this infinite loop case was discovered.

### Add error handling information to `withTransaction` API docs
The `withTransaction` function API docs now contain more information on how to avoid an application entering an infinite loop in the event that a failed command aborts the transaction and causes the transaction to be retried.

### DIFF
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -446,6 +446,15 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    *
    * Checkout a descriptive example here:
    * @see https://www.mongodb.com/blog/post/quick-start-nodejs--mongodb--how-to-implement-transactions
+   * 
+   * If a command inside this function fails:
+   * - It may cause the transaction on the server to be aborted. 
+   * - This situation is normally handled transparently by the driver. 
+   * - However, if the application catches such an error and does not re-raise it, the driver will not be able to determine whether the transaction was aborted or not. 
+   * - The driver will then retry the transaction indefinitely.
+   * 
+   * To avoid this situation, the application must not silently handle errors within this function.
+   * If the application needs to handle errors within, it must await all operations such that if an operation is rejected, the rejection becomes the rejection of the function passed into withTransaction.
    *
    * @param fn - callback to run within a transaction
    * @param options - optional settings for the transaction
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -446,15 +446,15 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    *
    * Checkout a descriptive example here:
    * @see https://www.mongodb.com/blog/post/quick-start-nodejs--mongodb--how-to-implement-transactions
-   * 
+   *
    * If a command inside this function fails:
-   * - It may cause the transaction on the server to be aborted. 
-   * - This situation is normally handled transparently by the driver. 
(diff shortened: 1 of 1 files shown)


src/sessions.ts
```diff
@@ -447,6 +447,15 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
    * Checkout a descriptive example here:
    * @see https://www.mongodb.com/blog/post/quick-start-nodejs--mongodb--how-to-implement-transactions
    *
+   * If a command inside this function fails:
+   * - It may cause the transaction on the server to be aborted.
+   * - This situation is normally handled transparently by the driver.
+   * - However, if the application catches such an error and does not re-raise it, the driver will not be able to determine whether the transaction was aborted or not.
```
nbbeeken, line 453:
```suggestion
   * - However, if the application catches such an error and does not rethrow it, the driver will not be able to determine whether the transaction was aborted or not.
```

src/sessions.ts
```diff
@@ -447,6 +447,15 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
...
+   * - The driver will then retry the transaction indefinitely.
+   *
+   * To avoid this situation, the application must not silently handle errors within this function.
+   * If the application needs to handle errors within, it must await all operations such that if an operation is rejected it becomes the rejection of the callback function passed into withTransaction.
```
nbbeeken, line 457: Not on this line but above, I want the "quick" important sentence to call out this detail too if that sounds good. I think you can leave what is written here as is.

> **IMPORTANT:** This method requires the user to return a Promise, and `await` all operations.

update to:

> **IMPORTANT:** This method requires the function passed in to return a Promise. That promise must be made by `await`-ing all operations in such a way that rejections are propagated to the returned promise.

nbbeeken (changes requested):

src/sessions.ts
aditi-khare-mongoDB: Made the change!

src/sessions.ts
```diff
@@ -435,7 +435,7 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
   /**
    * Starts a transaction and runs a provided function, ensuring the commitTransaction is always attempted when all operations run in the function have completed.
    *
-   * **IMPORTANT:** This method requires the user to return a Promise, and `await` all operations.
+   * **IMPORTANT:** This method requires the function passed in to return a Promise. That promise must be made by `await`-ing all operations in such a way that rejections are propagated to the returned promise.
    *
    * @remarks
    * This function:
```
nbbeeken, line 441: One more fix up, since we are here, "this function" doesn't make sense for the bullets below this except for the last one. Maybe at some point they were phrased differently. Can you delete "This function:" and rephrase the last bullet. "If the driver needs... it may be called..."

nbbeeken (changes requested):

PR #4042, opened 2024-03-15
fix(NODE-5745): Ignore Read/Write Concern in Atlas Search Index Helpers

aditi-khare-mongoDB:
### Description
Synced tests and remove r/w concern from relevant method's `options` value.

#### What is changing?
Added spec tests to ensure `readConcern` and `writeConcern` are not sent to command with Atlas Search Helpers.
Explicitly remove `readConcern` and `writeConcern` from `Collection.listSearchIndexes`.

No.

Spec compliance.

### Release Highlight

This type change is a correctness fix.

`Collection.listSearchIndexes` is an Atlas specific method, and  Atlas' search indexes do not support `readConcern` and `writeConcern` options. The types for this function now reflect this functionality. 


nbbeeken (changes requested):
Is this a feature? it doesn't modify any driver source. What about the txn unknown mentioned in the ticket? (Are txn's even relevant?)

aditi-khare-mongoDB:
There's a small code change, I mistakenly marked this as Ready for Review preemptively, but it is ready for review now. I'm not sure how transactions are relevant, since we never pass wr concern into the command anyway.

nbbeeken:
Transactions are controlled by the session, if a session is in a transaction it will apply the transaction's wc/rc settings, so how are we handling that? Do we need to handle that?

src/collection.ts
```diff
@@ -1072,14 +1072,18 @@ export class Collection<TSchema extends Document = Document> {
   ): ListSearchIndexesCursor {
     options =
       typeof indexNameOrOptions === 'object' ? indexNameOrOptions : options == null ? {} : options;
+
+    /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+    const cleanedOptions = (({ readConcern, writeConcern, ...rest }) => rest)(options);
```
baileympearson, line 1077: whoa what a clever way to remove fields from an object!  I've never seen this before but I like it 🙂

nbbeeken:
```suggestion
    const { readConcern, writeConcern, ...cleanedOptions } = options;
```

What is the purpose of the closure?

nbbeeken (changes requested):

src/collection.ts
```diff
@@ -1061,7 +1061,7 @@ export class Collection<TSchema extends Document = Document> {
    * Returns all search indexes for the current collection.
    *
    * @param name - The name of the index to search for.  Only indexes with matching index names will be returned.
-   * @param options - The options for the list indexes operation.
+   * @param options - The options for the list indexes operation. This method will ignore 'readConcern' and 'writeConcern' keys if they are present in 'options' (since there is no Atlas support).
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Done + updated the release highlights to reflect this change!

src/collection.ts
aditi-khare-mongoDB: No reason for this extra layer, removed it.

Edit: Removed the entire line of code since with the new type checking in the argument it is no longer relevant.

test/integration/index-management/node-specific/searchIndexIgnoresReadWriteConcern-with-transaction.json
aditi-khare-mongoDB, line 1: This is a replica of the spec test, except this test file passes in a transaction session into the operation arguments.

nbbeeken: Where did the transaction start? I'm not seeing a `startTransaction` or `withTransaction` operation.

I'm surprised we went with making a unified test out of this, I would've gone with writing it in code I think it would be succinct but not against this approach.

src/collection.ts
nbbeeken: I'm not sure I fully agree with typescript being the final source of trust that we don't attach those settings but willing to be convinced otherwise. I feel as though the one liner to prevent the error is worth keeping. @baileympearson thoughts?

nbbeeken (changes requested):

nbbeeken (changes requested):
Is this really a feat? shouldn't it be a fix?

src/collection.ts
aditi-khare-mongoDB: When I add the line back in it says that the options type does not contain readConcern / writeConcern with the updated options typing.

nbbeeken (approved):
LGTM, @baileympearson is the linked NODE ticket appropriate or should this be socialized via a DRIVERS ticket? This may only be a quirk of our driver.

src/collection.ts
baileympearson: I don't feel strongly.  We do, however, have precedent for removing write concern from options where it should never exist according to TS: https://github.com/mongodb/node-mongodb-native/blob/918fe69d104d8573966d52605b916a9de904d11a/src/operations/indexes.ts#L358

test/integration/index-management/search-index-management.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      commandStartedEvents = [];
+      client.on('commandStarted', e => commandStartedEvents.push(e));
+    });
+
+    afterEach(async function () {
+      await client.close();
+    });
+
+    context('when listSearchIndexes operation is run withTransaction', function () {
+      it.skip('should not include write concern or read concern in command', async function () {
```
nbbeeken, line 24: Can we flip the assertions and unskip

nbbeeken (changes requested):

test/integration/index-management/search-index-management.test.ts
```diff
@@ -21,64 +21,80 @@ describe('Search Index Management Integration Tests', function () {
...
 
     context('when listSearchIndexes operation is run with snapshot on', function () {
-      it.skip('should not include write concern or read concern in command', async function () {
-        await client.withSession({ snapshot: true }, async session => {
-          try {
+      // TODO(NODE-6047): Ignore read/write concern in applySession for Atlas Search Index Helpers
+      it('should include write concern or read concern in command - TODO(NODE-6047)', {
+        metadata: {
+          requires: {
+            topology: ['replicaset', 'sharded'],
```
aditi-khare-mongoDB, line 81: Load-balanced is excluded from topologies. This is because Mongodb versioning for snapshots is not captured by load-balanced test variants. For example, a rapid test will say, "snapshot requires MongoDB >= 5.0." Here's a [link to an example](https://spruce.mongodb.com/task/mongo_node_driver_next_rhel80_large_Node20_test_rapid_load_balanced_patch_abf8bdfc8dbae85356a7cef9d476e619a79e53d9_660f196e6650da000703866d_24_04_04_21_20_47/logs?execution=0&sortBy=STATUS&sortDir=ASC).

src/collection.ts
```diff
@@ -1056,7 +1056,9 @@ export class Collection<TSchema extends Document = Document> {
    *
    * @remarks Only available when used against a 7.0+ Atlas cluster.
    */
-  listSearchIndexes(options?: ListSearchIndexesOptions): ListSearchIndexesCursor;
+  listSearchIndexes(
+    options?: Omit<ListSearchIndexesOptions, 'readConcern' | 'writeConcern'>
```
baileympearson, line 1060: Is there ever a scenario where `ListSearchIndexesOptions` will allow a read / write concern?  If not, maybe we should just omit these options on the `ListSearchIndexesOptions` interface instead

aditi-khare-mongoDB: IIUC, we can run ListSearchIndexes with runCommand outside of Atlas.

baileympearson (approved):

nbbeeken (approved):

PR #4043, opened 2024-03-18
fix(NODE-5971): attach `v` to createIndexes command when `version` is specified

baileympearson:
#### What is changing?

`createIndex` and `createIndexes` now support specifying an index version with the `version` option.

One note is that `createIndexes()` requires that users specify the `version` as a part of their index specification - a `version` provided as a part of the options is ignored.  This behavior is consistent with other index options provided to `createIndexes`.

(release highlight left empty since `v` is an undocumented feature, no need for it to appear in our release notes)

### DIFF
--- a/src/operations/indexes.ts
+++ b/src/operations/indexes.ts
@@ -167,10 +167,11 @@ function isSingleIndexTuple(t: unknown): t is [string, IndexDirection] {
   return Array.isArray(t) && t.length === 2 && isIndexDirection(t[1]);
 }
 
-function makeIndexSpec(
-  indexSpec: IndexSpecification,
-  options?: CreateIndexesOptions
-): IndexDescription {
+/**
+ * Converts an `IndexSpecification`, which can be specified in multiple formats, into a
+ * valid `key` for the createIndexes command.
+ */
+function constructIndexDescriptionMap(indexSpec: IndexSpecification): Map<string, IndexDirection> {
   const key: Map<string, IndexDirection> = new Map();
 
   const indexSpecs =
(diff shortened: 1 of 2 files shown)


src/operations/indexes.ts
```diff
@@ -212,16 +245,12 @@ export class CreateIndexesOperation extends CommandOperation<string[]> {
 
     this.options = options ?? {};
     this.collectionName = collectionName;
-    this.indexes = indexes.map(userIndex => {
+    this.indexes = indexes.map((userIndex: IndexDescription): ResolvedIndexDescription => {
       // Ensure the key is a Map to preserve index key ordering
       const key =
         userIndex.key instanceof Map ? userIndex.key : new Map(Object.entries(userIndex.key));
       const name = userIndex.name != null ? userIndex.name : Array.from(key).flat().join('_');
```
W-A-James, line 252: Optional, but while we're here
```suggestion
      const name = userIndex.name ?? Array.from(key).flat().join('_');
```

W-A-James (commented):
Looks good, just one optional change

W-A-James (approved):

src/operations/indexes.ts
```diff
@@ -193,14 +194,46 @@ function makeIndexSpec(
...
 }
 
+/**
+ * Receives an index description and returns a modified index description which has had invalid options removed
+ * from the description and has mapped the `version` option to the `v` option.
+ */
+function resolveIndexDescription(
+  description: IndexDescription
+): Omit<ResolvedIndexDescription, 'key'> {
+  const validProvidedOptions = Object.entries({ ...description }).filter(([optionName]) =>
```
nbbeeken, line 207:
```suggestion
  const validProvidedOptions = Object.entries(description).filter(([optionName]) =>
```

Isn't the spread redundant? Or did I miss something about the input type?

baileympearson: it's a speed-up loop: https://thedailywtf.com/articles/The-Speedup-Loop

Just joking 😄 this change should be safe, it was just copied from the original implementation.

W-A-James (approved):

PR #4052, opened 2024-03-25
feat(NODE-6044): add detailed types for `SearchIndexDescription`

noseworthy:
### Description

Adds more detailed type definitions for `SearchIndexDescription` and `SearchIndexDefinition` based on the documentation found at [Create an Atlas Search Index](https://www.mongodb.com/docs/atlas/atlas-search/create-index/).

The existing type:

```typescript
export interface SearchIndexDescription {
  name?: string;
  definition: Document;
}
```

allows users to use any object for the `definition` property. Search index definitions are fairly complicated and having more strict types defined for this property will make it easier to use the `createSearchIndex()` and `updateSearchIndex()` apis.

#### What is changing?

Added full types for defining a search index definition.

I don't think any new hand-written documentation is necessary, but new tsdoc documentation should be generated to capture the new types.

Using the `createSearchIndex()` and `updateSearchIndex()` apis when the definition is just specified to be a `bson.Document` is a bit unwieldy.  The search index definitions can be fairly complicated, and it's very helpful to have TypeScript help validate the definitions.


W-A-James:
Hi @noseworthy thanks so much for all your work on this PR. We're currently reviewing this and will get back to you soon.

W-A-James:
Hi @noseworthy! Thanks for your patience and thanks again for all your hard work on this PR. After reviewing it with the team, we've decided that accepting these changes into the driver isn't the right choice for us at the moment. Taking on the burden of keeping these types accurate as the Atlas Search product continues to be developed is not something we can commit to at this time.
The driver also might not be the right place for these changes as the type definitions would then be bound to specific releases of the Node driver. This is something we generally try to avoid as we put a lot of work into having our driver be compatible and easy to use across all our currently supported server versions.
We will leave this PR open for consideration later down the line.

noseworthy:
> Hi @noseworthy! Thanks for your patience and thanks again for all your hard work on this PR. After reviewing it with the team, we've decided that accepting these changes into the driver isn't the right choice for us at the moment. Taking on the burden of keeping these types accurate as the Atlas Search product continues to be developed is not something we can commit to at this time. The driver also might not be the right place for these changes as the type definitions would then be bound to specific releases of the Node driver. This is something we generally try to avoid as we put a lot of work into having our driver be compatible and easy to use across all our currently supported server versions. We will leave this PR open for consideration later down the line.

No worries, @W-A-James . Thanks for taking a look and considering it. We're really excited to start using the Atlas Search features more now that we can run it locally for our dev environments but I found definitions for the indices really complicated. I was hoping that this would've been able to make it a bit more approachable than a simple `bson.Document` declaration. But I totally understand the burden of maintaining these types. Like I said, they're complicated and I'm sure with vector search changing frequently too it'd be a huge pain.

Hopefully as the product evolves we can get better typescript support for some of these features 🤞. Thanks for all the team's work on the driver!

PR #4053, opened 2024-03-25
refactor(NODE-3332): withTransaction uses async/await

aditi-khare-mongoDB:
### Description
Remove callbacks for `withTransaction.`

#### What is changing?
See Description.

No.

Remove usage of callbacks within node driver codebase.

###

### DIFF
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -553,33 +553,33 @@ function isMaxTimeMSExpiredError(err: MongoError) {
   );
 }
 
-function attemptTransactionCommit<T>(
+async function attemptTransactionCommit<T>(
   session: ClientSession,
   startTime: number,
   fn: WithTransactionCallback<T>,
   result: any,
   options: TransactionOptions
 ): Promise<T> {
-  return session.commitTransaction().then(
-    () => result,
-    (err: MongoError) => {
-      if (
-        err instanceof MongoError &&
-        hasNotTimedOut(startTime, MAX_WITH_TRANSACTION_TIMEOUT) &&
-        !isMaxTimeMSExpiredError(err)
-      ) {
-        if (err.hasErrorLabel(MongoErrorLabel.UnknownTransactionCommitResult)) {
-          return attemptTransactionCommit(session, startTime, fn, result, options);
-        }
-
-        if (err.hasErrorLabel(MongoErrorLabel.TransientTransactionError)) {
-          return attemptTransaction(session, startTime, fn, options);
-        }
+  try {
+    await session.commitTransaction();
+    return result;
+  } catch (err) {
+    if (
+      err instanceof MongoError &&
+      hasNotTimedOut(startTime, MAX_WITH_TRANSACTION_TIMEOUT) &&
+      !isMaxTimeMSExpiredError(err)
+    ) {
+      if (err.hasErrorLabel(MongoErrorLabel.UnknownTransactionCommitResult)) {
+        return attemptTransactionCommit(session, startTime, fn, result, options);
       }
 
-      throw err;
+      if (err.hasErrorLabel(MongoErrorLabel.TransientTransactionError)) {
+        return attemptTransaction(session, startTime, fn, options);
+      }
     }
-  );
+
+    throw err;
+  }
 }
 
 const USER_EXPLICIT_TXN_END_STATES = new Set<TxnState>([
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -570,11 +570,11 @@ async function attemptTransactionCommit<T>(
       !isMaxTimeMSExpiredError(err)
     ) {
       if (err.hasErrorLabel(MongoErrorLabel.UnknownTransactionCommitResult)) {
-        return attemptTransactionCommit(session, startTime, fn, result, options);
+        return await attemptTransactionCommit(session, startTime, fn, result, options);
       }
 
       if (err.hasErrorLabel(MongoErrorLabel.TransientTransactionError)) {
-        return attemptTransaction(session, startTime, fn, options);
+        return await attemptTransaction(session, startTime, fn, options);
       }
     }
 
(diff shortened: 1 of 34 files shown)


W-A-James (changes requested):
Looks great so far. Just a couple small changes

src/sessions.ts
```diff
@@ -764,20 +739,37 @@ function endTransaction(
...
+      session.client,
+      new RunAdminCommandOperation(command, {
+        session,
+        readPreference: ReadPreference.primary,
+        bypassPinningCheck: true
+      })
+    );
+    await partiallyHandleFirstCommandAttempt();
+    commandHandler();
+  } catch (err) {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Yep!

src/sessions.ts
```diff
@@ -788,29 +780,23 @@ function endTransaction(
...
+        await executeOperation(
+          session.client,
+          new RunAdminCommandOperation(command, {
+            session,
+            readPreference: ReadPreference.primary,
+            bypassPinningCheck: true
+          })
+        );
+        commandHandler();
+      } catch (err2) {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Yep, I'll name it `secondAttemptError`

src/sessions.ts
```diff
@@ -610,65 +608,48 @@ function attemptTransaction<T>(
...
-const endTransactionAsync = promisify(
-  endTransaction as (
-    session: ClientSession,
-    commandName: 'abortTransaction' | 'commitTransaction',
-    callback: (error: Error) => void
-  ) => void
-);
+    throw err;
+  }
+}
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Thanks!

src/sessions.ts
```diff
@@ -555,33 +553,33 @@ function isMaxTimeMSExpiredError(err: MongoError) {
...
-          return attemptTransactionCommit(session, startTime, fn, result, options);
-        }
-
-        if (err.hasErrorLabel(MongoErrorLabel.TransientTransactionError)) {
-          return attemptTransaction(session, startTime, fn, options);
-        }
+  try {
+    await session.commitTransaction();
+    return result;
+  } catch (err) {
```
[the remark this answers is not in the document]
W-A-James: bumping this

W-A-James (changes requested):

src/sessions.ts
```diff
@@ -764,20 +739,37 @@ function endTransaction(
...
+    // send the command
+    await executeOperation(
+      session.client,
+      new RunAdminCommandOperation(command, {
+        session,
+        readPreference: ReadPreference.primary,
+        bypassPinningCheck: true
+      })
+    );
+    await partiallyHandleFirstCommandAttempt();
```
nbbeeken, line 766: Can we inline the logic here rather than keeping the callback?

src/sessions.ts
```diff
@@ -764,20 +737,37 @@ function endTransaction(
...
+    commandHandler();
```
nbbeeken, line 767: Can we remove `commandHandler` in favor of inlining the logic? it is still callback-like

nbbeeken (changes requested):

src/sessions.ts
aditi-khare-mongoDB: The logic for command handler is called 4 separate times, which is why i kept it as a function. To make it less callback like, I can change its name to not include 'handler.' Lmk what you think!

src/sessions.ts
aditi-khare-mongoDB: It would mean a small bit of code duplication, but done!

src/sessions.ts
nbbeeken: It's called 4 times  but its logic is conditional, we don't need the exact same code to run in all of these places when a portion of it will never be entered and if we need to re-run some of the same code we can employ a loop. Both tatics will reduce the amount of code executed, and remove a function created on each invocation of `endTransaction`. I think it is worth breaking it out here for that benefit.

nbbeeken (changes requested):

src/sessions.ts
```diff
@@ -555,33 +553,33 @@ function isMaxTimeMSExpiredError(err: MongoError) {
   );
 }
 
-function attemptTransactionCommit<T>(
+async function attemptTransactionCommit<T>(
   session: ClientSession,
   startTime: number,
   fn: WithTransactionCallback<T>,
   result: any,
```
nbbeeken, line 560:
```suggestion
  result: T,
```

This does not matter too much, but it technically more correct

src/sessions.ts
```diff
@@ -594,7 +592,7 @@ function userExplicitlyEndedTransaction(session: ClientSession) {
   return USER_EXPLICIT_TXN_END_STATES.has(session.transaction.state);
 }
 
-function attemptTransaction<T>(
+async function attemptTransaction<T>(
   session: ClientSession,
   startTime: number,
   fn: WithTransactionCallback<T>,
```
nbbeeken, line 598: I can't leave a comment on it but can we change this to return `Promise<T>` instead, it is just a correctness improvement.

nbbeeken (changes requested):

src/sessions.ts
```diff
@@ -610,65 +608,48 @@ function attemptTransaction<T>(
   }
 
   if (!isPromiseLike(promise)) {
-    session.abortTransaction().catch(() => null);
-    return Promise.reject(
-      new MongoInvalidArgumentError('Function provided to `withTransaction` must return a Promise')
+    await session.abortTransaction().catch(() => null);
```
nbbeeken, line 611:
```suggestion
    try {
      await session.abortTransaction();
    } catch (error) {
      squashError(error);
    }
```

Can you remove the `/* eslint-disable github/no-then */` at the top of the file and align this bit with the rest of the way we squash errors?

nbbeeken (changes requested):

src/sessions.ts
aditi-khare-mongoDB: Bookkeeping from Slack DM discussion: 
- Non-error case is moved to inline
- Error case function is no longer nested and renamed from `commandHandler` to `handleEndTransactionError`

nbbeeken (approved):

W-A-James (approved):

PR #4056, opened 2024-03-26
feat(NODE-5825): Add minRTT to serverdescription and pipe through Conn

W-A-James:

### DIFF
--- a/src/sdam/server.ts
+++ b/src/sdam/server.ts
@@ -172,12 +172,11 @@ export class Server extends TypedEventEmitter<ServerEvents> {
 
     this.monitor.on('resetServer', (error: MongoServerError) => markServerUnknown(this, error));
     this.monitor.on(Server.SERVER_HEARTBEAT_SUCCEEDED, (event: ServerHeartbeatSucceededEvent) => {
-      this.emit(
-        Server.DESCRIPTION_RECEIVED,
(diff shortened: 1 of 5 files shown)


src/sdam/server.ts
```diff
@@ -172,12 +172,11 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
     this.monitor.on('resetServer', (error: MongoServerError) => markServerUnknown(this, error));
     this.monitor.on(Server.SERVER_HEARTBEAT_SUCCEEDED, (event: ServerHeartbeatSucceededEvent) => {
-      this.emit(
-        Server.DESCRIPTION_RECEIVED,
-        new ServerDescription(this.description.hostAddress, event.reply, {
-          roundTripTime: calculateRoundTripTime(this.description.roundTripTime, event.duration)
-        })
+      this.description.updateWithHelloResponse(event.reply);
+      this.description.addRTTSample(
+        calculateRoundTripTime(this.description.roundTripTime, event.duration)
```
W-A-James, line 177: Question for reviewers: If we have the logic for this calculation live on `ServerDescription` is this how we want to update `roundTripTime` on that class, or would it be better to have this logic live in the `Monitor` class and leave the `rountTripTime` calculation as-is on the `ServerDescription` class?

PR #4058, opened 2024-03-26
refactor(NODE-5916): enforce return-await

nbbeeken:
#### What is changing?

- enforce return-await
- make `.then` discouraged

No

Async stack traces

### DIFF
--- a/src/cmap/handshake/client_metadata.ts
+++ b/src/cmap/handshake/client_metadata.ts
@@ -1,4 +1,3 @@
-import { promises as fs } from 'fs';
 import * as os from 'os';
 import * as process from 'process';
 
--- a/src/cmap/handshake/client_metadata.ts
+++ b/src/cmap/handshake/client_metadata.ts
@@ -5,6 +5,7 @@ import * as process from 'process';
 import { BSON, type Document, Int32 } from '../../bson';
 import { MongoInvalidArgumentError } from '../../error';
 import type { MongoOptions } from '../../mongo_client';
+import { fileIsAccessible } from '../../utils';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const NODE_DRIVER_VERSION = require('../../../package.json').version;
@@ -156,18 +157,12 @@ export function makeClientMetadata(options: MakeClientMetadataOptions): ClientMe
   return metadataDocument.toObject() as ClientMetadata;
 }
 
-let isDocker: Promise<void> | boolean | undefined;
+let dockerPromise: Promise<boolean>;
 /** @internal */
 async function getContainerMetadata() {
   const containerMetadata: Record<string, any> = {};
-  if (typeof isDocker !== 'boolean') {
-    try {
-      await (isDocker ??= fs.access('/.dockerenv'));
-      isDocker = true;
-    } catch {
-      isDocker = false;
-    }
-  }
+  dockerPromise ??= fileIsAccessible('/.dockerenv');
+  const isDocker = await dockerPromise;
 
   const { KUBERNETES_SERVICE_HOST = '' } = process.env;
   const isKubernetes = KUBERNETES_SERVICE_HOST.length > 0 ? true : false;
--- a/src/cmap/handshake/client_metadata.ts
+++ b/src/cmap/handshake/client_metadata.ts
@@ -156,15 +156,18 @@ export function makeClientMetadata(options: MakeClientMetadataOptions): ClientMe
   return metadataDocument.toObject() as ClientMetadata;
 }
 
-let dockerPromise: Promise<boolean>;
+let isDocker: Promise<void> | boolean | undefined;
 /** @internal */
 async function getContainerMetadata() {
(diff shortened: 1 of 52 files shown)


src/sessions.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable github/no-then */
```
nbbeeken, line 1: This will be removed in https://github.com/mongodb/node-mongodb-native/pull/4053

src/gridfs/download.ts
```diff
@@ -183,24 +183,13 @@ function doRead(stream: GridFSBucketReadStream): void {
...
-    doc
-  }: { error: Error; doc: null } | { error: null; doc: any }) => {
-    if (stream.destroyed) {
-      return;
-    }
-    if (error) {
-      stream.destroy(error);
-      return;
-    }
+  const handleReadResult = (doc: Document | null) => {
```
nbbeeken, line 186: Moved the error case to the rejection handler since it was a one-liner. It will make a future conversion (if ever) to async easier

src/operations/list_databases.ts
```diff
@@ -63,7 +63,8 @@ export class ListDatabasesOperation extends CommandOperation<ListDatabasesResult
       cmd.comment = this.options.comment;
     }
 
-    return super.executeCommand(server, session, cmd) as TODO_NODE_3286;
+    const res: TODO_NODE_3286 = super.executeCommand(server, session, cmd);
```
baileympearson, line 66: How come there's no `await` here?

baileympearson:
```suggestion
return await super.executeCommand(server, session, cmd) as Promise<TODO_NODE_3286>;
```

maybe?

src/encrypter.ts
```diff
@@ -108,13 +108,18 @@ export class Encrypter {
   }
 
   async close(client: MongoClient, force: boolean): Promise<void> {
-    const maybeError: Error | void = await this.autoEncrypter.teardown(!!force).catch(e => e);
+    let error;
+    try {
+      await this.autoEncrypter.teardown(!!force);
```
baileympearson, line 113:
```suggestion
      await this.autoEncrypter.teardown(force);
```

src/connection_string.ts
```diff
@@ -552,9 +552,11 @@ export function parseOptions(
 
   mongoOptions.metadata = makeClientMetadata(mongoOptions);
 
-  mongoOptions.extendedMetadata = addContainerMetadata(mongoOptions.metadata).catch(() => {
-    /* rejections will be handled later */
-  });
+  // eslint-disable-next-line github/no-then
+  mongoOptions.extendedMetadata = addContainerMetadata(mongoOptions.metadata).then(
+    undefined,
+    squashError
```
baileympearson, line 558: can we leave the comment?

src/connection_string.ts
```diff
@@ -72,9 +73,8 @@ export async function resolveSRVRecord(options: MongoOptions): Promise<HostAddre
   // the SRV record is resolved before starting a second DNS query.
   const lookupAddress = options.srvHost;
   const txtResolutionPromise = dns.promises.resolveTxt(lookupAddress);
-  txtResolutionPromise.catch(() => {
-    /* rejections will be handled later */
-  });
+  // eslint-disable-next-line github/no-then
+  txtResolutionPromise.then(undefined, squashError);
```
baileympearson, line 77: same here

src/cmap/handshake/client_metadata.ts
```diff
@@ -156,15 +156,18 @@ export function makeClientMetadata(options: MakeClientMetadataOptions): ClientMe
...
-  );
-  const isDocker = await dockerPromise;
+  if (typeof isDocker !== 'boolean') {
+    try {
+      await (isDocker ??= fs.access('/.dockerenv'));
+      isDocker = true;
+    } catch {
+      isDocker = false;
+    }
+  }
```
baileympearson, line 170: Can we exempt this one?  This is significantly less readable imo

baileympearson (changes requested):

src/connection_string.ts
nbbeeken: I can, but would we consider naming squashError something to communicate the same thing? or do you prefer the comment?

nbbeeken: `handleRejection`?

src/cmap/handshake/client_metadata.ts
nbbeeken: Tried an improvement, lmk what you think

src/connection_string.ts
baileympearson: I think these are special cases.  Unlike other `.catch(noop)`s, we're not squashing the errors - they'll still be handled when we await the promise later.  We just need catch handlers here in case there's a rejection before we handle the promise.  It's useful to communicate that we still care about the rejection, but we can't handle it yet.

baileympearson (approved):

PR #4059, opened 2024-03-27
feat(NODE-5825): add minRTT and average rtt measures to Monitor

W-A-James:

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -28,6 +28,7 @@ import { MongoLoggableComponent, type MongoLogger, SeverityLevel } from '../mong
 import { type CancellationToken, TypedEventEmitter } from '../mongo_types';
 import { ReadPreference, type ReadPreferenceLike } from '../read_preference';
 import { ServerType } from '../sdam/common';
+import { type Monitor } from '../sdam/monitor';
 import { applySession, type ClientSession, updateSessionFromResponse } from '../sessions';
 import {
   BufferPool,
@@ -124,6 +125,8 @@ export interface ConnectionOptions
   extendedMetadata: Promise<Document>;
   /** @internal */
   mongoLogger?: MongoLogger | undefined;
+  /** @internal */
+  parent?: Monitor;
 }
 
 /** @public */
--- a/src/sdam/monitor.ts
+++ b/src/sdam/monitor.ts
@@ -8,8 +8,15 @@ import { LEGACY_HELLO_COMMAND } from '../constants';
 import { MongoError, MongoErrorLabel, MongoNetworkTimeoutError } from '../error';
 import { MongoLoggableComponent } from '../mongo_logger';
 import { CancellationToken, TypedEventEmitter } from '../mongo_types';
-import type { Callback, EventEmitterWithState } from '../utils';
-import { calculateDurationInMs, makeStateMachine, now, ns } from '../utils';
+import {
+  calculateDurationInMs,
+  type Callback,
+  type EventEmitterWithState,
+  makeStateMachine,
+  MovingWindow,
+  now,
+  ns
+} from '../utils';
 import { ServerType, STATE_CLOSED, STATE_CLOSING } from './common';
 import {
   ServerHeartbeatFailedEvent,
@@ -25,8 +32,6 @@ const kServer = Symbol('server');
 const kMonitorId = Symbol('monitorId');
 /** @internal */
 const kCancellationToken = Symbol('cancellationToken');
-/** @internal */
-const kRoundTripTime = Symbol('roundTripTime');
 
 const STATE_IDLE = 'idle';
 const STATE_MONITORING = 'monitoring';
@@ -100,6 +105,7 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
   rttPinger?: RTTPinger;
   /** @internal */
   override component = MongoLoggableComponent.TOPOLOGY;
+  private rttSamplesMS: MovingWindow;
 
   constructor(server: Server, options: MonitorOptions) {
     super();
@@ -121,6 +127,7 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
     });
     this.isRunningInFaasEnv = getFAASEnv() != null;
     this.mongoLogger = this[kServer].topology.client?.mongoLogger;
+    this.rttSamplesMS = new MovingWindow(10);
 
     const cancellationToken = this[kCancellationToken];
     // TODO: refactor this to pull it directly from the pool, requires new ConnectionPool integration
@@ -135,7 +142,8 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
       useBigInt64: false,
       promoteLongs: true,
       promoteValues: true,
-      promoteBuffers: true
+      promoteBuffers: true,
+      parent: this
     };
 
     // ensure no authentication is used for monitoring
@@ -202,6 +210,22 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
     // close monitor
     this.emit('close');
     stateTransition(this, STATE_CLOSED);
+  }
+
+  get roundTripTime(): number {
+    return this.rttSamplesMS.average();
+  }
+
+  get minRoundTripTime(): number {
+    return this.rttSamplesMS.min();
+  }
+
+  addRttSample(rtt: number) {
+    this.rttSamplesMS.addSample(rtt);
+  }
+
+  clearRttSamples() {
+    this.rttSamplesMS.clear();
   }
 }
 
@@ -275,6 +299,8 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
       hello.isWritablePrimary = hello[LEGACY_HELLO_COMMAND];
     }
 
+    // FIXME: Figure out how to set this. Should duration be the instantaneous rtt or the averaged
+    // rtt?
     const duration =
       isAwaitable && monitor.rttPinger
         ? monitor.rttPinger.roundTripTime
@@ -329,6 +355,7 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
 
     if (isAwaitable && monitor.rttPinger == null) {
       monitor.rttPinger = new RTTPinger(
+        monitor,
         monitor[kCancellationToken],
         Object.assign(
           { heartbeatFrequencyMS: monitor.options.heartbeatFrequencyMS },
@@ -456,23 +483,27 @@ export class RTTPinger {
   /** @internal */
   [kCancellationToken]: CancellationToken;
   /** @internal */
-  [kRoundTripTime]: number;
-  /** @internal */
   [kMonitorId]: NodeJS.Timeout;
+  /** @internal */
+  monitor: Monitor;
   closed: boolean;
 
-  constructor(cancellationToken: CancellationToken, options: RTTPingerOptions) {
+  constructor(monitor: Monitor, cancellationToken: CancellationToken, options: RTTPingerOptions) {
     this.connection = undefined;
     this[kCancellationToken] = cancellationToken;
-    this[kRoundTripTime] = 0;
     this.closed = false;
+    this.monitor = monitor;
 
     const heartbeatFrequencyMS = options.heartbeatFrequencyMS;
     this[kMonitorId] = setTimeout(() => measureRoundTripTime(this, options), heartbeatFrequencyMS);
   }
 
   get roundTripTime(): number {
-    return this[kRoundTripTime];
+    return this.monitor.roundTripTime;
+  }
+
+  get minRoundTripTime(): number {
+    return this.monitor.minRoundTripTime;
   }
 
   close(): void {
@@ -503,7 +534,7 @@ function measureRoundTripTime(rttPinger: RTTPinger, options: RTTPingerOptions) {
       rttPinger.connection = conn;
     }
 
-    rttPinger[kRoundTripTime] = calculateDurationInMs(start);
+    rttPinger.monitor.addRttSample(calculateDurationInMs(start));
     rttPinger[kMonitorId] = setTimeout(
       () => measureRoundTripTime(rttPinger, options),
       heartbeatFrequencyMS
@@ -518,7 +549,7 @@ function measureRoundTripTime(rttPinger: RTTPinger, options: RTTPingerOptions) {
       },
       () => {
         rttPinger.connection = undefined;
-        rttPinger[kRoundTripTime] = 0;
+        rttPinger.monitor.clearRttSamples();
       }
     );
     return;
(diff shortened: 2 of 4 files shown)


src/sdam/monitor.ts
```diff
@@ -275,6 +299,8 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
       hello.isWritablePrimary = hello[LEGACY_HELLO_COMMAND];
     }
 
+    // FIXME: Figure out how to set this. Should duration be the instantaneous rtt or the averaged
```
W-A-James, line 302: This duration is linked to a particular heartbeat, so we should be using the instantaneous duration here.

src/sdam/server_selection.ts
```diff
@@ -223,9 +223,8 @@ function latencyWindowReducer(
   servers: ServerDescription[]
 ): ServerDescription[] {
   const low = servers.reduce(
-    (min: number, server: ServerDescription) =>
-      min === -1 ? server.roundTripTime : Math.min(server.roundTripTime, min),
-    -1
+    (min: number, server: ServerDescription) => Math.min(server.roundTripTime, min),
+    Infinity
```
W-A-James, line 227: Drive-by cleanup: don't need the ternary

src/sdam/monitor.ts
```diff
@@ -486,42 +530,48 @@ export class RTTPinger {
...
-  const start = now();
-  options.cancellationToken = rttPinger[kCancellationToken];
-  const heartbeatFrequencyMS = options.heartbeatFrequencyMS;
-
+function measureAndReschedule(
+  rttPinger: RTTPinger,
+  options: RTTPingerOptions,
+  start?: number,
+  conn?: Connection
+) {
```
W-A-James, line 538: Refactored to use fewer closures and clarify control flow, but willing to retain the previous structure

src/sdam/server.ts
```diff
@@ -467,15 +468,6 @@ export class Server extends TypedEventEmitter<ServerEvents> {
...
 
-function calculateRoundTripTime(oldRtt: number, duration: number): number {
-  if (oldRtt === -1) {
-    return duration;
-  }
-
-  const alpha = 0.2;
-  return alpha * duration + (1 - alpha) * oldRtt;
-}
-
```
W-A-James, line 478: Removed in favour of average calculation

src/sdam/monitor.ts
```diff
@@ -458,23 +495,30 @@ export class RTTPinger {
...
+  latestRTT?: number;
 
-  constructor(cancellationToken: CancellationToken, options: RTTPingerOptions) {
+  constructor(monitor: Monitor, cancellationToken: CancellationToken, options: RTTPingerOptions) {
     this.connection = undefined;
     this[kCancellationToken] = cancellationToken;
-    this[kRoundTripTime] = 0;
     this.closed = false;
+    this.monitor = monitor;
+    this.latestRTT = monitor.latestRTT;
```
W-A-James, line 510: Done in accordance with spec: https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.md#calculation-of-average-round-trip-times

src/sdam/server_description.ts
```diff
@@ -58,6 +60,7 @@ export class ServerDescription {
   minWireVersion: number;
   maxWireVersion: number;
   roundTripTime: number;
+  minRoundTripTime: number;
```
nbbeeken, line 63: Since this is public can we add documentation. Ideally we should have some for every field... but we can at least clarify what this is.

src/cmap/connection.ts
```diff
@@ -125,6 +126,8 @@ export interface ConnectionOptions
   extendedMetadata: Promise<Document>;
   /** @internal */
   mongoLogger?: MongoLogger | undefined;
+  /** @internal */
+  monitor?: Monitor;
```
nbbeeken, line 130: This is not exactly how I pictured this, because now pooled connections will need to be given access to the monitor of the server that they come from rather than the pool or server which is a more common shared structure. I think we should have less of a strict binding of exactly what type is needed here, connections may want to have a `parent` that fits a common type contract as little as `{server: Server}` or less `{serverDescription: ServerDescription}`, then monitor, pool, or anything else needs to provide a getter to satisfy that. 

We can skip working on this in this PR, I figured it fit because where we ended up storing the relvant information affected the way we obtained it. I think now that it is clear the serverDescription provides the RTT data we can figure out how to provide that later.

src/sdam/monitor.ts
```diff
@@ -275,11 +305,15 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
       hello.isWritablePrimary = hello[LEGACY_HELLO_COMMAND];
     }
 
+    // NOTE: here we use the latestRtt as this measurment corresponds with the value
```
nbbeeken, line 308:
```suggestion
    // NOTE: here we use the latestRtt as this measurement corresponds with the value
```

src/sdam/monitor.ts
```diff
@@ -329,6 +363,7 @@ function checkServer(monitor: Monitor, callback: Callback<Document | null>) {
 
     if (isAwaitable && monitor.rttPinger == null) {
       monitor.rttPinger = new RTTPinger(
+        monitor,
         monitor[kCancellationToken],
```
nbbeeken, line 367: Could this ctor just take the monitor now rather than passing these in seperately? 

I don't feel strongly about this, just clean up.

src/sdam/monitor.ts
nbbeeken: Fine by me, while we're at it, why not make this a class method, same with `measureRoundTripTime`

src/sdam/server.ts
```diff
@@ -175,7 +175,8 @@ export class Server extends TypedEventEmitter<ServerEvents> {
       this.emit(
         Server.DESCRIPTION_RECEIVED,
         new ServerDescription(this.description.hostAddress, event.reply, {
-          roundTripTime: calculateRoundTripTime(this.description.roundTripTime, event.duration)
+          roundTripTime: this.monitor?.roundTripTime,
+          minRoundTripTime: this.monitor?.minRoundTripTime
```
nbbeeken, line 179: So in LB mode how does this work? I guess there's never heartbeats and rtt is "assumed" 0. is that correct?

src/utils.ts
```diff
@@ -1344,3 +1344,56 @@ export async function fileIsAccessible(fileName: string, mode?: number) {
     return false;
   }
 }
+
+/** @internal */
+export class MovingWindow {
```
nbbeeken, line 1349: Correct me if I'm wrong, but the logic within this seems specific to our RTT rules, like the way less than 2 samples are handled. `class RTTSampler`?

src/utils.ts
```diff
@@ -1344,3 +1344,56 @@ export async function fileIsAccessible(fileName: string, mode?: number) {
...
+  length: number;
+  samples: Float64Array;
+
+  constructor(windowSize = 10) {
+    this.samples = new Float64Array(windowSize);
+    this.length = 0;
+    this.writeIndex = 0;
+  }
+
+  addSample(sample: number) {
```
nbbeeken, line 1361: Can you add documentation for ourselves about the expectations of this and the other methods? (specifically the circular buffer / window logic)

src/utils.ts
```diff
@@ -1344,3 +1344,56 @@ export async function fileIsAccessible(fileName: string, mode?: number) {
     return false;
   }
 }
+
+/** @internal */
+export class MovingWindow {
+  /** Index of the next slot to be overwritten */
+  private writeIndex: number;
+  length: number;
+  samples: Float64Array;
```
nbbeeken, line 1353: `private` until proven need elsewhere?

test/unit/sdam/monitor.test.ts
```diff
@@ -306,6 +307,47 @@ describe('monitoring', function () {
...
+              }
+              expect(monitor.roundTripTime).to.be.greaterThanOrEqual(heartbeatDurationMS);
+              monitor.close();
+            });
+          });
+        });
+      }
+    });
+
+    //describe('minRoundTripTime');
```
nbbeeken, line 350: Do we want 'minRoundTripTime' coverage?

test/unit/sdam/monitor.test.ts
```diff
@@ -306,6 +307,47 @@ describe('monitoring', function () {
...
+              const server = new MockServer(mockServer.address());
+              if (topologyVersion) server.description.topologyVersion = topologyVersion;
+              monitor = new Monitor(server as any, { serverMonitoringMode } as any);
+              monitor.connect();
+
+              for (let i = 0; i < 5; i++) {
+                await once(monitor, 'serverHeartbeatSucceeded');
+                monitor.requestCheck();
+              }
+              expect(monitor.roundTripTime).to.be.greaterThanOrEqual(heartbeatDurationMS);
```
nbbeeken, line 342: IIUC the test always uses the same delay, can we use a variation of delays so that if this stops being an average it would fail?

nbbeeken (changes requested):

src/utils.ts
baileympearson, line 1353: Any reason not to use a `List`?  that would simplify a number of methods in this class

src/utils.ts
```diff
@@ -1344,3 +1344,56 @@ export async function fileIsAccessible(fileName: string, mode?: number) {
...
+    this.samples[this.writeIndex++] = sample;
+    if (this.length < this.samples.length) {
+      this.length++;
+    }
+
+    this.writeIndex %= this.samples.length;
+  }
+
+  min(): number {
+    if (this.length < 2) return 0;
```
baileympearson, line 1371: Is this correct?  Wouldn't the min rtt when there's a single sample be `this.samples[0]`?

W-A-James: We're implementing the [CSOT spec](https://github.com/mongodb/specifications/blob/bbb335e60cd7ea1e0f7cd9a9443cb95fc9d3b64d/source/client-side-operations-timeout/client-side-operations-timeout.md#drivers-use-minimum-rtt-to-short-circuit-operations) here, so I can rename the class as Neal recommended to be more explicit about what this class is doing and why it's calculating the minimum this way.

From the linked spec:
> A previous version of this spec used the 90th percentile RTT to short circuit operations that might otherwise fail with a socket timeout. We decided to change this logic to avoid canceling operations that may have a high chance of succeeding and also remove a dependency on t-digest. Instead, drivers use the minimum RTT from the last 10 samples, or 0 until at least 2 samples have been recorded.

test/unit/sdam/monitor.test.ts
W-A-James: So something like giving it a 100ms, 200ms, 300ms, 400ms, 500ms and checking the average should be around 750ms?

test/unit/sdam/monitor.test.ts
W-A-James: Forgot to remove this, but I think we have enough coverage in for this with the `MovingWindow` class, which I'll rename and possibly move to the `monitor.ts` since that's the only place it's used and it implements spec-required behaviour.

src/cmap/connection.ts
W-A-James: I'll remove this from this PR then.

src/utils.ts
W-A-James: The reason behind doing it this way is that since this is a hot path (adding samples will happen every heartbeat for every server), using a circular buffer will likely be faster than creating and removing nodes from the list that then have to be garbage-collected.

src/sdam/server.ts
W-A-James: In load-balanced mode, we don't initialize the monitor at all, so minRoundTripTime stays by default as 0. The CSOT spec doesn't mention any special handling for load-balanced clusters.

src/utils.ts
```diff
@@ -1344,3 +1344,92 @@ export async function fileIsAccessible(fileName: string, mode?: number) {
...
+   * */
+  get last(): number | null {
+    if (this._length === 0) return null;
+    return this._rttSamples[this.writeIndex === 0 ? this._length - 1 : this.writeIndex - 1];
+  }
+
+  /** @remarks Added for testing */
+  get length(): number {
+    return this._length;
+  }
```
nbbeeken, line 1419: The tests can still access the `private` properties, I don't see the need for this indirection, and despite the comment this just makes length "public" again. If you want to avoid type errors in tests you can write `expect(rttSampler).to.have.property('length', 1);`

test/unit/sdam/monitor.test.ts
```diff
@@ -306,6 +307,53 @@ describe('monitoring', function () {
...
+                await once(monitor, 'serverHeartbeatSucceeded');
+                monitor.requestCheck();
+              }
+
+              const avgRtt = monitor.roundTripTime;
+              const expectedRtt = 300; // (100 + 200 + 300 + 400 + 500)/5 = 300
+              // avgRtt will strictly be greater than expectedRtt since setTimeout sets a minimum
+              // delay from the time of scheduling to the time of callback execution
+              expect(avgRtt - expectedRtt).greaterThanOrEqual(0);
+              expect(avgRtt - expectedRtt).lessThan(50);
```
nbbeeken, line 349: Can we express this as a `within` chai assertion? And is +/-50ms the best we can do?

nbbeeken (changes requested):

src/utils.ts
baileympearson: It's true that we sample per heartbeat, but does that make it a hot code path?  Heartbeats are processed every `heartbeatFrequencyMS`, which defaults to 10 seconds.  So by default, we're only processing rtt samples once per 10 seconds.

src/utils.ts
baileympearson: If this is specifically for RTT calculations and we're only using it in the monitor, can we move it out of the utils file?

test/unit/sdam/monitor.test.ts
W-A-James: Will refactor to use the `within` assertion

Running locally, I've seen up to 25ms discrepancy overall, so I doubled that to account for any differences in my hardware and running in CI.

nbbeeken: Fine by me then, that range isn't critical and might as well keep the test far away from the flaky zone

nbbeeken (approved):

src/utils.ts
W-A-James: Fair point, but since it's something that's user-configurable, it would still be good for us make this have a minimal performance impact. And in general, since this doesn't introduce too much extra complexity while reducing the number of objects we have to create and destroy, I think it's still better for us to do it this way.

baileympearson: That's fine.  Generally though I don't like to optimize before we have a functional implementation that is benchmarked.  I'd rather implement an unoptimized, simpler solution, let our benchmarks show it's a problem and _then_ optimize than optimize for things that we haven't shown to cause performance issues.

src/utils.ts
baileympearson: @W-A-James

W-A-James: Moved to `src/sdam/monitor.ts`

nbbeeken (approved):

baileympearson (approved):

PR #4061, opened 2024-03-28
refactor(NODE-6055): implement OnDemandDocument

nbbeeken:
#### What is changing?

- Adds OnDemandDocument
  - getValue(), getNumber(), *valuesAs() toObject()
- Adds StringFinder
  - caches strings as byte sequences
- Add a new `includes` byte comparison method

No

OnDemandDocument will provide the foundation for doing less up front parsing work.


src/cmap/wire_protocol/on_demand/string_finder.ts
```diff
@@ -0,0 +1,23 @@
+import { ByteUtils } from '../../../utils';
+
+/**
+ * @internal
+ * StringFinder is declared as a class so we can delete/test the cache
+ */
+export class StringFinder {
+  private static cache: Record<string, Uint8Array> = Object.create(null);
```
nbbeeken, line 8: Since our intention for now is to only use OnDemand internally this should only grow to be as large as the keys depended on by the driver (numbering in the 100s at absolute most). If this were made public we would need to think about eviction.

src/cmap/wire_protocol/on_demand/document.ts
```diff
@@ -0,0 +1,296 @@
...
+export type JSTypeOf = {
+  [BSONType.int]: number;
+  [BSONType.long]: bigint;
+  [BSONType.timestamp]: Timestamp;
+  [BSONType.binData]: Binary;
+  [BSONType.bool]: boolean;
+  [BSONType.objectId]: ObjectId;
+  [BSONType.string]: string;
+  [BSONType.date]: Date;
+  [BSONType.object]: OnDemandDocument;
```
nbbeeken, line 36: We kind of assume that the server would omit keys rather than set them to null or undefined. What / how should support for those values look like if at all?

baileympearson: This is no longer relevant, right?

src/cmap/wire_protocol/on_demand/document.ts
```diff
@@ -0,0 +1,320 @@
...
+  [BSONType.array]: OnDemandDocument;
+};
+
+/** @internal */
+export class OnDemandDocument {
+  /** Caches the existence of a property */
+  private readonly existenceOf: Record<string, boolean> = Object.create(null);
+  /** Caches a look up of name to element */
+  private readonly elementOf: Record<string, BSONElement> = Object.create(null);
```
baileympearson, line 47: isn't `existenceOf[name]` basically the same as `elementOf[name] != null`?  do we need `existenceOf`?

src/cmap/wire_protocol/on_demand/document.ts
```diff
@@ -0,0 +1,320 @@
...
+  nameOffset = 1,
+  nameLength = 2,
+  offset = 3,
+  length = 4
+}
+
+export type JSTypeOf = {
+  [BSONType.null]: null;
+  [BSONType.undefined]: null;
+  [BSONType.double]: number;
```
baileympearson, line 29: I'd expect this to be `Double`, and then `getNumber` / `getFlexibleNumber` would coerce the value to a `number`.  But right now, there's no way to decode a BSON double into a boxed type

src/utils.ts
```diff
@@ -1301,18 +1301,18 @@ export async function once<T>(ee: EventEmitter, name: string): Promise<T> {
   }
 }
 
-export function maybeAddIdToDocuments(
-  coll: Collection,
+export function maybeAddIdToDocuments<T extends Document>(
```
baileympearson, line 1304: Is this related to the work in this PR or just a nice improvement?

src/cmap/wire_protocol/on_demand/document.ts
```diff
@@ -0,0 +1,320 @@
...
+  /** Caches the revived javascript value */
+  private readonly valueOf: Record<string, any> = Object.create(null);
+  /** Caches the index of elements that have been named */
+  private readonly indexFound: Record<number, boolean> = Object.create(null);
+
+  /** All bson elements in this document */
+  private readonly elements: BSONElement[];
+
+  /** The number of elements in the BSON document */
+  public readonly length: number;
```
baileympearson, line 57: `numberOfElements`?

baileympearson (commented):
Leaving some comments

src/utils.ts
nbbeeken: Accidentally left over from the PoC, we may want this at some point, but we can do it when necessary.

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: Yes, resolved!

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: I would expect the opposite since the boxed double type isn't useful to the driver. and we _only_ need double for one case which is "ok" so sending it right to a JS number skips a step we will always perform anyway.

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: I'd still lean toward `size()` here, but less is more, we can revist this if it is needed. For example, cursor might want to know how many docs are left but we should solve for that case.

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: Not quite, if I look up a key that doesn't exist, I won't know of its non-existence until I get to the end of all the elements. Without this cache non-existence will look it up again.

src/cmap/wire_protocol/on_demand/document.ts
baileympearson: This is the only type in this list that has an implicit, lossy conversion.  It feels wrong to have one special case here.  Isn't the purpose of getFlexibleNumber to provide a wrapper that the driver can use that doesn't care about numeric box types?

nbbeeken:
https://github.com/mongodb/js-bson/pull/665 Has BSON clean-up changes needed to get this branch off a git hash.

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: from slack: this approach still maintains precision

W-A-James (dismissed):

package.json
```diff
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "@mongodb-js/saslprep": "^1.1.5",
-    "bson": "^6.5.0",
+    "bson": "github:mongodb/js-bson#NODE-6059-cleanup",
```
addaleax, line 29: TBD, right?

src/bson.ts
```diff
@@ -25,6 +27,21 @@ export {
...
 
+export type BSONElement = BSON.OnDemand['BSONElement'];
+
+export function parseToElementsToArray(bytes: Uint8Array, offset?: number): BSONElement[] {
+  const res = BSON.onDemand.parseToElements(bytes, offset);
+  return Array.isArray(res) ? res : [...res];
+}
+
+// eslint-disable-next-line @typescript-eslint/unbound-method
+export const getInt32LE = BSON.onDemand.NumberUtils.getInt32LE;
```
addaleax, line 38: Fwiw, I know https://github.com/mongodb/js-bson/pull/665 is already merged but you should be able to get around this eslint warnings in a 'clean' way by typing the methods as e.g.

```js
  getInt32LE: (this: void, source: Uint8Array, offset: number) => number;
```

src/cmap/wire_protocol/on_demand/document.ts
```diff
@@ -0,0 +1,315 @@
...
+};
+
+/** @internal */
+export class OnDemandDocument {
+  /** Caches the existence of a property */
+  private readonly existenceOf: Record<string, boolean> = Object.create(null);
+  /** Caches a look up of name to element */
+  private readonly elementOf: Record<string, BSONElement> = Object.create(null);
+  /** Caches the revived javascript value */
+  private readonly valueOf: Record<string, any> = Object.create(null);
```
addaleax, line 49: Since a primary goal of `OnDemandDocument` is to provide a more performant interface for accessing BSON, it might be worth grouping these into a single `Record<string, { element: BSONElement; value: any; ... }>` to avoid creating multiple dictionary objects, because currently accessing a single subdocument already creates at least 6 new objects instead of 1 :)

baileympearson: Also related to this and https://github.com/mongodb/node-mongodb-native/pull/4061#discussion_r1544748906 - could we simply store whether or not we've deserialized the whole BSON?  That would remove the need for `existenceOf` entirely.

package.json
nbbeeken: Yes, we're working on finalizing a bson release, and we'll put that here before merging.

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: I think I can manage to collapse the existence, element, and value cache's into one. I think the indexFound one will need to remain seperate. I went in this direction because implementing the cache that way creates a new object per element as it is found, but I guess an allocation is an allocation, so it doesn't matter much might as well consolidate. I'll give that a try

> could we simply store whether or not we've deserialized the whole BSON?

Can you expand on what you invision here? (specifcally "whole BSON") do you mean caching the result of a toObject call or tracking that we've finished finding all elements?

src/cmap/wire_protocol/on_demand/document.ts
```diff
@@ -0,0 +1,322 @@
...
+
+    if (cachedElement != null) {
+      return cachedElement;
+    }
+
+    for (let index = 0; index < this.elements.length; index++) {
+      const element = this.elements[index];
+
+      // skip this element if it has already been associated with a name
+      if (!this.indexFound[index] && this.isElementName(name, element)) {
```
baileympearson, line 111: instead of a second cache of the indexes that we've visited - could we just keep track of the furthest index we've looked into `this.elements`?

- before this loop, we can easily check if `furthestIndex === this.elements.length` and if the element doesn't exist in the cache, we know it doesn't exist in the document
- if `furthestIndex < this.elements.length`, then this loop starts at `furthestIndex`

package.json
nbbeeken: Updated BSON version

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: In the current implementation, this would make elements that proceed `furthestIndex` inaccessible.

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: I removed existence, value, and element cache, in favor of one that tracks all of those pieces of work. The `cache[name]` being `false` is not checked outside of the getElement function so as long as any methods go through getElement the same nullish checks are used for the element itself and its potentially revived value.

baileympearson (approved):

PR #4062, opened 2024-03-28
refactor(NODE-6056): implement MongoDBResponse class

nbbeeken:
#### What is changing?

- Fixes and closes: [NODE-3946](https://jira.mongodb.org/browse/NODE-3946)
- Moves BSON parsing to OnDemandDocument
- Adds a subclass that provides getters for common command response properties

No

- In service of future improvements and optimizations
- Typesafe responses

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -416,7 +416,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   private async *sendWire(
     message: WriteProtocolMessageType,
     options: CommandOptions,
-    returnAs?: typeof MongoDBResponse
+    responseType?: typeof MongoDBResponse
   ): AsyncGenerator<MongoDBResponse> {
     this.throwIfAborted();
 
@@ -443,7 +443,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
         this.socket.setTimeout(0);
         const bson = response.parse();
 
-        const document = new (returnAs ?? MongoDBResponse)(bson, 0, false);
+        const document = new (responseType ?? MongoDBResponse)(bson, 0, false);
 
         yield document;
         this.throwIfAborted();
@@ -463,7 +463,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     ns: MongoDBNamespace,
     command: Document,
     options: CommandOptions = {},
-    returnAs?: typeof MongoDBResponse
+    responseType?: typeof MongoDBResponse
   ) {
     const message = this.prepareCommand(ns.db, command, options);
 
@@ -482,7 +482,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     let document;
     try {
       this.throwIfAborted();
-      for await (document of this.sendWire(message, options, returnAs)) {
+      for await (document of this.sendWire(message, options, responseType)) {
         if (options.session != null) {
           updateSessionFromResponse(options.session, document);
         }
@@ -520,7 +520,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
           );
         }
 
-        if (returnAs == null) {
+        if (responseType == null) {
           // If `documentsReturnedIn` not set or raw is not enabled, use input bson options
           // Otherwise, support raw flag. Raw only works for cursors that hardcode firstBatch/nextBatch fields
           const bsonOptions =
@@ -578,7 +578,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     ns: MongoDBNamespace,
     command: Document,
     options: CommandOptions | undefined,
-    returnAs: T | undefined
+    responseType: T | undefined
   ): Promise<T extends undefined ? Document : InstanceType<T>>;
 
   public async command(
@@ -599,10 +599,10 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     ns: MongoDBNamespace,
     command: Document,
     options: CommandOptions = {},
-    returnAs?: typeof MongoDBResponse
+    responseType?: typeof MongoDBResponse
   ): Promise<Document> {
     this.throwIfAborted();
-    for await (const document of this.sendCommand(ns, command, options, returnAs)) {
+    for await (const document of this.sendCommand(ns, command, options, responseType)) {
       return document;
     }
     throw new MongoUnexpectedServerResponseError('Unable to get response from server');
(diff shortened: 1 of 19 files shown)


src/sdam/common.ts
```diff
@@ -58,7 +58,7 @@ export function drainTimerQueue(queue: TimerQueue): void {
 /** @public */
 export interface ClusterTime {
   clusterTime: Timestamp;
-  signature: {
+  signature?: {
```
nbbeeken, line 61: Public API was incorrect

nbbeeken (commented):
https://spruce.mongodb.com/version/66085dc31549960007d3a292

test/unit/cmap/wire_protocol/responses.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    expect(new MongoDBResponse(BSON.serialize({ ok: 1 }))).to.be.instanceOf(OnDemandDocument);
+  });
+
+  context('get isError', () => {
+    it('returns true when ok is 0', () => {
+      const doc = new MongoDBResponse(BSON.serialize({ ok: 0 }));
+      expect(doc.isError).to.be.true;
+    });
+
+    it('short circuits detection of $err, errmsg, code', () => {
```
baileympearson, line 17: are we missing tests to check that `isError` returns true if `$err`, `errmsg` or `code` are present and truthy?

test/unit/cmap/wire_protocol/responses.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      expect(doc).to.not.have.property('cache.$err');
+      expect(doc).to.not.have.property('cache.errmsg');
+      expect(doc).to.not.have.property('cache.code');
+    });
+  });
+
+  context('utf8 validation', () => {
+    afterEach(() => sinon.restore());
+
+    context('when enableUtf8Validation is not specified', () => {
```
baileympearson, line 29: Can we preserve the utf8 validation tests from `commands.test.ts` that actually confirm we throw on invalid utf8 instead of just spying with sinon?

test/types/connection.test-d.ts
```diff
@@ -0,0 +1,25 @@
+import { expectError, expectType } from 'tsd';
+
+import { type Connection, type Document, MongoDBResponse, ns } from '../mongodb';
+
+declare const connection: Connection;
+
+expectType<Document>(await connection.command(ns('a'), { cmd: 1 }));
+expectType<Document>(await connection.command(ns('a'), { cmd: 1 }, undefined));
+expectType<Document>(await connection.command(ns('a'), { cmd: 1 }, { socketTimeoutMS: 1 }));
+// TODO fix TS or simplify arguments
```
baileympearson, line 10: todo for this PR or future ticket?

src/sdam/common.ts
baileympearson: Seems fine as a fix but in a separate PR.  One line fix PR maybe?

baileympearson (changes requested):
nothing major, looks good overall

test/types/connection.test-d.ts
nbbeeken: Either, probably this PR, I'll reach out in the team channel, I am having trouble getting the API to properly return `Document` vs the 4th argument's type. The issue is for now, we have all our usages calling this function with 3 arguments so we get back document but our layers of the driver are going to need to ferry the correct type based on that third argument. CryptoConnection.command is an example. So I need a way to type something that passes in a 4th argument that is potentially nullish accurately (at some point, not an blocker for this PR).

test/unit/cmap/wire_protocol/responses.test.ts
nbbeeken: My thinking was what coverage does that provide that isn't provided by BSON's testing of the utf validation options/behavior? I'll see if I can add it back the tough part is that hex blob needs to be refactored 😅

test/unit/cmap/wire_protocol/responses.test.ts
nbbeeken: Good call, previously the code was `document.ok === 0 || document.$err || document.errmsg || document.code` so it wasn't just that those properties existed but that they were truthy values. Do you think it is acceptable to change this to a has check?

src/cmap/connection.ts
```diff
@@ -509,14 +513,29 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
             new CommandSucceededEvent(
               this,
               message,
-              options.noResponse ? undefined : document,
+              options.noResponse ? undefined : document.toObject(options),
```
nbbeeken, line 516: This and the other instances of toObject make the deep clone we preform for command monitoring redundant. Should we remove that here or should I strive to pass through the same instance? @baileympearson (sorry should've called this out sooner)

baileympearson: Should we be caching the full deserialized object in `MongoDBResponse.toObject()`?  If so, wouldn't we still need the deep clone?

It seems like caching would be necessary to avoid redundant deserialization when command monitoring is enabled

nbbeeken: Yea, we can defer changing this, I'll reuse the same deserialized object and leave the deep cloning be. 

I say defer because I think at one point we considered that deserializing twice would be better than cloning, we can make that choice intentionally later.

test/types/connection.test-d.ts
nbbeeken: Cleaned up the overloads, we don't need to support this usage.

test/unit/cmap/wire_protocol/responses.test.ts
baileympearson: I can't think of a scenario where the server might return a response with one of these fields present with a falsey value.  I don't think it would muck up the implementation much to do something like the following, so I guess I don't see much of a reason not to leave the logic as-is:

```typescriopt
get isError(): boolean {
  return Boolean(
    this.ok ?? this.$err ?? this.errmsg ?? this.code
  )
}
```

nbbeeken: https://github.com/mongodb/mongo/blob/master/src/mongo/db/dbmessage.cpp#L159-L162

https://github.com/mongodb/mongo/blob/master/src/mongo/db/basic_types.idl#L442-L444

I have confidence in our change here to existence check, and at least currently we still run the error object through the same deserialization so we don't have to opt into binding these fields to a specific type.

baileympearson (approved):

W-A-James (approved):
Looks great! Looking at the perf tests, we're also doing well with these changes. 🚀

PR #4063, opened 2024-04-01
fix(NODE-5785): do not deserialize heartbeat events with `promoteBuffers` set to `true`

baileympearson:
#### What is changing?

This PR sets `promoteBuffers` to false for the monitor class.  This change is almost entirely invisible to users because monitors are not publicly exposed.

Server heartbeat events and log message _are_ impacted by this change, but we don't currently include the types of each field in the server's heartbeat as a part of our semver-supported API.

No.

Setting `promoteBuffers: true` increases the size of an EJSON.stringified heartbeat event, increasing the likelihood that log messages are truncated.

### DIFF
--- a/src/sdam/monitor.ts
+++ b/src/sdam/monitor.ts
@@ -135,7 +135,7 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
       useBigInt64: false,
       promoteLongs: true,
       promoteValues: true,
-      promoteBuffers: true
+      promoteBuffers: false
     };
 
     // ensure no authentication is used for monitoring
--- a/test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring_logging.test.ts
+++ b/test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring_logging.test.ts
@@ -33,7 +33,7 @@ describe('SDAM Logging Integration Tests', function () {
   });
 
   it(
-    'does not promote buffers',
+    'the monitor does not promote buffers in server heartbeats',
     {
       requires: {
         topology: ['replicaset', 'sharded']
@@ -50,7 +50,7 @@ describe('SDAM Logging Integration Tests', function () {
   );
 
   it(
-    'does not log messages `reply` field',
+    'heartbeat log messages do not promote buffers in the `reply` field',
     {
       requires: {
         topology: ['replicaset', 'sharded']
--- a/test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring_logging.test.ts
+++ b/test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring_logging.test.ts
@@ -1,19 +1,13 @@
 import { expect } from 'chai';
+import { setTimeout } from 'timers/promises';
 
-import {
-  Binary,
-  EJSON,
-  type MongoClient,
-  type MongoDBLogWritable,
-  type ServerHeartbeatSucceededEvent
-} from '../../mongodb';
+import { Binary, type MongoDBLogWritable, type ServerHeartbeatSucceededEvent } from '../../mongodb';
 
 describe('SDAM Logging Integration Tests', function () {
-  let client: MongoClient;
   const monitoringEvents: Array<ServerHeartbeatSucceededEvent> = [];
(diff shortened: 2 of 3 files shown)


test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring_logging.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    await setTimeout(100);
+    await client.close();
+  });
+
+  it(
+    'the monitor does not promote buffers in server heartbeats',
+    {
+      requires: {
+        topology: ['replicaset', 'sharded']
+      }
```
W-A-James, line 40: Any reason not to run these on standalone as well?

W-A-James (commented):
LGTM, just one question

test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring_logging.test.ts
baileympearson: The $clusterTime field isn't present on heartbeats from standalone servers

W-A-James (approved):

test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring_logging.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        mongodbLogComponentSeverities: { topology: 'trace' },
+        [Symbol.for('@@mdb.enableMongoLogger')]: true
+      }
+    );
+
+    client.on('serverHeartbeatSucceeded', monitoringEvents.push.bind(monitoringEvents));
+
+    await client.connect();
+    // give the driver a chance to connect to all servers and collect some heartbeats
+    await setTimeout(100);
```
aditi-khare-mongoDB, line 31: LGTM, but just a knowledge question, why is 100ms a sufficient time?

nbbeeken:
What is the public-facing fix here?

aditi-khare-mongoDB (approved):

test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring_logging.test.ts
baileympearson: Also because you might notice this sleep is after the awaited `connect`.  Connect will (by default) resolve when the driver finds a primary.  That means that `connect` can resolve before connecting to secondaries, arbiters or other servers.

baileympearson:
> What is the public-facing fix here?

@nbbeeken server heartbeat events I guess.  Are you thinking this is a "refactor" instead?

nbbeeken:
Oh I see, ServerHeartbeatSucceededEvent's `reply` will now be different. Is this a breaking change? While `reply` is server version/topology dependent, this change is caused by a client-side hardcoded setting that is not otherwise configurable.

baileympearson:
> Oh I see, ServerHeartbeatSucceededEvent's `reply` will now be different. Is this a breaking change? While `reply` is server version/topology dependent, this change is caused by a client-side hardcoded setting that is not otherwise configurable.

My thought is no - because we type the `reply` as a `Document`.  I interpret that to mean that the contents of the reply are not guaranteed to be stable (in addition to the fact that `reply` varies based on topology and server version).

nbbeeken:
I mentioned the fact that it changes with server version/topology because I think that it makes sense that is the _only_ way `reply` might change. There's server documentation for how commands respond for each of those cases. This change is exclusively client-side and there's no documentation. A lot of our APIs return `Document` but I do not think we can make intentional changes to the shape of the return value because of that.

src/sdam/monitor.ts
```diff
@@ -135,7 +135,7 @@ export class Monitor extends TypedEventEmitter<MonitorEvents> {
       useBigInt64: false,
       promoteLongs: true,
       promoteValues: true,
-      promoteBuffers: true
+      promoteBuffers: false
```
nbbeeken, line 138: Leaving a comment for viz, discussing public API concerns on slack.

nbbeeken (changes requested):

PR #4067, opened 2024-04-03
feat(NODE-5762): include `cause` and package name for all `MongoMissingDependencyError`s

baileympearson:
#### What is changing?

`MongoMissingDependencyError` now includes two required properties: 

- `dependencyName` - the name of the missing dependency
- `cause` - the error thrown when attempting to `require` the missing dependency.

### cause and package name included for MongoMissingDependencyErrors

`MongoMissingDependencyError`s now include a `cause` and a `dependencyName` field, which can be used to programmatically determine which package is missing and why the driver failed to load it.

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -695,7 +695,7 @@ export class CryptoConnection extends Connection {
   ): Promise<Document> {
     const { autoEncrypter } = this;
     if (!autoEncrypter) {
-      // TODO(NODE-6065): throw a MongoMissingDependencyError in Node V7
+      // TODO(NODE-6065): throw a MongoRuntimeError in Node V7
       // @ts-expect-error No cause provided because there is no underlying error.
       throw new MongoMissingDependencyError('No AutoEncrypter available for encryption', {
         dependencyName: 'n/a'
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -13,8 +13,8 @@ import {
   UNPINNED
 } from '../constants';
 import {
+  MongoAPIError,
   MongoCompatibilityError,
-  MongoMissingDependencyError,
   MongoNetworkError,
   MongoNetworkTimeoutError,
   MongoParseError,
@@ -695,7 +695,7 @@ export class CryptoConnection extends Connection {
   ): Promise<Document> {
     const { autoEncrypter } = this;
     if (!autoEncrypter) {
-      throw new MongoMissingDependencyError('No AutoEncrypter available for encryption');
+      throw new MongoAPIError('No AutoEncrypter available for encryption');
     }
 
     const serverWireVersion = maxWireVersion(this);
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -13,8 +13,8 @@ import {
   UNPINNED
 } from '../constants';
 import {
-  MongoAPIError,
   MongoCompatibilityError,
+  MongoMissingDependencyError,
   MongoNetworkError,
   MongoNetworkTimeoutError,
   MongoParseError,
@@ -695,7 +695,11 @@ export class CryptoConnection extends Connection {
   ): Promise<Document> {
     const { autoEncrypter } = this;
     if (!autoEncrypter) {
-      throw new MongoAPIError('No AutoEncrypter available for encryption');
+      // TODO(NODE-6065): throw a MongoMissingDependencyError in Node V7
+      // @ts-expect-error No cause provided because there is no underlying error.
+      throw new MongoMissingDependencyError('No AutoEncrypter available for encryption', {
+        dependencyName: 'n/a'
+      });
     }
 
     const serverWireVersion = maxWireVersion(this);
--- a/src/deps.ts
+++ b/src/deps.ts
@@ -29,8 +29,7 @@ export function getKerberos(): Kerberos {
     kerberos = makeErrorModule(
       new MongoMissingDependencyError(
         'Optional module `kerberos` not found. Please install it to enable kerberos authentication',
-        'kerberos',
-        { cause: error }
+        { cause: error, dependencyName: 'kerberos' }
       )
     );
   }
@@ -69,8 +68,7 @@ export function getZstdLibrary(): ZStandardLib | { kModuleError: MongoMissingDep
     ZStandard = makeErrorModule(
       new MongoMissingDependencyError(
         'Optional module `@mongodb-js/zstd` not found. Please install it to enable zstd compression',
-        'zstd',
-        { cause: error }
+        { cause: error, dependencyName: 'zstd' }
       )
     );
   }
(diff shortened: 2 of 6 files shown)


src/cmap/connection.ts
```diff
@@ -695,7 +695,7 @@ export class CryptoConnection extends Connection {
   ): Promise<Document> {
     const { autoEncrypter } = this;
     if (!autoEncrypter) {
-      throw new MongoMissingDependencyError('No AutoEncrypter available for encryption');
+      throw new MongoAPIError('No AutoEncrypter available for encryption');
```
baileympearson, line 698: This isn't really a "missing dependency" error - if anything, I think it's probably best captured by a `MongoRuntimeError` because we shouldn't be instantiating crypto connections if an autoEncrypter is not provided.  I chose to use a `MongoAPIError` because that mostly preserves the error hierarchy (`MongoAPIError` is the superclass of `MongoMissingDependencyError`).

If we decide this change is unacceptable, I can revert it and use a `ts-expect-error` instead.

src/error.ts
```diff
@@ -1019,8 +1022,10 @@ export class MongoMissingDependencyError extends MongoAPIError {
    *
    * @public
    **/
-  constructor(message: string, options: { cause?: Error } = {}) {
+  constructor(message: string, options: { cause: Error; dependencyName: string }) {
     super(message, options);
+    this.dependencyName = options.dependencyName;
+    this.cause = options.cause;
```
nbbeeken, line 1028: Why did we add this? can't we just let cause travel up to the super class?

src/encrypter.ts
```diff
@@ -128,7 +128,11 @@ export class Encrypter {
     if ('kModuleError' in mongodbClientEncryption) {
       throw new MongoMissingDependencyError(
         'Auto-encryption requested, but the module is not installed. ' +
-          'Please add `mongodb-client-encryption` as a dependency of your project'
+          'Please add `mongodb-client-encryption` as a dependency of your project',
+        {
+          cause: mongodbClientEncryption['kModuleError'].cause,
```
nbbeeken, line 133:
```suggestion
          cause: mongodbClientEncryption['kModuleError'],
```

Why not include the entire chain of errors? then we don't lose the full stack

src/deps.ts
```diff
@@ -256,10 +268,10 @@ export function getMongoDBClientEncryption():
     // Cannot be moved to helper utility function, bundlers search and replace the actual require call
     // in a way that makes this line throw at bundle time, not runtime, catching here will make bundling succeed
     mongodbClientEncryption = require('mongodb-client-encryption');
-  } catch (cause) {
+  } catch (error) {
     const kModuleError = new MongoMissingDependencyError(
       'Optional module `mongodb-client-encryption` not found. Please install it to use auto encryption or ClientEncryption.',
-      { cause }
+      { cause: error, dependencyName: 'mongodb-client-encryption' }
```
nbbeeken, line 274: Why the refactor away from shorthand properties?

src/cmap/connection.ts
nbbeeken: While an accurate analysis is this an acceptable API change?

> I can revert it and use a ts-expect-error here instead.

To exempt what issue exactly?

nbbeeken (changes requested):

src/deps.ts
baileympearson: Do you mean `cause -> error`?  Primarily for consistency with the rest of the file, since they all catch errors named `error`.

src/error.ts
baileympearson: Because I `override cause: Error` in this class (it's now required to construct a MissingDependencyError), TS complains that we didn't initialize the `cause` property in the constructor even though it's assigned in the superclass.

src/deps.ts
nbbeeken: Sure, I don't mind either way, I prefer naming it cause for the shorthand but just my 🧅

src/cmap/connection.ts
baileympearson: There would be a type error because we didn't provide a cause or dependencyName to the MongoMissingDependencyError  constructor.

> While an accurate analysis is this an acceptable API change?

I think you could argue either way.  I think this is a correctness issue but I'm also happy leaving it as-is.

src/encrypter.ts
baileympearson: kModuleError is already a MissingDependencyError, so we'd end up with a missing dependency error wrapping a missing dependency error wrapping an import error.  The logic here is basically equivalent to:

```
mongodbClientEncryption['kModuleError'].message = 'Please add `mongodb-client-encryption` as a dependency of your project'
throw mongodbClientEncryption['kModuleError']
```

But I didn't want to modify the global `kModuleError` property, so I chose to continue instantiating a new error instead.

src/error.ts
nbbeeken: I assume you added `override cause: Error` because it will type `cause` to 1. always exist and 2. be an error type rather than (I think) `unknown`? 

Could we use `!` like: `override cause!: Error`? By writing `this.cause =` this "cause" property has a different descriptor than if we allow the error super class to attach it.

src/cmap/connection.ts
nbbeeken: It wouldn't be entirely off base to pass in `'mongodb-client-encryption'` as the dependency name here. I am leaning towards leaving the error type as is to reduce the chance of surprising anyone. I agree it may be a correctness bug but only in the errors meaning whereas fixing it has the potential to break, we can correct it in a major.

src/encrypter.ts
nbbeeken: > we'd end up with a missing dependency error wrapping a missing dependency error wrapping an import error

What's wrong with having that chain of errors though? It is the "truth" in the sense of what a "cause" is supposed to mean and improves the traceability. It is one step away today but that might not always be the case

baileympearson: > It is the "truth" in the sense of what a "cause" is supposed to mean

Not really.  The first missing dependency error did not "cause" the second, we're just transforming the error message.

src/cmap/connection.ts
baileympearson: Done.

src/encrypter.ts
nbbeeken: True, but the full stack trace is what I'm hoping to preserve here

src/error.ts
nbbeeken: (maybe got lost) can we make this change?

nbbeeken (changes requested):

nbbeeken (approved):

src/cmap/connection.ts
```diff
@@ -695,7 +695,11 @@ export class CryptoConnection extends Connection {
   ): Promise<Document> {
     const { autoEncrypter } = this;
     if (!autoEncrypter) {
-      throw new MongoMissingDependencyError('No AutoEncrypter available for encryption');
+      // TODO(NODE-6065): throw a MongoMissingDependencyError in Node V7
```
addaleax, line 698:
```suggestion
      // TODO(NODE-6065): throw a MongoRuntimeError in Node V7
```

according to the ticket

addaleax (dismissed):

nbbeeken (approved):

PR #4070, opened 2024-04-04
fix(NODE-5530): make topology descriptions JSON stringifiable 

baileympearson:
#### What is changing?

This is basically entirely invisible to users.  However, the TopologyDescription is a field on `MongoServerSelectionError`s, so if a user is calling JSON.stringify on their errors, they will now see the `servers` field in the output instead of an empty object.

No.

### TopologyDescription now properly stringifies itself to JSON

The `TopologyDescription` class is exposed by the driver in server selection errors and topology monitoring events to provide insight into the driver's current representation of the server's topology and to aid in debugging.  However, the TopologyDescription uses `Map`s internally, which get serialized to `{}` when JSON stringified.  We recommend using Node's `util.inspect()` helper to print topology descriptions because `inspect` properly handles all JS types and all types we use in the driver.  However, if JSON must be used, the `TopologyDescription` now provides a custom `toJSON()` hook:

```typescript
client.on('topologyDescriptionChanged', ({ newDescription }) => {
   // recommended!
	console.log('topology description changed', inspect(newDescription, { depth: Infinity, colors: true }))

    // now properly prints the entire topology description
	console.log('topology description changed', JSON.stringify(newDescription))
});
```

### DIFF
--- a/src/sdam/topology_description.ts
+++ b/src/sdam/topology_description.ts
@@ -343,6 +343,12 @@ export class TopologyDescription {
     return this.servers.has(address);
   }
 
+  /**
+   * Returns a JSON-serializable representation of the TopologyDescription.  This is primarily
+   * intended for use with JSON.stringify().
+   *
+   * This method will not throw.
+   */
   toJSON() {
     return EJSON.serialize(this);
   }
--- a/src/sdam/topology_description.ts
+++ b/src/sdam/topology_description.ts
@@ -1,4 +1,4 @@
-import type { ObjectId } from '../bson';
+import { EJSON, type ObjectId } from '../bson';
 import * as WIRE_CONSTANTS from '../cmap/wire_protocol/constants';
 import { type MongoError, MongoRuntimeError } from '../error';
 import { compareObjectId, shuffle } from '../utils';
@@ -341,6 +341,10 @@ export class TopologyDescription {
    */
   hasServer(address: string): boolean {
     return this.servers.has(address);
+  }
+
+  toJSON() {
+    return EJSON.serialize(this);
   }
 }
 
--- a/test/integration/server-discovery-and-monitoring/topology_description.test.ts
+++ b/test/integration/server-discovery-and-monitoring/topology_description.test.ts
@@ -14,7 +14,22 @@ describe('TopologyDescription (integration tests)', function () {
     await client.close();
   });
 
+  beforeEach(async function () {
+    client = this.configuration.newClient();
+    await client.connect();
+  });
+
   context('options', function () {
+    let client: MongoClient;
+
+    afterEach(async function () {
+      await client.close();
+    });
+
+    beforeEach(async function () {
+      client = this.configuration.newClient();
+    });
+
     context('localThresholdMS', function () {
       it('should default to 15ms', async function () {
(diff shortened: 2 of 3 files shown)


src/sdam/topology_description.ts
```diff
@@ -342,6 +342,10 @@ export class TopologyDescription {
   hasServer(address: string): boolean {
     return this.servers.has(address);
   }
+
+  toJSON() {
```
aditi-khare-mongoDB, line 346: Can `EJSON.serialize` throw? To my knowledge, it `EJSON.serialize`, uses `EJSON.stringify` then reparses it, and stringify throws on a number of occasions.

test/integration/server-discovery-and-monitoring/topology_description.test.ts
```diff
@@ -14,7 +14,22 @@ describe('TopologyDescription (integration tests)', function () {
     await client.close();
   });
 
+  beforeEach(async function () {
+    client = this.configuration.newClient();
```
aditi-khare-mongoDB, line 18: What's the rationale behind adding this new beforeEach block without a `afterEach client.close`?

aditi-khare-mongoDB (changes requested):

test/integration/server-discovery-and-monitoring/topology_description.test.ts
baileympearson: The `afterEach` was already there (lines 13-15)

src/sdam/topology_description.ts
baileympearson: Yes, EJSON.serialize() can throw.

aditi-khare-mongoDB: Do we want to add in a try catch here, with a default result, or a doc comment that this function might throw?

aditi-khare-mongoDB (changes requested):

src/sdam/topology_description.ts
baileympearson: I can add a doc comment if you'd think that's helpful.  This API isn't intended for direct consumption (although a user could call `TopologyDescription.toJSON()` since it's public).  `toJSON` hooks are called automatically by JSON.stringify(), so it's primarily for use with something like `JSON.stringify(description)`.

I think it's alright as-is.  JSON.stringify() can throw, so anyone calling JSON.stringify() on an object they don't control should be try-catching it.

aditi-khare-mongoDB: I'd prefer to add a doc comment since it's technically a use case, but also I'm up to ask the team if you want more input!

baileympearson: I did add a doc comment but I also looked more closely at the TopologyDescription.  There's nothing that `toJSON()` serializes that would throw when passed to EJSON.serialize() so I marked the function as not throwing in the doc comment.  How does that look?

aditi-khare-mongoDB (approved):

W-A-James (commented):
LGTM but can we have a short release note since this is public facing

aditi-khare-mongoDB (approved):

W-A-James (approved):

PR #4072, opened 2024-04-08
fix(NODE-6029): update types for collection listing indexes

prenaissance:
#### What is changing?

Update the return type of `Collection.prototype.indexes` and `Collection.prototype.indexInformation` to an array of indexes with full properties when the `full` option is true and to a Record of \<index name\> - \<keys and direction\> when the full option is false.

No.

The old types were incorrect in regards to the `full` option and were very wide. This will avoid making some type mistakes when using these methods.

### Index information helpers have accurate Typescript return types

`Collection.indexInformation()`, `Collection.indexes()` and `Db.indexInformation()` are helpers that return index information for a given collection or database.  These helpers take an option, `full`, that configures whether the return value contains full index descriptions or a compact summary:

```typescript
collection.indexes({ full: true });   // returns an array of index descriptions
collection.indexes({ full: false });  // returns an object, mapping index names to index keys
```

However, the Typescript return type of these helpers was always `Document`.  Thanks to @prenaissance, these helpers now have accurate type information!  The helpers return a new type,  `IndexDescriptionCompact | IndexDescriptionInfo[]`, which accurately reflects the return type of these helpers.  The helpers also support type narrowing by providing a boolean literal as an option to the API:

```typescript
collection.indexes();   // returns `IndexDescriptionCompact | IndexDescriptionInfo[]`
collection.indexes({ full: false });  // returns an `IndexDescriptionCompact`
collection.indexes({ full: true });  // returns an `IndexDescriptionInfo[]`

collection.indexInfo();   // returns `IndexDescriptionCompact | IndexDescriptionInfo[]`
collection.indexInfo({ full: false });  // returns an `IndexDescriptionCompact`
collection.indexInfo({ full: true });  // returns an `IndexDescriptionInfo[]`

db.indexInfo();   // returns `IndexDescriptionCompact | IndexDescriptionInfo[]`
db.indexInfo({ full: false });  // returns an `IndexDescriptionCompact`
db.indexInfo({ full: true });  // returns an `IndexDescriptionInfo[]`
```

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -701,13 +701,21 @@ export class Collection<TSchema extends Document = Document> {
    *
    * @param options - Optional settings for the command
    */
-  async indexInformation<TFull extends boolean = false>(
-    options?: IndexInformationOptions<TFull>
-  ): Promise<IndexesInformation<TFull>> {
-    return (await this.indexes({
+  indexInformation(
+    options: IndexInformationOptions & { full: true }
+  ): Promise<IndexDescriptionInfo[]>;
+  indexInformation(
+    options: IndexInformationOptions & { full?: false }
+  ): Promise<IndexDescriptionCompact>;
+  indexInformation(
+    options: IndexInformationOptions & { full: boolean }
+  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
+  indexInformation(): Promise<IndexDescriptionCompact>;
+  async indexInformation(options?: IndexInformationOptions) {
+    return await this.indexes({
       ...options,
       full: options?.full ?? false
-    })) as IndexesInformation<TFull>;
+    });
   }
 
   /**
@@ -811,20 +819,24 @@ export class Collection<TSchema extends Document = Document> {
    *
    * @param options - Optional settings for the command
    */
-  async indexes<TFull extends boolean = true>(
-    options?: IndexInformationOptions<TFull>
-  ): Promise<IndexesInformation<TFull>> {
+  indexes(options: IndexInformationOptions & { full?: true }): Promise<IndexDescriptionInfo[]>;
+  indexes(options: IndexInformationOptions & { full: false }): Promise<IndexDescriptionCompact>;
+  indexes(
+    options: IndexInformationOptions & { full: boolean }
+  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
+  indexes(): Promise<IndexDescriptionInfo[]>;
+  async indexes(options?: IndexInformationOptions) {
     const indexes: IndexDescriptionInfo[] = await this.listIndexes(options).toArray();
     const full = options?.full ?? true;
     if (full) {
-      return indexes as IndexesInformation<TFull>;
+      return indexes;
     }
 
     const object: IndexDescriptionCompact = Object.fromEntries(
       indexes.map(({ name, key }) => [name, Object.entries(key)])
     );
 
-    return object as IndexesInformation<TFull>;
+    return object;
   }
 
   /**
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -701,7 +701,7 @@ export class Collection<TSchema extends Document = Document> {
     options: IndexInformationOptions & { full?: false }
   ): Promise<IndexDescriptionCompact>;
   indexInformation(
-    options: IndexInformationOptions & { full: boolean }
+    options: IndexInformationOptions
   ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
   indexInformation(): Promise<IndexDescriptionCompact>;
   async indexInformation(
@@ -817,9 +817,9 @@ export class Collection<TSchema extends Document = Document> {
   indexes(options: IndexInformationOptions & { full?: true }): Promise<IndexDescriptionInfo[]>;
   indexes(options: IndexInformationOptions & { full: false }): Promise<IndexDescriptionCompact>;
   indexes(
-    options: IndexInformationOptions & { full: boolean }
+    options: IndexInformationOptions
   ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
-  indexes(): Promise<IndexDescriptionInfo[]>;
+  indexes(options?: ListIndexesOptions): Promise<IndexDescriptionInfo[]>;
   async indexes(
     options?: IndexInformationOptions
   ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]> {
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -704,7 +704,9 @@ export class Collection<TSchema extends Document = Document> {
     options: IndexInformationOptions & { full: boolean }
   ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
   indexInformation(): Promise<IndexDescriptionCompact>;
-  async indexInformation(options?: IndexInformationOptions) {
+  async indexInformation(
+    options?: IndexInformationOptions
+  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]> {
     return await this.indexes({
       ...options,
       full: options?.full ?? false
@@ -818,7 +820,9 @@ export class Collection<TSchema extends Document = Document> {
     options: IndexInformationOptions & { full: boolean }
   ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
   indexes(): Promise<IndexDescriptionInfo[]>;
-  async indexes(options?: IndexInformationOptions) {
+  async indexes(
+    options?: IndexInformationOptions
+  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]> {
     const indexes: IndexDescriptionInfo[] = await this.listIndexes(options).toArray();
     const full = options?.full ?? true;
     if (full) {
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -52,7 +52,8 @@ import type {
   CreateIndexesOptions,
   DropIndexesOptions,
   IndexDescription,
-  IndexDirection,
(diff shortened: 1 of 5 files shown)


src/operations/indexes.ts
```diff
@@ -214,6 +214,18 @@ function resolveIndexDescription(
...
 
+/**
+ * @public
+ * The index information returned by the listIndexes command. https://www.mongodb.com/docs/manual/reference/command/listIndexes/#mongodb-dbcommand-dbcmd.listIndexes
+ */
+export type IndexDescriptionInfo = Omit<IndexDescription, 'key' | 'version'> & {
+  key: { [key: string]: IndexDirection };
+  v?: IndexDescription['version'];
+};
+
```
baileympearson, line 225: We generally try to keep our type definitions for server responses flexible so that if the server modifies a response in a future server version the driver is still compatible with it.  Could we `& Document` here?  I think that will provide type safety on known keys but will allow unknown keys without a Typescript error.

src/collection.ts
```diff
@@ -693,8 +701,13 @@ export class Collection<TSchema extends Document = Document> {
    *
    * @param options - Optional settings for the command
    */
-  async indexInformation(options?: IndexInformationOptions): Promise<Document> {
-    return await this.indexes({ ...options, full: options?.full ?? false });
+  async indexInformation<TFull extends boolean = false>(
```
baileympearson, line 704: It's preferrable to avoid computed types as the return types from our APIs, since they don't provided users with information about the return type of a function until a user 1) reads the source of the computed type or 2) writes code using the API and checks the computed return type.  Could we accomplish the same here using overloads and removing all `TFull` generics?  Something like:

```typescript
  indexInformation(
    options: IndexInformationOptions & { full: true }
  ): Promise<IndexDescriptionInfo[]>;
  indexInformation(
    options: IndexInformationOptions & { full: false }
  ): Promise<IndexDescriptionCompact>;
  indexInformation(
    options: IndexInformationOptions & { full: boolean }
  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
```

This has the additional benefits of 

- better documentation, because our tooling will generate a doc entry for each overload
- clearly defined overloads with existing IDE tooling
- each overload can be documented originally (if desired).

for example - here's what VSCode could hypothetically display if we used overloads:

Screenshot 2024-04-09 at 12 23 01 PM
Screenshot 2024-04-09 at 12 22 55 PM

src/operations/indexes.ts
```diff
@@ -72,7 +72,7 @@ export type IndexSpecification = OneOrMore<
 >;
 
 /** @public */
-export interface IndexInformationOptions extends ListIndexesOptions {
+export interface IndexInformationOptions<TFull extends boolean> extends ListIndexesOptions {
```
baileympearson, line 75: We could consider this a breaking change, since existing users' code that makes use of `IndexInformationOptions` would break because they wouldn't be providing a generic argument.  We could fix this by providing a default, but I think overloads are preferable (see my comment on `Collection.indexInformation()`).

baileympearson (changes requested):
Hey @prenaissance - thanks for the submission!  I left some comments about the implementation.

prenaissance:
> Hey @prenaissance - thanks for the submission! I left some comments about the implementation.

Thank you for the feedback. I fixed the comments and also added tests for `Db.prototype.indexInformation`

src/collection.ts
```diff
@@ -121,6 +122,13 @@ export interface CollectionPrivate {
   writeConcern?: WriteConcern;
 }
 
+/** @public */
+export type IndexesInformation<TFull extends boolean = boolean> = TFull extends true
+  ? IndexDescriptionInfo[]
+  : TFull extends false
+  ? IndexDescriptionCompact
+  : IndexDescriptionInfo[] | IndexDescriptionCompact;
+
```
baileympearson, line 131:
```suggestion
```

src/collection.ts
```diff
@@ -693,8 +701,21 @@ export class Collection<TSchema extends Document = Document> {
...
+    options: IndexInformationOptions & { full: true }
+  ): Promise<IndexDescriptionInfo[]>;
+  indexInformation(
+    options: IndexInformationOptions & { full?: false }
+  ): Promise<IndexDescriptionCompact>;
+  indexInformation(
+    options: IndexInformationOptions & { full: boolean }
+  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
+  indexInformation(): Promise<IndexDescriptionCompact>;
+  async indexInformation(options?: IndexInformationOptions) {
```
baileympearson, line 714:
```suggestion
  async indexInformation(options?: IndexInformationOptions): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]> {
```

We usually explicitly annotate the return type of all overloads, even when it can be inferred by TS.

src/collection.ts
```diff
@@ -798,19 +819,23 @@ export class Collection<TSchema extends Document = Document> {
...
    */
-  async indexes(options?: IndexInformationOptions): Promise<Document[]> {
-    const indexes = await this.listIndexes(options).toArray();
+  indexes(options: IndexInformationOptions & { full?: true }): Promise<IndexDescriptionInfo[]>;
+  indexes(options: IndexInformationOptions & { full: false }): Promise<IndexDescriptionCompact>;
+  indexes(
+    options: IndexInformationOptions & { full: boolean }
+  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
+  indexes(): Promise<IndexDescriptionInfo[]>;
+  async indexes(options?: IndexInformationOptions) {
```
baileympearson, line 828:
```suggestion
  async indexes(options?: IndexInformationOptions): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]> {
```

src/db.ts
```diff
@@ -482,8 +484,23 @@ export class Db {
...
+  indexInformation(
+    name: string,
+    options: IndexInformationOptions & { full?: false }
+  ): Promise<IndexDescriptionCompact>;
+  indexInformation(
+    name: string,
+    options: IndexInformationOptions & { full: boolean }
+  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
+  indexInformation(name: string): Promise<IndexDescriptionCompact>;
+  async indexInformation(name: string, options?: IndexInformationOptions) {
```
baileympearson, line 500:
```suggestion
  async indexInformation(name: string, options?: IndexInformationOptions): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]> {
```

baileympearson (changes requested):
one last round of small comments, otherwise LGTM

src/collection.ts
baileympearson: (we'll also need to remove the export from the `index.ts`)

src/db.ts
prenaissance: My bad!

src/collection.ts
```diff
@@ -693,8 +694,23 @@ export class Collection<TSchema extends Document = Document> {
...
+  indexInformation(
+    options: IndexInformationOptions & { full: true }
+  ): Promise<IndexDescriptionInfo[]>;
+  indexInformation(
+    options: IndexInformationOptions & { full?: false }
+  ): Promise<IndexDescriptionCompact>;
+  indexInformation(
+    options: IndexInformationOptions & { full: boolean }
+  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
+  indexInformation(): Promise<IndexDescriptionCompact>;
```
baileympearson, line 706: I was testing a bit more and I think we're missing one last overload.  Currently, the following has a type error:

```typescript
async function info(options: IndexInformationOptions) {
  const ret = await collection.indexInformation(options);
}

No overload matches this call.
  Overload 1 of 4, '(options: IndexInformationOptions & { full: true; }): Promise<IndexDescriptionInfo[]>', gave the following error.
    Argument of type 'IndexInformationOptions' is not assignable to parameter of type 'IndexInformationOptions & { full: true; }'.
      Type 'IndexInformationOptions' is not assignable to type '{ full: true; }'.
        Types of property 'full' are incompatible.
          Type 'boolean' is not assignable to type 'true'.
  Overload 2 of 4, '(options: IndexInformationOptions & { full?: false; }): Promise<IndexDescriptionCompact>', gave the following error.
    Argument of type 'IndexInformationOptions' is not assignable to parameter of type 'IndexInformationOptions & { full?: false; }'.
      Type 'IndexInformationOptions' is not assignable to type '{ full?: false; }'.
        Types of property 'full' are incompatible.
          Type 'boolean' is not assignable to type 'false'.
  Overload 3 of 4, '(options: IndexInformationOptions & { full: boolean; }): Promise<IndexDescriptionInfo[] | IndexDescriptionCompact>', gave the following error.
    Argument of type 'IndexInformationOptions' is not assignable to parameter of type 'IndexInformationOptions & { full: boolean; }'.
      Type 'IndexInformationOptions' is not assignable to type '{ full: boolean; }'.
        Property 'full' is optional in type 'IndexInformationOptions' but required in type '{ full: boolean; }'.ts(2769)
```

It seems like the problem is that the overload that matches a generic `IndexInformationOptions` contains a boolean literal (`{ full?: false }`), but because we don't have a const boolean, no overload is matched.

If we removed the `IndexInformationOptions & { full?: boolean }` overload and instead provided an overload:
```
indexInformation(
    options: IndexInformationOptions
  ): Promise<IndexDescriptionCompact | IndexDescriptionInfo[]>;
```

The issue is resolved.   This has the downside of losing the overload which provides a default return type when and empty options object is provided.

prenaissance: I figured why this happens. This type `IndexInformationOptions & { full: boolean }` has the `full` property required and that makes `IndexInformationOptions` unassignable. Nice catch, I'll add a test case for this.

durran (approved):
I just updated the description from `Document.prototype` call outs to `Collection.prototype`

PR #4075, opened 2024-04-11
fix(NODE-5925): driver throws error when non-read operation in a transaction has a ReadPreferenceMode other than 'primary'

aditi-khare-mongoDB:
### Description
Within a transaction, readPreferences != 'primary' are not supported, and our error logic must support that to comply with spec. An error should only be thrown in these circumstances when a read operation is being executed. 

#### What is changing?
In a transaction, when a user provides a readPreference != 'primary' and then tries to perform a command that involves a read, the Node Driver will now throw a 

Yes, update [Transactions Node.js Manual](https://www.mongodb.com/docs/drivers/node/current/fundamentals/transactions/) to reflect desired readPreference inputs.

See Release Highlights

### Don't throw error when non-read operation in a transaction has a `ReadPreferenceMode` other than `'primary'`
The following error will now only be thrown  when a user provides a `ReadPreferenceMode` other than `primary` and then tries to perform a command that involves a read:
```javascript
new MongoTransactionError('Read preference in a transaction must be primary');
```
Prior to this change, the Node Driver would incorrectly throw this error even when the operation does not perform a read. 
Note: a `RunCommandOperation` is treated as a read operation for this error.

### DIFF
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -122,7 +122,7 @@ export async function executeOperation<
   if (
     inTransaction &&
     !readPreference.equals(ReadPreference.primary) &&
-    operation.hasAspect(Aspect.READ_OPERATION)
+    (operation.hasAspect(Aspect.READ_OPERATION) || operation.commandName === 'runCommand')
   ) {
     throw new MongoTransactionError(
       `Read preference in a transaction must be primary, not: ${readPreference.mode}`
--- a/src/operations/execute_operation.ts
+++ b/src/operations/execute_operation.ts
@@ -119,7 +119,11 @@ export async function executeOperation<
   const readPreference = operation.readPreference ?? ReadPreference.primary;
   const inTransaction = !!session?.inTransaction();
 
-  if (inTransaction && !readPreference.equals(ReadPreference.primary)) {
+  if (
+    inTransaction &&
+    !readPreference.equals(ReadPreference.primary) &&
+    operation.hasAspect(Aspect.READ_OPERATION)
+  ) {
     throw new MongoTransactionError(
       `Read preference in a transaction must be primary, not: ${readPreference.mode}`
     );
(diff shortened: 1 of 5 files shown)


src/sessions.ts
```diff
@@ -194,7 +195,7 @@ export class ClientSession extends TypedEventEmitter<ClientSessionEvents> {
 
     this.operationTime = undefined;
     this.owner = options.owner;
-    this.defaultTransactionOptions = Object.assign({}, options.defaultTransactionOptions);
```
aditi-khare-mongoDB, line 197: This isn't directly related to the code changes, but is nice to have and helps us have less `Object.assign` calls in our code

test/tools/unified-spec-runner/entities.ts
```diff
@@ -620,9 +620,8 @@ export class EntitiesMap<E = Entity> extends Map<string, E> {
               WriteConcern.fromOptions(defaultOptions);
           }
           if (defaultOptions.readPreference) {
-            options.defaultTransactionOptions.readPreference = ReadPreference.fromOptions(
-              defaultOptions.readPreference
-            );
+            options.defaultTransactionOptions.readPreference =
+              ReadPreference.fromOptions(defaultOptions);
```
aditi-khare-mongoDB, line 624: This isn't in my PR description, but we were originally parsing `defaultTransactionOptions` incorrectly in our entity creation in the UTR, and it needed to be fixed for the tests to pass.

src/operations/execute_operation.ts
```diff
@@ -119,7 +119,11 @@ export async function executeOperation<
   const readPreference = operation.readPreference ?? ReadPreference.primary;
   const inTransaction = !!session?.inTransaction();
 
-  if (inTransaction && !readPreference.equals(ReadPreference.primary)) {
+  if (
+    inTransaction &&
+    !readPreference.equals(ReadPreference.primary) &&
+    (operation.hasAspect(Aspect.READ_OPERATION) || operation.commandName === 'runCommand')
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Yep, I've moved the isReadOperation up in the function!

nbbeeken (approved):

PR #4078, opened 2024-04-15
introduce timeout abstraction

W-A-James:


src/timeout.ts
```diff
@@ -0,0 +1,129 @@
+import { clearTimeout, setTimeout } from 'timers';
+
+import { MongoError, MongoInvalidArgumentError } from './error';
+import { noop } from './utils';
+
+/** @internal */
+export class CSOTError extends MongoError {
```
nbbeeken, line 7:
```suggestion
export class MongoDBTimeoutError extends MongoError {
```

We're in this tight spot with naming 😅 every timeout we fire is "client side" but since CSOT refers to the `timeoutMS` feature we can change this name and this error is going to serve as a communication of what kind of timeout we fired.

nbbeeken (changes requested):

src/timeout.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /**
+   * Implement maxTimeMS calculation detailed in https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#command-execution
+   * */
+  public getMaxTimeMS(minRoundTripTime: number): any {
+    if (!Number.isFinite(this.remainingTime)) return 0;
+    if (minRoundTripTime < this.remainingTime) return this.remainingTime - minRoundTripTime;
+    throw CSOTError.from(this.timeoutError);
+  }
```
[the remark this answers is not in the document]
nbbeeken: Go ahead and clean up any and all unused code, we should introduce it in the PR that uses and tests it

test/unit/timeout.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    context('when called with a duration greater than 0', function () {
+      beforeEach(() => {
+        timeout = Timeout.expires(2000);
+      });
+      it('creates a timeout instance that will not keep the Node.js event loop active', function () {
+        expect(timeout).to.have.property('id');
+        const id = timeout['id'];
+        expect(id?.hasRef()).to.be.false;
+      });
+      it('throws a CSOTError when it expires', async function () {
```
nbbeeken, line 33: Not sure why but our: `"mocha/consistent-spacing-between-blocks": "error",` rule is supposed to be preventing this 🤔

test/unit/timeout.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    });
+
+    context('when called with a duration greater than 0', function () {
+      beforeEach(() => {
+        timeout = Timeout.expires(2000);
+      });
+      it('creates a timeout instance that will not keep the Node.js event loop active', function () {
+        expect(timeout).to.have.property('id');
+        const id = timeout['id'];
+        expect(id?.hasRef()).to.be.false;
```
nbbeeken, line 31:
```suggestion
        // @ts-expect-error: accessing private property
        const id = timeout.id;
```

Rather than avoid the TS error, I'd rather acknoledge it and clean up the code if it were no longer disallowed.

test/unit/timeout.test.ts
```diff
@@ -0,0 +1,89 @@
...
+describe('Timeout', function () {
+  let timeout: Timeout;
+
+  beforeEach(() => {
+    if (Timeout.is(timeout)) {
+      timeout.clear();
+    }
+  });
+
+  describe('constructor()', function () {
```
nbbeeken, line 16:
```suggestion
  describe('expires()', function () {
```

test/unit/timeout.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import { expect } from 'chai';
+
+import { CSOTError, MongoInvalidArgumentError, Timeout } from '../mongodb';
+
+describe('Timeout', function () {
+  let timeout: Timeout;
+
+  beforeEach(() => {
+    if (Timeout.is(timeout)) {
```
nbbeeken, line 11: Can we test the `is` method?

src/timeout.ts
```diff
@@ -0,0 +1,129 @@
...
+    super((_, promiseReject) => {
+      reject = promiseReject;
+
+      executor(noop, promiseReject);
+    });
+
+    // NOTE: Construct timeout error at point of Timeout instantiation to preserve stack traces
+    // TODO(NODE-5679): Come up with better default message for CSOT error
+    this.timeoutError = new CSOTError('Timeout!');
+
```
nbbeeken, line 75: Along with the rename I suppose we can update this to something generic "expired after X ms" or something since this error will always be converted to the domain specific one.

src/timeout.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.duration = duration;
+    this.start = Math.trunc(performance.now());
+
+    if (this.duration > 0) {
+      this.id = setTimeout(() => {
+        this.ended = Math.trunc(performance.now());
+        this.timedOut = true;
+        reject(this.timeoutError);
+      }, this.duration);
+      // Ensure we do not keep the NodeJS event loop running
```
nbbeeken, line 85:
```suggestion
      // Ensure we do not keep the Node.js event loop running
```

nit (I think I wrote it this way, sorry 😅)

src/timeout.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (typeof this.id.unref === 'function') {
+        this.id.unref();
+      }
+    }
+  }
+
+  /**
+   * Clears the underlying timeout. This method is idempotent
+   * */
+  public clear(): void {
```
nbbeeken, line 95:
```suggestion
   */
```

src/timeout.ts
```diff
@@ -0,0 +1,129 @@
...
+   * Implement maxTimeMS calculation detailed in https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#command-execution
+   * */
+  public getMaxTimeMS(minRoundTripTime: number): any {
+    if (!Number.isFinite(this.remainingTime)) return 0;
+    if (minRoundTripTime < this.remainingTime) return this.remainingTime - minRoundTripTime;
+    throw CSOTError.from(this.timeoutError);
+  }
+
+  /** Create a new pending Timeout with the same duration */
+  public clone() {
```
nbbeeken, line 110: Do we use this? if not we can remove it

src/timeout.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  static from(error: CSOTError) {
+    return new CSOTError(error.message, { cause: error });
+  }
+}
+
+type Executor = ConstructorParameters<typeof Promise<never>>[0];
+type Reject = Parameters<ConstructorParameters<typeof Promise<never>>[0]>[1];
+/** @internal
```
nbbeeken, line 29: `@internal` needs to be on its own line, and this isn't CSOT specific anymore, we can update the docs.

nbbeeken (changes requested):

src/cmap/connection_pool.ts
```diff
@@ -149,6 +144,7 @@ export type ConnectionPoolEvents = {
 export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
   options: Readonly<ConnectionPoolOptions>;
   [kPoolState]: (typeof PoolState)[keyof typeof PoolState];
+  /** @internal */
```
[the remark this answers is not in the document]
W-A-James: Oh my bad, I had converted it to a non Symbol property but rolled back that change to keep the PR minimal

nbbeeken (approved):

src/timeout.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    super((_, promiseReject) => {
+      reject = promiseReject;
+
+      executor(noop, promiseReject);
+    });
+
+    // NOTE: Construct timeout error at point of Timeout instantiation to preserve stack traces
+    this.timeoutError = new TimeoutError(`Expired after ${duration}ms`);
```
baileympearson, line 59: How does this work with CSOT, when a single timeout will be reused for multiple operations?

src/timeout.ts
```diff
@@ -0,0 +1,100 @@
+import { clearTimeout, setTimeout } from 'timers';
+
+import { MongoInvalidArgumentError } from './error';
+import { noop } from './utils';
+
+/** @internal */
+export class TimeoutError extends Error {
```
baileympearson, line 7: The only place this is thrown is in the `Timeout` class, and we immediately catch it and throw a different error ( this error is internal ).  

Why use an error at all?  Could we just throw a string or something else?

src/timeout.ts
baileympearson: looked more closely at this class - if this is internal, I don't see why capturing it in the constructor for the stack trace is important

src/timeout.ts
W-A-James: Good point. I possibly misunderstood the [spec](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#errors) here. I confused the error we throw to signify a timeout has occurred in this class with the error that we'll have to wrap the server selection timeout/waitqueue timeout error with.

I'm open to just throwing a string here. @nbbeeken thoughts?

src/timeout.ts
W-A-James: In keeping with my response to your other question, I'm open to just throwing a string.

src/timeout.ts
nbbeeken: We can just throw anything since we should/need to always convert the error. However, I do think there's value in keeping it as an error since it will make a stack trace available if it is needed for debugging. If we use a string or symbol etc, and we accidentally miss converting it somehow there is no way to trace it back to where it was created. 

Also, we _could_ set this expiry error as the cause of the error that we convert to, not the most useful but it would identify where the timeout was created (ex. in selectServer or at a higher layer for CSOT) again just for debugging not really a user-facing feature.

W-A-James: I agree that for future debugging purposes, it would be useful to keep this as an error.

src/timeout.ts
W-A-James: Also as far as the message with the duration string is concerned, part of the use case for CSOT would be, for each of the blocking sections, we construct a new Timeout from the Timeout constructed for the whole operation that either uses the remaining timeoutMS or another timeout value as detailed in the spec. For example, when doing server selection, the [spec](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#server-selection) requires that we timeout that operation with the remaining timeoutMS or serverSelectionTimeoutMS. Which would look something like

```ts
const timeout = options.timeout
					? Timeout.expires(min(options.timeout.remainingTime, options.serverSelectionTimeoutMS))
					: Timeout.expires(options.serverSelectionTimeoutMS)
```

baileympearson: Okay, that works.  Early CSOT designs re-used timeouts (or timeout controllers) and just reset the internal timeout instead of creating a new timeout each time, which would make the stack trace incorrect.  If we're creating a new timeout for every async part of the driver, that's not an issue.

PR #4079, opened 2024-04-16
feat(NODE-3639): add a general stage to the aggregation pipeline builder

prenaissance:
#### What is changing?

add a general stage method to the aggregation pipeline builder

This feature would allow arbitrary aggregation stages to be used builder style with the aggregation cursor, simplifying the syntax needed to use the stages otherwise. Currently only the most common stages have explicit APIs available for chaining and adding an explicit API for every possible aggregation stage would not be maintainable.

This change will also help making progress in other aggregation pipeline typification features, as the aggregation pipeline builder can be used for both stages which have explicit builder helpers and those which do not.

### Aggregation pipelines can now add stages manually.

When creating an aggregation pipeline cursor, a new generic method `addStage()` has been added in the fluid API for users to add stages in a general manner.

```ts
const documents = await users.aggregate().addStage({ $project: { name: true } }).toArray();
```

Thank you @prenaissance for contributing this feature!

### DIFF
--- a/src/cursor/aggregation_cursor.ts
+++ b/src/cursor/aggregation_cursor.ts
@@ -86,33 +86,45 @@ export class AggregationCursor<TSchema = any> extends AbstractCursor<TSchema> {
     );
   }
 
+  /** Add a stage to the aggregation pipeline
+   * @example
+   * ```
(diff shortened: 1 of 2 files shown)


durran (approved):

nbbeeken (approved):
Thanks so much for this contribution!

PR #4081, opened 2024-04-17
feat(NODE-6094) RFC: Allow the Mongo-AWS creds provider to handle assuming an AWS IAM role.

jwhitaker-gridcog:
#### What is changing?
The MONGODB-AWS credential provider will now check if you pass a var AWS_ROLE_ARN as part of your auth mechanism properties. If you do, it will assume that role for you using the AWS SDK's credential provider, which will handle session expiry and re-authentication.

Yes.

In our org, we're using the mongo creds provider with an assumed role. We are assuming the role ourselves and passing the resulting key, secret, and session token to the mongo connector. This works fine...

... until the STS session expires.

At this point we need to implement full expiry and reconnection logic ourselves. This sucks - it's finicky, and the code to handle it already exists in both this repository and in the aws-sdk, probably more reliably than we'll ever manage.
It's also annoying because it has to exist outside of the connection pool, and must kill the whole pool. This is the sort of thing a pool should manage for you.

Because of this, we'd really like if the Mongo connector could handle assuming a role for the user. That way we don't need to do all this invalidation + reconnection stuff ourselves, and it lives inside the pool where it belongs. As a selling point to you guys, it actually lives in the aws sdks, not mongo. There is no reconnection logic added as part of this PR as none is needed with this approach.

I haven't set up automated tests with this approach yet, I'd ask for a go/no-go before going to that effort :)

Thanks for your consideration!

### Code notes

Most of the changes are just to pass through the authmechanismproperties to where they are needed. This is the first commit.

You'll see the second commit, actually doing the role assumption, is quite small.

### DIFF
--- a/src/client-side-encryption/auto_encrypter.ts
+++ b/src/client-side-encryption/auto_encrypter.ts
@@ -502,7 +502,8 @@ export class AutoEncrypter {
    * the original ones.
    */
   async askForKMSCredentials(): Promise<KMSProviders> {
-    return await refreshKMSCredentials(this._kmsProviders);
+    const creds = this._client.options.credentials?.mechanismProperties || {};
+    return await refreshKMSCredentials(this._kmsProviders, creds);
   }
 
   /**
--- a/src/client-side-encryption/client_encryption.ts
+++ b/src/client-side-encryption/client_encryption.ts
@@ -654,7 +654,8 @@ export class ClientEncryption {
    * the original ones.
    */
   async askForKMSCredentials(): Promise<KMSProviders> {
-    return await refreshKMSCredentials(this._kmsProviders);
+    const authProps = this._client.options.credentials?.mechanismProperties || {};
+    return await refreshKMSCredentials(this._kmsProviders, authProps);
   }
 
   static get libmongocryptVersion() {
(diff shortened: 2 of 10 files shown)


durran:
Hi @jwhitaker-gridcog thanks for the PR. We'll discuss and then update you here.

aditi-khare-mongoDB:
Hi [@jwhitaker-gridcog](https://github.com/jwhitaker-gridcog), thanks again for submitting this PR!

We will be closing this PR. Since `AuthMechanismProperties` type is part of the drivers specification, it can't be changed without a specifications change ([see here](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#gssapi)). I've filed a follow-up ticket [NODE-6141](https://jira.mongodb.org/browse/NODE-6141) to handle this issue in a spec compliant way and the specifications owners have been notified of this issue, as well, to see if any what changes should be made drivers-wide.

aditi-khare-mongoDB:
Hi [@jwhitaker-gridcog](https://github.com/jwhitaker-gridcog), thanks again for submitting this PR!

We will be closing this PR. Since AuthMechanismProperties type is part of the drivers specification, it can't be changed without a specifications change ([see here](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#gssapi)). I've filed a follow-up ticket [NODE-6141](https://jira.mongodb.org/browse/NODE-6141) to handle this issue in a spec compliant way and the specifications owners have been notified of this issue, as well, to see if any what changes should be made drivers-wide.

PR #4083, opened 2024-04-17
chore(NODE-6095): upgrade dev dependencies in driver

baileympearson:
#### What is changing?

Quarterly dev dependency bump.  `chai`, `chalk` were not upgraded because they only support esmodules in the latest versions.  `eslint` and other eslint plugins were left alone because upgrading caused lots of peer dependency issues - it seems like not all eslint plugin packages have adjusted their peer dependency range to support eslint@9 yet.


test/unit/client-side-encryption/state_machine.test.ts
```diff
@@ -83,6 +83,7 @@ describe('StateMachine', function () {
   });
 
   describe('kmsRequest', function () {
+    let sandbox: sinon.SinonSandbox;
```
baileympearson, line 86: I refactored `this.sinon.stub` -> `sandbox.stub` while debugging, and although that didn't fix the tests, I left it in because we generally use scoped variables instead of binding variables to `this` of the executing test.

W-A-James (approved):

aditi-khare-mongoDB (approved):

PR #4085, opened 2024-04-17
refactor(NODE-6057): implement CursorResponse for lazy document parsing

nbbeeken:
#### What is changing?

- Implement a `CursorResponse` subclass of MongoDBResponse
  - Implements logic to abstract List vs CursorResponse handling
- Make FindCursors use CursorResponse

TBD. Is this something we want to apply to all cursors? Should we progressively migrate to this?

An attempt at increasing performance. A sucessful decoupling of driver BSON and user BSON settings.

Only for FindCursor, BSON documents from the current batch are now parsed on each iteration. 

One clear improvement is that any cursor that is closed before exhaustion will take significantly less run time.

### DIFF
--- a/src/client-side-encryption/auto_encrypter.ts
+++ b/src/client-side-encryption/auto_encrypter.ts
@@ -10,7 +10,7 @@ import { MongoDBResponse, type MongoDBResponseConstructor } from '../cmap/wire_p
 import { getMongoDBClientEncryption } from '../deps';
 import { MongoRuntimeError } from '../error';
 import { MongoClient, type MongoClientOptions } from '../mongo_client';
-import { MongoDBCollectionNamespace } from '../utils';
+import { isUint8Array, MongoDBCollectionNamespace } from '../utils';
 import * as cryptoCallbacks from './crypto_callbacks';
 import { MongoCryptInvalidArgumentError } from './errors';
 import { MongocryptdManager } from './mongocryptd_manager';
@@ -480,7 +480,7 @@ export class AutoEncrypter {
   ): Promise<Document> {
     const buffer = MongoDBResponse.is(response)
       ? response.toBytes()
-      : Buffer.isBuffer(response)
+      : isUint8Array(response)
       ? response
       : serialize(response, options);
 
--- a/src/client-side-encryption/auto_encrypter.ts
+++ b/src/client-side-encryption/auto_encrypter.ts
@@ -6,11 +6,10 @@ import {
 
 import { deserialize, type Document, serialize } from '../bson';
 import { type CommandOptions, type ProxyOptions } from '../cmap/connection';
-import { MongoDBResponse, type MongoDBResponseConstructor } from '../cmap/wire_protocol/responses';
 import { getMongoDBClientEncryption } from '../deps';
 import { MongoRuntimeError } from '../error';
 import { MongoClient, type MongoClientOptions } from '../mongo_client';
-import { isUint8Array, MongoDBCollectionNamespace } from '../utils';
+import { MongoDBCollectionNamespace } from '../utils';
 import * as cryptoCallbacks from './crypto_callbacks';
 import { MongoCryptInvalidArgumentError } from './errors';
 import { MongocryptdManager } from './mongocryptd_manager';
@@ -474,19 +473,8 @@ export class AutoEncrypter {
   /**
    * Decrypt a command response
    */
-  async decrypt(
-    response: Uint8Array | Document | MongoDBResponse,
-    options: CommandOptions = {}
-  ): Promise<Document> {
-    const buffer = MongoDBResponse.is(response)
-      ? response.toBytes()
-      : isUint8Array(response)
-      ? response
-      : serialize(response, options);
-
-    const responseType = MongoDBResponse.is(response)
-      ? (response.constructor as MongoDBResponseConstructor)
-      : undefined;
+  async decrypt(response: Uint8Array | Document, options: CommandOptions = {}): Promise<Document> {
+    const buffer = Buffer.isBuffer(response) ? response : serialize(response, options);
 
     const context = this._mongocrypt.makeDecryptionContext(buffer);
 
@@ -499,11 +487,10 @@ export class AutoEncrypter {
     });
 
     const decorateResult = this[kDecorateResult];
-    const result = await stateMachine.execute(this, context, responseType);
+    const result = await stateMachine.execute<Document>(this, context);
     if (decorateResult) {
       decorateDecryptionResult(result, response);
     }
-
     return result;
   }
 
--- a/src/client-side-encryption/auto_encrypter.ts
+++ b/src/client-side-encryption/auto_encrypter.ts
@@ -6,6 +6,7 @@ import {
 
 import { deserialize, type Document, serialize } from '../bson';
 import { type CommandOptions, type ProxyOptions } from '../cmap/connection';
+import { MongoDBResponse } from '../cmap/wire_protocol/responses';
 import { getMongoDBClientEncryption } from '../deps';
 import { MongoRuntimeError } from '../error';
 import { MongoClient, type MongoClientOptions } from '../mongo_client';
@@ -473,8 +474,15 @@ export class AutoEncrypter {
   /**
    * Decrypt a command response
    */
-  async decrypt(response: Uint8Array | Document, options: CommandOptions = {}): Promise<Document> {
-    const buffer = Buffer.isBuffer(response) ? response : serialize(response, options);
+  async decrypt(
+    response: Uint8Array | Document | MongoDBResponse,
+    options: CommandOptions = {}
+  ): Promise<Document> {
+    const buffer = MongoDBResponse.is(response)
+      ? response.toBytes()
+      : Buffer.isBuffer(response)
+      ? response
+      : serialize(response, options);
 
     const context = this._mongocrypt.makeDecryptionContext(buffer);
 
@@ -487,10 +495,11 @@ export class AutoEncrypter {
     });
 
     const decorateResult = this[kDecorateResult];
-    const result = await stateMachine.execute<Document>(this, context);
+    const result = await stateMachine.execute(this, context, response.constructor);
     if (decorateResult) {
       decorateDecryptionResult(result, response);
     }
+
     return result;
   }
 
--- a/src/client-side-encryption/auto_encrypter.ts
+++ b/src/client-side-encryption/auto_encrypter.ts
@@ -6,7 +6,7 @@ import {
 
 import { deserialize, type Document, serialize } from '../bson';
 import { type CommandOptions, type ProxyOptions } from '../cmap/connection';
-import { MongoDBResponse } from '../cmap/wire_protocol/responses';
+import { MongoDBResponse, type MongoDBResponseConstructor } from '../cmap/wire_protocol/responses';
 import { getMongoDBClientEncryption } from '../deps';
 import { MongoRuntimeError } from '../error';
 import { MongoClient, type MongoClientOptions } from '../mongo_client';
@@ -484,6 +484,10 @@ export class AutoEncrypter {
       ? response
       : serialize(response, options);
 
+    const responseType = MongoDBResponse.is(response)
+      ? (response.constructor as MongoDBResponseConstructor)
+      : undefined;
+
     const context = this._mongocrypt.makeDecryptionContext(buffer);
 
     context.id = this._contextCounter++;
@@ -495,7 +499,7 @@ export class AutoEncrypter {
     });
 
     const decorateResult = this[kDecorateResult];
-    const result = await stateMachine.execute(this, context, response.constructor);
+    const result = await stateMachine.execute(this, context, responseType);
     if (decorateResult) {
       decorateDecryptionResult(result, response);
     }
--- a/src/client-side-encryption/state_machine.ts
+++ b/src/client-side-encryption/state_machine.ts
@@ -153,13 +153,29 @@ export class StateMachine {
   ) {}
 
   /**
-   * Executes the state machine according to the specification
+   * Executes the state machine according to the specification.
+   * Will construct the result using `responseType`.
+   */
+  async execute<R extends MongoDBResponseConstructor>(
+    executor: StateMachineExecutable,
+    context: MongoCryptContext,
+    responseType?: R
+  ): Promise<InstanceType<R>>;
+
+  /**
+   * Executes the state machine according to the specification.
+   * Will return a document from the default BSON deserializer.
    */
   async execute<T extends Document>(
     executor: StateMachineExecutable,
+    context: MongoCryptContext
+  ): Promise<T>;
+
+  async execute<T extends Document, R extends MongoDBResponseConstructor>(
+    executor: StateMachineExecutable,
     context: MongoCryptContext,
-    responseType?: MongoDBResponseConstructor
-  ): Promise<T> {
+    responseType?: R
+  ): Promise<T | InstanceType<R>> {
     const keyVaultNamespace = executor._keyVaultNamespace;
     const keyVaultClient = executor._keyVaultClient;
     const metaDataClient = executor._metaDataClient;
--- a/src/client-side-encryption/state_machine.ts
+++ b/src/client-side-encryption/state_machine.ts
@@ -11,6 +11,7 @@ import {
   serialize
 } from '../bson';
 import { type ProxyOptions } from '../cmap/connection';
+import { type MongoDBResponseConstructor } from '../cmap/wire_protocol/responses';
 import { getSocks, type SocksLib } from '../deps';
 import { type MongoClient, type MongoClientOptions } from '../mongo_client';
 import { BufferPool, MongoDBCollectionNamespace, promiseWithResolvers } from '../utils';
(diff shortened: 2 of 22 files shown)


src/client-side-encryption/auto_encrypter.ts
```diff
@@ -473,8 +474,19 @@ export class AutoEncrypter {
   /**
    * Decrypt a command response
    */
-  async decrypt(response: Uint8Array | Document, options: CommandOptions = {}): Promise<Document> {
-    const buffer = Buffer.isBuffer(response) ? response : serialize(response, options);
+  async decrypt(
```
nbbeeken, line 477: I can make FLE work or we can exempt it like we have to for explain

src/cursor/abstract_cursor.ts
```diff
@@ -382,14 +389,7 @@ export abstract class AbstractCursor<
...
 
-    const doc = await next<TSchema>(this, { blocking: true, transform: false });
-
-    if (doc) {
-      this[kDocuments].unshift(doc);
-      return true;
-    }
-
-    return false;
+    return await next(this, { blocking: true, transform: false, hasNext: true });
```
nbbeeken, line 392: If you look through my commits I originally supported unshift so that hasNext worked without changes but it was more logic than it was worth to hold on to the unshifted document. I think this is a good improvement regardless of CursorResponse, I wouldn't mind merging it separately

test/unit/assorted/collations.test.js
```diff
@@ -55,7 +55,7 @@ describe('Collation', function () {
         request.reply(primary[0]);
       } else if (doc.aggregate) {
         commandResult = doc;
-        request.reply({ ok: 1, cursor: { id: 0, firstBatch: [], ns: 'collation_test' } });
+        request.reply({ ok: 1, cursor: { id: 0n, firstBatch: [], ns: 'collation_test' } });
```
nbbeeken, line 58: Should we use getFlexibleNumber? a real cursor id will ALWAYS be an int64

src/cmap/wire_protocol/responses.ts
```diff
@@ -83,27 +95,98 @@ export class MongoDBResponse extends OnDemandDocument {
...
+  public batchSize = 0;
+
+  private batch: OnDemandDocument | null = null;
+  private values: Generator<OnDemandDocument, void, void> | null = null;
+  private iterated = 0;
+
+  constructor(b: Uint8Array, o?: number, a?: boolean) {
+    super(b, o, a);
+
+    if (this.isError) return;
```
nbbeeken, line 147: I don't like this. I would like to potentially write a helper that figures out before calling the responseType constructor if it is an error and then never calls the passed in responseType since we will always convert an ok:0 to a MongoServerError.

src/cmap/wire_protocol/responses.ts
```diff
@@ -83,27 +95,98 @@ export class MongoDBResponse extends OnDemandDocument {
...
+
+    const id = cursor.get('id', BSONType.long, true);
+    this.id = new Long(Number(id & 0xffff_ffffn), Number((id >> 32n) & 0xffff_ffffn));
+
+    const namespace = cursor.get('ns', BSONType.string) ?? '';
+    if (namespace) this.ns = ns(namespace);
+
+    if (cursor.has('firstBatch')) this.batch = cursor.get('firstBatch', BSONType.array, true);
+    else if (cursor.has('nextBatch')) this.batch = cursor.get('nextBatch', BSONType.array, true);
+    else throw new MongoUnexpectedServerResponseError('Cursor document did not contain a batch');
```
nbbeeken, line 159: This and the other bson `get` calls with `required: true` have the chance of reaching users IFF the server replies with a malformed response. I am still comfortable with this because previously we would have failed somewhere in cursor code if `response.cursor... etc` was undefined with a TypeError, now we would pin point the problem at the layer of BSON missing something we expected to always be there.

src/cursor/abstract_cursor.ts
```diff
@@ -144,7 +145,13 @@ export abstract class AbstractCursor<
   /** @internal */
   [kNamespace]: MongoDBNamespace;
   /** @internal */
-  [kDocuments]: List<TSchema>;
+  [kDocuments]: {
+    length: number;
+    shift(bsonOptions?: any): TSchema | null;
```
nbbeeken, line 150: List.shift will get bson options passed into it that it will make no use of. We won't accidentally introduce new list API usage that isn't supported by CursorResponse

src/cmap/wire_protocol/responses.ts
```diff
@@ -83,27 +95,98 @@ export class MongoDBResponse extends OnDemandDocument {
...
+  public override toObject(options?: BSONSerializeOptions): Record<string, any> {
     const exactBSONOptions = {
-      useBigInt64: options.useBigInt64,
-      promoteLongs: options.promoteLongs,
-      promoteValues: options.promoteValues,
-      promoteBuffers: options.promoteBuffers,
-      bsonRegExp: options.bsonRegExp,
-      raw: options.raw ?? false,
-      fieldsAsRaw: options.fieldsAsRaw ?? {},
+      useBigInt64: options?.useBigInt64,
```
nbbeeken, line 100: I ran the de-opt explorer over this, and since `{}` empty object was often provided then none of these keys exist.

src/client-side-encryption/auto_encrypter.ts
baileympearson: I don't feel strongly.  how much effort do you think it would be to implement for FLE as well?

nbbeeken: Sorry this wasn't clear, it is working, the changes here make it work but I am offering to defer impacting FLE since originally we _only_ wanted to change the find cursor, I don't mind or feel strongly either way.

src/cursor/abstract_cursor.ts
nbbeeken: Here's the commit where I removed it: [`329ca55` (#4085)](https://github.com/mongodb/node-mongodb-native/pull/4085/commits/329ca55bcd58d5c8c798cf25936cf2c18fd71a30)

I have to buffer the unshifted documents to be returned by shift at a later point, in my view there is complexity & risk to maintaining this look-aside list that must be iterated first to ensure we don't repeat parsing effort and maintain the correct order, better to just not pull things out of a list.

Happy to change the name and also move this change out of this PR to keep it isolated. `checkNext`? `iterate: false`? `dequeue: false`?

src/client-side-encryption/state_machine.ts
```diff
@@ -156,14 +157,15 @@ export class StateMachine {
...
+    context: MongoCryptContext,
+    responseType?: MongoDBResponseConstructor
   ): Promise<T> {
     const keyVaultNamespace = executor._keyVaultNamespace;
     const keyVaultClient = executor._keyVaultClient;
     const metaDataClient = executor._metaDataClient;
     const mongocryptdClient = executor._mongocryptdClient;
     const mongocryptdManager = executor._mongocryptdManager;
-    let result: T | null = null;
+    let result: any | null = null;
```
addaleax, line 168: I mean, TypeScript would be kind of right to complain here? The return type doesn't necessarily `extends Document` anymore

I think for "proper" typing you'd need to have overloads that account for the presence or absence of `responseType`:

```ts
    async execute<R extends typeof MongoDBResponseConstructor & {new(): R['prototype']}>(..., responseType: R): Promise<R['prototype']>;
    async execute<T extends Document>(...): Promise<T>;
    async execute<T extends Document, R extends typeof MongoDBResponseConstructor & {new(): R['prototype']}>(..., responseType?: R): Promise<T | R['prototype']> {
```

([TS playground](https://www.typescriptlang.org/play?ssl=14&ssc=158&pln=11&pc=12#code/JYOwLgpgTgZghgYwgAgCIHsEFcC2FzIDeAUMmcgNoDWEAngFzIDOYUoA5gLqMtsjsBuYgF9ixUJFiIUAFXQAHADIQAbhAA2GbHgIQAHpBAATJmky58YIqXIB9YEZ6sOQ0cQTq4TUwFl0-dFQAIQAlCCZ5fyYIAGEo1iwEMHQoIjcPL1MY9VMScmQvWhAEZH0IbEgAHhDSg3wTZDBaeQh0GGQ-AOCwiKjY+KhE5NSAMiIQCAB3AAoASkYQigByeSh0ZKaWpc5hAD5pqHDIkGiZZogF+eQABTWcYGjq5dX19fPt3aF8wuLa8qwqjJaoYGloLOB9ldbuh7o8ZJ8bGQfiUyhUIJUgfoQaYwTowAAaZA1LH1UybVrtTrsQKhI59OInBJJFLIMaECYzK6LFZrDbvHb7Q69E4QM4tAD8l0Y0Nh6KBAB8ic9eW8tpxdtZ8uRgO0DnSRWKILNkIcwFgoCBkByTfrTuc5l8tTazRaiDB0OhGEsAEZwKBLYSOsiidLxZCgFhwYooAC8VqmyGyTAd7jD7vQyDjEbAUaQADpUQCINMqTSesdogzeEMUrMhAgw77UlnGbmIAW9P9ICmG4zkL6AF6Z8Ot6MdrtyhTKNSacx4yFCIA), wanted to check if this actually works 🙂 )

src/cmap/wire_protocol/responses.ts
```diff
@@ -83,27 +95,98 @@ export class MongoDBResponse extends OnDemandDocument {
...
-
     return { utf8: { writeErrors: false } };
   }
 }
+
+/** @internal */
+export class CursorResponse extends MongoDBResponse {
+  static emptyGetMore = new CursorResponse(
+    Buffer.from(
+      'NgAAABBvawABAAAAA2N1cnNvcgAhAAAAEmlkAAAAAAAAAAAABG5leHRCYXRjaAAFAAAAAAAA',
```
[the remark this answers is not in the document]
addaleax: Yeah, if not this, then leaving a comment with instructions for how to create this string might be a good idea :)

src/cursor/abstract_cursor.ts
nbbeeken: WIP https://github.com/mongodb/node-mongodb-native/pull/4093

src/cmap/wire_protocol/responses.ts
```diff
@@ -83,27 +95,98 @@ export class MongoDBResponse extends OnDemandDocument {
...
+
+    if (cursor.has('firstBatch')) this.batch = cursor.get('firstBatch', BSONType.array, true);
+    else if (cursor.has('nextBatch')) this.batch = cursor.get('nextBatch', BSONType.array, true);
+    else throw new MongoUnexpectedServerResponseError('Cursor document did not contain a batch');
+
+    this.batchSize = this.batch.size();
+  }
+
+  get length() {
+    return Math.max(this.batchSize - this.iterated, 0);
```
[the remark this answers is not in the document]
nbbeeken: batchSize is the size of this batch, not the requested amount

src/cmap/wire_protocol/responses.ts
```diff
@@ -83,27 +95,98 @@ export class MongoDBResponse extends OnDemandDocument {
...
+
+  constructor(b: Uint8Array, o?: number, a?: boolean) {
+    super(b, o, a);
+
+    if (this.isError) return;
+
+    const cursor = this.get('cursor', BSONType.object, true);
+
+    const id = cursor.get('id', BSONType.long, true);
+    this.id = new Long(Number(id & 0xffff_ffffn), Number((id >> 32n) & 0xffff_ffffn));
```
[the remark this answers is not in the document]
nbbeeken: Creating a Long from the BigInt, apply a 32bit mask to the bottom of the number, then shift down and apply the same mask.

src/cmap/wire_protocol/responses.ts
```diff
@@ -83,27 +95,98 @@ export class MongoDBResponse extends OnDemandDocument {
...
+    this.batchSize = this.batch.size();
+  }
+
+  get length() {
+    return Math.max(this.batchSize - this.iterated, 0);
+  }
+
+  shift(options?: BSONSerializeOptions): any {
+    this.iterated += 1;
+    this.values ??= this.batch?.valuesAs(BSONType.object) ?? null;
```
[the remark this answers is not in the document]
nbbeeken: We cannot because of the isError check, I left a comment on that line, I'll experiment with how that can be removed but if I recall my last response PR work correctly I don't think it is easy unless we move writeConcernErrors out to the operation layer

src/cmap/wire_protocol/responses.ts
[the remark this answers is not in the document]
baileympearson: `Long.fromBigInt`?

src/cmap/wire_protocol/responses.ts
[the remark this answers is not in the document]
nbbeeken: done

src/cmap/wire_protocol/responses.ts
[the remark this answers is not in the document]
nbbeeken: Added comment

src/cmap/wire_protocol/responses.ts
nbbeeken: Fixed

src/cmap/wire_protocol/responses.ts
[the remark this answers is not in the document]
nbbeeken: This is fixed without changing anything related to wc errors

src/cmap/wire_protocol/responses.ts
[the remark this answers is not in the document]
baileympearson: Why not fix BSON's fromBigInt implementation and then use fromBigInt here?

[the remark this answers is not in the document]
nbbeeken: We can, I'll file a ticket but there are requirements around using bigint literals in BSON that make this a bit difficult to implement efficiently over there.

src/client-side-encryption/auto_encrypter.ts
```diff
@@ -473,8 +474,19 @@ export class AutoEncrypter {
...
+    response: Uint8Array | Document | MongoDBResponse,
+    options: CommandOptions = {}
+  ): Promise<Document> {
+    const buffer = MongoDBResponse.is(response)
+      ? response.toBytes()
+      : Buffer.isBuffer(response)
```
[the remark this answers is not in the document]
baileympearson: I don't feel strongly, but BSON's types specify `Uint8Array`s but detect runtime and return `Buffer`s if running in Nodejs.  Since our FLE bindings only support Nodejs, using `Buffer` APIs is safe

src/cmap/wire_protocol/responses.ts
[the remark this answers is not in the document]
baileympearson: Any reason you chose a comment over just serializing the document?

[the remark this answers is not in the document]
nbbeeken: Running the BSON serializer at import time seems undesirable to me. There _should_ be nothing wrong with this but it's opting into more than logic than I think is necessary for this task.

Ideally, I don't need to make this instance at all maybe there's something more clever than hard-coded bytes here. 🤔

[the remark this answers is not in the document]
nbbeeken: Avoiding making a CursorResponse altogether seems like a win. Our AbstractCursor lowered its "contract" requirements from List to just `length and .shift` (still has the push(Many) methods but it will not call them when `CursorResponse.is()` is true)

src/client-side-encryption/auto_encrypter.ts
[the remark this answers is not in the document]
nbbeeken: "Safe" within some degree of assumption about the environment you are running in. I can make BSON in Node.js return Uint8Arrays:

```js
Object.defineProperty(Buffer.prototype, '_isBuffer', {value: true}); 
Buffer.isBuffer(require('bson').serialize({a:1}));
```

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -297,4 +297,111 @@ describe('Client Side Encryption Functional', function () {
       });
     });
   });
+
+  describe(
```
baileympearson, line 301: Is auto_encrypter.test.ts a better place for this test?

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -297,4 +297,111 @@ describe('Client Side Encryption Functional', function () {
...
+    'when @@mdb.decorateDecryptionResult is set on autoEncrypter',
+    { requires: { mongodb: '>=7.0' } },
```
baileympearson, line 303: What about this test requires >= 7.0?

src/operations/find.ts
```diff
@@ -77,13 +77,10 @@ export class FindOperation extends CommandOperation<Document> {
...
   filter: Document;
 
-  constructor(
-    collection: Collection | undefined,
-    ns: MongoDBNamespace,
-    filter: Document = {},
-    options: FindOptions = {}
-  ) {
-    super(collection, options);
+  public encryptionEnabled = false;
```
baileympearson, line 80: why is this necessary?  without this, when encryption is enabled:

- the find operation will provide a MonogoDBResponse to the server
- the server will provide it to the connection
- the crypto connection won't pass it along, so a raw response is returned

baileympearson (changes requested):

test/integration/client-side-encryption/driver.test.ts
nbbeeken: https://spruce.mongodb.com/task/mongo_node_driver_next_rhel80_large_Node20_test_4.0_sharded_cluster_patch_ca2bfb0c536b665862a944c214e7cc51b49dbf1e_662fd327eb2a880007c01c39_24_04_29_17_04_51/logs?execution=0&sortBy=STATUS&sortDir=ASC

7.0 was too agressive, the $jsonSchema format I've used can't be used on 4.0 or below. I tried figuring out when this was added but it is not clear to me, do you happen to know? anyway lowered the filter requirement.

test/integration/client-side-encryption/driver.test.ts
nbbeeken: In the unit directory? I set this up to be integration because the unit coverage we already have did not catch the issue introduced, I want to know that `find` will return the right object

src/operations/find.ts
nbbeeken: You're right, I think I did this in parallel with removing the pass through for crypto and didn't go back, good catch ty!

src/client-side-encryption/auto_encrypter.ts
nbbeeken: Yea I think we do, and I think we should get FLE on to the response train 🚂 but I would need to refactor the state machine and the decorate logic a bit more than is comfortable for this PR, so I think we're better off exempting crypto here and we can refactor in a follow up

baileympearson (approved):

PR #4089, opened 2024-04-18
test(NODE-6052): support assertions on topologyDescriptionChanged in expectEvents

W-A-James:
#### What is changing?
- Pull in spec tests from DRIVERS-2875
- Fix subtle bug in unified test runner where sdam event matching was exiting prematurely
- Skip tests that are failing due to aforementioned runner bug changes. To be addressed in NODE-6045

No

NODE-6052


test/tools/unified-spec-runner/schema.ts
```diff
@@ -95,15 +96,15 @@ export interface UnifiedSuite {
   tests: Test[];
   _yamlAnchors?: Document;
 }
-export const TopologyType = Object.freeze({
+export const TopologyId = Object.freeze({
```
aditi-khare-mongoDB, line 99: Why are we making this equal `TopologyID`?

aditi-khare-mongoDB (changes requested):

test/tools/unified-spec-runner/schema.ts
W-A-James: It's a pattern we use when we want to create a type that is an enum without actually using Typescript's builtin enums. The main reason we don't use Typescript's builtin enums is that they were invented by Typescript don't map cleanly onto any existing Javascript construct. (@nbbeeken, @baileympearson correct me if I'm wrong here) 

Also doing it this way frees up the actual `TopologyType` "enum" we have in the driver for ease of use in this file.

For another example of this pattern, see how the driver's `TopologyType` "enum" is defined [here](https://github.com/mongodb/node-mongodb-native/blob/eece8c1b9fcd0aa7c12722f4fe4e50e2123102b8/src/sdam/common.ts#L17-L27)

aditi-khare-mongoDB: Bookkeeping:
Discussed this thread with Warren over slack specifically clarifying that I'm wondering about the naming choice not the enum. And he will update the name to something else.

aditi-khare-mongoDB (approved):
LGTM! Just waiting for CI to finish before I move to team review

test/integration/server-discovery-and-monitoring/server_discovery_and_monitoring.spec.test.ts
```diff
@@ -3,11 +3,17 @@ import * as path from 'path';
 import { loadSpecTests } from '../../spec';
 import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
 
+const skipTable: { pattern: string; reason: string }[] = [
+  { pattern: 'Topology lifecycle', reason: 'see NODE-5723' },
+  { pattern: 'connect with serverMonitoringMode=stream >=4.4', reason: 'NODE-6045' },
+  { pattern: 'connect with serverMonitoringMode=auto >=4.4', reason: 'NODE-6045' }
```
baileympearson, line 9: We usually use the pattern `TODO(NODE-xxxx)` in our skip reasons - can we update them?

Also, can we describe why the test is skipped instead of just linking a jira ticket?  That's easier to grok when scanning test output.

test/tools/unified-spec-runner/match.ts
```diff
@@ -561,63 +561,65 @@ function compareEvents(
           );
         }
       }
-      return;
     } else if (expectedEvent.serverHeartbeatStartedEvent) {
```
baileympearson, line 564: Why are the early returns a bug?  Aren't these all mutually exclusive?  for example - we can't expect than an event is both a heartbeat started and a heartbeat failed.

baileympearson (changes requested):

test/tools/unified-spec-runner/match.ts
aditi-khare-mongoDB: I imagine Warren's thinking is that these early returns are redundant since we have `if/else if/else` blocks encompassing the entire function.

W-A-James: It's not that they're redundant, it's that the early returns are occurring while we're looping over the array of events we want to assert against. Effectively what was happening when asserting on sdam events was that we'd exit after the first event when we had other events to check.

aditi-khare-mongoDB (approved):

baileympearson (approved):

PR #4093, opened 2024-04-22
refactor(NODE-6201): cursor to use fetchBatch when current batch is empty

nbbeeken:
#### What is changing?

- Move `next` into cursor and rename to fetchBatch
- Remove transform and blocking flags
- tryNext/next now call fetchBatch as appropriate to get the same behavior the flags provided
- move cleanupCursor logic to cleanup()

No

The `next` floating function owned too many details of the implementation of each cursor API, tryNext and next now do a small amount more lifting and we have a fetchBatch method that is focused on obtaining the next batch either with the init command or with a getMore.

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -533,7 +533,7 @@ export abstract class AbstractCursor<
     const oldTransform = this[kTransform] as (doc: TSchema) => TSchema; // TODO(NODE-3283): Improve transform typing
     if (oldTransform) {
       this[kTransform] = doc => {
-        return transform(oldTransform(doc));
+        return this.makeSafeTransform(transform)(oldTransform(doc));
       };
     } else {
       this[kTransform] = this.makeSafeTransform(transform);
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -786,7 +786,7 @@ export abstract class AbstractCursor<
     } catch (error) {
       squashError(error);
     } finally {
-      if (!session.hasEnded && session?.owner === this) {
+      if (session?.owner === this) {
         await session.endSession({ error });
       }
       if (!session?.inTransaction()) {
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -720,7 +720,9 @@ export abstract class AbstractCursor<
 
     if (this[kId] == null) {
       await this.cursorInit();
-      return;
+      // If the cursor died or returned documents, return
+      if (this[kDocuments].length !== 0 || this.isDead) return;
+      // Otherwise, run a getMore
     }
 
     // otherwise need to call getMore
@@ -868,6 +870,11 @@ class ReadableCursorStream extends Readable {
   }
 
   private _readNext() {
+    if (this._cursor[kId] === Long.ZERO) {
+      this.push(null);
+      return;
+    }
+
     // eslint-disable-next-line github/no-then
     this._cursor.next().then(
       result => {
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -341,7 +341,7 @@ export abstract class AbstractCursor<
 
   stream(options?: CursorStreamOptions): Readable & AsyncIterable<TSchema> {
     if (options?.transform) {
-      const transform = this.makeSafeTransform(options.transform);
+      const transform = options.transform;
       const readable = new ReadableCursorStream(this);
 
       const transformedStream = readable.pipe(
@@ -390,12 +390,11 @@ export abstract class AbstractCursor<
       throw new MongoCursorExhaustedError();
     }
 
-    const transform = this[kTransform];
-
     do {
       const doc = this[kDocuments].shift();
       if (doc != null) {
-        return transform != null ? transform(doc) : doc;
+        if (this[kTransform] != null) return await this.transformDocument(doc);
+        return doc;
       }
       await this.fetchBatch();
     } while (!this.isDead || this[kDocuments].length !== 0);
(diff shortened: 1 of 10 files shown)


src/sessions.ts
```diff
@@ -523,7 +523,11 @@ export function maybeClearPinnedConnection(
     const servers = Array.from(topology.s.servers.values());
     const loadBalancer = servers[0];
 
-    if (options?.error == null || options?.force) {
+    if (
+      options?.error == null ||
+      options?.error?.name === 'MongoExpiredSessionError' ||
+      options?.force
+    ) {
```
nbbeeken, line 530: Without this change, endSession is sensitive to the exact order of endSession calls that were made by our cursor before this refactor.

src/cursor/abstract_cursor.ts
```diff
@@ -309,33 +305,26 @@ export abstract class AbstractCursor<
...
+          return;
+        }
+
+        if (this.closed && this[kDocuments].length === 0) {
+          return;
+        }
+
+        if (this[kId] != null && this.isDead && this[kDocuments].length === 0) {
+          return;
+        }
```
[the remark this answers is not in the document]
nbbeeken: Yes, and ty! I intended these to be written as if you would read them:
- if killed - return
- if closed AND there's no more documents (because you can be closed but still finishing the last batch) - return
- if there is a truthy Id (which means we already ran the initial command) the cursor isDead (generally, id = 0) and there are no more documents - return

src/cursor/abstract_cursor.ts
```diff
@@ -137,13 +133,13 @@ export abstract class AbstractCursor<
   CursorEvents extends AbstractCursorEvents = AbstractCursorEvents
 > extends TypedEventEmitter<CursorEvents> {
   /** @internal */
-  [kId]: Long | null;
+  private [kId]: Long | null;
```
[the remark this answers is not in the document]
nbbeeken: Nope, I am happy to do that in another PR if this one is otherwise good to go, wouldn't take long

src/cursor/abstract_cursor.ts
```diff
@@ -691,243 +695,156 @@ export abstract class AbstractCursor<
...
-      if (!session.inTransaction()) {
-        maybeClearPinnedConnection(session, { error });
+  /** @internal */
+  private hasEmittedClose = false;
+  /** @internal */
+  private emitClose() {
+    try {
+      if (!this.hasEmittedClose && (this[kDocuments].length === 0 || this[kClosed])) {
+        // @ts-expect-error: not sure..
+        this.emit('close');
```
baileympearson, line 807: If we're going to leave a ts-expect-error in, can we document why it's an error?  but tbh I'd expect this to work, not sure why an expect error is needed here.

W-A-James (approved):

baileympearson (changes requested):

src/cursor/abstract_cursor.ts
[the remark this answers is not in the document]
nbbeeken: #4133

test/integration/node-specific/cursor_async_iterator.test.js
```diff
@@ -72,6 +72,18 @@ describe('Cursor Async Iterator Tests', function () {
       }
     });
 
+    it('should not iterate if closed immediately', async function () {
```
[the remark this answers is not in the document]
nbbeeken: From slack, we're showing "no documents are produced" here, a small edge case but proves we don't throw nor return `{value: null, done: false}` from our async iterator.

src/cursor/abstract_cursor.ts
nbbeeken: Yes sorry meant to get back to this is is declared in the same way ConnectionEvents are so I'm not sure what is going on here.

```ts
Argument of type '[]' is not assignable to parameter of type 'Parameters<CursorEvents["close"]>'.ts(2345)
```

Maybe it is because CursorEvents is generic so Parameters<CursorEvents["close"]> may not be assignable to `[]`. Not sure how to require extenders do not add parameters.

src/cursor/abstract_cursor.ts
```diff
@@ -515,14 +529,14 @@ export abstract class AbstractCursor<
    * @param transform - The mapping transformation method.
    */
   map<T = any>(transform: (doc: TSchema) => T): AbstractCursor<T> {
-    assertUninitialized(this);
+    this.throwIfInitialized();
     const oldTransform = this[kTransform] as (doc: TSchema) => TSchema; // TODO(NODE-3283): Improve transform typing
     if (oldTransform) {
       this[kTransform] = doc => {
         return transform(oldTransform(doc));
       };
```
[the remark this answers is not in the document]
nbbeeken: Good catch, added a test for this too

src/cursor/abstract_cursor.ts
```diff
@@ -691,243 +695,156 @@ export abstract class AbstractCursor<
...
+          const message =
+            'Cursor returned a `null` document, but the cursor is not exhausted.  Mapping documents to `null` is not supported in the cursor transform.';
+          throw new MongoAPIError(message);
+        }
 
-  if (session.hasEnded) {
-    return await completeCleanup();
+        return result;
+      } catch (error) {
+        // eslint-disable-next-line github/no-then
```
baileympearson, line 828: (optional) it might make more sense to propagate errors from `transform` and instead try-catch the application of the transform each place we use it (which, I think is just one place? `next()`?)

It seems like a small code smell to wrap a synchronous function in another synchronous function that sometimes kicks off an async operation but doesn't await it, when we can remove the async portion of the transform wrapper by calling await cursor.close() in the caller, not the transform.

baileympearson (dismissed):
One optional comment.

src/cursor/abstract_cursor.ts
baileympearson: Or there's no reason `makeSafeTransform` can't be async, since we await it everywhere we use it anyways.

W-A-James (dismissed):

src/cursor/abstract_cursor.ts
nbbeeken: `transformDocument` now awaits the close! and I removed the makeTransform from stream because we didn't disallow null / try/catch previously

baileympearson (approved):

W-A-James (approved):

PR #4095, opened 2024-04-23
feat(NODE-6520): add client side operations timeoutMS

W-A-James:
##### New Error
- Added `MongoOperationTimeoutError` class that is thrown when a CSOT timeout is encountered

##### Changes to `Timeout`
- Add `Timeout.throwIfExpired()` method
- Add `Timeout.remainingTime` getter method

##### Updates to `AbstractOperation`
- Add `timeout` field
	- `timeout` is set at construction if the `timeoutMS` option is provided
##### Implementing CSOT behaviour for server selection
- Update `Topology.selectServer` to accept a `timeout` option which it will use determine whether it has timed out when defined. Otherwise, constructs a `Timeout` using the `serverSelectionMS` option as before
- Update `Topology.selectServer` to throw a `MongoOperationTimeoutError` on timeout when `options.timeout` is provided and retain previous error behaviour otherwise.
- Update `Topology._connect` to pass down `timeout` to `Server.command` call used to execute ping on first connection

##### Implementing CSOT behaviour for connection checkout
- Update `Server.command` to accept `timeout` option.
- Update `ConnectionPool.checkOut` to accept `timeout` option
	- only uses the passed in timeout if the configured `serverSelectionTimeoutMS` is greater than the duration on the `timeout`, otherwise, computes the time elapsed since server selection completed and creates timeout for the `serverSelectionTimeoutMS` deadline

##### Test changes
- Implement Server Selection prose tests from CSOT spec. Left out last two since they require implementation of the connection creation cancellation behaviour that was decided against in DRIVERS-2347
- Implement unit tests from CSOT spec related to server selection and connection checkout. Skipped tests with appropriate messages

##### Misc changes
- Update `resolveOptions` to handle `timeoutMS` option propagation
- - Add `csotMin` helper method that implements the CSOT min algorithm described [here](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#terms)


src/cmap/connect.ts
```diff
@@ -110,7 +113,25 @@ export async function performInitialHandshake(
...
+        options.timeout,
+        conn.command(ns('admin.$cmd'), handshakeDoc, handshakeOptions)
+      ]);
+    } catch (error) {
+      if (TimeoutError.is(error)) {
+        throw new MongoOperationTimeoutError('Timed out during handshake');
+      } else {
+        throw error;
+      }
+    }
```
W-A-James, line 131: https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#server-selection

>  All commands sent during the connection’s handshake MUST use the remaining computedServerSelectionTimeout as their timeoutMS value

src/cmap/connect.ts
```diff
@@ -321,14 +342,40 @@ export async function makeSocket(options: MakeConnectionOptions): Promise<Stream
   const connectTimeoutMS = options.connectTimeoutMS ?? 30000;
   const existingSocket = options.existingSocket;
 
+  let timeout: Timeout | null;
+  if (options.timeout) {
+    options.timeout.throwIfExpired();
+    timeout = Timeout.expires(Timeout.min(connectTimeoutMS, options.timeout.remainingTime));
+  } else {
+    timeout = null;
+  }
```
W-A-James, line 351: > If a new connection is required, min(connectTimeoutMS, remaining computedServerSelectionTimeout) MUST be used as the timeout for TCP socket establishment

src/cmap/connection.ts
```diff
@@ -88,6 +89,11 @@ export interface CommandOptions extends BSONSerializeOptions {
   writeConcern?: WriteConcern;
 
   directConnection?: boolean;
+
+  /** @internal */
+  timeout?: Timeout;
+  /** @internal */
+  serverSelectionTimeout?: Timeout;
```
W-A-James, line 96: Added to facilitate the separation of the overall operation timeout and the server selection timeout requirements

src/cmap/connection_pool.ts
```diff
@@ -393,6 +400,9 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
             : 'Timed out while checking out a connection from connection pool',
           this.address
         );
+        if (options?.timeout) {
+          throw new MongoOperationTimeoutError('Timed out', { cause: timeoutError });
```
W-A-James, line 404: TODO: make message more useful here

src/cmap/connection_pool.ts
```diff
@@ -735,7 +746,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
     }
   }
 
-  private processWaitQueue() {
+  private processWaitQueue(options?: { timeout?: Timeout }) {
```
W-A-James, line 749: NOTE: this is added to allow us to pass timeouts through to `ConnectionPool.createConnection` in order to satisfy the CSOT spec's requirement to apply the serverSelectionTimeout to the creation of new connections when necessary

src/error.ts
```diff
@@ -702,6 +702,12 @@ export class MongoUnexpectedServerResponseError extends MongoRuntimeError {
   }
 }
 
+export class MongoOperationTimeoutError extends MongoRuntimeError {
```
W-A-James, line 705: New error added to support CSOT

src/operations/operation.ts
```diff
@@ -61,6 +62,12 @@ export abstract class AbstractOperation<TResult = any> {
 
   options: OperationOptions;
 
+  /** @internal */
+  timeout?: Timeout;
+
+  /** @internal */
+  serverSelectionTimeout?: Timeout;
```
W-A-James, line 69: populated at execute time

src/timeout.ts
```diff
@@ -78,6 +84,10 @@ export class Timeout extends Promise<never> {
     this.id = undefined;
   }
 
+  throwIfExpired(): void {
```
W-A-James, line 87: add optional message argument to make these errors more useful

src/admin.ts
```diff
@@ -78,7 +78,8 @@ export class Admin {
       new RunAdminCommandOperation(command, {
         ...resolveBSONOptions(options),
         session: options?.session,
-        readPreference: options?.readPreference
+        readPreference: options?.readPreference,
+        timeoutMS: options?.timeoutMS ?? this.s.db.timeoutMS
```
W-A-James, line 82: NOTE FOR REVIEWERS: Needed to do this to ensure that unit tests using ping work

W-A-James:
[Evergreen link](https://spruce.mongodb.com/version/663562e134c6bf0008c9c9e4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

src/operations/operation.ts
```diff
@@ -61,6 +62,12 @@ export abstract class AbstractOperation<TResult = any> {
 
   options: OperationOptions;
 
+  /** @internal */
+  timeout?: Timeout;
```
baileympearson, line 66: I'm confused about the purpose of the `timeout`.  I commented in `Server.ts`, but at no point that I know of will we ever use a generic "CSOT" timeout.  Instead, at each point where we apply CSOT's logic to async operations, we will create custom timeouts that apply CSOT logic (i.e., a timeout that applies to socket read specifically or socket write).  Am I misunderstanding?

Given that, it feels odd to create a promise that we won't use.  And it seems odd to use our generic Timeout class to store CSOT-specific context (`Timeout.min`, `Timeout.remainingTime`).

Thoughts on a generic `CSOTContext` class?  That's basically what the `TimeoutControllerFactory` was in the initial design.  Something that holds CSOT context, is instantiated at operation construction time, and generates timeouts for certain parts of the driver (i.e., `context.timeoutForServerSelection()`, which encapsulates CSOT logic for server selection into a single, unit testable class).

(this relates to my other comment about consolidating timeout logic for server selection - we'll need it eventually for change streams too so it makes sense not to implement it three times).

src/sdam/topology.ts
```diff
@@ -628,6 +647,11 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
           );
         }
 
+        if (options.timeout) {
+          throw new MongoOperationTimeoutError('Timed out during server selection', {
```
baileympearson, line 651: Should we include the TopologyDescription in this error (regular server selection timeout errors do).

src/operations/execute_operation.ts
```diff
@@ -152,9 +153,21 @@ export async function executeOperation<
     selector = readPreference;
   }
 
+  const timeout = operation.timeout;
+  timeout?.throwIfExpired();
+  // TODO: construct serverSelection timeout here, pass it into topology.selectServer and store on
+  // operation for use in operation.execute when we have to perform connection checkout
+  operation.serverSelectionTimeout =
```
baileympearson, line 160: thoughts on consolidating this logic into a function or something?  We have similar logic in the topology

baileympearson (changes requested):
left some small comments.  I'm out this week, so don't consider them blocking!

aditi-khare-mongoDB:
Bookkeeping: 
Waiting for Warren to respond to Bailey's comments and will proceed with review then.

src/sdam/server.ts
```diff
@@ -323,8 +325,23 @@ export class Server extends TypedEventEmitter<ServerEvents> {
 
     try {
       try {
-        return await conn.command(ns, cmd, finalOptions, responseType);
+        if (finalOptions.operationTimeout) {
+          finalOptions.operationTimeout.throwIfExpired();
+          return await Promise.race([
+            finalOptions.operationTimeout,
+            conn.command(ns, cmd, finalOptions)
+          ]);
```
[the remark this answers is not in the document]
W-A-James: Good catch here. Meant to remove this change, but missed it.

src/sdam/server.ts
```diff
@@ -333,6 +350,7 @@ export class Server extends TypedEventEmitter<ServerEvents> {
         operationError.code === MONGODB_ERROR_CODES.Reauthenticate
       ) {
         await this.pool.reauthenticate(conn);
+        // TODO(NODE-5682): Implement CSOT support for socket read/write at the connection layer
         try {
```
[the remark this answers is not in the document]
W-A-James: This was an artefact of my intention to remove the command execution CSOT functionality from this PR

src/sdam/topology.ts
W-A-James: Is adding the originally thrown error as a cause not sufficient for this?

src/operations/operation.ts
W-A-James: The `CSOTContext` class sounds somewhat similar to the `OperationContext` idea that you had explored earlier on in the CSOT design. I'm not against it on principle, but I think we scrapped that idea at some point. Will look for any documents we had on why and report back.

src/sdam/topology.ts
```diff
@@ -457,8 +459,17 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
       }
     }
 
+    const timeoutMS = this.client.options.timeoutMS;
+    const timeout = timeoutMS != null ? Timeout.expires(timeoutMS) : undefined;
+    const serverSelectionTimeoutMS = this.s.serverSelectionTimeoutMS;
+    const serverSelectionTimeout =
+      timeoutMS != null ? Timeout.expires(csotMin(timeoutMS, serverSelectionTimeoutMS)) : undefined;
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: We can also just  do 
```
serverSelectionTimeout  = Timeout.expires(csotMin(timeoutMS, serverSelectionTimeoutMS)
```
and change `csotMin` to allow and detect undefined parameters and return the other parameter, if both are undefined, return 0 (CSOT's Infinity)

```
export function csotMin(duration1: number, duration2: number): number {
  if (duration1 == null && duration2 == null) return 0;
  if (duration1 == null || duration1 === 0) return duration2;
  if (duration2 == null || duration2 === 0) return duration1;
    return Math.min(duration1, duration2);
}
```

Just a suggestion, feel free to ignore, I just feel it would mean less checking for undefined variables in the future.

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
```diff
@@ -1,51 +1,93 @@
...
     () => {}
-  );
+  ).skipReason =
+    'TODO(NODE-5682): Add CSOT support for socket read/write at the connection layer for CRUD APIs';
 
-  context(
+  context.skip(
     'The remaining timeoutMS value should apply to HTTP requests against KMS servers for CSFLE.',
     () => {}
-  );
```
aditi-khare-mongoDB, line 40: nit, but do we want to consolidate the two skipped tests that are marked with the same skip reason `TODO(NODE-5686)`, just for neatness

src/sdam/server.ts
```diff
@@ -310,14 +310,16 @@ export class Server extends TypedEventEmitter<ServerEvents> {
     this.incrementOperationCount();
     if (conn == null) {
       try {
-        conn = await this.pool.checkOut();
+        conn = await this.pool.checkOut({ timeout: options.serverSelectionTimeout });
```
aditi-khare-mongoDB, line 313: let's flag this path to only occur when csot is on explicitly, same logic just a bit different looking code

src/sdam/topology.ts
```diff
@@ -467,7 +478,11 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
 
       const skipPingOnConnect = this.s.options[Symbol.for('@@mdb.skipPingOnConnect')] === true;
       if (!skipPingOnConnect && server && this.s.credentials) {
-        await server.command(ns('admin.$cmd'), { ping: 1 }, {});
+        await server.command(
+          ns('admin.$cmd'),
+          { ping: 1 },
+          { operationTimeout: timeout, serverSelectionTimeout }
```
aditi-khare-mongoDB, line 484: Let's remove the `operationTimeout` from here and from `commandOptions` (relic of CSOT command execution ticket (NODE-5682) which not be done in this ticket.

aditi-khare-mongoDB (changes requested):

W-A-James:
@aditi-khare-mongoDB @nbbeeken should we conditionally clear the timeout on success/a non-timeout failure based on whether or not we created the timeout inside the function that we race the timeout with?
I ask because we don't always own Timeout instance we're using in `Topology.selectServer` or in `ConnectionPool.checkOut`.

src/sdam/topology.ts
W-A-James: Decided to rename the `timeout` option on `AbstractOperation` to `operationTimeout` to make clear it's referring to the time out for the entire operation. This avoids the name conflict with the find options in future and clarifies the intention behind the option/field name.

src/cmap/connection_pool.ts
```diff
@@ -354,17 +355,31 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+    const serverSelectionTimeoutMS = this[kServer].topology.s.serverSelectionTimeoutMS;
 
     const { promise, resolve, reject } = promiseWithResolvers<Connection>();
 
-    const timeout = Timeout.expires(waitQueueTimeoutMS);
+    let timeout: Timeout;
+    if (options?.timeout) {
+      // CSOT enabled
+      // Determine if we're using the timeout passed in or a new timeout
+      if (options.timeout.duration === 0 || serverSelectionTimeoutMS < options.timeout.duration) {
```
W-A-James, line 373:
```
if configured serverSelectionTimeoutMS < configured timeoutMS then:
	create a timeout that expires after (serverSelectionTimeoutMS - timeElapsed)
else:
	reuse timeout passed in via options
```

aditi-khare-mongoDB (dismissed):

src/operations/operation.ts
nbbeeken: Just to close the loop here: We do use the "csot" timeout when it is the lower bound of time remaining before the operation is meant to expire. Server selection and connection checkout need to never take more than `serverSelectionTimeoutMS` but can be given less time if a csot timeout is smaller than that setting. So the logic in this PR does some gte/lte checks and determines if we need to make a new timer to maintain that requirement.

src/sdam/server.ts
```diff
@@ -310,7 +310,11 @@ export class Server extends TypedEventEmitter<ServerEvents> {
     this.incrementOperationCount();
     if (conn == null) {
       try {
-        conn = await this.pool.checkOut();
+        if (options.operationTimeout) {
+          conn = await this.pool.checkOut({ timeout: options.operationTimeout });
+        } else {
+          conn = await this.pool.checkOut();
+        }
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: I feel Warren's current code is easier to read, but if we do end up going with this suggestion can we leave a clarifying comment?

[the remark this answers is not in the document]
nbbeeken: I am surprised because breaking this up into two calls to checkOut based on a condition that does matter is more to read without more meaningful context given. Whether or not timeout exists, there is no change to how checkOut is, practically, invoked because the typescript reports that field as optional.

I would actually take this further:
```js
conn = await this.pool.checkOut(options);
```
Why do we need to make a new object here? passing through options should be fine right? Less branching paths the less there is to debug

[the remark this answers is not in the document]
nbbeeken: Can you elaborate on what may accidentally break the spec without a test warning us?

src/operations/operation.ts
```diff
@@ -61,6 +62,9 @@ export abstract class AbstractOperation<TResult = any> {
 
   options: OperationOptions;
 
+  /** @internal */
+  operationTimeout?: Timeout;
+
```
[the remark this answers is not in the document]
W-A-James: I decided to replace `timeout` with `operationTimeout` anywhere else it comes up on the operation. There's an option on `FindOptions` called `timeout` that's a boolean, but I can override the value passed into `server.command` if needs be if we'd prefer to just keep it as `timeout`

src/cmap/connection_pool.ts
```diff
@@ -354,17 +355,33 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
+      ) {
+        timeout = Timeout.expires(serverSelectionTimeoutMS - options.timeout.timeElapsed);
+      } else {
+        timeout = options.timeout;
+      }
+    } else {
+      timeout = Timeout.expires(waitQueueTimeoutMS);
+    }
+
+    timeout.throwIfExpired();
```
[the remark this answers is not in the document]
W-A-James: Good call. Do we want to encapsulate the throwIfExpired calls into a `Timeout.race<T>(value: Promise<T>)` method? it would be pretty much just call throwIfExpired and then do the same Promise.race call as we do currently.

That would make it such that we only call throwIfExpired just as we go to complete the blocking work, which would guarantee that we catch the timeout error whenever it occurs.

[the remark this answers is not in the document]
nbbeeken: On paper, sounds good, in practice, if there are any instances where we need to not even begin work if this is already expired that is hard to accomplish with a helper that isn't aware of the context you're in.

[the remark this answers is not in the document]
W-A-James: That's a good point about needing the context. As far as the question about whether or not `checkout started` is correct if we're given an expired timeout. We're considering the case here that server selection succeeded, some amount of time passed between server selection succeeding and entering `ConnectionPool.checkOut` and the timeout expiring in that intermediate time.

I can see a strong argument for emitting the checkout started event as we enter `checkOut` (which we do currently), but not putting the waitqueue member in the queue at all if the timeout has already expired so as not to waste work. As you point out, the only time this would occur is when CSOT is enabled and the configured `serverSelectionTimeoutMS` is greater than the configured `timeoutMS`, meaning we'd be using the timeout passed in on the options. In that case, it would be safest and make the most sense to throw a `MongoOperationTimeoutError` with the timeout being the cause, but I'm not sure how that gels with existing expectations for the CMAP spec.

If we just want to proceed with this PR, I'd argue that we should call throwIfExpired just before the call to Promise.race (after we put the waitqueue member in the queue) since that's the beginning of the blocking work which we'd then cancel and that's what the CSOT spec is primarily concerned with and we can file a ticket to address this if it ends up causing further issues.

src/cmap/connection_pool.ts
```diff
@@ -354,17 +355,33 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
 
     const { promise, resolve, reject } = promiseWithResolvers<Connection>();
 
-    const timeout = Timeout.expires(waitQueueTimeoutMS);
+    let timeout: Timeout;
+    if (options?.timeout) {
+      // CSOT enabled
+      // Determine if we're using the timeout passed in or a new timeout
+      if (
+        csotMin(options.timeout.duration, serverSelectionTimeoutMS) === serverSelectionTimeoutMS
```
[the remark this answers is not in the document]
W-A-James: Bailey had the opposing opinion: that it should be a standalone helper because its only purpose is to implement CSOT logic.

Re the naming, I'm not sure we'll end up with a satisfactorily descriptive name for this without it being long and unwieldy. How do we feel about just adding a detailed doc comment?

src/cmap/connection_pool.ts
```diff
@@ -367,19 +367,21 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
 
     const { promise, resolve, reject } = promiseWithResolvers<Connection>();
 
-    let timeout: Timeout;
-    let clearTimeout: boolean;
+    let timeout: Timeout | null = null;
+    let clearTimeout: boolean | null = null;
```
nbbeeken, line 371: would it be wrong to default this to false?

nbbeeken (changes requested):

src/cmap/connection_pool.ts
```diff
@@ -354,35 +356,54 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
     );
 
     const waitQueueTimeoutMS = this.options.waitQueueTimeoutMS;
+    const serverSelectionTimeoutMS = this[kServer].topology.s.serverSelectionTimeoutMS;
 
     const { promise, resolve, reject } = promiseWithResolvers<Connection>();
 
-    const timeout = Timeout.expires(waitQueueTimeoutMS);
+    let timeout: Timeout | null = null;
+    let clearTimeout = false;
```
nbbeeken, line 371: Can we use instance equality to determine this?

nbbeeken (changes requested):

src/cmap/connection_pool.ts
```diff
@@ -354,35 +354,49 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
...
     const { promise, resolve, reject } = promiseWithResolvers<Connection>();
 
-    const timeout = Timeout.expires(waitQueueTimeoutMS);
+    let timeout: Timeout | null = null;
+    if (options?.timeout) {
+      // CSOT enabled
+      // Determine if we're using the timeout passed in or a new timeout
+      if (options.timeout.duration > 0 || serverSelectionTimeoutMS > 0) {
+        if (
+          csotMin(options.timeout.duration, serverSelectionTimeoutMS) === serverSelectionTimeoutMS
```
nbbeeken, line 374: This is still and equals check, sorry I think we discussed it but didn't leave a comment, if duration is the same then we'll create a new timeout when we can use the existing one.

src/sdam/topology.ts
```diff
@@ -467,7 +473,7 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
 
       const skipPingOnConnect = this.s.options[Symbol.for('@@mdb.skipPingOnConnect')] === true;
       if (!skipPingOnConnect && server && this.s.credentials) {
-        await server.command(ns('admin.$cmd'), { ping: 1 }, {});
+        await server.command(ns('admin.$cmd'), { ping: 1 }, { timeout: timeout });
```
nbbeeken, line 476:
```suggestion
        await server.command(ns('admin.$cmd'), { ping: 1 }, { timeout });
```

nbbeeken (changes requested):

nbbeeken (approved):

src/operations/operation.ts
baileympearson: > The CSOTContext class sounds somewhat similar to the OperationContext idea that you had explored earlier on in the CSOT design. I'm not against it on principle, but I think we scrapped that idea at some point. Will look for any documents we had on why and report back.

The context would be similar to the timeout factory, not the operation context.  We decided not to implement an operation context because we can pass CSOT-related data on the options objects in the driver.  I'm proposing a CSOT context to encapsulate CSOT logic, which we would then pass through the driver on the options objects.

---
I do not consider this thread resolved.  I think the current implementation is more complicated than necessary because we only sometimes re-use the timeout.

- This means server selection and connection checkout must be responsible for determining when to reuse the timeout.
- We can't unconditionally clear the timeout because it may be re-used later.

Instead, I propose we either:

1. Always create a new timeout for server selection and connection checkout.
3. Always reuse the same Timeout object, but just reset the Timeout's interval for server selection and connection checkout.

Regardless of which is chosen, I think the resultant code is simpler because server selection and connection checkout 1) do not worry about whether or not they need to use a cached timeout or create a new one 2) they can always clear the timeout.

This works especially nicely with the TimeoutFactory or a TimeoutContext, because we can encapsulate all timeout related logic into a single place that's easily unit testable.  I'm partial to the factory approach:

```typescript
class TimeoutFactory {
  private timeoutMS: number | null;
  private started = now();

  getTimeoutForServerSelection(): Timeout {
    // returns a timeout, handling CSOT vs Legacy timeout logic
  }
}

class Topology {

  selectServer(options: { ..., timeoutFactory: TimeoutFactory }) {
    ...
    const timeout = timeoutFactory.getTimeoutForServerSelection();
    try {
		....
	} finally {
      timeout.clear();
    }
  }
}
```

Note that with an approach like this, whether or not we reuse a timeout can easily be encapsulated into the TimeoutFactory by instantiating a timeout when the factory is constructed and returning the cached timeout where needed.

But a context class could suffice too:

```typescript
class TimeoutContext {
  private timeoutMS: number | null;
  private started = now();

  getTimeoutForServerSelection(): number {}
}


class Topology {

  selectServer(options: { ..., timeoutContext: TimeoutContext }) {
    ...
    const timeout = Timeout.expires(timeoutFactory.getTimeoutForServerSelection());
    try {
		....
	} finally {
      timeout.clear();
    }
  }
}
```

An approach like this consolidates CSOT logic and can be reused outside of the main code path (i.e., topology connect).  

I don't think this work needs to block this PR.  But I do want to make sure we discuss this, and I'd like to consider one of these approaches in a future ticket.

src/sdam/topology.ts
```diff
@@ -457,8 +458,14 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
       }
     }
 
+    const timeoutMS = this.client.options.timeoutMS;
+    const timeout = timeoutMS != null ? Timeout.expires(timeoutMS) : undefined;
```
baileympearson, line 462: Is this correct if we're auto-connecting the client?

baileympearson (dismissed):

aditi-khare-mongoDB (approved):

src/sdam/topology.ts
```diff
@@ -889,8 +922,6 @@ function drainWaitQueue(queue: List<ServerSelectionRequest>, drainError: MongoDr
       continue;
     }
```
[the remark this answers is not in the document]
W-A-James: We rely on the latter case. The logic being that all server selection requests are created in `selectServer`, so even if we have multiple operations running concurrently, e.g. 

```ts
Promise.all([
   coll.insertOne({a:10}),
   coll.insertOne({a:11})
]);
```

and `drainWaitQueue` gets called before server server selection completes for both of of them (e.g. if the client gets closed) they would each have a catch handler that appropriately clears the timeout.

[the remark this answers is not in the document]
baileympearson: Is there any harm in clearing the timeout here as well?  I'd imagine it's a no-op to clear an already cleared timeout.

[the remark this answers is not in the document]
W-A-James: It shouldn't be an issue to clear again, no. I can go ahead and add that logic.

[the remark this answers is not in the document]
W-A-James: Wait, looking again, the ServerSelectionRequest instances don't actually have a reference to the Timeout, so I'd have to change that if we want the behaviour you're requesting.

src/sdam/server.ts
```diff
@@ -336,6 +346,7 @@ export class Server extends TypedEventEmitter<ServerEvents> {
         operationError.code === MONGODB_ERROR_CODES.Reauthenticate
       ) {
         await this.pool.reauthenticate(conn);
+        // TODO(NODE-5682): Implement CSOT support for socket read/write at the connection layer
```
nbbeeken, line 349:
```suggestion
```

src/sdam/topology.ts
```diff
@@ -178,8 +178,11 @@ export interface SelectServerOptions {
   session?: ClientSession;
   operationName: string;
   previousServer?: ServerDescription;
-  /** @internal*/
-  timeout?: Timeout;
+  /**
+   * @internal
+   * TODO(NODE-5685): Make this required
+   * */
+  timeoutContext?: TimeoutContext;
```
nbbeeken, line 185: Change streams didn't end up using LegacyTimeoutContext so this is still optional

etc/notes/errors.md
```diff
@@ -109,6 +109,10 @@ This class should **never** be directly instantiated.
 | **MongoGridFSChunkError**   | Thrown when a malformed or invalid chunk is encountered when reading from a GridFS Stream. |
 | **MongoUnexpectedServerResponseError**   | Thrown when the driver receives a **parsable** response it did not expect from the server. |
 
+### `MongoOperationTimeoutError`
+
+- TODO(NODE-5688): Add MongoOperationTimeoutError documentation
```
nbbeeken, line 114: Needs docs but we can fill this out later

src/error.ts
```diff
@@ -128,6 +128,9 @@ function isAggregateError(e: unknown): e is Error & { errors: Error[] } {
  * mongodb-client-encryption has a dependency on this error, it uses the constructor with a string argument
  */
 export class MongoError extends Error {
+  get [Symbol.toStringTag]() {
+    return this.name;
+  }
```
nbbeeken, line 133: I think we were using this at one point and stopped -- we should remove it before unnecessarily introducing it

src/connection_string.ts
```diff
@@ -1092,6 +1092,7 @@ export const OPTIONS = {
     type: 'string'
   },
   socketTimeoutMS: {
+    deprecated: 'Please use timeoutMS instead',
```
nbbeeken, line 1095: Undo the deprecations since we decided to ship timeoutMS expiremental. We will deprecate later.

src/timeout.ts
```diff
@@ -76,10 +100,16 @@ export class Timeout extends Promise<never> {
...
+    if (this.timedOut) throw new TimeoutError('Timed out', { duration: this.duration });
   }
 
-  public static expires(durationMS: number, unref?: boolean): Timeout {
-    return new Timeout(undefined, durationMS, unref);
+  public static expires(duration: number, unref?: true): Timeout {
+    return new Timeout(undefined, { duration, unref });
   }
 
   static is(timeout: unknown): timeout is Timeout {
```
nbbeeken, line 115: We do not use this is, remove it along with the toStringTag, less code to maintain

test/benchmarks/driverBench/common.js
```diff
@@ -24,7 +24,9 @@ function loadSpecString(filePath) {
 }
 
 function makeClient() {
-  this.client = new MongoClient(process.env.MONGODB_URI || 'mongodb://127.0.0.1:27017');
+  this.client = new MongoClient(process.env.MONGODB_URI || 'mongodb://127.0.0.1:27017', {
+    timeoutMS: 0
```
nbbeeken, line 28: Let's merge without this. We want our benchmarks to run in default mode. we'll merge more bench tasks that change this setting

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -471,3 +694,589 @@ describe('Range Explicit Encryption with JS native types', function () {
     });
   });
 });
+
+describe('CSOT', function () {
+  describe('Auto encryption', function () {
```
nbbeeken, line 699: Should this be nested above? unusual to have multiple top-level suites.

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -1,8 +1,48 @@
...
+  MongoServerSelectionError,
+  now,
+  ObjectId,
+  promiseWithResolvers,
+  squashError
+} from '../../mongodb';
+import { type FailPoint, makeMultiBatchWrite, measureDuration } from '../../tools/utils';
+import { filterForCommands } from '../shared';
+
 // TODO(NODE-5824): Implement CSOT prose tests
```
nbbeeken, line 31: RM TODO. 4 is intentionally omitted.

test/integration/client-side-operations-timeout/client_side_operations_timeout.spec.test.ts
```diff
@@ -1,9 +1,66 @@
...
 
 import { loadSpecTests } from '../../spec';
 import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
 
-// TODO(NODE-5823): Implement unified runner operations and options support for CSOT
-describe.skip('CSOT spec tests', function () {
-  runUnifiedSuite(loadSpecTests(join('client-side-operations-timeout')));
+const skippedSpecs = {};
+
+const skippedTests = {
```
nbbeeken, line 9: NODE-6305 is closed but it is marked as the reason for skipping some of these tests

test/integration/client-side-operations-timeout/client_side_operations_timeout.spec.test.ts
```diff
@@ -1,9 +1,66 @@
...
+  'timeoutMS can be configured on a MongoClient - createChangeStream on client': 'TODO(NODE-6305)',
+  'timeoutMS applies to whole operation, not individual attempts - createChangeStream on client':
+    'TODO(NODE-6305)',
+  'Tailable cursor iteration timeoutMS is refreshed for getMore - failure': 'TODO(NODE-6305)',
+  'Tailable cursor awaitData iteration timeoutMS is refreshed for getMore - failure':
+    'TODO(NODE-6305)',
+  'command is not sent if RTT is greater than timeoutMS': 'TODO(DRIVERS-2965)',
```
nbbeeken, line 16: DRIVERS-2965 is about a contradiction in the spec, these skips are expected

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
```diff
@@ -1,51 +1,257 @@
...
+    );
+  });
+
+  context.skip('Socket timeouts', function () {
+    context(
+      'If timeoutMS is unset, operations fail after two non-consecutive socket timeouts.',
+      () => {}
+    );
+  }).skipReason =
+    'TODO(NODE-5682): Add CSOT support for socket read/write at the connection layer for CRUD APIs';
```
nbbeeken, line 110: NODE-5682 is closed but this test wasn't implemented

test/integration/client-side-encryption/driver.test.ts
dariakp: The previous block says "CSOT on client encryption", so should this be "CSOT on auto encryption"? What is the difference between what the previous block tests vs this one? ("client encryption" is a bit general)

src/connection_string.ts
dariakp: Use NODE-6491 for the un-deprecation and tag a TODO to that ticket (mention the undeprecation in NODE-6491)

src/change_stream.ts
```diff
@@ -538,7 +540,12 @@ export type ChangeStreamEvents<
   end(): void;
   error(error: Error): void;
   change(change: TChange): void;
-} & AbstractCursorEvents;
+  /**
+   * @remarks Note that the `close` event is currently emitted whenever the internal `ChangeStreamCursor`
+   * instance is closed, which can occur multiple times for a given `ChangeStream` instance.
+   */
+  close(): void;
```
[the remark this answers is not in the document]
dariakp: Do we want to link the ticket as a TODO here?

etc/notes/errors.md
dariakp: We can add a subtask on the documentation ticket and retag to the subtask since all the other docs changes already merged, unless you were thinking of filing a separate ticket entirely?

src/sdam/topology.ts
dariakp: To clarify, should it be optional? Or should it be required?

test/types/mongodb.test-d.ts
```diff
@@ -20,9 +20,39 @@ declare const options: MongoDBDriver.MongoClientOptions;
 expectDeprecated(options.w);
 expectDeprecated(options.journal);
 expectDeprecated(options.wtimeoutMS);
+expectDeprecated(options.socketTimeoutMS);
```
nbbeeken, line 23: Fix deprecation tests

test/benchmarks/driverBench/common.js
dariakp: Do we have a ticket tracking creating more benchmarks? (I assume we want the tasks that you have on your perf test branch)

test/unit/explain.test.ts
```diff
@@ -50,4 +50,42 @@ describe('class Explain {}', function () {
       });
     });
   });
+
+  describe('parseTimeoutOptions()', function () {
```
nbbeeken, line 54: minor: Test name does not match helper name

test/integration/client-side-operations-timeout/client_side_operations_timeout.spec.test.ts
dariakp: I assume we want to try unskipping and see what happens?

nbbeeken (changes requested):
Done with todo notes

src/sdam/topology.ts
nbbeeken: The desire is for it to be required from a code maintenance perspective ideally we always have a context (either CSOT or not) passed this point so code doesn't have to think about the two separate cases. - this does not have API impact so we can improve this later.

etc/notes/errors.md
nbbeeken: I was thinking separate ticket, this file is more our notes for the hierarchy of errors but it is not entirely only useful to us so we should update it soon (so not blocking)

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
dariakp: Was this supposed to be a spec test? The file isn't spec, so just curious how the placeholder ended up here to begin with

test/benchmarks/driverBench/common.js
nbbeeken: I think [NODE-6459](https://jira.mongodb.org/browse/NODE-6459) covers the tracking we need. Here we can just revert this setting.

test/integration/client-side-encryption/driver.test.ts
nbbeeken: Because we parallelized this work the tests merged as separate blocks. I thought I could wholesale merge this suite with the one above but there is mixed nesting that would need to be undone. 

This CSOT block covers Aditi's AutoEncryption, state machine, and createEncryptedCollection changes, the one above is all the other helpers on ClientEncryption that I worked on. So I would want to move createEncryptedCollection into that block and de-nest auto and statemachine.

But there's no actual test changes so let's follow up on this.

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
dariakp: Still need to remove TODO

test/integration/client-side-operations-timeout/client_side_operations_timeout.spec.test.ts
```diff
@@ -1,9 +1,61 @@
...
+  'Tailable cursor iteration timeoutMS is refreshed for getMore - failure': 'TODO(DRIVERS-2965)',
+  'Tailable cursor awaitData iteration timeoutMS is refreshed for getMore - failure':
+    'TODO(DRIVERS-2965)',
+  'command is not sent if RTT is greater than timeoutMS': 'TODO(DRIVERS-2965)',
+  'Non-tailable cursor iteration timeoutMS is refreshed for getMore if timeoutMode is iteration - failure':
+    'TODO(DRIVERS-2965)',
+  'maxTimeMS value in the command is less than timeoutMS':
+    'TODO(DRIVERS-2970): see modified test in unified-csot-node-specs',
+  'timeoutMS is refreshed for getMore - failure':
+    'TODO(DRIVERS-2965): see modified test in unified-csot-node-specs', // Skipping for both tailable awaitData and tailable non-awaitData cursors
```
dariakp, line 19: confused about this comment

test/integration/sessions/sessions.prose.test.ts
```diff
@@ -131,10 +134,15 @@ describe('Sessions Prose Tests', () => {
...
 
     before(() => {
-      childProcess = spawn('mongocryptd', ['--port', mongocryptdTestPort, '--ipv6'], {
-        stdio: 'ignore',
-        detached: true
-      });
+      const pidFile = path.join(os.tmpdir(), new ObjectId().toHexString());
+      childProcess = spawn(
+        'mongocryptd',
+        ['--port', mongocryptdTestPort, '--ipv6', '--pidfilepath', pidFile],
```
dariakp, line 140: why was this needed for CSOT (but not main)?

dariakp (changes requested):

etc/notes/errors.md
```diff
@@ -109,6 +109,10 @@ This class should **never** be directly instantiated.
 | **MongoGridFSChunkError**   | Thrown when a malformed or invalid chunk is encountered when reading from a GridFS Stream. |
 | **MongoUnexpectedServerResponseError**   | Thrown when the driver receives a **parsable** response it did not expect from the server. |
 
+### `MongoOperationTimeoutError`
+
+- TODO(NODE-6491): Add MongoOperationTimeoutError documentation
```
dariakp, line 114:
```suggestion
- TODO(NODE-6516): Add MongoOperationTimeoutError documentation
```

src/client-side-encryption/auto_encrypter.ts
```diff
@@ -310,7 +310,10 @@ export class AutoEncrypter {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore: TS complains as this always returns true on versions where it is present.
       if (net.getDefaultAutoSelectFamily) {
-        Object.assign(clientOptions, autoSelectSocketOptions(this._client.options));
+        // AutoEncrypter is made inside of MongoClient constructor while options are being parsed,
+        // we do not have access to the options that are in progress.
+        // TODO(NODE-NODE-6449): AutoEncrypter does not use client options for autoSelectFamily
```
dariakp, line 315:
```suggestion
        // TODO(NODE-6449): AutoEncrypter does not use client options for autoSelectFamily
```

src/collection.ts
```diff
@@ -669,7 +686,9 @@ export class Collection<TSchema extends Document = Document> {
         new DropIndexOperation(this as TODO_NODE_3286, '*', resolveOptions(this, options))
       );
       return true;
-    } catch {
+    } catch (error) {
+      if (error instanceof MongoOperationTimeoutError) throw error; // TODO: Check the spec for index management behaviour/file a drivers ticket for this
```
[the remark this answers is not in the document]
dariakp: What are we planning to do about this?

src/gridfs/download.ts
```diff
@@ -28,7 +30,7 @@ export interface GridFSBucketReadStreamOptions {
    * to be returned by the stream. `end` is non-inclusive
    */
   end?: number;
-  /** @internal TODO(NODE-5688): make this public */
+  /** @public */
```
dariakp, line 33: Should we have an experimental tag in this file?

src/gridfs/upload.ts
```diff
@@ -35,7 +42,7 @@ export interface GridFSBucketWriteStreamOptions extends WriteConcernOptions {
    * @deprecated Will be removed in the next major version. Add an aliases field to the metadata document instead.
    */
   aliases?: string[];
-  /** @internal TODO(NODE-5688): make this public */
+  /** @public */
```
dariakp, line 45: should this be marked experimental?

src/sdam/topology.ts
```diff
@@ -457,17 +460,29 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
       }
     }
 
+    // TODO(NODE-6223): auto connect cannot use timeoutMS
```
dariakp, line 463: Anything we need to do here still? (NODE-6223 is closed)

src/sessions.ts
```diff
@@ -58,8 +59,11 @@ export interface ClientSessionOptions {
   snapshot?: boolean;
   /** The default TransactionOptions to use for transactions started on this session. */
   defaultTransactionOptions?: TransactionOptions;
-  /** @internal
-   * The value of timeoutMS used for CSOT. Used to override client timeoutMS */
+  /**
+   * @public
+   * An overriding timeoutMS value to use for a client-side timeout.
```
dariakp, line 64: Should this be marked experimental, too? (If so, we should mention it in our revisit experimental tags ticket)

test/unit/index.test.ts
```diff
@@ -153,12 +149,4 @@ describe('mongodb entrypoint', () => {
   it('exports only the expected keys', () => {
     expect(setDifference(Object.keys(mongodb), EXPECTED_EXPORTS)).to.be.empty;
   });
-
-  it('should export keys added by ts-node as undefined', () => {
```
dariakp, line 157: Did this test start failing? Did we update something about ts-node in just this PR?

test/unit/operations/get_more.test.ts
```diff
@@ -69,7 +69,7 @@ describe('GetMoreOperation', function () {
         const call = stub.getCall(0);
         expect(call.args[0]).to.equal(namespace);
         expect(call.args[1]).to.deep.equal(expectedGetMoreCommand);
-        expect(call.args[2]).to.deep.equal(opts);
+        expect(call.args[2]).to.containSubset(opts);
```
dariakp, line 72: this test was ensuring that the command is called with the expected options, changing this to a subset changes that guarantee - can we update it instead to the expected options?

dariakp (changes requested):

test/integration/sessions/sessions.prose.test.ts
dariakp: from slack: consistency when updating other tests

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
dariakp: Interesting, we should probably note that this is a spec test in some way

etc/notes/errors.md
dariakp: (Resolve when https://github.com/mongodb/node-mongodb-native/pull/4315/files merges)

test/unit/index.test.ts
nbbeeken: - https://github.com/mongodb/node-mongodb-native/pull/4292#discussion_r1826104349
- https://github.com/mongodb/node-mongodb-native/pull/4292#pullrequestreview-2408002618

test/unit/operations/get_more.test.ts
nbbeeken: cc: @aditi-khare-mongoDB

src/collection.ts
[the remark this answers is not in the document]
dariakp: Addressed in https://github.com/mongodb/node-mongodb-native/pull/4315/files

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
dariakp: Addressed in https://github.com/mongodb/node-mongodb-native/pull/4315/files

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
dariakp: Addressed in https://github.com/mongodb/node-mongodb-native/pull/4315/files

test/integration/client-side-operations-timeout/client_side_operations_timeout.spec.test.ts
dariakp: Moving discussion to https://github.com/mongodb/node-mongodb-native/pull/4315/files

src/client-side-encryption/auto_encrypter.ts
dariakp: resolved in https://github.com/mongodb/node-mongodb-native/pull/4315/files

src/gridfs/download.ts
dariakp: addressed in https://github.com/mongodb/node-mongodb-native/pull/4315/files

src/gridfs/upload.ts
dariakp: addressed in https://github.com/mongodb/node-mongodb-native/pull/4315/files

src/sdam/topology.ts
dariakp: addressed in https://github.com/mongodb/node-mongodb-native/pull/4315/files

src/sessions.ts
dariakp: addressed in https://github.com/mongodb/node-mongodb-native/pull/4315/files

test/unit/operations/get_more.test.ts
dariakp: addressed in https://github.com/mongodb/node-mongodb-native/pull/4315/files

nbbeeken (approved):

dariakp (commented):
LGTM pending perf comparison

dariakp (approved):

PR #4109, opened 2024-05-09
fix(NODE-6165): useBigInt64 causes compareTopologyVersion to throw

nbbeeken:
#### What is changing?

- compareTopologyVersion handles bigint inputs

No

BSON settings should not make driver internals throw

### Fixed error when useBigInt64=true was set on Db or MongoClient

Fixed an issue where when setting `useBigInt64`=`true` on MongoClients or Dbs an internal function `compareTopologyVersion` would throw an error when encountering a bigint value.


src/sdam/server_description.ts
```diff
@@ -258,10 +258,19 @@ export function compareTopologyVersion(
   }
 
   // TODO(NODE-2674): Preserve int64 sent from MongoDB
-  const currentCounter = Long.isLong(currentTv.counter)
-    ? currentTv.counter
-    : Long.fromNumber(currentTv.counter);
-  const newCounter = Long.isLong(newTv.counter) ? newTv.counter : Long.fromNumber(newTv.counter);
+  const currentCounter =
```
baileympearson, line 261: Instead of fixing it only here - should we instead fix it by adding coercion to TopologyVersion?  Right now the types are inaccurate, because TopologyVersion says counter is always a bigint.

nbbeeken: From slack, we have the follow-up `raw: true` issue that will take a closer look at BSON option handling. We want to prioritize unblocking `useBigInt64` usage.

baileympearson (approved):

PR #4110, opened 2024-05-14
fix(NODE-6171): RTT set to zero when serverMonitoringMode=stream

W-A-James:

No

NODE-6171

### Server Selection performance regression due to incorrect RTT measurement

Starting in version [6.6.0](https://github.com/mongodb/node-mongodb-native/commit/0e3d6ead735ed067bd044c8d0c9c307d970f1986#), when using the `stream` server monitoring mode, heartbeats were incorrectly timed as having a duration of 0, leading to server selection viewing each server as equally desirable for selection.

### DIFF
--- a/test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
+++ b/test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
@@ -1,10 +1,14 @@
-import { EventEmitter, once } from 'node:events';
 import { setTimeout } from 'node:timers/promises';
 
 import { expect } from 'chai';
 import * as sinon from 'sinon';
 
-import { Connection, type MongoClient, type ServerHeartbeatSucceededEvent } from '../../mongodb';
+import {
+  Connection,
+  type MongoClient,
+  promiseWithResolvers,
+  type ServerHeartbeatSucceededEvent
+} from '../../mongodb';
 import { loadSpecTests } from '../../spec';
 import { runUnifiedSuite } from '../../tools/unified-spec-runner/runner';
 
@@ -19,36 +23,12 @@ describe('SDAM Unified Tests (Node Driver)', function () {
 describe('Monitoring rtt tests', function () {
   let client: MongoClient;
   let heartbeatDurations: Record<string, number[]>;
-  const HEARTBEATS_TO_COLLECT = 200; // Wait for 200 total heartbeats. This is high enough to work for standalone, sharded and our typical 3-node replica set topology tests
-  const IGNORE_SIZE = 20;
+  const HEARTBEATS_TO_COLLECT_PER_NODE = 65;
+  const IGNORE_SIZE = 5;
   const DELAY_MS = 10;
-  let count: number;
-  const ee = new EventEmitter();
-
-  const listener = (ev: ServerHeartbeatSucceededEvent) => {
-    if (!client.topology.s.servers.has(ev.connectionId)) return;
-    count++;
-    if (count < IGNORE_SIZE) {
-      return;
-    }
-
-    heartbeatDurations[ev.connectionId].push(ev.duration);
-
-    // We ignore the first few heartbeats since the problem reported in NODE-6172 showed that the
-    // first few heartbeats were recorded properly
-    if (count === IGNORE_SIZE) {
-      return;
-    }
-
-    if (count >= HEARTBEATS_TO_COLLECT + IGNORE_SIZE) {
-      client.off('serverHeartbeatSucceeded', listener);
-      ee.emit('done');
-    }
-  };
 
   beforeEach(function () {
-    count = 0;
-    heartbeatDurations = {};
+    heartbeatDurations = Object.create(null);
   });
 
   afterEach(async function () {
@@ -67,7 +47,7 @@ describe('Monitoring rtt tests', function () {
             serverMonitoringMode
           });
 
-          // make send command delay for DELAY_MS ms to ensure that the actual time between sending
+          // make sendCommand delay for DELAY_MS ms to ensure that the actual time between sending
           // a heartbeat and receiving a response don't drop below 1ms. This is done since our
           // testing is colocated with its mongo deployment so network latency is very low
           const stub = sinon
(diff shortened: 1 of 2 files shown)


test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
```diff
@@ -8,3 +15,112 @@ describe('SDAM Unified Tests (Node Driver)', function () {
...
+        });
+
+        it(
+          'heartbeat duration is not incorrectly reported as zero on ServerHeartbeatSucceededEvents',
+          {
+            metadata: {
+              requires: { topology: '!load-balanced' }
+            },
+            test: async function () {
+              for (const server in heartbeatDurations) {
```
nbbeeken, line 98: Can you add an assertion that `heartbeatDurations` has a non-zero length? the test will pass if the for loop doesn't run (same for the other test as well)

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
```diff
@@ -8,3 +15,112 @@ describe('SDAM Unified Tests (Node Driver)', function () {
...
+  let client: MongoClient;
+  let heartbeatDurations: Record<string, number[]>;
+  const HEARTBEATS_TO_COLLECT = 200; // Wait for 200 total heartbeats. This is high enough to work for standalone, sharded and our typical 3-node replica set topology tests
+  const IGNORE_SIZE = 20;
+  const DELAY_MS = 10;
+  let count: number;
+  const ee = new EventEmitter();
+
+  const listener = (ev: ServerHeartbeatSucceededEvent) => {
+    if (!client.topology.s.servers.has(ev.connectionId)) return;
```
nbbeeken, line 29: When do we have connectionIds that aren't in the servers map?

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
```diff
@@ -8,3 +15,112 @@ describe('SDAM Unified Tests (Node Driver)', function () {
...
+    count++;
+    if (count < IGNORE_SIZE) {
+      return;
+    }
```
nbbeeken, line 33: Just suggestions to pair down the logic here 
- could you just collect all the HBs and then test on a slice of IGNORE_SIZE and up?
- could you inline the listener into the beforeEach and use a promiseWithResolvers to block until you have 200 HBs? then you shouldn't need a seperate ee.emit(done) communication channel

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
nbbeeken: Also could we use for-of `Object.entries/keys/values` here?

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
```diff
@@ -8,3 +15,112 @@ describe('SDAM Unified Tests (Node Driver)', function () {
...
+
+    heartbeatDurations[ev.connectionId].push(ev.duration);
+
```
nbbeeken, line 36:
```suggestion
    (heartbeatDurations[ev.connectionId] ??= []).push(ev.duration);
```

Not trying to code golf you here, but then you don't need to "setup" heartbeatDurations. This combined with removing the !has check may change the output, so I'm curious to first understand what connectionIds aren't in the server map.

nbbeeken (changes requested):

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
```diff
@@ -8,3 +15,112 @@ describe('SDAM Unified Tests (Node Driver)', function () {
...
+            requires: { topology: '!load-balanced' }
+          },
+          test: async function () {
+            for (const server in heartbeatDurations) {
+              const averageDuration =
+                heartbeatDurations[server].reduce((acc, x) => acc + x) /
+                heartbeatDurations[server].length;
+              expect(
+                client.topology.description.servers.get(server).roundTripTime
+              ).to.be.approximately(averageDuration, 1);
```
nbbeeken, line 119: Failing on Windows with:

> AssertionError: expected 13.1 to be close to 11.761194029850746 +/- 1

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
W-A-James: I can refactor the tests to collect a smaller number of heartbeats per server and that would enable me to do the slicing.

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
W-A-James: loosened the approximate assertion

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
W-A-James: That generally shouldn't be the case, but was just writing it defensively.

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
nbbeeken: Why do we have to change the total?

nbbeeken (approved):

test/integration/server-discovery-and-monitoring/server_discover_and_monitoring.test.ts
W-A-James: I had set the number of heartbeats to collect to be 200 total before to be a simple way to get all topologies we test on be able to collect decently large number of heartbeats. That meant that  previously the test was just assuming that each server in the topology would correct roughly the same number of heartbeats, whereas now, we're guaranteeing that we get 65 heartbeats per server, which for the repl set topology works out to 195 heartbeats total, but distributed evenly across each of the servers.

We don't necessarily _need_ to drop the number of heartbeats collected, but collecting 200 heartbeats per server didn't seem necessary.

PR #4131, opened 2024-06-04
refactor(NODE-6187): refactor to use TimeoutContext abstraction

W-A-James:


src/sdam/server.ts
```diff
@@ -273,7 +273,7 @@ export class Server extends TypedEventEmitter<ServerEvents> {
   public async command(
     ns: MongoDBNamespace,
     command: Document,
-    options?: CommandOptions
+    options: CommandOptions
```
W-A-James, line 276: there are no instances in the driver where we don't pass options into this method

src/sdam/topology.ts
```diff
@@ -459,12 +458,19 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
+      serverSelectionTimeoutMS,
+      waitQueueTimeoutMS: this.client.options.waitQueueTimeoutMS
+    });
+
     const selectServerOptions = {
       operationName: 'ping',
-      timeout,
-      ...options
+      ...options,
+      timeoutContext
```
W-A-James, line 473: Note: We construct a `TimeoutContext` here as we treat the ping command used for autoconnect as its own command.

src/sdam/topology.ts
```diff
@@ -563,24 +569,10 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
-        }
-      } else {
-        timeout = null;
-      }
-    } else {
-      timeout = Timeout.expires(serverSelectionTimeoutMS);
+    let timeout;
+    if (options.timeoutContext) timeout = options.timeoutContext.serverSelectionTimeout;
+    else {
+      timeout = Timeout.expires(options.serverSelectionTimeoutMS ?? 0);
```
W-A-James, line 575: This is (or should be) dead code

src/operations/operation.ts
```diff
@@ -67,6 +67,9 @@ export abstract class AbstractOperation<TResult = any> {
   /** @internal */
   timeoutMS?: number;
 
+  /** @internal */
+  timeoutContext!: TimeoutContext;
+
```
W-A-James, line 72: NOTE: This is marked with a non-null assertion because the field will be set during execute_operation and only referenced within that function in the following places at which point, if it is not defined, this would indicate a bug in the driver. I can add an inline comment explaining this

1. `Topology.selectServer`
2. `Server.command`
3. `ConnectionPool.checkOut`
4. `Connection.command`

src/timeout.ts
```diff
@@ -108,3 +108,131 @@ export class Timeout extends Promise<never> {
...
+/** @internal */
+export class CSOTTimeoutContext extends TimeoutContext {
+  timeoutMS: number;
+  serverSelectionTimeoutMS: number;
+  socketTimeoutMS?: number;
+
+  private _maxTimeMS?: number;
+
+  private _serverSelectionTimeout?: Timeout | null;
+  private _connectionCheckoutTimeout?: Timeout | null;
```
baileympearson, line 159: Just my thought - it might be simpler to never cache promises, at least until CSOT's implementation is complete and we can benchmark the difference between caching and not caching.  If caching a promise doesn't provide a noticable performance increase, then why add the complexity? (this also simplifies the interface, because now places that use timeouts can always clear the timeout.)

ex:
```typescript
override get serverSelectionTimeout(): Timeout | null {
    const timeout = this.serverSelectionTimeoutMS !== 0 &&
      csotMin(this.timeoutMS, this.serverSelectionTimeoutMS)

    return timeout ? Timeout.expires(timeout) : null;
}
```

as another minor modification, the Timeout class does nothing if the timeout parameter is 0, so we can actually always a return a timeout, right?
```typescript
override get serverSelectionTimeout(): Timeout {
    const timeout = this.serverSelectionTimeoutMS !== 0 &&
      csotMin(this.timeoutMS, this.serverSelectionTimeoutMS);

    return timeout ? Timeout.expires(timeout ?? 0);
}

// server selection, for example:
// unified code path, regardless of whether or not a timeout is being applied
try {
	timeout.throwIfExpired(); // <- as a question: is this really required? If the timeout has expired, wouldn't promise.race immediately reject with the timeout error?
	return Promise.race([serverPromise, timeout]);
} catch (error) {
	...
} finally {
  timeout.clear();
}
```

src/timeout.ts
```diff
@@ -108,3 +108,131 @@ export class Timeout extends Promise<never> {
...
+  }
+
+  static create(options: TimeoutContextOptions): TimeoutContext {
+    if (options.timeoutMS != null) return new CSOTTimeoutContext(options);
+    else return new LegacyTimeoutContext(options);
+  }
+
+  get serverSelectionTimeout(): Timeout | null {
+    return null;
+  }
```
baileympearson, line 139: If we're going to use an abstract class to define the interface, let's keep all logic on it to a minimum.  It's generally best practice for abstract classes to only have implementations that will be shared across all subclasses (i.e., if there's a method that can be written in terms of the class' abstract members, then you can put it in the abstract class instead of duplicating it in the children).


```suggestion
  abstract get serverSelectionTimeout(): Timeout | null;
  abstract get connectionCheckoutTimeout(): Timeout | null;
  abstract get csotEnabled(): this is CSOTTimeoutContext;
```

src/sdam/topology.ts
baileympearson: I think we can consider this more fully later, but wouldn't that mean that in the case that an operation auto-connects it's possible for the operation to take up to 2x timeoutMS?

src/sdam/topology.ts
baileympearson: which part in particular is dead code?  The else block?

test/unit/timeout.test.ts
```diff
@@ -115,3 +123,202 @@ describe('Timeout', function () {
...
+    });
+  });
+
+  describe('CSOTTimeoutContext', function () {
+    let ctx: CSOTTimeoutContext;
+
+    afterEach(() => {
+      // @ts-expect-error private property
+      ctx.overallTimeout?.clear();
+    });
```
baileympearson, line 159:
```suggestion
```

as far as I can tell, this property no longer exists on the CSOTTimeoutContext.  right?

test/unit/operations/get_more.test.ts
```diff
@@ -69,7 +69,10 @@ describe('GetMoreOperation', function () {
         const call = stub.getCall(0);
         expect(call.args[0]).to.equal(namespace);
         expect(call.args[1]).to.deep.equal(expectedGetMoreCommand);
-        expect(call.args[2]).to.deep.equal(opts);
+        for (const k in opts) {
```
baileympearson, line 72: Is it possible to use `.containSubset` here?

ex:
```typescript
expect(args[2]).to.containSubset(opts);

// or

const { timeoutContext, ...rest } = args[2];
expect(rest).to.deep.equal(opts);
```

src/cmap/connection_pool.ts
```diff
@@ -354,41 +354,17 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
    * will be held by the pool. This means that if a connection is checked out it MUST be checked back in or
    * explicitly destroyed by the new owner.
    */
-  async checkOut(options?: { timeout?: Timeout }): Promise<Connection> {
+  async checkOut(options?: { timeoutContext?: TimeoutContext }): Promise<Connection> {
     this.emitAndLog(
```
baileympearson, line 358: What issues does it cause to make both options and timeoutContext required?

If it's not too much of a hassle, I'd tackle it now - it feels like it defeats the purpose of having a legacy timeout context if we still need `Timeout.expires(this.options.waitQueueTimeoutMS);` here.

baileympearson (changes requested):
Looks great so far!

src/operations/execute_operation.ts
```diff
@@ -118,6 +118,14 @@ export async function executeOperation<
     );
   }
 
+  const timeoutContext = TimeoutContext.create({
```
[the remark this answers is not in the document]
W-A-James: My intention was that we construct exactly one `TimeoutContext` per operation with a lifetime equal to that of the operation. The first convergent place we have access to both the client and operation options would be in `executeOperation`, but the absolute first place is in each of our operation helper methods (insertOne, updateOne, etc).

I am open to updating our `AbstractOperation` constructor to accept the client options we'd need to construct the `TimeoutContext` and updating these methods since that would be cleaner and make clear that each operation "owns" its `TimeoutContext`.

Wouldn't constructing the `TimeoutContext` in `resolveOptions` lead to multiple `TimeoutContext` objects being constructed that likely wouldn't be used.

src/timeout.ts
W-A-James: Don't know how I forgot the abstract keyword doesn't just apply to the class itself 😅

src/cmap/connection_pool.ts
W-A-James: We can make the options required without issue, so I'll do that here.
For making `timeoutContext` required, we'd have to change `CommandOptions` to require `timeoutContext` or change the type of the `options` parameter for `Server.command` to use a different type than `Connection.command`. The latter would be a band-aid fix until we got to implementing the socket timeouts for CSOT.

W-A-James:
> I think we can consider this more fully later, but wouldn't that mean that in the case that an operation auto-connects it's possible for the operation to take up to 2x timeoutMS?

Yes, but more specifically it would be that it could take up to `client.timeoutMS + operation.timeoutMS`.

W-A-James:
> which part in particular is dead code? The else block?

On closer inspection, the only place where this isn't dead code is when change streams are selecting the server on encountering an error (see `ChangeStream._processErrorIteratorMode` and `ChangeStream._processErrorStreamMode`), since they don't have a `TimeoutContext` to pass in. Should we address this once we start working on change streams in NODE-5685 or do we want to clean this up here?

src/timeout.ts
W-A-James: Re the second comment, true, the Timeout constructor doesn't create an actual NodeJS.Timeout unless the duration passed in is greater than zero, so I can refactor that.

The reason I had done the caching approach was in anticipation of the socket timeout and maxTimeMS portions of the work. That's where I intend to introduce the overall timeout and use that for the the remaining sections of blocking work.

src/timeout.ts
```diff
@@ -108,3 +108,131 @@ export class Timeout extends Promise<never> {
...
   }
 }
+
+/** @internal */
+export type TimeoutContextOptions = {
+  timeoutMS?: number;
+  serverSelectionTimeoutMS: number;
+  waitQueueTimeoutMS: number;
+  socketTimeoutMS?: number;
+};
```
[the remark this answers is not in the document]
W-A-James: If we were to make `serverSelectionTimeoutMS` optional, would we want to return a null/0 timeout when requested for the `serverSelectionTimeout`?

src/operations/execute_operation.ts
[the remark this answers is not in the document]
baileympearson: > I am open to updating our AbstractOperation constructor to accept the client options we'd need to construct the TimeoutContext and updating these methods since that would be cleaner and make clear that each operation "owns" its TimeoutContext.

Yeah, you're right that resolveOptions would lead to instantiation of timeouts we don't need, specifically when we construct dbs and collections.  I'm not sure an operation owning a timeout context is the right choice either - I think that cursors / change streams will need to "own" their timeout context and they'll perform multiple operations inside the same context.  Maybe we can leave it as-is for now and reconsider once we've implemented CSOT for cursors.

src/sdam/topology.ts
baileympearson: I filed [NODE-6223](https://jira.mongodb.org/browse/NODE-6223) to handle auto connect, we can worry about what the correct behavior is later if that works for you?

baileympearson:
> On closer inspection, the only place where this isn't dead code is when change streams are selecting the server on encountering an error (see `ChangeStream._processErrorIteratorMode` and `ChangeStream._processErrorStreamMode`), since they don't have a `TimeoutContext` to pass in. Should we address this once we start working on change streams in [NODE-5685](https://jira.mongodb.org/browse/NODE-5685) or do we want to clean this up here?

Maybe address it later?  We don't actually provide a timeout to `selectServer` when resuming a change stream, I guess it would default to whatever was set on the client 🤔

src/timeout.ts
baileympearson: Were you intending to refactor the timeout context to always return timeouts (second point)?  Or did you intended to leave it as-is?

W-A-James: Looking more into the refactor, it would end up with us creating a lot more promises on hot paths than we'd need. For one, we'd create the 0 timeout, and then in both connection checkout and server selection, we'd end up creating a new promise for the race with the timeout, so I think it's better for us to just return null and eat the complexity which that requires in `TimeoutContext`

src/cmap/connection_pool.ts
```diff
@@ -424,7 +399,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
       }
       throw error;
     } finally {
-      if (timeout !== options?.timeout) timeout?.clear();
+      if (options?.timeoutContext?.clearConnectionCheckoutTimeout) timeout?.clear();
```
baileympearson, line 402:
```suggestion
      if (options.timeoutContext.clearConnectionCheckoutTimeout) timeout?.clear();
```

src/cmap/connection_pool.ts
```diff
@@ -415,7 +390,7 @@ export class ConnectionPool extends TypedEventEmitter<ConnectionPoolEvents> {
             : 'Timed out while checking out a connection from connection pool',
           this.address
         );
-        if (options?.timeout) {
+        if (options?.timeoutContext?.csotEnabled()) {
```
baileympearson, line 393:
```suggestion
        if (options.timeoutContext.csotEnabled()) {
```

src/operations/execute_operation.ts
```diff
@@ -118,6 +118,12 @@ export async function executeOperation<
     );
   }
 
+  timeoutContext ??= TimeoutContext.create({
```
baileympearson, line 121: Shouldn't we be constructing the timeout with the `options` from the operation?  Otherwise, `timeoutMS` and other options won't be properly inherited from the client -> db -> collection -> operation.

src/sdam/server.ts
```diff
@@ -266,20 +272,20 @@ export class Server extends TypedEventEmitter<ServerEvents> {
   public async command<T extends MongoDBResponseConstructor>(
     ns: MongoDBNamespace,
     command: Document,
-    options: CommandOptions | undefined,
+    options: ServerCommandOptions | undefined,
```
baileympearson, line 275:
```suggestion
    options: ServerCommandOptions,
```

src/sdam/topology.ts
```diff
@@ -459,12 +461,19 @@ export class Topology extends TypedEventEmitter<TopologyEvents> {
...
     const timeoutMS = this.client.options.timeoutMS;
-    const timeout = timeoutMS != null ? Timeout.expires(timeoutMS) : undefined;
+    const serverSelectionTimeoutMS = this.client.options.serverSelectionTimeoutMS;
     const readPreference = options.readPreference ?? ReadPreference.primary;
+
+    const timeoutContext = TimeoutContext.create({
+      timeoutMS,
+      serverSelectionTimeoutMS,
+      waitQueueTimeoutMS: this.client.options.waitQueueTimeoutMS
+    });
```
baileympearson, line 471:
```suggestion
    const timeoutContext = TimeoutContext.create(client.s.options);
```

and while we're here - can we also adjust these `client.options` calls to use `client.s.options`?

baileympearson (changes requested):
just a few small comments

src/operations/execute_operation.ts
W-A-James: timeoutMS is the only option here that we get from the operation. `waitQueueTimeoutMS` and `serverSelectionTimeoutMS` are strictly client options that can't be set at the database, collection or operation level

test/unit/timeout.test.ts
```diff
@@ -115,3 +123,197 @@ describe('Timeout', function () {
     });
   });
 });
+
+describe('TimeoutContext', function () {
+  describe('TimeoutContext.create', function () {
+    describe('when timeoutMS is a number', function () {
```
durran, line 129: nit: would you mind changing all the `describe` blocks with "when" in their descriptions to `context` to match our testing guidelines? If not I can change them in mine if it's a hassle.

durran (changes requested):

src/timeout.ts
```diff
@@ -108,3 +108,165 @@ export class Timeout extends Promise<never> {
...
+
+  set maxTimeMS(v: number) {
+    this._maxTimeMS = v;
+  }
+
+  csotEnabled(): this is CSOTTimeoutContext {
+    return true;
+  }
+
+  override get serverSelectionTimeout(): Timeout | null {
```
durran, line 208: The `override` keywords in the subclasses are not necessary as the parent class defines them as abstract.

durran (approved):

baileympearson (approved):

baileympearson (approved):

PR #4133, opened 2024-06-05
refactor: replace symbol properties in cursor classes

nbbeeken:
#### What is changing?

- `AbstractCursor`
  - Removed all symbol properties in favor of equivalently named strings.
  - Changed "options" to cursorOptions, no need for the getter indirection
  - Added `is` in front of booleans that have getters
- `FindCursor`
  - kFilter -> cursorFilter since there is a "filter" method
  - builtOptions -> findOptions, I think the naming came from TS translation days, it is true that the options are built from all the helpers (if one uses those) but in the end they are options for find
- `AggregationCursor`
  - pipline is a public readonly property which is equivalent to the getter that fetched the symbol property before
- `ChangeStreamCursor`
  - No symbol properties here, but the "options" property shared the same name as AbstractCursor's "options" until I changed it to "cursorOptions". 
  - But that led me to take a look at the class properties, we never set resumeAfter/startAfter on the class, thos remain on the options, unless I missed something.

[patch](https://spruce.mongodb.com/version/6660be9352ce9e00070984f5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

No.

Symbol properties served as "private" before we had typescript. We prefer using plain string properties now for better DX, and testing internals. 

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -532,7 +532,7 @@ export abstract class AbstractCursor<
     const oldTransform = this[kTransform] as (doc: TSchema) => TSchema; // TODO(NODE-3283): Improve transform typing
     if (oldTransform) {
       this[kTransform] = doc => {
-        return transform(oldTransform(doc));
+        return this.makeSafeTransform(transform)(oldTransform(doc));
       };
     } else {
       this[kTransform] = transform;
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -532,7 +532,7 @@ export abstract class AbstractCursor<
     const oldTransform = this[kTransform] as (doc: TSchema) => TSchema; // TODO(NODE-3283): Improve transform typing
     if (oldTransform) {
       this[kTransform] = doc => {
-        return this.makeSafeTransform(transform)(oldTransform(doc));
+        return transform(oldTransform(doc));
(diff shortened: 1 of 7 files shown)


nbbeeken:
evergreen patch

src/cursor/abstract_cursor.ts
```diff
@@ -648,15 +620,29 @@ export abstract class AbstractCursor<
...
+    if (this.cursorId == null) {
+      throw new MongoRuntimeError(
+        'Unexpected null cursor id. A cursor creating command should have set this'
+      );
+    }
+    if (this.selectedServer == null) {
+      throw new MongoRuntimeError(
+        'Unexpected null selectedServer. A cursor creating command should have set this'
+      );
+    }
```
W-A-James, line 632: Is this replacing the non-null assertion?

W-A-James (approved):

src/cursor/abstract_cursor.ts
nbbeeken: Yes, happy to revert this, (or make sure we have coverage?). My thinking is that if this somehow was null and we pass it down the code below trusts it always exists but it doesn't necessarily mean the code below will use it in a way that's crashing, like dot access, it could go all the way to the server and we get "cursor.id cannot be null"

W-A-James (approved):

PR #4138, opened 2024-06-07
refactor: move CountDocument logic into collection API

nbbeeken:
#### What is changing?

- Delete CountDocument operation class
- Move logic into collection.countDocuments

The subclass is unneccessary and make typing difficult. See: https://github.com/mongodb/node-mongodb-native/pull/4112#discussion_r1630101527


src/collection.ts
```diff
@@ -764,10 +771,23 @@ export class Collection<TSchema extends Document = Document> {
...
+    if (typeof options.limit === 'number') {
+      pipeline.push({ $limit: options.limit });
+    }
+
+    pipeline.push({ $group: { _id: 1, n: { $sum: 1 } } });
+
+    const cursor = this.aggregate<{ n: number }>(pipeline, options);
+    const doc = await cursor.next();
+    await cursor.close();
+    return doc?.n ?? 0;
```
baileympearson, line 790:
```suggestion
    try {
      const doc = await cursor.next();
      return doc?.n ?? 0;
    } finally {
       await cursor.close().catch(error => squashError(error);
    }
```

Do you think we need to be careful about closing the cursor?  I'm thinking it can't hurt 🤷

nbbeeken: I followed how findOne was written, I recall there being a reason but I can't recall and I can't say I disagree so sure 🤔

nbbeeken: I recall now, the expectation is that our cursor's next cleans itself up if it throws and we also did not want to squash errors from close because we expect it not to throw. The thinking is both of these expectations are part of the API, if they aren't satisfied then something is broken and if we squash / use finally we are covering up a problem that needs fixing. 

Added testing!

test/integration/crud/crud_api.test.ts
```diff
@@ -105,9 +105,9 @@ describe('CRUD API', function () {
         const spy = sinon.spy(Collection.prototype, 'find');
         const result = await collection.findOne({});
         expect(result).to.deep.equal({ _id: 1 });
-        expect(events.at(0)).to.be.instanceOf(CommandSucceededEvent);
-        expect(spy.returnValues.at(0)).to.have.property('closed', true);
-        expect(spy.returnValues.at(0)).to.have.nested.property('session.hasEnded', true);
+        expect(events[0]).to.be.instanceOf(CommandSucceededEvent);
+        expect(spy.returnValues[0]).to.have.property('closed', true);
+        expect(spy.returnValues[0]).to.have.nested.property('session.hasEnded', true);
```
nbbeeken, line 110: I could have sworn we had to go back and remove `.at` usage... did this get backported to v16? or am I remembering when we supported node v14?

baileympearson: We broke this on v14.  Array.at is supported on Node16

baileympearson (approved):

PR #4141, opened 2024-06-11
feat(NODE-5908): support range v2

baileympearson:
#### What is changing?

This PR adds support for the QE range v2 protocol.  This is a breaking change to users of QE range v1 but our support is experimental and subject to change, so this is okay.


nbbeeken:
NODE-5908 From standup: We want to have an official release of the bindings before merging this so we can lift the experimental tag and land the final API at the same time

.evergreen/config.in.yml
```diff
@@ -13,6 +13,11 @@ command_type: system
 # That roughly accounts for variable system performance for various buildvariants
 exec_timeout_secs: 2100
 
+patch_aliases:
```
baileympearson, line 16: Happy to revert.  I was thinking that once this PR lands, we can set up a nightly job that runs using this alias.

I _think_ this works but I haven't tested it yet.

.evergreen/config.in.yml
```diff
@@ -272,10 +277,31 @@ functions:
     - command: shell.exec
       type: test
       params:
+        shell: bash
         working_dir: src
         timeout_secs: 300
         script: |
           ${PREPARE_SHELL}
+          # Disable xtrace (just in case it was accidentally set).
```
baileympearson, line 285: turns out we didn't actually run FLE tests in load balanced mode because `CSFLE_KMS_PROVIDERS` was not set

test/tools/runner/filters/client_encryption_filter.ts
```diff
@@ -67,15 +74,21 @@ export class ClientSideEncryptionFilter extends Filter {
     }
 
     // TODO(NODE-3401): unskip csfle tests on windows
-    if (process.env.TEST_CSFLE && !this.enabled && process.platform !== 'win32') {
-      throw new Error('Expected CSFLE to be enabled in the CI');
+    if (process.env.TEST_CSFLE && process.platform !== 'win32') {
```
baileympearson, line 77: The logic is the same, it's just refactored to have descriptive skip reasons.

test/tools/runner/hooks/configuration.ts
```diff
@@ -81,6 +81,12 @@ const testSkipBeforeEachHook = async function () {
   if (metadata && metadata.requires && Object.keys(metadata.requires).length > 0) {
     const failedFilter = filters.find(filter => filter.filter(this.currentTest) !== true);
     if (failedFilter) {
+      const maybeSkipReason = failedFilter.filter(this.currentTest);
```
baileympearson, line 84: Our `Filter` class supports returning skip reasons, but we ignored them.

test/tools/spec-runner/index.js
```diff
@@ -187,9 +187,12 @@ function generateTopologyTests(testSuites, testContext, filter) {
         shouldRun = false;
       }
 
-      if (shouldRun && typeof filter === 'function' && !filter(spec, this.configuration)) {
-        this.currentTest.skipReason = `filtered by custom filter passed to generateTopologyTests`;
-        shouldRun = false;
+      if (shouldRun && typeof filter === 'function') {
+        const result = filter(spec, this.configuration);
```
baileympearson, line 191: Minor improvement to allow for skip reasons in predicates provided to `generateTopologyTests`.

.evergreen/config.in.yml
aditi-khare-mongoDB: Could we move this to a separate PR? Since supporting range v2 will affect our public API, I think it's best to have a clean commit users can look at. Also, I think it would make the scope of this ticket an easier task while reviewing.

src/client-side-encryption/client_encryption.ts
```diff
@@ -957,42 +957,3 @@ export interface RangeOptions {
   sparsity: Long;
   precision?: number;
 }
-
-/**
```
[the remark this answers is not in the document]
baileympearson: No reason, just an oversight.

.evergreen/generate_evergreen_tasks.js
```diff
@@ -579,33 +579,54 @@ BUILD_VARIANTS.push({
...
+        vars: {
+          'LIBMONGOCRYPT_VERSION': 'latest'
+        }
+      },
+      { func: 'run custom csfle tests' }
+    ]
+  }
+];
 
-const FLE_PINNED_COMMIT = '974a4614f8c1c3786e5e39fa63568d83f4f69ebd';
```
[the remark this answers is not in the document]
baileympearson: It's no longer needed.  Previously we tested:

- with the latest version of mongodb-client-encryption (the latest version we'd released)
- the pinned commit, which was the furthest into `master` in libmongocrypt that we'd developed
- libmongocrypt master, so we were alerted of changes in libmongocrypt that broke our bindings early


Now, this isn't necessary.  We test

- with the latest version of mongodb-client-encryption (the latest version we'd released)
- `master` of mongodb-client-encryption, which is the furthest we've developed
- with libmongocrypt `master`,  so we were alerted of changes in libmongocrypt that broke our bindings early


the coverage is the same, we're just getting it in different tests.

.evergreen/config.in.yml
baileympearson: I can remove the patch alias, but we will need the other CI changes in this PR.

test/tools/runner/hooks/configuration.ts
nbbeeken: Not a big deal but the check above has already run the filter, could we preserve the result from that? I just think it could catch us by surprise in the future if a filter runs twice on the same input.

test/integration/node-specific/client_encryption.test.ts
```diff
@@ -406,10 +416,10 @@ describe('ClientEncryption integration tests', function () {
     });
 
     it(
-      `does not throw if algorithm has different casing than 'rangePreview'`,
+      `does not throw if algorithm has different casing than 'range'`,
```
nbbeeken, line 419: interesting that this test is "does not throw" rather than "works with any casing" I don't think we have to change something here but I would test for success rather than "not error", what if it ignores this option, does this throw or is it nonsensical usage?

test/integration/client-side-encryption/client_side_encryption.prose.22.range_explicit_encryption.test.ts
```diff
@@ -202,6 +211,11 @@ describe('Range Explicit Encryption', function () {
           keyVaultNamespace: 'keyvault.datakeys',
           kmsProviders: getKmsProviders()
         };
+        console.error({
```
nbbeeken, line 214: debug cruft?

.evergreen/install-mongodb-client-encryption.sh
```diff
@@ -0,0 +1,33 @@
+#! /usr/bin/env bash
+set +o xtrace # Do not write AWS credentials to stderr
+
+# Initiail checks for running these tests
+if [ -z ${INSTALL_DIR+omitted} ]; then echo "INSTALL_DIR is unset" && exit 1; fi
```
nbbeeken, line 5:
```suggestion
# Initial checks for running these tests
```

nbbeeken, line 5: Don't mind this. but could we just leave it as `mongodb-client-encryption`? When is `INSTALL_DIR` an alternate name, and why?

.evergreen/install-mongodb-client-encryption.sh
```diff
@@ -0,0 +1,33 @@
...
+source "${PROJECT_DIRECTORY}/.evergreen/init-node-and-npm-env.sh"
+
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+CWD=$(pwd)
+
+rm -rf $INSTALL_DIR
+git clone https://github.com/mongodb-js/mongodb-client-encryption.git $INSTALL_DIR
+cd $INSTALL_DIR
```
nbbeeken, line 17:
```suggestion
pushd $INSTALL_DIR
```

.evergreen/install-mongodb-client-encryption.sh
```diff
@@ -0,0 +1,33 @@
...
+    npm run install:libmongocrypt -- --libVersion $LIBMONGOCRYPT_VERSION
+else
+	# otherwise use whatever is specified in the package.json.
+    npm run install:libmongocrypt
+fi
+
+echo "finished installing libmongocrypt"
+BINDINGS_DIR=$(pwd)
+
+cd $CWD
```
nbbeeken, line 30:
```suggestion
popd
```

.evergreen/install-mongodb-client-encryption.sh
```diff
@@ -0,0 +1,33 @@
...
+
+echo "linking mongodb-client-encrytion"
+npm link $BINDINGS_DIR
```
nbbeeken, line 33: related to the install dir comment this would always just be `npm link ./mongodb-client-encryption` (or expand . to an abspath)

nbbeeken (changes requested):
pretty much LGTM, a couple of small improvements that we can also pass over, lmk if you agree

.evergreen/install-mongodb-client-encryption.sh
baileympearson: What change are you suggesting here?  Removing the guard on INSTALL_DIR?

test/tools/runner/hooks/configuration.ts
baileympearson: you might think that something like this would work:
```
    const failedFilter = filters.filter(filter => filter.filter(this.currentTest) !== true)[0];
```
but it turns out the default print statement prints which filter failed by using the constructor's name (`else` block).  I think this is useful so I decided to preserve it

You might then think we could do something like this:
```
    const failedFilter = filters.find(filter => {
       const result = filter.filter(this.currentTest);
       if (result) return { result, filter };
       return null
    })[0];
```
but that seemed less nice than just filtering again.  So I chose to just run the filter again, especially because the filters are designed to be idempotent.

.evergreen/install-mongodb-client-encryption.sh
nbbeeken: Oh this landed on the wrong line, did this review from vscode, `Initiail` -> `Initial`

test/tools/runner/hooks/configuration.ts
baileympearson: a better solution is to just ensure that filters return strings, not booleans, and then we can make the change you suggested because every failing filter will return a skip reason.

nbbeeken (approved):
🚀

test/tools/runner/hooks/configuration.ts
nbbeeken: Whatever happened to a good ol' for loop? 🙂 

```js
let failedFilter: Filter;
let shouldSkip: boolean | string;
for (const filter of filters) {
  shouldSkip = filter.filter(this.currentTest);
  if (shouldSkip) {
    failedFilter = filter;
    break;
  }
}
```

Or you can leave the `find` as is and assign `shouldSkip` in the closure.

1000 ways to reach the same goal, I think it would be nice to not double call a filter, but it causes no bugs so that works too if you wanna make each one return its name.

.evergreen/generate_evergreen_tasks.js
```diff
@@ -758,7 +754,8 @@ fileData.tasks = (fileData.tasks || [])
   .concat(SINGLETON_TASKS)
   .concat(AUTH_DISABLED_TASKS)
   .concat(AWS_LAMBDA_HANDLER_TASKS)
-  .concat(MONGOCRYPTD_CSFLE_TASKS);
+  .concat(MONGOCRYPTD_CSFLE_TASKS)
+  .concat(NIGHTLY_TASKS);
```
nbbeeken, line 758: NIGHTLY_TASKS is not defined

nbbeeken (changes requested):

nbbeeken (approved):

aditi-khare-mongoDB (approved):
LGTM! Moving to team review

PR #4149, opened 2024-06-13
test(NODE-6202): use single mongos in unacknowledged write tests

durran:
#### What is changing?

Syncs unacknowledged write CLAM spec tests per DRIVERS-2921/DRIVERS-2877 and implements `ignoreExtraMessages` in the unified runner.

None

NODE-6202 / NODE-6061


test/spec/command-logging-and-monitoring/logging/unacknowledged-write.json
durran, line 1: This incorporates the spec test changes from both [DRIVERS-2921](https://jira.mongodb.org/browse/DRIVERS-2921) and [DRIVERS-2877](https://jira.mongodb.org/browse/DRIVERS-2877)

test/tools/unified-spec-runner/match.ts
```diff
@@ -241,13 +241,13 @@ export function resultCheck(
       }
       for (const [index, value] of expectedEntries) {
         path.push(`[${index}]`);
-        checkNestedDocuments(index, value, false);
+        checkNestedDocuments(index, value, checkExtraKeys);
```
durran, line 244: For these we pass the `checkExtraKeys` value through all the way - no more hardcoded assumptions.

test/tools/unified-spec-runner/match.ts
```diff
@@ -675,11 +675,16 @@ export function filterExtraLogs(
 export function compareLogs(
   expected: ExpectedLogMessage[],
   actual: ExpectedLogMessage[],
-  entities: EntitiesMap
+  entities: EntitiesMap,
+  ignoreExtraMessages = false
 ): void {
-  expect(actual).to.have.lengthOf(expected.length);
+  if (!ignoreExtraMessages) {
```
durran, line 681: `ignoreExtraMessages` was in the schema previously, but now we have tests that use it so we ignore any extra messages after the expected ones from the unified suite.

test/unit/tools/unified_spec_runner.test.ts
```diff
@@ -37,7 +37,7 @@ describe('Unified Spec Runner', function () {
           it('throws AssertionError when it finds extra keys', function () {
             actual =
               '{"data": {"$numberLong": "100"}, "a": {"$numberInt": "10"}, "b": {"$numberInt": "100"}}';
-            expect(() => resultCheckSpy(actual, expected, entitiesMap, [])).to.throw(
+            expect(() => resultCheckSpy(actual, expected, entitiesMap, [], true)).to.throw(
```
aditi-khare-mongoDB, line 40: I'm surprised this test was previously passing since `resultCheck`'s `checkExtraKeys` variable defaulted to `false`. Can we add a test asserting the inverse of this one?

aditi-khare-mongoDB (changes requested):

test/tools/unified-spec-runner/match.ts
durran: I've renamed the method as asked and added comments to the 2 properties directly from the spec so it's clear on what their purpose is. As for the filtering, I left it as it just before the call to `compareLogs` to leave the method signature there as is - seemed a bit clearer to understand.

test/unit/tools/unified_spec_runner.test.ts
durran: I've added another test here for the inverse.

test/unit/tools/unified_spec_runner.test.ts
```diff
@@ -37,12 +37,18 @@ describe('Unified Spec Runner', function () {
...
               AssertionError,
               /object has more keys than expected/
             );
           });
 
+          it('does not throw when it finds extra keys but checking is disabled', function () {
+            actual =
+              '{"data": {"$numberLong": "100"}, "a": {"$numberInt": "10"}, "b": {"$numberInt": "100"}}';
+            resultCheckSpy(actual, expected, entitiesMap, [], false);
```
aditi-khare-mongoDB, line 49:
```suggestion
            expect(() => resultCheckSpy(actual, expected, entitiesMap, [], false)).to.not.throw;
```

test/tools/unified-spec-runner/match.ts
aditi-khare-mongoDB: Looks good, thanks!

aditi-khare-mongoDB (changes requested):

test/unit/tools/unified_spec_runner.test.ts
nbbeeken: Is it not already the case that a test will fail if a function throws?

aditi-khare-mongoDB: It will but it's best practice in my opinion to wrap and assert the exception, rather than let it potentially propagate through.

durran: I'm fine with being explicit about the assertion given the method returns void. I've updated.

aditi-khare-mongoDB (approved):

PR #4155, opened 2024-06-17
fix(NODE-5720): on pre-4.4 sharded servers, the node driver uses `error.writeConcern.code` to determine retryability

aditi-khare-mongoDB:
### Description
On pre-4.4 `Mongos`servers, no longer add a `RetryableWriteLabel` to `WriteConcernError` due to its `code` property.

#### What is changing?
Sync some retryable write spec tests and adhere to spec guidelines.

No.

Spec compliance.

### On pre-4.4 sharded servers, the node driver uses the 'writeConcernError.code` field determine retryability
This fix was made to comply with [spec guidelines](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.md#retryablewriteerror-labels).


aditi-khare-mongoDB:
Passing on all relevant CI configurations: [see here.](https://spruce.mongodb.com/version/6670a443863b1a000783bb6d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -63,16 +79,17 @@ describe('Legacy Retryable Writes Specs', function () {
...
       for (const spec of suite.tests) {
-        // Step 2: Run the test
-        const mochaTest = it(spec.description, async () => await executeScenarioTest(spec, ctx));
-
-        // A pattern we don't need to repeat for unified tests
-        // In order to give the beforeEach hook access to the
-        // spec test so it can be responsible for skipping it
-        // and executeScenarioSetup
-        mochaTest.spec = spec;
+        if (!LEGACY_SKIP_TESTS_4_4_SHARDED.includes(spec.description)) {
```
aditi-khare-mongoDB, line 83: Unsure how to detect server type and version within this file (the `Context` given is empty and so is `this.configuration`). Because of this, I've skipped all of the legacy tests rather than only `serverVersion <= 4.4`  and `topologyType === 'Sharded'`.

Let me know your thoughts and suggestions!

baileympearson: See my comment in `error.ts`, we shouldn't have to skip tests.  

But in general - `this.configuration` is only accessible in `before`/`after`/`test` functions, not in the `context`/`describe` blocks.  If you put skip logic into a `beforeEach`, you should be able to skip based on server version and topology.

baileympearson (changes requested):

src/error.ts
```diff
@@ -1237,7 +1241,9 @@ export function needsRetryableWriteLabel(error: Error, maxWireVersion: number):
   }
 
   if (error instanceof MongoWriteConcernError) {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good! Thanks for catching my mistake. The tests are still running but they should pass - based on this[ evergreen patch ](https://spruce.mongodb.com/version/66831d8f30f5a40007135722/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)I ran.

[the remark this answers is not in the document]
aditi-khare-mongoDB: Update, the tests were wrong, and I've updated them according to this [spec commit](https://github.com/mongodb/specifications/commit/2c080e879cc54c386a382b33020bcc8154682acf).

src/error.ts
```diff
@@ -1237,11 +1242,13 @@ export function needsRetryableWriteLabel(error: Error, maxWireVersion: number):
...
-    return RETRYABLE_WRITE_ERROR_CODES.has(error.result?.code ?? error.code ?? 0);
+    return serverType === 'Mongos' && maxWireVersion < 9
```
baileympearson, line 1245: I don't think this is right still - the spec says that for <4.4 mongoses, we _do_ check the top level code, we just don't check writeConcernError.code.  As written, I think that this will always return false if a `writeConcernError` is returned with a top-level retryable error code, when instead we should be hitting the if-statement on line 1250.

Right?

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -40,6 +57,14 @@ describe('Legacy Retryable Writes Specs', function () {
 
         await utilClient.close();
 
+        if (
+          this.configuration.topologyType === 'Sharded' &&
+          lt(this.configuration.version, '4.4.0')
+        ) {
```
baileympearson, line 63: We also need to check the description, right? we don't want to skip all tests on <4.4 mongoses.

Also - we usually attach a skip reason:

```typescript
this.currentTest.skipReason = ...
this.skip();
```

src/error.ts
baileympearson: Right, I'm saying that I think it's supposed to reach line 1250 but for a pre-4.4 mongos writeConcernError, we will always return `false` here.

src/error.ts
```diff
@@ -1246,11 +1253,16 @@ export function needsRetryableWriteLabel(error: Error, maxWireVersion: number):
...
+        error.result.writeConcernError.code ?? Number(error.code) ?? 0
+      );
+    }
+    return RETRYABLE_WRITE_ERROR_CODES.has(Number(error.originalCode) ?? 0);
   }
 
-  if (error instanceof MongoError && typeof error.code === 'number') {
-    return RETRYABLE_WRITE_ERROR_CODES.has(error.code);
+  if (error instanceof MongoError) {
+    return RETRYABLE_WRITE_ERROR_CODES.has(Number(error.code));
```
baileympearson, line 1265: Is the coercion to number here just because it seems correct or do tests fail without it? I agree this seems correct but just curious.

src/error.ts
```diff
@@ -1246,11 +1253,16 @@ export function needsRetryableWriteLabel(error: Error, maxWireVersion: number):
   }
 
   if (error instanceof MongoWriteConcernError) {
-    return RETRYABLE_WRITE_ERROR_CODES.has(error.result.writeConcernError.code ?? error?.code ?? 0);
+    if (!(serverType === 'Mongos' && maxWireVersion < 9)) {
+      return RETRYABLE_WRITE_ERROR_CODES.has(
+        error.result.writeConcernError.code ?? Number(error.code) ?? 0
+      );
+    }
+    return RETRYABLE_WRITE_ERROR_CODES.has(Number(error.originalCode) ?? 0);
```
baileympearson, line 1261:
```suggestion
    return RETRYABLE_WRITE_ERROR_CODES.has(Number(error.result.code) ?? 0);
```

just a thought since the top-level code is already available on the result document.  

If you'd prefer to leave `originalCode` on the write concern error, can we add the originalCode property to the write concern error class as an internal field?  This works right now because MongoError has `[key: string]: any` in it, so it allows any keys.  But if we expect to use `originalError`, we should explicitly add it to our error type.

baileympearson (changes requested):
Looks great! two small comments but otherwise LGTM.

src/error.ts
aditi-khare-mongoDB: The coercion to number is because error.code can be a string or a number as define on `MongoError`.

src/error.ts
aditi-khare-mongoDB: I see, I'm good with using `error.result.code` here instead of adding a property explicitly.

src/error.ts
aditi-khare-mongoDB: Bookkeeping, we are now checking only the top-level code (not nested code) for 4.2 sharded clusters

baileympearson (approved):

PR #4157, opened 2024-06-21
refactor(NODE-6230): executeOperation to use iterative retry mechanism

W-A-James:
#### What is changing?

* Refactor `executeOperation` to use iterative retry mechanism
* Replaced `retryOperation` helper function with `executeOperationWithRetry` helper function
* Changed `AbstractOperation`'s `canRetryWrite` and `canRetryRead` getter methods to use opeation aspects as source of truth for retryability instead of defaulting to `true` for both (note that these values are being relied upon in `executeOperation` in a different manner than prior to the refactor so the inconsistency that existed prior did not affect correctness before, but had to be updated to ensure correctness here)
* Changed `CommandOperation`'s `canRetryWrite` getter to fallback on the parent class' value of `canRetryWrite`

* No

NODE-6230


nbbeeken (approved):
LGTM

PR #4161, opened 2024-06-27
fix(NODE-6242): close becomes true after calling close when documents still remain

nbbeeken:
#### What is changing?

Fix a change in behavior to close introduced in a recent cursor refactor. The expectation was that close became true when no more documents remain locally.

Yes, API docs updated.

The closed flag is used to determine if there are docs remaing in mongosh

### Clarified cursor state properties

The cursor has a few properties that represent the current state from the perspective of the driver and server. This PR corrects an issue that never made it to a release but we would like to take the opportunity to re-highlight what each of these properties mean.

- `cursor.closed` - `cursor.close()` has been called, and there are no more documents stored in the cursor.
- `cursor.killed` - `cursor.close()` was called while the cursor still had a non-zero id, and the driver sent a killCursors command to free server-side resources
- `cursor.id == null` - The cursor has yet to send it's first command (ex. `find`, `aggregate`)
- `cursor.id.isZero()` - The server sent the driver a cursor id of `0` indicating a cursor no longer exists on the server side because all data has been returned to the driver.
- `cursor.bufferedCount()` - The amount of documents stored locally in the cursor. 

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -209,6 +209,17 @@ export abstract class AbstractCursor<
     }
   }
 
+  /**
+   * The cursor has no id until it receives a response from the initial cursor creating command.
+   *
+   * It is non-zero for as long as the database has an open cursor.
+   *
+   * The initiating command may receive a zero id if the entire result is in the `firstBatch`.
+   *
+   * The id value may still be non-zero after calling `cursor.close()`
+   * since the value is only updated when the server returns a new id value,
+   * and a cursor may be closed before iteration is complete.
+   */
   get id(): Long | undefined {
     return this.cursorId ?? undefined;
   }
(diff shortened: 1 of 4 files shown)


src/cursor/abstract_cursor.ts
```diff
@@ -209,6 +209,17 @@ export abstract class AbstractCursor<
...
   }
 
+  /**
+   * The cursor has no id until it receives a response from the initial cursor creating command.
+   *
+   * It is non-zero for as long as the database has an open cursor.
+   *
+   * The initiating command may receive a zero id if the entire result is in the `firstBatch`.
+   *
+   * The id value may still be non-zero after calling `cursor.close()`
```
baileympearson, line 219: This is another regression - we previously zero'ed out the cursor id when closing it:

```
if (cursorId == null || server == null || cursorId.isZero() || cursorNs == null) {
    if (needsToEmitClosed) {
      cursor[kClosed] = true;
      cursor[kId] = Long.ZERO;  <-- zero out id
      cursor.emit(AbstractCursor.CLOSE);
    }

    if (session) {
      if (session.owner === cursor) {
        await session.endSession({ error });
        return;
      }

      if (!session.inTransaction()) {
        maybeClearPinnedConnection(session, { error });
      }
    }

    return;
  }

```

nbbeeken: Good catch, fixed!

durran (approved):

PR #4167, opened 2024-07-02
fix(NODE-6249): prevent make invocation on falsy responseType

fredmaggiowski:
### Description

By [NODE-6249](https://jira.mongodb.org/browse/NODE-6249) we are receiving a `TypeError` in the `changeStream.on('error') handler`:

```sh
TypeError: responseType ?? responses_1.MongoDBResponse).make is not a function
```

This `TypeError` can occur either because:

* `responseType` is _falsy_ (but not _nullish_); or
* an unexpected `responseType` that doesn't fit the `MongoDBResponseConstructor` type is received

#### What is changing?

This PR proposes a change that makes the `(responseType/MongoDBResponse).make()` more robust, by falling-back to `MongoDBResponse` on all the falsy values rather than only the nullish ones.

don't think so

As said above, we are receiving a TypeError that feels a lot like a programming error; this may help mitigate the issue; I couldn't find a way (if given some help/insight I can try and further investigate) to understand the actual value for responseType.

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -449,7 +449,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
         this.socket.setTimeout(0);
         const bson = response.parse();
 
-        const document = (responseType ?? MongoDBResponse).make(bson);
+        const document = (responseType || MongoDBResponse).make(bson);
 
         yield document;
         this.throwIfAborted();


fredmaggiowski:
I found out the issue was related to an incompatibility between mongodb driver `6.8.0` and `@opentelemetry/instrumentation-mongodb <0.46.0`.

Updating the latter, the bug stopped happening

PR #4171, opened 2024-07-09
perf(NODE-5906): optimize toArray to use batches

aditi-khare-mongoDB:
### Description
Performance optimization on `AbstractCursor.toArray`

#### What is changing?
Rather than `await`ing each document, `toArray` now only `await`s a new batch. 

No.

Performance improvement.

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -297,6 +297,7 @@ export abstract class AbstractCursor<
 
     return bufferedDocs;
   }
+
   async *[Symbol.asyncIterator](): AsyncGenerator<TSchema, void, void> {
     if (this.isClosed) {
       return;
@@ -457,9 +458,24 @@ export abstract class AbstractCursor<
    * cursor.rewind() can be used to reset the cursor.
    */
   async toArray(): Promise<TSchema[]> {
-    const array = [];
+    const array: TSchema[] = [];
+
+    // when each loop iteration ends,documents will be empty and a 'await (const document of this)' will run a getMore operation
     for await (const document of this) {
       array.push(document);
+      let docs = this.readBufferedDocuments();
+      if (this.transform != null) {
+        docs = await Promise.all(
+          docs.map(async doc => {
+            if (doc != null) {
+              return await this.transformDocument(doc);
+            } else {
+              throw Error;
+            }
+          })
+        );
+      }
+      array.push(...docs);
     }
     return array;
   }


src/cursor/abstract_cursor.ts
```diff
@@ -457,9 +458,24 @@ export abstract class AbstractCursor<
...
+
+    // when each loop iteration ends,documents will be empty and a 'await (const document of this)' will run a getMore operation
     for await (const document of this) {
       array.push(document);
+      let docs = this.readBufferedDocuments();
+      if (this.transform != null) {
+        docs = await Promise.all(
+          docs.map(async doc => {
+            if (doc != null) {
+              return await this.transformDocument(doc);
```
baileympearson, line 471: transformDocument is only async because it cleans up the cursor.  Does performance improve if we inline the call to `transform` here instead?

baileympearson: That would prevent unnecessary promises from being constructed.

nbbeeken: Our benchmarks do not run a cursor with a transform set but it would be great to get an idea of "how bad" this way of writing the code is.

src/cursor/abstract_cursor.ts
```diff
@@ -457,9 +458,24 @@ export abstract class AbstractCursor<
...
+            } else {
+              throw Error;
```
nbbeeken, line 473: Not sure this is what you intended, this is throwing the `Error` constructor. Is it possible for readBufferedDocuments to return an array that has nullish values in it?

src/cursor/abstract_cursor.ts
```diff
@@ -457,9 +458,24 @@ export abstract class AbstractCursor<
...
-    const array = [];
+    const array: TSchema[] = [];
+
+    // when each loop iteration ends,documents will be empty and a 'await (const document of this)' will run a getMore operation
     for await (const document of this) {
       array.push(document);
+      let docs = this.readBufferedDocuments();
+      if (this.transform != null) {
+        docs = await Promise.all(
+          docs.map(async doc => {
```
nbbeeken, line 469: After looking into Bailey's comment on removing the need to await a promise, if we still have `await` I'm not sure if it is desirable for us to create all these promises without awaiting them, and then in parallel find out if one of them threw. We should probably change this into a loop proper, rather than an `.all(.map)` style call.

nbbeeken (changes requested):
When you get the chance can you fill out the release highlights? 

And (I guess on slack?) can you share the numbers you're seeing, in BSON and previous driver performance improvements I have often included a conservative estimation of improvements downstream projects can expect to see from the perf change.

src/cursor/abstract_cursor.ts
aditi-khare-mongoDB: I ran tests with a transform and without, as well as making `transformDocuments` sync/async, doesn't seem like there's any local performance changes either way. Here are my [results](https://docs.google.com/document/d/1hm0cKfyz3FTYqQIIwY1pnQsBdunJ0ZGrmWF1UbAcnOI/edit#heading=h.8yw5ksibnwoj).

src/cursor/abstract_cursor.ts
aditi-khare-mongoDB: Yep! Actually, since `readBufferedDocuments` removes non-nullish values, I changed it to a NonNullable<TSchema> (we'll see how EVG fares with this, but locally my tests are passing).

src/cursor/abstract_cursor.ts
aditi-khare-mongoDB: Done!

src/cursor/abstract_cursor.ts
```diff
@@ -457,13 +458,22 @@ export abstract class AbstractCursor<
...
-    const array = [];
+    const array: TSchema[] = [];
+    // at the end of the loop (since readBufferedDocuments is called) the buffer will be empty
+    // then, the 'await of' syntax will run a getMore call
     for await (const document of this) {
       array.push(document);
+      const docs = this.readBufferedDocuments();
+      if (this.transform != null) {
+        for (const doc of docs) {
+          array.push(await this.transformDocument(doc));
```
baileympearson, line 469: The PR description says we no longer await each document when building the array.  That's true if there isn't a transform, but are we planning on also fixing the unnecessary `awaits` when there is a transform?

nbbeeken: I think the reply here: https://github.com/mongodb/node-mongodb-native/pull/4171#discussion_r1698991071 alludes to it not making a difference. @aditi-khare-mongoDB is that correct?

Maybe we should call out that the optimization only applies to cursors without a transform? 

I think the only way around this would either be making transform sync again or migrating logic from transform into toArray, which we can file a follow-up for, this still successfully addresses common usage.

nbbeeken (approved):

baileympearson (approved):

PR #4173, opened 2024-07-10
fix(NODE-6259): replace dynamically assigned length property with a static getter

sis0k0:
### Description

The length property was declared on the `BulkOperationBase` interface in the `@types/mongodb` package. With version 4.0, installing `@types/mongodb` is no longer required. However, the declarations generated in the `mongodb` package don't have the `length` property.

#### What is changing?

The `BulkOperationBase` type now has a `length` property. This was the behaviour when using `@types/mongodb`.

No

See [this discussion](https://www.mongodb.com/community/forums/t/how-to-check-if-bulkop-is-empty-or-not/287055/4) in the Community Forums.

### `BulkOperationBase` (super class of `UnorderedBulkOperation` and `OrderedBulkOperation`) now reports `length` property in Typescript

The `length` getter for these classes was defined manually using `Object.defineProperty` which hid it from typescript. Thanks to @sis0k0 we now have the getter defined on the class, which is functionally the same, but a greatly improved DX when working with types. 🎉  

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -911,8 +911,6 @@ export class BulkWriteShimOperation extends AbstractOperation {
 /** @public */
 export abstract class BulkOperationBase {
   isOrdered: boolean;
-  // Declare dynamically assigned property
-  declare length: number;
   /** @internal */
   s: BulkOperationPrivate;
   operationId?: number;
@@ -1180,6 +1178,10 @@ export abstract class BulkOperationBase {
     );
   }
 
+  get length(): number {
+    return this.s.currentIndex;
+  }
+
   get bsonOptions(): BSONSerializeOptions {
     return this.s.bsonOptions;
   }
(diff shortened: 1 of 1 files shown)


src/bulk/common.ts
```diff
@@ -911,6 +911,8 @@ export class BulkWriteShimOperation extends AbstractOperation {
 /** @public */
 export abstract class BulkOperationBase {
   isOrdered: boolean;
+  // Declare dynamically assigned property
+  declare length: number;
```
nbbeeken, line 915: Can we declare the getter here?

```ts
  public get length(): number {
    return this.s.currentIndex;
  }
```

nbbeeken (changes requested):
Thanks so much for jumping on this @sis0k0! 

The `Object.defineProperty(BulkOperationBase.prototype, 'length'` style of declaring class members is no longer needed since we're writing code with `class` syntax now. This field was overlooked when the driver was rewritten in TS, thanks for catching it and improving things!

src/bulk/common.ts
sis0k0: I added the getter a little further down in the file, before the other getters.

nbbeeken (approved):

PR #4174, opened 2024-07-11
feat(NODE-5682): set maxTimeMS on commands and preempt I/O

nbbeeken:
#### What is changing?

- add timeoutMS option to onData helper. Makes the iterator throw the same way an error event would.
- Add `maxTimeMS` field to commands when CSOT is enabled
- If minRoundTrip defined, add it to the timeoutContext
- listDatabases and runCommand respect timeoutMS settings

No

Part of the goal of CSOT. This enables the actual interruption of operations, and hints to the server (through maxTimeMS) how long it should permit an operation to run for.

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -29,7 +29,7 @@ import { type CancellationToken, TypedEventEmitter } from '../mongo_types';
 import { ReadPreference, type ReadPreferenceLike } from '../read_preference';
 import { ServerType } from '../sdam/common';
 import { applySession, type ClientSession, updateSessionFromResponse } from '../sessions';
-import { type Timeout } from '../timeout';
+import { type TimeoutContext } from '../timeout';
 import {
   BufferPool,
   calculateDurationInMs,
@@ -96,7 +96,7 @@ export interface CommandOptions extends BSONSerializeOptions {
   directConnection?: boolean;
 
   /** @internal */
-  timeout?: Timeout;
+  timeoutContext?: TimeoutContext;
 }
 
 /** @public */
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -29,7 +29,7 @@ import { type CancellationToken, TypedEventEmitter } from '../mongo_types';
 import { ReadPreference, type ReadPreferenceLike } from '../read_preference';
 import { ServerType } from '../sdam/common';
 import { applySession, type ClientSession, updateSessionFromResponse } from '../sessions';
-import { type TimeoutContext } from '../timeout';
+import { Timeout, type TimeoutContext, TimeoutError } from '../timeout';
 import {
   BufferPool,
   calculateDurationInMs,
@@ -416,6 +416,11 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       ...options
     };
 
+    if (options.timeoutContext?.csotEnabled()) {
+      const { maxTimeMS } = options.timeoutContext;
+      if (maxTimeMS > 0 && Number.isFinite(maxTimeMS)) cmd.maxTimeMS = maxTimeMS;
+    }
+
     const message = this.supportsOpMsg
       ? new OpMsgRequest(db, cmd, commandOptions)
       : new OpQueryRequest(db, cmd, commandOptions);
@@ -430,7 +435,9 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   ): AsyncGenerator<MongoDBResponse> {
     this.throwIfAborted();
 
-    if (typeof options.socketTimeoutMS === 'number') {
+    if (options.timeoutContext?.csotEnabled()) {
+      this.socket.setTimeout(0);
+    } else if (typeof options.socketTimeoutMS === 'number') {
       this.socket.setTimeout(options.socketTimeoutMS);
     } else if (this.socketTimeoutMS !== 0) {
       this.socket.setTimeout(this.socketTimeoutMS);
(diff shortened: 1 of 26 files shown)


src/timeout.ts
```diff
@@ -150,6 +150,11 @@ function isCSOTTimeoutContextOptions(v: unknown): v is CSOTTimeoutContextOptions
 
 /** @internal */
 export abstract class TimeoutContext {
+  start: number;
+  constructor() {
+    this.start = Math.trunc(performance.now());
```
baileympearson, line 155: any reason you're not using our `now()` helper?

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -384,7 +384,7 @@ describe('CSOT spec prose tests', function () {
         clock.restore();
       });
 
-      it('serverSelectionTimeoutMS honored if timeoutMS is not set', async function () {
+      it.skip('serverSelectionTimeoutMS honored if timeoutMS is not set', async function () {
```
baileympearson, line 387: Instead of skipping these tests, can't we just call `connect()` manually on the client before performing the operation that is using timeoutMS?

baileympearson (changes requested):

src/cmap/connection.ts
```diff
@@ -430,7 +436,9 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   ): AsyncGenerator<MongoDBResponse> {
     this.throwIfAborted();
 
-    if (typeof options.socketTimeoutMS === 'number') {
+    if (options.timeoutContext?.csotEnabled()) {
+      this.socket.setTimeout(0);
+    } else if (typeof options.socketTimeoutMS === 'number') {
       this.socket.setTimeout(options.socketTimeoutMS);
     } else if (this.socketTimeoutMS !== 0) {
       this.socket.setTimeout(this.socketTimeoutMS);
```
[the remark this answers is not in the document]
nbbeeken: Currently, this will timeout the `'data'` event we wait on which represents a whole message whereas the `socket.setTimeout` represents chunks, for CSOT we do need to change the way we timeout a message from how it was done previously.

We could consider changing the old way to work similar to CSOT.

src/cmap/connection.ts
```diff
@@ -416,6 +417,11 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       ...options
     };
 
+    if (options.timeoutContext?.csotEnabled()) {
+      const { maxTimeMS } = options.timeoutContext;
+      if (maxTimeMS > 0 && Number.isFinite(maxTimeMS)) cmd.maxTimeMS = maxTimeMS;
```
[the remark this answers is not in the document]
nbbeeken: Correct, but this is a check for greater than 0, so it confirms that maxTimeMS does not equal 0. The checks for if we should throw are colocated with the write/read steps.

src/cmap/connection.ts
```diff
@@ -622,7 +641,11 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
    */
   private async writeCommand(
     command: WriteProtocolMessageType,
-    options: { agreedCompressor?: CompressorName; zlibCompressionLevel?: number }
+    options: {
+      agreedCompressor?: CompressorName;
+      zlibCompressionLevel?: number;
+      timeoutContext?: TimeoutContext;
```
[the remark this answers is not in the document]
nbbeeken: There are plenty of usages that do not define any options when calling `connection.command`

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
nbbeeken: No because `connect` throws unfortunately, and if you squash the error then you get a client unconnected error

src/cmap/connection.ts
```diff
@@ -634,8 +657,32 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
+        );
+      }
+    }
+
     if (this.socket.write(buffer)) return;
-    return await once(this.socket, 'drain');
+
+    const drainEvent = once<void>(this.socket, 'drain');
+    if (options.timeoutContext?.csotEnabled()) {
+      const timeout = Timeout.expires(options.timeoutContext.remainingTimeMS);
```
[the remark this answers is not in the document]
nbbeeken: I'll look into it, but if infinity is returned then we can avoid making the promise, one less resource, might as well avoid it.

src/timeout.ts
nbbeeken: We are already using `performance.now()` in the rest of this file, I kept up with the pattern.

src/timeout.ts
```diff
@@ -150,6 +150,11 @@ function isCSOTTimeoutContextOptions(v: unknown): v is CSOTTimeoutContextOptions
 
 /** @internal */
 export abstract class TimeoutContext {
+  start: number;
```
baileympearson, line 153: This likely belongs in the CSOTTimeoutContext specifically, since it's not used in the abstract base class or in the legacy timeout context.

baileympearson (changes requested):

src/cmap/connection.ts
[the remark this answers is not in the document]
nbbeeken: Oh I see, yea that would work, how do you intend we resolve the `options.socketTimeoutMS` vs the one on the class `this.socketTimeoutMS` currently there's this fallback behavior, I need to hand the context an instance of the connection class and the options.

Are you thinking something along these lines:
```ts
const t = context.socketTimeoutMS(this, options) // CSOT returns 0, legacy resolves options vs connection
socket.setTimeout(t)
```

src/timeout.ts
nbbeeken: Good call, moved!

PR #4177, opened 2024-07-16
feat(NODE-5614): Add support for explicit resource management 

baileympearson:
#### What is changing?

This PR adds support for explicit resource management to our async disposable resources as outlined in the [design](https://docs.google.com/document/d/1eT-7ntnX--M49OPmLO41RNe1PJ8Vore6TE2Kz3IY_cI/edit#heading=h.b1os3ai9s8t3).

##### Beta Namespace

A new entrypoint for the driver has been added, `mongodb/lib/beta`.  This namespace will contain any experimental features our APIs we want to expose to users.

API extractor natively supports tagging things as `beta` and producing a beta namespace, so this PR configures API extractor to emit a beta namespace for us.

Users can access the beta namespace by importing from `mongodb/lib/beta`.

##### Symbol.asyncDispose methods

Symbol.asyncDispose methods have been added to the MongoClient, ClientSession, Cursors and ChangeStream classes.  These methods are conditionally defined only when `Symbol` has an `asyncDispose` property (i.e., the user has polyfilled this property themselves or they're running on a compatible version of Nodejs where it's available).

Our readable streams (streaming cursors and download streams) are also disposable on compatible versions of Nodejs because they inherit `Symbol.asyncDispose` from `Nodejs.Readable`.

Additionally, a method is provided to users to allow them to attach the `Symbol.asyncDispose` methods to each class manually, if desired.  This avoids complications for users when the user is polyfilling `Symbol.asyncDispose` but they may import the driver (and load its classes) before the polyfill runs.  In this case, since Symbol.asyncDispose is not defined when evaluating the driver's source code, no disposal methods are added.  The `configureExplicitResourceManagement` can be called manually to add the disposal methods to each class in this scenario.

##### Testing

There are three major CI and testing changes:

1. A new manual test has been added that tests whether or not Symbol.asyncDispose is added to our resources in certain Nodejs environments.  This test suite also asserts on the expected behavior of asyncDispose - namely, that each implementation calls the correct cleanup method for the resource.
2. A new test suite has been added that smoke tests driver resource to confirm that they can actually be used with resource management.  This is a new subpackage in our testing directory because it requires tslib and additional TS configuration.  This test suite confirms that our resources can be used with `await using` and that they are cleaned up appropriately.
3. (optional) I enhanced the TS testing matrix a bit during development and decided to leave it in the PR.  The changes here primarily test with different versions of Typescript and @types/node.  The versions I chose are based on what currently pass to more explicitly test our compatibility.  I can revert these changes if desired.

### Support for explicit resource management

The driver now natively supports [explicit resource management](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management) for MongoClient, ClientSession, ChangeStreams and cursors.  Additionally, on compatible Nodejs versions, explicit resource management can be used with `cursor.stream()` and the `GridFSDownloadStream`, since these classes inherit resource management from Nodejs' readable streams.

This feature is experimental and subject to changes at any time.  This feature will remain experimental until the proposal has reached stage 4 and Nodejs declares its implementation of async disposable resources as stable. 

To use explicit resource management with the Node driver, you must:

- Use Typescript 5.2 or greater (or another bundler that supports resource management)
- Enable tslib polyfills for your application
- Either use a compatible Nodejs version or polyfill `Symbol.asyncDispose` (see the [TS 5.2 release announcement](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management:~:text=Because%20this%20feature%20is%20so%20recent%2C%20most%20runtimes%20will%20not%20support%20it%20natively.%20To%20use%20it%2C%20you%20will%20need%20runtime%20polyfills%20for%20the%20following%3A) for more information).

Explicit resource management is a feature that ensures that resources' disposal methods are always called when the resources' scope is exited.  For driver resources, explicit resource management guarantees that the resources' corresponding `close` method is called when the resource goes out of scope.

```typescript
// before:
{
  try {
    const client = MongoClient.connect('<uri>');
    try {
      const session = client.startSession();
      const cursor = client.db('my-db').collection("my-collection").find({}, { session });
      try {
        const doc = await cursor.next();
      } finally {
        await cursor.close();
      }
    } finally {
      await session.endSession();
    }
  } finally {
    await client.close();
  }
}

// with explicit resource management:
{
  await using client = MongoClient.connect(<uri>);

  await using session = client.startSession();
  await using cursor = client.db('my-db').collection('my-collection').find({}, { session });

  const doc = await cursor.next();
}
// outside of scope, the cursor, session and mongo client will be cleaned up automatically.
```

The full explicit resource management proposal can be found [here](https://github.com/tc39/proposal-explicit-resource-management).


src/index.ts
```diff
@@ -111,6 +111,7 @@ export { ReturnDocument } from './operations/find_and_modify';
 export { ProfilingLevel } from './operations/set_profiling_level';
 export { ReadConcernLevel } from './read_concern';
 export { ReadPreferenceMode } from './read_preference';
+export { AsyncDisposable } from './resource_management';
```
nbbeeken, line 114:
```suggestion
export { type AsyncDisposable } from './resource_management';
```

Also, minor, this is grouped with the enums, but that's something we've manually kept organized.

src/beta.ts
```diff
@@ -0,0 +1 @@
+export * from './index';
```
nbbeeken, line 1: The generated JS is:

```js
__exportStar(require("./index"), exports);
```

Does having tslib installed make that helper available on the global somehow? 

If so, "`or another bundler that supports resource management`" is there any other tool that works, or is this for TS only? (I also see that we say "enable tslib", is there a special import for that?)

nbbeeken (changes requested):
Pretty much good to go! two small functional changes and some api-extractor fixes.

test/explicit-resource-management/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "build": "tsc",
+    "test": "npm run build && mocha --config .mocharc.json main.test.js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@types/chai": "^4.3.16",
+    "chai": "^4.4.1",
+    "mongodb": "../../mongodb-6.8.0.tgz",
```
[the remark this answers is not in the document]
baileympearson: I wanted the smoke test to use a packaged driver, this was intentional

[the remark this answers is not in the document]
nbbeeken: Sure, and fine to leave it as is, for my understanding What is or Is there a difference? what is a package testing that a symlink doesn't? IIUC npm prepare is the step that runs in either case, is there something else?

test/explicit-resource-management/main.test.ts
```diff
@@ -0,0 +1,123 @@
...
+async function setUpCollection(client: MongoClient) {
+	const collection = client.db('foo').collection<{ name: string }>('bar');
+	const documents: Array<{ name: string }> = Array.from({ length: 5 }).map(i => ({
+		name: String(i)
+	}));
+	await collection.insertMany(documents)
+	return collection;
+}
+
+describe('explicit resource management smoke tests', function () {
```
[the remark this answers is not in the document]
baileympearson: I don't think that's necessary, the contents of that function are straightforward.  This is intended as a smoke test, not exhaustive testing.

src/beta.ts
baileympearson: No, it depends on your ts config settings.  I'm surprised this is what's exported.

Do you have any ideas how we can export the entirety of the index without duplicating the entire file?

src/resource_management.ts
```diff
@@ -0,0 +1,57 @@
+import { AbstractCursor, ChangeStream } from './beta';
+import { MongoClient } from './mongo_client';
+import { ClientSession } from './sessions';
+
+/**
+ * @public
+ * @experimental
```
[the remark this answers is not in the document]
baileympearson: No.  beta and public are mutually exclusive.  But we need to export this, otherwise API extractor complains because public classes implement this interface.

src/index.ts
baileympearson: What is the value of organization of this file?  This frequently comes up in reviews and it's primarily just you requesting changes for organization purposes.  But best I can tell, there's no impact for us or users if it's not organized

test/explicit-resource-management/package.json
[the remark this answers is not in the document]
baileympearson: I wanted it to be as close as possible to what users will experience, that's all.  Ideally we'd install a real beta driver to test with, but since we can't do that per-CI run, npm pack was the next best option.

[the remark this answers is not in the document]
baileympearson: What's wrong with npm pack?

src/resource_management.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  Symbol.asyncDispose &&
+    (AbstractCursor.prototype[Symbol.asyncDispose] = async function () {
+      await this.close();
+    });
+
+  Symbol.asyncDispose &&
+    (ChangeStream.prototype[Symbol.asyncDispose] = async function () {
+      await this.close();
+    });
```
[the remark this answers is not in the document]
baileympearson: > What if we just called this function in this module and removed the other definitions throughout the driver?

I considered it, but decided that the code was simple enough to duplicate, so I didn't really see a benefit

src/resource_management.ts
```diff
@@ -36,22 +36,42 @@ export interface AsyncDisposable {
...
+      configurable: true,
+      writable: true
     });
 
   Symbol.asyncDispose &&
-    (ChangeStream.prototype[Symbol.asyncDispose] = async function () {
-      await this.close();
+    Object.defineProperty(ClientSession.prototype, Symbol.asyncDispose, {
+      value: async function asyncDispose(this: { endSession(): Promise<void> }) {
+        await this.endSession();
```
nbbeeken, line 71: This was passing in `{force:true}` in the previous commit, I think you still want to do that correct? (Need to update the other copy of this as well)

src/index.ts
nbbeeken: No need to change it, that's why I call out that it has been manually maintained, it does provide visual organization for us when reading this file.

The important part here is the `type` export

src/beta.ts
nbbeeken: Does the `export =` solution you have now work? I am unsure but I think this may make this file import strangely if it's a genuine ESM import, but that's something you can manually verify works. 

Another option may be to write `export type * from 'index'` and write the _exportStar logic in TS: essentially `Object.assign(module.exports, require('./index'))`.

src/resource_management.ts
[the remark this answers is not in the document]
nbbeeken: If there's changes to options (like force: true for the sessions), only one place needs updating. But I don't feel strongly.

test/explicit-resource-management/main.test.ts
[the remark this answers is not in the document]
nbbeeken: Since we have two copies of the dispose functions, and possibly may add more (thinking txn), is there a way we can assess that the functions are identical?

test/explicit-resource-management/package.json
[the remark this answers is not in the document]
nbbeeken: > What's wrong with npm pack?

Nothing :) and I'll resolve this as it was a suggestion anyway

nbbeeken (changes requested):

nbbeeken (dismissed):

etc/docs/build.ts
```diff
@@ -24,7 +24,7 @@ const RELEASES_JSON_FILE = './template/static/versions.json';
 
 const copyGeneratedDocsToDocsFolder = () => exec(`cp -R temp/. ../../docs/.`);
 const removeTempDirectory = () => exec('rm -rf temp');
-const installDependencies = () => exec('npm i --no-save --legacy-peer-deps typedoc@0.24.8');
```
aditi-khare-mongoDB, line 27: Is there a reason why we're updating the `typedoc` dependency within this PR?

test/explicit-resource-management/.gitignore
aditi-khare-mongoDB, line 1: why is the `explicit-resource-management` folder top-level in the test folder? Most of these tests seem like unit tests

test/manual/resource_management.test.ts
aditi-khare-mongoDB, line 1: Another question on testing structure, when do we decide to put tests in the `manual` folder?

aditi-khare-mongoDB (changes requested):

etc/docs/build.ts
aditi-khare-mongoDB: Nevermind, just read your clarification in the PR description.

test/explicit-resource-management/.gitignore
baileympearson: These are essentially unit tests but they need to run in a special configuration - this is a standalone npm project with its own package.json.  These tests will fail as a part of any unit test / other CI configuration, since they rely on the special tsconfig in this project.  

I'm not sure there's a "great" place for tests like this but I chose a top-level folder because that's what we do for lambda tests.  lambda tests are similar in that they run inside their own npm package.

api-extractor.json
```diff
@@ -10,7 +10,8 @@
   "dtsRollup": {
     "enabled": true,
     "untrimmedFilePath": "",
-    "publicTrimmedFilePath": "<projectFolder>/<unscopedPackageName>.d.ts"
+    "publicTrimmedFilePath": "<projectFolder>/<unscopedPackageName>.d.ts",
+    "betaTrimmedFilePath": "<projectFolder>/lib/beta.d.ts"
```
dariakp, line 14: Are there any other features currently marked experimental that we want to expose via the beta tag? Have we talked to Alex about how we are going to differentiate messaging in experimental-tagged vs beta-tagged features? (e.g., we currently have a "use at your own risk" disclaimer for experimental features, which is missing from the release highlight for this feature - I imagine we want to add something about the beta aspects). Similarly, how will users know about the existence of the beta namespace and how to access it?

dariakp (changes requested):
Some meta comments ahead of code review proper

test/explicit-resource-management/.gitignore
aditi-khare-mongoDB: If we create a folder with the path `test/unit/explicit-resource-management` and put the config information there, would it fail? I thought, no, but I might be mistaken.

.evergreen/config.in.yml
```diff
@@ -348,10 +348,36 @@ functions:
...
+      params:
+        working_dir: "src"
+        timeout_secs: 60
+        env:
+          PROJECT_DIRECTORY: ${PROJECT_DIRECTORY}
+        binary: bash
+        args:
+          - "${PROJECT_DIRECTORY}/.evergreen/run-resource-management.sh"
+
+  "check resource management smoke tests":
```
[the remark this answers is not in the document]
baileympearson: Sure.  Did you have a different name in mind?

api-extractor.json
baileympearson: > Are there any other features currently marked experimental that we want to expose via the beta tag?

No.

> Have we talked to Alex about how we are going to differentiate messaging in experimental-tagged vs beta-tagged features?

No, but I can ask.  There's also no reason we can't tag everything that is tagged beta as experimental - beta is only used to indicate to API extractor what should be extracted into the beta `.d.ts` file.

> Similarly, how will users know about the existence of the beta namespace and how to access it?

I don't think the intention is for users to "find" the beta namespace and explore it.  Instead, users will find a beta feature, either through release notes or documentation or a blog post if Alex wants to write one, and our reference will point users towards the beta namespace when showing them how to use the feature.  There's already a docsp ticket filed for this work.

test/explicit-resource-management/main.test.ts
```diff
@@ -0,0 +1,120 @@
...
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+import { AbstractCursor, ChangeStream, ClientSession, GridFSBucket, MongoClient } from 'mongodb/lib/beta';
+import * as sinon from 'sinon';
+import { Readable } from 'stream';
+import { pipeline } from 'stream/promises';
+import { setTimeout } from 'timers/promises';
+
+async function setUpCollection(client: MongoClient) {
+	const collection = client.db('foo').collection<{ name: string }>('bar');
```
dariakp, line 11: it looks like this file is using tab-formatting which makes it look different than the other test file in this PR (and in the code base) - can we update it to be consistent with the other files?

test/explicit-resource-management/package.json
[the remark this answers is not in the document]
baileympearson: That's a good point.  We'll probably need to rename this file.  6.8.0 does include the current `version` from the package.json, so this will change after the next release.  

I'll rename the packed file to a stable filename and depend on that instead.

test/explicit-resource-management/main.test.ts
```diff
@@ -0,0 +1,120 @@
...
+	afterEach(function () {
+		clientSpy.resetHistory();
+		cursorSpy.resetHistory();
+		endSessionSpy.resetHistory();
+		changeStreamSpy.resetHistory();
+		readableSpy.resetHistory();
+	});
+
+	describe('MongoClient', function () {
+		it('can be used with await-using syntax', async function () {
```
dariakp, line 35: "can be used with await-using syntax" is somewhat vague intent-wise, e.g., to me, it means "doesn't crash when used with await-using syntax". I think we should be testing the part of the user expectation that really matters here, i.e., that `it('is cleaned up automatically when used with await-using syntax')` and assert that the corresponding `close` method is called as expected, instead of the `asyncDispose` (or in addition to it, I don't have strong opinion on whether or not to keep that part of the assertion)

what I am looking for is a set of tests that covers the expected user-facing behavior: it can be here or it can be in a different suite if this one is meant to serve a more limited purpose

test/explicit-resource-management/main.test.ts
```diff
@@ -0,0 +1,120 @@
...
+			{
+				await using client = new MongoClient(process.env.MONGODB_URI!);
+				await client.connect();
+
+				const collection = await setUpCollection(client);
+
+				await using cursor = collection.find();
+				await cursor.next();
+				await cursor.next();
+				await cursor.next();
```
dariakp, line 56: what's special about the 3 calls to next? should there also be a test where the cursor isn't iterated at all, and perhaps, a test where the cursor is fully exhausted, too?

test/manual/resource_management.test.ts
```diff
@@ -0,0 +1,89 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+
+import { AbstractCursor, ChangeStream, ClientSession, MongoClient } from '../mongodb';
+
+describe('Explicit Resource Management Tests', function () {
```
dariakp, line 6: The explicit resource management spec is more than just the `asyncDispose` symbol; considering the scope of these tests, I think it would be more accurate to say that this test file is for `'Symbol.asyncDispose [functionality in the driver]'`.

test/manual/resource_management.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  describe('Symbol.asyncDispose defined', function () {
+    beforeEach(function () {
+      if (!('asyncDispose' in Symbol)) {
+        this.currentTest.skipReason = 'Test must run with asyncDispose available.';
+        this.skip();
+      }
+    });
+
+    describe('MongoClient', function () {
+      it('closes the the client', async function () {
```
dariakp, line 22: I suggest flattening the tests (getting rid of the class-level describes) and making the `it` blocks read `'calls <method> when invoked on the <class name>'` in order to make it clear what is performing the action under test and when

dariakp (changes requested):

.evergreen/config.in.yml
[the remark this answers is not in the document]
dariakp: Off the top of my head, perhaps "feature integration" tests or "feature end to end" tests? As opposed to the others that are more of "implementation" or "internal implementation" tests. We could rename both for more clarity, certainly open to other suggestions from author/team here.

test/explicit-resource-management/.gitignore
baileympearson: That's an interesting question - I don't think so, but I'm not positive.  Do you think I should look into it?

These tests are slightly different from our unit tests or integration tests.  These tests do require a server, so they can't be unit tests, although I could use the mock server.  Also, there's additional setup here - these tests specifically depend on a packed driver (they have their own package.json file in this folder, and the `run-resource-management-smoke-tests.sh` script packs the driver and installs it into the resource management tests).

test/explicit-resource-management/main.test.ts
dariakp: I do think there is still value in testing the feature in the context in which it will be used by our users. If, hypothetically speaking, the runtime/bundler has an issue where the way we have implemented asyncDispose is no longer sufficient to ensure proper cleanup, it would be better to find out about it in our test suite as opposed to from user reports. There is also the additional benefit of ensuring that our cleanup methods themselves adhere to the idempotency expected from the asyncDispose implementation: that if the user's code inside the context block results in our driver automatically cleaning up the resource, then calling it again implicitly at the end of a context block doesn't cause any issues.

test/explicit-resource-management/main.test.ts
```diff
@@ -0,0 +1,84 @@
...
+async function setUpCollection(client: MongoClient) {
+  const collection = client.db('foo').collection<{ name: string }>('bar');
+  const documents: Array<{ name: string }> = Array.from({ length: 5 }).map(i => ({
+    name: String(i)
+  }));
+  await collection.insertMany(documents)
+  return collection;
+}
+
+describe('explicit resource management smoke tests', function () {
```
dariakp, line 17:
```suggestion
describe('explicit resource management feature integration tests', function () {
```

test/manual/resource_management.test.ts
dariakp: The previous suggestion was just one way of trying to make it more clear what condition is being tested. The `asyncDispose` being defined doesn't in itself cause the MongoClient to close, which is what the current prose suggests, and MongoClient can be closed regardless of whether `asyncDispose` exists. So you need to add something along the lines of `closes <resource> via direct call to Symbol.asyncDispose` in order to fully convey the intent of the test. (In contrast, the symbol not being defined as opposed to not defined IS what is responsible for the absence of the resulting definition in the negative cases, so those correctly convey their intent).

dariakp (changes requested):

test/explicit-resource-management/main.test.ts
baileympearson: I disagree - it is not our responsibility to test the implementation of bundlers or runtimes in this way.  The resource management proposal is more-or-less stable and the changes being made are all minor, internal implementation details.  I mentioned this in the [spike](https://docs.google.com/document/d/1eT-7ntnX--M49OPmLO41RNe1PJ8Vore6TE2Kz3IY_cI/edit#heading=h.hxndodn23tzy).  This feature has had a year of adoption in the community and any breakage in tooling would disrupt the ecosystem enough that it would be caught quickly.

The implementation of asyncDispose is tested separately.  Our dispose methods proxy to the resources' underlying disposal method, and any behavioral requirements for `close()` or `endSession()` should be tested elsewhere.

I believe the current testing sufficient, so unless you consider additional testing a blocker to merging this PR, can we consider this resolved?  If not, can you elaborate on exactly what testing you'd like to see added before you would consider this PR unblocked?

test/explicit-resource-management/main.test.ts
baileympearson: No reason.  I removed 2 calls.

test/explicit-resource-management/main.test.ts
dariakp: I believe that it is in the best interest of our users that we test any workflow that we advertise to them, i.e., we should contain tests that, at a minimum, cover the syntax we use in our documentation, and ideally mimic any end-to-end workflow that we would consider supported.

It's the same reason we write integration tests for other driver features and don't consider it sufficient to just unit test each individual component: things can break down in unexpected ways when put together no matter how thoroughly you try to define and test the individual functionalities and the contracts between them, so the best way to guarantee a smooth end-to-end experience is to test the end-to-end experience.

Based on my understanding of the purpose of this work, supported user flows concern automatic cleanup of the client, sessions, and cursors. I would like to see the happy path and main error scenarios covered for these flows. Happy path would cover successfully instantiating the resource and then, if it is stateful, checking that the cleanup happens as expected for each state (e.g., for cursors we'd want to cover instantiation, mid-iteration, and exhaustion separately). Error means we get the expected error with `using` even if the resource fails to instantiate (e.g., uri validation in the client) or fails later with an error. I think you could collaborate with the primary reviewer (@nbbeeken) on the exact lineup of test cases that can help ensure that users don't run into surprising behavior when they go to try out this feature.

dariakp (changes requested):

test/readme.md
```diff
@@ -32,6 +32,7 @@ Below is a summary of the types of test automation in this repo.
...
+| Explicit Resource Management           | `/test/integration/explicit-resource-management`    | Tests that use explicit resource management with the driver's disposable resources.                                                                                                                                                                                                                                                                                                                                                                                                             | `bash .evergreen/run-resource-management-feature-integration.sh`                                                                                                                                                                                                                                                                                      |
```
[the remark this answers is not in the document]
baileympearson: Typo, it's intended to be top-level, following the pattern of the lambda tests.

test/explicit-resource-management/main.test.ts
baileympearson: Met with @nbbeeken and I've added tests for the scenarios we discussed.

test/explicit-resource-management/.gitignore
aditi-khare-mongoDB: Ah I see, why doesn't this fall under integration tests? I hope this isn't too nit-picky; I don't want to block your work over a testing structure issue.

aditi-khare-mongoDB (commented):
Left small comment, feel free to message me on slack for clarification.

api-extractor.json
dariakp: Do you mean that we could label the same thing as both beta and experimental and it's the experimental tag that indicates non-stability? Could you clarify what the beta tag should signify to users?

baileympearson: Yes.  That's also what is currently implemented in this PR.

The beta tag, if users see it, indicates the same thing.  But it's only present because API extractor requires it to extract the beta API.

dariakp: (I assume it's just for actually surfacing the feature in the docs and would always be paired with experimental, but I wanted to make sure - it's something that we'll want to capture either in our internal team FAQ or in this repo's readme for future reference)

dariakp: Thanks - I've now added a note to our internal FAQ on pre-release features about the beta tag. Could you add our standard note on `@experimental` features to the release highlight text? Once that's done, we can resolve this thread.

test/explicit-resource-management/.gitignore
baileympearson: I think these are good questions to ask, it's making me think about the testing structure more carefully even if we decide to leave it as-is.

There are just a few special considerations for these tests.

1. we need separate typescript and linting configurations to use the new syntax
2. they import from a special, `beta` namespace from the driver unlike the rest of the integration tests
3. related to 2), and how the tests currently run:  we currently run them in a separate task that packs the driver into a tarball, installs it into the testing package (explicit-resource-management is its own npm package with its own tsconfig and package.json) and then runs the tests.  this setup would not work with the integration test suite

But that being said, it may be possible to somehow have a separate tsconfig for a single file in the integration test suite that has the settings these tests require and I can take a look into it if you'd like me to.

test/explicit-resource-management/main.test.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+describe('explicit resource management feature integration tests', function () {
+  describe('MongoClient', function () {
+    it('does not crash or error when used with await-using syntax', async function () {
+      await using client = new MongoClient(process.env.MONGODB_URI!);
+      await client.connect();
+    })
+
+    it('always cleans up the client, regardless of thrown errors', async function () {
```
dariakp, line 27: for the cases that say "cleans up X", can we make the assertion somehow relate to the clean up (either assert that the resource is closed or that the relevant method was called with the expected arguments where applicable?) or, if not,  can we align the test case description with what the test actually guarantees?

dariakp (changes requested):

test/explicit-resource-management/main.test.ts
baileympearson: I'd prefer not to.  I think "cleans up x resource" is accurate.  I also don't want to introduce any sinon spying into this test suite - we are already covered with sinon assertions asserting we call the correct disposal method Symbol.asyncDispose is called in the manual integration tests.

baileympearson: Definitely a good addition, although it doesn't help us here.  We can't check `client.closed` because after we leave scope there is no client 🙂

nbbeeken:
```js
let autoCleanedUpClient;
const error = await (async () => {
  await using client = new MongoClient(process.env.MONGODB_URI!);
  autoCleanedUpClient = client;
  // ...
expect(autoCleanedUpClient).to.have.nested.property('s.hasBeenClosed', true); 
```

Seems like you can save what you need to check outside of the scope. And we can use what we already have internally instead of a `closed` prop 

(although I don't disagree, that `x.closed` is nice, and maybe a pattern we should make sure is shared on everything in design close work)

baileympearson: Okay, I feel that perhaps I haven't adequately communicated this.  Spying is an option, so is saving a reference to the client, or other things like overriding the dispose method with a custom implementation, etc.

I feel that any testing of whether or not `Symbol.asyncDispose` is called or not is just testing the TS polyfill, and that since we already have tests that use sinon to assert that the correct cleanup methods are called when `Symbol.asyncDispose` is called, asserting on client state / spying in the feature smoke tests isn't necessary.

dariakp: My recommendation is based on two testing best practices:
1) test prose should accurately describe what the author intended to safeguard by writing the test
2) tests should _always_ fail if what they say *should* happen does not happen

Both of these speak to the "DAMP not DRY" idea of testing ([here](https://abseil.io/resources/swe-book/html/ch12.html) is a good read on the subject of writing clear tests with that principle in context).

So, if a test says, e.g., "always cleans up the client", then that test needs to fail if the client is not, in fact, cleaned up. Right now, this test won't fail if the `asyncDispose` becomes a no-op. That behavior would be surprising to any future developer looking at this test case. This can be fixed either by adjusting the assertions or the description. Either is acceptable to me personally, though the consensus among the three reviewers here seems to be in favor of assertions on the cleanup.

test/explicit-resource-management/main.test.ts
dariakp: Resolving this thread in favor of the other test-specific thread since some of the concerns have now been addressed in the docs changes and the tests

test/explicit-resource-management/main.test.ts
baileympearson: I added sinon assertions to these tests.

src/beta.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @internal
+ *
+ * Since we don't bundle tslib helpers, we need to polyfill this method.
+ *
+ * This is used in the generated JS.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function __exportStar(mod: Document) {
+  for (const key of Object.keys(mod)) {
+    exports[key] = void 0;
```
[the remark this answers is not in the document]
baileympearson: Done.

That line isn't necessary, it got added while I was experimenting with alternate approaches.  I removed it.

dariakp (approved):

dariakp (approved):

nbbeeken (approved):

test/explicit-resource-management/.gitignore
aditi-khare-mongoDB: No need, don't want to hold up this PR!

aditi-khare-mongoDB (approved):

PR #4180, opened 2024-07-17
test(NODE-3237): migrate legacy spec tests to unified format

baileympearson:
### Description

NODE-3211,NODE-3213,NODE-5975,NODE-5976,NODE-5984,NODE-6004,NODE-6005

#### What is changing?

Migrates all legacy spec tests except for FLE to the unified test format.  All spec tests are from the latest `master` in the specs repo.

This PR syncs:

- read/write concern
- gridfs
- retryable writes
- retryable reads
- crud v1
- crud v2
- Atlas data lake tests

No.


src/operations/find_and_modify.ts
```diff
@@ -192,12 +192,7 @@ export class FindAndModifyOperation extends CommandOperation<Document> {
       ...this.cmdBase
     };
 
-    // Have we specified collation
-    try {
-      decorateWithCollation(cmd, coll, options);
-    } catch (err) {
-      return err;
-    }
+    decorateWithCollation(cmd, coll, options);
```
baileympearson, line 195: I can revert this change, but without this the API doesn't error but returns an error.  That's not indicated by TS at all, so I figured it would be better to just throw.

nbbeeken:
```js
    // Have we specified collation
    try {
      decorateWithCollation(cmd, coll, options);
    } catch (err) {
      return callback(err);
    }
```
I can see in 4.x this is the code, so it was an oversight in async/await conversion that this was left as a `return` rather than changed to a `throw`. Good catch!

nbbeeken (approved):

test/integration/retryable-writes/retryable_writes.spec.test.ts
```diff
@@ -1,207 +1,27 @@
...
-  runUnifiedSuite(loadSpecTests(path.join('retryable-writes', 'unified')));
+  runUnifiedSuite(loadSpecTests(path.join('retryable-writes', 'unified')), ({ description }) => {
+    return clientBulkWriteTests.includes(description)
+      ? `TODO(NODE-6257): implement client-level bulk write.`
+      : description ===
+        'RetryableWriteError label is not added based on writeConcernError in pre-4.4 mongos response'
+      ? 'TODO(NODE-5720)'
+      : false;
+  });
 });
```
aditi-khare-mongoDB, line 27: Looks great! @nbbeeken, this is what I mentioned I wanted to check for in our meeting.

aditi-khare-mongoDB (approved):
LGTM!

PR #4181, opened 2024-07-22
test(NODE-6098): enable unified valid fail tests

durran:
### Description

Enables the unifed test format's valid-fail tests and fixes all failures.

#### What is changing?

- Syncs the valid-fail tests in the specs directory to latest.
- Adds the necessary validation in the unified runner to pass the tests

None

NODE-6098

### DIFF
--- a/test/tools/unified-spec-runner/runner.ts
+++ b/test/tools/unified-spec-runner/runner.ts
@@ -306,13 +306,21 @@ async function runUnifiedTest(
  */
 export function runUnifiedSuite(
   specTests: uni.UnifiedSuite[],
-  skipFilter: uni.TestFilter = () => false
+  skipFilter: uni.TestFilter = () => false,
+  expectRuntimeError = false
 ): void {
   for (const unifiedSuite of specTests) {
     context(String(unifiedSuite.description), function () {
       for (const [index, test] of unifiedSuite.tests.entries()) {
         it(String(test.description === '' ? `Test ${index}` : test.description), async function () {
-          await runUnifiedTest(this, unifiedSuite, test, skipFilter);
+          if (expectRuntimeError) {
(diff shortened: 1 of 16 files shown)


test/tools/unified-spec-runner/runner.ts
```diff
@@ -306,13 +306,21 @@ async function runUnifiedTest(
...
   for (const unifiedSuite of specTests) {
     context(String(unifiedSuite.description), function () {
       for (const [index, test] of unifiedSuite.tests.entries()) {
         it(String(test.description === '' ? `Test ${index}` : test.description), async function () {
-          await runUnifiedTest(this, unifiedSuite, test, skipFilter);
+          if (expectRuntimeError) {
+            const error = await runUnifiedTest(this, unifiedSuite, test, skipFilter).catch(
+              error => error
+            );
+            expect(error).to.exist;
```
durran, line 320: We just assert existence of an error since the unified spec does not define any specific types of errors or what their messages should be.

nbbeeken: Looks like there are not many cases that aren't an assertion error. Maybe we can change the manually thrown errors to a chai assertion and change this check to AssertionError or MongoServerError. My thinking is that we want to test we're failing for assertion reasons not "crash" reasons, to check the correctness of the runner.

durran: Good suggestion. I've updated the unified runner itself to only throw `AssertionError`, and then narrowed the check down to look for only 4: `AssertionError`, `MongoServerError`, `TypeError`, and `MongoParseError`. The later 2 are coming from EJSON parsing of the spec file and URI parsing, so out of the unified runner's control.

nbbeeken: That sounds good to me, a bit more specificity should help catch bugs.

nbbeeken (approved):

PR #4185, opened 2024-07-29
feat(NODE-5754): allow auto select family options

durran:
### Description

Allows users to specify auto select family options passed to the underlying connections.

#### What is changing?
- Adds `autoSelectFamily` option to the client which defaults to `true`.
- Adds `autoSelectFamilyAttemptTimeout` option to the client with no default.
- Adds tests to verify the options are passed through to regular and TLS connections.

Add note in documentation about the options.

NODE-5754

### Driver now supports auto selecting between IPv4 and IPv6 connections

For users on Node versions that support the `autoSelectFamily` and `autoSelectFamilyAttemptTimeout` options (Node 18.13+), they can now be provided to the `MongoClient` and will be passed through to socket creation. `autoSelectFamily` will default to `true` with `autoSelectFamilyAttemptTimeout` by default not defined. Example:

```ts
const client = new MongoClient(process.env.MONGODB_URI, { autoSelectFamilyAttemptTimeout: 100 });
```


test/manual/mocharc.json
```diff
@@ -1,5 +1,8 @@
 {
-  "require": "ts-node/register",
+  "require": [
+    "ts-node/register",
+    "test/tools/runner/chai_addons.ts"
```
durran, line 4: Unrelated to the feature but needed for sinon-chai assertions in the manual tests.

nbbeeken (changes requested):
Changes LGTM! Can you fill out the release highlights? this is def a feature to show off :tada:

durran:
> Changes LGTM! Can you fill out the release highlights? this is def a feature to show off 🎉

Ah yeah missed that. Updated now.

durran:
> Question: should we also be using `autoSelectFamily` when establishing sockets for KMS requests in the state machine?

Good question. I think the answer to that should be yes? If we agree I can add to this PR.

baileympearson:
@durran We could potentially consume `makeSocket` directly from the KMS fetching in the state machine.  That's a nice improvement regardless.  Then we'd automatically support the new option, so long as we pass it into the state machine.  

want to spend a bit of time looking at that approach?

nbbeeken (changes requested):
I am just putting back my request changes state since you're working on the above, given the reactions I think we're all on board with expanding this to KMS. 

Using `makeSocket` sounds great but you may run into more refactoring than is worth, I'm thinking about how connecting will be inside that function now and you may have to catch and re-throw in "interesting" ways, additionally, I am unsure if the socks5 behavior is precisely equivalent. 

If none of that ends up being a concern GREAT! But if it is, as long as we can get the autoFamily option through we can rest this work there, and file a follow up to consolidate 🚀 Good catch @baileympearson

durran:
The refactor to use `makeSocket` did prove to be a bit more invasive for my liking, so I've left it out here and will create a ticket for that. PR has now been updated to pass through the new socket options to the state machine to use with KMS requests.

src/client-side-encryption/state_machine.ts
```diff
@@ -289,14 +302,27 @@ export class StateMachine {
...
       host: parsedUrl[0],
       servername: parsedUrl[0],
       port
     };
     const message = request.message;
     const buffer = new BufferPool();
 
+    const socketOptions = this.options.socketOptions || {};
+    if ('autoSelectFamily' in socketOptions) {
+      options.autoSelectFamily = socketOptions.autoSelectFamily;
```
nbbeeken, line 320: We want to default `autoSelectFamily` to `true` if it is not given right?

src/client-side-encryption/state_machine.ts
```diff
@@ -289,14 +302,27 @@ export class StateMachine {
...
+    }
+    if ('autoSelectFamilyAttemptTimeout' in socketOptions) {
+      options.autoSelectFamilyAttemptTimeout = socketOptions.autoSelectFamilyAttemptTimeout;
+    }
+
     const netSocket: net.Socket = new net.Socket();
```
nbbeeken, line 326: commenting here, but the code I'm thinking about is lowerdown. Should `netSocket.connect` also get this option? even though it is targeting the proxy I think you'd still want the consistent family selection behavior right? 

Does the same apply for SOCKS5 inside the `connect.ts` file? I think it does already because the socks5 negotiates over socket we will have made with the autoSelectFamily option set, but good to double check.

src/client-side-encryption/auto_encrypter.ts
```diff
@@ -101,6 +105,8 @@ export interface AutoEncryptionOptions {
   proxyOptions?: ProxyOptions;
   /** The TLS options to use connecting to the KMS provider */
   tlsOptions?: CSFLEKMSTlsOptions;
+  /** Options for KMS socket requests. */
+  socketOptions?: ClientEncryptionSocketOptions;
```
nbbeeken, line 109: Is this something we want to be seperately configurable from the client options? I can understand an argument for "yes" because this controls KMS connections vs DB connections. But since this is (at this moment) just for autoFam controls, does it make sense to keep the knob count low and only let this be configured via the client settings?

nbbeeken (changes requested):
couple of Qs

src/client-side-encryption/auto_encrypter.ts
```diff
@@ -379,7 +387,8 @@ export class AutoEncrypter {
       promoteValues: false,
```
baileympearson, line 387: (hidden in the diff) - should we also apply this option to the MongoClient instantiated in the constructor of the AutoEncrypter?

src/client-side-encryption/state_machine.ts
durran: 1. I've updated fot `netSocket.connect` to get these options now.
2. Yes the raw socket for socks5 is going through `makeSocket` which will have the options provided.

src/client-side-encryption/auto_encrypter.ts
durran: I've refactored this now to always pull the options directly from the client and remove the `socketOptions` property here.

src/client-side-encryption/state_machine.ts
durran: These all now go through a common utility method that will default it.

nbbeeken (approved):
LGTM!

baileympearson (approved):

baileympearson (approved):

PR #4189, opened 2024-08-05
fix(NODE-6284): make sparsity and trimFactor optional

baileympearson:
#### What is changing?

Sparsity and trimFactor, two options for the QE Range, have been marked optional.  All range v2 spec tests are syncced.  CSFLE prose test 23 has been implemented.

Additionally, the crypto callbacks tests have been removed.  The bindings now automatically load crypto callbacks from C++ with all compatible mongodb-client-encryption versions.  This means we can no longer spy on or sinon stub the cypto callbacks, so these tests are broken.  


test/integration/node-specific/crypto_callbacks.test.ts
```diff
@@ -1,179 +0,0 @@
...
-const kmsProviders = {
-  aws: { accessKeyId: AWS_ACCESS_KEY_ID, secretAccessKey: AWS_SECRET_ACCESS_KEY }
-};
-const dataKeyOptions = { masterKey: { key: AWS_CMK_ID, region: AWS_REGION } };
-
-const SKIP_AWS_TESTS = [AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION, AWS_CMK_ID].some(
-  secret => !secret
-);
-
-describe('cryptoCallbacks', function () {
```
W-A-James, line 25: I see the rationale for removing the tests since they're broken, but is there any way we can still test this functionality?

test/integration/client-side-encryption/client_side_encryption.spec.test.ts
```diff
@@ -97,6 +97,9 @@ describe('Client Side Encryption (Legacy)', function () {
     }
 
     if (isServerless) {
+      if (description === "Compact works with 'range' fields") {
+        return 'TODO(CLOUDP-267864): fix compactStructuredEncryptionData against serverless';
```
W-A-James, line 101: Can we file a NODE project ticket that's blocked on the CLOUDP ticket and link that instead. I feel we're more likely to action it faster if we have the NODE ticket.

W-A-James (changes requested):

test/integration/client-side-encryption/client_side_encryption.spec.test.ts
baileympearson: Done.

test/integration/node-specific/crypto_callbacks.test.ts
W-A-James: Do we think that we have sufficient coverage for these functions in `mongodb-client-encryption`?

baileympearson: Ha.  No.  We have no tests for them really in mongodb-client-encryption.

But these tests aren't any better - they also don't actually test the crypto hooks, they spy on them and assert that they're called.  Or make the spy error and assert the error is propagated to the user.

W-A-James: Maybe I'm not quite understanding, how do we know that the functions work? Are we just relying on the tests that live inside the libmongocrypt itself and hoping that the bindings just work™️?

baileympearson: The JS crypto callbacks are simple: https://github.com/mongodb-js/mongodb-client-encryption/blob/main/src/crypto_callbacks.ts

The C++ crypto callbacks are basically copy-pasting code from libmongocrypt, with some minor modifications: https://github.com/mongodb-js/mongodb-client-encryption/blob/main/addon/openssl-crypto.cc and https://github.com/mongodb/libmongocrypt/blob/18cb9e4e900c45c0b9b71fd34e159f8cb29fe1de/src/crypto/libcrypto.c.

They're implicitly covered with all our FLE tests.  We never invoke these functions directly - they're called from libmongocrypt when we invoke FLE operations.  Our existing FLE test suite implicitly covers the crypto hooks.

We probably could test them explicitly, although I personally think that there are diminishing returns to testing them directly.  Outside of the happy path (successfully getting an encrypted or decrypted value), testing other scenarios is involved for marginal benefit (without reworking our c++ crypto callbacks, for example, I'm not sure how to test anything but the happy path.  how do we make opensll give us the errors we expect?).

W-A-James (approved):

src/client-side-encryption/client_encryption.ts
```diff
@@ -948,12 +956,13 @@ export interface ClientEncryptionRewrapManyDataKeyResult {
...
- * min, max, sparsity, and range must match the values set in the encryptedFields of the destination collection.
+ * min, max, sparsity, trimFactor and range must match the values set in the encryptedFields of the destination collection.
  * For double and decimal128, min/max/precision must all be set, or all be unset.
  */
 export interface RangeOptions {
   min?: any;
   max?: any;
-  sparsity: Long;
+  sparsity?: Long;
+  trimFactor?: Int32;
```
nbbeeken, line 966: Why is this strictly the Int32 type, couldn't `number` also be allowed? Same for sparsity and `bigint` right?

test/integration/client-side-encryption/client_side_encryption.prose.23.range_encryption_defaults.test.ts
```diff
@@ -0,0 +1,144 @@
+import { type Binary, EJSON, Int32, Long } from 'bson';
```
nbbeeken, line 1: We should import BSON from the driver here, to match our expectation of the user's usage

nbbeeken (changes requested):
We're putting the release highlights of this change into: https://github.com/mongodb/node-mongodb-native/pull/4141 correct? Could you fill those out with the changes we're making here?

test/integration/client-side-encryption/client_side_encryption.prose.23.range_encryption_defaults.test.ts
baileympearson: That's not something we enforce in our tests, just searching `from 'bson'` yields a bunch of results.  I'm also not convinced of its value tbh.  But regardless, unless you'd like me to fix all occurrences of importing from bson directly in this PR, I'm going to leave this as-is.

src/client-side-encryption/client_encryption.ts
baileympearson: Yeah.  I chose to use the exact BSON types because 1) the spec is particular about the exact BSON types, so we should be as well. Ensuring that users take the extra few characters to explicitly provide the BSON type we expect seems fine.

baileympearson:
> We're putting the release highlights of this change into: #4141 correct? Could you fill those out with the changes we're making here?

This is a fix to changes made in another PR that hasn't been released.  There are no release notes for this - as far as users are concerned, this work never happened.

src/client-side-encryption/client_encryption.ts
nbbeeken: Arguably fair for `number` but it's more than a few characters, its an object vs a number, it won't respond to `typeof` or other operations the same way a number would. 

I suggest we consider validating, wraping in an Int32 or, similar to how we treat server-side values, allow it to passthrough and fail as an incorrect type. In general, we define intuitive and idiomatic APIs for developing applications with MongoDB, so we want to meet users where they are rather than requiring them to write code specific to our domain (unless necessary). 

However, for bigint, I don't think the same inexactness applies, it can only ever be an int64

I imagine these values work without modification, the question is whether our Typescript is telling the truth about the reality of what works or what we chose to support, is that correct?

test/integration/client-side-encryption/client_side_encryption.prose.23.range_encryption_defaults.test.ts
nbbeeken: Our tests should match the way users use the API, for now there's no bug and I am sure we have some mistakes around our tests where we don't catch this. But we ask our users to use the BSON exported from the driver for correctness reasons the tests should match otherwise they don't reproduce the expected use case. 

I think it is acceptable to just address the new code added in the PR, but feel free to fix more than asked if you feel so inclined.

nbbeeken:
> This is a fix

Maybe I am overlooking something in this PR. Is it not the case that `trimFactor?: Int32;` is a new line of code? I do not see `trimFactor` in `src` on main. 

I figure that we can still call this a "fix" PR but that this "feature" is a logical part of #4141 in the sense that it is "adding support for range v2"

baileympearson:
> > This is a fix
> 
> Maybe I am overlooking something in this PR. Is it not the case that `trimFactor?: Int32;` is a new line of code? I do not see `trimFactor` in `src` on main.
> 
> I figure that we can still call this a "fix" PR but that this "feature" is a logical part of #4141 in the sense that it is "adding support for range v2"

I guess that was an oversight.  The full feature will be documented in https://github.com/mongodb/node-mongodb-native/pull/4190 when the time comes.

src/client-side-encryption/client_encryption.ts
baileympearson: Ensuring users provide the correct type helps both us and users - we can be sure that the user has provided the type we expect, and users see a self-documenting API that doesn't let them provide incorrect data.  Even if primitives do work, it's a more explicit and clear user experience if we ensure that users provide us the type of data we expect.

nbbeeken: The best documentation is actual documentation, if this type concern is critical, we should add some prose that informs them. Prose can be translated and disseminated more broadly than a compiler error that _only implies_ something about a part of the functionality relating to this encryption API.

For the `number` scenario, I disagree with introducing artificial limitations when code already automatically supports a broader set of inputs that can be tested and maintained by passing both types, it would only be one more case to cover.

For the `bigint` scenario, what is unclear, unspecific, or unexpected about that input?

baileympearson: > The best documentation is actual documentation, if this type concern is critical, we should add some prose that informs them. Prose can be translated and disseminated more broadly than a compiler error that only implies something about a part of the functionality relating to this encryption API.

I fundamentally disagree with this.  Types are documentation and they are inherently superior to prose/doc comments.  Prose easily gets stale and out-of-date - types are self documenting.  I don't just mean compiler errors - when a user sees a function signature with concrete types in it, they know exactly what they're expected to provide.  In this scenario, an `Int32` clearly communicates that the type is numeric, integral and has to be inside the range of a signed, 32bit integer.  

Furthermore, w.r.t. number and bigint, we would need to provide additional doc comments and test cases.  Numbers can have decimal places or be out of range of int32.  Bigints can be greater than Longs.  All of this would require explicit documentation and potentially additional test cases.

test/integration/client-side-encryption/client_side_encryption.prose.23.range_encryption_defaults.test.ts
baileympearson: Done.

src/client-side-encryption/client_encryption.ts
nbbeeken: Prose documentation is not instead of accurate types we want to have and maintain both because they help us as well as our users. Explicit documentation should be added regardless of the input type.

We do not ask for the `Int32` boxed type anywhere else in the driver despite there being a number of options that can also only be within this bit size. `bigint` does not share the same problem that `numbers` do with respect to how they are handled in BSON, while the bigint may be larger, it can never be serialized to more or less than a 64 bit number. 

I found what I think is a [comparable example](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/crypt/explicit_encryption_context.rb#L139-L143) ruby changes the input to a boxed type, maybe it throws, maybe it coerces, but they're aligning with the common driver goal of providing an idiomatic API for their users.

W-A-James: I feel using the BSON types here is appropriate as a means of making the API more specific and explicitly spec-compliant, but regardless of the option we go with the types, we should still have some prose documentation to specify the valid range of values. e.g. can the `trimFactor` be negative, or is it limited to zero and positive integers?

src/client-side-encryption/client_encryption.ts
```diff
@@ -956,13 +964,19 @@ export interface ClientEncryptionRewrapManyDataKeyResult {
...
 export interface RangeOptions {
+  /** min is the minimum value for the encrypted index. Required if precision is set. */
   min?: any;
+  /** max is the minimum value for the encrypted index. Required if precision is set. */
   max?: any;
-  sparsity: Long;
+  /** sparsity may be used to tune performance. must be non-negative. When omitted, a default value is used. */
+  sparsity?: Long | bigint | number;
+  /** trimFactor may be used to tune performance. must be non-negative. When omitted, a default value is used. */
+  trimFactor?: Int32 | bigint | number;
```
nbbeeken, line 978:
```suggestion
  /** sparsity may be used to tune performance. must be non-negative. When omitted, a default value is used. */
  sparsity?: Long | bigint;
  /** trimFactor may be used to tune performance. must be non-negative. When omitted, a default value is used. */
  trimFactor?: Int32 | number;
```

I'm not sure `bigint` should be permitted for `trimFactor` and `number` for `sparsity`. The difference with those is that `sparsity` must be an `int64` and the only way to accomplish that reliably is with `Long` or `bigint`, you can get an int64 out of a number but it must be very large, seems unlikely that would be a desirable input, but we can leave that one as is if you feel strongly. For `trimFactor` it can be no larger than `int32.max` so only supporting `number` or `Int32` makes sense since bigint will always fail. 

Let's remove `bigint` from `trimFactor` and add tests for these input types. Docs look great, thanks!

nbbeeken (changes requested):

nbbeeken:
FYI, there's a lint error in the test file.

src/client-side-encryption/client_encryption.ts
baileympearson: I'll adjust the types, sure.  But I don't think testing is necessary here - as you've pointed out, this is just standard coercion we're letting users use and if it's not allowed, the server will error.

baileympearson: > For trimFactor it can be no larger than int32.max so only supporting number or Int32 makes sense since bigint will always fail.
Do you mean that it always fails because bigints serialize to Longs?  bigints do fit inside the range of int32s.

nbbeeken: Yes bigints are always serialized as a 64-bit int, so unless we coerce the value to Int32/Number then it will never have the correct BSON type, so I think it would make sense to not list it.

nbbeeken: Some of it is validated inside libmongocrypt, but I think that's equivalently external and authoritative validation that is better to rely on. I was hoping for some happy path coverage of these types, the tests only use Long and Int32, we can just dupe one of them to use bigint/number too.

baileympearson: Okay, I've added tests.

nbbeeken (approved):

nbbeeken (approved):

nbbeeken (approved):

W-A-James (approved):

PR #4192, opened 2024-08-07
feat(NODE-6312): add error transformation for server timeouts

nbbeeken:
#### What is changing?

- When the server returns a max time error we turn it into a client side timeout error.

Not for this PR. General CSOT documentation should cover the cases where a timeout error is encountered

When CSOT is enabled the goal is to provide a consistent mechanism for timing out operations. Ususally a client side timeout should expire before a server is able to, but depending on the variance in RTT calculation it is possible for the maxTimeMS setting that is sent to the server to be shorter than what remains on the client side. In this case, if we get a server timeout we want to throw the same error that would normally occur in CSOT circumstances.

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -614,6 +614,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
     for await (const document of this.sendCommand(ns, command, options, responseType)) {
       if (options.timeoutContext?.csotEnabled()) {
         if (MongoDBResponse.is(document)) {
+          // TODO(NODE-5684): test coverage to be added once cursors are enabling CSOT
           if (document.isMaxTimeExpiredError) {
             throw new MongoOperationTimeoutError('Server reported a timeout error', {
               cause: new MongoServerError(document.toObject())
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -621,7 +621,10 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
           }
         } else {
           if (
-            document?.writeErrors?.[0]?.code === MONGODB_ERROR_CODES.MaxTimeMSExpired ||
+            (Array.isArray(document?.writeErrors) &&
+              document.writeErrors.some(
+                error => error?.code === MONGODB_ERROR_CODES.MaxTimeMSExpired
+              )) ||
             document?.writeConcernError?.code === MONGODB_ERROR_CODES.MaxTimeMSExpired
           ) {
             throw new MongoOperationTimeoutError('Server reported a timeout error', {
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -15,6 +15,7 @@ import {
 } from '../constants';
 import {
   MongoCompatibilityError,
+  MONGODB_ERROR_CODES,
   MongoMissingDependencyError,
   MongoNetworkError,
   MongoNetworkTimeoutError,
@@ -537,6 +538,11 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
         }
 
         if (document.ok === 0) {
+          if (options.timeoutContext?.csotEnabled() && document.isMaxTimeExpiredError) {
+            throw new MongoOperationTimeoutError('Server reported a timeout error', {
+              cause: new MongoServerError((object ??= document.toObject(bsonOptions)))
+            });
+          }
           throw new MongoServerError((object ??= document.toObject(bsonOptions)));
         }
 
@@ -606,6 +612,25 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   ): Promise<Document> {
     this.throwIfAborted();
     for await (const document of this.sendCommand(ns, command, options, responseType)) {
+      if (options.timeoutContext?.csotEnabled()) {
+        if (MongoDBResponse.is(document)) {
+          if (document.isMaxTimeExpiredError) {
+            throw new MongoOperationTimeoutError('Server reported a timeout error', {
+              cause: new MongoServerError(document.toObject())
+            });
+          }
+        } else {
+          if (
+            document?.writeErrors?.[0]?.code === MONGODB_ERROR_CODES.MaxTimeMSExpired ||
+            document?.writeConcernError?.code === MONGODB_ERROR_CODES.MaxTimeMSExpired
+          ) {
+            throw new MongoOperationTimeoutError('Server reported a timeout error', {
+              cause: new MongoServerError(document)
+            });
+          }
+        }
+      }
+
       return document;
     }
     throw new MongoUnexpectedServerResponseError('Unable to get response from server');
--- a/src/cmap/wire_protocol/responses.ts
+++ b/src/cmap/wire_protocol/responses.ts
@@ -8,7 +8,7 @@ import {
   pluckBSONSerializeOptions,
   type Timestamp
 } from '../../bson';
-import { MongoUnexpectedServerResponseError } from '../../error';
+import { MONGODB_ERROR_CODES, MongoUnexpectedServerResponseError } from '../../error';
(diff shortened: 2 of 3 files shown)


W-A-James (changes requested):

src/cmap/wire_protocol/responses.ts
```diff
@@ -104,6 +104,27 @@ export class MongoDBResponse extends OnDemandDocument {
...
+   * - ok is 1 and the writeConcern object contains a code === 50
+   */
+  get isMaxTimeExpiredError() {
+    return (
+      // {ok: 0, code: 50 ... }
+      (this.ok === 0 && this.code === MONGODB_ERROR_CODES.MaxTimeMSExpired) ||
+      // {ok: 1, writeErrors: [{code: 50 ... }]}
+      (this.ok === 1 &&
+        this.get('writeErrors', BSONType.array)?.get(0, BSONType.object)?.getNumber('code') ===
+          MONGODB_ERROR_CODES.MaxTimeMSExpired) ||
```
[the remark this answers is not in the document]
nbbeeken: Possibly, but I am getting conflicting results in terms of reproducing this error and whether or not it is possible (linked you to a thread on slack). I kinda want a real example before I continue programming defensively against something we haven't seen.

But it wouldn't be hard to implement, not sure what is best. 🤔 Thoughts given that context?

src/cmap/connection.ts
```diff
@@ -606,6 +612,28 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
     this.throwIfAborted();
     for await (const document of this.sendCommand(ns, command, options, responseType)) {
+      if (options.timeoutContext?.csotEnabled()) {
+        if (MongoDBResponse.is(document)) {
+          if (document.isMaxTimeExpiredError) {
+            throw new MongoOperationTimeoutError('Server reported a timeout error', {
+              cause: new MongoServerError(document.toObject())
+            });
+          }
+        } else {
```
W-A-James, line 622: Can we add a TODO comment for NODE-5684 to add tests covering this branch

W-A-James (changes requested):

src/cmap/connection.ts
nbbeeken: Added the TODO and updated the ticket to mention it.

src/cmap/connection.ts
```diff
@@ -537,6 +538,11 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
         }
 
         if (document.ok === 0) {
+          if (options.timeoutContext?.csotEnabled() && document.isMaxTimeExpiredError) {
+            throw new MongoOperationTimeoutError('Server reported a timeout error', {
+              cause: new MongoServerError((object ??= document.toObject(bsonOptions)))
```
aditi-khare-mongoDB, line 543:
```suggestion
              cause: new MongoServerError(document.toObject(bsonOptions))
```
Before this, object is only assigned to `null`, what's the reasoning behind adding `object ??= ..` ?

src/cmap/wire_protocol/responses.ts
```diff
@@ -104,6 +104,40 @@ export class MongoDBResponse extends OnDemandDocument {
...
+    const isWriteConcern =
+      this.get('writeConcernError', BSONType.object)?.getNumber('code') ===
+      MONGODB_ERROR_CODES.MaxTimeMSExpired;
+    if (isWriteConcern) return true;
+
+    const writeErrors = this.get('writeErrors', BSONType.array);
+    if (writeErrors?.size()) {
+      for (let i = 0; i < writeErrors.size(); i++) {
+        const isWriteError =
+          writeErrors.get(i, BSONType.object)?.getNumber('code') ===
```
aditi-khare-mongoDB, line 130: I'm not that familiar with wire protocols, but is the reason we're using `get` calls rather than just `i?.code` that we're in a wire protocol?

src/cmap/wire_protocol/responses.ts
```diff
@@ -104,6 +104,40 @@ export class MongoDBResponse extends OnDemandDocument {
...
+
+    // {ok: 1, writeConcernError: {code: 50 ... }}
+    const isWriteConcern =
+      this.get('writeConcernError', BSONType.object)?.getNumber('code') ===
+      MONGODB_ERROR_CODES.MaxTimeMSExpired;
+    if (isWriteConcern) return true;
+
+    const writeErrors = this.get('writeErrors', BSONType.array);
+    if (writeErrors?.size()) {
+      for (let i = 0; i < writeErrors.size(); i++) {
```
aditi-khare-mongoDB, line 128: can we rename i to `err` something more descriptive?

aditi-khare-mongoDB (changes requested):

aditi-khare-mongoDB:
Bookkeeping, I'm taking over Warren's review while he's OOO.

src/cmap/connection.ts
nbbeeken: This caches the work done by toObject so it is not repeated. Once we've decoded the BSON once we won't again, for command monitoring.

src/cmap/wire_protocol/responses.ts
nbbeeken: `writeErrors` here is an `OnDemandDocument` so we need to call `get` to fetch values from inside the document. That `get` call also returns a document so we call getNumber to get the code property from that document.

src/cmap/wire_protocol/responses.ts
nbbeeken: `i` is the index into the `writeErrors` array whereas `err` might imply I am doing something like a for-of loop where the variable is the value inside the array rather than the index. `i` is pretty typical for indexing, the only other name that would fit would be `index` or `offset` but seems like overkill given the context, no?

aditi-khare-mongoDB: Other possible names I was thinking were `currErr` or `currWriteErr`? I don't feel too strongly it just was less easy to grok for me with the current naming.

src/cmap/connection.ts
aditi-khare-mongoDB: In the case of this function, we reassign 'object = null' at the start of every loop, and it's only used once if the error is thrown (since the `throw` statement is not wrapped in a `try-catch`). In my understanding, it doesn't make sense to cache in this way if we're just going to throw the error anyways.

nbbeeken: You're correct, but it is not something I changed in this PR I would prefer to leave it put because of the consistency it visually provides when looking at the function. This is a holdover until we get the entire driver migrated to using MongoDBResponse, until then this code block is carefully written to prevent us from double serializing every message. 

So that's my feeling, If you feel strongly I will change it

src/cmap/wire_protocol/responses.ts
nbbeeken: Both of those names imply the same thing I was mentioning, that the variable is assigned to the "current error" rather than the index of the error. Is there a name that relates to that detail that you would prefer?

aditi-khare-mongoDB: Usually I think `currentBlank` implies theres multiple `Blanks` and this is one of them. 

We can stick with `i` if no specific var name gets the point across super clearly - since it is a small loop body anyway.

src/cmap/connection.ts
aditi-khare-mongoDB: I see what you're saying about visual similarity / wanting this error to look the same as the other one. I'm good with keeping it as is.

aditi-khare-mongoDB (approved):

src/cmap/wire_protocol/responses.ts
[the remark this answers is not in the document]
aditi-khare-mongoDB: Bookkeeping: Likely, the error will be caught on the first index. In the unknown case it won't be, we're looping through the whole writeErrors array.

PR #4196, opened 2024-08-15
chore(NODE-6268): update dev dependencies and add dependabot config

baileympearson:
#### What is changing?

This is a pretty "large" PR - large in terms of files changed, not large in terms of actual substantial changes.  This PR

- performs the quarterly dependency upgrade
- upgrades to eslint v9, upgrades all other eslint-related packages to their eslint-v9 corresponding versions and fixes all the issues that arose
- upgrades prettier to 3.x (necessary for compatibility with eslint v9 and eslint-plugin-prettier), which results in a minor reformatting of the codebase
- adds a dependabot config for future dependency upgrades

I've tried to comment on any substantial changes made as a result of the eslint upgrade (spoiler: there aren't many).

No.


.eslintrc.json
```diff
@@ -276,7 +293,18 @@
       ],
       "parser": "@typescript-eslint/parser",
       "rules": {
-        "unused-imports/no-unused-imports": "error"
+        "unused-imports/no-unused-imports": "error",
+        "@typescript-eslint/no-empty-object-type": "off",
+        "@typescript-eslint/no-unsafe-declaration-merging": "off",
+        "no-duplicate-imports": "off",
+        "@typescript-eslint/consistent-type-imports": [
```
baileympearson, line 300: inline type imports were [added in TS 4.5](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names).  Since we technically "support" TS 4.4, we need use "separate-type-imports" instead of "inline-type-imports" so that our types are parsable on TS 4.4.

package.json
```diff
@@ -62,54 +62,52 @@
...
+    "eslint-plugin-tsdoc": "^0.3.0",
+    "eslint-plugin-unused-imports": "^4.0.0",
     "express": "^4.19.2",
     "gcp-metadata": "^5.3.0",
     "js-yaml": "^4.1.0",
     "mocha": "^10.4.0",
     "mocha-sinon": "^2.1.2",
-    "mongodb-client-encryption": "^6.1.0-alpha.0",
-    "mongodb-legacy": "^6.0.1",
+    "mongodb-client-encryption": "6.1.0-alpha.0",
```
baileympearson, line 99: I had to pin this to the alpha, CI is red until https://github.com/mongodb/node-mongodb-native/pull/4189#pullrequestreview-2243087220 merges.

src/bulk/common.ts
```diff
@@ -294,7 +294,7 @@ export class BulkWriteResult {
   }
 
   toString(): string {
-    return `BulkWriteResult(${this.result})`;
+    return `BulkWriteResult(${EJSON.stringify(this.result)})`;
```
baileympearson, line 297: I'm happy to revert this  but the new eslint rules actually pointed out that this uses the default stringification (`[object Object]`).

src/error.ts
```diff
@@ -1255,9 +1255,8 @@ export function needsRetryableWriteLabel(
       // use original top-level code from server response
       return RETRYABLE_WRITE_ERROR_CODES.has(error.result.code ?? 0);
     }
-    return RETRYABLE_WRITE_ERROR_CODES.has(
-      error.result.writeConcernError.code ?? Number(error.code) ?? 0
-    );
+    const code = error.result.writeConcernError.code ?? Number(error.code);
```
baileympearson, line 1258: eslint correctly points out that `Number(...)` never returns null/undefined, so the nullish coelescing was unused.  I believe what we want instead is a Number.isNumber check.

test/explicit-resource-management/main.test.ts
```diff
@@ -1,52 +1,57 @@
-
+import { expect } from 'chai';
```
baileympearson, line 1: This file was not linted until now because eslint's TS parser only supported TS 5.2 until typescript-eslint@8.x.

package.json
```diff
@@ -136,7 +134,7 @@
     "check:lambda": "mocha --config test/mocha_lambda.json test/integration/node-specific/examples/handler.test.js",
     "check:lambda:aws": "mocha --config test/mocha_lambda.json test/integration/node-specific/examples/aws_handler.test.js",
     "check:lint": "npm run build:dts && npm run check:dts && npm run check:eslint && npm run check:tsd",
-    "check:eslint": "eslint -v && eslint --max-warnings=0 --ext '.js,.ts' src test",
+    "check:eslint": "npm run  build:dts && ESLINT_USE_FLAT_CONFIG=false eslint -v && ESLINT_USE_FLAT_CONFIG=false eslint --max-warnings=0 --ext '.js,.ts' src test",
```
W-A-James, line 137:
```suggestion
    "check:eslint": "npm run build:dts && ESLINT_USE_FLAT_CONFIG=false eslint -v && ESLINT_USE_FLAT_CONFIG=false eslint --max-warnings=0 --ext '.js,.ts' src test",
```

test/tools/mongodb-mock/src/server.js
```diff
@@ -108,7 +108,7 @@ class MockServer extends EventEmitter {
           // this is when the driver closes connections
           // change false to true to start printing the logs,
           // keep CI logs clear for test reporting
-          false && console.warn('connection error: ', e);
+          console.warn('connection error: ', e);
```
W-A-James, line 111: Why change to always print the logs?

baileympearson: This causes a lint failure, because the condition is redundant - we never log anything.  Alternatively, I can just remove this log entirely.  Do you have a preference?

W-A-James: No preference, was just curious.

W-A-James (approved):

nbbeeken (approved):
LGTM, the no-then removals seem to have left empty lines that I would've expected prettier to clean up, fine to leave as is but mentioning if you didn't notice

PR #4205, opened 2024-08-21
feat(NODE-6327): new client bulk write types and builders

durran:
### Description

Creates the new client bulk write model types and options as well as the builders for the bulk write command.

#### What is changing?

- Creates the new client bulk write model types.
- Creates a client bulk write command builder that delegates to the various model builders to build the operations.
- Ensures the `ops` field of the bulk write command is a document sequence
- Ensures the `nsInfo` field of the bulk write command is a unique document sequence.
- Adds a bunch-o-unit-tests

None. docs changes will come on the last ticket of the epic as well as the release highlights.

NODE-6327

### DIFF
--- a/src/operations/client_bulk_write/command_builder.ts
+++ b/src/operations/client_bulk_write/command_builder.ts
@@ -1,5 +1,8 @@
 import { type Document } from '../../bson';
 import { DocumentSequence } from '../../cmap/commands';
+import type { Filter, OptionalId, UpdateFilter, WithoutId } from '../../mongo_types';
+import { type CollationOptions } from '../command';
+import { type Hint } from '../operation';
 import type {
   AnyClientBulkWriteModel,
   ClientBulkWriteOptions,
@@ -66,9 +69,7 @@ export class ClientBulkWriteCommandBuilder {
       }
     }
 
-    const nsInfo = Array.from(namespaces.keys()).map(ns => {
-      return { ns: ns };
-    });
+    const nsInfo = Array.from(namespaces.keys(), ns => ({ ns }));
 
     // The base command.
     const command: ClientBulkWriteCommand = {
@@ -83,11 +84,17 @@ export class ClientBulkWriteCommandBuilder {
       command.bypassDocumentValidation = this.options.bypassDocumentValidation;
     }
     // Add let if it was present in the options.
-    if ('let' in this.options) {
+    if (this.options.let) {
       command.let = this.options.let;
     }
     return [command];
   }
+}
+
+/** @internal */
+export interface InsertOperation<TSchema extends Document = Document> {
+  insert: number;
+  document: OptionalId<TSchema>;
 }
 
 /**
@@ -97,12 +104,21 @@ export class ClientBulkWriteCommandBuilder {
  * @returns the operation.
  */
 export const buildInsertOneOperation = (model: ClientInsertOneModel, index: number): Document => {
-  const document: Document = {
+  const document: InsertOperation = {
     insert: index,
     document: model.document
   };
   return document;
 };
+
+/** @internal */
+export interface DeleteOperation<TSchema extends Document = Document> {
+  delete: number;
+  multi: boolean;
+  filter: Filter<TSchema>;
+  hint?: Hint;
+  collation?: CollationOptions;
+}
 
 /**
  * Build the delete one operation.
@@ -131,8 +147,8 @@ function createDeleteOperation(
   model: ClientDeleteOneModel | ClientDeleteManyModel,
   index: number,
   multi: boolean
-): Document {
-  const document: Document = {
+): DeleteOperation {
+  const document: DeleteOperation = {
     delete: index,
     multi: multi,
     filter: model.filter
@@ -146,13 +162,27 @@ function createDeleteOperation(
   return document;
 }
 
+/** @internal */
+export interface UpdateOperation<TSchema extends Document = Document> {
+  update: number;
+  multi: boolean;
+  filter: Filter<TSchema>;
+  updateMods: UpdateFilter<TSchema> | Document[];
+  hint?: Hint;
+  upsert?: boolean;
+  arrayFilters?: Document[];
+}
+
 /**
  * Build the update one operation.
  * @param model - The update one model.
  * @param index - The namespace index.
  * @returns the operation.
  */
-export const buildUpdateOneOperation = (model: ClientUpdateOneModel, index: number): Document => {
+export const buildUpdateOneOperation = (
+  model: ClientUpdateOneModel,
+  index: number
+): UpdateOperation => {
   return createUpdateOperation(model, index, false);
 };
 
@@ -162,7 +192,10 @@ export const buildUpdateOneOperation = (model: ClientUpdateOneModel, index: numb
  * @param index - The namespace index.
  * @returns the operation.
  */
-export const buildUpdateManyOperation = (model: ClientUpdateManyModel, index: number): Document => {
+export const buildUpdateManyOperation = (
+  model: ClientUpdateManyModel,
+  index: number
+): UpdateOperation => {
   return createUpdateOperation(model, index, true);
 };
 
@@ -173,8 +206,8 @@ function createUpdateOperation(
   model: ClientUpdateOneModel | ClientUpdateManyModel,
   index: number,
   multi: boolean
-): Document {
-  const document: Document = {
+): UpdateOperation {
+  const document: UpdateOperation = {
     update: index,
     multi: multi,
     filter: model.filter,
@@ -192,14 +225,27 @@ function createUpdateOperation(
   return document;
 }
 
+/** @internal */
+export interface ReplaceOneOperation<TSchema extends Document = Document> {
+  update: number;
+  multi: boolean;
+  filter: Filter<TSchema>;
+  updateMods: WithoutId<TSchema>;
+  hint?: Hint;
+  upsert?: boolean;
+}
+
 /**
  * Build the replace one operation.
  * @param model - The replace one model.
  * @param index - The namespace index.
  * @returns the operation.
  */
-export const buildReplaceOneOperation = (model: ClientReplaceOneModel, index: number): Document => {
-  const document: Document = {
+export const buildReplaceOneOperation = (
+  model: ClientReplaceOneModel,
+  index: number
+): ReplaceOneOperation => {
+  const document: ReplaceOneOperation = {
     update: index,
     multi: false,
     filter: model.filter,
--- a/src/operations/client_bulk_write/command_builder.ts
+++ b/src/operations/client_bulk_write/command_builder.ts
@@ -60,8 +60,9 @@ export class ClientBulkWriteCommandBuilder {
     const namespaces = new Map<string, number>();
     for (const model of this.models) {
       const ns = model.namespace;
-      if (namespaces.has(ns)) {
-        operations.push(buildOperation(model, namespaces.get(ns) as number));
+      const index = namespaces.get(ns);
+      if (index != null) {
+        operations.push(buildOperation(model, index));
       } else {
         namespaces.set(ns, currentNamespaceIndex);
         operations.push(buildOperation(model, currentNamespaceIndex));
@@ -92,7 +93,7 @@ export class ClientBulkWriteCommandBuilder {
 }
 
 /** @internal */
-export interface InsertOperation<TSchema extends Document = Document> {
+interface ClientInsertOperation<TSchema extends Document = Document> {
   insert: number;
   document: OptionalId<TSchema>;
 }
@@ -103,8 +104,11 @@ export interface InsertOperation<TSchema extends Document = Document> {
  * @param index - The namespace index.
  * @returns the operation.
  */
-export const buildInsertOneOperation = (model: ClientInsertOneModel, index: number): Document => {
-  const document: InsertOperation = {
+export const buildInsertOneOperation = (
+  model: ClientInsertOneModel,
+  index: number
+): ClientInsertOperation => {
+  const document: ClientInsertOperation = {
     insert: index,
     document: model.document
   };
@@ -112,7 +116,7 @@ export const buildInsertOneOperation = (model: ClientInsertOneModel, index: numb
 };
 
 /** @internal */
-export interface DeleteOperation<TSchema extends Document = Document> {
+export interface ClientDeleteOperation<TSchema extends Document = Document> {
   delete: number;
   multi: boolean;
   filter: Filter<TSchema>;
@@ -147,8 +151,8 @@ function createDeleteOperation(
   model: ClientDeleteOneModel | ClientDeleteManyModel,
   index: number,
   multi: boolean
-): DeleteOperation {
-  const document: DeleteOperation = {
+): ClientDeleteOperation {
+  const document: ClientDeleteOperation = {
     delete: index,
     multi: multi,
     filter: model.filter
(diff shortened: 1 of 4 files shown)


src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,302 @@
...
+    }
+    // Add let if it was present in the options.
+    if ('let' in this.options) {
+      command.let = this.options.let;
+    }
+
+    // Iterate the models to build the ops and nsInfo fields.
+    const operations = [];
+    let currentNamespaceIndex = 0;
+    const namespaces = new Map<string, number>();
```
durran, line 62: JS Maps retain insertion order so the index will be correct when we add a new ns to the map.

durran, line 62: I chose to use a map here to track the namespace indexes since they retain insertion order when adding new elements.

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  /**
+   * Build the operation.
+   * @param index - The namespace index.
+   * @returns the operation.
+   */
+  buildOperation(index: number): Document {
+    const document: Document = {
+      insert: index,
+      document: this.model.document
```
baileympearson, line 111: We likely need to autogenerate an object id using our `maybeAddIdToDocuments` helper (although, this method currently accepts a collection object - we'll have to change that :) ).

That also leads me to a second question - should we support CRUD-method specific options too?  For example - we support `pkFactory` and `forceServerObjectId` to insert operations.  We should likely support those for the new bulk write too.  Although pkFactory is tricky - because we usually get the pkFactory from the collection, which we don't have.

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,302 @@
...
+    }
+    if (this.model.upsert) {
+      document.upsert = this.model.upsert;
+    }
+    return document;
+  }
+}
+
+const BUILDERS: Map<string, (model: AnyClientBulkWriteModel) => OperationBuilder> = new Map();
+BUILDERS.set('insertOne', model => new InsertOneOperationBuilder(model as ClientInsertOneModel));
```
baileympearson, line 288: If you want to avoid type-casting the model - the typical pattern in TS is to use a switch statement on the discriminator in your union type, which TS uses to narrow the type for you:

```typescript
export function builder(model: AnyClientBulkWriteModel): OperationBuilder {
  switch (model.name) {
    case 'insertOne':
      return new InsertOneOperationBuilder(model);
    // ... TS can auto-generate all possible cases for the switch
    default:
      throw new MongoInvalidArgumentError(`Could not load builder for model ${model.name}`);
  }
}
```

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,302 @@
...
+   * @param index - The namespace index.
+   * @returns the operation.
+   */
+  buildOperation(index: number): Document {
+    return createUpdateOperation(this.model, index, true);
+  }
+}
+
+/**
+ * Creates a delete operation based on the parameters.
```
baileympearson, line 228:
```suggestion
 * Creates an update operation based on the parameters.
```

src/operations/client_bulk_write/common.ts
```diff
@@ -0,0 +1,133 @@
...
+  collation?: CollationOptions;
+  /** The index to use. If specified, then the query system will only consider plans using the hinted index. */
+  hint?: Hint;
+}
+
+/** @public */
+export interface ClientDeleteManyModel<TSchema extends Document = Document>
+  extends ClientWriteModel {
+  name: 'deleteMany';
+  /** The filter to limit the deleted documents. */
```
baileympearson, line 58: (same suggestion for deleteOne)
```suggestion
  /** 
   * The filter used to determine if a document should be deleted. For a deleteOne operation, the first match is removed.
  */
```

baileympearson: (also maybe a similar change for the updateOne updateMany models)

src/operations/client_bulk_write/command_builder.ts
baileympearson: edit: I found it in the spec, we do need to add _ids: https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#insert

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    // Iterate the models to build the ops and nsInfo fields.
+    const operations = [];
+    let currentNamespaceIndex = 0;
+    const namespaces = new Map<string, number>();
+    for (const model of this.models) {
+      const ns = model.namespace;
+      if (namespaces.has(ns)) {
+        operations.push(builderFor(model).buildOperation(namespaces.get(ns) as number));
+      } else {
```
baileympearson, line 67: Surprising that TS can't narrow this scenario.  I guess it's because maybe TS doesn't know that `ns` hasn't changed between the has call and the subsequent get call?  

Regardless, moving index into a standalone variable lets TS narrow this properly.
```suggestion
      const ns = model.namespace;
      const index = namespaces.get(ns);
      if (index) {
        operations.push(builderFor(model).buildOperation(index));
      } else {
```

src/operations/client_bulk_write/command_builder.ts
baileympearson: Could we just use an array, pushing into the back?

Then we can construct nsInfo with:

```typescript

    const nsInfo = namespaces.map((ns) => {
      return { ns: ns };
    });

```

baileympearson: I guess that has the tradeoff that checking whether or not the ns exists in the array is O(number of namespaces in array).  Feel free to disregard

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,302 @@
...
+    if ('verboseResults' in this.options) {
+      return !this.options.verboseResults;
+    }
+    return true;
+  }
+
+  /**
+   * Build the bulk write commands from the models.
+   */
+  buildCommands(): Document[] {
```
baileympearson, line 43: (optional) since we're starting from scratch with a new command - thoughts on typing the new bulkWrite command?

baileympearson: Doesn't have to be too strong, maybe just the top-level fields:

```
type ClientBulkWriteCommand = {
	bulkWrite: 1,
	errorsOnly: boolean,
	...
}
```

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    // Iterate the models to build the ops and nsInfo fields.
+    const operations = [];
+    let currentNamespaceIndex = 0;
+    const namespaces = new Map<string, number>();
+    for (const model of this.models) {
+      const ns = model.namespace;
+      if (namespaces.has(ns)) {
+        operations.push(builderFor(model).buildOperation(namespaces.get(ns) as number));
```
baileympearson, line 66: This creates a new object for each write model just to build the command, only to immediately discard the builders.  Thoughts on structuring this in a way that allocates fewer objects?

Two ideas:

1. create one of each model builder before the loop, and just re-assign the model and then build it.
2. use functions instead of classes:

```typescript
type Builder = (model: AnyClientBulkWriteModel, index: number) => Document;

const insertOneBuilder: Builder = (model: InsertOneModel, index: number) => { ... };
```

baileympearson (changes requested):
Looks good!  The testing made the PR straightforward to review.

src/operations/client_bulk_write/command_builder.ts
durran: I left as-is because this suggestion would cause the functionality to break since index can be `0` which would incorrectly return false here and explicitly checking for a value of undefined fails one of our eslint rules.

src/operations/client_bulk_write/command_builder.ts
durran: I've add the interface for the command now.

src/operations/client_bulk_write/command_builder.ts
durran: I've updated to use functions now instead instantiating the objects.

src/operations/client_bulk_write/command_builder.ts
durran: Fixed comments.

src/operations/client_bulk_write/command_builder.ts
durran: I've updated to use a switch on the new functions now.

src/operations/client_bulk_write/command_builder.ts
durran: Note as I mentioned I am doing _id generation in the next ticket.

src/operations/client_bulk_write/common.ts
durran: Updated all comments.

src/operations/client_bulk_write/common.ts
```diff
@@ -0,0 +1,148 @@
...
+  hint?: Hint;
+}
+
+/** @public */
+export interface ClientDeleteManyModel<TSchema extends Document = Document>
+  extends ClientWriteModel {
+  name: 'deleteMany';
+  /**
+   * The filter used to determine if a document should be deleted.
+   * For a deleteOne operation, all matches are removed.
```
baileympearson, line 63:
```suggestion
   * For a deleteMany operation, all matches are removed.
```

durran: Fixed

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,233 @@
...
+      nsInfo: new DocumentSequence(nsInfo)
+    };
+    // Add bypassDocumentValidation if it was present in the options.
+    if ('bypassDocumentValidation' in this.options) {
+      command.bypassDocumentValidation = this.options.bypassDocumentValidation;
+    }
+    // Add let if it was present in the options.
+    if ('let' in this.options) {
+      command.let = this.options.let;
+    }
```
nbbeeken, line 88: https://github.com/mongodb/node-mongodb-native/blob/NODE-6327/src/operations/aggregate.ts#L121-L127

It's a little strange to me that we've switched up the check to ingest these arguments. The code I link to isn't perfect either by only using truthiness, but this check seems to go out of its way to put `let: null` onto the command if that is present on the options. Is that intentional?

src/operations/client_bulk_write/common.ts
```diff
@@ -0,0 +1,148 @@
...
+  /** Specifies a collation. */
+  collation?: CollationOptions;
+  /** The index to use. If specified, then the query system will only consider plans using the hinted index. */
+  hint?: Hint;
+  /** When true, creates a new document if no document matches the query. */
+  upsert?: boolean;
+}
+
+/** @public */
+export type AnyClientBulkWriteModel<TSchema extends Document = Document> =
```
nbbeeken, line 142: Can we add documentation for what is `TSchema` means for this type? 

IIUC Each one of these models has a `namespace` property that can be different, that probably means there's a mixture of schemas, how does the use case for a `ClientUpdateOneModel<Books>` and `ClientDeleteOneModel<Authors>` work?

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,233 @@
...
+function createDeleteOperation(
+  model: ClientDeleteOneModel | ClientDeleteManyModel,
+  index: number,
+  multi: boolean
+): Document {
+  const document: Document = {
+    delete: index,
+    multi: multi,
+    filter: model.filter
+  };
```
nbbeeken, line 139:
```suggestion
  const document: {
    delete: number;
    multi: boolean;
    filter: Filter<TSchema>;
    hint?: Hint;
    collation?: CollationOptions;
  } = {
    delete: index,
    multi: multi,
    filter: model.filter
  };
```

Typing all the commands accurately doesn't seem to introduce any hurdles to overcome and would help future maintainers to make sure the optional assignments (like `if(hint)`) remain correct.

nbbeeken (changes requested):

src/operations/client_bulk_write/command_builder.ts
durran: I've added types now for all the operations.

src/operations/client_bulk_write/common.ts
durran: Docs added.

src/operations/client_bulk_write/command_builder.ts
durran: I updated the check for `let` to not allow null as in the example linked. Nice catch.

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,233 @@
...
+        namespaces.set(ns, currentNamespaceIndex);
+        operations.push(buildOperation(model, currentNamespaceIndex));
+        currentNamespaceIndex++;
+      }
+    }
+
+    const nsInfo = Array.from(namespaces.keys()).map(ns => {
+      return { ns: ns };
+    });
+
```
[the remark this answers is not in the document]
durran: Updated per suggestion.

src/operations/client_bulk_write/command_builder.ts
durran: Good point. I generally don't like `!=` in JS an prefer exact matching but I've updated it now to that.

src/operations/client_bulk_write/common.ts
nbbeeken:
```ts
type Book = { _id: UUID };
type Author = { _id: ObjectId };

const array: AnyClientBulkWriteModel<Book | Author> = [
  { namespace: 'db.books', name: 'insertOne', document: { _id: new ObjectId() } },
  { namespace: 'db.authors', name: 'updateOne', filter: {}, update: { $set: { _id: new UUID() } } }
];
```

(ignoring the fact that $set doesn't make sense for _id)

If I just union my schemas together then I don't get safety when there is overlap, so I don't see how in practice this API has a good DX for the typescript user. 

But we did say:

> TSchema can be different on each of the individual models must always match the appropriate namespace that it defines provided to each of the models. 

Is the intention to cast or declare separate variables? Would it be possible to "lookup" the schema the way we do event listeners if the namespace string is a literal type?

nbbeeken (changes requested):

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -0,0 +1,283 @@
...
+    // The base command.
+    const command: ClientBulkWriteCommand = {
+      bulkWrite: 1,
+      errorsOnly: this.errorsOnly,
+      ordered: this.options.ordered ?? true,
+      ops: new DocumentSequence(operations),
+      nsInfo: new DocumentSequence(nsInfo)
+    };
+    // Add bypassDocumentValidation if it was present in the options.
+    if ('bypassDocumentValidation' in this.options) {
```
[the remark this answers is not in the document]
durran: Update this as well.

src/operations/client_bulk_write/common.ts
durran: I've gone ahead and done that. Can maybe revisit options in the next ticket for the full happy path.

nbbeeken: I am okay with revisiting, can we actually file a ticket specifically for investigating the TS DX of the API, we can finalize the functionality of the API with weak types and consider possibly a type intended to be used as an annotation to opt into an autocomplete experience (like `StrictFilter`)

nbbeeken (approved):
LGTM. Keep in mind this is going in as a feat but no release notes understandably because a later PR will include the actual API. So we'll just want to remember to omit this bullet from next release.

baileympearson (approved):

baileympearson:
CI failures are unrelated.

PR #4207, opened 2024-08-23
feat(NODE-6342): support maxTimeMS for explain commands

baileympearson:
#### What is changing?

Explain commands can now have a maxTimeMS specified for them.  

Corresponds to https://github.com/mongodb/specifications/pull/1640.  The CSOT-related changes for explain helpers will be done in https://jira.mongodb.org/browse/NODE-6375.

No.

### Explain now supports maxTimeMS

Driver CRUD commands can be explained by providing the `explain` option:

```typescript
collection.find({}).explain('queryPlanner'); // using the fluent cursor API
collection.deleteMany({}, { explain: 'queryPlanner' }); // as an option
```

However, if maxTimeMS was provided, the maxTimeMS value was applied to the command to explain, and consequently the server could take more than maxTimeMS to respond.

Now, maxTimeMS can be specified as a new option for explain commands:

```typescript
collection.find({}).explain({ verbosity: 'queryPlanner', maxTimeMS: 2000 }); // using the fluent cursor API
collection.deleteMany({}, { 
	explain: {
		verbosity: 'queryPlanner',
		maxTimeMS: 2000
	}
); // as an option
```

If a top-level maxTimeMS option is provided in addition to the explain maxTimeMS, the explain-specific maxTimeMS is applied to the explain command, and the top-level maxTimeMS is applied to the explained command:

```typescript
collection.deleteMany({}, { 
	maxTimeMS: 1000,
	explain: {
		verbosity: 'queryPlanner',
		maxTimeMS: 2000
	}
);

// the actual command that gets sent to the server looks like:
{
	explain: { delete: <collection name>, ..., maxTimeMS: 1000 },
    verbosity: 'queryPlanner',
	maxTimeMS: 2000 
}
```


test/integration/crud/explain.test.ts
```diff
@@ -117,6 +119,111 @@ describe('CRUD API explain option', function () {
       });
     }
   }
+
+  describe('explain helpers w/ maxTimeMS', function () {
```
dariakp, line 123: I think in order to capture intentionality, whichever way the spec lands, we'll want to add cases for when maxTimeMS is specified as a command level option but not an explain-level option.

It may be easier to enumerate all these test cases if we use describe blocks for `describe('when maxTimeMS <is/is not> set as an explain option')` with `it('<what happens when it is/isn't set on the command-level>')` (or the reverse, with the is/is not set as command-level option in the `describe` and the explain in the `it`), we could also use nested describes to fully separate the `when` from the `what happens`, but I don't know whether that would help or hurt the readability.

test/integration/crud/explain.test.ts
```diff
@@ -117,6 +119,111 @@ describe('CRUD API explain option', function () {
...
+                verbosity: 'queryPlanner',
+                maxTimeMS: 2000
+              }
+            }
+          )
+          .toArray();
+
+        const [{ command }] = commands;
+        expect(command).to.have.property('maxTimeMS', 2000);
+      });
```
addaleax, line 187: There should probably be tests for what happens when `maxTimeMS` is only specified as a command-level option, and not as an explain option at all

src/utils.ts
```diff
@@ -251,12 +251,29 @@ export function decorateWithReadConcern(
...
+export function decorateWithExplain(
+  command: Document,
+  explain: Explain
+): {
+  explain: Document;
+  verbosity: ExplainVerbosity;
+  maxTimeMS?: number;
+} {
+  type ExplainCommand = ReturnType<typeof decorateWithExplain>;
+  const isExplainCommand = (doc: Document): doc is ExplainCommand => 'explain' in command;
```
W-A-James, line 263: Why not define these outside of this function?

test/integration/shared.js
```diff
@@ -42,6 +42,26 @@ function dropCollection(dbObj, collectionName, options = {}) {
...
+ * client.on('commandStarted', filterForCommands(['ping', 'find'], commands));
+ * // custom predicate
+ * client.on('commandStarted', filterForCommands((command) => command.commandName === 'find', commands));
+ * ```
+ * @param {string | string[] | (arg0: string) => boolean} commands A set of commands to look for.  Either
+ * a single command name (string), a list of command names (string[]) or a predicate function that
+ * determines whether or not a command should be kept.
+ * @param {Array} bag the output for the filtered commands
+ * @returns a function that collects the specified comment events
+ */
```
W-A-James, line 64: Nice addition here

test/unit/explain.test.ts
```diff
@@ -0,0 +1,53 @@
+import { expect } from 'chai';
+import { it, test } from 'mocha';
+
+import { Explain, ExplainVerbosity } from '../mongodb';
+
+describe('class Explain {}', function () {
+  describe('static .fromOptions()', function () {
+    test('when no options are provided, it returns undefined', function () {
```
W-A-James, line 8: Why the move to using `mocha.test` instead of `mocha.it`. I know they're aliases for each other, but we typically stick with the [BDD patterns](https://mochajs.org/#bdd) as opposed to mixing with the [TDD patterns](https://mochajs.org/#tdd).

test/unit/utils.test.ts
```diff
@@ -1003,4 +1006,42 @@ describe('driver utils', function () {
     describe('when given an object that does not respond to Symbol.toStringTag', () =>
       it('returns false', () => expect(isUint8Array(Object.create(null))).to.be.false));
   });
+
+  describe('decorateWithExplain()', function () {
+    test('when the command is a valid explain command, the command is returned unmodified', function () {
```
W-A-James, line 1011: Same question as above

W-A-James (changes requested):

src/utils.ts
baileympearson: They're an implementation detail and aren't used anywhere else, so they shouldn't be available anywhere else.

test/unit/explain.test.ts
baileympearson: I usually use test when I specify both the context and behavior in one string because "it when no options are provided, it returns undefined" doesn't sound right.  I can change these if you'd like though

nbbeeken: a practical reason to stick with `it` is our lint rule won't work on this block

https://github.com/lo1tuma/eslint-plugin-mocha/blob/main/lib/rules/consistent-spacing-between-blocks.js#L4-L11

baileympearson: Oh, good point.  

I adjusted both files to use `it`.

W-A-James (dismissed):

src/explain.ts
```diff
@@ -20,33 +18,72 @@ export type ExplainVerbosity = string;
...
     } else {
       this.verbosity = verbosity;
     }
+
+    this.maxTimeMS = maxTimeMS;
   }
 
-  static fromOptions(options?: ExplainOptions): Explain | undefined {
-    if (options?.explain == null) return;
+  static fromOptions({ explain }: ExplainOptions = {}): Explain | undefined {
```
nbbeeken, line 79: If I pass null here it throws where it didn't before, is it possible that null could've been passed in here?

src/explain.ts
```diff
@@ -20,33 +18,72 @@ export type ExplainVerbosity = string;
...
+ *
+ * ```typescript
+ * // limits the `explain` command to no more than 2 seconds
+ * collection.find({ name: 'john doe' }, { explain:
+ *  {
+ *    verbosity: 'queryPlanner',
+ *    maxTimeMS: 2000
+ *  }
+ * });
+ * ```
```
nbbeeken, line 55:
```
// limits the `explain` command to no more than 2 seconds
collection.find({ name: 'john doe' }, { explain: { verbosity: 'queryPlanner', maxTimeMS: 2000 } });
```

just offering: prettier reformats this to the above. 

The line break after explain rather than before made this a bit weird to read at first.

src/utils.ts
```diff
@@ -251,12 +251,29 @@ export function decorateWithReadConcern(
...
-  if (command.explain) {
+export function decorateWithExplain(
+  command: Document,
+  explain: Explain
+): {
+  explain: Document;
+  verbosity: ExplainVerbosity;
+  maxTimeMS?: number;
+} {
+  type ExplainCommand = ReturnType<typeof decorateWithExplain>;
```
nbbeeken, line 262: This type changes to whatever I make this function return, the self referencing means I don't really get any warning if I do something unexpected with the return type.

src/utils.ts
nbbeeken: Why is the condition now explain "in" instead of explain truthy?

test/integration/crud/crud.prose.test.ts
```diff
@@ -143,4 +150,42 @@ describe('CRUD Prose Spec Tests', () => {
...
+      client.on('commandStarted', filterForCommands('explain', commands));
+      commands.length = 0;
+    });
+
+    afterEach(async function () {
+      await client.close();
+    });
+
+    it('sets maxTimeMS on explain commands, when specfied', async function () {
+      await collection
```
nbbeeken, line 175:
```suggestion
    it('sets maxTimeMS on explain commands, when specified', async function () {
```

nbbeeken (changes requested):

src/explain.ts
baileympearson: No.  This is called in exactly one place, with either an object or undefined.

src/utils.ts
baileympearson: I don't care about the value, only that the key is present.

test/integration/crud/crud.prose.test.ts
```diff
@@ -143,4 +150,42 @@ describe('CRUD Prose Spec Tests', () => {
...
+
+      client.on('commandStarted', filterForCommands('explain', commands));
+      commands.length = 0;
+    });
+
+    afterEach(async function () {
+      await client.close();
+    });
+
+    it('sets maxTimeMS on explain commands, when specfied', async function () {
```
nbbeeken, line 174:
```suggestion
    it('sets maxTimeMS on explain commands, when specified', async function () {
```

src/utils.ts
baileympearson: That's not the point of this type.  It's only here to satisfy the return case when we already have a key of `explain` in the document.

Two other options, if you'd prefer:
- remove this and the type assertion, revert the logic change and just type cast this as-any
- remove this logic entirely.  it's unused (we never decorate a command twice).  I chose to leave it in because it seemed like a nice safeguard in case that changes in the future.

src/utils.ts
nbbeeken: What changed from before that we no longer care about the truthiness of the value? Doesn't an explain always have to be a document so it needs to be a non-null object or Map? It appears as though this has a functional change

baileympearson: Sure, it's a change in behavior.  If you see my response to your comment above though, I've determined that it's safe to make.  We never decorate explain commands twice, we could alternatively just remove this logic.

baileympearson: But also, not validating the type of this field aligns with our philosophy of not validating user input and instead letting the server error.

src/utils.ts
nbbeeken: It depends on what is important to maintain 🤔 maybe casting makes sense in order to not change the existing functionality. But if it is true that we never can rewrap, then maybe this isn't necessary. 

Maybe this was added if explain gets retried?

src/utils.ts
baileympearson: @nbbeeken I left the type in, only because its useful when typing the `command` being built.  Since I initialize it to a literal, TS infers the type as { explain: Document, verbosity: string }.  If I don't type it like this (or as any or Document or something), TS complains that `maxTimeMS` isn't allowed on the command when I attach it.

nbbeeken (approved):

W-A-James (approved):

PR #4210, opened 2024-08-23
fix(NODE-6346): import performance from perf_hooks for better nodejs compatibility

dougludlow:
#### What is changing?

Importing `performance` from `perf_hooks`.

No.

In Node.js app bundled with webpack, when using `mongodb` to connect to a database the following error occurs:

```
TypeError: Cannot read properties of undefined (reading 'now')
    at new Timeout (/path/to/project/node_modules/.pnpm/mongodb@6.7.0_socks@2.8.3/node_modules/mongodb/src/timeout.ts:58:41)
```

Importing `performance` from `perf_hooks` provides better compatibility.

### DIFF
--- a/src/timeout.ts
+++ b/src/timeout.ts
@@ -1,3 +1,5 @@
+import { performance } from 'node:perf_hooks';
+
 import { clearTimeout, setTimeout } from 'timers';
 
 import { MongoInvalidArgumentError } from './error';


baileympearson:
As per the discussion in https://jira.mongodb.org/browse/NODE-6346, we will continue to use the global performance object as its available as a global in all of our supported runtimes, and other server-side JS runtimes (Deno and Bun).

PR #4214, opened 2024-08-30
fix(NODE-6355): respect utf8 validation options when iterating cursors

baileympearson:
#### What is changing?

This PR ensures that any BSON utf8 validation options are provided to the deserializer when deserializing cursors.

Since on-demand BSON, cursors are deserialized on-demand.  Documents are pulled out of the cursor using `MongoDBResponse.get()`, which returns an unparsed on-demand document.  We then deserialize it using `toObject()`.

However, although `MongoDBResponse` correctly applies BSON utf8 options to its deserialization logic, the on-demand document does not.  This results in documents retrieved from a cursor being deserialized without the utf8 validation options.

### DIFF
--- a/.evergreen/install-mongodb-client-encryption.sh
+++ b/.evergreen/install-mongodb-client-encryption.sh
@@ -10,7 +10,7 @@ set -o xtrace   # Write all commands first to stderr
 set -o errexit  # Exit the script with error if any of the commands fail
 
 rm -rf mongodb-client-encryption
-git clone https://github.com/mongodb-js/mongodb-client-encryption.git
+git clone https://github.com/mongodb-js/mongodb-client-encryption.git -b explicit-lifetime-chaining
 pushd mongodb-client-encryption
 
 if [ -n "${LIBMONGOCRYPT_VERSION}" ]; then
--- a/.evergreen/install-mongodb-client-encryption.sh
+++ b/.evergreen/install-mongodb-client-encryption.sh
@@ -10,7 +10,7 @@ set -o xtrace   # Write all commands first to stderr
 set -o errexit  # Exit the script with error if any of the commands fail
 
 rm -rf mongodb-client-encryption
-git clone https://github.com/mongodb-js/mongodb-client-encryption.git -b explicit-lifetime-chaining
+git clone https://github.com/mongodb-js/mongodb-client-encryption.git
 pushd mongodb-client-encryption
 
 if [ -n "${LIBMONGOCRYPT_VERSION}" ]; then
--- a/src/cmap/wire_protocol/on_demand/document.ts
+++ b/src/cmap/wire_protocol/on_demand/document.ts
@@ -47,6 +47,13 @@ type CachedBSONElement = { element: BSONElement; value: any | undefined };
 /** @internal */
 export class OnDemandDocument {
   /**
+   * @internal
+   *
+   * Used for testing purposes.
+   */
+  private static BSON: typeof BSON = BSON;
+
+  /**
    * Maps JS strings to elements and jsValues for speeding up subsequent lookups.
    * - If `false` then name does not exist in the BSON document
    * - If `CachedBSONElement` instance name exists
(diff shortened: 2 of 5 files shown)


test/integration/node-specific/bson-options/utf8_validation.test.ts
```diff
@@ -153,3 +156,177 @@ describe('class MongoDBResponse', () => {
...
+   * bytes of the character 'é', to produce invalid utf8.
+   */
+  async function insertDocumentWithInvalidUTF8() {
+    const stub = sinon.stub(net.Socket.prototype, 'write').callsFake(function (...args) {
+      const providedBuffer = args[0].toString('hex');
+      const targetBytes = Buffer.from('é').toString('hex');
+
+      if (providedBuffer.includes(targetBytes)) {
+        if (providedBuffer.split(targetBytes).length !== 2) {
+          throw new Error('received buffer more than one `c3a9` sequences.  or perhaps none?');
```
nbbeeken, line 176:
```suggestion
          sinon.restore();
          throw new Error('received buffer more than one `c3a9` sequences. or perhaps none?');
```

Maybe this is why we didn't see the error?

test/integration/node-specific/bson-options/utf8_validation.test.ts
```diff
@@ -153,3 +156,177 @@ describe('class MongoDBResponse', () => {
...
+    await client.connect();
+    const db = client.db('test');
+    collection = db.collection('invalidutf');
+
+    await collection.deleteMany({});
+    await insertDocumentWithInvalidUTF8();
+  });
+
+  afterEach(async function () {
+    await client.close();
```
nbbeeken, line 207:
```suggestion
    sinon.restore();
    await client.close();
```

Just to be extra sure we revert to normal.

nbbeeken (changes requested):

test/integration/node-specific/bson-options/utf8_validation.test.ts
baileympearson: Once this method returns / errors, we'll still have to continue up the sinon callstack back into the driver, regardless of whether or not the stub has been restored.

nbbeeken (changes requested):
The `c3a9` condition is failing. Perhaps we need to be a bit more robust, searching for `c3a9` after finding `"insert"` or `"documents"` so we know we're in the command.

jia-zhen:
Please update and release it as soon as possible, as we are currently troubled by this issue. / Please update and publish it promptly, as we are being bothered by this problem.

.evergreen/install-mongodb-client-encryption.sh
```diff
@@ -10,7 +10,7 @@ set -o xtrace   # Write all commands first to stderr
 set -o errexit  # Exit the script with error if any of the commands fail
 
 rm -rf mongodb-client-encryption
-git clone https://github.com/mongodb-js/mongodb-client-encryption.git
+git clone https://github.com/mongodb-js/mongodb-client-encryption.git -b explicit-lifetime-chaining
```
nbbeeken, line 13: Seems unrelated, accidentally added this when testing the fle PR?

src/cmap/wire_protocol/on_demand/document.ts
```diff
@@ -45,6 +46,13 @@ type CachedBSONElement = { element: BSONElement; value: any | undefined };
 
 /** @internal */
 export class OnDemandDocument {
+  /**
+   * @internal
+   *
+   * Used for testing purposes.
+   */
+  private static BSON: typeof BSON = BSON;
```
nbbeeken, line 54: Why do we need to do this but not something similar for `net`?

nbbeeken: Let's determine the reason before introducing this injectable. Are you able to debug? I will take a look when I'm at my desk

nbbeeken (approved):
LGTM

PR #4218, opened 2024-09-03
feat(NODE-6305): Add CSOT support to tailable cursors

W-A-James:
##### `AbstractCursor`
* Added logic to implement CSOT logic for tailable awaitData and non-awaitData cursors

##### `onData`
* Only relevant change here is the addition of the commented code used for debugging purposes. Other diffs in this area are due to a messy rebase, but have been checked and are consistent with the expected state of this code. note that all previous CSOT spec, prose and unit tests continue to pass.

##### `RunCommandCursor`
* Added support for `tailable` and `awaitData` flags

##### `MongoClient`
* added internal `timeoutMS` getter 

##### Benchmarks
* Duplicated all existing benchmarks except the return document benchmark with csot enabled (with an infinite timeout)

##### Testing
* Unskipped (and fixed) tailable cursor Blocking Iteration method test
* Unskipped tailable awaitData and non-awaitData spec tests, duplicated and modified some tests which are only failing due to the the emission of a `killCursors` event when cleaning up the test resources. Original tests that were skipped have accompanying TODO with linked ticket
* Added int tests for tailable awaitData and non-awaitData cursors
* Updated UTR to interpret `cursorType`
* Updated UTR to pass through correct options for `RunCommandCursor`

Not for this PR


test/integration/client-side-operations-timeout/unified-csot-node-specs/tailable-awaitData.json
```diff
@@ -0,0 +1,146 @@
...
+          "object": "tailableCursor",
+          "expectError": {
+            "isTimeoutError": true
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "ignoreExtraEvents": true,
```
W-A-James, line 110: This line is the only change from the original test

test/integration/client-side-operations-timeout/unified-csot-node-specs/tailable-non-awaitData.json
```diff
@@ -0,0 +1,151 @@
...
+          "object": "tailableCursor",
+          "expectError": {
+            "isTimeoutError": true
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "ignoreExtraEvents": true,
```
W-A-James, line 110: This is the only difference from the original test

test/tools/unified-spec-runner/operations.ts
```diff
@@ -258,6 +258,18 @@ operations.set('createCollection', async ({ entities, operation }) => {
 operations.set('createFindCursor', async ({ entities, operation }) => {
   const collection = entities.getEntity('collection', operation.object);
   const { filter, ...opts } = operation.arguments!;
+  switch (opts.cursorType) {
```
W-A-James, line 261: Added to support spec tests in lieu of adding `cursorType` option to all cursors in this PR. Can file follow-up PR for this

test/tools/unified-spec-runner/operations.ts
```diff
@@ -793,9 +817,25 @@ operations.set('runCursorCommand', async ({ entities, operation }: OperationFunc
...
+  }
+  delete opts.cursorType;
   const cursor = collection.runCursorCommand(command, {
     readPreference: ReadPreference.fromOptions({ readPreference: opts.readPreference }),
-    session: opts.session
+    session: opts.session,
+    tailable: opts.tailable,
+    awaitData: opts.awaitData,
+    timeoutMode: opts.timeoutMode,
+    timeoutMS: opts.timeoutMS
```
W-A-James, line 838: passing through relevant options to ensure that runCommandCursors are properly configured

src/cursor/abstract_cursor.ts
```diff
@@ -133,6 +133,7 @@ export type InternalAbstractCursorOptions = Omit<AbstractCursorOptions, 'readPre
   partial?: boolean;
 
   omitMaxTimeMS?: boolean;
+  useMaxAwaitTimeMSAsMaxTimeMS?: boolean;
```
W-A-James, line 136: open to just putting this on the `AbstractCursor` object itself

src/mongo_client.ts
```diff
@@ -477,6 +477,11 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
     return this.s.bsonOptions;
   }
 
+  /** @internal */
+  get timeoutMS(): number | undefined {
+    return this.options.timeoutMS;
+  }
```
W-A-James, line 483: Added to ensure that `resolveOptions` options inheritance is actually able to get option from `MongoClient`

src/operations/create_collection.ts
```diff
@@ -17,6 +17,7 @@ import { Aspect, defineAspects } from './operation';
 const ILLEGAL_COMMAND_FIELDS = new Set([
   'w',
   'wtimeout',
+  'timeoutMS',
```
W-A-James, line 20: Added here as the benchmarks were adding the `timeoutMS` to the actual command being sent to the server when creating a collection

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -291,11 +291,11 @@ describe('CSOT spec prose tests', function () {
        * 1. Verify that a `find` command and two `getMore` commands were executed against the `db.coll` collection during the test.
        */
 
-      it.skip('send correct number of finds and getMores', async function () {
+      it('send correct number of finds and getMores', async function () {
```
W-A-James, line 294: This test seems unreliable

test/integration/client-side-operations-timeout/unified-csot-node-specs/tailable-non-awaitData.json
baileympearson: Can we leave comments explaining why these tests are skipped?

Semi-related - the PR description says:
> Unskipped tailable awaitData and non-awaitData spec tests, duplicated and modified some tests which are only failing due to the the emission of a killCursors event when cleaning up the test resources.

That doesn't seem right.  We match on command events during test iteration, before cleaning up any resources associated with the test.  This implies that the killCursors is coming from inside the test execution - the test seems to imply that we shouldn't be cleaning up a tailable cursor on iteration timeout failure.  I confirmed the test passes if, in `fetchBatch`, we don't clean up the cursor if the error we received is a timeout error.

Can you confirm what the expected behavior is here?  This probably applies for the awaitData tests too.

test/integration/client-side-operations-timeout/node_csot.test.ts
```diff
@@ -576,6 +576,190 @@ describe('CSOT driver tests', metadata, () => {
...
+  describe('Tailable cursors', function () {
+    let client: MongoClient;
+    let internalClient: MongoClient;
+    let commandStarted: CommandStartedEvent[];
+
+    const failpoint: FailPoint = {
+      configureFailPoint: 'failCommand',
+      mode: 'alwaysOn',
+      data: {
+        failCommands: ['aggregate', 'find'],
```
baileympearson, line 588:
```suggestion
        failCommands: ['aggregate', 'find', 'getMore'],
```

src/cursor/abstract_cursor.ts
```diff
@@ -767,15 +785,27 @@ export abstract class AbstractCursor<
...
+      omitMaxTimeMS:
+        this.cursorOptions.omitMaxTimeMS || this.cursorOptions.useMaxAwaitTimeMSAsMaxTimeMS
+    };
+
+    if (
+      this.cursorOptions.tailable &&
+      this.cursorOptions.awaitData &&
+      this.cursorOptions.maxAwaitTimeMS != null
+    ) {
+      getMoreOptions.maxTimeMS = this.cursorOptions.maxAwaitTimeMS;
```
baileympearson, line 801: This does nothing - `GetMoreOperation.execute()` already conditionally attaches a maxTimeMS equal to `maxAwaitTimeMS`

src/cursor/abstract_cursor.ts
```diff
@@ -841,6 +871,8 @@ export abstract class AbstractCursor<
     // otherwise need to call getMore
     const batchSize = this.cursorOptions.batchSize || 1000;
     this.cursorOptions.omitMaxTimeMS = this.cursorOptions.timeoutMS != null;
+    this.cursorOptions.useMaxAwaitTimeMSAsMaxTimeMS =
```
baileympearson, line 874: This will never change after the cursor is initialized.  Right?  Why re-assign it?

test/tools/unified-spec-runner/operations.ts
baileympearson: Do we want to support these options for all our cursors?  Like, what use does a tailable await data list collections cursor have?

test/tools/unified-spec-runner/operations.ts
```diff
@@ -793,9 +817,25 @@ operations.set('runCursorCommand', async ({ entities, operation }: OperationFunc
...
+      opts.tailable = true;
+      opts.awaitData = true;
+      break;
+    case 'tailable':
+      opts.tailable = true;
+      break;
+    default:
+      break;
+  }
+  delete opts.cursorType;
```
baileympearson, line 831:
```suggestion
  const { command, cursorType, ...opts } = operation.arguments!;
  switch (cursorType) {
    case 'tailableAwait':
      opts.tailable = true;
      opts.awaitData = true;
      break;
    case 'tailable':
      opts.tailable = true;
      break;
    default:
      break;
  }
```

If you destructure the cursor type too, there shouldn't be a reason to delete it from `opts`.

baileympearson (changes requested):

src/cursor/abstract_cursor.ts
W-A-James: Good catch here

test/benchmarks/driverBench/common.js
```diff
@@ -27,6 +27,12 @@ function makeClient() {
   this.client = new MongoClient(process.env.MONGODB_URI || 'mongodb://127.0.0.1:27017');
 }
```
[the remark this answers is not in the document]
W-A-James: Fair point. I duplicated them explicitly to see the difference side-by-side on the pr, but it makes sense to just let the perf tooling handle it.

test/tools/unified-spec-runner/operations.ts
W-A-James: Fair question. I don't really see that they would have utility. Do we want to remove these options from the types for the `listCollections`, `listIndexes` and similar functions? If we do, would that constitute a breaking API change?

W-A-James: Looking at the code, it seems like `listIndexes` is the only method that still incorrectly included the `tailable` and `awaitData` flags in the type for its `options` parameter, I can put up a separate PR to remove those from main.

src/cursor/abstract_cursor.ts
```diff
@@ -767,15 +781,26 @@ export abstract class AbstractCursor<
...
+    };
+
+    if (
+      this.cursorOptions.tailable &&
+      this.cursorOptions.awaitData &&
+      this.cursorOptions.maxAwaitTimeMS != null
+    ) {
+      getMoreOptions.maxTimeMS = this.cursorOptions.maxAwaitTimeMS;
+    }
+
```
baileympearson, line 798: thoughts on just calculating omitMaxTimeMS inline and not storing it on the cursor's options?
```typescript
const getMoreOptions = {
	...this.cursorOptions,
	session: this.cursorSession,
	batchSize,
};

if (timeoutMS != null) {
	if (...) options.omitMaxTimeMS = true;
}
```

src/cursor/abstract_cursor.ts
```diff
@@ -767,15 +781,26 @@ export abstract class AbstractCursor<
...
+      omitMaxTimeMS: this.cursorOptions.omitMaxTimeMS
+    };
+
+    if (
+      this.cursorOptions.tailable &&
+      this.cursorOptions.awaitData &&
+      this.cursorOptions.maxAwaitTimeMS != null
+    ) {
+      getMoreOptions.maxTimeMS = this.cursorOptions.maxAwaitTimeMS;
+    }
```
baileympearson, line 797:
```suggestion
```

GetMoreOperation looks for [maxAwaitTimeMS, not maxTimeMS](https://github.com/mongodb/node-mongodb-native/blob/681ddd8d539f77a2db80c7b9af3710c63a1b112c/src/operations/get_more.ts#L87).  This likely is passing because you spread `this.cursorOptions` into getMoreOptions already, so `maxAwaitTimeMS` is provided to the GetMoreOperation.

src/cursor/abstract_cursor.ts
```diff
@@ -204,10 +204,22 @@ export abstract class AbstractCursor<
       ...pluckBSONSerializeOptions(options)
     };
     this.cursorOptions.timeoutMS = options.timeoutMS;
```
baileympearson, line 206: (optional)  We have two patterns for attaching options from `options` into `cursorOptions`:

1. CSOT options are attached after the object is built
2. Readpreference is attached inline, with a conditional

Can we consolidate?

```typescript
    readPreference:
        options.readPreference && options.readPreference instanceof ReadPreference
          ? options.readPreference
          : ReadPreference.primary,
    ...pluckBSONSerializeOptions(options),
	timeoutMS: options.timeoutMS,
	tailable: options.tailable,
	awaitData: options.awaitData
```

(or the other way around would be fine too.)

test/integration/client-side-operations-timeout/node_csot.test.ts
```diff
@@ -26,13 +26,15 @@ import { type FailPoint } from '../../tools/utils';
 const metadata = { requires: { mongodb: '>=4.4' } };
 
 describe('CSOT driver tests', metadata, () => {
+  const minPoolSize = 20;
```
baileympearson, line 29: Can we add a comment explaining why we need minPoolSize?

test/integration/client-side-operations-timeout/node_csot.test.ts
```diff
@@ -576,6 +581,201 @@ describe('CSOT driver tests', metadata, () => {
...
+        it('applies timeoutMS to initial command', metadata, async function () {
+          cursor = client
+            .db('db')
+            .collection('coll')
+            .find({}, { timeoutMS: 50, tailable: true, awaitData: true, batchSize: 1 });
+          const maybeError = await cursor.next().then(
+            () => null,
+            e => e
+          );
+          expect(maybeError).to.be.instanceOf(MongoOperationTimeoutError);
```
baileympearson, line 653: nothing here asserts that the aggregate specifically timed out, only that cursor execution timed out.

Perhaps we could assert on the commandStartedEvents and confirm that there was only one command sent?

test/integration/client-side-operations-timeout/node_csot.test.ts
```diff
@@ -576,6 +581,201 @@ describe('CSOT driver tests', metadata, () => {
...
+        it('applies timeoutMS to initial command', metadata, async function () {
+          cursor = client
+            .db('db')
+            .collection('coll')
+            .find({}, { timeoutMS: 50, tailable: true, batchSize: 1 });
+          const maybeError = await cursor.next().then(
+            () => null,
+            e => e
+          );
+          expect(maybeError).to.be.instanceOf(MongoOperationTimeoutError);
```
baileympearson, line 732: Same comment as the `awaitData` section

baileympearson (changes requested):

src/cmap/wire_protocol/on_data.ts
```diff
@@ -105,9 +112,15 @@ export function onData(
 
   function errorHandler(err: Error) {
     const promise = unconsumedPromises.shift();
+    const timeoutError = TimeoutError.is(err)
```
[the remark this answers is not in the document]
W-A-James: Good catch, I can remove this.

src/cursor/abstract_cursor.ts
W-A-James: Wouldn't we still need to calculate it for the initial command? It works better in my opinion to continue to have it on the internal cursor options since it lets us just spread the options  wherever needed across all of our cursor subclasses. I can change the getMore code here to just use the value obtained by the spread though. Also instead of recalculating it every time we enter `fetchBatch` as we do currently, we could just set it to the value needed after `_initialize` is called in `cursorInit`

baileympearson: Yeah, I'm suggesting we just calculate it for the initial cursor command inline (in `_initialize()`).

Here are my thoughts - if you disagree, that's okay and we can resolve this conversation.  This is non-blocking.

`omitMaxTimeMS` isn't a cursor option - it's an option provided to operation execution that cursors calculate given  cursor options and state.  We have a history of attaching data on large objects because its convenient, but I don't think that's the right approach because it leads to a blurring of the boundaries between the data our various objects contain (for example, I think there's a reason we separate out BSON options with pluckBSONSerializationOptions - we have all our options on a shared object and need to separate them out with functions instead of just separating them out when constructing the options object).

I'm also a proponent of keeping data as closely scoped to where its needed as possible. `omitMaxTimeMS` is only needed in exactly two places: when constructing the initialization options and when constructing the getMore options.  It is a little strange to calculate the value of an option in the constructor that only gets used in `_initialize()`.  

Additionally, `omitMaxTimeMS` isn't used in any subclasses and I don't know of any CSOT requirement that will require this in the future.  Storing it in a protected property leaks implementation details of the abstract cursor.

I see the point about convenience about spreading cursor options, but is it that much simpler than something like this?:

```typescript
_initialize() {
  const options = { 
    ...this.cursorOptions, 
    omitMaxTimeMS: this.timeoutMS != null && ... 
  }
}

getMore() {
  const options = { 
    ...this.cursorOptions, 
    omitMaxTimeMS: this.timeoutMS != null 
  }
}
```

baileympearson (approved):

PR #4219, opened 2024-09-03
feat(NODE-6060): set fire-and-forget protocol when writeConcern is w: 0

aditi-khare-mongoDB:
### Description
When `writeConcern.w = 0` is set on an operation that's run through OP_MSG, the driver will set moreToCome on the request so that the server will not send a response. 

#### What is changing?

- When an operation being run through OP_MSG has `writeConcern.w === 0`, the driver
  - sets the moreToCome bit on the outgoing request.
  - returns from operation execution without waiting for a server response
  - set `CommandSucceededEvent.reply` to `{ ok: 1 }` regardless of server reply
 - TSDoc for WriteConcern now contains extra context.

No, none other than the TSDoc changes.

Support spec-defined [fire-and-forget](https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.md#q--a) protocol.

### Write Operations with `WriteConcern.w` set to `0` trigger unacknowledged write protocol
When `WriteConcern.w` is set to `0` on a write operation, either through operation options or the client settings, writes will be unacknowledged by the server. This means the server will not send a response after the request is sent and therefore, the return value from the operation will not contain any information from the server, including whether the server has processed the request yet. 

In addition, when command monitoring is enabled, the `reply` field of a `CommandSucceededEvent` of an unacknowledged write will always be `{ ok: 1 }`.


src/cmap/connection.ts
```diff
@@ -526,7 +526,11 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
             new CommandSucceededEvent(
               this,
               message,
-              options.noResponse ? undefined : (object ??= document.toObject(bsonOptions)),
+              options.noResponse
+                ? undefined
+                : 'moreToCome' in message && message.moreToCome
+                  ? { ok: 1 }
```
aditi-khare-mongoDB, line 532: The command monitoring spec tests expect `{ok: 1}` in the `event.reply` field for a unacknowledged write, [see here for the test details](https://github.com/mongodb/specifications/blob/39db05c6861f3ef6dba8f2cd00b3c7296f561bcd/source/command-logging-and-monitoring/tests/logging/unacknowledged-write.json#L112-L119).

test/integration/read-write-concern/write_concern.test.ts
```diff
@@ -94,6 +99,7 @@ describe('Write Concern', function () {
 
         afterEach(async function () {
           await db.dropDatabase();
+          await sleep(1000);
           await client.close();
```
aditi-khare-mongoDB, line 103: By no longer waiting for the server's response, if we plan on relying on an unacknowledged write immediately after its run, we need to set a timeout.

aditi-khare-mongoDB: Should this be included in the release notes / API docs?

durran: Putting sleeps in tests with arbitrary values is generally going to be flaky at times so I would avoid this. In this specific case I'm not sure why it's even necessary - why do we need to sleep before closing the client?

test/integration/read-write-concern/write_concern.test.ts
```diff
@@ -130,9 +136,10 @@ describe('Write Concern', function () {
           await col.createIndex({ b: -1 });
           await col.createIndex({ a: 1, b: -1 });
 
+          await sleep(1000);
```
durran, line 139: Why is this sleep necessary? We are just asserting that `listIndexes` does not error. If we wanted to ensure it was returning the correct indexes I would explicitly set the write concern to 'majority' when creating the indexes.

test/integration/read-write-concern/write_concern.test.ts
```diff
@@ -161,11 +168,145 @@ describe('Write Concern', function () {
               { writeConcern: { w: 'majority' } }
             );
 
+            await sleep(1000);
```
durran, line 171: Same here. Why do we need to sleep here?

test/integration/read-write-concern/write_concern.test.ts
```diff
@@ -161,11 +168,145 @@ describe('Write Concern', function () {
...
     });
   });
+
+  describe('fire-and-forget protocol', function () {
+    context('when writeConcern = 0 and OP_MSG is used', function () {
+      const writeOperations: { name: string; command: any; expectedReturnVal: any }[] = [
+        {
+          name: 'insertOne',
+          command: client => client.db('test').collection('test').insertOne({ a: 1 }),
+          expectedReturnVal: { acknowledged: false }
```
durran, line 187: I feel like the `expectedReturnVal` for these tests isn't necessary as it's just showing that we return custom values from these methods and only the assertion on the command suceeded event's reply is really needed here. Maybe we can remove these and the assertions on them?

test/unit/commands.test.ts
```diff
@@ -109,3 +109,14 @@ describe('class OpCompressedRequest', () => {
     }
   });
 });
+
+describe('OpMsgRequest', () => {
+  describe('fire-and-forget', () => {
```
durran, line 114:
```suggestion
  describe('#constructor', () => {
```
Since we are describing behaviour inside the class' constructor.

durran (changes requested):

src/cmap/commands.ts
```diff
@@ -465,7 +465,7 @@ export class OpMsgRequest {
 
     // flags
     this.checksumPresent = false;
-    this.moreToCome = options.moreToCome || false;
+    this.moreToCome = options.moreToCome || command.writeConcern?.w === 0 || false;
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Is there documentation you could point me to for this? I was interpreting[ the fire-and-forget protocol](https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.md#q--a) differently!

src/cmap/connection.ts
```diff
@@ -439,7 +439,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
         zlibCompressionLevel: this.description.zlibCompressionLevel
       });
 
-      if (options.noResponse) {
+      if (options.noResponse || ('moreToCome' in message && message.moreToCome)) {
```
nbbeeken, line 442: https://github.com/mongodb/node-mongodb-native/blob/NODE-6060/fire-and-forget/src/cmap/commands.ts#L315-L316

Can we do the same trick we applied to OpReply to OpQuery? Just hard code moreToCome to false since it never is something that is used for that protocol, then we don't have to check for property presence.

aditi-khare-mongoDB: Sure!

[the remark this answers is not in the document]
aditi-khare-mongoDB: I decided to leave removing `noResponse` out of this PR since it would require editing of the public API for `CommandOperationOptions` and `RunAdminCommandOperation` which may be a breaking change. I suggest if we want this change, we should defer creating a separate ticket and working on it there.

test/integration/read-write-concern/write_concern.test.ts
durran: I'm fine leaving them as-is then. Cannot hurt.

src/cmap/commands.ts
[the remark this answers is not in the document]
durran: Resolving as per just setting based on write concern only for now.

src/cmap/connection.ts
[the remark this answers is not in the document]
durran: Can you then deprecate the `noResponse` option on the options then with a comment and link to a ticket to remove it please?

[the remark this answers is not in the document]
aditi-khare-mongoDB: Yep, will do!

src/cmap/commands.ts
```diff
@@ -465,7 +465,7 @@ export class OpMsgRequest {
 
     // flags
     this.checksumPresent = false;
-    this.moreToCome = options.moreToCome || false;
+    this.moreToCome = command.writeConcern?.w === 0 || false;
```
nbbeeken, line 468: Could we check the options first? 

I am thinking about a runCommand that I want to set fire-and-forget for but is not a command that supports a writeConcern field, that should work right? (then we can use this for the endSessions in client.close())

src/cmap/connection.ts
nbbeeken: could noResponse also use ok:1? 

I know there's a test that expects undefined, but changing it to ok:1 doesn't break the TS API given its flexibility and communicates the same lack of information about the actual result of the command.

src/operations/run_command.ts
```diff
@@ -51,6 +51,7 @@ export class RunAdminCommandOperation<T = Document> extends AbstractOperation<T>
   constructor(
     public command: Document,
     public override options: RunCommandOptions & {
+      /** @deprecated Will be removed in the next major version, and replaced with an option to set writeConcern.w to 0 */
       noResponse?: boolean;
```
nbbeeken, line 55: This isn't publicly viewable, so if it was possible to transition this to using w=0 we could remove this from here. Close this if we don't go with my question about checking options.

src/operations/command.ts
```diff
@@ -55,6 +55,7 @@ export interface CommandOperationOptions
   // Admin command overrides.
   dbName?: string;
   authdb?: string;
+  /** @deprecated Will be removed in the next major version. Set writeConcern.w to 0 instead. */
   noResponse?: boolean;
```
nbbeeken, line 59: More than this being deprecated this option was never meant to be public. I'm glad we're doing the safe thing and marking it before removing it but can we also add (loud) warnings about using this? 

If passed to a normal operation this will leave the socket in a nondeterminate state and it will be put back into the pool with an unread reply. I would argue removing this option is a bug fix because "if set this option causes bugs", but we can at least document it.

nbbeeken (changes requested):

src/cmap/connection.ts
aditi-khare-mongoDB: I'd rather not add new functionality to `noResponse` if we're saying that it buggy and deprecated. What do you think?

src/cmap/commands.ts
aditi-khare-mongoDB: Okay, I've added `moreToCome` as an option back.

aditi-khare-mongoDB: @durran , what are your thoughts on this?

src/operations/command.ts
aditi-khare-mongoDB: Resolved on slack, will be done in a separate ticket NODE-6377.

src/cmap/commands.ts
durran: Fine with me.

durran (approved):

src/cmap/connection.ts
aditi-khare-mongoDB: will do in NODE-6377

src/cmap/commands.ts
nbbeeken: Oh that's not what I meant sorry! Can we check the options for the writeConcern setting and can we check it first before checking the command.

nbbeeken: from slack: checking w = 0 on options is problematic because all options are inherited, fire-and-forget should only be used by writes so checking the command is correct. `moreToCome` will be out option for setting fire-and-forget for special runCommand cases.

nbbeeken (approved):

durran (approved):

PR #4220, opened 2024-09-04
fix(NODE-6355): respect utf8 validation options when iterating cursors (6.8)

baileympearson:
#### What is changing?

Cursors now respect `enableUtf8Validation`.

CI won't run on 6.8 - patch build here: https://spruce.mongodb.com/version/66db46990ef9890007c0f2cb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Performance comparison with the 6.8 commit: https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=66db3c2adbc39a02ac0aef75

### Fixed `enableUtf8Validation` option

Starting in v6.8.0 we inadvertently removed the ability to disable UTF-8 validation when deserializing BSON. Validation is normally a good thing, but it was always meant to be configurable and the recent Node.js runtime issues (v22.7.0) make this option indispensable for avoiding errors from mistakenly generated invalid UTF-8 bytes.

### DIFF
--- a/src/cmap/wire_protocol/on_demand/document.ts
+++ b/src/cmap/wire_protocol/on_demand/document.ts
@@ -1,8 +1,9 @@
+import { type DeserializeOptions } from 'bson';
+
 import {
   Binary,
   type BSONElement,
   BSONError,
-  type BSONSerializeOptions,
   BSONType,
   deserialize,
   getBigInt64LE,
@@ -10,7 +11,6 @@ import {
   getInt32LE,
   ObjectId,
   parseToElementsToArray,
-  pluckBSONSerializeOptions,
   Timestamp,
   toUTF8
 } from '../../../bson';
@@ -43,6 +43,15 @@ export type JSTypeOf = {
 
 /** @internal */
 type CachedBSONElement = { element: BSONElement; value: any | undefined };
+
+/**
+ * @internal
+ *
+ * Options for `OnDemandDocument.toObject()`. Validation is required to ensure
+ * that callers provide utf8 validation options. */
+export type OnDemandDocumentDeserializeOptions = DeserializeOptions & {
+  validation: NonNullable<DeserializeOptions['validation']>;
+};
 
 /** @internal */
 export class OnDemandDocument {
@@ -330,14 +339,12 @@ export class OnDemandDocument {
    * Deserialize this object, DOES NOT cache result so avoid multiple invocations
    * @param options - BSON deserialization options
    */
-  public toObject(options?: BSONSerializeOptions): Record<string, any> {
-    const exactBSONOptions = {
-      ...pluckBSONSerializeOptions(options ?? {}),
-      validation: this.parseBsonSerializationOptions(options),
+  public toObject(options?: OnDemandDocumentDeserializeOptions): Record<string, any> {
+    return deserialize(this.bson, {
+      ...options,
       index: this.offset,
       allowObjectSmallerThanBufferSize: true
-    };
-    return deserialize(this.bson, exactBSONOptions);
+    });
   }
 
   private parseBsonSerializationOptions(options?: { enableUtf8Validation?: boolean }): {
--- a/src/cmap/wire_protocol/on_demand/document.ts
+++ b/src/cmap/wire_protocol/on_demand/document.ts
@@ -49,9 +49,8 @@ type CachedBSONElement = { element: BSONElement; value: any | undefined };
  *
  * Options for `OnDemandDocument.toObject()`. Validation is required to ensure
  * that callers provide utf8 validation options. */
-export type OnDemandDocumentDeserializeOptions = DeserializeOptions & {
-  validation: NonNullable<DeserializeOptions['validation']>;
-};
+export type OnDemandDocumentDeserializeOptions = Omit<DeserializeOptions, 'validation'> &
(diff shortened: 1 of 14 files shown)


src/cmap/wire_protocol/on_demand/document.ts
```diff
@@ -329,14 +339,24 @@ export class OnDemandDocument {
...
-    return BSON.deserialize(this.bson, {
+  public toObject(options?: OnDemandDocumentDeserializeOptions): Record<string, any> {
+    return deserialize(this.bson, {
       ...options,
       index: this.offset,
       allowObjectSmallerThanBufferSize: true
     });
   }
 
+  private parseBsonSerializationOptions(options?: { enableUtf8Validation?: boolean }): {
```
dariakp, line 350: is this still necessary? it looks like we are just calling the bson parseUtf8ValidationOptions now

test/integration/node-specific/bson-options/utf8_validation.test.ts
```diff
@@ -1,125 +1,52 @@
...
-          passOptionTo === 'operation' ? option : {}
-        );
-
-        expect(bsonSpy).to.have.been.called;
-        const result = bsonSpy.lastCall.returnValue;
-        expect(result).to.deep.equal(EXPECTED_VALIDATION_ENABLED_ARGUMENT);
-      });
-    }
-  });
+          error = await test.insertOne({ _id }).catch(error => error);
```
dariakp, line 41: It looks like we only test the default behavior here, could we add a case with utf8 validation explicitly enabled also?

test/integration/node-specific/bson-options/utf8_validation.test.ts
```diff
@@ -153,3 +65,182 @@ describe('class MongoDBResponse', () => {
     }
   );
 });
+
+describe('utf8 validation with cursors', function () {
```
dariakp, line 69: Just want to make the fact that we are explicitly testing invalid documents in all the below cases more visible
```suggestion
describe('parsing of utf8-invalid documents with cursors', function () {
```

test/integration/node-specific/bson-options/utf8_validation.test.ts
```diff
@@ -153,3 +65,182 @@ describe('class MongoDBResponse', () => {
...
+    const stub = sinon.stub(net.Socket.prototype, 'write').callsFake(function (...args) {
+      const providedBuffer = args[0].toString('hex');
+      const targetBytes = Buffer.from(document.field, 'utf-8').toString('hex');
+
+      if (providedBuffer.includes(targetBytes)) {
+        if (providedBuffer.split(targetBytes).length !== 2) {
+          sinon.restore();
+          const message = `too many target bytes sequences: received ${
+            providedBuffer.split(targetBytes).length
+          }\n.  command: ${inspect(deserialize(args[0]), { depth: Infinity })}`;
```
nbbeeken, line 88: Isn't args[0] an op_msg payload not BSON? so this shouldn't be parsable I think

test/integration/node-specific/bson-options/utf8_validation.test.ts
```diff
@@ -153,3 +65,182 @@ describe('class MongoDBResponse', () => {
...
+      await fn();
+      expect.fail('expected the provided callback function to reject, but it did not.');
+    } catch (error) {
+      expect(error).to.match(/Invalid UTF-8 string in BSON document/);
+      expect(error).to.be.instanceOf(BSONError);
+    }
+  }
+
+  context('when utf-8 validation is explicitly enabled', function () {
+    it('a for-await loop throw a BSON error', async function () {
```
dariakp, line 166:
```suggestion
    it('a for-await loop throws a BSON error', async function () {
```

test/integration/node-specific/bson-options/utf8_validation.test.ts
```diff
@@ -153,3 +65,182 @@ describe('class MongoDBResponse', () => {
...
+
+        while (await cursor.hasNext()) {
+          await cursor.tryNext();
+        }
+      });
+    });
+  });
+
+  context('utf-8 validation defaults to enabled', function () {
+    it('a for-await loop throw a BSON error', async function () {
```
dariakp, line 207:
```suggestion
    it('a for-await loop throws a BSON error', async function () {
```

src/cmap/wire_protocol/on_demand/document.ts
nbbeeken: This is "unused" according to vscode graying it out, should be good to remove

src/cmap/wire_protocol/responses.ts
```diff
@@ -267,12 +265,13 @@ export class CursorResponse extends MongoDBResponse {
       this.cursor.get('postBatchResumeToken', BSONType.object)?.toObject({
         promoteValues: false,
         promoteLongs: false,
-        promoteBuffers: false
+        promoteBuffers: false,
+        validation: { utf8: true }
```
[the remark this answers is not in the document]
nbbeeken: validation is now required, we have added it to a few places as a result even though it is repeating a default is better that we are forced to set this so we don't miss it.

nbbeeken (approved):

dariakp (approved):

PR #4231, opened 2024-09-11
fix(NODE-6367):  enable mixed use of iteration APIs

W-A-James:

No

Fixing mixed use of `cursor.next` and `for await (const doc of cursor)`.

### Restore mixed use of `cursor.next()` and `cursor[Symbol.asyncIterator]`

In 6.8.0, we inadvertently prevented the use of `cursor.next()` along with using `for await` syntax to iterate cursors. This use-case has now been restored.

### DIFF
--- a/test/integration/crud/find_cursor_methods.test.js
+++ b/test/integration/crud/find_cursor_methods.test.js
@@ -1,6 +1,11 @@
 'use strict';
 const { expect } = require('chai');
 const { filterForCommands } = require('../shared');
+const {
+  promiseWithResolvers,
+  MongoExpiredSessionError,
+  MongoCursorExhaustedError
+} = require('../../mongodb');
 
 describe('Find Cursor', function () {
   let client;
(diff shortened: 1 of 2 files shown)


test/integration/crud/find_cursor_methods.test.js
```diff
@@ -361,4 +361,93 @@ describe('Find Cursor', function () {
...
+        while ((doc = (await cursor.next()) && doc != null)) {
+          /** empty */
+        }
+
+        expect(count).to.equal(2);
+      });
+    });
+
+    context('when there are documents are not retrieved in the first batch', function () {
+      it('allows combining iteration modes', async function () {
```
nbbeeken, line 419: From slack, a few more cases:
- Async iterator first, cursor.next() after.
- Is it our expectation that cursor.next/asyncIterator/tryNext are all UB if someone has made a cursor.stream()? (This was my understanding)
- We have some `readBufferedDocuments()` tests in this file, but not interleaved with next calls, nor in relation to cursor state, what does this "iteration" method do when the cursor is closed? 
- Should we double check tryNext here as well? it doesn't share code paths with next()
- I think after toArray is called the cursor is clearly "done" what do next/asyncIterator/tryNext/readBufferedDocuments/.stream() do after that?

nbbeeken (changes requested):

nbbeeken (approved):

PR #4240, opened 2024-09-16
fix(NODE-6377): remove noResponse option

aditi-khare-mongoDB:
### Description
Now that unacknowledged writes are supported through `writeConcern.w = 0`, remove the buggy `noResponse option` internally and in public APIs.

#### What is changing?
- Remove noResponse from public `CommandOperationOptions` API
- Update `client.close()` and `RunAdminCommandOperation` to use `writeConcern` instead

None, except removing the option from the Public API.

NODE-6060.

### Remove `CommandOperationOptions.noResponse` option

Setting the `noResponse` option on `CommandOperationOptions` would result in indeterminate socket behavior leading to either silent or thrown errors on any following operations. Since this option is broken and causes bugs, we're removing it to call attention to any code that may have been using it inadvertently.

In order to correctly request that the driver recieves 'no response' from the server (usually to perform an unacknowledged write) setting the `WriteConcern.w` property to `0` is the recommended migration.


test/integration/connection-monitoring-and-pooling/connection.test.ts
```diff
@@ -144,6 +144,21 @@ describe('Connection', function () {
         }
       }
     });
+
```
aditi-khare-mongoDB, line 147: Moved this out of unit tests so it could be configured properly / not default to `OpQuery` incorrectly

test/integration/collection-management/collection.test.ts
```diff
@@ -422,6 +429,41 @@ describe('Collection', function () {
     });
   });
 
+  describe('#createIndex', () => {
```
aditi-khare-mongoDB, line 432: Moved this out of unit tests so it could be configured properly / not default to OpQuery incorrectly. We can also make `mockServer` default to a specific `maxWireVersion >= 7` to stop this error, but I figured since it's only 2 tests for now, its not worth the effort.

nbbeeken (changes requested):
Just some quick comment fixes

src/mongo_client.ts
```diff
@@ -628,7 +628,7 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
             this,
             new RunAdminCommandOperation(
               { endSessions },
-              { readPreference: ReadPreference.primaryPreferred, noResponse: true }
+              { readPreference: ReadPreference.primaryPreferred, moreToCome: true }
```
durran, line 631: Note that the description on the PR says we're using `writeConcern` instead of `noResponse`, so I had the expectation that we'd be setting `writeConcern` to `0` here and not setting `moreToCome`. If this is the true approach can we either update the PR description or change this to use `writeConcern`?

durran (changes requested):

test/integration/node-specific/mongo_client.test.ts
```diff
@@ -698,7 +698,7 @@ describe('class MongoClient', function () {
       expect(startedEvents).to.have.lengthOf(1);
       expect(startedEvents[0]).to.have.property('commandName', 'endSessions');
       expect(endEvents).to.have.lengthOf(1);
-      expect(endEvents[0]).to.have.property('reply', undefined); // noReponse: true
+      expect(endEvents[0]).to.containSubset({ reply: { ok: 1 } }); // writeConcern.w = 0
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Yep, good catch

src/cmap/connection.ts
```diff
@@ -94,6 +93,9 @@ export interface CommandOptions extends BSONSerializeOptions {
   writeConcern?: WriteConcern;
 
   directConnection?: boolean;
+
+  // Triggers fire-and-forget protocol for commands that don't support WriteConcern
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good, made the change!

src/mongo_client.ts
aditi-khare-mongoDB: `RunAdminCommandOperation` is an internal command, and it uses the `moreToCome` flag to run on commands which don't support `writeConcern`. 

The only user-facing API change is to `CommandOperationOptions`  which does already support `writeConcern` as an option, so users should be able to make the switch from using `noResponse` to `writeConcern` pretty easily.

durran (approved):

nbbeeken (changes requested):
CI needs a closer look 👁️

nbbeeken (changes requested):
Still seeing tasks timeout ⏲️

PR #4242, opened 2024-09-19
feat(NODE-6337): implement client bulk write batching

durran:


src/cmap/commands.ts
```diff
@@ -429,10 +429,68 @@ export interface OpMsgOptions {
...
+    this.header?.writeInt32LE(4 + this.field.length + 1 + this.serializedDocumentsLength, 1);
+    return this.serializedDocumentsLength;
+  }
+
+  /**
+   * Get the fully serialized bytes for the document sequence section.
+   * @returns The section bytes.
+   */
+  toBin(): Uint8Array {
+    // TODO: What to do if no documents?
```
baileympearson, line 492: Nothing, right?  We should probably check but I'd assume that the server would deserialize an OP_MSG with a document sequence that has no documents as an empty array.

durran: My todo here is that if there are no documents currently the 4 byte length in the header isn't set yet.

durran:
Did a clean rebase here: https://github.com/mongodb/node-mongodb-native/pull/4248

PR #4243, opened 2024-09-19
feat(NODE-6389): Add support for timeoutMS in StateMachine.execute()

aditi-khare-mongoDB:


test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -57,17 +60,55 @@ describe('CSOT spec prose tests', function () {
...
+        sinon.restore();
+      });
+
+      it('maxTimeMS is not set', async function () {
+        const commandStarted = [];
+        client.on('commandStarted', ev => commandStarted.push(ev));
+        await client
+          .db('admin')
+          .command({ ping: 1 })
+          .catch(e => squashError(e));
```
aditi-khare-mongoDB, line 104: The squashed error on my machine is `MongoServerError: no such command { ping : 1 }`. Before the command is sent to the server we can use command monitoring to ensure there is no `maxTimeMS` field, but I'm aware this is avoiding the problem of why this error is there in the first place.

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
```diff
@@ -93,17 +104,35 @@ describe('CSOT spec unit tests', function () {
   }).skipReason =
     'TODO(NODE-5682): Add CSOT support for socket read/write at the connection layer for CRUD APIs';
 
-  context.skip('Client side encryption', function () {
-    context(
-      'The remaining timeoutMS value should apply to HTTP requests against KMS servers for CSFLE.',
-      () => {}
-    );
+  context('Client side encryption', function () {
+    it('The remaining timeoutMS value should apply to HTTP requests against KMS servers for CSFLE.', async function () {
```
baileympearson, line 108: a couple testing notes

---

First, tests should describe the behavior under test.  `The remaining timeoutMS value should apply to HTTP requests against KMS servers for CSFLE.'` isn't accurate because you're not testing the "remaining" portion anywhere - this tests that timeoutMS is respected as a timeout.  I find that writing out the test case as in the format "given <conditions>, it <behavior>" helps make it clear whats under test.  For example:

```typescript
  context('when StateMachine.kmsRequest() is passed a `timeoutMS`', function () {
	it('timeoutMS is used as the timeout for the KMS request")
  }); 
```

This also helps uncover missing test cases, in this case:

```typescript
  context('when StateMachine.kmsRequest() is not passed a `timeoutMS`', function () {
	it('the kms request never times out")
  }); 
```

---

Second, we need testing for the other async methods on StateMachine.  Unless we have coverage of your changes elsewhere?  

---

I think we should consider our testing approach.  If we add unit tests like this to each other async method on the state machine, we don't actually test that the state machine correctly times out each implementation (there's nothing the ensures we use the remaining timeoutMS from the timeoutContext as the timeoutMS value for each method.  

Finally, we also need to consider what happens when timeoutMS is not enabled, and ensure that the existing behavior is unchanged.

I'm happy to meet later today / tomorrow if you want to discuss testing in person! Just let me know.

src/client-side-encryption/state_machine.ts
```diff
@@ -255,9 +282,12 @@ export class StateMachine {
         }
```
baileympearson, line 282: I can't comment on it because it isn't in the diff.  Can we add a TODO to line 179 linking this ticket? [DRIVERS-2671](https://jira.mongodb.org/browse/DRIVERS-2671)

We need to make automatic kms credential refresh respect CSOT too, but that drivers work has not been completed.

baileympearson (changes requested):

src/client-side-encryption/state_machine.ts
```diff
@@ -522,12 +558,20 @@ export class StateMachine {
...
+  ): Promise<Uint8Array> {
     const { db } = MongoDBCollectionNamespace.fromString(ns);
-    const rawCommand = deserialize(command, options);
+    const bsonOptions = { promoteLongs: false, promoteValues: false };
+    const rawCommand = deserialize(command, bsonOptions);
 
-    const response = await client.db(db).command(rawCommand, options);
+    const response = await client.db(db).command(rawCommand, {
+      ...bsonOptions,
+      ...(typeof timeoutMS === 'number' ? { timeoutMS, omitMaxTimeMS: true } : undefined)
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: `timeoutMS` is often typed as `timeoutMS?: number` type, so it can either be undefined or a number. Here and in other places, we're passing in a `number | null value`, which cannot be passed in as is.

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
aditi-khare-mongoDB: Bookkeeping: added testing for `remainingTimeM`S for all async `StateMachine.execute helpers` and added an explicit `sleep` in between execution in tests

src/client-side-encryption/state_machine.ts
```diff
@@ -182,7 +184,11 @@ export class StateMachine {
   /**
    * Executes the state machine according to the specification
    */
-  async execute(executor: StateMachineExecutable, context: MongoCryptContext): Promise<Uint8Array> {
+  async execute(
+    executor: StateMachineExecutable,
+    context: MongoCryptContext,
+    timeoutContext?: CSOTTimeoutContext
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good! Made the change!

src/client-side-encryption/state_machine.ts
aditi-khare-mongoDB: Done!

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
```diff
@@ -93,17 +105,73 @@ describe('CSOT spec unit tests', function () {
...
+          });
+        });
+
+        afterEach(async function () {
+          sinon.restore();
+        });
+
+        it('the kms request does not timeout within 30 seconds', async function () {
+          const sleepingFn = async () => {
+            await sleep(55000);
```
baileympearson, line 161:
```suggestion
            await sleep(30000);
```

Also - Node has a setTimeout that returns a promise we can use now, if you import it from `timers/promises`

src/client-side-encryption/state_machine.ts
```diff
@@ -543,15 +586,21 @@ export class StateMachine {
...
 
+    if (timeoutContext?.csotEnabled() && timeoutContext?.remainingTimeMS <= 0) {
+      throw new MongoOperationTimeoutError('Timed out before dataKey fetched.');
+    }
     return client
       .db(dbName)
       .collection<DataKey>(collectionName, { readConcern: { level: 'majority' } })
-      .find(deserialize(filter))
+      .find(deserialize(filter), {
+        timeoutMS: timeoutContext?.csotEnabled() ? timeoutContext?.remainingTimeMS : undefined
```
baileympearson, line 602: The default timeout mode is LIFETIME - but just to be on the safe side, can we pass in `timeoutMode: CursorTimeoutMode.LIFETIME` too? That prevents the timeout behavior in these functions from silently changing if the default timeout mode ever changes.

aditi-khare-mongoDB: Sounds good!

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
```diff
@@ -93,17 +105,73 @@ describe('CSOT spec unit tests', function () {
...
+          sinon.stub(TLSSocket.prototype, 'connect').callsFake(async function (..._args) {
+            return {} as TLSSocket;
+          });
+        });
+
+        afterEach(async function () {
+          sinon.restore();
+        });
+
+        it('the kms request does not timeout within 30 seconds', async function () {
```
baileympearson, line 159: For this test - can we use fake timers so that the test suite doesn't have a 30 second delay?  

You can find examples using fake timers by looking for `createTimerSandbox()` and `useFakeTimers`, but basically it lets us mock timers so that we can advance time synchronously / instantaneously. 

We'll also need to update `sleep` to look something like `export const sleep = (ms: number): Promise<void> => promisify(setTimeout)(ms);`

src/client-side-encryption/state_machine.ts
```diff
@@ -543,15 +576,19 @@ export class StateMachine {
...
+    timeoutContext?: TimeoutContext
   ): Promise<Array<DataKey>> {
     const { db: dbName, collection: collectionName } =
       MongoDBCollectionNamespace.fromString(keyVaultNamespace);
 
     return client
       .db(dbName)
       .collection<DataKey>(collectionName, { readConcern: { level: 'majority' } })
-      .find(deserialize(filter))
+      .find(deserialize(filter), {
```
baileympearson, line 588: Lots of red in CI caused by this - `timeoutMode` can only be specified if timeoutMS is specified too.  Otherwise cursors throw an error.

Also - lint is failing

test/integration/client-side-operations-timeout/client_side_operations_timeout.unit.test.ts
aditi-khare-mongoDB: Done!

src/client-side-encryption/state_machine.ts
aditi-khare-mongoDB: Yep, changed it!

baileympearson (approved):

PR #4246, opened 2024-09-24
feat(NODE-6275): Add CSOT support to GridFS

W-A-James:
##### `GridFSUploadStream`
* Add CSOT support by piping `TimeoutContext` down through all operations over lifetime of `GridFSUploadStream`
* Migrate `checkIndexes` call from constructor to async `construct()` method to accomodate CSOT error handling

##### `GridFSDownloadStream`
* Add CSOT support by piping `TimeoutContext` down through all operations over lifetime of `GridFSDownloadStream`

##### `GridFSBucket`
* Add support for `timeoutMS` option to:
	* `GridFSBucket.openUploadStream()`
	* `GridFSBucket.openUploadStreamWithId()`
	* `GridFSBucket.openDownloadStream()`
	* `GridFSBucket.openDownloadStreamByName()`
	* `GridFSBucket.find()`
	* `GridFSBucket.drop()`
	* `GridFSBucket.rename()`
	* `GridFSBucket.delete()`
	
##### Testing
* Unskip CSOT gridfs spec tests
* Implement Prose test [6](https://github.com/mongodb/specifications/tree/master/source/client-side-operations-timeout/tests#6-gridfs---upload) and [7](https://github.com/mongodb/specifications/tree/master/source/client-side-operations-timeout/tests#7-gridfs---download) 
* Add int tests that verify that non-spec-required methods (`GridFSBucket.openUploadStreamWithId()` and  `GridFSBucket.openDownloadStreamByName()`) properly handle `timeoutMS`
* Add int test verifying that `'index'` event is emitted exactly once by `GridFSBucket` when creating upload streams

Yes, punt to NODE-5688, also see NODE-6275 for specifics.

NODE-6275


W-A-James:
Note, tests currently failing since mongodb-legacy does not currently pass through the options for some methods that now have an options argument

src/gridfs/download.ts
```diff
@@ -196,7 +204,8 @@ export class GridFSBucketReadStream extends Readable {
   async abort(): Promise<void> {
     this.push(null);
     this.destroy();
-    await this.s.cursor?.close();
+    const remainingTimeMS = getRemainingTimeMSOrThrow(this.s.timeoutContext);
+    await this.s.cursor?.close({ timeoutMS: remainingTimeMS });
```
W-A-James, line 208: The spec says that abort should fall under the overall lifetime of the stream, but wouldn't this leak server resources if it were to time out before the command is written?

src/timeout.ts
```diff
@@ -352,3 +352,20 @@ export class LegacyTimeoutContext extends TimeoutContext {
     return;
   }
 }
+
+/** @internal
+ * When `timeoutContext` is defined, throws a MongoOperationTimeoutError if the context has expired.
+ * If the context has not expired, returns the `remainingTimeMS`
+ * If the `timeoutContext` is not defined, returns undefined
+ * */
+export function getRemainingTimeMSOrThrow(
```
W-A-James, line 361: Added this helper to prevent excessive duplication of code

test/integration/client-side-operations-timeout/client_side_operations_timeout.spec.test.ts
```diff
@@ -9,11 +9,6 @@ const skippedSpecs = {
   'change-streams': 'TODO(NODE-6035)',
   'convenient-transactions': 'TODO(NODE-5687)',
   'deprecated-options': 'TODO(NODE-5689)',
-  'gridfs-advanced': 'TODO(NODE-6275)',
-  'gridfs-delete': 'TODO(NODE-6275)',
-  'gridfs-download': 'TODO(NODE-6275)',
-  'gridfs-find': 'TODO(NODE-6275)',
-  'gridfs-upload': 'TODO(NODE-6275)',
```
W-A-James, line 16: Unskipping tests

src/gridfs/upload.ts
```diff
@@ -147,8 +153,23 @@ export class GridFSBucketWriteStream extends Writable {
...
+        () => {
+          this.bucket.s.checkedIndexes = true;
+          this.bucket.emit('index');
+        },
+        error => {
+          if (error instanceof MongoOperationTimeoutError) {
+            handleError(this, error, callback);
+          } else {
+            squashError(error);
+            process.nextTick(callback);
```
nbbeeken, line 169: We're already inside of a promise rejection here, the nextTick made (alleged) sense when it was calling the callback sync but we're async here so we can just call the callback

src/gridfs/upload.ts
```diff
@@ -147,8 +153,23 @@ export class GridFSBucketWriteStream extends Writable {
...
+          }
+        }
+      );
     }
     this.bucket.once('index', callback);
```
nbbeeken, line 174: Can we call the callback in the promise resolution now? I think this indirection made sense when we needed to link the promise started in the CTOR to this, but now it is all here, we can avoid setting a listener.

src/timeout.ts
nbbeeken: Don't feel strongly but I would've expected this to be a method on the class, and the null case can be handled by `?.`

src/gridfs/upload.ts
```diff
@@ -270,21 +311,48 @@ function checkDone(stream: GridFSBucketWriteStream, callback: Callback): void {
...
-    stream.files.insertOne(gridFSFile, { writeConcern: stream.writeConcern }).then(
-      () => {
-        stream.gridFSFile = gridFSFile;
-        callback();
-      },
-      error => handleError(stream, error, callback)
-    );
+    const remainingTimeMS = stream.timeoutContext?.remainingTimeMS;
+    if (remainingTimeMS != null && remainingTimeMS <= 0) {
+      handleError(
```
nbbeeken, line 316: `return handleError` (just consistent, and easy to see early return)

src/gridfs/index.ts
```diff
@@ -109,7 +111,10 @@ export class GridFSBucket extends TypedEventEmitter<GridFSBucketEvents> {
     filename: string,
     options?: GridFSBucketWriteStreamOptions
   ): GridFSBucketWriteStream {
-    return new GridFSBucketWriteStream(this, filename, options);
+    return new GridFSBucketWriteStream(this, filename, {
+      timeoutMS: this.s.db.timeoutMS,
```
nbbeeken, line 115: We stored timeoutMS to this bucket but then reach to the db for the value, is that intentional?

src/gridfs/index.ts
```diff
@@ -198,18 +223,36 @@ export class GridFSBucket extends TypedEventEmitter<GridFSBucketEvents> {
...
+        timeoutMS,
+        serverSelectionTimeoutMS: this.s.db.client.options.serverSelectionTimeoutMS
+      });
+    }
+
+    if (timeoutContext) {
+      await this.s._filesCollection.drop({ timeoutMS: timeoutContext.remainingTimeMS });
+      const remainingTimeMS = timeoutContext.remainingTimeMS;
+      if (remainingTimeMS <= 0)
+        throw new MongoOperationTimeoutError(`Timed out after ${timeoutMS}ms`);
```
nbbeeken, line 251: Get remaining or throw helper?

src/gridfs/download.ts
nbbeeken: Is this question "new"? in the sense that it is the same one we asked about timing out iterations on the cursor right?

nbbeeken (changes requested):

src/gridfs/index.ts
W-A-James: This was an oversight. Good catch.

src/gridfs/upload.ts
```diff
@@ -97,6 +104,8 @@ export class GridFSBucketWriteStream extends Writable {
    * ```
    */
   gridFSFile: GridFSFile | null = null;
+  /** @internal */
+  timeoutContext?: CSOTTimeoutContext;
```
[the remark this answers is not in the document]
W-A-James: Since the majority of the implementation of the class is done in floating functions that take in an instance of a `GridFSUploadStream`, this either has to remain public or we can refactor to turn those floating functions into private methods. I'm open to either.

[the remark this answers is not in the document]
nbbeeken: Oh right, no worries leaving this 'as is' is fine

test/tools/unified-spec-runner/operations.ts
```diff
@@ -832,9 +833,16 @@ operations.set('updateOne', async ({ entities, operation }) => {
...
-  const collection = entities.getEntity('collection', operation.object);
-  const { to, ...options } = operation.arguments!;
-  return collection.rename(to, options);
+  try {
+    const collection = entities.getEntity('collection', operation.object);
+    const { to, ...options } = operation.arguments!;
+    return collection.rename(to, options);
+  } catch {
+    // Ignore
+  }
```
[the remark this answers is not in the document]
W-A-James: As far as where I referenced this change, I was looking at the spec tests and we just didn't have handling for GridFSBucket.rename. If it's not in the UTR spec, then we should probably file a DRIVERS ticket.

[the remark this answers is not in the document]
W-A-James: Yeah, looking at the UTR spec, it seems this operation doesn't exist 😕 
I'll still go ahead and update this implementation to what you recommended so we don't drop any errors unexpectedly

W-A-James:
Re the question about leaking cursors on the server, you're right, it's not a new question for CSOT, but thought it might still be good to flag it here. Not sure if there's anything we want to or can do about it rn.

test/tools/unified-spec-runner/operations.ts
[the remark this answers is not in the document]
nbbeeken: "rename" is not in the UTR spec but it is in spec test files and we have an implementation in our runner? Fascinating :) Yea drivers ticket sounds good but seems like everyone (generally) would have to have this for some other use of rename that already exists. 🤔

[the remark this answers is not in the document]
nbbeeken: The `getEntity` doesn't take an options object, it just takes a boolean, there are TS errors "in editor" but there's no type check in CI

src/timeout.ts
nbbeeken: Just want your thoughts here in case this was missed

src/gridfs/download.ts
nbbeeken: Okay yes worth calling out but something we should defer and track the answer to. I think the prevailing feeling is that we rely on auto clean up server-side but we should document the answer when it is clear to us and users what the expectation is.

nbbeeken (changes requested):

nbbeeken (approved):

src/gridfs/upload.ts
```diff
@@ -131,14 +140,11 @@ export class GridFSBucketWriteStream extends Writable {
...
-    if (!this.bucket.s.calledOpenUploadStream) {
-      this.bucket.s.calledOpenUploadStream = true;
-
-      checkIndexes(this).then(() => {
-        this.bucket.s.checkedIndexes = true;
-        this.bucket.emit('index');
-      }, squashError);
-    }
+    if (options.timeoutMS != null)
+      this.timeoutContext = new CSOTTimeoutContext({
```
baileympearson, line 144: Any reason we're not using `TimeoutContext.create()`?

src/gridfs/upload.ts
```diff
@@ -147,10 +153,26 @@ export class GridFSBucketWriteStream extends Writable {
...
+        error => {
+          if (error instanceof MongoOperationTimeoutError) {
+            return handleError(this, error, callback);
+          }
+          squashError(error);
+          callback();
+        }
+      );
+    } else {
+      callback();
```
baileympearson, line 174:
```suggestion
      return process.nextTick(callback);
```

In general, it's important to ensure that APIs are either 1. always async or 2. never async.  We aim to avoid "sometimes async" APIs.

src/gridfs/upload.ts
```diff
@@ -194,7 +216,10 @@ export class GridFSBucketWriteStream extends Writable {
     }
 
     this.state.aborted = true;
-    await this.chunks.deleteMany({ files_id: this.id });
+    const remainingTimeMS = this.timeoutContext?.getRemainingTimeMSOrThrow(
+      `Upload timed out after ${this.timeoutContext?.timeoutMS}ms`
+    );
+    await this.chunks.deleteMany({ files_id: this.id, timeoutMS: remainingTimeMS });
   }
```
baileympearson, line 223: Is this really the correct behavior?  This would leave the user's data in a partially-written state.  That doesn't seem right

src/gridfs/index.ts
```diff
@@ -144,11 +154,27 @@ export class GridFSBucket extends TypedEventEmitter<GridFSBucketEvents> {
...
 
+    const remainingTimeMS = timeoutContext?.remainingTimeMS;
+    if (remainingTimeMS != null && remainingTimeMS <= 0)
+      throw new MongoOperationTimeoutError(`Timed out after ${timeoutMS}ms`);
     // Delete orphaned chunks before returning FileNotFound
-    await this.s._chunksCollection.deleteMany({ files_id: id });
+    await this.s._chunksCollection.deleteMany({ files_id: id }, { timeoutMS: remainingTimeMS });
 
     if (deletedCount === 0) {
       // TODO(NODE-3483): Replace with more appropriate error
```
baileympearson, line 180: How feasible is it to make our public APIs take a timeoutContext? (internal option)
```suggestion
	const timeoutContext = TimeoutContext.create(resolveOptions(this.s.db, options));
    
    const { deletedCount } = await this.s._filesCollection.deleteOne(
      { _id: id },
      { timeoutContext }
	);

    // Delete orphaned chunks before returning FileNotFound
    await this.s._chunksCollection.deleteMany({ files_id: id }, { timeoutContext });
```

If all operations are sharing a timeout context, we don't need to manually check if its expired because the first server selection after it expires (in the subsequent operation) will throw a timeout error.

Also - can we use `resolveOptions`?

src/gridfs/index.ts
```diff
@@ -198,18 +224,36 @@ export class GridFSBucket extends TypedEventEmitter<GridFSBucketEvents> {
...
       throw new MongoRuntimeError(`File with id ${id} not found`);
     }
   }
 
   /** Removes this bucket's files collection, followed by its chunks collection. */
-  async drop(): Promise<void> {
-    await this.s._filesCollection.drop();
-    await this.s._chunksCollection.drop();
+  async drop(options?: { timeoutMS: number }): Promise<void> {
+    const timeoutMS = options?.timeoutMS ?? this.s.db.timeoutMS;
```
baileympearson, line 238: Same comment here, we can unify the code path for legacy and CSOT if we accept timeoutContexts in our public API.

baileympearson: (that would also remove a lot of manual checks to see if the timeout has expired in upload.ts).

baileympearson (changes requested):

src/gridfs/download.ts
```diff
@@ -352,7 +361,17 @@ function init(stream: GridFSBucketReadStream): void {
         filter['n'] = { $gte: skip };
       }
     }
-    stream.s.cursor = stream.s.chunks.find(filter).sort({ n: 1 });
+
+    const remainingTimeMS = stream.s.timeoutContext?.getRemainingTimeMSOrThrow(
```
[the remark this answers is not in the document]
W-A-James: Good catch here

src/gridfs/upload.ts
W-A-James: The only time we'll have a timeout context on a GridFSStream is when CSOT is enabled, so we don't need the broader abstraction here.

src/gridfs/upload.ts
W-A-James: This is the behaviour expected by the spec([1](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#gridfs-api), [2](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#gridfs-streams-behavior), [3](https://github.com/mongodb/specifications/tree/master/source/client-side-operations-timeout/tests#aborting-an-upload-stream-can-be-timed-out)), but I do see your point here as it could leave user's collections in an unexpected state. I can see an argument for treating `abort` the same way we treat `close` for cursors in that it seems more user friendly to refresh the timeout before calling `deleteMany`

src/gridfs/index.ts
W-A-James: It should be possible to do this by updating the `OperationOptions`, adding a `timeoutContext` property to `AbstractOperation` and updating `executeOperation` to have some preference for which `timeoutContext` it gives precedence to (the one passed into `executeOperation` directly or the one on the operation), but I think that can be a future improvement because we'd need additional tests to ensure that this doesn't lead to any unexpected behaviour changes that our existing tests might not account for.

baileympearson: Okay.

Can we still use resolve options though?

```typescript
const timeoutMS = resolveOptions(this.db, options);
const timeoutContext = TimeoutContext.create(options);
```

src/gridfs/upload.ts
baileympearson: Filed DRIVERS-2996

src/gridfs/index.ts
```diff
@@ -188,7 +215,7 @@ export class GridFSBucket extends TypedEventEmitter<GridFSBucketEvents> {
       this.s._filesCollection,
       this.s.options.readPreference,
       { filename },
-      { ...options, sort, skip }
+      { timeoutMS: this.s.db.timeoutMS, ...options, sort, skip }
```
baileympearson, line 218:
```suggestion
      { timeoutMS: this.s.options.timeoutMS, ...options, sort, skip }
```

src/gridfs/index.ts
```diff
@@ -144,11 +155,27 @@ export class GridFSBucket extends TypedEventEmitter<GridFSBucketEvents> {
    *
    * @param id - The id of the file doc
    */
-  async delete(id: ObjectId): Promise<void> {
-    const { deletedCount } = await this.s._filesCollection.deleteOne({ _id: id });
+  async delete(id: ObjectId, options?: { timeoutMS: number }): Promise<void> {
+    const { timeoutMS } = resolveOptions(this.s.db, options);
```
baileympearson, line 159: I looked more closely at this - the constructor already resolves timeoutMS based on the `db` object and the bucket options.  These helpers should be resolving it based on the option set on the bucket and the options provided to each method - right?

```typescript
// constructor
const resolvedOptions = resolveOptions(this.db, { 
	...DEFAULT_GRIDFS_BUCKET_OPTIONS,
	...options,
})

// (write concern gets handled in resolveOptions)

// in each method
const timeoutMS = options.timeoutMS ?? this.s.options.timeoutMS;
```

baileympearson (changes requested):

PR #4248, opened 2024-09-25
feat(NODE-6337): implement client bulk write batching

durran:
### Description

Implements batching of client bulk writes.

#### What is changing?
- Updates `DocumentSequence` to be able to serialize itself and return its documents byte length when pushing to it.
- Updates the command builder to batch bulk write commands based on the max message size or max batch size.
- Updates the result merger to properly link indexes back to the original user provided models.
- Exposes `maxMessageSizeBytes` and `maxWriteBatchSize` up to the topology level.
- Implements CRUD prose tests #3 and #4

None

NODE-6337


src/cmap/commands.ts
```diff
@@ -429,10 +429,66 @@ export interface OpMsgOptions {
...
+      }
+    }
+  }
+
+  /**
+   * Initialize the buffer chunks.
+   */
+  private init() {
+    // Document sequences starts with type 1 at the first byte.
+    const buffer = Buffer.allocUnsafe(1 + 4 + this.field.length + 1);
```
baileympearson, line 460: This came up in the initial implementation but I think it got missed.  Can we add a comment that we expect all document sequence keys to be utf8?  Otherwise, we'd want to use `Buffer.byteLength(this.field)`.

src/cmap/commands.ts
```diff
@@ -429,10 +429,66 @@ export interface OpMsgOptions {
 
 /** @internal */
 export class DocumentSequence {
+  field: string;
   documents: Document[];
+  serializedDocumentsLength: number;
+  private chunks: Uint8Array[];
+  private header?: Buffer;
```
baileympearson, line 436:
```suggestion
  private header: Buffer;
```

header can't be null, right?

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -62,32 +67,148 @@ export class ClientBulkWriteCommandBuilder {
...
+        }
+        // We've added a new namespace, increment the namespace index.
         currentNamespaceIndex++;
       }
     }
 
-    const nsInfo = Array.from(namespaces.keys(), ns => ({ ns }));
+    // After we've finisihed iterating all the models put the last current command
+    // only if there are operations in it.
+    if (currentCommand.ops.documents.length > 0) {
```
baileympearson, line 176: Good catch here!

src/cmap/commands.ts
```diff
@@ -429,10 +429,66 @@ export interface OpMsgOptions {
...
+    this.chunks.push(buffer);
+    this.header = buffer;
+  }
+
+  /**
+   * Push a document to the document sequence. Will serialize the document
+   * as well and return the current serialized length of all documents.
+   * @param document - The document to add.
+   * @param buffer - The serialized document in raw BSON.
+   * @returns The new totoal document sequence length.
```
baileympearson, line 473:
```suggestion
   * @returns The new total document sequence length.
```

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -62,32 +67,148 @@ export class ClientBulkWriteCommandBuilder {
...
+        const operationBuffer = BSON.serialize(operation);
+
+        // Check if the operation buffer can fit in the current command. If it can,
+        // then add the operation to the document sequence and increment the
+        // current length as long as the ops don't exceed the maxWriteBatchSize.
+        if (
+          currentCommandLength + operationBuffer.length < maxMessageSizeBytes &&
+          currentCommand.ops.documents.length < maxWriteBatchSize
+        ) {
+          // Pushing to the ops document sequence returns the bytes length added.
```
baileympearson, line 108:
```suggestion
          // Pushing to the ops document sequence returns the total byte length of the document sequence.
```

baileympearson: (this applies to other places in this file too)

baileympearson (changes requested):

src/sdam/server_description.ts
```diff
@@ -18,6 +18,12 @@ const DATA_BEARING_SERVER_TYPES = new Set<ServerType>([
   ServerType.LoadBalancer
 ]);
 
+/** Default in case we are in load balanced mode. */
+const MAX_MESSAGE_SIZE_BYTES = 48000000;
```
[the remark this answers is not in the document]
durran: This is what the 8.0 server returns in its hello responses, so since they are only being used for bulk write I used that. The spec doesn't mention this but without a monitoring connection in load balanced mode we have to default this to something or we cannot batch.

src/operations/client_bulk_write/executor.ts
```diff
@@ -49,6 +50,23 @@ export class ClientBulkWriteExecutor {
    * @returns The result.
    */
   async execute(): Promise<ClientBulkWriteResult | { ok: 1 }> {
+    const topologyDescription = this.client.topology?.description;
```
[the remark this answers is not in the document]
durran: 1. It is determined before server selection if the monitoring connections have gotten their first responses, otherwise it would be on the initial handshake of the connection. See #2 on complexity.
2. It seemed strange to me to have access to a connection itself at this level and wasn't sure what that would really look like. I guess we'd have to select a server and check out a connection before even building the commands, then somehow pass them both down through the cursor and operation levels. It seemed a bit heavy compared to just having a sensible default.

src/operations/client_bulk_write/command_builder.ts
durran: Updated.

src/cmap/commands.ts
durran: Updated.

src/cmap/commands.ts
durran: Updated so it is never undefined.

src/cmap/commands.ts
durran: Added command.

src/operations/client_bulk_write/executor.ts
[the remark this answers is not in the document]
baileympearson: This works if the user has connected their client (or at least, monitoring has been started).  But this won't work with auto connected, which we do need to support.  Although, if we need to get this out for server 8.0, auto connection support could be a follow up.

I think there are two related questions here:

- how can we make sure this works for auto connected clients?  (usually we do this by 
- how can this work in LB mode?  I [asked in the bulk write channel](https://mongodb.slack.com/archives/C035ZJL6CQN/p1727392398800859) what other drivers do.  If nobody has anything better, I think defaulting is fine for now because nobody can use this on servers with values different than the defaults you've used from server 8.0.  But we probably should figure out a way to determine the actual server limits in LB mode.

Thoughts?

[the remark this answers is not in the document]
durran: Well since I've got the weekend as well, I think I can try for all of the new bulk write a similar solution to what we did for LB mode in general with connection pinning. After thinking about this more I think this current implementation is incorrect but no unified test or prose test is exposing it. I think this would mean that:

1. Server selection and connection checkout will need to happen in the `BulkWriteExecutor`.
2. Implicit session creation will need to happen at this layer now, and passed down to the cursor. (I actually think this is correct now, and the current implementation is incorrect with a potential new implicit session per cursor)
3. We pin the connection to the session like we do in LB mode, and unpin when the executor is finished.

I think this is fairly straightforward. Thoughts?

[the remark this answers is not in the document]
baileympearson: Are you imagining that the executor will run server selection itself and connection checkout itself?  I'm not a fan of that idea because we've put a lot of effort into ensuring all operations are executed by execute_operation (which handles server selection, retryablity, session acquisition, SDAM and CMAP updating after command success/failures, etc).   (correct me if I misunderstood your proposal)

I think there are 2 different pieces of work here (depending on the approach we take).  

1. we need to get batch construction done after server selection so that auto-connect works
2. we need to build commands after connection checkout.  I think this applies to the whole driver - any command that is obtaining metadata from the selected server's server description won't have accurate info in LB mode (I think you encountered this recently in a different ticket?).

I think 1 can be done post-release if necessary.  2 definitely can be done post release - that's a pretty big undertaking across the whole driver.

What would you think of making the builder lazy (so that it produces on-demand) and then attaching it as a property on the operation?  Then the operation's execute method can build the next batch on-demand with the selected server's information.

```typescript
// in the executor:
const builder = new Builder(writeModels);
while (builder.hasBatch()) {
  const operation = new ClientBulkWriteOperation(..., builder);
  await executeOperation(operation);
}

// ClientBulkWriteOperation.execute();
execute(server, session) {
  const batch = this.builder.nextBatch(server.description.maxBSONSize, server.description.maxMessageSize);
  return server.command(batch);
}
```

That doesn't solve our LB mode issues though, but I think we could get away with using defaults for now.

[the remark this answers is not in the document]
durran: Just a note from offline the command builder has been adjusted to return one command at a time and we've moved the building of each command to post executeOperation so we are sure we've got a live connection. LB mode in this case will move pinning up one step.

src/operations/client_bulk_write/client_bulk_write.ts
```diff
@@ -37,9 +37,34 @@ export class ClientBulkWriteOperation extends CommandOperation<ClientBulkWriteCu
...
+    if (server.description.type === ServerType.LoadBalancer) {
+      // Checkout a connection to build the command.
+      const connection = await server.pool.checkOut();
+      command = this.commandBuilder.buildBatch(
+        connection.hello?.maxMessageSizeBytes,
+        connection.hello?.maxWriteBatchSize
+      );
+      // Pin the connection to the session so it get used to execute the command and we do not
+      // perform a double check-in/check-out.
+      session?.pin(connection);
```
baileympearson, line 51:
```typescript
const connection = await server.pool.checkOut();
session?.pin(connection);
command = this.commandBuilder.buildBatch(
  connection.hello?.maxMessageSizeBytes,
  connection.hello?.maxWriteBatchSize
);
```

flip the order, perhaps, in case buildBatch throws for some reason?

Also - I don't think session can ever be undefined but if it is, we'd leak a connection.  Should we maybe assert that it's defined?

baileympearson (changes requested):
small changes!

src/operations/client_bulk_write/client_bulk_write.ts
durran: I've updated per the suggestions here.

src/sdam/server_description.ts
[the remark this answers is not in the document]
durran: Removed.

src/sdam/server_description.ts
```diff
@@ -111,6 +117,9 @@ export class ServerDescription {
     this.setVersion = hello?.setVersion ?? null;
     this.electionId = hello?.electionId ?? null;
     this.logicalSessionTimeoutMinutes = hello?.logicalSessionTimeoutMinutes ?? null;
+    this.maxMessageSizeBytes = hello?.maxMessageSizeBytes ?? 0;
```
baileympearson, line 120: I don't know if setting it to zero is the correct default - I think `null` to indicate "not provided" is preferable.  `0` could cause other parts of the driver to behave strangely whereas null will force users of this data to check is its defined and handle the case where its not explicitly.

Although, I wish there were a way we could make these fields required on the `hello` somehow.  I'd be surprised if they're ever undefined on our supported server versions.

durran: The reason I defaulted to 0 here instead of null is that the end result is going to be the same but without having the extra checks for existence. If any of these values is either 0 or null no messages will be able to be added - and since these are new properties on the server description they haven't been used anywhere else as of yet. Let me know if you feel strongly or not about it.

durran: Per conversation we've defaulted to null and will throw later if they are actually null when building the command batches.

baileympearson (approved):

PR #4249, opened 2024-09-25
fix(NODE-6394): pause message stream when not listening for events

baileympearson:
#### What is changing?

The message stream is paused when we're not reading from it, so we do not miss `data` events.

### Fixed indefinite hang bug for high write load scenarios

When performing large and many write operations, the driver will likely encounter buffering at the socket layer. The logic that waited until buffered writes were complete would mistakenly drop `'data'` (reading from the socket), causing the driver to hang indefinitely or until a socket timeout. Using pausing and resuming mechanisms exposed by Node streams we have eliminated the possibility for data events to go unhandled. 

Shout out to @hunkydoryrepair for debugging and finding this issue!

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -237,6 +237,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       .on('error', this.onError.bind(this));
     this.socket.on('close', this.onClose.bind(this));
     this.socket.on('timeout', this.onTimeout.bind(this));
+
+    this.messageStream.pause();
   }
 
   public get hello() {
@@ -651,6 +653,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   private async *readMany(): AsyncGenerator<OpMsgResponse | OpReply> {
     try {
       this.dataEvents = onData(this.messageStream);
+      this.messageStream.resume();
       for await (const message of this.dataEvents) {
         const response = await decompressResponse(message);
         yield response;
@@ -661,6 +664,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       }
     } finally {
       this.dataEvents = null;
+      this.messageStream.pause();
       this.throwIfAborted();
     }
   }
--- a/test/unit/cmap/connection.test.ts
+++ b/test/unit/cmap/connection.test.ts
@@ -1,11 +1,15 @@
+import { Socket } from 'node:net';
+
 import { expect } from 'chai';
 import * as sinon from 'sinon';
+import { setTimeout } from 'timers/promises';
 
 import {
   connect,
   Connection,
   isHello,
   MongoClientAuthProviders,
+  MongoDBCollectionNamespace,
   MongoNetworkTimeoutError,
   ns
 } from '../../mongodb';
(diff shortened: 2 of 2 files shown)


src/cmap/connection.ts
```diff
@@ -237,6 +237,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       .on('error', this.onError.bind(this));
     this.socket.on('close', this.onClose.bind(this));
     this.socket.on('timeout', this.onTimeout.bind(this));
+
+    this.messageStream.pause();
```
baileympearson, line 241: Streams start paused, so I'm happy to remove this.  This just made it a bit more explicit - without this and knowing that streams start paused, it wasn't immediately obvious how the new tests pass for the first message from the message stream.

test/unit/cmap/connection.test.ts
```diff
@@ -142,4 +146,181 @@ describe('new Connection()', function () {
...
+    });
+
+    const validResponse = Buffer.from(
+      'a30000002a0800004b010000dd07000000000000008e000000016f6b00000000000000f03f0324636c757374657254696d65005800000011636c757374657254696d65001c00000093f6f266037369676e61747572650033000000056861736800140000000072d8d6eab4e0703d2d50846e2db7adb5d2733cc4126b65794964000200000026f6f2660000116f7065726174696f6e54696d65001c00000093f6f26600',
+      'hex'
+    );
+
+    const chunks = [validResponse.slice(0, 10), validResponse.slice(10)];
+
+    describe('when data is emitted before drain', function () {
```
baileympearson, line 174: If the changes in `connection.ts` are reverted, most of these tests pass.  

1. all the `first command` tests pass because the message stream starts paused.
2. the `when there is a delay between data and drain` test in the not first command section fails.  
3. everything else passes.

I did some debugging, and determined that there is some async delay after the message stream calls its callback with the full message and when the message stream emits a data event.  without digging further into Nodejs source, I'm not sure 1. if this is expected and 2. why this is.

I'm not convinced it matters for these tests - I think that there's adequate coverage of the various scenarios here that is resilient to changes in async behavior.

test/unit/cmap/connection.test.ts
```diff
@@ -142,4 +146,181 @@ describe('new Connection()', function () {
...
+        describe('when there is no delay between data and drain', function () {
+          it('does not hang', async function () {
+            const result$ = connection.command(
+              MongoDBCollectionNamespace.fromString('foo.bar'),
+              { ping: 1 },
+              {}
+            );
+            // there is an await in writeCommand, we must move the event loop forward just enough
+            // so that we reach the `await drain`.  Otherwise, we'll emit both data and drain before
+            // listeners are attached.
```
baileympearson, line 185: I'm not thrilled about this setTimeout.  But the only alternative I could think of was to set up a promise that resolves when `mockSocket.write()` is called, which also seems like too much work.  So I left it as-is.

src/cmap/connection.ts
addaleax: > Streams start paused, so I'm happy to remove this.

(This probably isn't particularly relevant to this particular PR, but just as a fun fact, stream's don't act as if `.pause()` or `.resume()` had been called on them initially; instead, they are in a secret third state (in particular, `.on('data', ...)` implicitly starts reading, which it doesn't if it's a `.pause()`d stream). A topic that's been brought up from time to time in Node.js core is actually that there's no way to "reset" a stream into its initial state once either of these methods have been called 🙂)

baileympearson: That's actually included in the documentation now (maybe it has been for a while?  I found it earlier today): https://nodejs.org/api/stream.html#implementing-a-transform-stream:~:text=data%27%20event%20handler.-,Three%20states,-%23

But for us, it just means that we won't miss data events on the first message.  Subsequent messages can miss data events, because `onData` removes its `data` listener but doesn't pause the message stream in its cleanup logic.

src/cmap/connection.ts
```diff
@@ -427,11 +429,8 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
   ): AsyncGenerator<MongoDBResponse> {
     this.throwIfAborted();
 
-    if (typeof options.socketTimeoutMS === 'number') {
-      this.socket.setTimeout(options.socketTimeoutMS);
-    } else if (this.socketTimeoutMS !== 0) {
-      this.socket.setTimeout(this.socketTimeoutMS);
-    }
+    const socketTimeoutMS = options.socketTimeoutMS ?? this.socketTimeoutMS ?? 0;
+    this.socket.setTimeout(socketTimeoutMS);
```
nbbeeken, line 433: Could this change be done separately from this bug fix or does it relate directly? IIRC the "not zero" condition was actually significant rather than an elimination of null (where as the typeof number was probably just eliminating null, but I do question removing the type safety)

src/cmap/connection.ts
```diff
@@ -455,11 +454,7 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
         yield document;
         this.throwIfAborted();
 
-        if (typeof options.socketTimeoutMS === 'number') {
-          this.socket.setTimeout(options.socketTimeoutMS);
-        } else if (this.socketTimeoutMS !== 0) {
-          this.socket.setTimeout(this.socketTimeoutMS);
-        }
+        this.socket.setTimeout(socketTimeoutMS);
```
nbbeeken, line 457: Same change. leaving a comment to track.

src/cmap/connection.ts
nbbeeken: I'm fine with calling pause out the gate, I don't think we have use for that "third" state, but very interesting to know!

test/unit/cmap/connection.test.ts
nbbeeken: If I understand correctly, that is a guarantee. The stream makes sure the callback is run "nextTick" (might use a different API) so that you don't get an inconsistent execution experience. Otherwise sometimes (ex. `write(.., cb); nextLine;`) `cb` would run before `nextLine`

I agree, this coverage is worth having, maybe future stream implementations or different runtimes could introduce this scenario and we're defended against it

test/unit/cmap/connection.test.ts
nbbeeken: Seems fine

nbbeeken (changes requested):

src/cmap/connection.ts
baileympearson: Yeah, it can be done separately.  I can revert this if you'd like, just let me know.  But to your point - it makes no difference:

- when not in use, socket timeouts are set to 0.
- if the options socket timeout is non-null, we should use that.
- if the socket timeout on the connection is non-null, we should use that.  if it is explicitly zero, then this.socket.setTimeout(0) is no change in behavior from _not_ setting it, because it's already defaulted to 0.

test/unit/cmap/connection.test.ts
baileympearson: The `callback` that users provide to `write()` is not the same as the callback applied in `transform` (see [this](https://nodejs.org/api/stream.html#transform_transformchunk-encoding-callback:~:text=The%20callback%20function%20must%20be%20called%20only,In%20other%20words%2C%20the%20following%20are%20equivalent%3A)).  

Also, I think that for transform streams the behavior you've mentioned is not guaranteed - it would depend on whether the implementation of `write` is synchronous or not (there's no guarantee, best I can tell: https://nodejs.org/api/stream.html#transform_transformchunk-encoding-callback:~:text=transform._transform()%20is%20never%20called%20in%20parallel%3B%20streams%20implement%20a%20queue%20mechanism%2C%20and%20to%20receive%20the%20next%20chunk%2C%20callback%20must%20be%20called%2C%20either%20synchronously%20or%20asynchronously.). 

What you can rely on being true is that the ordering of the following is strictly defined:

```typescript
transform.write(chunk, () => { 
	nextLine;
});
```

src/cmap/connection.ts
nbbeeken: I think you are correct, I just remember writing this exact improvement and having to revert it back to these conditions when I refactored the connection layer. Digging through history now, I'll let you know if I come up empty.

nbbeeken: FYI sent you some historical context on slack. Let me know if that changes things here. But I think we can keep this once we get our CI back on track and nothing fails.

baileympearson: Given the unreliability of our CI right now, I reverted the changes.

nbbeeken (approved):

nbbeeken:
Failures unrelated

hunkydoryrepair:
any chance of a patch release until 6.10 is out? Maybe a 6.9.1? Especially since 6.10 removes mongo 3.6 support, some won't have option to upgrade.

alexbevi:
Based on the [MongoDB Software Lifecycle Schedules](https://www.mongodb.com/legal/support-policy/lifecycles) MongoDB 3.6 reached EOL in April 2021.

It's unlikely there'd be enough demand to support this server version to act as a forcing function for a patch release.

hunkydoryrepair:
Fair enough. we are stuck on 6.3 since this has been broken for a while and would love to be able to use the optimizations done since then.

PR #4252, opened 2024-09-25
refactor(NODE-6398): bulkWrite internals to use async/await

nbbeeken:
#### What is changing?

- Refactor executeCommands to use async/await

No

Stack traces, debugging

### DIFF
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -557,20 +557,25 @@ async function executeCommands(
     }
 
     if (thrownError != null) {
-      if (!(thrownError instanceof MongoWriteConcernError)) {
+      if (thrownError instanceof MongoWriteConcernError) {
+        mergeBatchResults(batch, bulkOperation.s.bulkResult, thrownError, result);
+        const writeResult = new BulkWriteResult(
+          bulkOperation.s.bulkResult,
+          bulkOperation.isOrdered
+        );
+
+        throw new MongoBulkWriteError(
+          {
+            message: thrownError.result.writeConcernError.errmsg,
+            code: thrownError.result.writeConcernError.code
+          },
+          writeResult
+        );
+      } else {
         // Error is a driver related error not a bulk op error, return early
         throw new MongoBulkWriteError(
           thrownError,
           new BulkWriteResult(bulkOperation.s.bulkResult, bulkOperation.isOrdered)
-        );
-      }
-
-      if (thrownError instanceof MongoWriteConcernError) {
-        handleMongoWriteConcernError(
-          batch,
-          bulkOperation.s.bulkResult,
-          bulkOperation.isOrdered,
-          thrownError
         );
       }
     }
@@ -585,23 +590,6 @@ async function executeCommands(
   const writeResult = new BulkWriteResult(bulkOperation.s.bulkResult, bulkOperation.isOrdered);
   bulkOperation.handleWriteError(writeResult);
   return writeResult;
-}
-
-function handleMongoWriteConcernError(
-  batch: Batch,
-  bulkResult: BulkResult,
-  isOrdered: boolean,
-  err: MongoWriteConcernError
-): never {
-  mergeBatchResults(batch, bulkResult, undefined, err.result);
-
-  throw new MongoBulkWriteError(
-    {
-      message: err.result.writeConcernError.errmsg,
-      code: err.result.writeConcernError.code
-    },
-    new BulkWriteResult(bulkResult, isOrdered)
-  );
 }
 
 /**
--- a/src/bulk/common.ts
+++ b/src/bulk/common.ts
@@ -1,5 +1,3 @@
-import { promisify } from 'util';
-
 import { type BSONSerializeOptions, type Document, EJSON, resolveBSONOptions } from '../bson';
 import type { Collection } from '../collection';
 import {
@@ -7,6 +5,7 @@ import {
   MongoBatchReExecutionError,
   MONGODB_ERROR_CODES,
   MongoInvalidArgumentError,
+  MongoRuntimeError,
   MongoServerError,
   MongoWriteConcernError
 } from '../error';
@@ -22,7 +21,6 @@ import type { Topology } from '../sdam/topology';
 import type { ClientSession } from '../sessions';
 import {
   applyRetryableWrites,
-  type Callback,
   getTopology,
   hasAtomicOperators,
   maybeAddIdToDocuments,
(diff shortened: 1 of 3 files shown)


src/bulk/common.ts
```diff
@@ -500,86 +498,46 @@ export function mergeBatchResults(
...
   if (bulkOperation.s.batches.length === 0) {
-    return callback(
-      undefined,
-      new BulkWriteResult(bulkOperation.s.bulkResult, bulkOperation.isOrdered)
-    );
+    return new BulkWriteResult(bulkOperation.s.bulkResult, bulkOperation.isOrdered);
   }
 
-  const batch = bulkOperation.s.batches.shift() as Batch;
+  for (const batch of bulkOperation.s.batches) {
```
nbbeeken, line 509: Note to reviewers: take a careful look at how we interact with batches, previously we shifted the first batch out, and "looped" until the condition above, when no batches remain. I have changed this to just a loop that does not modify the batches until the end:

```ts
bulkOperation.s.batches.length = 0;
```

src/bulk/common.ts
```diff
@@ -500,86 +498,46 @@ export function mergeBatchResults(
...
+      finalOptions.checkKeys = false;
     }
 
-    if (isDeleteBatch(batch)) {
-      finalOptions.retryWrites =
-        finalOptions.retryWrites && !batch.operations.some(op => op.limit === 0);
+    if (finalOptions.retryWrites) {
+      if (isUpdateBatch(batch)) {
+        finalOptions.retryWrites =
+          finalOptions.retryWrites && !batch.operations.some(op => op.multi);
```
nbbeeken, line 532: operations is never empty at this point, so this won't hit the "return false for empty" case. 

Also this is n^2 :( should we / can we move this out of the for loop?

src/bulk/common.ts
```diff
@@ -500,86 +498,46 @@ export function mergeBatchResults(
...
+      }
+
+      if (isDeleteBatch(batch)) {
+        finalOptions.retryWrites =
+          finalOptions.retryWrites && !batch.operations.some(op => op.limit === 0);
```
nbbeeken, line 537: n^2

test/integration/crud/crud_api.test.ts
```diff
@@ -1093,32 +1093,24 @@ describe('CRUD API', function () {
...
-    },
-
-    test: async function () {
-      const ops = [];
-      // Create a set of operations that go over the 1000 limit causing two messages
-      let i = 0;
-      for (; i < 1005; i++) {
-        ops.push({ insertOne: { _id: i, a: i } });
-      }
+  it('should correctly throw error on illegal callback when unordered bulkWrite encounters error', async function () {
```
nbbeeken, line 1096: drive by cleanup as I fixed a bug caught here. Nothing functionally changed.

src/bulk/unordered.ts
```diff
@@ -20,12 +19,12 @@ export class UnorderedBulkOperation extends BulkOperationBase {
     super(collection, options, false);
   }
 
-  override handleWriteError(callback: Callback, writeResult: BulkWriteResult): boolean {
+  override handleWriteError(writeResult: BulkWriteResult): void {
     if (this.s.batches.length) {
-      return false;
+      return;
```
nbbeeken, line 24: length is always a truthy value while in the for loop, after we finish the array is cleared so we will go to the next line which reuses the logic in the parent class. This is the difference between an ordered operation that is interrupted by an error and an unordered one that settles everything and you check the result for what succeeded and what did not.

src/bulk/common.ts
```diff
@@ -588,38 +546,61 @@ function executeCommands(
           ? new DeleteOperation(bulkOperation.s.namespace, batch.operations, finalOptions)
           : null;
 
-    if (operation != null) {
-      executeOperation(bulkOperation.s.collection.client, operation).then(
-        result => resultHandler(undefined, result),
-        error => resultHandler(error)
-      );
+    if (operation == null) throw new MongoRuntimeError(`Unknown batchType: ${batch.batchType}`);
```
nbbeeken, line 549: TS narrowing 🤷🏻 what can ya do

src/bulk/common.ts
```diff
@@ -588,38 +546,61 @@ function executeCommands(
...
+
+    let thrownError = null;
+    let result;
+    try {
+      result = await executeOperation(bulkOperation.s.collection.client, operation);
+    } catch (error) {
+      thrownError = error;
+    }
```
nbbeeken, line 557: This may not be our preferred structure for error handling, but to reduce refactoring, I need to have the error and result available to me at the same time so I can call mergeBatchResults the same way it is written now.

src/bulk/common.ts
```diff
@@ -500,86 +498,46 @@ export function mergeBatchResults(
...
-    if (err && 'message' in err && !(err instanceof MongoWriteConcernError)) {
-      return callback(
-        new MongoBulkWriteError(
-          err,
-          new BulkWriteResult(bulkOperation.s.bulkResult, bulkOperation.isOrdered)
-        )
-      );
+    if (finalOptions.bypassDocumentValidation !== true) {
+      delete finalOptions.bypassDocumentValidation;
     }
```
baileympearson, line 517: Did you intentionally remove `operationId`?

src/bulk/common.ts
baileympearson: This is spec-compliant:

> Supported multi-statement write operations include insertMany() and bulkWrite(). The ordered option may be true or false. For both the collection-level and client-level bulkWrite() methods, a bulk write batch is only retryable if it does not contain any multi: true writes (i.e. UpdateMany and DeleteMany). Drivers MUST evaluate eligibility for each write command sent as part of the bulkWrite() (after order and batch splitting) individually. Drivers MUST NOT alter existing logic for order and batch splitting in an attempt to maximize retryability for operations within a bulk write.

We could determine retryability per-batch when constructing the batch though.

(also - I think this is still O(n).  We iterate over all mdoels once to build batches, and then each batch iterates over only its operations, not all the operations).

src/bulk/common.ts
baileympearson: I'm not sure I follow.  What problematic narrowing are you referring to here? 

On line 347 we explicitly assign to `null`, so we need to narrow.  Alternatively, the ternary could just throw if an unexpected batch type was encountered (MongoRuntimeError - should never happen).

src/bulk/common.ts
```diff
@@ -588,38 +546,61 @@ function executeCommands(
...
-    // Merge top level error and return
-    mergeBatchResults(batch, bulkOperation.s.bulkResult, err, undefined);
-    callback();
+
+    mergeBatchResults(batch, bulkOperation.s.bulkResult, thrownError, result);
+    const writeResult = new BulkWriteResult(bulkOperation.s.bulkResult, bulkOperation.isOrdered);
+    bulkOperation.handleWriteError(writeResult);
   }
+
+  bulkOperation.s.batches.length = 0;
```
baileympearson, line 583: Are you assiging length to 0 to perserve the behavior where `batches.length = 0` after a bulk write?

If so, this is broken when we encounter errors.  Previously, the batchs' length would be accurate.  Now its always equal to the total number of batches.

src/bulk/common.ts
```diff
@@ -588,38 +546,61 @@ function executeCommands(
...
 }
 
 function handleMongoWriteConcernError(
   batch: Batch,
   bulkResult: BulkResult,
   isOrdered: boolean,
-  err: MongoWriteConcernError,
-  callback: Callback<BulkWriteResult>
-) {
+  err: MongoWriteConcernError
```
baileympearson, line 594: Thoughts on inlining this method?  Its only called in one place

src/bulk/common.ts
```diff
@@ -588,38 +546,61 @@ function executeCommands(
...
+    if (thrownError != null) {
+      if (!(thrownError instanceof MongoWriteConcernError)) {
+        // Error is a driver related error not a bulk op error, return early
+        throw new MongoBulkWriteError(
+          thrownError,
+          new BulkWriteResult(bulkOperation.s.bulkResult, bulkOperation.isOrdered)
+        );
+      }
+
+      if (thrownError instanceof MongoWriteConcernError) {
```
baileympearson, line 568: This if condition is unnecessary, right?

test/integration/crud/crud_api.test.ts
baileympearson: Can we fix the test description?  I think this is testing that when a multi-batch unordered bulk write encounters an error, we handle the error properly and return an error to the user?

baileympearson (changes requested):

src/bulk/common.ts
nbbeeken: Yea you got it right, I just called it out for the reviewer that is wondering why it is necessary. 

I can't make the ternary throw (yet!) so this is the best we got.

src/bulk/common.ts
nbbeeken: done!

test/integration/crud/crud_api.test.ts
nbbeeken: Updated the description, also could not resist improving the assertions a bit too to match the title.

src/bulk/common.ts
baileympearson: Yes, it's effectively just an `else` block

src/bulk/common.ts
baileympearson: Okay, if this isn't public, I don't understand why we need to clear out the batches after successfully executing all of them.  Can you elaborate?

nbbeeken: In the unordered case, [we jump over handling errors until the batches are empty](https://github.com/mongodb/node-mongodb-native/pull/4252#discussion_r1781538410).

PR #4256, opened 2024-09-27
feat(NODE-6387): Add CSOT support to change streams

W-A-James:
##### `ChangeStream`
* Added support for CSOT to change stream iterator mode methods (`ChangeStream.next()`, `ChangeStream.hasNext()`, `ChangeStream.tryNext()`)
* Added support for CSOT to change stream emitter mode
	* Change streams will now emit an error event with a `MongoOperationTimeoutError` when the internal cursor iteration times out

##### `Connection`
* Close socket when a command times out (temporary change while #4273 is in progress)

##### Testing
* Unskip and fix change stream prose test
* Unskip change stream tests
* Add integration tests to verify that CSOT behviour for ChangeStream emitter mode

Yes, punted to NODE-5688

NODE-6387

### DIFF
--- a/src/change_stream.ts
+++ b/src/change_stream.ts
@@ -3,7 +3,7 @@ import type { Readable } from 'stream';
 import type { Binary, Document, Timestamp } from './bson';
 import { Collection } from './collection';
 import { CHANGE, CLOSE, END, ERROR, INIT, MORE, RESPONSE, RESUME_TOKEN_CHANGED } from './constants';
-import type { AbstractCursorEvents, CursorStreamOptions } from './cursor/abstract_cursor';
+import { type CursorStreamOptions, CursorTimeoutContext } from './cursor/abstract_cursor';
 import { ChangeStreamCursor, type ChangeStreamCursorOptions } from './cursor/change_stream_cursor';
 import { Db } from './db';
 import {
@@ -11,6 +11,7 @@ import {
   isResumableError,
   MongoAPIError,
   MongoChangeStreamError,
+  MongoOperationTimeoutError,
   MongoRuntimeError
 } from './error';
 import { MongoClient } from './mongo_client';
@@ -20,6 +21,7 @@ import type { CollationOptions, OperationParent } from './operations/command';
 import type { ReadPreference } from './read_preference';
 import { type AsyncDisposable, configureResourceManagement } from './resource_management';
 import type { ServerSessionId } from './sessions';
+import { CSOTTimeoutContext, type TimeoutContext } from './timeout';
 import { filterOptions, getTopology, type MongoDBNamespace, squashError } from './utils';
 
 /** @internal */
@@ -538,7 +540,13 @@ export type ChangeStreamEvents<
   end(): void;
   error(error: Error): void;
   change(change: TChange): void;
-} & AbstractCursorEvents;
+  /**
+   * @remarks Note that the `close` event is currently emitted whenever the internal `ChangeStreamCursor`
+   * instance is closed, which can occur multiple times for a given `ChangeStream` instance.
+   * When this event is emitted is subject to change outside of major versions.
+   */
+  close(): void;
+};
 
 /**
  * Creates a new Change Stream instance. Normally created using {@link Collection#watch|Collection.watch()}.
@@ -609,6 +617,13 @@ export class ChangeStream<
    */
   static readonly RESUME_TOKEN_CHANGED = RESUME_TOKEN_CHANGED;
 
+  private timeoutContext?: TimeoutContext;
+  /**
+   * Note that this property is here to uniquely identify a ChangeStream instance as the owner of
+   * the {@link CursorTimeoutContext} instance (see {@link ChangeStream._createChangeStreamCursor}) to ensure
+   * that {@link AbstractCursor.close} does not mutate the timeoutContext.
+   */
+  private contextOwner: symbol;
   /**
    * @internal
    *
@@ -624,20 +639,25 @@ export class ChangeStream<
 
     this.pipeline = pipeline;
     this.options = { ...options };
+    let serverSelectionTimeoutMS: number;
     delete this.options.writeConcern;
 
     if (parent instanceof Collection) {
       this.type = CHANGE_DOMAIN_TYPES.COLLECTION;
+      serverSelectionTimeoutMS = parent.s.db.client.options.serverSelectionTimeoutMS;
     } else if (parent instanceof Db) {
       this.type = CHANGE_DOMAIN_TYPES.DATABASE;
+      serverSelectionTimeoutMS = parent.client.options.serverSelectionTimeoutMS;
     } else if (parent instanceof MongoClient) {
       this.type = CHANGE_DOMAIN_TYPES.CLUSTER;
+      serverSelectionTimeoutMS = parent.options.serverSelectionTimeoutMS;
     } else {
       throw new MongoChangeStreamError(
         'Parent provided to ChangeStream constructor must be an instance of Collection, Db, or MongoClient'
       );
     }
 
+    this.contextOwner = Symbol();
     this.parent = parent;
     this.namespace = parent.s.namespace;
     if (!this.options.readPreference && parent.readPreference) {
@@ -662,6 +682,13 @@ export class ChangeStream<
         this[kCursorStream]?.removeAllListeners('data');
       }
     });
+
+    if (this.options.timeoutMS != null) {
+      this.timeoutContext = new CSOTTimeoutContext({
+        timeoutMS: this.options.timeoutMS,
+        serverSelectionTimeoutMS
+      });
+    }
   }
 
   /** @internal */
@@ -681,22 +708,31 @@ export class ChangeStream<
     // This loop continues until either a change event is received or until a resume attempt
     // fails.
 
-    while (true) {
-      try {
-        const hasNext = await this.cursor.hasNext();
-        return hasNext;
-      } catch (error) {
+    this.timeoutContext?.refresh();
+    try {
+      while (true) {
+        const cursorInitialized = this.cursor.id != null;
         try {
-          await this._processErrorIteratorMode(error);
+          const hasNext = await this.cursor.hasNext();
+          return hasNext;
         } catch (error) {
           try {
-            await this.close();
+            await this._processErrorIteratorMode(error, cursorInitialized);
           } catch (error) {
-            squashError(error);
+            if (error instanceof MongoOperationTimeoutError && cursorInitialized) {
+              throw error;
+            }
+            try {
+              await this.close();
+            } catch (error) {
+              squashError(error);
+            }
+            throw error;
           }
-          throw error;
         }
       }
+    } finally {
+      this.timeoutContext?.clear();
     }
   }
 
@@ -706,24 +742,33 @@ export class ChangeStream<
     // Change streams must resume indefinitely while each resume event succeeds.
     // This loop continues until either a change event is received or until a resume attempt
     // fails.
+    this.timeoutContext?.refresh();
 
-    while (true) {
-      try {
-        const change = await this.cursor.next();
-        const processedChange = this._processChange(change ?? null);
-        return processedChange;
-      } catch (error) {
+    try {
+      while (true) {
+        const cursorInitialized = this.cursor.id != null;
         try {
-          await this._processErrorIteratorMode(error);
+          const change = await this.cursor.next();
+          const processedChange = this._processChange(change ?? null);
+          return processedChange;
         } catch (error) {
           try {
-            await this.close();
+            await this._processErrorIteratorMode(error, cursorInitialized);
           } catch (error) {
-            squashError(error);
+            if (error instanceof MongoOperationTimeoutError && cursorInitialized) {
+              throw error;
+            }
+            try {
+              await this.close();
+            } catch (error) {
+              squashError(error);
+            }
+            throw error;
           }
-          throw error;
         }
       }
+    } finally {
+      this.timeoutContext?.clear();
     }
   }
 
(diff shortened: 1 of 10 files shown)


src/cursor/abstract_cursor.ts
```diff
@@ -170,7 +170,9 @@ export abstract class AbstractCursor<
   private cursorClient: MongoClient;
   /** @internal */
   private transform?: (doc: TSchema) => any;
-  /** @internal */
+  /** @internal
+   * This is true whether or not the first command fails. It only indicates whether or not the first
+   * command has been run. */
```
nbbeeken, line 175:
```suggestion
  /** 
   * @internal
   * This is true whether or not the first command fails. It only indicates whether or not the first
   * command has been run. 
   */
```

src/cursor/abstract_cursor.ts
```diff
@@ -489,7 +493,7 @@ export abstract class AbstractCursor<
         await this.fetchBatch();
       } while (!this.isDead || (this.documents?.length ?? 0) !== 0);
     } finally {
-      if (this.cursorOptions.timeoutMode === CursorTimeoutMode.ITERATION && this.cursorId != null) {
+      if (shouldRefresh) {
```
nbbeeken, line 496: Since this line is after fetchBatch the value at cursorId could have changed right? so is the stored condition still the same one we want to refresh on?

src/cursor/change_stream_cursor.ts
```diff
@@ -55,7 +55,7 @@ export class ChangeStreamCursor<
     pipeline: Document[] = [],
     options: ChangeStreamCursorOptions = {}
   ) {
-    super(client, namespace, options);
+    super(client, namespace, { ...options, tailable: true, awaitData: true });
```
nbbeeken, line 58: Surprising these weren't set before but also why now?

src/change_stream.ts
```diff
@@ -609,6 +615,8 @@ export class ChangeStream<
    */
   static readonly RESUME_TOKEN_CHANGED = RESUME_TOKEN_CHANGED;
 
+  private timeoutContext?: TimeoutContext;
+  private symbol: symbol;
```
nbbeeken, line 619: Maybe `contextOwner` and a doc comment explaining what this is

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -435,11 +435,11 @@ describe('CSOT spec prose tests', function () {
...
-      it.skip('sends correct number of aggregate and getMores', metadata, async function () {
+      it('sends correct number of aggregate and getMores', metadata, async function () {
         const changeStream = client
           .db('db')
           .collection('coll')
-          .watch([], { timeoutMS: 20, maxAwaitTimeMS: 19 });
+          .watch([], { timeoutMS: 120, maxAwaitTimeMS: 10 });
         const maybeError = await changeStream.next().then(
           () => null,
           e => e
```
nbbeeken, line 445: Can we leave a comment about why we don't check for "return a change stream with a non-zero ID.".

test/integration/client-side-operations-timeout/node_csot.test.ts
```diff
@@ -804,6 +805,182 @@ describe('CSOT driver tests', metadata, () => {
...
+          cs = client.db('db').collection('coll').watch([], { timeoutMS: 120 });
+          const _changePromise = once(cs, 'change');
+          await once(cs.cursor, 'init');
+
+          await internalClient.db().admin().command(failpoint);
+        });
+
+        it('emits an error event', async function () {
+          let [err] = await once(cs, 'error'); // getMore failure
+          [err] = await once(cs, 'error'); // aggregate failure
```
nbbeeken, line 971: Do we want to check something about each error instead of just the last?

I am surprised we emit more than one error event. I supposed that is fine but is there a close event that follows all of these?

src/cmap/connection.ts
```diff
@@ -746,12 +747,17 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
...
-        throw new MongoOperationTimeoutError(
+        const error = new MongoOperationTimeoutError(
           `Timed out during socket read (${readError.duration}ms)`
         );
+        this.dataEvents = null;
+        this.onError(error);
+        throw error;
       }
-      throw readError;
+      this.cleanup(err);
```
nbbeeken, line 759: This isn't correct right? it should just be the clean up inside the timeout if block

src/cmap/connection.ts
```diff
@@ -712,10 +711,12 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       try {
         return await Promise.race([drainEvent, timeout]);
       } catch (error) {
+        let err = error;
         if (TimeoutError.is(error)) {
-          throw new MongoOperationTimeoutError('Timed out at socket write');
+          err = new MongoOperationTimeoutError('Timed out at socket write');
         }
-        throw error;
+        this.cleanup(error);
```
nbbeeken, line 718: Same Q, this cleanup is new for every error but if you saw something related to CSOT shouldn't it be conditional?

src/cursor/change_stream_cursor.ts
W-A-James: Setting those here allows the `AbstractCursor` constructor to do the appropriate validation and to correctly determine whether or not we send the `maxTimeMS` on the command

test/integration/client-side-operations-timeout/node_csot.test.ts
W-A-James: Good catch here. It does seem that there are two close events being emitted here; one for the initial timeout error that will be resumed and the second for the aggregate that would have been the resume attempt.

Would we prefer not to emit an error event on the getMore's timeout? We could instead add a custom timeout event that reports the timeout error but doesn't cause the change stream to emit the close event.

src/cursor/abstract_cursor.ts
W-A-James: Yes because the only time that `cursorId` is null is when the cursor is uninitialized, but to guard against any changes we make to that logic, we can check for existence of `timeoutContext` instead, because they're effectively testing the same thing.

nbbeeken: Let me know what you find. 

It sounds correct; my thinking is that if we've hit an error we should still have a null id because the operation didn't give us one, maybe we still set it to 0 somewhere or this logic runs in either case.

test/integration/client-side-operations-timeout/node_csot.test.ts
nbbeeken: Post the link here when you have it and we can resolve TY

nbbeeken (changes requested):
LGTM, just request me back when we can resolve the two threads. Moved to team review

test/integration/client-side-operations-timeout/node_csot.test.ts
W-A-James: Filed: NODE-6434

src/cursor/abstract_cursor.ts
W-A-James: The only two times we set the cursorId to null are in
* `AbstractCursor#constructor()`
* `AbstractCursor.rewind()`

Anywhere else, we are setting the `cursorId` from the server response (which is never null, provided that nothing went wrong server-side), so regardless of where we check the `cursorId`, in `AbstractCursor.next()/.tryNext()/.hasNext()`, it will still yield the correct behaviour of only calling `timeoutContext.refresh()/.clear()` when the cursor is already initialized and in iteration mode.

If it's initially null, then we'd still want to clear the clear the timeout if the cursor gets killed (id set to zero).

nbbeeken (approved):

test/integration/client-side-operations-timeout/node_csot.test.ts
```diff
@@ -808,6 +809,183 @@ describe('CSOT driver tests', metadata, () => {
...
+            configureFailPoint: 'failCommand',
+            mode: { times: 1 }, // fail twice to account for executeOperation's retry attempt
+            data: {
+              failCommands: ['aggregate'],
+              blockConnection: true,
+              blockTimeMS: 130
+            }
+          };
+
+          await internalClient.db().admin().command(failpoint);
```
baileympearson, line 864: Just sharing - I added `configureFailPoint`/`clearFailPoint` helpers to `test/utils.ts` to make it easier to setup and teardown failpoints.

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -436,11 +436,13 @@ describe('CSOT spec prose tests', function () {
        *    - Expect this to fail with a timeout error.
        * 1. Verify that an `aggregate` command and two `getMore` commands were executed against the `db.coll` collection during the test.
        */
-      it.skip('sends correct number of aggregate and getMores', metadata, async function () {
+      it('sends correct number of aggregate and getMores', metadata, async function () {
+        // NOTE: we don't check for a non-zero ID since we lazily send the initial aggregate to the
```
baileympearson, line 440: Our version of this would be to use `await cursorInit(changeStream)` (or `csInit`, can't remember the exact name) to force the initialization of the change stream without iterating it.  Then we can check its id.  Is there a reason we don't do that here?

src/cursor/abstract_cursor.ts
```diff
@@ -489,7 +493,7 @@ export abstract class AbstractCursor<
         await this.fetchBatch();
       } while (!this.isDead || (this.documents?.length ?? 0) !== 0);
     } finally {
-      if (this.cursorOptions.timeoutMode === CursorTimeoutMode.ITERATION && this.cursorId != null) {
+      if (this.cursorOptions.timeoutMode === CursorTimeoutMode.ITERATION) {
```
baileympearson, line 496: Just curious - why are we removing the check for `cursorId != null`?

src/cursor/change_stream_cursor.ts
baileympearson: Until CSOT, tailable / awaitData didn't impact the actual cursor commands we send to the server.  So it was implicitly a tailable await data cursor I guess, and our `Cursor.next()` logic works regardless of the cursor's configuration (we rely on the server to tell us when to stop iterating). :shrug

src/cursor/change_stream_cursor.ts
```diff
@@ -130,6 +130,7 @@ export class ChangeStreamCursor<
     const aggregateOperation = new AggregateOperation(this.namespace, this.pipeline, {
       ...this.cursorOptions,
       ...this.changeStreamCursorOptions,
+      omitMaxTimeMS: false,
```
baileympearson, line 133:
```suggestion
```

This should already be calculated and provided by the AbstractCursor, right?

src/change_stream.ts
```diff
@@ -538,7 +540,13 @@ export type ChangeStreamEvents<
   end(): void;
   error(error: Error): void;
   change(change: TChange): void;
-} & AbstractCursorEvents;
+  /**
+   * @remarks Note that the `close` event is currently emitted whenever the internal `ChangeStreamCursor`
+   * instance is closed, which can occur multiple times for a given `ChangeStream` instance.
+   * When this event is emitted is subject to change outside of major versions.
+   */
+  close(): void;
```
baileympearson, line 548: Should we instead emit a new `ChangeStreamCursor:Close` event from the change stream when the underlying cursor is closed?  If we think that's preferable, I suggest that we 1. update the comment only to mention that the current behavior is buggy and 2. file a ticket (that can be done outside of CSOT).

src/change_stream.ts
```diff
@@ -969,15 +1026,16 @@ export class ChangeStream<
...
       throw new MongoAPIError(CHANGESTREAM_CLOSED_ERROR);
     }
 
     if (
-      this.cursor.id == null ||
-      !isResumableError(changeStreamError, this.cursor.maxWireVersion)
+      !cursorInitialized ||
+      (!isResumableError(changeStreamError, this.cursor.maxWireVersion) &&
+        !(changeStreamError instanceof MongoOperationTimeoutError))
     ) {
```
baileympearson, line 1039: (optional) I struggle with lots of negated conditions - I find it easier to understand a single negation of a number of possive conditions.  Something like:

```suggestion
	const canResume = 
      cursorInitialized &&
      (isResumableError(changeStreamError) ||
        changeStreamError instanceof MongoOperationTimeoutError);
    if (!canResume) {
```

(`(!A || !B && !C) is equivalent to !(A && (B || C))`, I think)

src/change_stream.ts
```diff
@@ -609,6 +617,13 @@ export class ChangeStream<
    */
   static readonly RESUME_TOKEN_CHANGED = RESUME_TOKEN_CHANGED;
 
+  private timeoutContext?: TimeoutContext;
```
baileympearson, line 620: Do we ever actually instantiate a timeout context?  I don't see it anywhere in the change stream.

src/change_stream.ts
```diff
@@ -893,14 +941,17 @@ export class ChangeStream<
     const stream = this[kCursorStream] ?? cursor.stream();
     this[kCursorStream] = stream;
     stream.on('data', change => {
+      this.timeoutContext?.refresh();
```
baileympearson, line 944: When we refresh/clear the context here - what are we hoping to time out?  If we've entered this listener, we already have received a change event from the underlying cursor's stream.  The processing of events (this function) is synchronous, and there is nothing that waits on the context.

baileympearson (changes requested):

src/change_stream.ts
nbbeeken: de morgan's law!

src/change_stream.ts
W-A-James: We have a ticket filed to address this (NODE-6434), but I can your suggestion to emit the `ChangeStreamCursor:Close` event as a comment for consideration when we eventually go to refine that ticket.

src/change_stream.ts
W-A-James: Good catch, the timeout context here is getting refreshed and cleared by the underlying cursor and any timeouts happening there get bubbled up as error events.

src/change_stream.ts
nbbeeken: Something I was recently considering, we have that ticket set for v7, but wouldn't close have emitted many more times with the recent-ish bug report about close being emitted multiple times from a cursor?

I think we could consider it a bug that it happens more than once, but we can still do the work in that follow up ticket.

src/cursor/abstract_cursor.ts
W-A-James: Regardless of whether or not the cursor id is null after the fetch batch, we should still clear the timeout context since we would have initialized it in the `cursorInit` call.
I don't feel very strongly about this, so I can revert this if needed, but regardless of the path the cursor takes, clearing the timeoutContext at the end of the iteration is correct.

* uninitialized -> successfully initialized; timeoutContext is initialized, should be cleared
* uninitialized -> failed to initialize -> cleanup is called;  timeoutContext would have been cleared and set to undefined, but the nullish check makes this alright
* initialized -> successful getMore; timeout context should be cleared
* initialized -> failed getMore -> cleanup is called; same situation as failing to initialize

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
W-A-James: No reason in particular, I can do that here.

src/change_stream.ts
baileympearson: I don't think we should say "when we emit this event is subject to change inside major versions".  It should always be emitted exactly once, when the change stream itself is closed.  I agree with Neal that if that's now how its working now, that seems like a bug.

baileympearson (approved):

nbbeeken (approved):

PR #4257, opened 2024-09-30
feat(NODE-6350): add typescript support to client bulkWrite API

nbbeeken:
#### What is changing?

- Adds namespace to schema mapping type
- Adds tests/examples to our tsd tests

- Some important choices:
  - We should default only where needed ("near the top"), the Models now require an input schema, Document can be used when none is needed. 
  - `ClientBulkWriteModel` is exported on its own and is _not_ an array type. We can offer this seperate from the client API so annotations can be used to get this string to schema behavior (but I would advocate for putting it directly into the API)

Yes, I think the API doc on client.bulkWrite should explain how to use the TS a bit. I started in the TS doc for `ClientBulkWriteModel` 

We mostly use "inline" models to demo the API, however, this TS change does not cause friction in wrapping this API, it does have some considerations though, if you make a helper that returns InsertOneModels you will want the return type to still preserve the type of "namespace" where appropriate to get autocomplete and assertions.  -- we should write a small example of this.

Users of this method can get thier schema asserted on the various bulkWrite models making sure input data is not adjusted accidentally or that filters aren't written in a way that doees not match the data format at all. 

### New client bulk write API supported

A new bulk write API on the `MongoClient` is now supported for users on server versions 8.0 and higher.
This API is meant to replace the existing bulk write API on the `Collection` as it supports a bulk
write across multiple databases and collections in a single call.

#### Usage

Users of this API call `MongoClient#bulkWrite` and provide a list of bulk write models and options.
The models have a structure as follows:

#### Insert One

* Note that when no `_id` field is provided in the document, the driver will generate a BSON `ObjectId`
automatically. *

```ts
{
  namespace: '<db>.<collection>',
  name: 'insertOne',
  document: Document
}
```

#### Update One

```ts
{
  namespace: '<db>.<collection>',
  name: 'updateOne',
  filter: Document,
  update: Document | Document[],
  arrayFilters?: Document[],
  hint?: Document | string,
  collation?: Document,
  upsert: boolean
}
```

#### Update Many

* Note that write errors occuring with an update many model present are not retryable. *

```ts
{
  namespace: '<db>.<collection>',
  name: 'updateMany',
  filter: Document,
  update: Document | Document[],
  arrayFilters?: Document[],
  hint?: Document | string,
  collation?: Document,
  upsert: boolean
}
```

#### Replace One

```ts
{
  namespace: '<db>.<collection>',
  name: 'replaceOne',
  filter: Document,
  replacement: Document,
  hint?: Document | string,
  collation?: Document
}
```

#### Delete One

```ts
{
  namespace: '<db>.<collection>',
  name: 'deleteOne',
  filter: Document,
  hint?: Document | string,
  collation?: Document
}
```

#### Delete Many

* Note that write errors occuring with a delete many model present are not retryable. *

```ts
{
  namespace: '<db>.<collection>',
  name: 'deleteMany',
  filter: Document,
  hint?: Document | string,
  collation?: Document
}
```

#### Example

Below is a mixed model example of using the new API:

```ts
const client = new MongoClient(process.env.MONGODB_URI);
const models = [
  {
    name: 'insertOne',
    namespace: 'db.authors',
    document: { name: 'King' }
  },
  {
    name: 'insertOne',
    namespace: 'db.books',
    document: { name: 'It' }
  },
  {
    name: 'updateOne',
    namespace: 'db.books',
    filter: { name: 'it' },
    update: { $set: { year: 1986 } }
  }
];
const result = await client.bulkWrite(models);

```

The bulk write specific options that can be provided to the API are as follows:

- `ordered`: Optional boolean that indicates the bulk write as ordered. Defaults to true.
- `verboseResults`: Optional boolean to indicate to provide verbose results. Defaults to false.
- `bypassDocumentValidation`: Optional boolean to bypass document validation rules. Defaults to false.
- `let`: Optional document of parameter names and values that can be accessed using $$var. No default.

The object returned by the bulk write API is:

```ts
interface ClientBulkWriteResult {
  // Whether the bulk write was acknowledged.
  readonly acknowledged: boolean;
  // The total number of documents inserted across all insert operations.
  readonly insertedCount: number;
  // The total number of documents upserted across all update operations.
  readonly upsertedCount: number;
  // The total number of documents matched across all update operations.
  readonly matchedCount: number;
  // The total number of documents modified across all update operations.
  readonly modifiedCount: number;
  // The total number of documents deleted across all delete operations.
  readonly deletedCount: number;
  // The results of each individual insert operation that was successfully performed.
  // Note the keys in the map are the associated index in the models array.
  readonly insertResults?: ReadonlyMap<number, ClientInsertOneResult>;
  // The results of each individual update operation that was successfully performed.
  // Note the keys in the map are the associated index in the models array.
  readonly updateResults?: ReadonlyMap<number, ClientUpdateResult>;
  // The results of each individual delete operation that was successfully performed.
  // Note the keys in the map are the associated index in the models array.
  readonly deleteResults?: ReadonlyMap<number, ClientDeleteResult>;
}
```

#### Error Handling

Server side errors encountered during a bulk write will throw a `MongoClientBulkWriteError`. This error
has the following properties:

- `writeConcernErrors`: Ann array of documents for each write concern error that occurred.
- `writeErrors`: A map of index pointing at the models provided and the individual write error.
- `partialResult`: The client bulk write result at the point where the error was thrown.

#### Schema assertion support

```ts
interface Book {
  name: string;
  authorName: string;
}

interface Author {
  name: string;
}

type MongoDBSchemas = {
  'db.books': Book;
  'db.authors': Author;
}

const model: ClientBulkWriteModel<MongoDBSchemas> = {
  namespace: 'db.books'
  name: 'insertOne',
  document: { title: 'Practical MongoDB Aggregations', authorName: 3 } 
  // error `authorName` cannot be number
};
```

Notice how authorName is type checked against the `Book` type because namespace is set to `"db.books"`. 

### DIFF
--- a/test/integration/crud/crud.prose.test.ts
+++ b/test/integration/crud/crud.prose.test.ts
@@ -3,7 +3,7 @@ import { once } from 'events';
 
 import { type CommandStartedEvent } from '../../../mongodb';
 import {
-  type AnyClientBulkWriteModel,
+  type ClientBulkWriteModel,
   type ClientSession,
   type Collection,
   type Document,
@@ -176,7 +176,7 @@ describe('CRUD Prose Spec Tests', () => {
     // firstEvent.operationId is equal to secondEvent.operationId.
     let client: MongoClient;
     let maxWriteBatchSize;
-    let models: AnyClientBulkWriteModel<Document>[] = [];
+    let models: ClientBulkWriteModel<Document>[] = [];
     const commands: CommandStartedEvent[] = [];
 
     beforeEach(async function () {
@@ -244,7 +244,7 @@ describe('CRUD Prose Spec Tests', () => {
     let maxBsonObjectSize;
     let maxMessageSizeBytes;
     let numModels;
-    let models: AnyClientBulkWriteModel<Document>[] = [];
+    let models: ClientBulkWriteModel<Document>[] = [];
     const commands: CommandStartedEvent[] = [];
 
     beforeEach(async function () {
@@ -315,7 +315,7 @@ describe('CRUD Prose Spec Tests', () => {
     // Assert that two CommandStartedEvents were observed for the bulkWrite command.
     let client: MongoClient;
     let maxWriteBatchSize;
-    let models: AnyClientBulkWriteModel<Document>[] = [];
+    let models: ClientBulkWriteModel<Document>[] = [];
     const commands: CommandStartedEvent[] = [];
 
     beforeEach(async function () {
@@ -383,7 +383,7 @@ describe('CRUD Prose Spec Tests', () => {
     // Construct a list of write models (referred to as models) with model repeated maxWriteBatchSize + 1 times.
     let client: MongoClient;
     let maxWriteBatchSize;
-    let models: AnyClientBulkWriteModel<Document>[] = [];
+    let models: ClientBulkWriteModel<Document>[] = [];
     const commands: CommandStartedEvent[] = [];
 
     beforeEach(async function () {
@@ -391,7 +391,7 @@ describe('CRUD Prose Spec Tests', () => {
       await client.connect();
       await client.db('db').collection('coll').drop();
       const hello = await client.db('admin').command({ hello: 1 });
-      await client.db('db').collection('coll').insertOne({ _id: 1 });
+      await client.db('db').collection<{ _id?: number }>('coll').insertOne({ _id: 1 });
       maxWriteBatchSize = hello.maxWriteBatchSize;
 
       client.on('commandStarted', filterForCommands('bulkWrite', commands));
@@ -472,7 +472,7 @@ describe('CRUD Prose Spec Tests', () => {
     // Assert that a CommandStartedEvent was observed for the getMore command.
     let client: MongoClient;
     let maxBsonObjectSize;
-    const models: AnyClientBulkWriteModel<Document>[] = [];
+    const models: ClientBulkWriteModel<Document>[] = [];
     const commands: CommandStartedEvent[] = [];
 
     beforeEach(async function () {
@@ -546,7 +546,7 @@ describe('CRUD Prose Spec Tests', () => {
     let client: MongoClient;
     let session: ClientSession;
     let maxBsonObjectSize;
-    const models: AnyClientBulkWriteModel<Document>[] = [];
+    const models: ClientBulkWriteModel<Document>[] = [];
     const commands: CommandStartedEvent[] = [];
 
     beforeEach(async function () {
@@ -633,7 +633,7 @@ describe('CRUD Prose Spec Tests', () => {
     // Assert that a CommandStartedEvent was observed for the killCursors command.
     let client: MongoClient;
     let maxBsonObjectSize;
-    const models: AnyClientBulkWriteModel<Document>[] = [];
+    const models: ClientBulkWriteModel<Document>[] = [];
     const getMoreCommands: CommandStartedEvent[] = [];
     const killCursorsCommands: CommandStartedEvent[] = [];
 
@@ -804,7 +804,7 @@ describe('CRUD Prose Spec Tests', () => {
     let opsBytes;
     let numModels;
     let remainderBytes;
-    let models: AnyClientBulkWriteModel<Document>[] = [];
+    let models: ClientBulkWriteModel<Document>[] = [];
     const commands: CommandStartedEvent[] = [];
 
     beforeEach(async function () {
@@ -860,7 +860,7 @@ describe('CRUD Prose Spec Tests', () => {
       it('executes in a single batch', {
         metadata: { requires: { mongodb: '>=8.0.0', serverless: 'forbid' } },
         async test() {
-          const sameNamespaceModel: AnyClientBulkWriteModel<Document> = {
+          const sameNamespaceModel: ClientBulkWriteModel<Document> = {
             name: 'insertOne',
             namespace: 'db.coll',
             document: { a: 'b' }
@@ -897,7 +897,7 @@ describe('CRUD Prose Spec Tests', () => {
         metadata: { requires: { mongodb: '>=8.0.0', serverless: 'forbid' } },
         async test() {
           const namespace = `db.${'c'.repeat(200)}`;
-          const newNamespaceModel: AnyClientBulkWriteModel<Document> = {
+          const newNamespaceModel: ClientBulkWriteModel<Document> = {
             name: 'insertOne',
             namespace: namespace,
             document: { a: 'b' }
@@ -951,7 +951,7 @@ describe('CRUD Prose Spec Tests', () => {
       it('raises a client error', {
         metadata: { requires: { mongodb: '>=8.0.0', serverless: 'forbid' } },
         async test() {
-          const model: AnyClientBulkWriteModel<Document> = {
+          const model: ClientBulkWriteModel<Document> = {
             name: 'insertOne',
             namespace: 'db.coll',
             document: { a: 'b'.repeat(maxMessageSizeBytes) }
@@ -977,7 +977,7 @@ describe('CRUD Prose Spec Tests', () => {
         metadata: { requires: { mongodb: '>=8.0.0', serverless: 'forbid' } },
         async test() {
           const namespace = `db.${'c'.repeat(maxMessageSizeBytes)}`;
-          const model: AnyClientBulkWriteModel<Document> = {
+          const model: ClientBulkWriteModel<Document> = {
             name: 'insertOne',
             namespace: namespace,
             document: { a: 'b' }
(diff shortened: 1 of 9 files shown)


test/integration/crud/crud.prose.test.ts
```diff
@@ -188,12 +189,12 @@ describe('CRUD Prose Spec Tests', () => {
       client.on('commandStarted', filterForCommands('bulkWrite', commands));
       commands.length = 0;
 
-      Array.from({ length: maxWriteBatchSize + 1 }, () => {
-        models.push({
+      models = Array.from({ length: maxWriteBatchSize + 1 }, () => {
```
durran, line 192: These `Array.from` changes are unrelated to the TS changes, just making them from a suggestion on the previous PR for the other test.

nbbeeken:
evergreen refresh

nbbeeken (commented):
I forgot this is my PR.. LGTM! @durran

test/types/client_bulk_write.test-d.ts
```diff
@@ -0,0 +1,258 @@
...
+  type WithoutId
+} from '../mongodb';
+
+declare const client: MongoClient;
+type Book = { title: string; released: Date };
+type Author = { name: string; published: number };
+type Store = { _id: UUID };
+
+// Baseline check that schema modifies the following fields for each type.
+expectType<ClientInsertOneModel<Book>['document']>(null as unknown as OptionalId<Book>);
```
baileympearson, line 26: `expectType` has the signature of `expectType<T>(expression: T)` - I think these are asserting backwards (as if we did `expect(expected).to.deep.equal(actual)`).  It took me a second to realize these are asserting that `ClientInsertOneModel<Book>['Document']` is of type `OptionalId<Book>`.

Usually we'd do something like:

```typescript
{ // check that the schema modifies the following fields for each type
  declare const model: ClientInsertOneModel<Book>;
  expectType<OptionalId<Book>(model.document);
  ...
}
```

test/types/client_bulk_write.test-d.ts
```diff
@@ -0,0 +1,258 @@
...
+    namespace: 'db.authors',
+    name: 'updateMany',
+    filter: { name: 'bob', published: 2 },
+    update: {}
+  },
+  { namespace: 'db.authors', name: 'deleteOne', filter: {} },
+  { namespace: 'db.authors', name: 'deleteMany', filter: {} }
+]);
+
+// No schemas - incorrect use - random namespaces, no type checking
```
baileympearson, line 75:
```suggestion
// No schemas - random namespaces, no type checking
```

if we expect this to compile, its still correct usage.  it just isn't getting type safety

test/types/client_bulk_write.test-d.ts
```diff
@@ -0,0 +1,258 @@
...
+  document: { name: 'bob', published: 2 }
+});
+
+type MongoDBSchemas = {
+  'db.books': Book;
+  'db.authors': Author;
+  'db.stores': Store;
+};
+
+expectError<ClientBulkWriteModel<MongoDBSchemas>>({
```
baileympearson, line 117: I thought we wanted this to compile but to fallback to no type safety.

test/types/client_bulk_write.test-d.ts
```diff
@@ -0,0 +1,258 @@
...
+    filter: { name: 'bob', published: 2 },
+    update: {}
+  },
+  { namespace: 'db.authors', name: 'deleteOne', filter: {} },
+  { namespace: 'db.authors', name: 'deleteMany', filter: {} }
+]);
+
+// Defined schemas - incorrect use
+expectError(
+  client.bulkWrite<MongoDBSchemas>([
```
baileympearson, line 166: I don't think this works - won't this fail if only one of the incorrect usages is causing an error?  I think each "incorrect" case needs a separate expectError

test/types/client_bulk_write.test-d.ts
```diff
@@ -0,0 +1,258 @@
...
+      update: {}
+    },
+    {
+      namespace: 'db.authors',
+      name: 'updateManyy', // unknown operation
+      filter: { name: 'bob', published: 2 },
+      update: {}
+    },
+    { namespace: 'db.authors', name: 'deleteOne', filter: {} },
+    { namespace: 'db.authors', name: 'deleteMany', filter: {} }
```
baileympearson, line 191: Why are these here?  These are correct

test/types/client_bulk_write.test-d.ts
```diff
@@ -0,0 +1,258 @@
...
+  ])
+);
+
+type MongoDBSchemasWithCalculations = {
+  // River Books uses star ratings
+  [key: `river-books.${string}`]: Book & { fiveStarRatings: number };
+  // Ocean literature uses thumbs up for ratings
+  [key: `ocean-literature.${string}`]: Book & { thumbsUp: number };
+};
```
baileympearson, line 200: What is the use case for dynamically-created namespaces?

baileympearson (changes requested):

test/types/client_bulk_write.test-d.ts
nbbeeken: Extra copy pasta, removed

test/types/client_bulk_write.test-d.ts
nbbeeken: It works but yea good catch we wouldn't catch the individual parts of this failing

test/types/client_bulk_write.test-d.ts
nbbeeken: This code snippet is kinda the use case, it is common to create namespaces that reuse the same type so you can avoid actually writing them all out and just define the prefix

test/types/client_bulk_write.test-d.ts
baileympearson: If the keys are strings, can we just fallback to Document?  

If I'm understanding correctly, the current API requires users to either completely strictly type their `bulkWrite` api (map all namespaces to all schemas) or default to document - there's no in between.

test/types/client_bulk_write.test-d.ts
baileympearson: Usually you'd just use the [polymorphic pattern](https://www.mongodb.com/developer/products/mongodb/polymorphic-pattern/) for something like this.

Sharing schemas is still possible, just orthogonal to dynamically computed namespaces:

```typescript
type MongoDBSchemasWithCalculations = {
  // River Books uses star ratings
  [key: `river-books.b`]: Book & { fiveStarRatings: number };
  // Ocean literature uses thumbs up for ratings
  [key: `ocean-literature.b`]: Book & { thumbsUp: number };
};
```

test/types/client_bulk_write.test-d.ts
nbbeeken: What is the in-between use case you're thinking of?

If you add `[key: string]: Document` to your SchemaMap you remove type assertion but also typo checking, so you revert to the default experience

Can you write out the TS you're imagining?

test/types/client_bulk_write.test-d.ts
nbbeeken: Yea I guess I gave the wrong impression with "withCalculations" I meant type calculations not dynamic at runtime necessarily. 

Here's a more simple case, not related to data design per se:

```ts
type MongoDBSchemas = { [key: `bookstores.${string}`]: Book };
```

Now all collections in the bookstores db are using the Book schema. I don't have to write out all my collections this way.

test/types/client_bulk_write.test-d.ts
baileympearson: I'd expect users to be able to do:

```
const schema = { 
  'foo.bar': { name: string }
}

// fine
client.bulkWrite<schema>([
  { name: 'insertOne', namespace: 'foo.bar', document: { name: 'bailey' } }
])

// also fine
client.bulkWrite<schema>([
  { name: 'insertOne', namespace: 'foo.other-ns', document: { name: 'bailey' } }
])

// also fine
client.bulkWrite<schema>([
  { name: 'insertOne', namespace: 'foo.bar', document: { name: 'bailey' } },
  { name: 'insertOne', namespace: 'foo.other-ns', document: { name: 'bailey' } }
])

// error - type checking on the supplied namespaces
client.bulkWrite<schema>([
  { name: 'insertOne', namespace: 'foo.bar', document: { name: 3 } }, // error
  { name: 'insertOne', namespace: 'foo.other-ns', document: { name: 'bailey' } }
])
```

nbbeeken: Thanks for the example, but I was wondering what the implementation for that would look like? That behavior would remove checking for namespace typos right?

durran:
evergreen retry

test/types/client_bulk_write.test-d.ts
nbbeeken: From standup: checking the namespaces for typos is incompatible with making plain strings map to Document, so I don't think we want to do this. 

I did look into this though, fell short of solution, I was toying around with deploying something like this:
```ts
type IsStringLiteral<T> = T extends string
  ? string extends T
    ? false
    : true
  : false;
```

But I think Namespace is always a string literal since it comes from keyof

baileympearson (approved):

PR #4262, opened 2024-10-01
feat(NODE-6338): implement client bulk write error handling

durran:
### Description

Implements error handling for `MongoClient#bulkWrite` operations.

#### What is changing?
- Unskips remaining unified tests around error handling and retries.
- Syncs retryable-writes spec tests
- Implements remaining CRUD prose tests around error handling.
- Implements all required error handling per the bulk write spec.

None

NODE-6338


src/operations/client_bulk_write/common.ts
```diff
@@ -181,6 +182,55 @@ export interface ClientBulkWriteResult {
...
+}
+
+/**
+ * An error indicating that an error occurred when executing the bulk write.
+ *
+ * @public
+ * @category Error
+ */
+export class MongoClientBulkWriteError extends MongoServerError {
+  /**
```
baileympearson, line 197: Thoughts on moving this to `errors.ts` so all of our bulk write eerrors live together?  Or moving the other bulk write error from errors.ts into this file?  Either way.

src/operations/client_bulk_write/executor.ts
```diff
@@ -65,15 +71,42 @@ export class ClientBulkWriteExecutor {
...
+          await resultsMerger.merge(cursor);
+        } catch (error) {
+          // Write concern errors are recorded in the writeConcernErrors field on MongoClientBulkWriteError.
+          // When a write concern error is encountered, it should not terminate execution of the bulk write
+          // for either ordered or unordered bulk writes. However, drivers MUST throw an exception at the end
+          // of execution if any write concern errors were observed.
+          if (error instanceof MongoServerError && !(error instanceof MongoClientBulkWriteError)) {
+            // Server side errors need to be wrapped inside a MongoClientBulkWriteError, where the root
+            // cause is the error property and a partial result is to be included.
+            const bulkWriteError = new MongoClientBulkWriteError({
```
baileympearson, line 86: Thoughts on using `cause` here instead of the `error` property? https://mongodb.slack.com/archives/C035ZJL6CQN/p1727975432135529

src/error.ts
```diff
@@ -643,6 +643,33 @@ export class MongoClientBulkWriteCursorError extends MongoRuntimeError {
   }
 }
 
+/**
+ * An error indicating that an error occurred when generating a bulk write update.
+ *
+ * @public
+ * @category Error
+ */
+export class MongoClientBulkWriteUpdateError extends MongoRuntimeError {
```
baileympearson, line 652: Seems like a MongoAPIError, since we throw this when users provide invalid update filters to the client bulk write.

src/operations/client_bulk_write/executor.ts
```diff
@@ -31,9 +33,13 @@ export class ClientBulkWriteExecutor {
...
     options?: ClientBulkWriteOptions
   ) {
+    if (operations.length === 0) {
+      throw new MongoClientBulkWriteExecutionError('No client bulk write models were provided.');
+    }
+
     this.client = client;
     this.operations = operations;
-    this.options = { ...options };
+    this.options = { ordered: true, ...options };
```
baileympearson, line 42: Do we need to apply defaults for bypassDocumentValidation or verboseResults too?

src/operations/client_bulk_write/results_merger.ts
```diff
@@ -84,18 +81,84 @@ export class ClientBulkWriteResultsMerger {
...
+            this.result.updateResults?.set(document.idx + this.currentBatchOffset, result);
           }
           // Handle delete results.
           if ('delete' in operation) {
-            this.result.deleteResults?.set(document.idx + currentBatchOffset, {
+            this.result.deleteResults?.set(document.idx + this.currentBatchOffset, {
               deletedCount: document.n
             });
           }
+        } else {
```
baileympearson, line 92: I think the if statement above needs to be adjusted - otherwise we'll enter the `else` block every time for any top-level error we receive where verbose results is false.

I think what we want is:
```typescript
if (document.ok === 1) {
	if (this.options.verboseResults) {
		...
	}
} else {
	
}

```

baileympearson (changes requested):

src/operations/client_bulk_write/results_merger.ts
durran: Updated, good catch.

src/operations/client_bulk_write/executor.ts
durran: Add defaults for both.

src/error.ts
durran: Changed uses of this to MongoAPIError and removed this error.

src/operations/client_bulk_write/executor.ts
durran: Updated to use cause, since it's idiomatic JS.

src/operations/client_bulk_write/common.ts
durran: Moved to error.ts

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -293,6 +337,22 @@ export const buildUpdateManyOperation = (
   return createUpdateOperation(model, index, true);
 };
 
+/**
+ * Validate the update document.
+ * @param update - The update document.
+ */
+function validateUpdate(update: Document) {
+  const keys = Object.keys(update);
+  if (keys.length === 0) {
```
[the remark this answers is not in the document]
durran: Updated both cases to use the helper.

baileympearson (changes requested):

src/operations/client_bulk_write/client_bulk_write.ts
```diff
@@ -27,6 +27,14 @@ export class ClientBulkWriteOperation extends CommandOperation<ClientBulkWriteCu
     this.ns = new MongoDBNamespace('admin', '$cmd');
   }
 
+  override resetBatch(): boolean {
+    return this.commandBuilder.resetBatch();
+  }
+
+  override get canRetryWrite(): boolean {
+    return this.commandBuilder.isRetryable && super.canRetryWrite;
```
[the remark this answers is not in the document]
durran: Removed the super call.

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -301,6 +375,20 @@ function createUpdateOperation(
...
 ): ClientUpdateOperation {
+  // Update documents provided in UpdateOne and UpdateMany write models are
+  // required only to contain atomic modifiers (i.e. keys that start with "$").
+  // Drivers MUST throw an error if an update document is empty or if the
+  // document's first key does not start with "$".
+  if (Array.isArray(model.update)) {
+    if (model.update.length === 0) {
+      throw new MongoAPIError('Client bulk write update model pipelines may not be empty.');
+    }
+    for (const update of model.update) {
```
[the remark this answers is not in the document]
durran: That's correct - I updated to remove the loop.

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -179,6 +222,25 @@ export class ClientBulkWriteCommandBuilder {
...
+  maxBsonObjectSize: number,
+  maxMessageSizeBytes: number
+) {
+  if (buffer.length > maxBsonObjectSize) {
+    throw new MongoInvalidArgumentError(
+      `Client bulk write operation ${name} of length ${buffer.length} exceeds the max bson object size of ${maxBsonObjectSize}`
+    );
+  }
+
+  if (buffer.length > maxMessageSizeBytes) {
```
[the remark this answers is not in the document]
durran: This is for the validation checks of individual ops not the total size for the prose tests. The total size checks happen on lines 145 and 182.

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -39,7 +40,9 @@ export class ClientBulkWriteCommandBuilder {
   options: ClientBulkWriteOptions;
   pkFactory: PkFactory;
   currentModelIndex: number;
+  previousModelIndex: number;
   lastOperations: Document[];
+  isRetryable: boolean;
```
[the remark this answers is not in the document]
durran: Comments added.

src/operations/client_bulk_write/client_bulk_write.ts
```diff
@@ -45,10 +53,13 @@ export class ClientBulkWriteOperation extends CommandOperation<ClientBulkWriteCu
         const connection = await server.pool.checkOut();
```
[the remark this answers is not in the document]
durran: Good catch, I've moved the checkout now.

src/operations/client_bulk_write/client_bulk_write.ts
```diff
@@ -59,23 +70,40 @@ export class ClientBulkWriteOperation extends CommandOperation<ClientBulkWriteCu
...
+        server.description.maxWriteBatchSize,
+        server.description.maxBsonObjectSize
       );
     }
-    return await super.executeCommand(server, session, command, ClientBulkWriteCursorResponse);
+
+    try {
+      return await super.executeCommand(server, session, command, ClientBulkWriteCursorResponse);
+    } finally {
+      if (server.description.type === ServerType.LoadBalancer) {
```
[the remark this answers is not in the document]
durran: Another good catch - I misread the spec there so now we are only pinned from initial bulkWrite for the batch to the end of cursor iteration.

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -54,7 +57,12 @@ export class ClientBulkWriteCommandBuilder {
     this.options = options;
     this.pkFactory = pkFactory ?? DEFAULT_PK_FACTORY;
     this.currentModelIndex = 0;
+    this.previousModelIndex = 0;
     this.lastOperations = [];
+    // Multi updates are not retryable.
+    this.isRetryable = !models.some(model => {
```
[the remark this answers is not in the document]
durran: You are correct. I've moved this determination to happen now when the batch is getting processed and changed the flag to note this in its new name `isBatchRetryable`.

src/operations/client_bulk_write/command_builder.ts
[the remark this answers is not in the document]
baileympearson: What I'm wondering is if `maxMessageSizeBytes` is ever less than or equal to `maxBSONObjectSize`.  My gut says that `maxMessageSizeBytes` is strictly greater than `maxBSONObjectSize` because OP_MSG is, in the simplest case, a few bytes of overhead + a BSON document and the the document's length is always less than or equal to `maxBSONObjectSize`.

If that's the case, then we'll never enter this if statement because `buffer.length > maxMessageSizeBytes > maxBSONObjectSize`, so we'd throw from the first if-statement.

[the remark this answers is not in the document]
durran: Oh apologies I was quite clear on what you meant before, but I do now. I think your statement is correct in that I think it's impossible for `maxMessageSizeBytes` to ever be less than or equal to `maxBSONObjectSize` when using OP_MSG or any part of our wire protocol. This is just defensive programming in case in the future for some strange reason that could happen, and the spec prose tests explicitly called both out. I've removed the second check as I think you were suggesting to do.

src/operations/client_bulk_write/results_merger.ts
```diff
@@ -47,55 +56,126 @@ export class ClientBulkWriteResultsMerger {
...
+        code: writeConcernError.code,
+        message: writeConcernError.errmsg
+      });
     }
 
     return this.result;
   }
+
+  /**
+   * Process an invididual document in the results.
```
nbbeeken, line 136:
```suggestion
   * Process an individual document in the results.
```

src/operations/client_bulk_write/command_builder.ts
```diff
@@ -76,27 +85,59 @@ export class ClientBulkWriteCommandBuilder {
...
         const operation = buildOperation(model, nsIndex, this.pkFactory);
-        const operationBuffer = BSON.serialize(operation);
+        let operationBuffer;
+        try {
+          operationBuffer = BSON.serialize(operation);
+        } catch (error) {
+          throw new MongoInvalidArgumentError(
+            `Could not serialize operation to BSON: ${error.message}.`
+          );
+        }
```
nbbeeken, line 138:
```suggestion
        try {
          operationBuffer = BSON.serialize(operation);
        } catch (cause) {
          throw new MongoInvalidArgumentError(
            `Could not serialize operation to BSON`, { cause }
          );
        }
```

Same for the other try catching around BSON, I think keeping the stack trace in the error is worth having because this error should be so rare such that we want the most information when it occurs.

nbbeeken (changes requested):

baileympearson (dismissed):

src/operations/client_bulk_write/results_merger.ts
durran: Fixed.

src/operations/client_bulk_write/command_builder.ts
durran: `MongoInvalidArgumentError`, which we had decided last week was the appropriate error for this case, only takes a string message in its constructor. This is why I was including the original message in it and not providing a `cause` option. Is your suggestion to create a new error that provide a cause option or modify the existing error to accept it?

nbbeeken: I would modify the existing one, we've added cause to errors as we've encountered the need, its parent MongoAPIError already supports it so only one level of changes is needed here.

durran: Updated InvalidArgumentError to accept a cause and changed the 2 usages here.

nbbeeken (approved):

PR #4264, opened 2024-10-02
refactor(NODE-6411): AbstractCursor accepts an external timeout context

baileympearson:
#### What is changing?

The abstract cursor now accepts an external timeout.  This lets operations which create cursors share their timeout contexts with the abstract cursor.  This required a handful of changes:

1. this can only be used in CURSOR_LIFETIME, since ITERATION mutates the timeout context after each operation
2. killcursors now refreshes the timeout context without mutating it using the new `refreshed` method

I also removed some usages of `isCSOTEnabled()` by adding a `maxTimeMS` property to the TimeoutContext.  We can iterate on this and remove `omitMaxTimeMS` too if we want by making cursors conditionally return `null` when they should not set maxTimeMS, but I considered this out of scope for this PR.

### DIFF
--- a/src/cursor/abstract_cursor.ts
+++ b/src/cursor/abstract_cursor.ts
@@ -21,7 +21,7 @@ import { ReadPreference, type ReadPreferenceLike } from '../read_preference';
 import { type AsyncDisposable, configureResourceManagement } from '../resource_management';
 import type { Server } from '../sdam/server';
 import { ClientSession, maybeClearPinnedConnection } from '../sessions';
-import { TimeoutContext } from '../timeout';
+import { type CSOTTimeoutContext, type Timeout, TimeoutContext } from '../timeout';
 import { type MongoDBNamespace, squashError } from '../utils';
 
 /**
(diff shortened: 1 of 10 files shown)


src/cursor/abstract_cursor.ts
```diff
@@ -205,20 +213,12 @@ export abstract class AbstractCursor<
...
-          this.cursorOptions.timeoutMode = CursorTimeoutMode.LIFETIME;
-        }
-      } else {
-        if (options.tailable && this.cursorOptions.timeoutMode === CursorTimeoutMode.LIFETIME) {
-          throw new MongoInvalidArgumentError(
-            "Cannot set tailable cursor's timeoutMode to LIFETIME"
-          );
-        }
-        this.cursorOptions.timeoutMode = options.timeoutMode;
+      if (options.tailable && this.cursorOptions.timeoutMode === CursorTimeoutMode.LIFETIME) {
```
baileympearson, line 216: Just a bit of cleanup, happy to revert.

W-A-James (approved):

durran (approved):

PR #4267, opened 2024-10-08
fix(NODE-6418): change stream resumes infinitely after failed aggregates

W-A-James:
#### What is changing?
* `changeStream` iteration methods now throw when the `aggregate` command that establishes the change stream on the server side fails

### Change stream infinite resume bug fixed
Before this fix, when change streams would fail to establish a cursor on the server, the driver would infinitely attempt to resume the change stream. Now, when the aggregate to establish the change stream fails, the driver will throw an error and clos the change stream.

### DIFF
--- a/test/integration/change-streams/change_stream.test.ts
+++ b/test/integration/change-streams/change_stream.test.ts
@@ -2061,6 +2061,34 @@ describe('ChangeStream resumability', function () {
           }
         );
       });
+
+      context('when the error occurs on the aggregate command', function () {
+        it(
+          'does not resume',
+          { requires: { topology: '!single', mongodb: '>=4.2' } },
+          async function () {
+            const resumableErrorCode = 7; // Host not found
+            await client.db('admin').command({
+              configureFailPoint: 'failCommand',
+              mode: { times: 2 }, // Fail twice to account for retry attempt in executeOperation which is separate from the change stream's resume attempt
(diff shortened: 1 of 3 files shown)


test/integration/change-streams/change_stream.test.ts
```diff
@@ -2551,6 +2635,40 @@ describe('ChangeStream resumability', function () {
...
+                failCommands: ['aggregate'],
+                errorCode: resumableErrorCode
+              }
+            } as FailPoint);
+
+            try {
+              // eslint-disable-next-line @typescript-eslint/no-unused-vars
+              for await (const change of changeStream) {
+                expect.fail('Change stream produced events on an unresumable error');
+              }
```
nbbeeken, line 2663:
```suggestion
              for await (const change of changeStream);
              expect.fail('Change stream produced events on an unresumable error');              
```

Right? just to catch a case where no iteration and no error happens?

nbbeeken (changes requested):

test/integration/change-streams/change_stream.test.ts
W-A-James: The intention here was to throw an error when a change stream is _successfully_ iterable after an unresumable error, but throwing on the case where it doesn't iterate is something that was missing here.

PR #4269, opened 2024-10-08
feat(NODE-6391): Add timeoutMS support to explicit encryption

aditi-khare-mongoDB:
#### What is changing?
Add timeoutMS support to 
- ClientEncryptionOptions
- the following ClientEncryption helper functions:
  - encrypt
  - decrypt
  - createDataKey
  - createEncryptedCollection

- Implemented spec prose tests for the above (createDataKey, encrypt, decrypt)
- Add integration test for createEncryptedCollection and CSOT
- Add unit tests for ClientEncryption getting a timeoutMS value from client and/or ClientEncryptionOptions

During release

Spec Compliance


test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -216,9 +275,42 @@ describe('CSOT spec prose tests', function () {
...
+
+        const err = await clientEncryption
+          .encrypt('hello', {
+            algorithm: `AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic`,
+            keyId: datakeyId
+          })
+          .catch(e => e);
+
+        expect(err).to.be.instanceOf(MongoOperationTimeoutError);
+        expect(commandStarted[0]).to.containSubset({ commandName: 'find' });
```
W-A-James, line 309: The prose test says to verify that we executed a find against the `keyvault.datakeys`, you can check that it executed against the right collection by checking `commandStarted[0].command.find`

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -192,6 +229,28 @@ describe('CSOT spec prose tests', function () {
...
+              failCommands: ['insert'],
+              blockConnection: true,
+              blockTimeMS: 150
+            }
+          } as FailPoint);
+        const commandStarted: CommandStartedEvent[] = [];
+        keyVaultClient.on('commandStarted', ev => commandStarted.push(ev));
+        const err = await clientEncryption.createDataKey('local').catch(e => e);
+        expect(err).to.be.instanceOf(MongoOperationTimeoutError);
+        expect(commandStarted[0]).to.containSubset({ commandName: 'insert' });
```
W-A-James, line 252: The prose test says to verify that we executed an insert against the keyvault.datakeys, you can check that it executed against the right collection by checking `commandStarted[0].command.insert`

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -216,9 +275,42 @@ describe('CSOT spec prose tests', function () {
...
+          })
+          .catch(e => e);
+
+        expect(err).to.be.instanceOf(MongoOperationTimeoutError);
+        expect(commandStarted[0]).to.containSubset({ commandName: 'find' });
+      });
     });
 
-    context('decrypt', () => {
+    context('decrypt', clientEncryptionMetadata, () => {
```
W-A-James, line 313: Putting the client metadata here doesn't prevent it from running in environments it shouldn;t, you need it in the `it` call

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -243,6 +335,40 @@ describe('CSOT spec prose tests', function () {
...
        */
+      it('times out due to timeoutMS', async function () {
+        const datakeyId = await clientEncryption.createDataKey('local');
+        expect(datakeyId).to.be.instanceOf(Binary);
+        expect(datakeyId.sub_type).to.equal(Binary.SUBTYPE_UUID);
+
+        const encrypted = Binary.createFromBase64(
+          'Af6ie/LRP0uoisAZthHPUs0CKzTBFIkJr8kxmOk1pV1C/6K54otT8QvNJgNTNG2CNpThhfdXaObuOMMReNlTgwapqPYCb/HJRQ1Nfma6uA3cTg==',
+          6
+        );
```
W-A-James, line 346: The prose test (on the [spec](https://github.com/mongodb/specifications/tree/master/source/client-side-operations-timeout/tests#decrypt) at least) says to encrypt the value 'hello'. I assume this is the precomputed encrypted value? Any reason not to just compute it here?

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -243,6 +335,40 @@ describe('CSOT spec prose tests', function () {
...
+              blockTimeMS: 150
+            }
+          } as FailPoint);
+
+        const commandStarted: CommandStartedEvent[] = [];
+        keyVaultClient.on('commandStarted', ev => commandStarted.push(ev));
+
+        const err = await clientEncryption.decrypt(encrypted).catch(e => e);
+        expect(err).to.be.instanceOf(MongoOperationTimeoutError);
+        expect(commandStarted[0]).to.containSubset({ commandName: 'find' });
```
W-A-James, line 370: Same as other comments regarding checking the command was executed against the correct namespace

test/unit/client-side-encryption/client_encryption.test.ts
```diff
@@ -158,7 +201,7 @@ describe('ClientEncryption', function () {
 
         expect(createDataKeySpy.callCount).to.equal(0);
         const options = createCollectionSpy.getCall(0).args[1];
-        expect(options).to.deep.equal({ encryptedFields: { fields: 'not an array' } });
+        expect(options).to.containSubset({ encryptedFields: { fields: 'not an array' } });
```
W-A-James, line 204: I assume this was changed to accommodate a new option? Was this the timeoutContext?

test/unit/client-side-encryption/client_encryption.test.ts
```diff
@@ -175,7 +218,7 @@ describe('ClientEncryption', function () {
 
         expect(createDataKeyStub.callCount).to.equal(1);
         const options = createCollectionSpy.getCall(0).args[1];
-        expect(options).to.deep.equal({
+        expect(options).to.containSubset({
```
W-A-James, line 221: Same question as above

W-A-James (changes requested):

test/unit/client-side-encryption/client_encryption.test.ts
aditi-khare-mongoDB: Yeah, rather than containSubset I can also just add in timeoutContext

test/unit/client-side-encryption/client_encryption.test.ts
aditi-khare-mongoDB: Same here

src/client-side-encryption/client_encryption.ts
```diff
@@ -224,7 +238,10 @@ export class ClientEncryption {
     const { insertedId } = await this._keyVaultClient
       .db(dbName)
       .collection<DataKey>(collectionName)
-      .insertOne(dataKey, { writeConcern: { w: 'majority' } });
+      .insertOne(dataKey, {
+        writeConcern: { w: 'majority' },
+        timeoutMS: timeoutContext?.csotEnabled() ? timeoutContext?.remainingTimeMS : undefined
+      });
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Made the change! Thanks for the pointer

src/client-side-encryption/client_encryption.ts
```diff
@@ -568,7 +593,8 @@ export class ClientEncryption {
     try {
       const collection = await db.createCollection<TSchema>(name, {
         ...createCollectionOptions,
-        encryptedFields
+        encryptedFields,
+        timeoutMS: timeoutContext?.remainingTimeMS
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: yeah that makes sense, made the change!

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -471,3 +476,85 @@ describe('Range Explicit Encryption with JS native types', function () {
...
+
+      afterEach(async function () {
+        await internalClient
+          .db()
+          .admin()
+          .command({
+            configureFailPoint: 'failCommand',
+            mode: 'off'
+          } as FailPoint);
+        internalClient.db('db').collection('newnew').drop();
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: good point, made the change!

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -172,6 +173,42 @@ describe('CSOT spec prose tests', function () {
...
+    afterEach(async function () {
+      await internalClient
+        .db()
+        .admin()
+        .command({
+          configureFailPoint: 'failCommand',
+          mode: 'off'
+        } as FailPoint);
+      await keyVaultClient.close();
+    });
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Good catch!

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
aditi-khare-mongoDB: I'll add a comment describing why. Essentially, if we call decrypt right after encrypt, some data is cached and the find command that we have a failPoint on might not run - resulting in no delay and no MongoOperationTimeoutError.

src/client-side-encryption/client_encryption.ts
```diff
@@ -541,13 +559,20 @@ export class ClientEncryption {
...
+      : undefined;
+
     if (Array.isArray(encryptedFields.fields)) {
       const createDataKeyPromises = encryptedFields.fields.map(async field =>
         field == null || typeof field !== 'object' || field.keyId != null
           ? field
           : {
               ...field,
-              keyId: await this.createDataKey(provider, { masterKey })
+              keyId: await this.createDataKey(provider, { masterKey, timeoutContext })
```
W-A-James, line 575: Bailey pointed out to me that this may be an issue since we're using the same timeout context concurrently, which is not something I think our timeoutContext can do. Something like this to `CSOTTimeoutContext` in  src/timeout.ts:

```ts
  clone(): CSOTTimeoutContext {
    const timeoutContext = new CSOTTimeoutContext({
      timeoutMS: this.timeoutMS,
      serverSelectionTimeoutMS: this.serverSelectionTimeoutMS
    });
    timeoutContext.start = this.start;

    return timeoutContext;
  }

```

And setting the timeoutContext being used in createDataKey here to `timeoutContext.clone()`

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
aditi-khare-mongoDB: Changed the location!

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
aditi-khare-mongoDB: Sounds good!

src/client-side-encryption/client_encryption.ts
aditi-khare-mongoDB: I see, I added in the .clone() function and call so there's no concurrent timeoutContexts being used!

src/client-side-encryption/client_encryption.ts
```diff
@@ -541,13 +561,23 @@ export class ClientEncryption {
...
     if (Array.isArray(encryptedFields.fields)) {
       const createDataKeyPromises = encryptedFields.fields.map(async field =>
         field == null || typeof field !== 'object' || field.keyId != null
           ? field
           : {
               ...field,
-              keyId: await this.createDataKey(provider, { masterKey })
+              keyId: await this.createDataKey(provider, {
+                masterKey,
+                timeoutContext: timeoutContext?.clone()
```
W-A-James, line 579: Can we add a comment here explaining that we're cloning the timeoutContext to avoid concurrent use of timeout contexts which would lead to sharing the same timeout for server selection and connection checkout across different operations.

src/timeout.ts
```diff
@@ -323,6 +323,15 @@ export class CSOTTimeoutContext extends TimeoutContext {
     return remainingTimeMS;
   }
 
+  clone(): CSOTTimeoutContext {
```
W-A-James, line 326: Can we also add a doc comment here explaining that this is intended to be used in situations where we need to perform concurrent operations but operating on the same deadline, but cannot share a single TimeoutContext.

W-A-James (changes requested):

src/client-side-encryption/client_encryption.ts
aditi-khare-mongoDB: Added it in!

src/client-side-encryption/client_encryption.ts
```diff
@@ -576,6 +576,8 @@ export class ClientEncryption {
               ...field,
               keyId: await this.createDataKey(provider, {
                 masterKey,
+                // clone the timeoutContext
+                // in order to avoid sharing the same timeout for server selection and connection checkout across different operations
```
W-A-James, line 580:
```suggestion
                // clone the timeoutContext
                // in order to avoid sharing the same timeout for server selection and connection checkout across different concurrent operations
```

src/timeout.ts
```diff
@@ -323,6 +323,11 @@ export class CSOTTimeoutContext extends TimeoutContext {
     return remainingTimeMS;
   }
 
+  /**
+   * @internal
+   * This function is intended to be used in situations where concurrent operation are on the same deadline, but cannot share a single `TimeoutContext`.
+   * Returns a deep copy of `this`.
+   */
```
W-A-James, line 330:
```suggestion
  /**
   * @internal
   * This method is intended to be used in situations where concurrent operation are on the same deadline, but cannot share a single `TimeoutContext` instance.
   * Returns a new instance of `CSOTTimeoutContext` constructed with identical options, but setting the `start` property to `this.start`.
   */
```

W-A-James (changes requested):

W-A-James (approved):

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -471,3 +476,89 @@ describe('Range Explicit Encryption with JS native types', function () {
...
   });
 });
+
+describe('CSOT', function () {
+  describe('Explicit Encryption', function () {
+    describe('#createEncryptedCollection', function () {
+      let keyVaultClient: MongoClient;
+      let internalClient: MongoClient;
+      let clientEncryption: ClientEncryption;
+      const LOCAL_KEY = Buffer.from(
```
baileympearson, line 486: Can we obtain the LOCAL_KEY from the env?  It is available in all CI environments where we run csfle tests , in the `CSFLE_KMS_PROVIDERS` environment variable.

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -471,3 +476,89 @@ describe('Range Explicit Encryption with JS native types', function () {
...
+        'times out due to timeoutMS',
+        {
+          requires: {
+            clientSideEncryption: true,
+            mongodb: '>=7.0.0',
+            topology: '!single'
+          }
+        },
+        async function () {
+          const createCollectionOptions = {
```
baileympearson, line 543: For composite operations (operations made up of multiple things that can time out), it is important that we test that

- timeoutMS applies to all operations cumulatively.  In this case, it needs to apply to all `createDataKey` invocations as well as in addition to the final createCollection
- timeoutMS applies to each async operation

We usually test this by

- making each async operation time out, confirming that each async op respects timeoutMS
- setting failcommands on all async operations such that in total they block for longer than `timeoutMS`.  In this case, that could look like a fail command on insert and createCollection, blocking for 500MS each, with timeoutMS of like 800ms.  Neither operation times out individually, but collectively they time out.

Finally, its important to measure the duration and confirm it fails in roughly the amount of time we expect (timeoutMS).

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -471,3 +476,89 @@ describe('Range Explicit Encryption with JS native types', function () {
...
+          const err = await clientEncryption
+            .createEncryptedCollection(db, 'newnew', {
+              provider: 'local',
+              createCollectionOptions,
+              masterKey: null
+            })
+            .catch(err => err);
+
+          expect(err).to.be.instanceOf(MongoCryptCreateEncryptedCollectionError);
+          expect(err.message).to.contain(
```
baileympearson, line 557: We usually test for `instanceof` - here I think you'd want to check for `(err.cause).to.be.instanceOf(...)`.

test/integration/client-side-encryption/driver.test.ts
baileympearson: (this applies to all tests)

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
baileympearson: Metadata still here 🙂 

(also, fwiw I _think_ metadata is inherited from a context/describe into any tests that are direct children of the context/describe.  but not any deeper)

src/client-side-encryption/client_encryption.ts
```diff
@@ -653,7 +679,13 @@ export class ClientEncryption {
       socketOptions: autoSelectSocketOptions(this._client.options)
     });
 
-    const { v } = deserialize(await stateMachine.execute(this, context));
+    const timeoutContext = this._timeoutMS
+      ? new CSOTTimeoutContext({
+          timeoutMS: this._timeoutMS,
+          serverSelectionTimeoutMS: this._client.options.serverSelectionTimeoutMS
+        })
+      : undefined;
```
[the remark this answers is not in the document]
baileympearson: If we do this, there's no need to resolve `timeoutMS` from the client in the constructor - because we'll obtain it from the client when we resolve options when instantiating contexts.

baileympearson (changes requested):

baileympearson (approved):

W-A-James (approved):

PR #4273, opened 2024-10-10
fix(NODE-6412): read stale response from previously timed out connection

nbbeeken:
#### What is changing?

- Close a connection that has been timed out

No

If the connection is not closed it will go back into the pool and a subsequent operation will read the reply to the previous operation

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -747,9 +747,12 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       }
     } catch (readError) {
       if (TimeoutError.is(readError)) {
-        throw new MongoOperationTimeoutError(
+        const error = new MongoOperationTimeoutError(
           `Timed out during socket read (${readError.duration}ms)`
         );
+        this.dataEvents = null;
+        this.onError(error);
+        throw error;
       }
       throw readError;
     } finally {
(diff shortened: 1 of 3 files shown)


src/cmap/connection.ts
```diff
@@ -747,9 +747,12 @@ export class Connection extends TypedEventEmitter<ConnectionEvents> {
       }
     } catch (readError) {
       if (TimeoutError.is(readError)) {
-        throw new MongoOperationTimeoutError(
+        const error = new MongoOperationTimeoutError(
           `Timed out during socket read (${readError.duration}ms)`
         );
+        this.dataEvents = null;
```
W-A-James, line 753: Does this pretty much drop all packets that have been received but not read?

W-A-James (approved):

src/cmap/connection.ts
W-A-James: Paraphrasing response from Slack
> No, I am setting it to null because the iterator has already thrown the error, connection.cleanup will call `dataEvents.throw(error)` if it exists, which it has already done because it was the one that threw
. This is just preventing us from doing the work twice.
 Typically errors come from the socket, so dataEvents.throw(e) is how you get the for-await loop to "unblock"
. It's like you're inserting the error into the iterator, this new code path is coming out of the iterator, and entering clean up, so needs a new "null it out" line of code.

PR #4276, opened 2024-10-14
feat(NODE-6409): new errors for unacknowledged bulk writes

durran:
### Description

Updates client bulk write to change the way it handles unacknowledged writes.

#### What is changing?
- Error is thrown when w:0 and ordered is true.
- Error is thrown when w:0 and verbose results are requested.
- Syncs unified tests to latest.
- Adds a new prose test for batching unacknowledged writes.

None

NODE-6409/DRIVERS-2993

https://github.com/mongodb/specifications/pull/1670

### DIFF
--- a/test/integration/crud/crud.prose.test.ts
+++ b/test/integration/crud/crud.prose.test.ts
@@ -738,7 +738,8 @@ describe('CRUD Prose Spec Tests', () => {
         async test() {
           const error = await client
             .bulkWrite([{ name: 'insertOne', namespace: 'db.coll', document: document }], {
-              writeConcern: { w: 0 }
+              writeConcern: { w: 0 },
+              ordered: false
             })
             .catch(error => error);
           expect(error.message).to.include('Client bulk write operation ops of length');
@@ -763,7 +764,7 @@ describe('CRUD Prose Spec Tests', () => {
           const error = await client
             .bulkWrite(
               [{ name: 'replaceOne', namespace: 'db.coll', filter: {}, replacement: document }],
-              { writeConcern: { w: 0 } }
+              { writeConcern: { w: 0 }, ordered: false }
             )
             .catch(error => error);
           expect(error.message).to.include('Client bulk write operation ops of length');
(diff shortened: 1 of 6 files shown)


test/integration/crud/crud.prose.test.ts
```diff
@@ -1079,4 +1080,85 @@ describe('CRUD Prose Spec Tests', () => {
...
+    const models: AnyClientBulkWriteModel[] = [];
+    const commands: CommandStartedEvent[] = [];
+
+    beforeEach(async function () {
+      const uri = this.configuration.url({
+        useMultipleMongoses: false
+      });
+      client = this.configuration.newClient(uri, { monitorCommands: true });
+      await client.connect();
+      await client.db('db').collection('coll').drop();
```
nbbeeken, line 1125:
```suggestion
      await client.db('db').collection('coll').drop().catch(() => null);
```

This will bug us later if you run this test without this ns existing it will throw (unfortunately), also I don't think lint will like my exact suggestion

test/integration/crud/crud.prose.test.ts
```diff
@@ -1079,4 +1080,85 @@ describe('CRUD Prose Spec Tests', () => {
...
+      await client.db('db').createCollection('coll');
+      const hello = await client.db('admin').command({ hello: 1 });
+      maxBsonObjectSize = hello.maxBsonObjectSize;
+      maxMessageSizeBytes = hello.maxMessageSizeBytes;
+      numModels = Math.floor(maxMessageSizeBytes / maxBsonObjectSize) + 1;
+      Array.from({ length: numModels }, () => {
+        models.push({
```
nbbeeken, line 1132: A little mixture of approaches here IIUC we're using the Array.from to loop, to push into an array that is scoped to the describe block, I am not seeing the array be reset, granted there's only one test but maybe this should've returned the array to overwrite the variable for each test?

test/integration/crud/crud.prose.test.ts
```diff
@@ -1079,4 +1080,85 @@ describe('CRUD Prose Spec Tests', () => {
...
+      await client.close();
+    });
+
+    it('performs all writes unacknowledged', {
+      metadata: { requires: { mongodb: '>=8.0.0', serverless: 'forbid' } },
+      async test() {
+        const result = await client.bulkWrite(models, { ordered: false, writeConcern: { w: 0 } });
+        expect(result).to.deep.equal({ ok: 1 });
+        expect(commands.length).to.equal(2);
+        expect(commands[0].command.ops.length).to.equal(numModels - 1);
```
nbbeeken, line 1155: > If the driver exposes `operationId`s in its CommandStartedEvents, assert that `firstEvent.operationId` is equal to `secondEvent.operationId`

Is this our `requestId`? or is this something bulk specific?

test/integration/crud/crud.prose.test.ts
```diff
@@ -1079,4 +1080,85 @@ describe('CRUD Prose Spec Tests', () => {
...
+        expect(commands[0].command.writeConcern.w).to.equal(0);
+        expect(commands[1].command.ops.length).to.equal(1);
+        expect(commands[1].command.writeConcern.w).to.equal(0);
+        const count = await client.db('db').collection('coll').countDocuments();
+        expect(count).to.equal(3);
```
nbbeeken, line 1160: > To force completion of the `w:0` writes, execute `coll.countDocuments` and expect the returned count is `maxMessageSizeBytes / maxBsonObjectSize + 1`.

should be this instead of `3` right?

nbbeeken (changes requested):

test/integration/crud/crud.prose.test.ts
durran: It is bulk specific and separate from requestId in the command monitoring spec. It's optional in the spec and we don't have it in any bulk implementation.

test/integration/crud/crud.prose.test.ts
durran: Updated to use `numModels` which is that calculation.

test/integration/crud/crud.prose.test.ts
durran: Updated to set the array instead of pushing to it.

test/integration/crud/crud.prose.test.ts
durran: Updated to suggestion, lint didn't complain.

nbbeeken (approved):
LGTM

PR #4278, opened 2024-10-14
chore: fix a few flaky CSOT tests

baileympearson:
#### What is changing?

I've noticed a number of flaky CSOT tests.  This PR attempts to fix the flakiness.

### DIFF
--- a/test/integration/client-side-operations-timeout/node_csot.test.ts
+++ b/test/integration/client-side-operations-timeout/node_csot.test.ts
@@ -649,7 +649,7 @@ describe('CSOT driver tests', metadata, () => {
       client = this.configuration.newClient(undefined, { monitorCommands: true, minPoolSize });
       commandStarted = [];
       client.on('commandStarted', ev => commandStarted.push(ev));
-      await client.connect();
+      await waitUntilPoolsFilled(client, AbortSignal.timeout(30_000), minPoolSize);
     });
 
     afterEach(async function () {
@@ -690,11 +690,13 @@ describe('CSOT driver tests', metadata, () => {
             .db('db')
             .collection('coll')
             .find({}, { timeoutMS: 150, tailable: true, awaitData: true, batchSize: 1 });
-          for (let i = 0; i < 5; i++) {
-            // Iterate cursor 5 times (server would have blocked for 500ms overall, but client
-            // should not throw
-            await cursor.next();
-          }
+          // Iterate cursor 5 times (server would have blocked for 500ms overall, but client
+          // should not throw
+          await cursor.next();
+          await cursor.next();
+          await cursor.next();
+          await cursor.next();
+          await cursor.next();
         });
 
         it('does not use timeoutMS to compute maxTimeMS for getMores', metadata, async function () {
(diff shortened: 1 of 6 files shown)


test/integration/client-side-operations-timeout/node_csot.test.ts
```diff
@@ -457,17 +457,17 @@ describe('CSOT driver tests', metadata, () => {
...
 
             expect(commandStarted).to.have.length.gte(3); // Find and 2 getMores
+
             expect(
               commandStarted.filter(ev => {
                 return (
-                  ev.command.find != null &&
-                  ev.command.getMore != null &&
-                  ev.command.maxTimeMS != null
+                  (ev.command.find != null && ev.command.maxTimeMS != null) ||
```
baileympearson, line 469: I think the previous condition was impossible, because it wanted both a find and a getMore in the same command.  So it was always true.

test/integration/node-specific/abstract_cursor.test.ts
```diff
@@ -473,39 +477,59 @@ describe('class AbstractCursor', function () {
...
+      afterEach(async function () {
+        await clearFailPoint(this.configuration, uri);
+      });
 
+      it(
+        'the provided timeoutContext is not modified',
+        {
+          requires: {
+            mongodb: '>=4.4',
+            topology: '!load-balanced'
```
baileympearson, line 512: No kill cursors in lb mode.

test/integration/node-specific/abstract_cursor.test.ts
```diff
@@ -473,39 +477,59 @@ describe('class AbstractCursor', function () {
...
             {
               timeoutContext: context,
-              timeoutMS: 1000,
+              timeoutMS: 150,
               timeoutMode: CursorTimeoutMode.LIFETIME,
               batchSize: 1
             }
           );
 
+          const refresh = sinon.spy(context, 'refresh');
```
baileympearson, line 526: Checking that `remainingTimeoutMS < 0` was flaky.  I don't particularly like this approach, because it relies on testing the current implementation, but this does fix the flakiness.

test/integration/client-side-operations-timeout/node_csot.test.ts
nbbeeken: If it returns the right result this is fine, but still seems like a confusing condition. If you feel similarly maybe we should break this up, 

const finds = filter finds;
const getMores = filter getMores;
expect finds do not have maxTimeMS
expect getMores do not have maxTimeMS

nbbeeken (approved):

nbbeeken:
```
                             Client Bulk Write
[2024/10/15 16:20:50.617]        CSOT enabled
[2024/10/15 16:20:50.617]          acknowledged writes
[2024/10/15 16:20:50.617]            when the timeout is reached while iterating the result cursor
[2024/10/15 16:20:50.617]              the bulk write operation times out:
[2024/10/15 16:20:50.617]      AssertionError: expected 1499 to be within 1500..2300
```

off by 1! should we expand the range?

baileympearson:
evergreen retry

baileympearson:
https://github.com/mongodb/node-mongodb-native/pull/4278#issuecomment-2417167872

That's not behavior I'd ever expect - the killCursors is so slow that it usually takes an extra 5-7 hundred ms.  I think we should leave this as-is, and if it continues to flake we should look into why it runs so much more quickly sometimes.  That might be a bug in our CSOT logic.

nbbeeken (approved):
SGTM, just seeing if I can clear the flakes, but will merge soon.

PR #4280, opened 2024-10-15
docs(NODE-6223): timeoutMS does not govern auto-connect

nbbeeken:
#### What is changing?

Added docs

New docs for docs? interesting idea

Auto connect or plain connect performs file system access, DNS lookup, monitor creation, mongocryptd launching, and more! each one of these steps has its own mechanism for cancellation that is not considered by the CSOT spec. Alternatively we considered ignoring cancelling those operations and simply racing with a timeout, however, the impact could be squashing important errors. 

Auto connect needs a future rework to better align with the drivers "always connected" perspective, for now, its best to explicitly connect if there's concern that the connecting operation won't complete in a globally set timeoutMS value.

#### Is there anything that the user can set that will bound the execution time of autoconnect in this case?

Nothing specific to auto-connect, any client options: `socketTimeoutMS`, `waitQueueTimeoutMS`, `serverSelectionTimeoutMS`, `connectTimeoutMS` are all general options that control those components of the driver for all operations. Auto-connect or just connect itself has its own set of operations that cannot be timed out at all, reading TLS files, SRV/TXT DNS lookup, monitoring start-up, server selection, and a ping. Also, auto-encryption can have an internal client that repeats the aforementioned as well.

Generally, folks observe a `MongoServerSelectionError` when the ping forces a server selection that the monitors fail to service in time. That cannot be set for only connect and only covers a common case.

### DIFF
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -47,7 +47,6 @@ import { readPreferenceServerSelector } from './sdam/server_selection';
 import type { SrvPoller } from './sdam/srv_polling';
 import { Topology, type TopologyEvents } from './sdam/topology';
 import { ClientSession, type ClientSessionOptions, ServerSessionPool } from './sessions';
-import { Timeout } from './timeout';
 import {
   COSMOS_DB_CHECK,
   COSMOS_DB_MSG,
@@ -528,19 +527,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
       return await this.connectionLock;
     }
 
-    const timeoutMS = this[kOptions].timeoutMS;
-
-    this.connectionLock = this._connect().finally(() => {
-      // Clear the lock only when this promise finishes:
+    try {
+      this.connectionLock = this._connect();
+      await this.connectionLock;
+    } finally {
+      // release
       this.connectionLock = undefined;
-    });
-
-    await (timeoutMS == null
-      ? this.connectionLock
-      : Promise.race([
-          this.connectionLock,
-          Timeout.expires(timeoutMS).catch(Timeout.convertError)
-        ]));
+    }
 
     return this;
   }
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -47,6 +47,7 @@ import { readPreferenceServerSelector } from './sdam/server_selection';
 import type { SrvPoller } from './sdam/srv_polling';
 import { Topology, type TopologyEvents } from './sdam/topology';
 import { ClientSession, type ClientSessionOptions, ServerSessionPool } from './sessions';
+import { Timeout } from './timeout';
 import {
   COSMOS_DB_CHECK,
   COSMOS_DB_MSG,
@@ -520,13 +521,19 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
       return await this.connectionLock;
     }
 
-    try {
-      this.connectionLock = this._connect();
-      await this.connectionLock;
-    } finally {
-      // release
+    const timeoutMS = this[kOptions].timeoutMS;
+
+    this.connectionLock = this._connect().finally(() => {
+      // Clear the lock only when this promise finishes:
       this.connectionLock = undefined;
-    }
+    });
+
+    await (timeoutMS == null
+      ? this.connectionLock
+      : Promise.race([
+          this.connectionLock,
+          Timeout.expires(timeoutMS).catch(Timeout.convertError)
+        ]));
 
     return this;
   }
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -514,6 +514,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
   /**
    * Connect to MongoDB using a url
    *
+   * @remarks
+   * Calling connect is optional as the first operation you preform will call connect if it's needed.
+   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
+   * However, when the operation being run is automatically connecting your MongoClient the timeoutMS will only be used for the operation portion of task.
+   * This means the time to setup the MongoClient does not count against timeoutMS.
+   * If you are using timeoutMS we recommend connecting your client explicitly in advance of any operation to avoid this inconsistent execution time.
+   *
    * @see docs.mongodb.org/manual/reference/connection-string/
    */
   async connect(): Promise<this> {
@@ -715,6 +722,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
 
   /**
    * Connect to MongoDB using a url
+   *
+   * @remarks
+   * Calling connect is optional as the first operation you preform will call connect if it's needed.
+   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
+   * However, when the operation being run is automatically connecting your MongoClient the timeoutMS will only be used for the operation portion of task.
+   * This means the time to setup the MongoClient does not count against timeoutMS.
+   * If you are using timeoutMS we recommend connecting your client explicitly in advance of any operation to avoid this inconsistent execution time.
    *
    * @remarks
    * The programmatically provided options take precedence over the URI options.
--- a/src/mongo_client.ts
+++ b/src/mongo_client.ts
@@ -515,8 +515,8 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
    *
    * @remarks
    * Calling `connect` is optional since the first operation you perform will call `connect` if it's needed.
-   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
-   * However, when the operation being run is automatically connecting your `MongoClient` the `timeoutMS` will only be used for the operation portion of task.
+   * `timeoutMS` will bound the time any operation can take before throwing a timeout error.
+   * However, when the operation being run is automatically connecting your `MongoClient` the `timeoutMS` will not apply to the time taken to connect the MongoClient.
    * This means the time to setup the `MongoClient` does not count against `timeoutMS`.
    * If you are using `timeoutMS` we recommend connecting your client explicitly in advance of any operation to avoid this inconsistent execution time.
    *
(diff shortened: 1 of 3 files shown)


nbbeeken:
Happy to take the tests out or adjust them, they were left over from when I started to implement connect timeout.

src/mongo_client.ts
```diff
@@ -706,6 +713,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
    * Connect to MongoDB using a url
    *
    * @remarks
+   * Calling connect is optional as the first operation you preform will call connect if it's needed.
+   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
+   * However, when the operation being run is automatically connecting your MongoClient the timeoutMS will only be used for the operation portion of task.
+   * This means the time to setup the MongoClient does not count against timeoutMS.
+   * If you are using timeoutMS we recommend connecting your client explicitly in advance of any operation to avoid this inconsistent execution time.
+   *
+   * @remarks
```
aditi-khare-mongoDB, line 722:
```suggestion
   * Calling `connect` is optional since the first operation you perform will call `connect` if it's needed.
   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
   * However, when the operation being run is automatically connecting your `MongoClient` the timeoutMS will only be used for the operation portion of task.
   * This means the time to setup the `MongoClient` does not count against `timeoutMS`.
   * If you are using `timeoutMS` we recommend connecting your client explicitly in advance of any operation to avoid this inconsistent execution time.
   *
   * @remarks
```

src/mongo_client.ts
```diff
@@ -509,6 +509,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
   /**
    * Connect to MongoDB using a url
    *
+   * @remarks
+   * Calling connect is optional as the first operation you preform will call connect if it's needed.
+   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
+   * However, when the operation being run is automatically connecting your MongoClient the timeoutMS will only be used for the operation portion of task.
+   * This means the time to setup the MongoClient does not count against timeoutMS.
+   * If you are using timeoutMS we recommend connecting your client explicitly in advance of any operation to avoid this inconsistent execution time.
+   *
```
aditi-khare-mongoDB, line 518:
```suggestion
   * @remarks
   * Calling `connect` is optional since the first operation you perform will call `connect` if it's needed.
   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
   * However, when the operation being run is automatically connecting your `MongoClient` the timeoutMS will only be used for the operation portion of task.
   * This means the time to setup the `MongoClient` does not count against `timeoutMS`.
   * If you are using `timeoutMS` we recommend connecting your client explicitly in advance of any operation to avoid this inconsistent execution time.
   *
```

aditi-khare-mongoDB (changes requested):

aditi-khare-mongoDB (approved):

src/mongo_client.ts
```diff
@@ -706,6 +713,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
    * Connect to MongoDB using a url
    *
    * @remarks
+   * Calling `connect` is optional since the first operation you perform will call `connect` if it's needed.
+   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
+   * However, when the operation being run is automatically connecting your `MongoClient` the timeoutMS will only be used for the operation portion of task.
```
dariakp, line 718:
```suggestion
   * However, when the operation being run is automatically connecting your `MongoClient` the `timeoutMS` will only be used for the operation portion of task.
```

src/mongo_client.ts
```diff
@@ -509,6 +509,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
   /**
    * Connect to MongoDB using a url
    *
+   * @remarks
+   * Calling `connect` is optional since the first operation you perform will call `connect` if it's needed.
+   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
+   * However, when the operation being run is automatically connecting your `MongoClient` the timeoutMS will only be used for the operation portion of task.
```
dariakp, line 515:
```suggestion
   * However, when the operation being run is automatically connecting your `MongoClient` the `timeoutMS` will only be used for the operation portion of task.
```

src/mongo_client.ts
```diff
@@ -509,6 +509,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
...
+   * This means the time to setup the `MongoClient` does not count against `timeoutMS`.
+   * If you are using `timeoutMS` we recommend connecting your client explicitly in advance of any operation to avoid this inconsistent execution time.
```
dariakp, line 517: Is there anything that the user can set that will bound the execution time of autoconnect in this case?

dariakp (changes requested):

src/mongo_client.ts
nbbeeken: Nothing specific to auto-connect, any client options: socketTimeoutMS, waitQueueTimeoutMS, serverSelectionTimeoutMS, connectTimeoutMS are all general options that control those components of the driver for all operations. Auto-connect or just connect itself has its own set of operations that cannot be timed out at all, reading TLS files, SRV/TXT DNS lookup, monitoring start-up, server selection, and a ping. Also, auto-encryption can have an internal client that repeats the aforementioned as well.

Generally, folks observe a MongoServerSelectionError when the ping forces a server selection that the monitors fail to service in time. That cannot be set for only connect and only covers a common case.

src/mongo_client.ts
```diff
@@ -509,6 +509,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
   /**
    * Connect to MongoDB using a url
    *
+   * @remarks
+   * Calling `connect` is optional since the first operation you perform will call `connect` if it's needed.
+   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
```
baileympearson, line 514:
```suggestion
   * `timeoutMS` will bound the time any operation can take before throwing a timeout error.
```

This is true, regardless of where timoeutMS is is specified.

src/mongo_client.ts
```diff
@@ -509,6 +509,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
...
+   * However, when the operation being run is automatically connecting your `MongoClient` the `timeoutMS` will only be used for the operation portion of task.
```
baileympearson, line 515:
```suggestion
   * However, when the operation being run is automatically connecting your `MongoClient` the `timeoutMS` will not apply to the time taken to connect the MongoClient.
```

I'd surmise that many users don't have a conception of the difference between "operation portion of the task" and "connect" - I think it's easier to just say that timeoutMS won't apply to the connecting logic.

src/mongo_client.ts
```diff
@@ -706,6 +713,13 @@ export class MongoClient extends TypedEventEmitter<MongoClientEvents> implements
    * Connect to MongoDB using a url
    *
    * @remarks
+   * Calling `connect` is optional since the first operation you perform will call `connect` if it's needed.
+   * `timeoutMS` specified at the client-level will bound the time any operation can take before throwing a timeout error.
```
baileympearson, line 717: (same comments as above to this doc comment too).

baileympearson (changes requested):

src/mongo_client.ts
nbbeeken: done!

src/mongo_client.ts
nbbeeken: done!

src/mongo_client.ts
dariakp: Can we capture this in the PR description and/or in Jira for posterity? @nbbeeken Otherwise good to resolve

test/integration/node-specific/auto_connect.test.ts
```diff
@@ -821,4 +823,80 @@ describe('When executing an operation for the first time', () => {
...
+
+    describe('when nothing is wrong', function () {
+      it('connects the client', async function () {
+        await client.connect();
+        expect(client).to.have.property('topology').that.is.instanceOf(Topology);
+      });
+    });
+
+    describe(
+      'when the server requires auth and ping is delayed',
```
baileympearson, line 846: This is slightly nit-picky, but since we don't care which parts of auto connect hang, could we simply just stub `connect()` and delay it longer than timeoutMS?

nbbeeken: I do agree, but since I need to keep the fail point around for the explicit connect (stubbing in that test would remove any coverage of connect accidentally respecting `timeoutMS`, you know something that is so easy to do 😅) seems easier to keep them in sync than to cover it in two different ways. Does that seem okay to you?

src/mongo_client.ts
nbbeeken: Captured in the PR body and filed: https://jira.mongodb.org/browse/NODE-6448 as well as updated the title of this ticket in jira to be accurate

test/integration/node-specific/auto_connect.test.ts
baileympearson: I'm not sure I follow.  You're saying that test additionally is giving us coverage that connect() does not support timeoutMS?  If we want to test that connect doesn't support timeoutMS, then we should probably test that separately.

nbbeeken: That is what the test is testing, when the ping is delayed: client.connect() takes as long as ping is delayed for and does not throw a timeout error

It is checking that the executeOperation portion of connect which is where we handle applying a timeoutMS to an operation doesn't mistakenly pull in and apply the client's timeoutMS.

baileympearson: I think we're missing a test that demonstrates:

```typescript
describe('when CSOT is enabled', function() {
  describe('when auto connect is used and connect() takes longer than timeoutMS', function() {
    it('the operation succeeds', async function() {});
  });
})
```

Or is this tested elsewhere?

baileympearson: I guess this is implicitly tested in the last test block in this file.  Was that the intended CSOT + auto connect test?

nbbeeken: Yes it was but we can have both, it would be good to have coverage that is agnostic of the connect implementation but also include the known ping that has a risk of using timeoutMS

nbbeeken: Added!

test/integration/node-specific/auto_connect.test.ts
```diff
@@ -891,12 +891,36 @@ describe('When executing an operation for the first time', () => {
...
           expect(res).to.have.property('acknowledged', true);
-          expect(end - start).to.be.within(2000, 2500); // timeoutMS is 1000, did not apply.
+          expect(end - start).to.be.within(1000, 1500); // timeoutMS is 1000, did not apply.
         });
       }
     );
+
+    describe('when auto connect is used and connect() takes longer than timeoutMS', function () {
+      // This test stubs the connect method to check that connect() does not get timed out
+      // vs. the test above makes sure that the `ping` does not inherit the client's timeoutMS setting
```
baileympearson, line 904: Can we make this clear with the test descriptions instead of a comment?  It seems like the second sentence of this comment is the actual intent of the previous test

```typescript
describe('when auto connect is used', async function() {
  it('timeoutMS from the client is not used for the internal `ping`', ... {} );
});
```

nbbeeken: updated!

baileympearson (approved):

PR #4281, opened 2024-10-16
feat(NODE-6392): add timeoutMS support to ClientEncryption helpers part 1

nbbeeken:
#### What is changing?

- Adds timeoutMS support to CE helpers

No

The timeoutMS setting given to CE will now be passed to the various helpers.


test/integration/client-side-encryption/driver.test.ts
```diff
@@ -1,20 +1,32 @@
...
+import { type FailPoint, getEncryptExtraOptions } from '../../tools/utils';
+
+const metadata: MongoDBMetadataUI = {
   requires: {
     mongodb: '>=4.2.0',
     clientSideEncryption: true
   }
 };
 
+const LOCAL_KEY = Buffer.from(
```
baileympearson, line 25: This is only used for CSOT tests.  Can we move this key to the appropriate describe block?

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +413,176 @@ describe('Client Side Encryption Functional', function () {
...
+          configureFailPoint: 'failCommand',
+          mode: { times: 2 },
+          data: {
+            failCommands: [command],
+            blockConnection: true,
+            blockTimeMS,
+            appName: 'clientEncryption'
+          }
+        } as FailPoint);
+        await utilClient.close();
```
baileympearson, line 434: We now have helpers for this, if you're interested:

```suggestion
        await configureFailPoint(this.configuration, {
          configureFailPoint: 'failCommand',
          mode: { times: 2 },
          data: {
            failCommands: [command],
            blockConnection: true,
            blockTimeMS,
            appName: 'clientEncryption'
          }
        });
```

(same for clearfailpoint)

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +413,176 @@ describe('Client Side Encryption Functional', function () {
...
+        }
+        expect(end - start).to.be.within(498, 1000);
+      };
+    }
+
+    let client: MongoClient;
+    let clientEncryption: ClientEncryption;
+
+    beforeEach(async function () {
+      if (!this.configuration.clientSideEncryption.enabled) {
```
baileympearson, line 470: Why not use `{ requires: { clientEncryption: 'enabled' }}`?

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +413,176 @@ describe('Client Side Encryption Functional', function () {
       });
     }
   );
+
```
baileympearson, line 416: There are some CSOT prose tests I think we need to implement too: https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/tests/README.md#3-clientencryption

src/client-side-encryption/client_encryption.ts
```diff
@@ -120,6 +123,7 @@ export class ClientEncryption {
     this._proxyOptions = options.proxyOptions ?? {};
     this._tlsOptions = options.tlsOptions ?? {};
     this._kmsProviders = options.kmsProviders || {};
+    this._timeoutMS = options.timeoutMS ?? client.options.timeoutMS;
```
baileympearson, line 126:
```suggestion
    const { timeoutMS } = resolveTimeoutOptions(client, options);
    this._timeoutMS = timeoutMS;
```

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +413,176 @@ describe('Client Side Encryption Functional', function () {
...
+        ...getEncryptExtraOptions()
+      });
+    });
+
+    afterEach(async function () {
+      await client?.close();
+    });
+
+    describe('rewrapManyDataKey', function () {
+      makeBlockingFailFor('update', 2000);
```
baileympearson, line 490: we’re missing a tests that cover: 
- timeoutMS is respected by the state machine
- timeoutMS applies to the state machine.execute() and bulkWrite collectively

We can probably test this by putting a failCommand on both a find and bulkWrite, such that the total blockTimeMS > timeoutMS.  Neither would fail individually, but it fails if timeoutMS is shared across both.

baileympearson (changes requested):

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +410,227 @@ describe('Client Side Encryption Functional', function () {
...
+            await clientEncryption.rewrapManyDataKey({ _id: key1Id }, { provider: 'local' });
+          })
+        );
+      });
+
+      // TOOD, does not make rewriteManyDataKey fail
+      describe.skip('when the find and bulk operation takes too long', function () {
+        // together they add up to 800, exceeding the timeout of 500
+        makeBlockingFailFor('update', 400);
+        makeBlockingFailFor('listCollections', 400);
```
baileympearson, line 550: there can only be a single failCommand active at a time.  If you need to fail multiple commands, specify both in the failCommands array:

```typescript
        makeBlockingFailFor(['update', 'listCollections'], 400);
```

Perhaps this is why your test isn't failing?

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
```diff
@@ -216,9 +289,35 @@ describe('CSOT spec prose tests', function () {
        *   - Expect this to fail with a timeout error.
        * 1. Verify that a `find` command was executed against the `keyvault.datakeys` collection as part of the `encrypt` call.
        */
+
+      afterEach(async function () {
+        await clearFailPoint(this.configuration);
+      });
+
+      it('throws a timeout error', metadata, async function () {
```
baileympearson, line 297: Did you mean to implement this prose test?  You didn't make the encrypt() helper respect timeoutMS.  I'm surprised this is passing 🤔

baileympearson: Same with decrypt.

baileympearson (changes requested):
Just some test questions!

src/client-side-encryption/auto_encrypter.ts
```diff
@@ -310,7 +310,10 @@ export class AutoEncrypter {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore: TS complains as this always returns true on versions where it is present.
       if (net.getDefaultAutoSelectFamily) {
-        Object.assign(clientOptions, autoSelectSocketOptions(this._client.options));
+        // AutoEncrypter is made inside of MongoClient constructor while options are being parsed,
+        // we do not have access to the options that are in progress.
+        // TODO(NODE-NODE-6449): AutoEncrypter does not use client options for autoSelectFamily
+        Object.assign(clientOptions, autoSelectSocketOptions(this._client.s?.options ?? {}));
```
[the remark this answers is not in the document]
nbbeeken: Unlike the BSON import which is just a 'nice-to-have' single location for all BSON interactions, the options getter on the client has real cost spreading and freezing on every access, we fixed a ~10% performance regression when we reverted its introduction to the Session constructor, so it is worth correcting it where we notice it. 

I do agree though we should be able to remove the s.options and change the getter to just storing the options directly on the client without breaking anything.

This specific case is a bug: https://jira.mongodb.org/browse/NODE-6449

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +410,227 @@ describe('Client Side Encryption Functional', function () {
       });
     }
   );
+
+  describe('CSOT on ClientEncryption', { requires: { clientSideEncryption: true } }, function () {
+    const metadata: MongoDBMetadataUI = {
+      requires: { clientSideEncryption: true, mongodb: '>=4.4', topology: '!single' }
```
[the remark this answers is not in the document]
nbbeeken: I was getting an error something like ~ "standalones do not support queryable encryption" 🤷🏻 I thought that was expected

test/integration/client-side-operations-timeout/client_side_operations_timeout.prose.test.ts
nbbeeken: LOL oops... 🤔

src/client-side-encryption/auto_encrypter.ts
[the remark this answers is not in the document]
nbbeeken: I did consider making these changes to main, but most of the usages are introduced for CSOT, maybe this will close the small gap in perf we saw. fyi: @W-A-James

test/integration/client-side-encryption/driver.test.ts
nbbeeken: fixed

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +410,227 @@ describe('Client Side Encryption Functional', function () {
...
+          'throws a timeout error',
+          metadata,
+          runAndCheckForCSOTTimeout(async () => {
+            await clientEncryption.rewrapManyDataKey({ _id: key1Id }, { provider: 'local' });
+          })
+        );
+      });
+
+      // TOOD, does not make rewriteManyDataKey fail
+      describe.skip('when the find and bulk operation takes too long', function () {
```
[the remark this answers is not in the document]
nbbeeken: fixed

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +410,227 @@ describe('Client Side Encryption Functional', function () {
...
+      const db = setupClient.db('db');
+      await db
+        .collection('explicit_encryption')
+        .drop()
+        .catch(() => null);
+      await db.createCollection('explicit_encryption', { encryptedFields }).catch(error => {
+        const expected =
+          'Driver support of Queryable Encryption is incompatible with server'.toLowerCase();
+        if (!error.message.toLowerCase().includes(expected)) throw error;
+      });
```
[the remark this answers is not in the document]
nbbeeken: removed

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +410,227 @@ describe('Client Side Encryption Functional', function () {
...
+
+    let client: MongoClient;
+    let clientEncryption: ClientEncryption;
+    let key1Id;
+
+    const data = path.join(__dirname, '..', '..', 'spec', 'client-side-encryption', 'etc', 'data');
+
+    beforeEach(async function () {
+      // Load the file encryptedFields.json as encryptedFields.
+      const encryptedFields = EJSON.parse(
```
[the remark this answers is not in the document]
nbbeeken: removed

test/integration/client-side-encryption/driver.test.ts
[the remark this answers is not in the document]
nbbeeken: should be able to remove this now

src/client-side-encryption/auto_encrypter.ts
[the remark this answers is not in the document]
baileympearson: I don't understand how this fixes the bug you reported.  We still instantiate an auto encrypter before adding the options to `client.s` - all you've done is explicitly default to {}.

Unlike sessions, none of the places you fixed in this PR are hot code paths (this usage included).  I disagree that this is burning and that important.  I'm trying to say that, if you believe this _needs_ to be addressed, please file a ticket for it so we can fix it somehow and put guardrails around this so we _can't_ make this mistake.  Otherwise more usages of `client.options` will continue to pop up (for example - there are literally more occurrences of this in Aditi's PR).

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -401,6 +415,211 @@ describe('Client Side Encryption Functional', function () {
       });
     }
   );
+
+  describe.only(
```
baileympearson, line 419: .only

nbbeeken: 😭

src/client-side-encryption/auto_encrypter.ts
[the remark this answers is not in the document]
nbbeeken: > I don't understand how this fixes the bug you reported.

Because it doesn't! The ticket I filed is for us to fix the bug, I did change the code to be explicit as a reminder for anyone who drops by this line of code.

> there are literally more occurrences of this in Aditi's PR

We should change the occurrences we can catch because of what we know to be true about the cost of using the incorrect API rather than giving up any attempt to improve what we are implementing. 

I filed a ticket to fix this particular bug but at that time we can make sure we're not making the same mistake elsewhere and potentially lint or just correct the API altogether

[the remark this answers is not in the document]
baileympearson: Is there a ticket to either 1. find a tooling solution to prevent us from using `client.options` or 2. remove the freeze or 3. some other fix that ensures we don't make that mistake going forward?

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -412,6 +415,207 @@ describe('Client Side Encryption Functional', function () {
...
+    let clientEncryption: ClientEncryption;
+    let commandsStarted: CommandStartedEvent[];
+
+    beforeEach(async function () {
+      const internalClient = this.configuration.newClient();
+      await internalClient
+        .db('keyvault')
+        .dropCollection('datakeys', { writeConcern: { w: 'majority' } })
+        .catch(() => null);
+      await sleep(500);
```
baileympearson, line 480: What is the sleep necessary for?

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -412,6 +415,207 @@ describe('Client Side Encryption Functional', function () {
...
+          await clientEncryption.deleteKey(new UUID());
+        })
+      );
+    });
+
+    describe('getKey', function () {
+      makeBlockingFailFor('find', 2000);
+
+      it(
+        'throws a timeout error if the bulk operation takes too long',
```
baileympearson, line 563:
```suggestion
        'throws a timeout error if the find takes too long',
```

copy-paste error?

test/integration/client-side-encryption/driver.test.ts
```diff
@@ -526,11 +730,12 @@ describe('CSOT', function () {
     });
 
     const getKmsProviders = (): KMSProviders => {
-      const result = EJSON.parse(process.env.CSFLE_KMS_PROVIDERS || '{}') as unknown as {
-        local: unknown;
+      const { local } = EJSON.parse(process.env.CSFLE_KMS_PROVIDERS || '{}') as {
```
baileympearson, line 733: Any reason not to use `getKmsProviders` in your new tests above? You currently have hard-coded a local key but we could get the provider from the environment.

baileympearson (changes requested):
just a minor test title fix

test/integration/client-side-encryption/driver.test.ts
nbbeeken: removed

test/integration/client-side-encryption/driver.test.ts
nbbeeken: fixed

test/integration/client-side-encryption/driver.test.ts
nbbeeken: made this into a shared function

baileympearson (approved):

PR #4284, opened 2024-10-17
fix(NODE-6436): only force majority write concern on commitTransaction retry

W-A-James:
#### What is changing?

Ensure that `commitTransaction` only overrides the write concern when the `commitTransaction` call is retried (both with builtin retry mechanism and user-initiated retry).

No

Needed by NODE-3914

### `ClientSession.commitTransaction()` no longer unconditionally overrides write concern
Prior to this change, `ClientSession.commitTransaction()` would always override any previously configured `writeConcern` on the initial attempt. This overriding behaviour now only applies to internal and user-initiated retries of `ClientSession.commitTransaction()` for a given transaction.

### DIFF
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -124,6 +124,7 @@ export class ClientSession
   owner?: symbol | AbstractCursor;
   defaultTransactionOptions: TransactionOptions;
   transaction: Transaction;
+  commitAttempted?: boolean;
   /** @internal */
   [kServerSession]: ServerSession | null;
   /** @internal */
@@ -417,6 +418,7 @@ export class ClientSession
       );
     }
 
+    this.commitAttempted = false;
     // increment txnNumber
     this.incrementTransactionNumber();
     // create transaction state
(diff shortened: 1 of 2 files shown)


test/integration/transactions/transactions.test.ts
```diff
@@ -231,8 +232,14 @@ describe('Transactions', function () {
 
   context('when completing a transaction', () => {
     let client: MongoClient;
+    let commandsStarted: CommandStartedEvent[];
     beforeEach(async function () {
-      client = this.configuration.newClient();
+      client = this.configuration.newClient(undefined, { monitorCommands: true });
+      commandsStarted = [];
+      client.on('commandStarted', ev => {
+        console.log(ev);
```
durran, line 240: Can we remove this console statement?

src/sessions.ts
```diff
@@ -494,16 +499,25 @@ export class ClientSession
...
         // SPEC-1185: apply majority write concern when retrying commitTransaction
         WriteConcern.apply(command, { wtimeoutMS: 10000, ...wc, w: 'majority' });
         // per txns spec, must unpin session in this case
         this.unpin({ force: true });
 
         try {
-          await executeOperation(this.client, operation);
+          await executeOperation(
+            this.client,
+            new RunAdminCommandOperation(command, {
```
durran, line 515: Is there a reason for instantiating a new operation here? I see before we just reused the old one.

durran (changes requested):

src/sessions.ts
durran: Ah yes, that's what I missed. Thanks!

durran (approved):

PR #4288, opened 2024-10-22
fix(CWE-74) cookie accepts cookie name, path, and domain with out of bounds characters

lamcodeofpwnosec:
### Description
cookie is a basic HTTP cookie parser and serializer for HTTP servers. The cookie name could be used to set other fields of the cookie, resulting in an unexpected cookie value. A similar escape can be used for path and domain, which could be abused to alter other fields of the cookie. Upgrade to 0.7.0, which updates the validation for name, path, and domain. The cookie name could be used to set other fields of the cookie, resulting in an unexpected cookie value. For example, serialize("userName=<script>alert('XSS3')</script>; Max-Age=2592000; a", value) would result in "userName=<script>alert('XSS3')</script>; Max-Age=2592000; a=test", setting userName cookie to <script> and ignoring value.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3529,9 +3529,9 @@
       }
     },
     "node_modules/body-parser": {
-      "version": "1.20.2",
-      "resolved": "https://registry.npmjs.org/body-parser/-/body-parser-1.20.2.tgz",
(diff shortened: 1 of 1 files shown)


nbbeeken:
These are dev-only dependencies and do not impact the driver or development workflow. NODE-6426 will probably be done soon and clear up any local development warnings here.

PR #4291, opened 2024-10-24
fix(NODE-6454): use timeoutcontext for state machine execute() cursor options

baileympearson:
#### What is changing?

Instead of using the remainingTimeMS as the value for `timeoutMS` when running cursor operations in the state machine, we provide a CursorTimeoutContext that wraps the original timeout.

### DIFF
--- a/src/client-side-encryption/state_machine.ts
+++ b/src/client-side-encryption/state_machine.ts
@@ -11,6 +11,7 @@ import {
   serialize
 } from '../bson';
 import { type ProxyOptions } from '../cmap/connection';
+import { CursorTimeoutContext } from '../cursor/abstract_cursor';
 import { getSocks, type SocksLib } from '../deps';
 import { MongoOperationTimeoutError } from '../error';
 import { type MongoClient, type MongoClientOptions } from '../mongo_client';
@@ -519,16 +520,16 @@ export class StateMachine {
   ): Promise<Uint8Array | null> {
     const { db } = MongoDBCollectionNamespace.fromString(ns);
 
-    const collections = await client
-      .db(db)
-      .listCollections(filter, {
-        promoteLongs: false,
-        promoteValues: false,
-        ...(timeoutContext?.csotEnabled()
-          ? { timeoutMS: timeoutContext?.remainingTimeMS, timeoutMode: 'cursorLifetime' }
-          : {})
-      })
-      .toArray();
+    const cursor = client.db(db).listCollections(filter, {
+      promoteLongs: false,
+      promoteValues: false,
+      timeoutContext: timeoutContext && new CursorTimeoutContext(timeoutContext, Symbol())
+    });
+
+    // There is always exactly zero or one matching documents, so this should always exhaust the cursor
+    // in a single batch.  We call `toArray()` just to be safe and ensure that the cursor is always
+    // exhausted and closed.
+    const collections = await cursor.toArray();
 
     const info = collections.length > 0 ? serialize(collections[0]) : null;
     return info;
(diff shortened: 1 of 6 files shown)


src/client-side-encryption/state_machine.ts
```diff
@@ -524,9 +525,7 @@ export class StateMachine {
       .listCollections(filter, {
         promoteLongs: false,
         promoteValues: false,
-        ...(timeoutContext?.csotEnabled()
-          ? { timeoutMS: timeoutContext?.remainingTimeMS, timeoutMode: 'cursorLifetime' }
-          : {})
+        timeoutContext: timeoutContext && new CursorTimeoutContext(timeoutContext, Symbol())
```
aditi-khare-mongoDB, line 528: Neat! I didn't know you could `&&` objects together.

aditi-khare-mongoDB (approved):

aditi-khare-mongoDB:
LGTM, I'll wait for CI to finish before approving

src/client-side-encryption/state_machine.ts
baileympearson: Yup - `&&` and `||` return the first falsy/truthy value they encounter: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND

`timeoutContext` is either a timeout context or undefined, so if it is undefined, `timeoutContext && new CursorTimeoutContext(timeoutContext, Symbol())` returns `undefined`.  Otherwise, we proceed to `new CursorTimeoutContext(timeoutContext, Symbol())`, which is truthy, so it is returned.

aditi-khare-mongoDB (changes requested):
Lint is failing, I'll wait on re-reviewing until it's fixed

test/integration/crud/client_bulk_write.test.ts
```diff
@@ -278,25 +279,35 @@ describe('Client Bulk Write', function () {
...
-          const start = now();
-          const timeoutError = await client
-            .bulkWrite(models, {
-              verboseResults: true,
-              timeoutMS: 1500
-            })
-            .catch(e => e);
+        it(
+          'the bulk write operation times out',
+          mergeTestMetadata(metadata, {
```
baileympearson, line 284: This should fix some of the flakiness we see in CI.

src/client-side-encryption/state_machine.ts
```diff
@@ -519,16 +521,16 @@ export class StateMachine {
...
-      .db(db)
-      .listCollections(filter, {
-        promoteLongs: false,
-        promoteValues: false,
-        ...(timeoutContext?.csotEnabled()
-          ? { timeoutMS: timeoutContext?.remainingTimeMS, timeoutMode: 'cursorLifetime' }
-          : {})
-      })
-      .toArray();
+    const cursor = new ListCollectionsCursor(client.db(db), filter, {
```
aditi-khare-mongoDB, line 524: Doesn't calling` client.db(db).listCollections(...).toArray` have the same behavior?

test/tools/utils.ts
```diff
@@ -689,3 +689,19 @@ export async function measureDuration<T>(f: () => Promise<T>): Promise<{
...
+  return {
+    requires: {
+      ...metadata.requires,
+      ...newMetadata.requires
+    },
+    sessions: {
+      ...metadata.sessions,
+      ...newMetadata.sessions
+    }
+  };
```
aditi-khare-mongoDB, line 706: nice helper function!

aditi-khare-mongoDB (changes requested):

src/client-side-encryption/state_machine.ts
baileympearson: If we call `listCollections()`, we'd need to add logic to support `timeoutContext` as a public API option.  I can do that if you'd prefer!

aditi-khare-mongoDB: If we keep `listCollections` as is, will it support `timeoutMS` reset upon cursor refresh?

aditi-khare-mongoDB (commented):
See above.

src/client-side-encryption/state_machine.ts
baileympearson: I was looking at `resolveOptions` and thought that we needed to add `timeoutContext` support so that the option would be inherited, but I was wrong.  I reverted this change

aditi-khare-mongoDB (approved):

nbbeeken (approved):

PR #4294, opened 2024-10-28
test(NODE-6459): Add comparative perf tests

W-A-James:
#### What is changing?
Re-introduce performance tests for legacy timeout behaviour and add test when CSOT is enabled with a timeoutMS of 120000. Note that this value was chosen to be greater than zero in order to ensure that CSOT is actually constructing timeouts and timeoutContexts. The reasoning for the specific value is to allow time for longer running operations to complete.

* Introduced `makeClientWithOpts` helper for benchmarks
* encapsulated some of the logic to make setting up the different tests with different configuration options a little cleaner
* Ensured that the benchmarks that use the legacy timeout options have the same name as they do on main for continuity of results once the CSOT feature branch merges.


test/benchmarks/driverBench/index.js
```diff
@@ -35,54 +35,64 @@ const benchmarkRunner = new Runner()
...
-
-    const benchmarkResults = {
-      singleBench,
-      multiBench,
-      parallelBench,
-      readBench,
-      writeBench,
-      driverBench,
-      ...microBench.parallel,
-      ...microBench.bsonBench,
```
W-A-James, line 82: bsonBench was already removed when we migrated those tests to the js-bson repo

PR #4299, opened 2024-10-30
test(NODE-5920): do not throw if the server returns unknown saslSupportedMechs

baileympearson:
#### What is changing?
- New prose test

Spec compliance


dariakp (changes requested):
Could we move this into a stand-alone prose file and make the context title match the spec prose test and include the spec steps in comments so that it follows the format of the rest of the prose tests? (The goal is easier auditability of spec tests - to know what prose we implemented and if the prose test later gets updated, which version of the test we implemented)

test/integration/mongodb-handshake/mongodb-handshake.prose.test.ts
```diff
@@ -109,4 +109,38 @@ describe('Handshake Prose Tests', function () {
...
+        // @ts-expect-error: sinon will place wrappedMethod there
+        const command = Connection.prototype.command.wrappedMethod.bind(this);
+        if (cmd.hello || cmd[LEGACY_HELLO_COMMAND]) {
+          return stub();
+        }
+        return command(ns, cmd, options);
+
+        async function stub() {
+          const response = await command(ns, cmd, options);
+          // console.error({ response });
```
dariakp, line 126: stray debug comment?

test/integration/mongodb-handshake/mongodb-handshake.test.ts
```diff
@@ -4,6 +4,7 @@ import type Sinon from 'sinon';
 import * as sinon from 'sinon';
 
 import {
+  connect,
```
dariakp, line 7: leftover from when the tests were in this file?

dariakp (changes requested):

test/integration/mongodb-handshake/mongodb-handshake.prose.test.ts
```diff
@@ -109,4 +109,37 @@ describe('Handshake Prose Tests', function () {
...
     });
   }
+
+  context(`Test 2: Test that the driver accepts an arbitrary auth mechanism`, function () {
+    beforeEach(() => {
+      // Mock the server response in a way that saslSupportedMechs array in the hello command response contains an arbitrary string.
+      sinon.stub(Connection.prototype, 'command').callsFake(async function (ns, cmd, options) {
+        // @ts-expect-error: sinon will place wrappedMethod there
+        const command = Connection.prototype.command.wrappedMethod.bind(this);
+        if (cmd.hello || cmd[LEGACY_HELLO_COMMAND]) {
```
dariakp, line 119: Do we know which pathway this test is expected to take - the hello or the legacy hello? It occurs to me that we should probably throw an assertion into the `it` block to ensure the stub is called at least once to ensure that the condition under test actually occurs, sorry for not noticing this earlier.

dariakp (changes requested):

test/integration/mongodb-handshake/mongodb-handshake.prose.test.ts
baileympearson: Both, it depends on the configuration (server api or load balanced make us use `hello`, otherwise we use the legacy hello command).

I added a check to ensure we actually call the stub.

dariakp (approved):

PR #4303, opened 2024-10-31
chore(NODE-6488): specify target=es2021 for ts compilation tests

baileympearson:
#### What is changing?

Set target to es2021 to avoid errors related to supporting JS syntax below our lowest supported node version.

Yes, added to the readme

### DIFF
--- a/.evergreen/run-typescript.sh
+++ b/.evergreen/run-typescript.sh
@@ -44,8 +44,12 @@ if [ "$TS_CHECK" == "COMPILE_DRIVER" ]; then
     npm run build:ts
 elif [ "$TS_CHECK" == "CHECK_TYPES" ]; then
     echo "checking driver types"
+    if [ "$TS_VERSION" == "4.4" ]; then
     # check compilation
+        node $TSC mongodb.d.ts --module commonjs --target es2021
+    else
     node $TSC mongodb.d.ts --module node16 --target es2021
+    fi
 else
     "Invalid value $TS_CHECK for TS_CHECK environment variable."
     exit 1
--- a/.evergreen/run-typescript.sh
+++ b/.evergreen/run-typescript.sh
@@ -45,7 +45,7 @@ if [ "$TS_CHECK" == "COMPILE_DRIVER" ]; then
 elif [ "$TS_CHECK" == "CHECK_TYPES" ]; then
     echo "checking driver types"
     # check compilation
-    node $TSC mongodb.d.ts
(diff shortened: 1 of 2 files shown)


.evergreen/run-typescript.sh
```diff
@@ -44,8 +44,12 @@ if [ "$TS_CHECK" == "COMPILE_DRIVER" ]; then
     npm run build:ts
 elif [ "$TS_CHECK" == "CHECK_TYPES" ]; then
     echo "checking driver types"
+    if [ "$TS_VERSION" == "4.4" ]; then
     # check compilation
-    node $TSC mongodb.d.ts
+        node $TSC mongodb.d.ts --module commonjs --target es2021
```
nbbeeken, line 49: Are we forced to set a module when specifying target?

nbbeeken (changes requested):
I went to leave a bit of breadcrumb in the PR description and when I got to the "documentation?" question is made me think that maybe we should update our "Typescript Version" section with this change. We say we support 4.4 but now we're specifically saying we support 4.4 with a target later than es2021 right?

If we introduced a private field like the node types did, our tests would pass, but anyone not using this target or later would no longer be able to compile with our types, just as we weren't able to with the node type changes, right?

baileympearson:
> I went to leave a bit of breadcrumb in the PR description and when I got to the "documentation?" question is made me think that maybe we should update our "Typescript Version" section with this change. We say we support 4.4 but now we're specifically saying we support 4.4 with a target later than es2021 right?
> 
> If we introduced a private field like the node types did, our tests would pass, but anyone not using this target or later would no longer be able to compile with our types, just as we weren't able to with the node type changes, right?

Yes, but I think that is fine because that aligns with our runtime behavior.  Node16 is the first Node version that supports 100% of es2021.  We are free to use 100% of the syntax in es2021 in our codebase (including private fields).  

Happy to reconsider though.

.evergreen/run-typescript.sh
baileympearson: Yup.  By default the target is es5, which has no built-in module system so TS doesn't need a module resolution system.  Anything es6 or later must have the `-m` option specified

nbbeeken (changes requested):
You sound confident enough for me! let's just capture the new expectations in the Readme in our 'TS Version' section. 

I assume this is a pattern we will follow moving forward right? so when we bump min node version we can also bump the required target to compile?

nbbeeken (approved):

PR #4313, opened 2024-11-04
chore: use file to store data in benchmark

nbbeeken:
#### What is changing?

- Take data out of ram and write it to file
- process file externally to find median
- Skip ldjsonMultiFileUpload

No

ldjsonMultiFileUpload is failing
returnDocument produces more data points than a JS array can hold

### DIFF
--- a/test/benchmarks/driverBench/index.js
+++ b/test/benchmarks/driverBench/index.js
@@ -47,7 +47,7 @@ benchmarkRunner
     const multiBench = average(Object.values(microBench.multiBench));
 
     const parallelBench = average([
-      // microBench.parallel.ldjsonMultiFileUpload,
+      microBench.parallel.ldjsonMultiFileUpload,
       microBench.parallel.ldjsonMultiFileExport,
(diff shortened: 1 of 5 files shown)


test/benchmarks/driverBench/index.js
```diff
@@ -47,7 +47,7 @@ benchmarkRunner
     const multiBench = average(Object.values(microBench.multiBench));
 
     const parallelBench = average([
-      microBench.parallel.ldjsonMultiFileUpload,
+      microBench.parallel.ldjsonMultiFileUpload ?? 0,
```
dariakp, line 50: I worry that setting it to 0 would mess up the tracked result, could we just take it out? Or set it to the currently known average value if we're concerned about preserving the parallel measurement history?

dariakp (changes requested):

test/benchmarks/driverBench/index.js
nbbeeken: I agree, removing it makes more sense. I'd rather have accurate numbers than simulate a value here if that sounds good.

PR #4314, opened 2024-11-05
ci(NODE-6501): use latest node in spec benchmarks

baileympearson:
#### What is changing?

We use latest LTS (v22.11.0) for performance testing.  Also, the runner now has a max number of executions set to 10 million, to prevent OOM errors on the returnDocument test.


.evergreen/config.in.yml
```diff
@@ -1157,7 +1157,7 @@ tasks:
         type: setup
         params:
           updates:
-            - { key: NODE_LTS_VERSION, value: v18.16.0 }
+            - { key: NODE_LTS_VERSION, value: v22.11.0 }
             - { key: NPM_VERSION, value: "9" }
```
nbbeeken, line 1161: We don't need the npm pin anymore either

nbbeeken (changes requested):

test/benchmarks/mongoBench/runner.js
```diff
@@ -162,12 +163,17 @@ class Runner {
...
     let count = 1;
 
     const taskTimer = benchmark._taskType === 'sync' ? timeSyncTask : timeAsyncTask;
 
-    while (time < maxExecutionTime && (time < minExecutionTime || count < minExecutionCount)) {
+    while (
+      time < maxExecutionTime &&
+      (time < minExecutionTime || count < minExecutionCount) &&
+      count < maxExecutionAttempts
+    ) {
```
nbbeeken, line 176:
```js
 _errorHandler(error) {
    let currentError = error;
    while (currentError) {
      this.reporter(
        `${currentError !== error ? 'Caused by' : 'Error'}: ${currentError.name} - ${currentError.message} - ${currentError.stack}`
      );
      currentError = currentError.cause;
    }
    throw error;
  }
```

I don't feel too strongly about the print logic, it was just to remove all those ids from the bulk error, but can we make errors throw now? I don't think we should have the code "tolerate" errors and continue and ruin the entire calculation with `NaN` anymore.

test/benchmarks/mongoBench/runner.js
```diff
@@ -57,6 +57,7 @@ class Runner {
     this.minExecutionTime = options.minExecutionTime || CONSTANTS.DEFAULT_MIN_EXECUTION_TIME;
```
nbbeeken, line 57:
```js
const sortedData = [].concat(rawData).sort();
```

I cannot comment on the line directly. This does not sort the values correctly

```js
[].concat([1.2, 10.1, 2.4, 20.2]).sort();
[ 1.2, 10.1, 2.4, 20.2 ]
```
10 doesn't come before 2 😢

test/benchmarks/mongoBench/runner.js
```diff
@@ -162,12 +163,17 @@ class Runner {
     const minExecutionCount = this.minExecutionCount;
     const minExecutionTime = this.minExecutionTime;
     const maxExecutionTime = this.maxExecutionTime;
+    const maxExecutionAttempts = this.minExecutionCount;
```
nbbeeken, line 166:
```suggestion
    const maxExecutionAttempts = this.maxExecutionCount;
```

🍝

nbbeeken (changes requested):

test/benchmarks/mongoBench/runner.js
```diff
@@ -162,12 +163,17 @@ class Runner {
...
     let time = performance.now() - start;
     let count = 1;
 
     const taskTimer = benchmark._taskType === 'sync' ? timeSyncTask : timeAsyncTask;
 
-    while (time < maxExecutionTime && (time < minExecutionTime || count < minExecutionCount)) {
+    while (
```
nbbeeken, line 172: Can't comment on the line. Inside `_runSuite`:

```js
this.reporter(` Executed Benchmark "${name}" =`, result[name]);
```

nbbeeken (approved):

test/benchmarks/mongoBench/runner.js
dariakp: It does in the alphabet 😿

test/benchmarks/mongoBench/runner.js
```diff
@@ -39,7 +39,7 @@ function calculateMicroBench(benchmark, data) {
   const rawData = data.rawData;
   const count = data.count;
 
-  const sortedData = [].concat(rawData).sort();
+  const sortedData = [].concat(rawData).sort((a, b) => a - b);
```
nbbeeken, line 42: One last fix, can we calculate the median properly? average of middle two if even and the middle value if odd. (we could delete the other percentiles here, we don't use them)

nbbeeken (changes requested):

test/benchmarks/mongoBench/runner.js
nbbeeken: https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.md#:~:text=For%20each%20micro,for%20every%20driver.

nbbeeken (approved):

PR #4321, opened 2024-11-08
perf(NODE-6525): remove setPrototype and defineProperty from hot path

nbbeeken:
#### What is changing?

- Making various performance improvements to the hot paths of the driver

No

Regain performance

### Improved performance

TBD...

### DIFF
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -970,9 +970,7 @@ export class ServerSession {
   isDirty: boolean;
 
   /** @internal */
-  constructor(
-    cloned?: { id: ServerSessionId; lastUse: number; txnNumber: number; isDirty: boolean } | null
-  ) {
+  constructor(cloned?: ServerSession | null) {
     if (cloned != null) {
       const idBytes = Buffer.allocUnsafe(16);
       idBytes.set(cloned.id.id.buffer);
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -299,11 +299,8 @@ export class ClientSession
         if (serverSession != null) {
           // release the server session back to the pool
           this.sessionPool.release(serverSession);
-          // Make sure a new serverSession never makes it onto this ClientSession
-          Object.defineProperty(this, kServerSession, {
-            value: ServerSession.clone(serverSession),
-            writable: false
-          });
+          // Store a clone of the server session for reference (debugging)
+          this[kServerSession] = new ServerSession(serverSession);
         }
         // mark the session as ended, and emit a signal
         this.hasEnded = true;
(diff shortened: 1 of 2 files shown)


src/sessions.ts
```diff
@@ -994,30 +1002,6 @@ export class ServerSession {
...
-   * after ClientSession has ended
-   */
-  static clone(serverSession: ServerSession): Readonly<ServerSession> {
-    const arrayBuffer = new ArrayBuffer(16);
-    const idBytes = Buffer.from(arrayBuffer);
-    idBytes.set(serverSession.id.id.buffer);
-
-    const id = new Binary(idBytes, serverSession.id.id.sub_type);
-
-    // Manual prototype construction to avoid modifying the constructor of this class
```
nbbeeken, line 1010: The constructor is internal

src/sessions.ts
```diff
@@ -299,11 +299,8 @@ export class ClientSession
         if (serverSession != null) {
           // release the server session back to the pool
           this.sessionPool.release(serverSession);
-          // Make sure a new serverSession never makes it onto this ClientSession
-          Object.defineProperty(this, kServerSession, {
-            value: ServerSession.clone(serverSession),
-            writable: false
```
nbbeeken, line 305: This is an over abundance of caution for something we can either enforce with a flag or tolerably let be, this symbol property is internal an unlikely to be written to after this point.

src/sessions.ts
```diff
@@ -973,7 +970,18 @@ export class ServerSession {
   isDirty: boolean;
 
   /** @internal */
-  constructor() {
+  constructor(
+    cloned?: { id: ServerSessionId; lastUse: number; txnNumber: number; isDirty: boolean } | null
```
dariakp, line 974: Why type this separately vs just using the ServerSession type?

nbbeeken: Good catch, fixed.

I did this manual type b/c originally I was modifying clone, and passed in the plain object we had in there. But once I figured the helper didn't have much value I merged the logic into the ctor.

addaleax (dismissed):

src/sessions.ts
baileympearson: Can we make the property private and use a `ts-expect-error` here with a comment explaining why we're assigning the readonly property?

nbbeeken: I can make it private (it is already our implicit form of private b/c its symbol and internal)

But I don't understand the readonly suggestion since it is a property we have to write to. What's the benefit of marking it readonly only to ignore that?

baileympearson: you implemented this to make it not writable by defining it as `writable: false`.  But you're removing that restriction.

An alternative approach is to make it readonly so that future changes can't write to this value without TS errors.

nbbeeken: It was not writable after this point in time, which is different than a property being readonly upon creation. 

If I make it readonly on the class I also have to ignore the error that shows up where we actually do write to the property as intended. I think that is undesirable since expect-error ignores _all_ ts checking not only the thing we're thinking about now. Also, it is the wrong intention, it is intentionally writable because it's lazily calculated to avoid TooManyLogicalSessions errors.

W-A-James (approved):

dariakp (approved):

PR #4322, opened 2024-11-08
chore(NODE-6502) Use a python venv wherever we use Python in driver CI

blink1073:
#### What is changing?

Use scripts from drivers-evergreen-tools instead of using the system python package manager.

Tested against https://github.com/mongodb-labs/drivers-evergreen-tools/pull/545 to ensure there are no regressions in requiring Python 3.8+ syntax.

https://spruce.mongodb.com/version/672e33e060d13d0007e3cdb4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

No

Some places in Node's CI we rely on the system python/pip, which can lead to compatibility issues when we call drivers-evergreen-tools scripts and broken CI.


baileympearson (changes requested):
It might be working just because the system's Python version is compatible with the drivers-evergreen-tools version, but I do see other places where I think we need to make this change

https://github.com/mongodb/node-mongodb-native/blob/19ecf07cf418480b70d065bcaa0cc0d3ef80edc9/.evergreen/run-socks5-tests.sh#L21, for example, should be run using a python venv from drivers-evergreen-tools that has boto3 installed, right?  

There are other places hard-coded into the config file that also source this script

blink1073:
I updated and ran again against https://github.com/mongodb-labs/drivers-evergreen-tools/pull/545.

https://spruce.mongodb.com/version/672e580262625300076c1050/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

blink1073:
New build: https://spruce.mongodb.com/version/672e7b7543cf44000751a5cf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

baileympearson (approved):

PR #4324, opened 2024-11-11
chore(NODE-6502) Address two other usages of system python

blink1073:
#### What is changing?
Follow up to #4322

No

Use scripts from drivers-evergreen-tools instead of using the system python package manager.


baileympearson (approved):
Every python usage I found in our evergreen tooling on this branch either doesn't use drivers-evergreen-tools or uses a venv.

PR #4338, opened 2024-11-21
fix(NODE-6583): upgrade to BSON v6.10.1 to remove internal unbounded type cache

nbbeeken:
#### What is changing?

- ~Improve connection error messages~
- Allow the driver path, bson path, and driver settings to be configured from env vars
- Unskip: ldjsonMultiFileUpload and ldjsonMultiFileExport
- Update to fixed BSON

No

We weren't able to run our parallel benchmarks without this fix.

### BSON upgraded to v6.10.1

See: https://github.com/mongodb/js-bson/releases/tag/v6.10.1

### DIFF
--- a/test/benchmarks/driverBench/common.js
+++ b/test/benchmarks/driverBench/common.js
@@ -4,10 +4,50 @@ const fs = require('fs');
 const path = require('path');
 const { Readable } = require('stream');
 const { pipeline } = require('stream/promises');
-const { MongoClient } = require('../../..');
-const { GridFSBucket } = require('../../..');
-// eslint-disable-next-line no-restricted-modules
-const { MONGODB_ERROR_CODES } = require('../../../lib/error');
+const child_process = require('child_process');
+
+const MONGODB_DRIVER_PATH = (() => {
+  let driverPath = process.env.MONGODB_DRIVER_PATH;
+  if (!driverPath?.length) {
+    driverPath = path.resolve(__dirname, '../../..');
+  }
+  return driverPath;
+})();
+
+const { MongoClient, GridFSBucket } = require(MONGODB_DRIVER_PATH);
+
+const { version: MONGODB_DRIVER_VERSION } = require(path.join(MONGODB_DRIVER_PATH, 'package.json'));
+
+const MONGODB_DRIVER_REVISION = (() => {
+  try {
+    return child_process
+      .execSync('git rev-parse --short HEAD', {
+        cwd: MONGODB_DRIVER_PATH,
+        encoding: 'utf8'
+      })
+      .trim();
+  } catch {
+    return 'unknown revision';
+  }
+})();
+
+const MONGODB_BSON_PATH = path.join(MONGODB_DRIVER_PATH, 'node_modules', 'bson');
+const { version: MONGODB_BSON_VERSION } = require(path.join(MONGODB_BSON_PATH, 'package.json'));
+const MONGODB_BSON_REVISION = (() => {
+  if (!fs.existsSync(path.join(MONGODB_BSON_PATH, '.git'))) {
+    return 'installed from npm';
+  }
+  try {
+    return child_process
+      .execSync('git rev-parse --short HEAD', {
+        cwd: path.join(MONGODB_BSON_PATH),
+        encoding: 'utf8'
+      })
+      .trim();
+  } catch {
+    return 'unknown revision';
+  }
+})();
 
 const DB_NAME = 'perftest';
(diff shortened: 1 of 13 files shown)


test/benchmarks/driverBench/common.js
```diff
@@ -4,10 +4,72 @@ const fs = require('fs');
...
-// eslint-disable-next-line no-restricted-modules
-const { MONGODB_ERROR_CODES } = require('../../../lib/error');
+const child_process = require('child_process');
+
+/**
+ * The path to the MongoDB Node.js driver.
+ * This MUST be set to the directory the driver is installed in
+ * NOT the file "lib/index.js" that is the driver's export.
+ */
+const MONGODB_DRIVER_PATH = (() => {
```
nbbeeken, line 14: Also happy to pull out all this revision/version stuff. I think I've adequately made it fail nicely when the files/env vars don't exist. Again, it provided value once so it could again.

test/benchmarks/driverBench/common.js
```diff
@@ -67,7 +129,7 @@ function initCollection() {
 
 function dropCollection() {
   return this.collection.drop().catch(e => {
-    if (e.code !== MONGODB_ERROR_CODES.NamespaceNotFound) {
+    if (e.code !== 26 /* NamespaceNotFound */) {
```
nbbeeken, line 132: This isn't public API so testing arbitrary driver versions made this difficult when only importing at the package level.

W-A-James (changes requested):
Changes all look good to me, but I'd prefer if we were to keep the BSON version bump and associated restoring of the previously failing benchmarks in this PR and move the network error changes to a separate PR.

W-A-James (dismissed):

baileympearson (approved):

test/readme.md
```diff
@@ -241,6 +241,52 @@ beforeEach(() => {
...
+
+```sh
+npm run check:bench
+```
+
+Refer to the `run-spec-benchmark-tests-node-server` task for Node.js version, MongoDB server version, and platform that we run benchmarks against in CI.
+
+The server is run in standalone mode and the server versions are aliased by this script: https://github.com/mongodb-labs/drivers-evergreen-tools/blob/5048cca80e9ca62642409de2d401058bbd7057fa/.evergreen/mongodl.py#L58 check the latest version to see what alias the driver is running against.
+
+The host used is described here: https://spruce.mongodb.com/distro/rhel90-dbx-perf-large/settings/general (Auth required to view)
```
W-A-James, line 254: Do we want to just describe it explicitly instead of linking to it? If external users or contributors want to do testing of their own, it may be useful for them to have an idea of what specs we're working with, especially if they want to do the bulk of their own performance work and present us with changes with perf improvements, it would cut down on PR review cycles if they had those details to spin up their own perf testing setups.

test/readme.md
```diff
@@ -241,6 +241,52 @@ beforeEach(() => {
...
+
+It is best to try reproductions against as similar a deployment as possible to isolate regressions.
+
+### Configuration
+
+The benchmarks can be directed to test different settings and driver versions.
+
+The following are environment variables and how the benchmark runner uses them:
+
+- `MONGODB_DRIVER_PATH` - if set MUST be set to the directory a driver version is in, usually another clone of the driver checked out to a different revision.
```
W-A-James, line 264: I assume it uses the currently checked out version? Worth stating explicitly.

test/readme.md
```diff
@@ -241,6 +241,52 @@ beforeEach(() => {
...
+- `MONGODB_CLIENT_OPTIONS` - if set MUST be a JSON string that will be parsed and passed as the second argument to the MongoClient constructor.
+- `MONGODB_URI` - if set MUST be a valid MongoDB connection string and it will be used as the host the benchmarks will run against.
```
W-A-James, line 266: Same for these as above. User friendly to mention the default

W-A-James (approved):

W-A-James (changes requested):
See comments made previously. Accidentally hit approve before 🤦🏿

test/readme.md
nbbeeken: added, although I don't expect users to reproduce the host, but I agree might as well document it.

test/readme.md
nbbeeken: done

test/readme.md
nbbeeken: done

W-A-James (approved):

W-A-James (approved):

PR #4345, opened 2024-12-03
fix(NODE-6600): set object mode for message chunking correctly

addaleax:

### DIFF
--- a/src/cmap/connection.ts
+++ b/src/cmap/connection.ts
@@ -769,7 +769,7 @@ export class SizedMessageTransform extends Transform {
   connection: Connection;
 
   constructor({ connection }: { connection: Connection }) {
-    super({ objectMode: false });
+    super({ writableObjectMode: false, readableObjectMode: true });
     this.bufferPool = new BufferPool();
     this.connection = connection;
   }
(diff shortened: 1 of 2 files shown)


baileympearson (approved):

nbbeeken (approved):

nbbeeken (changes requested):
What are the release notes for this?

addaleax:
@nbbeeken Do you think this is worth calling out in release notes?

nbbeeken:
Yes even if it is just a sentence that points to what is being corrected without elaborating in detail we should inform what folks may have faced if this is a bug that could have occurred on this or prior versions. The versions that have this correctness issue may be in use for years and we'll want a guide that describes in human terms what the problem was as a reference.

I think this could work if it sounds accurate to you:

### Socket read stream set to object mode

Socket data was being read with a stream set to buffer mode when it should be set to object mode to prevent inaccurate data chunking, which may have caused message parsing errors in rare cases.

addaleax:
@nbbeeken Sure, that works for me 🙂

nbbeeken (approved):

